PR #138, opened 2013-06-25
Osx transparentproxy diocnatlook

nlitsme:
Hi,

( 49b626a5b437e652fcf6e23ca84249fa9731c6b2 )
I implemented transparent proxy originaladdr resolving for osx using DIOCNATLOOK.

the original lsof solution did not work for me ( on mountain lion ), it went into an infinite loop,
as if each lsof was causing more connections.

( 07eec690287c4d3f6e35017c0ba6831542771f28 )
and i also fixed a bug in the hexdump output of POST contents


cortesi:
Closing for now. We'll re-evaluate moving to DIOCNATLOOK down the track.

PR #439, opened 2015-01-06
added support for specifying multiple files for flow playback (aka scenario support first dependencies) 

gato:
Hi Max, 

As we agreed in https://github.com/mitmproxy/mitmproxy/pull/432 i'm PR you with (server|client)_replay supporting a list of files and modified _readflow to read all of them. 
I've done it in DumpMaster (dump.py) but also in cmdline.py (for mitmproxy to support it as well)
I've also fixed a bug (that i previously added), replay_ignore_payload_params was not been passed to start_server_playback when called in mitmproxy (it was ok in mitmdump) 

Regards.
Marcelo

### DIFF
--- a/libmproxy/console/__init__.py
+++ b/libmproxy/console/__init__.py
@@ -496,11 +496,13 @@ class ConsoleMaster(flow.FlowMaster):
         self.eventlog = not self.eventlog
         self.view_flowlist()
 
-    def _readflow(self, path):
-        path = os.path.expanduser(path)
+    def _readflow(self, paths):
         try:
-            f = file(path, "rb")
-            flows = list(flow.FlowReader(f).stream())
+            flows = []
+            for path in paths:
+                path = os.path.expanduser(path)
+                f = file(path, "rb")
+                flows.extend(list(flow.FlowReader(f).stream()))
         except (IOError, flow.FlowReadError), v:
             return True, v.strerror
         return False, flows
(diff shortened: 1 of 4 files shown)


libmproxy/console/__init__.py
```diff
@@ -496,11 +496,13 @@ def toggle_eventlog(self):
...
-    def _readflow(self, path):
-        path = os.path.expanduser(path)
+    def _readflow(self, paths):
         try:
-            f = file(path, "rb")
-            flows = list(flow.FlowReader(f).stream())
+            flows = []
+            for path in paths:
+                path = os.path.expanduser(path)
+                f = file(path, "rb")
```
mhils: Could you adjust this to `with open(path, "rb"):` here and below? We should make sure that we close all files.

mhils:
Hey Marcelo,

sorry for the late reply - the last week was super hectic and this PR got buried below a bunch of other mails. Looks excellent for me so far, but I think we should fix the open file handle on the way as well. :-)

Cheers,
Max

gato:
Hi Max, 
No problem.
I will change the open's behavior this afternoon and update the PR.
Regards.
Marcelo

mhils:
:+1:

gato:
done

mhils:
Hi Marcelo,

I finally get around reviewing your PRs - thanks for the great work. I just merged this one (modulo your latest merge commit, which is why it is not shown as merged here). If you encounter any duplicate code as in this case, please don't refrain from refactoring and unifying functionality. :smiley: I cleaned this whole mess (which certainly has been there before your additions) up in https://github.com/mitmproxy/mitmproxy/commit/0ac3227b7bb27a76b6f49aa1605f1b1887a01a10 .

Cheers,
Max

gato:
Hi max, 
I haven't found any issue 
Regards.
Marcelo

Issue #441, opened 2015-01-08
Mitmproxy crashes when loading a bad server replay from the terminal


nickhs:
Mitmproxy doesn't like loading (bad) client or server replays from the command line, as it fails to display the error.

Note that server_playback_path is called in `__init__` before `self.statusbar` is initialized. Loading the file via the `S` key in mitmproxy shows the error just fine.

```
$ mitmproxy -S ~/Downloads/oldreplay.capture
Traceback (most recent call last):
  File "/usr/local/bin/mitmproxy", line 3, in <module>
    mitmproxy()
  File "/usr/local/lib/python2.7/site-packages/libmproxy/main.py", line 97, in mitmproxy
    m = console.ConsoleMaster(server, console_options)
  File "/usr/local/lib/python2.7/site-packages/libmproxy/console/__init__.py", line 426, in __init__
    self.server_playback_path(options.server_replay)
  File "/usr/local/lib/python2.7/site-packages/libmproxy/console/__init__.py", line 514, in server_playback_path
    self.statusbar.message(ret)
AttributeError: 'ConsoleMaster' object has no attribute 'statusbar'
```

mhils:
Excellent find. Thanks for the report!

nickhs:
Thanks Max! I've had a quick stab at resolving the issue, let me know if the above PR makes sense.

PR #442, opened 2015-01-08
Error to stdout when loading bad replays from the command line

nickhs:
Fixes bug #441


mhils:
Great. Thanks a lot! :smile:

PR #445, opened 2015-01-12
hotfix to mitmproxy missing start_server_playback() parameter added. 

gato:
Hi Max, 
I've just backported the fix in mitmproxy's start_server_playback() from PR [#439] to add the missing parameter (replay_ignore_payload_params) which is present in mitmdump but was missing in mitmproxy.
Regards
Marcelo


mhils:
Good find. Thanks!

PR #447, opened 2015-01-13
missing parameter in start_server_playback call (again)

gato:
Hi Max,
I found two more places were start_server_playback has missing the last parameter 
I grep the code and these seems like the only places were it was missing
Regards.
Marcelo


mhils:
Excellent find, thanks! :+1:

Issue #334, opened 2014-08-26
Feature request - copy response body to clipboard


rwozniak:
As in the title - it would be nice to be able to copy response body to clipboard with some shortcut.

mhils:
Thanks to @gato, the next mitmproxy version will ship with this feature. :-)

danielphillips:
How do you use this feature?

dufferzafar:
@danielphillips Depends on which mitmproxy version you're using.

In the code from latest master, you just press `C` on a flow and choose what you want to copy - body, headers etc.

shubham-mandal:
How do we copy the flow in the latest version of mitmproxy 3.0.3?

shubham-mandal:
How do we copy the flow in the latest version of mitmproxy 3.0.3?

kajojify:
@shubham-mandal Hi! 
`ctrl + L` (L in lowercase)
You will see `: cut.clip [flow] [cut]` in the statusbar, where `[flow]` is some view flow selector from https://docs.mitmproxy.org/stable/concepts-filters/ and [cut] is a part of a flow.

shubham-mandal:
@kajojify  Many thanks, I had tried the command but couldn't figure out the arguments. After a mild research, I finally understood them. Thanks again.

laxiK:
How to copy the request url and request headers in the latest version of mitmproxy 3.0.3 as I've tried these steps: ctrl + L (L in lowercase)
then : cut.clip @focus request.url
But it's not working.

alejoes:
@shubham-mandal would you mind sharing the arguments? I guess you mean [flow][cut], I have not been able to figure it out. Thanks

alejoes:
Hi, I just recently discovered mitmproxy, I'm not much of a gig and I'm having trouble figuring out what to do after clicking ctrl + L (L in lowercase),  when cut.clip [flow] [cut] shows in the statusbar
I would appreciate any help.
Thank you

PR #448, opened 2015-01-14
copy response body to clipboard [#334] implemented

gato:
Hi Max,
I've implemented Feature request - copy response body to clipboard [#334] in flowlist and in flowview 
It's able to copy response's content and header. 
I was going to implement the same for request but I'd rather like you or @cortesi to validate this implementation first. 
If it's ok I will replicate the same for request.
As a known issue:
When content has invalid utf8 characters pyperclip (or i guess is xclip implementation below) prints a warning that forces the screen to scroll
Regards,
Marcelo


libmproxy/protocol/http.py
```diff
@@ -953,6 +953,15 @@ def replace(self, pattern, repl, *args, **kwargs):
             c += self.response.replace(pattern, repl, *args, **kwargs)
         return c
 
+    def response_content(self):
+        with decoded(self.response):
+            s = self.response.content
+        return s
```
mhils: see get_decoded_content

libmproxy/protocol/http.py
```diff
@@ -953,6 +953,15 @@ def replace(self, pattern, repl, *args, **kwargs):
...
+
+    def response_headers(self):
+        with decoded(self.response):
+            s = str(self.response.headers) 
+        return s
```
mhils: the headers are never encoded (only the content is, which is specified in the content-encoding header), you can just str them and remove the function here.

libmproxy/console/flowlist.py
```diff
@@ -136,6 +138,17 @@ def server_replay_prompt(self, k):
                 self.state.last_saveload,
                 self.master.server_playback_path
             )
+    def server_copy_response(self, k):
+        if k == "c":
+            try:
+                pyperclip.copy(self.flow.response_content())
+            except TypeError:
+                self.master.statusbar.message("Content is binary or can be converted to text")
```
mhils: + not :-)
We should indicate here that the content can be dumped to disk by pressing w (?) alternatively.

mhils:
Hi @gato,

thanks for taking the initative here - really cool to see that being implemented! :smile: 
I'm a bit in a hurry now and will take a closer look later.

Cheers,
Max

gato:
Hi Max,
I've changed the decoding stuff. 
I've started with "pressing "w" to dump content to file" and found that in flowview the "b" hotkey does exactly that. also "W" & "w" are used in both cases (flowview and flowlist) to write complete flows.
what do you suggest? changing the message to suggest pressing "b" and add the functionality to flowlist too? or 
show a message like: 
"content is binary do you want to save it to a file instead?" 
>> yes / no

if yes ask for path as if "b" key was pressed? and save it ?
Regards.
Marcelo

mhils:
Hi Marcelo,

> what do you suggest? changing the message to suggest pressing "b" and add the functionality to flowlist too? or 
show a message like: 
"content is binary do you want to save it to a file instead?"
yes / no
if yes ask for path as if "b" key was pressed? and save it ?

I'd say both options are fine. Do what you feel fits better :-)
Also, if the `server_copy_response` method is getting larger, feel free to extract it and place it in libmproxy.console.common or so.

The thing I'm more worried about is that we now assume that xclip is installed on the system - otherwise, mitmproxy won't start. We don't have any other system package dependencies and I'd be very much inclined to keep that. How about just degrading gracefully if we can't import pyperclip?

```python
try:
    import pyperclip
except:
    pyperclip = False

# Then use the following pattern:
if pyperclip:
    # ...
else:
    self.master.statusbar.message("No clipboard support on your system, sorry.")
```
You can then exclude the travis workarounds (sudo implies a different test system and slows down the tests by a factor of ~4x) as well :-)

Thanks for the great work!

Cheers,
Max

gato:
Hi Max!
I've pushed the suggested changes (pyperclip is now optional, server_copy_response was moved to common and if content is binary the proxy ask if you want to save it instead)
Also moved save_body to common and refactored flowview to call it when "b" is pressed.
if this is ok, i will replicate the behavior for the request (for now is response only) and can also add "b" hotkey in flowlist.
Regards.
Marcelo

mhils:
Hi Marcelo,

looks good to me. Make sure that you rename `server_copy_response` to `copy_message` or someting along those lines and use it for both requests and responses. :smile: 

Cheers,
Max

gato:
Hi Max,
I finished implementing copy to clipboard for requests too. 
and added support for the "b" key (save content to file directly) in flowlist (was in flowview only)
I have no more changes to add, so fill free to merge this PR. you can also fell free to suggest some mods ;) 
Regards
Marcelo

mhils:
Excellent. It'll take a few days before I can take a closer look, so bear with me for the moment :-)

Cheers,
Max

gato:
No problem,  just let me know if you need anything.
Regards
Marcelo

gato:
Hi Max,
I've updated the PR to add support for copying the url to the clipboard.
This is part of what @gajus requested in issue #460 
Regards.
Marcelo

gato:
Hi Max!, 
forget about the number of commits (git rebase is to blame) changes are not much, if you want i can squash them to 1 commit and make a new PR.
Just let me know
Regards.
Marcelo

mhils:
Hi there,

I'm fine with merging it as-is, but feel free to do a rebase if you want to keep the commit history clean. ;-)

Whatsoever, I just did a quick test and ran into a major issue on Ubuntu 14.04:

```
$ sudo apt-get install xclip
```
```
$ mitmproxy
<Copy a flow into the clipboard>
<Exit mitmproxy>
$ mitmproxy
Error starting proxy server: error(98, 'Address already in use')
$ netstat -lnp | grep 8080
tcp        0      0 0.0.0.0:8080            0.0.0.0:*               LISTEN      4687/xclip
```

As it seems, pyperclip is [not closing child programs properly](https://github.com/asweigart/pyperclip/blob/2938a488e62843f2496b85d5e54fbb6d3ad5a7b9/pyperclip/__init__.py#L134)... Can we fix pyperclip first?

Cheers,
Max

gato:
Hi Max, 
I've done a rebase and that's why some commits are "duplicated" (after I rebased my local repo github (origin) branch and my local branch have different starting commits and also all my commits where replayed on the new base (changing it's uuids) then git try both branches as diverged and I had to  merge them. 
i will create a new branch and apply only the unique commits, then will remove remote branch and recreate it.
As the pyperclip issue i can try to make a PR for them and see what happens
I haven't had the issue (in ubuntu 14.04) as i have gnome and pyperclip fallbacks to gtk clipboard when xclip nor xsel are installed (my case). 
Regards.
Marcelo

mhils:
Excellent, thanks. Having clipboard functionality in mitmproxy would be great, but I'd be reluctant to have side effects with pyperclip that cause apparently unrelated errors.

If you feel that they're not moving, we can of course ship our own copy of pyperclip, but having it solved upstream would be preferable.

Thanks!
Max

libmproxy/console/common.py
```diff
@@ -161,6 +165,88 @@ def raw_format_flow(f, focus, extended, padding):
...
+                state,
+                flow.response.get_decoded_content(),
+            )
+        else:
+            master.statusbar.message("Flow has no response")
+
+## common copy_message parts 
+def copy_message( k, master, state, message):
+    if not message:
+        # only response could be None
```
mhils: This must not be the case. We want to get to a point where we have error-only flows as well.

gato: I created this check because in flowlist you can press "g" (copy response) in a flow that has no response yet. I guess i can move the check to flowlist, but has to be done somewhere. (i do not remember if this case can be triggered from flowview).  Do you want me to change this?

gato:
Hi Max, 
I was checking the duplicated commits issue and can't remove them without rewriting the history of my public repo (push -f) which is not only bad and ugly, but will almost surely break this PR so i can leave it like this or PR a clean branch. 
I've also PRed a fix to pyperclip (that works on my machine with xclip) so now we wait to see if they upgrade the lib or not.
Regards.
Marcelo

mhils:
Thanks for pushing this, @gato!
Looking at pyperclip, they are apparenty quick. I feel we can merge here once https://github.com/asweigart/pyperclip/pull/13 is merged as well and they shipped a release (we need to update our requirements file then).

Regarding the commits: I think we can just merge it as-is., screw the couple of additional entries we're getting.

Apart from that, I will follow up with inline comments now for all minor issues that I encountered. :smiley:

libmproxy/console/common.py
mhils: Forget about this comment - I was in a different context at that point, you're completely right here.

libmproxy/console/common.py
```diff
@@ -161,6 +165,88 @@ def raw_format_flow(f, focus, extended, padding):
...
     return urwid.Pile(pile)
 
+## common save body parts
+def save_body(path, master, state, content):
+    if not path:
+        return
+    state.last_saveload = path
+    path = os.path.expanduser(path)
+    try:
+        f = file(path, "wb")
```
mhils: I see you only copied this, but can we please go for `with open(...)` here as well? Also, I think we don't need to `str()` the content, this would even break in Python 3.

libmproxy/console/common.py
```diff
@@ -161,6 +165,88 @@ def raw_format_flow(f, focus, extended, padding):
...
+    if not message:
+        # only response could be None
+        master.statusbar.message("Flow has no response")                
+        return
+
+    if pyperclip:
+        if k == "c":
+            try:
+                pyperclip.copy(message.get_decoded_content())
+            except TypeError:
```
mhils: Can you actually trigger this? I can't.

libmproxy/console/common.py
```diff
@@ -161,6 +165,88 @@ def raw_format_flow(f, focus, extended, padding):
...
+                master.prompt_onekey(
+                    "Content is binary do you want to save it to a file instead?",
```
mhils: This message is missing a statement that the data has not been / can not be copied into the clipboard. How about
"Cannot copy binary data to clipboard. Save as file?" or something along those lines?

setup.py
```diff
@@ -20,7 +20,8 @@
     "pyasn1>0.1.2",
     "pyOpenSSL>=0.14",
     "tornado>=4.0.2",
-    "configargparse>=0.9.3"
+    "configargparse>=0.9.3",
+    "pyperclip>=1.5.7"
```
mhils: 1.5.8 :smile:

libmproxy/console/common.py
gato: not right now, I have osx at home, but will try tomorrow with an ubuntu box. As far as i recall using gtk clipboard (not tried xclip) when trying to copy payload of a binary response this exception is thrown. 
I will try tomorrow with gtk and with xclip and let you know.

gato: I've tested In ubuntu 14.04 and exception is thrown when using gtk clipboard and xsel. but not when xclip is installed.
also in mac osx exception is not thrown.

mhils:
Continued in #467.

Issue #426, opened 2014-12-15
Create New Requests Inside mitmproxy


revintec:
when mitmproxy opens, there is no request
then I want to issue a new request to the server to see how server responses
this seems can't be done

(why not set proxy and use curl ...)
because mitmproxy has a nice editor ...
(why I want this)
I think it's a valuable feature, sometimes you just want to provoke the server to see how
it responses, maybe try to squeeze some ajax api out of it, testing some variables
and the easy way is to make a new request instead of curl/browser/...

what I want:
could someone tell me is there a way to make new request right inside mitmproxy?
(I think it should have one, maybe I missed it?)
if not, could you add a shortcut "n" at main screen to start make and edit a new request?

gato:
Hi Max,
I'm trying to implement this feature but runned into an issue with ssl https://github.com/tekii/mitmproxy/tree/feature-426
the code in that branch can create a new request from the flowlist (with or without previous flows) and also can resend http request and it works ok, but if created request is a https one SSL.py throws an exception. 
```
Exception in thread RequestReplayThread:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/home/gato/trabajos/mitmproxy/libmproxy/protocol/http.py", line 1467, in run
    self.flow.response = HTTPResponse.from_stream(server.rfile, r.method,
  File "/home/gato/trabajos/mitmproxy/libmproxy/proxy/connection.py", line 144, in send
    self.wfile.write(message)
  File "/home/gato/trabajos/mitmproxy/netlib/netlib/tcp.py", line 94, in write
    return self.o.sendall(v)
  File "/home/gato/.local/lib/python2.7/site-packages/OpenSSL/SSL.py", line 969, in sendall
    raise TypeError("buf must be a byte string")
TypeError: buf must be a byte string
```
I must be missing some initialization in the creation of the request done here https://github.com/tekii/mitmproxy/blob/feature-426/libmproxy/flow.py#L767-L771
but i can't figure out what is missing, if you could point me in the right direction i could PR this soon.
Regards.
Marcelo

mhils:
Hey Marcelo,

thanks for picking this up! :smiley:

I'll take a look at the specific issue once I have access to a debugger later today or tomorrow.

[...]

PR #474, opened 2015-02-12
Feature 426

gato:
Hi Max, 
Adding str() fixed the issue thanks.
Here is the PR for issue #426 with that addition and the method renamed 
The created request is almost empty, only method and url are set by default, but everything else can be added using the flow editor. 
just let me know if you think more data needs to be precreated.
Regards.
Marcelo

### DIFF
--- a/libmproxy/console/flowlist.py
+++ b/libmproxy/console/flowlist.py
@@ -267,7 +267,7 @@ class FlowListBox(urwid.ListBox):
     def new_request(self, url, method):
         try:
             scheme, host, port, path = http.parse_url(url)
-            f = self.master.add_request(method, scheme, host, port, path)
+            f = self.master.create_request(method, scheme, host, port, path)
             self.master.view_flow(f)
         except ValueError:
             self.master.statusbar.message("Invalid Url")
--- a/libmproxy/console/flowlist.py
+++ b/libmproxy/console/flowlist.py
@@ -265,12 +265,13 @@ class FlowListBox(urwid.ListBox):
         self.master.prompt("URL:", "http://www.example.com/", self.new_request, method)
 
     def new_request(self, url, method):
-        try:
-            scheme, host, port, path = http.parse_url(str(url))
-            f = self.master.create_request(method, scheme, host, port, path)
-            self.master.view_flow(f)
-        except ValueError:
+        parts = http.parse_url(str(url))
+        if not parts:
             self.master.statusbar.message("Invalid Url")
+            return
+        scheme, host, port, path = parts
+        f = self.master.create_request(method, scheme, host, port, path)
+        self.master.view_flow(f)
 
     def keypress(self, size, key):
         key = common.shortcuts(key)
--- a/libmproxy/console/flowlist.py
+++ b/libmproxy/console/flowlist.py
@@ -17,7 +17,7 @@ def _mkhelp():
         ("g", "copy flow to clipboard"),
         ("l", "set limit filter pattern"),
         ("L", "load saved flows"),
-        ("n", "creates a new request"),
+        ("n", "create a new request"),
         ("r", "replay request"),
         ("V", "revert changes to request"),
         ("w", "save flows "),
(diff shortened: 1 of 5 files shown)


libmproxy/console/flowlist.py
```diff
@@ -16,6 +17,7 @@ def _mkhelp():
         ("g", "copy flow to clipboard"),
         ("l", "set limit filter pattern"),
         ("L", "load saved flows"),
+        ("n", "creates a new request"),
```
mhils: creates -> create

libmproxy/console/flowlist.py
```diff
@@ -245,6 +247,31 @@ def __init__(self, master):
...
+            self.master.prompt("Method:", "", self.get_method_raw)
+        else:
+            method = ""
+            for i in common.METHOD_OPTIONS:
+                if i[1] == k:
+                    method = i[0].upper()
+            self.get_url(method)
+
+    def get_url(self,method):
+        self.master.prompt("Url:", "http://www.example.com/", self.new_request, method)
```
mhils: Url -> URL (as in the flow editor)

mhils:
Hey Marcelo,

great work - thanks. Three issues left:

1. parse_url doesn't throw a ValueError (it probably should, but if we wanted to change that, we shouldn't do it in this PR :wink: )
```
  File "/media/sf_git/mitmproxy/libmproxy/console/__init__.py", line 628, in run
    self.ui.run_wrapper(self.loop)
  File "/usr/local/lib/python2.7/dist-packages/urwid/raw_display.py", line 268, in run_wrapper
    return fn()
  File "/media/sf_git/mitmproxy/libmproxy/console/__init__.py", line 848, in loop
    self.prompt_execute()
  File "/media/sf_git/mitmproxy/libmproxy/console/__init__.py", line 773, in prompt_execute
    msg = p(txt, *args)
  File "/media/sf_git/mitmproxy/libmproxy/console/flowlist.py", line 269, in new_request
    scheme, host, port, path = http.parse_url(url)
TypeError: 'NoneType' object is not iterable
```
2. I think you forgot including ServerConnection and  ClientConnection in this PR. They are strictly mandatory though, i.e. try to *D*uplicate a artificial flow.
3. We should keep the phrasing similar to the rest of the ui (see comments)

Thanks!
Max

libmproxy/flow.py
```diff
@@ -763,6 +763,21 @@ def tick(self, q, timeout):
...
         return self.load_flow(f.copy())
 
+    def create_request(self, method, scheme, host, port, path):
+        """
+        Creates a new request from params and add it to flow list.
+        created request is empty (except for method and url) but is able 
+        to be replayed 
+
+        """        
+        f = http.HTTPFlow(None,None);
```
mhils: Wrong language :wink:

libmproxy/flow.py
```diff
@@ -763,6 +764,29 @@ def tick(self, q, timeout):
...
+        c = ClientConnection.from_state(dict(
+                address=dict(address=(host, port), use_ipv6=False),
+                clientcert=None
+            ))
+
+        s = ServerConnection.from_state(dict(
+                address=dict(address=(host, port), use_ipv6=False),
+                state=[],
+                source_address=None, #source_address=dict(address=(host, port), use_ipv6=False),
+                cert=None
```
mhils: Depending on the scheme, we should add `ssl_established=True` and `sni=host` here. Generally speaking I feel that it's not optimal depending on these attributes of the server connection, but I don't really know of a better place to put them.

libmproxy/protocol/http.py
```diff
@@ -1457,9 +1457,11 @@ def run(self):
                     server = ServerConnection(server_address)
                     server.connect()
                     if r.scheme == "https":
-                        server.establish_ssl(self.config.clientcerts, sni=self.flow.server_conn.sni)
+                        sni = None
+                        if self.flow.server_conn:
+                            sni = self.flow.server_conn.sni
+                        server.establish_ssl(self.config.clientcerts, sni=sni)
```
mhils: These changes shouldn't be neccessary. We should always assume there's a ServerConnection.

mhils:
Excellent. Once the SNI is in, we're done here. :smiley:

mhils:
Great. :+1:

Issue #487, opened 2015-02-26
Display milliseconds for each request


cellularmitosis:
I spend a lot of time developing a mobile application against an API, and most of the time my phone is going through mitmproxy.

It would be great if each request showed how milliseconds (or seconds, yikes!) it took each request to come back from the server, so that I could easily keep track of which API endpoints need some optimization attention.

e.g. instead of this:

   POST https://p09-ckdatabase.icloud.com/api/client/user/quota
        ← 200 application/x-protobuf 78B 39.28kB/s

it would look something like this:

   POST https://p09-ckdatabase.icloud.com/api/client/user/quota
        ← 200 application/x-protobuf 78B 39.28kB/s (237ms)

or this:

   POST https://p09-ckdatabase.icloud.com/api/client/user/quota
        ← 200 application/x-protobuf 78B 39.28kB/s (1.32s)

or this:

   POST https://p09-ckdatabase.icloud.com/api/client/user/quota
        ← 200 application/x-protobuf 78B 39.28kB/s (27.9s)

(i.e. three significant digits would be great)

mhils:
Thanks for the suggestion @cellularmitosis and thanks for the incredibly quick PR, @gato. :smiley: I feel your pain - the web ui we're working on already features a way more detailed view on timing information: 
![screenshot](https://maximilianhils.com/upload/2015-02/2015-02-27_09-24-26.png)

Looking at mitmproxy, we should be careful that we don't clutter the UI with too much information. As it seems, @cortesi removed timing information in an earlier version (https://github.com/mitmproxy/mitmproxy/commit/a2643b52f909719d4e0fe97e17d3d30f43669341) precisely for this reason. Anyhow, I feel that the response time would be more useful than the rate, so maybe we should swap those two. Before we do that, we need to hear why @cortesi initially decided to display the rate only.

Cheers,
Max

gato:
Hi Max, 
maybe a hotkey can switch between values or as @cortesi has suggested in (a2643b5) this value or more detailed info could be shown only in flowview and not in flowlist.
Just let me know.
Regards.
Marcelo
P.S. new ui looks really good ;)

mhils:
Hey there,

I had a quick chat with @cortesi tonight. In short, I think we can pick up most points raised in this discussion here. :smiley:

1. Replace the download rate in the UI with the elapsed time between `request.timestamp_start` and `response.timestamp_end`. We then have size + elapsed time in the UI, the download rate could be directly derived.
[...]

PR #489, opened 2015-02-26
added time elapsed between request sent and response received

gato:
Hi Max, 
i've added to flowview and flowlist time elapsed to handle request. I've used the difference between end of request and end of response but flow has timestamps for begin and end for both request and response so other combinations can be shown either. 
Just let me know if this approach seems ok or you want to calculate it differently. Or perhaps add to view more times (req time: xxx ms., resp time: yyy ms. , whole time: zzz ms.)
I created this PR for issue #487 
Regards
Marcelo


mhils:
Excellent. Thanks! :smiley:

PR #492, opened 2015-02-27
use entry_points for Windows support

mhils:
We should use setuptool's `entry_points` instead of `scripts` - `scripts` don't work on Windows.


mhils:
This partially reverts https://github.com/mitmproxy/mitmproxy/commit/ec17e70d9e82bc6ae62300397b2f849c2d8dcc85 - what was the motivation for using scripts here?

Issue #469, opened 2015-02-08
Allow Selection of Upstream Cipher Suites


mhils:
Currently, a user can specify the ciphers presented to the client by using the `--ciphers` switch. As with `--ssl-version-client/--ssl-version-server`, we should introduce `--ciphers-server` and rename `--ciphers` to `--ciphers-client`. 

Relevant parts in the code:
- [libmproxy/proxy/config.py#L217](https://github.com/mitmproxy/mitmproxy/blob/c9240812d9239591915b521524e8a1dbbef05b0f/libmproxy/proxy/config.py#L217)
- [libmproxy/proxy/config.py#L58](https://github.com/mitmproxy/mitmproxy/blob/c9240812d9239591915b521524e8a1dbbef05b0f/libmproxy/proxy/config.py#L58)
- [libmproxy/proxy/server.py#L181-L218](https://github.com/mitmproxy/mitmproxy/blob/c9240812d9239591915b521524e8a1dbbef05b0f/libmproxy/proxy/server.py#L181-L218)

yzninja:
It would be helpful if the "--ciphers-server" switch could inherit the cipher suite of the client device. Perhaps using "--ciphers-server inherit".

yzninja:
Max - I did a search in the github repository to see where mitmproxy is analyzing "client hello" requests but couldn't find it. Could you point me in the right direction?

mhils:
@yzninja, take a look at the last excerpt i posted. I calls [TCPClient.convert_to_ssl](https://github.com/mitmproxy/netlib/blob/b6af3fddf4ee2c45a722bd5087c86cee59b8cfa0/netlib/tcp.py#L293), which accepts a cipher_list argument.

elitest:
I have been looking at how to implement this.  I can see how the client interacts with netlib as you referrenced, but I'm having trouble understanding how `self.server_conn.establish_ssl` interacts with it.    `convert_to_ssl` accepts `cipher_list` as a parameter, but `establish_ssl` does not.  Am I barking up the wrong tree?

mhils:
[...]

PR #493, opened 2015-03-02
Adding support for server facing SSL cipher suite specification

elitest:
This pull request adds support for specifying ciphers on the server side of the proxy as part of its CLIENTHELLO(--server-ciphers).  It also modifies the current cipher(--ciphers) parameter clarify that it is for the client facing side of the proxy(--client-ciphers).  I tested and observed, with wireshark, ciphers that I expected both leaving and coming to the proxy.

This is in regards to feature issue/request [mitmproxy#649](https://github.com/mitmproxy/mitmproxy/issues/469).


mhils:
Excellent work, thanks a lot! :smile:

Issue #504, opened 2015-03-05
mitmweb: fix filters


mhils:
Something broke the filters, not sure what.

legendtang:
Confirmed that it is EventEmitter related. See the PR.

Issue #503, opened 2015-03-05
mitmweb: fix live updates 


mhils:
mitmweb already connects to a WebSocket channel and fetches live updates. A recent commit apparently broke this functionality and the UI doesn't get refreshed.

PR #505, opened 2015-03-05
Replace `addListener` method with multiple bindings, fix for #503 #504

legendtang:
It's an EventEmitter issue.


mhils:
Hi @legendtang,

thanks for tracking this down! :smile:
Regarding your fix: I'd rather like to keep the EventEmitter from node than our own - less code is better code. :wink: 
I think the issue is that the `addListener` calls of node's emitter don't accept our shorthand "add remove update" syntax but require three separate calls instead. Rather than re-adding our EventEmitter, we should just adjust our calls to `addListener`.

Thanks!
Max

mhils:
refs #503, refs #504

legendtang:
Hi @mhils, 

Thanks for pointing out. Indeed, less code is better code. Maybe this feature can be a feature of node or a new npm package in the future.

Fixed just now.

Regards
Legend Tang

web/src/js/components/flowtable.js
```diff
@@ -65,15 +65,24 @@ var FlowTable = React.createClass({
...
         if (nextProps.view !== this.props.view) {
             if (this.props.view) {
-                this.props.view.removeListener("add update remove recalculate");
+                this.props.view.removeListener("add");
+                this.props.view.removeListener("update");
+                this.props.view.removeListener("remove");
+                this.props.view.removeListener("recalculate");
             }
-            nextProps.view.addListener("add update remove recalculate", this.onChange);
+            nextProps.props.view.addListener("add", this.onChange);
```
mhils: `nextProps.view.addListener` ?

mhils:
I have a few more places where we call addListener with more than one argument: 
```
$ grep -r . addListener
js/components/eventlog.js:        view.addListener("add recalculate", this.onEventLogChange);
js/components/flowtable.js:            this.props.view.addListener("add update remove recalculate", this.onChange);
js/components/flowtable.js:            nextProps.view.addListener("add update remove recalculate", this.onChange);
js/components/mainview.js:        view.addListener("recalculate", this.onRecalculate);
js/components/mainview.js:        view.addListener("add update remove", this.onUpdate);
js/components/mainview.js:        view.addListener("remove", this.onRemove);
js/components/proxyapp.js:        this.state.settings.addListener("recalculate", this.onSettingsChange);
js/store/view.js:    this.store.addListener("add", this.add);
js/store/view.js:    this.store.addListener("update", this.update);
js/store/view.js:    this.store.addListener("remove", this.remove);
js/store/view.js:    this.store.addListener("recalculate", this.recalculate);

```

web/src/js/components/flowtable.js
legendtang: Sorry:-/ I have just mistaken that and fixed.

legendtang:
@mhils wow.

mhils:
:+1:

PR #508, opened 2015-03-06
Add --replay-ignore-host option

onlywade:
Similar to the other --replay-ignore-* options, this prevents matching on the destination address while searching for flows to replay.

The use case necessitating this is replaying server responses (in transparent mode) for domains that use multiple IP addresses.

### DIFF
--- a/libmproxy/cmdline.py
+++ b/libmproxy/cmdline.py
@@ -192,7 +192,8 @@ def get_common_options(options):
         nopop=options.nopop,
         replay_ignore_content = options.replay_ignore_content,
         replay_ignore_params = options.replay_ignore_params,
-        replay_ignore_payload_params = options.replay_ignore_payload_params
+        replay_ignore_payload_params = options.replay_ignore_payload_params,
+        replay_ignore_host = options.replay_ignore_host
     )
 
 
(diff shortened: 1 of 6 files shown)


Kriechi:
@onlywade could you please rebase your commits onto master and get rid of your merge-commits?
And are you aware the git-committer that is set for your commits?

onlywade:
Thanks, Kriechi. I've cleaned up the PR and corrected my git-committer config.

mhils:
Excellent work. We had some regressions due to similar PRs lately (see #447), so I probably want to review this one carefully, which may take a few days before I get to it. Thank a lot so far! :smiley:

onlywade:
Cool, totally understandable. Come to think of it, my change may not be compatible with the console interface, as I tend to use only mitmdump for server replay - I'll take a look and make sure I'm not introducing any regressions there, and update my PR if needed.

onlywade:
@mhils, I've adjusted the changes to make sure server replay doesn't crash in console mode, and tested with both mitmdump and mitmproxy. Thanks for the reminder!

mhils:
Everything look good here - merged. Thanks for the PR! :smile:

PR #512, opened 2015-03-09
added bash autocompletion

gato:
Hi Max,
I've created a small script to add autocompletion (for bash now) to mitmproxy, mitmdump and mitmweb
to use it run
```
source autocomplete.bash
```
or copy / link it to /etc/bash_completion.d/ for system wide use
Cheers
Marcelo
P.S. will start with completion for zsh as is the other sh that i use

### DIFF
--- a/autocomplete.bash
+++ b/autocomplete.bash
@@ -16,10 +16,41 @@ _mitmproxy()
  	opts+="--setheader --nonanonymous --singleuser --htpasswd --cert --cert-forward --ciphers-client --ciphers-server " 
     opts+="--client-certs --no-upstream-cert --ssl-port --ssl-version-client --ssl-version-server -i --intercept " 
 
+    case "${prev}" in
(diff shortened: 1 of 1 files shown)


cortesi:
Hi Marcelo. Thank you for your work - this is very useful, and is something that I would personally use. I would prefer not to include it in the official repo, though, because I know just how difficult it is to keep things like this in sync with the main project. External helpers are difficult to test, and become impossible to maintain if developer interest flags, or if it deals with a tool primary developers don't use.

I have two possible suggestions, though. One, if you put the shell completion scripts in a repo, I'm happy to add a pointer to it in our official installation instructions. Second, this seems like a golden opportunity for a more ambitious project. The command-line interface for all these tools is defined by a standard argparse(-compatible) library - it should be possible to pick that up and auto-generate the command-line completions for bash, zsh and other shells. I've had a look at projects like this [argcomplete](https://github.com/kislyuk/argcomplete), but they've never worked well for me.

gato:
Hi Aldo,
Thanks for reviewing this PR .
I will follow your advice and try to autogenerate the suggestions.
Regards
Marcelo

PR #513, opened 2015-03-10
Added transcript.py example that makes a human readable session transcript

stevemadere:
I find this particular script invaluable for analyzing API protocols
and  suspect many others will too.

It creates a highly readable transcript of a mitmproxy/mitmdump session.

It  decodes URL query params, request content and response content and
pretty-prints them as JSON.


mhils:
Hi Steve,

thanks for the PR! I like what you're doing here, but I feel that writing reasonable representations of flows should be a core feature of mitmdump itself. Taking your script as inspiration, I improved `mitmdump -dd` so that we now strip off transfer-encoding before printing and indent JSON where possible.

Cheers,
Max

PR #517, opened 2015-03-14
simplify virtualenv

mhils:
- add dev.bat for Windows users
- remove superfluous pip install calls.
- remove `~/.pipcache` stuff.  
  If that causes issues with PyInstaller, we should use `pip --no-cache-dir` there or `rm -r pip.locations.USER_CACHE_DIR`.

@cortesi: third change works for you?

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -64,8 +64,8 @@ checkouts will be reflected live in the virtualenv.
 To confirm that you're up and running, activate the virtualenv, and run the
 mitmproxy test suite:
 
-```
-$ ../venv.mitmproxy/bin/activate
+```shell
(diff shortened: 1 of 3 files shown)


cortesi:
I can live with removal of the pip cache. Running pyinstaller without the cache would take too long, though - I run it hundreds of times while troubleshooting for each release. I'll figure something else out, we can re-assess later.

Other changes look good. We should output some text from each dev script saying what's being done (creating virtualenv PATH, etc.).

mhils:
I removed the echo statements as they were basically hidden by pip's verbosity. I'll add them at the end again and merge. :smile:

Issue #518, opened 2015-03-14
A clean `npm install` for mitmweb unavailable now due to `react@0.13.0` released


legendtang:
A clean `npm install` shows the following:
````
npm ERR! argv "node" "/usr/local/lib/node_modules/cnpm/node_modules/.bin/npm-for-cnpm" "--userconfig=/home/legend/.cnpmrc" "--disturl=http://npm.taobao.org/mirrors/node" "--cache=/home/legend/.cnpm" "--registry=http://registry.npm.taobao.org" "install"
npm ERR! node v0.10.36
npm ERR! npm  v2.6.0
npm ERR! code EPEERINVALID

npm ERR! peerinvalid The package react does not satisfy its siblings' peerDependencies requirements!
npm ERR! peerinvalid Peer react-router@0.12.4 wants react@0.12.x

npm ERR! Please include the following file with any support request:
npm ERR!     /home/legend/mitmproxy/web/npm-debug.log
````

Transparently, there is an compatibility issue between `react-router` and `react` at the moment.
A few days ago, package [react](https://www.npmjs.com/package/react) v0.13.0 published, but an amount of packages which suits to `react@0.12.x` only haven't updated yet.

Simply modify the `package.json` would work.

mhils:
fixed in #519.

PR #519, opened 2015-03-14
update package.json during transition period, see #518

legendtang:
Decide to modify or just wait `react-router` to update.


mhils:
Thanks!

PR #520, opened 2015-03-16
autocompletion using argcomplete

gato:
Hi, 
I've implemented autocompletion using argcomplete (https://github.com/kislyuk/argcomplete) as @cortesi suggested in #512 
It's fully working but has 2 issues: 
1.- as "--conf" parameter is added to parser in construction time https://github.com/mitmproxy/mitmproxy/blob/master/libmproxy/cmdline.py#L566-L575  I couldn't add a FilesCompleter to it (so no file suggestion for this param, the param itself is suggested). I have some ideas on how to solve this but all of them are hacks.
2.- completion is slow: The main reasons for this is that argcomplete runs the script to generate the suggestions every time tab is pressed (there is an issue requesting static completions, but i don't think will be done any time soon https://github.com/kislyuk/argcomplete/issues/85 ) And mitmproxy loads lots of libraries before parsing args so expect a delay between 1.2 and 2 seconds between pressing tab and seen suggestions on screen.
Regards
Marcelo

### DIFF
--- a/libmproxy/cmdline.py
+++ b/libmproxy/cmdline.py
@@ -5,6 +5,7 @@ import configargparse
 from netlib import http
 from . import filt, utils, version
 from .proxy import config
+from argcomplete.completers import FilesCompleter
 
 APP_HOST = "mitm.it"
 APP_PORT = 80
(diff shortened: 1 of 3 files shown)


cortesi:
Hi Marcelo. We'd like something like this, but as you say argcomplete is just too slow at this point to be a great user experience. Lets keep thinking about this.

Issue #516, opened 2015-03-13
Improving mitmproxy’s Documentation


elitest:
I have some suggestions and want to work on mitmproxy’s documentation.  I wanted to throw my thoughts out to the community so that I could gather some feedback and ideas before doing a bunch of work and dropping in with some PRs.  This is a great project, but I feel like it needs better docs!   If it would be better I can get these split out into different issues, so that they can be discussed and worked on separately separately.
#### 1. Spruce up the installation instructions.

The Ubuntu instructions need to be fleshed out a bit as I think there are steps missing that even a technical person might not think are obvious.  Been a while since I used  OS X, and even when I did use it, I’m not familiar with how python works on it(homebrew vs. XCode).  Perhaps we could have a pros and cons of why you would install from homebrew vs use the python that comes with XCode.  I assume it boils down to a “Do you mind old python vs more setup” thing.  I may need some help from someone more familiar with mitmproxy on OS X on this one.  Or perhaps make the decision to document one of the methods say, homebrew over XCode.
#### 2. Developer docs

Related to this but somewhat separate I think we should also create developer installation docs so that people know how to get set up to help contribute.  Explain what Countershape is and how to write documentation.
#### 3. Command line options in one place

I think we need to have a page that lists all of the command line options for each tool rather than having them scattered throughout the features/examples.  I’m aware that you could make the argument “just specify --help at the command line”.  The purpose of the command line options page would give someone who hasn’t even installed mitmproxy yet a one-pager on what you can do with mitmproxy, I know this because it was the first page I looked for when I found mitmproxy.  I think that it is important that this page be short and sweet and to the point, perhaps linking to related examples.
#### 4. Introduction docs

I would like to hear some feedback on how people feel about the “[How mitmproxy works](https://mitmproxy.org/doc/howmitmproxy.html)” page vs the “[Modes of Operation](https://mitmproxy.org/doc/modes.html)” page, I feel like this is probably amongst the most polished documentation within this project, but perhaps there is a bit of duplication between the two pages.  Maybe we can re-work these two pages by splitting each of mitmproxy’s modes into its own page, and then take the section related to that mode of operation from the “how mitmproxy works” page and put it on that page.
#### 5. Navigation

[...]

PR #524, opened 2015-03-18
Consolidate SSL documentation

elitest:
As discussed in #516, I have eliminated the platform specific SSL installation instructions.  Combined the Webapp and Overview pages, did a bit of reorganization and cleanup.


mhils:
Thanks for the PR! I'm in favor of the general change here, but we should be careful that we're not losing important bits. In particular, the docs for Java are quite helpful and not covered by the webapp. Also, I'm not sure about the IOS Simulator. I'll ponder a bit how we can structure this best...

elitest:
Though, to someone who has no experience with certificates the Java documentation is as useful as the caveats in the Android documentation.  I'm not sure that the Java documentation, as an example, tells you anything that you couldn't learn from Oracle's official [root cert documentation](http://docs.oracle.com/cd/E19906-01/820-4916/geygn/index.html).  Perhaps linking to official documentation for instructions on how to manage certificates for those clients that don't work with the webapp might be better?  Or even linking to official docs for instructions for all of the most common platforms that users of mitmproxy would be using?

With the iOS simulator that might cause a problem, as I couldn't find any official docs on installing a custom root cert into the iOS simulator and I'm sure Apple's stance is "Why would you want to do that?".  I have only used the Android simulator, not the iOS one, does it not allow you to use the browser inside the simulator to install root certs?  According to [this](http://codeblog.shape.dk/blog/2014/01/06/custom-ssl-certificate-with-charles-web-proxy/) it is as simple as dragging and dropping the certificate onto the simulator window.

mhils:
I wasn't aware of the Oracle documentation - sure, we should just link to it.

Having a single section "Install certificate manually", which provides links to all the relevant docs seems to be the best solution. I personally don't care if they are official docs, as long as they are good.

cortesi:
For what it's worth, I'm all for pruning back the SSL cert documentation, but I think this will increase our support burden. I'm in favor of: a) removing platform-specific instructions wherever the onboarding app works, b) consolidating all the SSL installation info onto one page, c) simplifying and referring to other documentation where it's good, available, and canonical.

elitest:
Sounds good.  I think the my only hesitation to linking to "not official" documentation is that it raises the risk of that documentation becoming a dead link somewhat.  I closed the pull request as I really needed to properly branch my fork, there are a couple other things I would like to work on as well and having my master locked up in a pull request is preventing that! :)  I'll flesh out the one pager more, and see if I can get something better submitted.

Jim

Issue #522, opened 2015-03-17
mitmproxy will stuck / crash if pressing `q` after entering the help section by pressing `?`


legendtang:
```
Traceback (most recent call last):
  File "/home/legend/mitmproxy/libmproxy/console/__init__.py", line 805, in run
    self.loop.run()
  File "/usr/lib/python2.7/site-packages/urwid/main_loop.py", line 278, in run
    self._run()
  File "/usr/lib/python2.7/site-packages/urwid/main_loop.py", line 375, in _run
    self.event_loop.run()
  File "/usr/lib/python2.7/site-packages/urwid/main_loop.py", line 678, in run
    self._loop()
  File "/usr/lib/python2.7/site-packages/urwid/main_loop.py", line 715, in _loop
    self._watch_files[fd]()
  File "/usr/lib/python2.7/site-packages/urwid/raw_display.py", line 392, in <lambda>
    event_loop, callback, self.get_available_raw_input())
  File "/usr/lib/python2.7/site-packages/urwid/raw_display.py", line 492, in parse_input
    callback(processed, processed_codes)
  File "/usr/lib/python2.7/site-packages/urwid/main_loop.py", line 396, in _update
    keys = self.input_filter(keys, raw)
  File "/usr/lib/python2.7/site-packages/urwid/main_loop.py", line 535, in input_filter
    return self._input_filter(keys, raw)
  File "/home/legend/mitmproxy/libmproxy/console/__init__.py", line 616, in input_filter
    elif k in self.onekey:
TypeError: 'in <string>' requires string as left operand, not tuple

mitmproxy has crashed!
Please lodge a bug report at:
	https://github.com/mitmproxy/mitmproxy
Shutting down...
```

I caught this after several tries. It's weird how `k` became a tuple.
I check the code around `libmproxy/console/__init__.py `:616 and have no idea about this output.

Occasionally, it would stuck without crashing and even `^C` lost its function. At this situation, I used `top` to track the CPU / Mem usage but it seemed to be fine.

I remembered that I didn't get this when I installed the git version of mitmproxy for the first time. It might relate to recent changes.

mhils:
Thanks for the report, `git bisect` says this is caused by 488bcaca6fbb1696e05254fbaf193393883467eb.

gato:
Hi @cortesi 
check the proposed fix in #526 
Regards.
Marcelo

legendtang:
Thanks for the fix!

legendtang:
[...]

PR #526, opened 2015-03-18
fix for #522

gato:
Hi,
Setting loop.widget was missing in return from help, also in flowdetailview.py which produced a similar issue.
with this changes i can't reproduce #522 any longer
Regards.
Marcelo


cortesi:
Thanks, Marcelo.

Issue #531, opened 2015-03-22
mitmproxy console: crash when copying binary data to clipboard


cortesi:
```
Traceback (most recent call last):
  File "/Users/cortesi/mitmproxy/mitmproxy/libmproxy/console/__init__.py", line 426, in run
    self.loop.run()
  File "/Users/cortesi/mitmproxy/venv.mitmproxy/lib/python2.7/site-packages/urwid/main_loop.py", line 278, in run
    self._run()
  File "/Users/cortesi/mitmproxy/venv.mitmproxy/lib/python2.7/site-packages/urwid/main_loop.py", line 375, in _run
    self.event_loop.run()
  File "/Users/cortesi/mitmproxy/venv.mitmproxy/lib/python2.7/site-packages/urwid/main_loop.py", line 678, in run
    self._loop()
  File "/Users/cortesi/mitmproxy/venv.mitmproxy/lib/python2.7/site-packages/urwid/main_loop.py", line 715, in _loop
    self._watch_files[fd]()
  File "/Users/cortesi/mitmproxy/venv.mitmproxy/lib/python2.7/site-packages/urwid/raw_display.py", line 392, in <lambda>
    event_loop, callback, self.get_available_raw_input())
  File "/Users/cortesi/mitmproxy/venv.mitmproxy/lib/python2.7/site-packages/urwid/raw_display.py", line 492, in parse_input
    callback(processed, processed_codes)
  File "/Users/cortesi/mitmproxy/venv.mitmproxy/lib/python2.7/site-packages/urwid/main_loop.py", line 399, in _update
    self.process_input(keys)
  File "/Users/cortesi/mitmproxy/venv.mitmproxy/lib/python2.7/site-packages/urwid/main_loop.py", line 499, in process_input
    k = self._topmost_widget.keypress(self.screen_size, k)
  File "/Users/cortesi/mitmproxy/mitmproxy/libmproxy/console/window.py", line 14, in keypress
    k = urwid.Frame.keypress(self, self.master.loop.screen_size, k)
  File "/Users/cortesi/mitmproxy/venv.mitmproxy/lib/python2.7/site-packages/urwid/container.py", line 1116, in keypress
    return self.footer.keypress((maxcol,),key)
  File "/Users/cortesi/mitmproxy/mitmproxy/libmproxy/console/statusbar.py", line 105, in keypress
    return self.ab.keypress(*args, **kwargs)
[...]

PR #532, opened 2015-03-24
handles UnicodeDecodeError (#531)

gato:
Hi @cortesi 
this PR solves #531
When feature #334 was implemented i found a similar error once (a TypeError) but couldn't reproduce it again, so we removed the catch for these type of error. (see https://github.com/mitmproxy/mitmproxy/pull/448#discussion-diff-24215901R226) As pyperclip uses different base tools to do the actual copying depending on OS and installed libraries (gtk, xsel, pbpaste, etc) is hard to know in advance what exceptions can be thrown. I could reproduce the UnicodeDecodeError on a Mac running Yosemite (10.10.3) and this solve it. (well actually just catch the exception and offers to save the content as file instead)
Regards
Marcelo

### DIFF
--- a/libmproxy/console/common.py
+++ b/libmproxy/console/common.py
@@ -252,7 +252,7 @@ def copy_flow(part, scope, flow, master, state):
     try:
         master.add_event(str(len(data)))
         pyperclip.copy(data)
-    except RuntimeError:
+    except (RuntimeError, UnicodeDecodeError):
         def save(k):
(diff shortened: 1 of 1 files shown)


mhils:
Excellent. Thanks!

By the way: If you include a "fix #531" somewhere in your commit message, the issue will be closed automatically on merge. :-)

Cheers,
Max

PR #537, opened 2015-03-25
fix #530 set client_playback to None when client_playback is done

### LINKED ISSUE #530: mitmproxy console: client playback indicator should be cleared after playback completes
Start a client playback. Observe that the client playback indicator persists after playback completes. Pressing "c" again to stop client playback does not clear the indicator either.

gato:
Hi,
I'm sending a PR for fixing #530 
Regards.
Marcelo


mhils:
:+1:

PR #538, opened 2015-03-26
mitmweb: start onboarding app if instructed to

peralta:
mitmweb is not starting the onboarding app even if commanded to. This patch fixes the issue.


mhils:
Good find. Thanks! :-)

PR #539, opened 2015-03-26
mitmweb: upgrade to react-router 0.13.2

vzvu3k6k:
react-router 0.13.2, which includes https://github.com/rackt/react-router/pull/957, is shipped.

Should I use "mitmweb" rather than "web" for the prefix of commit logs?

cc: @mhils


mhils:
Ah, excellent. Of course they introduced new "features", but at least we can work around it this time (dcb42b3016415865848cdc2607ff640c1fd50818). :smiley:

Cheers,
Max

PR #543, opened 2015-03-30
web: fix delete flows

legendtang:
Well, recent changes seemed to cause the issue. A small fix

Regards,
Wensheng Tang


mhils:
Good find. Thanks!

PR #544, opened 2015-03-31
Make the Websocket's connection header value case-insensitive

ryoqun:
Some servers respond with `"upgrade"`; so just normalize the field value.

(Just I realized that maybe, should also be done for the header name? I dunno..)


mhils:
Good find - thanks! :smile: 
Header names are already normalized by default, no need to do this manually.

Cheers,
Max

PR #547, opened 2015-04-03
Keep blank query values for flow hashing

macmantrl:
I've found that some websites use query values that are blank, e.g. foobar.com/index?thing.
The current master branch would hash '/index?thing' and '/index?other' as the same flow for server replay, which is undesirable. This patch causes these blank query values to be preserved and included when hashing.


mhils:
Good find. Thanks! :smiley:

PR #548, opened 2015-04-03
Prevent unnecessary upstream server connects

macmantrl:
Selectively connect to upstream server based on certforward and
no_upstream_cert options. When no_upstream_cert is used during server
replay, this patch prevents connecting to the upstream server unless
absolutely necessary.

### DIFF
--- a/libmproxy/protocol/primitives.py
+++ b/libmproxy/protocol/primitives.py
@@ -236,7 +236,7 @@ class LiveConnection(object):
         ssl_mismatch = (
             ssl is not None and
             (
-                ssl != self.c.server_conn.ssl_established
+                (self.c.server_conn.connection and ssl != self.c.server_conn.ssl_established)
                 or
                 (sni is not None and sni != self.c.server_conn.sni)
             )
--- a/libmproxy/flow.py
+++ b/libmproxy/flow.py
@@ -240,9 +240,13 @@ class ServerPlaybackState:
         _, _, path, _, query, _ = urlparse.urlparse(r.url)
         queriesArray = urlparse.parse_qsl(query, keep_blank_values=True)
 
+        # scheme should match the client connection to be able to replay
+        # although r.scheme may have been changed to http to connect to upstream server
+        scheme = "https" if flow.client_conn and flow.client_conn.ssl_established else "http"
+
         key = [
             str(r.port),
-            str(r.scheme),
+            str(scheme),
             str(r.method),
             str(path),
         ]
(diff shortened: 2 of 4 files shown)


macmantrl:
Well, looks like the buildbot is failing 2 tests even though I was getting all passing locally.
Seems to be related to me switching request.scheme to be based on client_conn rather than server_conn. It appears to be using SSL to the upstream reverse proxy, even though it should be sending  as HTTP (but the client is using SSL). I'll try to see what's going on.

macmantrl:
I was able to fix that issue while still getting the result that I wanted. I'm still not 100% confident that this doesn't break anything (that's not covered in a unit test)–I would appreciate someone else taking a look and recommending changes.

cortesi:
I'm just about to release mitmproxy 0.12, and this is too delicate a change to pull at this late stage. I'll look at this again just after the release.

macmantrl:
Rebased my change on mitmproxy 0.12 to fix merge conflicts and remove references to certforward (which was removed recently from mitmproxy).
Not sure what the AppVeyor failure is about–looks like that machine didn't install libxml properly. My pull request shouldn't have changed anything that would affect that.

Everything is succeeding on Travis CI.

libmproxy/protocol/primitives.py
```diff
@@ -236,7 +236,7 @@ def change_server(
         ssl_mismatch = (
             ssl is not None and
             (
-                ssl != self.c.server_conn.ssl_established
+                (self.c.server_conn.connection and ssl != self.c.server_conn.ssl_established)
```
mhils: Why is that required?

mhils:
Thanks for adjusting the PR. There are a few other places in the other proxy modes where we blindly connect upstream, which we certainly want to fix as well. This looks like a great start. Can I get a quick reply on my question above so that we can finally get this merged? :-)

Cheers,
Max

libmproxy/protocol/primitives.py
macmantrl: TL;DR this ensures that attempts to upgrade a server connection to SSL will not establish a new connection when one didn't exist before.

When ssl_mismatch is True, this code always tries to connect to the upstream server (in an attempt to correct the mismatch). My other code changes ensure that server_conn.ssl_established is always False because it hasn't connected to the server. That would mean that without this change to ssl_mismatch, it would still try to create a new connection to the server.

If I understand the rest of this method correctly, my change will only affect the case where the server is the same, it is attempting an upgrade to SSL, and doesn't already have a server connection.

macmantrl:
Replied to question.

Yeah, it would be great to fix the other modes too, but thankfully (for me) I don't use those modes.

mhils:
Merged, thanks!

Issue #554, opened 2015-04-11
Running mitmdump on Windows from the Path requires file extension


elitest:
I have been working on user installation documentation in regards to #516 and I'm recommending that when users install Python they "Add Python.exe to Path" which adds "C:\Python27\" and "C:\Python27\Scripts\" to %PATH%.  Then the user installs mitmproxy using "pip install mitmproxy".  Then when you try to run "mitmdump", it doesn't find it.  Turns out this is due to mitmdump being installed to "C:\Python27\Scripts\mitmdump" without a ".py" file extension.

The weird thing is that when I pull from master, and run setup.py manually mitmdump(and mitmweb) gets properly turned into an exe.  I looked through setup.py and there isn't anything obvious to me that is doing this.  Was this fixed on master and we are just waiting for 0.11.4 to be released and to be loaded onto PyPI or is pip introducing something into the mix?

mhils:
Hi @elitest,

the exe is generated using setuptools entry_points, introduced in 355f9fc40799c88f413ca260dd93691074fd758e. 0.11.4 should just generate executables :smiley:

Cheers,
Max

Issue #398, opened 2014-11-04
Document installation of dependencies.


mhils:
We have a variety of bug reports wrt. the installation procedure. While installing mitmproxy itself usually doesn't cause any issues, our dependencies often do ( #396, #379, ...).

We should document installation instructions for Linux, OSX and Windows, both for binary distributions as well as for installing from source.

refs: 
- https://gist.github.com/mhils/9315651
- #394
- https://groups.google.com/d/msg/mitmproxy/SoNVziqow3g/3dELw1BNjQsJ

elitest:
Having worked in Ubuntu and Windows today re:#555, I'll give some feedback on this.  Are these reports a problem with pip, or are they problems with the libraries we rely on(i.e. something they need to fix), or are they a conflict in dependencies due to people installing conflicting stuff or not updating stuff with pip?

I find pip very fragile on Ubuntu, and the only reliable way I have found to update it is to use apt, rather then self-updating using pip.  I also find not working of virtual environments troublesome on Ubuntu due to the wide range of python stuff already installed, due to the risk of pip clobbering something.

I was pleasantly surprised by how well it seemed to work on Windows.  Now that pip is shipping with Python, perhaps just updating python will be the best way to update pip?

Most of the issues you reference are issues on OS X, I'm guessing due to the weird kludge of different version you can get yourself into on that platform.  Perhaps the binary installer for end users is better than using pip on this platform.

mhils:
> Having worked in Ubuntu and Windows today re:#555, I'll give some feedback on this. Are these reports a problem with pip, or are they problems with the libraries we rely on(i.e. something they need to fix), or are they a conflict in dependencies due to people installing conflicting stuff or not updating stuff with pip?

The most common issues are old versions of pyOpenSSL lying around (see [here](https://github.com/mitmproxy/mitmproxy/blob/923503260ed48510da63e4dd1d6b36b4b859b110/libmproxy/main.py#L30-L48)) and missing dependencies (pip fails with a very obscure message if compilation of a binary module fails).

Given the fabolous state of Python packaging, we should always prefer binaries over manual installation. Maybe we could simplify the whole thing it as follows:

## Windows

`pip install mitmproxy` as the primary means (until we have Windows binaries). As you already described, this works pretty flawless.

### Installing from Source

Instructions for installing mitmproxy from source can be found in the README: https://github.com/mitmproxy/mitmproxy#hacking

## OSX

[...]

PR #555, opened 2015-04-12
User Installation Documentation update

elitest:
See the countershaped output [here](http://secstuff.org/userdoc/install.html)

This PR includes things like:
* de-emphasizing installing pip using the get-pip.py installation method, as Python 2.7.9 ships with pip.
* Linked to developer docs
* I didn't really touch the OS X instructions as I don't have a good way to test those, but if anyone has any comments on them let me know.
* Clarified Ubuntu installation instructions.  Tested on a fresh install of Ubuntu 14.04
* Created Windows installation instructions.  The instructions don't strictly work with the current mitmproxy on PyPI but will when 0.11.4 ships(which I assume is the next time these docs will be updated).  See #554 for more on that.


elitest:
ref:#398

doc-src/install.html
```diff
@@ -43,17 +38,32 @@
...
+Install the latest version of Python 2 from the <a href="https://www.python.org/downloads/windows/">Python website</a>.  The Windows MSI installer is fairly straightforward and even includes pip as long as you are installing a version of Python 2 after 2.7.9.  For the most part, you can accept the defaults but one option you should change during installation is "Add python.exe to Path".  This will make it easier to run Python.
+<img src="@!urlTo('winpythoninstaller.jpg')!@"/>
 
-- build-essential
-- python-dev
-- libffi-dev
-- libssl-dev
-- libxml2-dev
-- libxslt1-dev
+Now log out of Windows and back in to finish having Python added to your PATH.
```
mhils: I think it's sufficient to re-open the terminal nowadays.

doc-src/install.html
```diff
@@ -43,17 +38,32 @@
...
 </pre>
 
+From now on you can run <a href="mitmproxy.html">mitmproxy</a> or <a href="mitmdump.html">mitmdump</a> from a terminal.
+
+## <a id=docUbuntu></a>Installation On Ubuntu
+
+Ubuntu comes with Python but we need to install pip, python-dev and several libraries.  This was tested on a fully patched installation of Ubuntu 14.04.
+
+<pre class="terminal">
+$ sudo apt-get python-pip python-dev libffi-dev libssl-dev libxml2-dev libxslt1-dev
```
mhils: + install

doc-src/install.html
elitest: Terminal?  On Windows?  Not based on my testing, even did a "/taskkill /IM explorer.exe /F".  Couldn't find anything current on the issue, but [this](https://support.microsoft.com/en-us/kb/104011) says:
>However, note that modifications to the environment variables do not result in immediate change. For example, if you start another Command Prompt after making the changes, the environment variables will reflect the previous (not the current) values. The changes do not take effect until you log off and then log back on.

mhils:
:+1:

PR #561, opened 2015-04-23
Check overwrite, when save to file instaed of clipboard

tunz:
When save a content as a file instead of saving to clipboard, it asks if overwrite or not.


mhils:
Thanks!

PR #568, opened 2015-04-25
Fix toggle eventlog

tunz:
There are two changes.

1. When we toggle eventlog, it makes new windows at every toggle.

   If we toggle eventlog 5 times, then we have to press 'q' 5 times to quit program.

   So, just pop the new window, when a user toggles eventlog again.

2. I'm not sure that it is better change.

   When we toggle eventlog, it resets cursor position.

   I think it is better to keep cursor position.


mhils:
Good find, thanks! :smiley:

PR #569, opened 2015-04-26
Match all freebsd versions and not just freebsd10

mike-pt:
+ Use sys.platform.startswith("freebsd") instead of matching just freebs...
+ This means support for any freebsd version (note that I only tested 11-CURRENT and 9)

### DIFF
--- a/libmproxy/platform/__init__.py
+++ b/libmproxy/platform/__init__.py
@@ -8,7 +8,7 @@ if sys.platform == "linux2":
 elif sys.platform == "darwin":
     from . import osx
     resolver = osx.Resolver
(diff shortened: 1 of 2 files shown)


mhils:
Should be `.startswith("freebsd")` in both cases, no?

mike-pt:
Yes indeed.  That's what one gets for doing stuff in a hurry..  I'll fix that when I'm back at the laptop. 

Sorry

mike-pt:
Ok, just made a git commit --amend, the PR should be correct now.

mhils:
Great - thanks a lot! :smiley:

PR #576, opened 2015-05-08
Accurately timestamp start of request

nickraptis:
When building a request from a stream, try to get an accurate
start timestamp from the Reader. This was already in the code
and also used when building response objects, but was omitted
in commit ddf458b330bf9fe200cb1dbc3ddb5ae1a5d2102a

Without his logic and when the client is reusing a connection
to send requests, the timestamp_start of subsequent requests
is early and equal to when the connection started read blocking


mhils:
Thanks again! :smiley:

PR #580, opened 2015-05-12
windows 7 certificate installation docs update

jbremer:


cortesi:
Hi Jurriaan. Thanks for the patch. A few moments ago, I merged a documentation patch that removes the specific platform instructions in our documentation, in favour of pointers to canonical documentation elsewhere on the web. For Windows generally, we now have a pointer to this page:

http://windows.microsoft.com/en-ca/windows/import-export-certificates-private-keys#1TC=windows-7

Is this accurate for Windows 7, or is the command-line technique in this patch the only way? Perhaps we should also include a pointer to the certutil page:

https://technet.microsoft.com/en-us/library/cc732443.aspx

Kriechi:
@jbremer ping - any update from your side?

jbremer:
Well, do as you like of course, but for people that want to automate it (like me), the certutil trick is neat.
Up to you whether a link to the certutil page is "good enough" instead of just pointing out the exact usage in the docs ;)

mhils:
Fixed: http://docs.mitmproxy.org/en/latest/certinstall.html#windows-automated

PR #584, opened 2015-05-16
Make sure proxy returns the httpversion specified in the request

dlethin:
rather than hardcoding to 1.1.

This helped me to work around the scenario where I was testing some old code that broke because it was expecting to receive 'HTTP/1.0 200 Connection established'


cortesi:
Looks good! Merged.

PR #585, opened 2015-05-16
docs: Single Qoute for ignore pattern with "!"

TearsDontFalls:
because wouldnt work with default bash settings as its part of history expansion in bash


cortesi:
Thanks!

Issue #588, opened 2015-05-19
Cipher selection snafu on server that only accepts SSLv3 with RC4-SHA


aGHz:
### TL;DR

We need to specify `--ciphers-server 'RC4-SHA'` if we want to connect to this IIS server, `'ALL'` causes `ECONNRESET` when reading from the stream. However, using `'ALL'` in `openssl s_client` is able to negotiate into RC4-SHA just fine.

Not entirely sure if there's any problem, as netlib seems to just pass the cipher list to OpenSSL, but I'm asking here in case anyone has any idea that I can't see.

I'm also posting here instead of netlib because if there's no clear solution (we really don't want to force RC4-SHA on all our connections), I'm planning to make a PR to enable per-host cipher lists in the same way we have per-host client cert files. So I want to ask what you think about that

### Details

We're using mitmdump to collect all traffic between our server and third-party servers, and we've encountered a strange situation. Unfortunately I can't share details but I'll try to provide complete diagnostics and can also perform any test you may need.

The remote is an IIS server that seems to understand nothing but SSLv3 with the RC4-SHA cipher. We've arrived at this conclusion using a combination of brute-forcing mitmdump parameters and `openssl s_client`.

So we enabled SSLv3 with `mitmdump --ssl-version-server all`, and then we could at least connect to the remote server. However, we're getting `ECONNRESET` when we try to read from the connection:
```
<ip>:60920: clientconnect
<ip>:60920: request
  -> CONNECT <remote domain>:443 HTTP/1.1
<ip>:60920: Set new server address: <remote domain>:443
<ip>:60920: serverconnect
  -> <remote domain>:443
------ setting ciphers to: ALL
<ip>:60920: Received CONNECT request to SSL port. Upgrading to SSL...
<ip>:60920: Establish SSL
  -> with client
  -> with server (sni: None)
<ip>:60920: SNI received: <remote domain>
<ip>:60920: (server reconnect follows)
<ip>:60920: serverdisconnect
  -> <remote domain>:443
<ip>:60920: Set new server address: <remote domain>:443
<ip>:60920: serverconnect
  -> <remote domain>:443
------ setting ciphers to: ALL
<ip>:60920: Establish SSL
  -> with server (sni: <remote domain>)
<ip>:60920: Upgrade to SSL completed.
<ip>:60920: request
  -> GET <path> HTTP/1.1
---------- server_conn.send
---------- server_conn.send done
~~~ from_stream
~~~ read_response
***** read_response
[...]

PR #590, opened 2015-05-20
Added --per-host-ciphers-dir option

aGHz:
This adds the ability to specify allowed ciphers on a per-server basis in a similar way to the use of client certificates (see #588 for motivation).

### DIFF
--- a/libmproxy/proxy/config.py
+++ b/libmproxy/proxy/config.py
@@ -47,6 +47,7 @@ class ProxyConfig:
             tcp_hosts=[],
             ciphers_client=None,
             ciphers_server=None,
+            per_host_ciphers_dir=None,
             certs=[],
             certforward=False,
             ssl_version_client="secure",
@@ -58,6 +59,7 @@ class ProxyConfig:
         self.server_version = server_version
         self.ciphers_client = ciphers_client
         self.ciphers_server = ciphers_server
+        self.per_host_ciphers_dir = per_host_ciphers_dir
         self.clientcerts = clientcerts
         self.no_upstream_cert = no_upstream_cert
         self.body_size_limit = body_size_limit
@@ -192,6 +194,7 @@ def process_proxy_options(parser, options):
         authenticator=authenticator,
         ciphers_client=options.ciphers_client,
         ciphers_server=options.ciphers_server,
+        per_host_ciphers_dir=options.per_host_ciphers_dir,
         certs=certs,
         certforward=options.certforward,
         ssl_version_client=options.ssl_version_client,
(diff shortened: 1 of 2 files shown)


elitest:
So if I understand this right I would specify a folder which would contain something like ``google.com.cipher``?  In those would be something like ``RC4-SHA:AES128-SHA256``?  Does this work with multiple sites in the directory?

My only other comment with this is that looking at the code this only determines the ciphers used on one side of the proxy.  If I remember how this part of the code works you are setting the ciphers that the client-facing side of the proxy uses to figure out what cipher to send back in the SERVERHELLO.  This does not include what mitmproxy sends to the upstream server, ie. what is specified in the server-facing CLIENTHELLO.  I think it would be better to specify what side of the proxy it is on in the flag and mention it in the help text as well.

aGHz:
Thanks for the comments! You are correct and yes it would work with any sites you put in there. When it tries to connect to a host, it looks in that dir to see if there's an appropriately named file and uses its content to override the generic `--ciphers-server` option.

I must admit my understanding of the code isn't top shape, but from what I gather `ConnectionHandler.server_conn` is the connection **to** the server (i.e. presumably the one where mitmproxy says CLIENTHELLO, I'm not familiar with TLS specifics). So this code would set the ciphers on the connection to the upstream server.

I didn't think it would be useful to add per-client cipher configuration because typically you can control what the client does, and in any case it would be hard to identify the client unless it has a static IP.

elitest:
Ah gotcha, I guess I got mixed up then.

Being able to control the ciphers that my client connects to mitmproxy with is certainly as useful as setting it on the upstream connection.  I find it is often far easier and more temporary to set the cipher through mitmproxy than in the config of a client.  Also not every client is as flexible as a web browser(if you even consider them flexible) in terms of being able to turn off and on ciphers.  I could also specify deprecated or disabled ciphers if I'm compiled against the right version of OpenSSL.

Yeah I find the way we worded ``--ciphers-server`` and ``--ciphers-client`` confusing, as the person who contributed some of that code I have trouble keeping it straight in my head..  Maybe ``--ciphers-pre-proxy`` and ``--ciphers-post-proxy`` would have been better.

Kriechi:
I don't like the idea of using folders and files to specify ciphers...
Did you consider a simple config file, like yaml or json to specify a list of domains + ciphers?
This could probably go int `~/.mitmproxy` with all the other config stuff...

mhils:
We should probably just provide inline script hooks which allow modification of the SSL context.

mhils:
Thanks a lot for the PR @aGHz - As I mentioned in my last comment, this sounds like a feature that's very suitable for an inline script. I'd be happy to accept a PR that exposes the SSL Context to scripts, but I think we don't want to add another command line switch for this. Sorry!

Thanks!
Max

PR #591, opened 2015-05-21
improve autoenv script

Kriechi:
* properly quote DIR variable
  (it might contain spaces)
* use builtin string magic instead of `dirname`


mhils:
:+1:

PR #600, opened 2015-05-31
Print consistency

elitest:
Just a cleanup PR, bringing some consistency to how stuff is printed.  Also helps make stuff more Python 3 'ready'.  I didn't add the code to backport print() to 2.6 as I believe mitmproxy only supports 2.7.  Let me know if this is something we should add anyway.

There aren't a few other weird [examples](https://github.com/mitmproxy/mitmproxy/blob/master/libmproxy/filt.py#L264-L290) of print not behaving in a future friendly way.  I'll see if I can figure out a way to fix them in an elegantly.


cortesi:
Hey Paul. This looks good, but could you please exclude the stuff in libmproxy/contrib. We're tracking upstream to some extent in those libraries, and formatting changes like this will make it devilishly hard to tell what's changed.

mhils:
Thanks!

PR #603, opened 2015-05-31
replace contrib libraries with pypi dependencies

### LINKED ISSUE #602: Remove libmproxy.contrib
All the dependencies we have included in libmproxy.contrib are now on pypi as well. We should ditch libmproxy.contrib and list them as regular requirements instead.
Most of the libraries provide newer versions as well - we should test them well to not break anything.

Kriechi:
fixes #602

* jsbeautifier: not sure if the change made by @cortesi is still needed
* **our** wbxml is not available on pypi (maybe under a different name?)
  * https://github.com/davidpshaw/PyWBXMLDecoder


cortesi:
Careful with the jsbeautifier change.  The issue I was side-stepping is [here](https://github.com/mitmproxy/mitmproxy/blob/master/libmproxy/contrib/jsbeautifier/__init__.py#L248-L254). The way the packers are imported and used means that the pyinstaller dependency analysis fails to pick up the packers, which then causes a runtime failure. 

There are a few potential answers to this. First, we could amend this patch to explicitly list the packer files for packaging in the pyinstaller configs. I avoided this in the past, because it introduces a tricky corner case that can't be detected in our test suite and is hard to catch when rolling releases - if jsbeautifier ever introduced another packer, packaged mitmproxy would fail while viewing JS files at runtime.  Second, we could find an alternative library that does this (I haven't looked in ages) and doesn't do crazy dynamic imports. Third, we could leave jsbeautifier as is. 

The fourth option, of course, is to get rid of PyInstaller. I think trying to build a dependency tree for packages through analysis is absolutely nuts. This is responsible for 100% of the issues we see with PyInstaller again and again, constitutes the majority of the PyInstaller codebase, and provides (for most use cases) almost 0 incremental benefit over a system that just packages all specified dependencies with no cleverness. This is the system used by [pex](https://github.com/pantsbuild/pex) to produce a self-contained package. Unfortunately pex still relies on an existing Python interpreter on the target system to execute the package. I'd love to find a way to either combine pex with PyInstaller, or just to create a better binary distribution mechanism from scratch. I have a light experiment running that moves in this direction by using Rust for the executable wrapper. I'll explore further in my "copious free time".

cortesi:
Thumbs up to the broad thrust of this change, though. Getting rid of contrib would be nice.

Kriechi:
Let's take small steps - I have re-added jsbeautifier for now.
Now the only libs that are replaced by pypi dependencies are:
* html2text
* pyparsing

mhils:
LGTM. Let's throw that in once 0.12.1 is out! :+1:

PR #616, opened 2015-06-07
Force 0px height on main-view to restrain height to viewport

### LINKED ISSUE #615: Unable to scroll in mitmweb flow-table or flow-detail panes
Currently its not possible to scroll either the in the flow-table or flow-detail panes. Observing the same behaviour in Chrome, Firefox and IE.

Been observing this in 0.12.0 and 0.12.1

no_scroll

johnmccabe:
Looks like the problem is down to the height of main-view not being restrained to the viewport, I've attached a hacky fix to resolve #615 (theres probably a React way of fixing this).

The height with a build from master is larger than the viewport (depends on the size of the table or detail, whichever is bigger).  (see the size in the bottom right)
mitmweb_invalid_height

And with the hacky fix you can see its constrained to the view port
mitmweb_0px_height

### DIFF
--- a/libmproxy/web/static/app.css
+++ b/libmproxy/web/static/app.css
@@ -61,6 +61,7 @@ body,
   flex: 1 1 auto;
   display: flex;
   flex-direction: row;
+  height: 0px;
 }
 .main-view.vertical {
   flex-direction: column;


mhils:
Fixed in d389b9c59d7f4a4918a15b5a40d771ccc154d751. Thanks!

PR #619, opened 2015-06-09
Add copy as curl command

### LINKED ISSUE #596: Export as curl command
It would be very neat to have functionality in the UI to export as a curl command for a given http request

Fuzion24:
This does the trick for me, but might be a bit hacky.  Improvements/suggestions are welcome

fixes #596

### DIFF
--- a/libmproxy/console/common.py
+++ b/libmproxy/console/common.py
@@ -194,7 +194,7 @@ def raw_format_flow(f, focus, extended, padding):
 
 
 # Save file to disk
-def save_data(path, data, master, state):
+def save_data(path, data):
     if not path:
         return
     try:
@@ -204,32 +204,32 @@ def save_data(path, data, master, state):
         signals.status_message.send(message=v.strerror)
 
 
-def ask_save_overwite(path, data, master, state):
+def ask_save_overwrite(path, data):
     if not path:
         return
     path = os.path.expanduser(path)
     if os.path.exists(path):
-        def save_overwite(k):
+        def save_overwrite(k):
             if k == "y":
-                save_data(path, data, master, state)
+                save_data(path, data)
 
         signals.status_prompt_onekey.send(
-            prompt = "'" + path + "' already exists. Overwite?",
+            prompt = "'" + path + "' already exists. Overwrite?",
             keys = (
                 ("yes", "y"),
                 ("no", "n"),
             ),
-            callback = save_overwite
+            callback = save_overwrite
         )
     else:
-        save_data(path, data, master, state)
+        save_data(path, data)
 
 
-def ask_save_path(prompt, data, master, state):
+def ask_save_path(prompt, data):
     signals.status_prompt_path.send(
         prompt = prompt,
-        callback = ask_save_overwite,
-        args = (data, master, state)
+        callback = ask_save_overwrite,
+        args = (data)
     )
 
 
@@ -300,7 +300,7 @@ def copy_to_clipboard_or_prompt(data):
     except (RuntimeError, UnicodeDecodeError, AttributeError):
         def save(k):
             if k == "y":
-                ask_save_path("Save data", data, master, state)
+                ask_save_path("Save data", data)
         signals.status_prompt_onekey.send(
             prompt = "Cannot copy data to clipboard. Save as file?",
             keys = (
@@ -379,16 +379,12 @@ def ask_save_body(part, master, state, flow):
     elif part == "q" and request_has_content:
         ask_save_path(
             "Save request content",
-            flow.request.get_decoded_content(),
-            master,
-            state
+            flow.request.get_decoded_content()
         )
     elif part == "s" and response_has_content:
         ask_save_path(
             "Save response content",
-            flow.response.get_decoded_content(),
-            master,
-            state
+            flow.response.get_decoded_content()
         )
     else:
         signals.status_message.send(message="No content to save.")
--- a/libmproxy/console/common.py
+++ b/libmproxy/console/common.py
@@ -270,7 +270,7 @@ def copy_as_curl_command(flow):
 
     headerString = ""
     for k,v in flow.request.headers:
-      headerString += " -H \"" + k + ":" + v + "\" "
+      headerString += " -H '" + k + ":" + v + "' "
 
     data = "curl"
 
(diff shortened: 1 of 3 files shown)


libmproxy/console/common.py
```diff
@@ -263,6 +263,42 @@ def copy_flow_format_data(part, scope, flow):
...
+    headerString = ""
+    for k,v in flow.request.headers:
+      headerString += " -H \"" + k + ":" + v + "\" "
+
+    data = "curl" + " -X " + flow.request.method  + headerString + flow.request.scheme + "://" + flow.request.host + flow.request.path
+
+    if flow.request.content != None and flow.request.content != "":
+      data += " -d " + "'" + flow.request.content + "'"
+
+    # pyperclip calls encode('utf-8') on data to be copied without checking.
```
mhils: Can we extract everything below this line in a separate function which is shared between `copy_flow` and `copy_as_curl_command`?

libmproxy/console/flowview.py
```diff
@@ -525,6 +526,8 @@ def keypress(self, size, key):
                 callback = self.master.save_one_flow,
                 args = (self.flow,)
             )
+        elif key == "z":
+            common.copy_as_curl_command(flow)
```
gato: here is self.flow

libmproxy/console/common.py
```diff
@@ -263,6 +263,42 @@ def copy_flow_format_data(part, scope, flow):
...
+def copy_as_curl_command(flow):
+
+    if flow.request.content is None or flow.request.content == CONTENT_MISSING:
+        return None, "Request content is missing"
+
+    headerString = ""
+    for k,v in flow.request.headers:
+      headerString += " -H \"" + k + ":" + v + "\" "
+
+    data = "curl" + " -X " + flow.request.method  + headerString + flow.request.scheme + "://" + flow.request.host + flow.request.path
```
gato: full request path should be quoted, because & (param separators) and ?(for query part) will be misinterpreted by the shell (probably # for anchor will be too)

Fuzion24: I thought about this a bit.  You're probably right, we can quote it, but i was testing a curl command with query params and when copy and pasting into the shell, it auto escapes '=' and '?' properly

Fuzion24:
I cleaned up some of these requests

libmproxy/console/common.py
gato: I've tried on ubuntu and doesn't and i do not have any auto escape, and if you paste in a script file you won't have it either.

gato:
i guess "Z" is free

Fuzion24:
Changed control command to "Z".  Quoted the full url path.  Changed to --data-binary.  I don't think we need to worry about content-type, we can always assume that whatever we received from mitmproxy was properly encoded and just dump it into --data-binary?

gato:
I guess using --data-binary for all request should be ok.

cortesi:
Hi there. Good work, and a very handy addition. We're going to sprout a number of these export functions over the next while - I'd like a HAR exporter, and a "raw" exporter that just writes the headers and body to file as it would on the wire. I don't want to add a separate keyboard shortcut for each of these, though, so can we please:

- Add an "export" command bound to the key "E".
- Prompt for the export format, in the same style as the edit request prompt (lower-case "e" in the request view). 
- Then prompt for the filename to export to.

That keeps the bindings manageable, and gives us an easy-to-remember mnemonic.

cortesi:
@Fuzion24 Ryan, just a quick ping to make sure you saw my comment on this pull request (as opposed to the original ticket). Let me know if you're still working on this?

Fuzion24:
This feature seems to work well, when a clipboard is available. When it is not, there seems to be some issues. It seems there may be some broken code (prior to me touching this) in the case a clipboard isn't available.  I don't have an easy way to test this bit.

This probably requires closer investigation by someone more familiar with the code.  @gato may have been taking a look?

gato:
@Fuzion24 Ryan, I'm kind of busy now, but will try to take a look at this in the next couple of days.
Sorry for the delay.

babakgh:
:+1:

mhils:
What's the status over here? :smiley:

gato:
Hi Max,
Sorry, but i had a couple of really complicated months and couldn't do
anything about this. if no one gets it, i will try to take a look at this
and the current status of the proyect next week.
Regards.
Marcelo


On Wed, Aug 26, 2015 at 8:31 PM, Maximilian Hils <notifications@github.com>
wrote:

> What's the status over here? [image: :smiley:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mitmproxy/mitmproxy/pull/619#issuecomment-135205004>.
>

-- 
Antes de imprimir, piense en su responsabilidad y compromiso con el medio 
ambiente.  
La información contenida en esta comunicación es confidencial, podrá ser 
objeto de privilegios legales, puede constituir una información reservada, 
y está destinada sólo para el uso del destinatario. Es propiedad de TEKii 
S.R.L.. El uso no autorizado, divulgación o copia de esta comunicación o de 
cualquier parte de la misma está estrictamente prohibida y puede ser 
ilegal. Si usted ha recibido esta comunicación por error, por favor 
notifique inmediatamente a TEKii S.R.L. contestando este correo electrónico 
y destruya esta comunicación y todas sus copias, incluyendo todos los 
archivos adjuntos.

Before printing, keep in mind your responsibility and commitment with the 
environment.
The information contained in this communication is confidential, may be 
subject to legal privileges, may constitute inside information, and is 
intended only for the use of the addressee. It is the property of TEKii 
S.R.L.. Unauthorized use, disclosure or copying of this communication or 
any part thereof is strictly prohibited and may be unlawful. If you have 
received this communication in error, please notify TEKii S.R.L. 
immediately by return e-mail and destroy this communication and all copies 
thereof, including all attachments.

mhils:
:+1:

monperrus:
this would be an awesome feature

tunz:
Wow... I didn't realize that I was making five consecutive 'overwite' typo.. Besides, I made the 'overwite' typo again while typing this comment!! I have to fix this habit.

Anyway, this feature may be useful for me.

mhils:
Fixed by #884! :smiley:

Issue #623, opened 2015-06-11
Easier navigation to "favorite" flows


drahosj:
It is common to want to save three or four flows for replays, such as JSON POST requests, for frequent use when using MITMProxy for web application debugging or analysis. The application, however, may be generating a lot of other traffic. This makes it difficult to find or navigate to the "favorite" flows.

Currently, the best solution is to write the desired flows to a file, clear the screen, and reload the flows. This can become quite cumbersome, and it is very inconvenient to add another saved flow to the file.

There are several possibilities. The easiest would be to allow vim-like navigation to a specific flow number by pressing a hotkey, then typing the flow number. A more complete option would be to allow flows to be marked as a "favorite" This would place a visual marker on the flow, and enable additional navigation options, such as "next favorite flow" and "previous favorite flow". Additionally, adding a "favorite flows" option to the "write flows to file" submenu would allow for an easy way to persist the "favorite" flows across clears, or the "Clear All Flows" command could default to not deleting favorite flows.

Overall, these features would vastly improve the experience of using MITMProxy for web application debugging, especially for debugging of applications which make frequent requests to REST endpoints.

drahosj:
This was actually rather easy to implement. I created functionality for toggling a mark on flows and persisting those flows across clears and submitted a pull request here: https://github.com/mitmproxy/mitmproxy/pull/624

If this is popular, I can work on extending this functionality a bit more, but what I have so far has been a lifesaver for the project I'm working on right now.

cortesi:
I've merged @drahosj's change that implements this. Thanks!

PR #624, opened 2015-06-11
Added flow marking functionality in the console

drahosj:
Some functionality in https://github.com/mitmproxy/mitmproxy/issues/623 was actually fairly trivial to implement. The following features are included

In the flow list view:
- Press 'm' to toggle an easily visible mark on the current flow.
- When the flow list is cleared with 'C', marked flows are not deleted.
- If all flows in the list are marked, marked flows are deleted with 'C'

This means that pressing C twice will clear all flows, including marked ones. 

The mark is made large to stand out while scrolling, but that can be changed by changing SYMBOL_MARK in libmproxy/console/common.py.

### DIFF
--- a/libmproxy/console/__init__.py
+++ b/libmproxy/console/__init__.py
@@ -48,6 +48,7 @@ class ConsoleState(flow.State):
             self.set_focus(0)
         elif self.follow_focus:
             self.set_focus(len(self.view) - 1)
+        self.set_flow_marked(f, False)
         return f
 
     def update_flow(self, f):
@@ -100,12 +101,29 @@ class ConsoleState(flow.State):
         return ret
 
     def clear(self):
+        marked_flows = []
+        for f in self.flows:
+            if self.flow_marked(f):
+                marked_flows.append(f)
+                
         super(ConsoleState, self).clear()
+        
+        for f in marked_flows:
+            self.add_flow(f)
+            self.set_flow_marked(f, True)
+        
         if len(self.flows.views) == 0:
             self.focus = None
         else:
             self.focus = 0
         self.set_focus(self.focus)
+        
+    def flow_marked(self, flow):
+        return self.get_flow_setting(flow, "marked", False)
+    
+    def set_flow_marked(self, flow, marked):
+        self.add_flow_setting(flow, "marked", marked)
+        
 
 class Options(object):
     attributes = [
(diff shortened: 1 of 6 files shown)


cortesi:
Hi there. A very useful patch! I'd like one change before I merge this. I want to keep the protocol definitions as clean as possible, dealing only the actual protocols, and not with data used by the various front-ends. As it happens, mitmproxy console already has a mechanism for keeping "per flow" settings in a clean way - look at the flow settings functions [here](https://github.com/mitmproxy/mitmproxy/blob/master/libmproxy/console/__init__.py#L37-L43).

So, could you please:

- Remove the changes to libmproxy/protocol/primitives.py
- Move the tagging into a flow setting on the console app state instead

After that, I'll merge, and we can chat about how to take the functionality forward. I'm hesitant about having the C shortcut effectively mean two different things depending on a context that can change very rapidy in a live proxy session. I think we'd be better served with an "unmark all" shortcut or something similar. But these are details we can discuss once we have this in base. Thanks!

drahosj:
I made the changes you requested and moved all of the marking functionality out of the primitive and into the console. I also implemented "unmark all" on U instead of double-clear. It does make more sense to work that way.

Complete list of marking functionality:
- m in FlowList to toggle mark on the flow
- U in FlowList to unmark all flows
- C clears all unmarked flows, marked flows remain
- w now has an 'm' option to write all marked flows to a file
- When a marked flow is duplicated, the new copy will not be marked. This facilitates an easy duplicate, modify, replay workflow.

cortesi:
Excellent, thanks!

cortesi:
@drahosj: If you wanted to continue to polish the marking implementation, there are a few improvements that could be made. 

First, it would be useful to be able to mark/unmark from within a flow view. We should use the same keybinding, so "m" would do there too. At the moment, "m" is used to change the display mode in that view - we can move that to "M" instead. The help document erroneously claims that "M" is used to change the default display mode - that's now actually done in the options screen.

Second, we need a "marked" indicator in the flow view itself as well. 

Third, I don't love the fact that the marker is pre-pended in the flow list. It changes the alignment, and makes things jump around. I'd prefer that to be stable. I don't have a firm suggestion for how to do this - we could sacrifice a column to have the spacing there for all flows, we could integrate it into the current flow indicator column somehow. Something to think about.

Thanks again for the contribution - very handy indeed.

PR #638, opened 2015-06-18
HTTP Transparent Proxy

iroiro123:
Added http transparent proxy mode.
Unlike transparent proxy mode, this mode uses the Host header to connect to a server.

```
$ mitmproxy --http-transparent -p 8080

$ nc localhost 8080
GET / HTTP/1.1
Host: www.example.com

HTTP/1.1 200 OK
Content-Type: text/html
Content-Length: 1270
...
```

Example of use:
Host A (192.168.100.2) ... using mitmproxy
```
$ mitmproxy --http-transparent -p 8080
```
Host B (192.168.200.1) ... routing ip packets, redirecting tcp:http packets to A
```
$ sysctl -w net.ipv4.ip_forward=1
$ iptables -t nat -A PREROUTING -i eth1 -p tcp --dport 80 -j DNAT --to-destination 192.168.100.2:8080
```

### DIFF
--- a/libmproxy/cmdline.py
+++ b/libmproxy/cmdline.py
@@ -363,17 +363,17 @@ def common_options(parser):
         help="Set transparent proxy mode."
     )
     group.add_argument(
-        "-H", "--http-transparent",
-        action="store_true", dest="http_transparent_proxy", default=False,
-        help="Use the Host header to connect to server."
-    )
-    group.add_argument(
         "-U", "--upstream",
         action="store",
         type=parse_server_spec,
         dest="upstream_proxy",
         default=None,
         help="Forward all requests to upstream proxy server: http://host[:port]"
+    )
+    group.add_argument(
+        "--spoof",
+        action="store_true", dest="spoof_mode", default=False,
+        help="Use Host header to connect to HTTP server."
     )
 
     group = parser.add_argument_group(
--- a/libmproxy/cmdline.py
+++ b/libmproxy/cmdline.py
@@ -363,6 +363,11 @@ def common_options(parser):
         help="Set transparent proxy mode."
     )
     group.add_argument(
+        "-H", "--http-transparent",
+        action="store_true", dest="http_transparent_proxy", default=False,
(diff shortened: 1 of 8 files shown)


byt3bl33d3r:
@iroiro123 this could be something that would be very useful for me in the near future, using this mode you can directly connect to the proxy or forward packets to it correct?

mhils:
Thanks for the interesting PR. We currently provide similar functionality using an inline script  ([dns_spoofing.py](https://github.com/mitmproxy/mitmproxy/blob/master/examples/dns_spoofing.py)), so I'm not sure if we want to include that in the core directly. @cortesi, thoughts?

iroiro123:
Yes, if I correctly understand the English...
This mode operates as a transparent forward proxy.
It is like virtual hosts of an http server.

cortesi:
@iroiro123 Thanks for the patch. I definitely think this feature is useful enough to be one of our base supported modes. A few comments:

- We should use SNI information when available, as in the script linked by Max (though perhaps not by using the pretty_host function, which is intended to be for presentation, as the name implies).
- If we do this, calling the feature "HTTP transparent" doesn't work anymore. I suggest "Spoof Mode", and we should change the class and command-line flag accordingly.
- Could you please include unit tests for the feature?

Thanks again for your work - with the changes above, this will be a very useful addition.

iroiro123:
Thank you for your advice. I will try to improve my patch.

iroiro123:
I implemented HTTP/HTTPS spoof mode as separated options. If my English in help messages is wrong, please correct it. And it takes probably some time to add test codes because I'm a python beginner.

cortesi:
@iroiro123: This is looking good, thank you (and the English in the help messages is faultless). Any chance you could add some unit tests for this functionality? We're trying to keep the test coverage on the project up. We've let it slip a bit for the latest release, which is something I want to address for the next iteration.

iroiro123:
I managed to add unit tests for this functionality.

cortesi:
Excellent, thank you very much.

PR #661, opened 2015-06-29
Enabling upstream server verification.

kyle-m:
This PR adds 3 new flags:
--verify-upstream-cert      ==> flag that enables strict upstream certificate validation
--upstream-trusted-ca      ==> takes a path to a pem formatted file containing a trusted CA certificate
--upstream-trusted-cadir  ==> takes a path to a directory of trusted CA certificates created by c_rehash

These flags enable upstream server verification to be enabled and verification logic to be tweaked. Both of the trusted ca flags are completely optional (by default the certifi library is used in netlib).

If certificate verification is enabled errors will abort the SSL handshake and print both the certificate chain depth and error number which identify the issue to error logs.

If it is not enabled then the same message is printed to error logs however the SSL handshake will continue as normal.

No work was done to optimize for the interactive mitmproxy, although the certificate error information is now stored as a dictionary on the ServerConnection object as ssl_verification_error so the necessary information is present for whoever wants to take the next steps of displaying it more informatively.


mhils:
This looks like a very strong start. :+1:
LGTM from my end. Ultimately, we should aim to enable cert verification by default. :smiley:

cortesi:
Excellent. For what it's worth, I don't think enabling cert verification by default is the way to go. I think "warn by default, error optionally" probably better serves our users, the majority of whom still use mitmproxy in assessment scenarios where dropping connections on failure is probably not the desired outcome.

PR #675, opened 2015-07-12
HTTP protocol refactoring

Kriechi:
This PR starts of a larger refactoring process to unify everything related to HTTP semantics and split the actual "on-the-wire" protocol from the overall flow behind it.

Mostly moving code around between projects, modules, and classes.

This PR requires mitmproxy/netlib#81 to work properly.
This PR requires mitmproxy/pathod#31 to work properly.

### DIFF
--- a/libmproxy/console/grideditor.py
+++ b/libmproxy/console/grideditor.py
@@ -5,9 +5,11 @@ import re
 import os
 import urwid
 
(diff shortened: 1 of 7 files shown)


Kriechi:
I updated the PR with the new file & module structure from netlib.

PR #676, opened 2015-07-13
Add source address filter

isra17:
This filter allow to match on the request source and destination addresses in the form
of `<IP>:<Port>` with `~src` and `~dst`.
Ex:

  * Match client IP: `~src 192.168.0.123`
  * Match source subnet: `~src 192.168.0.*`
  * Match port: `~dst :8080`

Also fixed the parsing grammar to add a `WordEnd` after each filter
name. That way, `~src` doesn't match `~s` instead and keep the behavior
consistent with `~hq` != `~h`.

An alternative, which look betters IMO, but isn't consistent with the others filter name would be using name such as:

 * `~src.ip`
 * `~src.port`

This look similar to Wireshark notation and could be extended to other filter such as `~request.headers.host google.com` which might look better and be easier to remember than `~hq 'host.*google.com'`.

### DIFF
--- a/libmproxy/filt.py
+++ b/libmproxy/filt.py
@@ -241,6 +241,19 @@ class FUrl(_Rex):
     def __call__(self, f):
         return re.search(self.expr, f.request.url)
 
+class FSrc(_Rex):
+    code = "src"
+    help = "Match source address"
+
+    def __call__(self, f):
+        return f.client_conn and re.search(self.expr, repr(f.client_conn.address))
+
+class FDst(_Rex):
+    code = "dst"
+    help = "Match destination address"
+
+    def __call__(self, f):
+        return f.server_conn and re.search(self.expr, repr(f.server_conn.address))
 
 class _Int(_Action):
     def __init__(self, num):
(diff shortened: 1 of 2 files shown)


libmproxy/filt.py
```diff
@@ -241,6 +241,23 @@ def make(klass, s, loc, toks):
...
         return re.search(self.expr, f.request.url)
 
+class FSrc(_Rex):
+    code = "src"
+    help = "Match source address"
+
+    def __call__(self, f):
+        return f.client_conn and re.search(self.expr, repr(f.client_conn.address))
+
+class _Int(_Action):
```
mhils: dupe?

mhils:
Thanks for the PR and the suggestions! @cortesi has some (vague) plans to improve the filter stuff in general, this PR looks like a very good addition in the meantime.

Could you please remove the duplicate `_Int`? Looks good to merge otherwise.

isra17:
Great to hear about the filter enhencement!
I removed the `_Int` duplicate :+1:

PR #694, opened 2015-07-25
[WIP] Proxy Refactor

mhils:
A first attempt at separating concerns in the proxy core to fix a variety of issues and provide clean abstractions for other protocols. This PR is intended for comments and discussion.

Requirements to run:
  - cryptography from master
  - `ssl_peek` branch of pyOpenSSL: [ssl_peek@mhils/pyopenssl](https://github.com/mhils/pyopenssl/tree/ssl_peek)
  - `ssl_peek` branch of netlib [ssl_peek@mitmproxy/netlib](https://github.com/mitmproxy/netlib/tree/ssl_peek)

Open points:
  - Using `yield` to communicate between layers is great as it forces a single way for communication. There are a few disadvantages though:
    - `yield`ing from pyOpenSSL callbacks is a major PITA, see `SslLayer._establish_ssl_with_client_and_server` - all that threading stuff is just there to yield from the main thread.
    - Getting responses from lower layers (do we need that?) involves workarounds again
    - Alternatively, callbacks would fix the issues mentioned above, but are (subjectively) not so nice. Thoughts?
  - Protocol Choice Strategy:
    - I hacked together a quick `AutoLayer`, which tries to intelligently guess the next layer. 
      -  Implement as a Mixin rather than a layer?
      - The user must be able to specify a strategy at some point, at least in inline scripts. What's the best interface for this?


libmproxy/protocol2/auto.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import (absolute_import, print_function, division, unicode_literals)
+from .layer import Layer
+
+
+class AutoLayer(Layer):
+    def __call__(self):
+        d = self.client_conn.rfile.peek(1)
+        if d[0] == "\x16":
```
Kriechi: I guess this is a SSL/TLS handshake magic value defined in the spec?
I don't know the actual SSL/TLS handshake on the wire - so this really is magic for me.
Maybe a short comment about it would help...

libmproxy/protocol2/socks.py
```diff
@@ -0,0 +1,26 @@
...
+        except ProxyError as e:
+            # TODO: Unmonkeypatch
+            raise ProxyError2(str(e), e)
+
+        self._set_address(address)
+
+        if address[1] == 443:
+            layer = AutoLayer(self)
+        else:
+            layer = AutoLayer(self)
```
Kriechi: This behaves the same way - I guess this is still WIP?

libmproxy/protocol2/socks.py
```diff
@@ -0,0 +1,26 @@
...
+        for message in layer():
+            if not self._handle_server_message(message):
+                yield message
+
+from .auto import AutoLayer
```
Kriechi: imports should go to the top of the file.

libmproxy/protocol2/auto.py
mhils: Fair point. This was really a quick one-off earlier today, there's much more we can do for the heuristics, which we then should document.

Regarding our actual question:
`0x16` signifies a ClientHello, http://www.moserware.com/2009/06/first-few-milliseconds-of-https.html#client-hello

libmproxy/protocol2/socks.py
mhils: yep. this is a relict of having an explicit ssl/no-ssl switch here.

Kriechi:
This looks really promising! Nice architecture!
This will cleanly separate the HTTP on-the-wire stuff (writing ASCII to the socket vs. writing h2-frames)  from HTTP semantics (read requests, write responses).

Just to mention it again for documentation purposes:
ALPN needs to be addressed in a similar fashion as SNI in the SSL layer.
Depending on which callback fires first we need to workaround a dependency problem.
See: http://openssl.6102.n7.nabble.com/SNI-ALPN-ordering-td58754.html

Additionally I want to propose the replacement of **SSL** with **TLS**.
SSL is dead. Long lives TLS.

(or maybe something more neutral, such as *encryption layer*?)

mhils:
Alternative implementation using callbacks instead of generators: https://github.com/mitmproxy/mitmproxy/compare/proxy-refactor-cb
@resam, @cortesi, what do you prefer?

Kriechi:
@mhils I think I'm a bit more in favour of the layered approach. The callback stuff probably looks like spaghetti code very soon...

Just to get this right:
If I want to send control frames in HTTP/2, such as window control, settings, etc.,
I would add a new message and yield to process it? and once this is done the workflow resumes by waiting for a new request/response?

Kriechi:
This branch is abandoned. Future work will be done in `proxy-refactor-cb`.

Issue #669, opened 2015-07-07
Unable to enter "Replacement mode" in 0.12.1


jjarava:
##### Steps to reproduce the problem:
1. Start mitmproxy
2. Press "R" (Uppercase R) to enter "Replacement" mode to enter expressions
3. Nothing happens

##### What is the expected behavior?
The screen to enter "Replacement" expressions should appear to enable entering the filters and regexes to replace as per the doc in http://mitmproxy.org/doc/features/replacements.html

It's noted that the "on-screen" help doesn't show "R" as a shortcut.

##### What went wrong?
Not sure - no editor appears so there's no option to enter replacement

##### Any other comments?
This is a key feature of mitmproxy for me. 

---
mitmproxy version: `mitmproxy 0.12.1`
Operating System:
```
ProductName:	Mac OS X
ProductVersion:	10.9.5
BuildVersion:	13F34
```

mhils:
The feature moved into the options screen - you can press `o` `R` to open the editor. If you feel strong enough 

Cheers,
Max

jjarava:
Thanks for the information - shouldn't the docs be updated to reflect this?

mhils:
Sure - thanks for pointing that out. This basically applies to all features in the option screen, we definitely need to review the docs. If anyone wants to give it a go, help is highly appreciated.

PR #699, opened 2015-07-29
Update docs to reflect current shortcut keys for features which have …

onlywade:
Update docs to reflect current shortcut keys for features which have been moved into the options menu.

This covers the discrepancy noted in Issue #669.


mhils:
Thanks a lot, really appreciated. You're taking the worst kind of work from our shoulders. :smiley:

Issue #521, opened 2015-03-17
mitmdump output format


raymontag:
Hey,

it would be great if one could get (optional) an output format similar to the one of mitmproxy for POST data and response data. The hex dump is nice but I would like to see something which is more convenient for searching in the data.

Kind regards

albertzaharovits:
Hi Max,

I would like to start working on this! At the moment I see two ways to go: 
1) write some classes similar to the View ones in libmproxy/console/contentiew.py
2) use the current urwid.Widget generation, and 'pretty print' the root urwid.ListBox widget
What do you think?

Albert

mhils:
Hi Albert,

Looking at option 2, I'm afraid we can't really do that - urwid is Linux/OSX only, so we cannot use the console parts in mitmdump (which also runs on Windows). 
Option 1 sounds better here. It would be very cool if we could re-use the common bits for mitmproxy/mitmdump/mitmweb. For example, it would be useful to have common views which return simple data structures (for the time being, either `dict` or `list`), which are then processed by mitmdump for (raw) text output or by mitmproxy for limited/formatted output.

@cortesi, thoughts?

Cheers,
Max

albertzaharovits:
Hi Max,

After a first attempt to rewrite (contentview)[https://github.com/mitmproxy/mitmproxy/blob/master/libmproxy/console/contentview.py], I realised it would imply quite a lot of duplicate code, trying to avoid urwid formating (or worse bloating code with if then/else).
Wouldn't it be more convenient to have an inline script for dump pretty printing?

raymontag:
@albertzaharovits @mhils I received the permission of my employer to work on this issue, too. Is there any progress on this thing or some new thoughts? Are there other things that must happen before this one?

cortesi:
I think content formatting would be a great addition to mitmdump, and I'd definitely pull patches that work towards this. If this happens, we'll have content views in 3 places - mitmproxy, mitmweb and mitmdump - so it's time to think about what we can share and how. 

For mitmproxy and mitmdump, I think it's clear that we can share a great deal. The mitmproxy content views basically produce line-by-line output, with a sprinkling of highlighting. I propose that:

- We move the contentview formatters into the top level of the project
- We refactor the current contentview module to remove reliance on urwid. Instead, contentviews yield lists of lists of (style, content) tuples, where each list of tuples represents a line. So, for hex view, it would look like this:

```
[
    [
        ("offset", "0000000000"), 
[...]

PR #700, opened 2015-07-30
Web plugins

zbuc:
This pull requests encompasses support for plugins in the web interface.

Remaining tasks:

* pulling this upward into `FlowMaster` so it can be used in `mitmdump` and `mitmproxy`
* unit tests (in progress)

There are multiple new features:

* **View Plugins** - these allow you to change the visual presentation of the contents of flow response/request without modifying them on-the-wire
* **Plugin Options** - these persist on the plugin across every flow and can be used for configuring its behavior
* **Plugin Output** - this allows you to view stateful or debug information about the plugin
* **Plugin Actions** - these are available on a one-off basis to run against a flow & type (request/response)
* **Plugin Actions Every Flow** - all plugin actions are also available to apply to every flow automatically

Plugins are registered by passing them with the `-s` argument. They can use `context.plugins.register_action` and `context.plugins.register_view` to add new actions, options, and views to the front-end UI.

screen shot 2015-07-30 at 1 34 40 am
screen shot 2015-07-30 at 1 34 30 am
screen shot 2015-07-30 at 1 33 45 am

Example plugin:

```
import string
import cStringIO
import inspect
import os

from libmproxy.protocol.http import decoded

from bs4 import BeautifulSoup


def piglify(src):
    words = string.split(src)
    ret = ''
    for word in words:
        idx = -1;
        while word[idx] in string.punctuation and (idx * -1) != len(word): idx -= 1
        if word[0].lower() in 'aeiou':
            if idx == -1: ret += word[0:] + "hay"
            else: ret += word[0:len(word)+idx+1] + "hay" + word[idx+1:]
        else:
            if idx == -1: ret += word[1:] + word[0] + "ay"
            else: ret += word[1:len(word)+idx+1] + word[0] + "ay" + word[idx+1:]
        ret += ' '
    return ret.strip()


def piglify_response(context, flow):
    with decoded(flow.response):
        html_response = False
        for header in flow.response.headers:
            if header[0] == 'Content-Type' and 'text/html' in header[1]:
                html_response = True

        if not html_response:
            return

        soup = BeautifulSoup(flow.response.content, 'html.parser')
        replace = []
        for text in soup._all_strings():
            replace.append(text)

        for text in replace:
            if soup.find(text=text):
                soup.find(text=text).replaceWith(piglify(text))

        flow.response.content = str(soup.prettify().encode('utf-8'))


def disable_cache(context, flow):
    with decoded(flow.request):
        flow.request.headers['Cache-Control'] = ['no-cache']
        flow.request.headers['If-Modified-Since'] = []


def puglify_image(context, flow):
    with decoded(flow.response):
        try:
            if flow.response.headers.get_first("content-type", "").startswith("image") or flow.response.content.startswith("\xFF\xD8\xFF\xE0\x00\x10\x4A\x46\x49\x46\x00"):
                img = cStringIO.StringIO(open(os.path.normcase(os.path.dirname(inspect.stack()[0][1]) + '/pug.jpg'), 'rb').read())
                flow.response.content = img.getvalue()
                flow.response.headers["content-type"] = ["image/jpeg"]
        except Exception, e:
            print "Error puglifying:"
            print repr(e)


def colorify(context, flow):
    with decoded(flow.response):
        try:
            color = context.plugins.get_option_value('colorswitcher', 'color')
            css_response = False
            for header in flow.response.headers:
                if header[0].lower() == 'content-type' and ('css' in header[1].lower() or 'html' in header[1].lower()):
                    css_response = True
            for header in flow.request.headers:
                if header[0].lower() == 'accept' and ('css' in header[1].lower() or 'html' in header[1].lower()):
                    css_response = True

            if not css_response:
                return

            import re
            pattern_hex = re.compile('#[0-9]+', re.IGNORECASE)
            flow.response.content = pattern_hex.sub(color, flow.response.content)
        except Exception, e:
            print "Error colorifying:"
            print repr(e)


def start(context, argv):
    def hexdump(flow, **kwargs):
        target = kwargs.get('target')
        src = ''
        if target == 'request':
            src = flow.request.content
        elif target == 'response':
            src = flow.response.content
        length = kwargs.get('length') or 8
        result = []
        digits = 4 if isinstance(src, unicode) else 2
        for i in xrange(0, len(src), length):
            s = src[i:i+length]
            hexa = b' '.join(["%0*X" % (digits, ord(x)) for x in s])
            text = b''.join([x if 0x20 <= ord(x) < 0x7F else b'.' for x in s])
            result.append(b"%04X   %-*s   %s" %
                         (i, length * (digits + 1), hexa, text))
        return b'\n'.join(result)

    # plugins are currently only supported on the WebMaster class
    if hasattr(context, 'plugins') and context.plugins:
        context.plugins.register_view('hex',
                                      title='Hex View Plugin',
                                      transformer=hexdump)

        context.plugins.register_action('colorswitcher',
                                        title='Color Switcher',
                                        options=[{
                                            'title': 'Color',
                                            'id': 'color',
                                            'state': {
                                                'value': '#EA5C7A',
                                            },
                                            'type': 'text',
                                        }],
                                        actions=[{
                                            'title': 'Colorify',
                                            'id': 'colorify',
                                            'state': {
                                                'every_flow': True,
                                            },
                                            'possible_hooks': ['response'],
                                        },
                                        {
                                            'title': 'Disable Cache',
                                            'id': 'disable_cache',
                                            'state': {
                                                'every_flow': True,
                                            },
                                            'possible_hooks': ['request'],
                                        }
                                        ])

        context.plugins.register_action('piglify',
                                        title='P[iu]g Latin Plugin',
                                        actions=[
                                                 {
                                                  'title': 'Piglify Response',
                                                  'id': 'piglify_response',
                                                  'state': {
                                                    'every_flow': True,
                                                  },
                                                  'possible_hooks': ['response'],
                                                 },
                                                 { 
                                                  'title': 'Puglify Image',
                                                  'id': 'puglify_image',
                                                  'possible_hooks': ['response'],
                                                  'state': {
                                                    'every_flow': True,
                                                    },

                                                 },
                                                 ])
```


mhils:
Thanks for the great work, @zbuc. This PR brings a few very cool additions, although we should have started with smaller PRs waaaay earlier - this is really though to review. Anway, I'm not complaining. :wink: 

The Web plugins stuff is pretty much independent of the rest and I think we can more or less merge it as-is. :+1: 

I'm not so satisfied with the content view/view plugins yet. Custom content views is something that should be available to mitmproxy and mitmdump as well (#521 deals with the same issue). In particular, reusing the existing content views in `libmproxy.console.contentview` right from the start would be a huge addition. Can we get this rolling together? I'm happy to help out with both the console and the mitmdump part, maybe let's just coordinate this on Slack over the next few days?

mhils:
We'll continue with that in smaller PRs! :smiley:

PR #729, opened 2015-08-18
[WIP] Add an example to beautify JavaScript

ryoqun:
First off, thanks for actively maintaining this great project!

As the title says all, this adds a very simple script to beautify Javascript.

well, it seems there is no other simple solution on the net. So I created a proof of concept prototype.

Though, actually I don't want to improve this script further in the areas of content-type handling (and content sniffing) and encoding support on par with browsers..

Do you know other better approach??

### DIFF
--- a/examples/beautify_javascript.py
+++ b/examples/beautify_javascript.py
@@ -1,0 +1,34 @@
+from libmproxy.protocol.http import decoded
+from libmproxy import version
+import jsbeautifier
+import sys
+import bs4
+import platform
+
+def start(context, argv):
+    if len(argv) != 1:
+        raise ValueError('Usage: -s "beautify_javascript"')
+
+def response(context, flow):
+    try:
+        header = '// beautified by jsbeautifier ' + jsbeautifier.__version__ + ' via mitmproxy ' + version.VERSION
+        server_header = '// running on Python ' + sys.version.replace('\n', '') + ' ' + (' '.join(platform.uname()))
+
+        content_type = flow.response.headers.get_first('Content-Type', '').lower()
+        # this sucks; should use more better logic....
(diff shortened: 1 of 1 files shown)


ryoqun:
As a note: I'm not really good at writing Python.

examples/beautify_javascript.py
```diff
@@ -0,0 +1,34 @@
...
+import platform
+
+def start(context, argv):
+    if len(argv) != 1:
+        raise ValueError('Usage: -s "beautify_javascript"')
+
+def response(context, flow):
+    try:
+        header = '// beautified by jsbeautifier ' + jsbeautifier.__version__ + ' via mitmproxy ' + version.VERSION
+        server_header = '// running on Python ' + sys.version.replace('\n', '') + ' ' + (' '.join(platform.uname()))
```
ryoqun: This would be needn't; But I added it for diagnostic purpose (There WAS multiple pythons on my mac....) and for fun (meh, this kind of comments ticks my nerd mentality, doesn't it?) :p

ryoqun:
Another note: I developed this using mitmproxy 0.12.1

mhils:
Thanks for the PR (and sorry for the slow response - this one slipped through)! May I ask what's the exact use case for this? We already have jsbeautifier included for the content view in mitmproxy, which should display things nicely in the UI.

ryoqun:
@mhils well, the use case for this is to debug scripts of third-party site (i.e, can't have an access to the origina code) As far as I know, most browsers doesn't nicely handle minified scripts when debugging especially for backtraces and breakpoints...

mhils:
You can now use mitmproxy's content views from inline scripts, which also support pretty-printing. This is a great example, but it doesn't demonstrate any mitmproxy-unique features that aren't already present in the modify_response_body example. I think we want to focus on smaller "official" examples that demonstrate a single use-case for now. Thank you very much anyway, happy to see people building great stuff on top of that!

Issue #710, opened 2015-08-04
UnicodeDecodeError when dumping with read_dumpfile


FreeArtMan:
Used example script  flowwriter.py to collect some data
Then tryed to run it with example script read_dumpfile 

getting :
```
...
        "code": 200, 
        "content": Traceback (most recent call last):
  File "read_dumpfile", line 16, in <module>
    json.dump(f.get_state(), sys.stdout, indent=4)
  File "/usr/lib/python2.7/json/__init__.py", line 189, in dump
    for chunk in iterable:
  File "/usr/lib/python2.7/json/encoder.py", line 434, in _iterencode
    for chunk in _iterencode_dict(o, _current_indent_level):
  File "/usr/lib/python2.7/json/encoder.py", line 408, in _iterencode_dict
    for chunk in chunks:
  File "/usr/lib/python2.7/json/encoder.py", line 390, in _iterencode_dict
    yield _encoder(value)
UnicodeDecodeError: 'utf8' codec can't decode byte 0x8b in position 1: invalid start byte
...
```

Example file that was collected with flowwrite.py is attached. I filltered out one http response that causes this bug

 
![logfile](https://cloud.githubusercontent.com/assets/214155/9064656/7a4207f0-3ac5-11e5-8599-5536580ae3fc.png)

>md5sum logfile
fc0285d300cccc88b757e4ff70bf128b  logfile

Tested with mitmproxy commit adf2877020c563fa9dc9a4df26bfd8b243e8e41a

cortesi:
Hi there,

Could you please test with a recent checkout of mitmproxy? Reading the dumpfile you attached works with no error in the latest master.

If the issue is still there, please re-open and specify your platform, mitmproxy version and any other details that could help us reproduce.


Aldo

FreeArtMan:
Hi, ofcourse it works without "no error". There is just try/catch used. Problem was I think becouse 
if you have some binary file like image then it contains non printable characters. While json.dump couldnt
dump it it stil valid file/packet. This error happened with files that have in headers Contet-Type specified compression format and in content compressed content. Probably now some one will expirience non-dump valid packets.

cortesi:
Hi Arthur. Further to our email just now, base64 encoding to make binary data JSON-safe is the way to go. Let's get a working test case up and running here, so we can see exactly what we're dealing with. Using the test dump you provided, and the example script in the repo, I don't see the error. The output is as expected, and an error is not hidden by a try/catch block.

[...]

PR #731, opened 2015-08-18
Fixing issue #710. Prety printing in read_dumpfile

FreeArtMan:


mhils:
Thanks a lot. Can we just use `pprint.pprint(data, indent=4)` please (or create just one pprint instance, not a new one for every flow)?

PR #733, opened 2015-08-19
Proxy refactor cb

Kriechi:


libmproxy/proxy/connection.py
```diff
@@ -98,6 +98,7 @@ def __init__(self, address):
         self.timestamp_tcp_setup = None
         self.timestamp_ssl_setup = None
         self.protocol = None
+        self.finished = False
```
mhils: This should be in TCPClient?

mhils:
Thanks for porting. LGTM otherwise. :+1:

PR #739, opened 2015-08-26
fix return value and empty requests

Kriechi:


libmproxy/protocol2/http.py
```diff
@@ -237,6 +238,9 @@ def __call__(self):
 
                 self.log("request", "debug", [repr(request)])
 
+                if not request:
+                    return
+
```
mhils: Why is this part neccessary?

Kriechi: just a safe guard...

mhils: Ok. Let's keep that out -`HTTPRequest.from_protocol` should raise a NetLibError if it is unable to return a request, which we catch above.

Kriechi: fixed.

PR #742, opened 2015-08-27
split layers from proxy-modes

Kriechi:


mhils:
Thanks - implemented in a86ec56012136664688fa4a8efcd866b5e3e17a8.

PR #750, opened 2015-09-03
Update modify_response_body.py

internety:
should be: 

from libmproxy.protocol.http import decoded


mhils:
That was quick :stuck_out_tongue:  - I changed this an hour ago (a18db90ea3123b37e3f1fd3f10da001a8ea893c2) because we moved the decorator into `models`. It may move it again because it doesn't really fit there semantically, but the code itself is correct for the latest mitmproxy master. Thanks for bothering to send a PR though!

Cheers,
Max

PR #752, opened 2015-09-06
HTTP/2: implement stream multiplexing

Kriechi:
requires https://github.com/mitmproxy/netlib/pull/91

### DIFF
--- a/libmproxy/models/connections.py
+++ b/libmproxy/models/connections.py
@@ -2,6 +2,9 @@ from __future__ import (absolute_import, print_function, division)
 
 import copy
 import os
+import threading
+import time
+import StringIO
 
 from netlib import tcp, certutils
 from .. import stateobject, utils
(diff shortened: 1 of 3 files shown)


libmproxy/models/connections.py
```diff
@@ -26,6 +27,7 @@ def __init__(self, client_connection, address, server):
         self.timestamp_end = None
         self.timestamp_ssl_setup = None
         self.protocol = None
+        self.socket_lock = threading.Lock()
```
mhils: As you said, we probably don't want to touch ServerConnection at all. 
I think we can just use a wrapper for the HTTP2 protocol which does the locking?
```python
class ThreadsafeServerConnection(object):
    def __init__(self, socket_lock, ctx):
        self.socket_lock = socket_lock
        self.ctx = ctx

    def __getattr__(self, item):
        return getattr(self.ctx.server_conn, item)
    
    def send(self, message):
        with self.socket_lock:
            self.ctx.server_conn.send(msg)
```

libmproxy/protocol/http.py
```diff
@@ -249,6 +223,78 @@ def handle_unexpected_frame_from_server(self, frame):
...
+        # Single use means discarding it afterwards.
+        return True
+
+    def __call__(self):
+        layer = HttpLayer(self, self.mode)
+        layer()
+
+    def connect(self):
+        # do not mess with the server connection - all streams share it.
+        pass
```
mhils: should raise `NotImplementedError` (same for set_server)

PR #788, opened 2015-09-30
Allow reading scripts from an anonymous pipe

sethp-jive:
Bash (and many other shells) provide a nifty feature in "anonymous pipe" or "anonymous fifo" whereby the output of a subshell may be treated as a simple file by the parent shell: http://unix.stackexchange.com/a/156088

Unfortunately, libmproxy complains because that "file" is not a regular file, as os.path.isfile checks, e.g. giving the error "Not a file: /dev/fd/11". This patch is intended to provide for the following use-case:

```
mitmdump -s <(echo "def response(context, flow):\n  flow.response.headers['newheader'] = [`hostname`]")
```

where `hostname` may be replaced with a more complicated lookup.

### DIFF
--- a/libmproxy/script.py
+++ b/libmproxy/script.py
@@ -83,7 +83,7 @@ class Script:
                  "If your script path contains spaces, "
                  "make sure to wrap it in additional quotes, e.g. -s \"'./foo bar/baz.py' --args\".") %
                 args[0])
-        elif not os.path.isfile(args[0]):
+        elif os.path.isdir(args[0]):
             raise ScriptError("Not a file: %s" % args[0])
         return args
 


mhils:
Great stuff - thanks for the PR! :relaxed:

sethp-jive:
Thanks for the fast turnaround!

vapor99:
Ok, now that is just friggin sweet!

enesunal:
Only looking os.path.isdir() is not the appropriate way doing it. Entry should be verified with stat(). I'm going to correct it when I have some time.

sethp-jive:
@enesunal Do I read you correctly in that you'd prefer the condition specifically whitelisted FIFOs as per:

```
import stat
...
elif not os.path.isfile(args[0]) and not stat.S_ISFIFO(os.stat(args[0]).st_mode):
    raise ScriptError("Not a file: %s" % args[0])
```

My feeling was that of the (file types)[https://en.wikipedia.org/wiki/Unix_file_types] I knew about, the only one that couldn't possibly work as an `execfile` argument was a directory – named or anonymous pipes, sockets, or device files (most obviously `/dev/null`, or maybe in extreme cases a block device) might provide valid python code that a user may want to feed in as a script & this was a cheap way to support those admittedly weird edge cases. In the event that the user fed in the wrong input, the `execfile` itself would provide the user an error message.

Anyway, let me know if you'd prefer to specifically allow just regular files & FIFOs and I'll send a PR that changes the code to do that.

mhils:
I think the current way is perfectly fine. Alternatively, we can do the 
```python
try:
    with open("abc", "rb"):
        pass
except:
    # print error
```
dance, but I think the current solution is more than good enough.

PR #791, opened 2015-10-04
Prevent flowview from creating duplicated windows

tunz:
When accepting intercepted flows with 'a' or 'A' command in flowview, mitmproxy makes new window every time. So, I need to press 'q' button several times to return the flow list.


tunz:
I cannot understand why it fails in CI. The test failure is not reproduced in my ubuntu 14.04 or windows 10 environment.

mhils:
Thanks for the PR! :cake: 
Somehow forgot to merge this - sorry! :smile:

PR #793, opened 2015-10-06
added another example for local content

enesunal:
binary file response.


mhils:
Thanks for the PR. This seems like it fits into `modify_response_body.py` rather than the redirect example. Also, we should probably read the file on startup and then use the cached copy from memory. Once that's in, I'd be happy to merge this.

PR #799, opened 2015-10-11
export as curl command 

gato:
Implements #596 based on #619 PR with using "E" as key see @cortesi comment on https://github.com/mitmproxy/mitmproxy/pull/619#issuecomment-110547758 
also fixes issues when pyperclip is not installed. (https://github.com/mitmproxy/mitmproxy/pull/619#issuecomment-114899022)
also fixes copy to clipboard when "h" key (header + content) was pressed, it was broken on master.


mhils:
Fixed by #884! :smiley:

PR #805, opened 2015-10-23
Update ubuntu installation to includ libjpeg

Badg:
Otherwise install fails with error "--enable-jpeg requested but jpeg not found"


mhils:
Thanks, great catch. This seems to be caused by https://github.com/python-pillow/Pillow/issues/1457.

https://pillow.readthedocs.org/en/3.0.0/installation.html#linux-installation recommends `libjpeg8-dev zlib1g-dev`. Can you adjust the PR to include these please?

Thanks!
Max

Badg:
Should be updated accordingly. Cheers!

mhils:
Thanks! :cake:

PR #809, opened 2015-10-29
Better error handling/reporting for pfctl failures on OS X

zbuc:
On my system, I was receiving a `CalledProcessError` when the pfctl subprocess executed, resulting in a vague "Transparent mode failure: CalledProcessError()" message.

This should make it more likely to report the correct error.

### DIFF
--- a/libmproxy/platform/osx.py
+++ b/libmproxy/platform/osx.py
@@ -19,8 +19,15 @@ class Resolver(object):
 
     def original_addr(self, csock):
         peer = csock.getpeername()
-        stxt = subprocess.check_output(self.STATECMD, stderr=subprocess.STDOUT)
-        if "sudo: a password is required" in stxt:
+        try:
+            stxt = subprocess.check_output(self.STATECMD, stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError, e:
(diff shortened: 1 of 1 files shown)


libmproxy/platform/osx.py
```diff
@@ -19,8 +19,16 @@ class Resolver(object):
...
-        stxt = subprocess.check_output(self.STATECMD, stderr=subprocess.STDOUT)
-        if "sudo: a password is required" in stxt:
+        insufficient_priv = False
+        try:
+            stxt = subprocess.check_output(self.STATECMD, stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError, e:
+            if "sudo: a password is required" in e.output:
+                insufficient_priv = True
+            else:
+                raise RuntimeError("Error getting pfctl state: " + repr(e))
```
mhils: Can we please make an 
```python
except:
    ...
else:
    insufficient_priv = "sudo: a password is required" in stxt
```

block here and then just check for `insufficient_priv`?

mhils:
Good catch, thanks :relaxed:. Can we make the code a bit more Pythonic (see comment)? Looks great otherwise.

zbuc:
Done!

PR #810, opened 2015-10-30
Use ctrl+f and ctrl+b to scroll a page

gecko655:
I want to use ctrl+f/b to scroll down/up one page, just like `less` and `vi(m)`.

Since b (save request/response body) and f (load full body data) are used as other commands, I only implemented ctrl+f/b, unlike `less`.


mhils:
Thanks! :sparkles:

PR #819, opened 2015-11-05
fix copy to clipboard command when copying content

gato:
Hi @mhils 
 
This PR has only the changes to fix the exception thrown when trying to copy to clipboard option: headers+content (h key)

Cheers
Marcelo


mhils:
:+1: :cake:

Issue #713, opened 2015-08-06
Hotkey to reload all inline scripts


ojab:
When writing/testing inline scripts there is AFAIU two options to load fresh script version from filesystem:

1. Restart mitmproxy.
2. Remove and add again script via "Editing scripts menu".

It will be handy to have scripts reloaded with just one keystroke.

mhils:
Good suggestions. We'd be happy to accept a PR for this! :smiley:

PR #820, opened 2015-11-06
Add hotkey to reload all inline scripts #713

MatthewShao:
Simply add hotkey 's' on the flowview window, I'm not sure whether 's' key is the best choice for this function, but it works fine.

### DIFF
--- a/libmproxy/console/__init__.py
+++ b/libmproxy/console/__init__.py
@@ -225,7 +225,6 @@ class ConsoleMaster(flow.FlowMaster):
                 if err:
                     print >> sys.stderr, "Script load error:", err
                     sys.exit(1)
-                script.ObserveScripts(self, i)
 
         if options.outfile:
             err = self.start_stream_to_path(
--- a/libmproxy/console/__init__.py
+++ b/libmproxy/console/__init__.py
@@ -225,6 +225,7 @@ class ConsoleMaster(flow.FlowMaster):
                 if err:
                     print >> sys.stderr, "Script load error:", err
                     sys.exit(1)
+                script.ObserveScripts(self, i)
 
         if options.outfile:
             err = self.start_stream_to_path(
--- a/libmproxy/console/__init__.py
+++ b/libmproxy/console/__init__.py
@@ -731,3 +731,7 @@ class ConsoleMaster(flow.FlowMaster):
         if f:
             self.process_flow(f)
         return f
+
+    def script_change(self, script):
+        self.masterq.put(("script_change", script))
+        signals.status_message.send(message="<{}> reloaded.".format(script.args[0]))
--- a/libmproxy/flow.py
+++ b/libmproxy/flow.py
@@ -667,6 +667,10 @@ class FlowMaster(controller.Master):
             self.add_event("Script error:\n" + str(e), "error")
         self.scripts.remove(script_obj)
 
+    def reload_scripts(self):
+        for s in self.scripts[:]:
+            s.load()
+    
     def load_script(self, command):
         """
             Loads a script. Returns an error description if something went
--- a/libmproxy/flow.py
+++ b/libmproxy/flow.py
@@ -642,6 +642,7 @@ class FlowMaster(controller.Master):
 
         self.stream = None
         self.apps = AppRegistry()
+        script.sig_script_change.connect(self.script_change)
 
     def start_app(self, host, port):
         self.apps.add(
@@ -666,10 +667,6 @@ class FlowMaster(controller.Master):
         except script.ScriptError as e:
             self.add_event("Script error:\n" + str(e), "error")
         self.scripts.remove(script_obj)
-
-    def reload_scripts(self):
-        for s in self.scripts[:]:
-            s.load()
     
     def load_script(self, command):
         """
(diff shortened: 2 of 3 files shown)


mhils:
Great - thanks a lot! :cake: 
It'd be great if we could also reload scripts in the flow list (not just the flow view). Then again, it's kind of odd that the script list is hidden behind `o` ``S`, whereas reloading scripts is possible by simply pressing `s`.

I'm happy to merge this if you could add it to the flow list as well. If you have a spare afternoon, it would be super super awesome if we could use https://pypi.python.org/pypi/watchdog and reload scripts automatically. :smile:

MatthewShao:
Thanks for the suggestions, it's really awesome! :smiley: 
So I add a *watchdog.observers.Observer* to catch the modify events on the script files.It works fine on my computer, but checks failed. Don't know why this happened, could you explain?

libmproxy/console/__init__.py
```diff
@@ -225,6 +225,7 @@ def __init__(self, server, options):
                 if err:
                     print >> sys.stderr, "Script load error:", err
                     sys.exit(1)
+                script.ObserveScripts(self, i)
```
mhils: We should initialize the observation in the constructor of the Script class - otherwise, we don't have this feature for mitmdump.

libmproxy/script.py
```diff
@@ -4,6 +4,9 @@
 import threading
 import shlex
 import sys
+from watchdog.observers import Observer
+from watchdog.events import PatternMatchingEventHandler, FileModifiedEvent
+from .console import signals
```
mhils: Unfortunately, we cannot use the "signals" from the `console` submodule here, which is specific to mitmproxy. Instead, we should add a log event to the master's queue (the "Channel" part in http://docs.mitmproxy.org/en/stable/dev/architecture.html) as it is done [here](https://github.com/mitmproxy/mitmproxy/blob/3739e1fe82c83ada3289dea5e24f9efcc7175696/libmproxy/proxy/root_context.py#L115). This is a bit tricky, I'm happy to take this over if you like.

mhils:
Thanks for the adapting this! The issues you are seeing are caused by a circular import of modules, as you import parts of `console` in `script`. I added two comments above, which I'd like to see fixed before we merge this. Also, it'd be great if you could adjust your code to be somewhat [PEP8](https://www.python.org/dev/peps/pep-0008/) compliant - function names and variables should start with a lowercase letter.

Thanks!
Max

MatthewShao:
I really appreciate such a detailed explaination! :cake: 
Now,this function works in both ```mitmproxy``` and ```mitmdump```, you can see the message about the script has been reloaded in the event log after modifying a script. However, I want to push this one step forward —— display the message at the bottom of the window in mitmproxy. Clearly, ```signals.status_message.send``` should be called, but I don't know how to call this correctly when a modify event is caught. Do you have any idea?

libmproxy/script.py
```diff
@@ -4,6 +4,8 @@
 import threading
 import shlex
 import sys
+from watchdog.observers import Observer
+from watchdog.events import PatternMatchingEventHandler, FileModifiedEvent
```
mhils: 1. Add a blinker event (as in [libmproxy.console.signals](https://github.com/mitmproxy/mitmproxy/blob/master/libmproxy/console/signals.py)) for script_change here.

libmproxy/script.py
```diff
@@ -192,3 +195,23 @@ def _concurrent(ctx, obj):
...
+
+class ScriptModified(PatternMatchingEventHandler):
+    
+    def __init__(self, flow_master):
+        self.flow_master = flow_master
+        PatternMatchingEventHandler.__init__(self, ignore_directories=True, patterns=["*.py"])
+        self.context = ScriptContext(self.flow_master)
+
+    def on_modified(self, event=FileModifiedEvent):
+        self.flow_master.reload_scripts()
```
mhils: 2. Instead of using the flow master here, trigger the blinker event with the script object as argument.

libmproxy/flow.py
```diff
@@ -667,6 +667,10 @@ def unload_script(self, script_obj):
             self.add_event("Script error:\n" + str(e), "error")
         self.scripts.remove(script_obj)
 
+    def reload_scripts(self):
```
mhils: 3) Subscribe to the script change signal in `__init__`. The event handler should call `self.masterq.put(("script_change", script))`.
4) Add a `handle_script_change` function, that once called, takes the script object and calls `script.reload()`.

mhils:
Thanks so much for the changes! I think there are two places where we can improve the code significantly:

1. Decouple `libmproxy.script` from `libmproxy.flow`: The script module should not know the inner workings of the flow module. Instead, let's expose a signal that Flow can subscribe to!
2. In `libmproxy.flow`, we don't process the signal immediately - the "the script has changed" notification comes from a different thread, so this could cause some nasty race conditions with the main mitmproxy thread. Instead, put it on the queue, so that it get's processed in the main FlowMaster thread.

If you have any questions, please feel free to ask here or join us on Slack! :relaxed:

MatthewShao:
Many thanks! I spend an afternoon on the code and finally make myself clear about the structure. I change the code as you said. And now it can display a  message on the status bar of ```mitmproxy``` when the script is changed. Unfortunately, it fails on the ```appveyor``` checks as you see above, but I guess the error is not cause by my code. Please check this and if there is something wrong, please let me know.:relieved:
Really enjoy working with you guys!:punch:

mhils:
I just did some final adjustments and merged everything. This is a great feature, thank you very much! :relaxed:

PR #821, opened 2015-11-06
Fix failing protobuf contentview test

zbuc:
I think the test code should *probably* use odict stuff but this fixes the failing test case

```
================================================================================================================== FAILURES ==================================================================================================================
_________________________________________________________________________________________________________ test_view_protobuf_request _________________________________________________________________________________________________________

    def test_view_protobuf_request():
        v = cv.ViewProtobuf()

        p = tutils.test_data.path("data/protobuf01")
        content_type, output = v(file(p, "rb").read())
        assert content_type == "Protobuf"
>       assert output[0].text == '1: "3bbc333c-e61c-433b-819a-0b9a8cc103b8"'
E       TypeError: 'generator' object has no attribute '__getitem__'

test/test_contentview.py:236: TypeError
```

### DIFF
--- a/test/test_contentview.py
+++ b/test/test_contentview.py
@@ -233,7 +233,7 @@ if cv.ViewProtobuf.is_available():
         p = tutils.test_data.path("data/protobuf01")
         content_type, output = v(file(p, "rb").read())
         assert content_type == "Protobuf"
-        assert output[0].text == '1: "3bbc333c-e61c-433b-819a-0b9a8cc103b8"'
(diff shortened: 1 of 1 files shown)


mhils:
Thanks! What's more concerning is that we don't catch this on travis. :open_mouth: 
I thought https://github.com/mitmproxy/mitmproxy/tree/travis-contentviews should catch it, but apparently the tests run fine - any idea why that is the case?

zbuc:
My only idea is that this check is failing on Travis and not running the test case:

```if cv.ViewProtobuf.is_available():```

PR #832, opened 2015-11-12
Custom contentviews

zbuc:
This adds the ability to register custom contentviews.

I'm not convinced this is the best way. I liked the original implementation I had in https://github.com/mitmproxy/mitmproxy/pull/700 but it wasn't contentview friendly.

The new implementation requires scripts to define subclasses of `libmproxy.contentviews.View` and produce lists of formatted tuples rather than the simple string-based approach used in PR #700.

### DIFF
--- a/examples/custom_contentviews.py
+++ b/examples/custom_contentviews.py
@@ -1,0 +1,60 @@
+import string
+from libmproxy import script, flow, utils
+import libmproxy.contentviews as cv
+from netlib.http import Headers
+import lxml.html
+import lxml.etree
+
+
+def start(context, argv):
+    class ViewPigLatin(cv.View):
+        name = "pig_latin_HTML"
+        prompt = ("pig latin HTML", "l")
+        content_types = ["text/html"]
+
+        def __call__(self, data, **metadata):
+            if utils.isXML(data):
+                parser = lxml.etree.HTMLParser(
+                    strip_cdata=True,
+                    remove_blank_text=True
+                )
+                d = lxml.html.fromstring(data, parser=parser)
+                docinfo = d.getroottree().docinfo
+
+                def piglify(src):
+                    words = string.split(src)
+                    ret = ''
+                    for word in words:
+                        idx = -1
+                        while word[idx] in string.punctuation and (idx * -1) != len(word): idx -= 1
+                        if word[0].lower() in 'aeiou':
+                            if idx == -1: ret += word[0:] + "hay"
+                            else: ret += word[0:len(word)+idx+1] + "hay" + word[idx+1:]
+                        else:
+                            if idx == -1: ret += word[1:] + word[0] + "ay"
+                            else: ret += word[1:len(word)+idx+1] + word[0] + "ay" + word[idx+1:]
+                        ret += ' '
+                    return ret.strip()
+
+                def recurse(root):
+                    if hasattr(root, 'text') and root.text:
+                            root.text = piglify(root.text)
+                    if hasattr(root, 'tail') and root.tail:
+                            root.tail = piglify(root.tail)
+
+                    if len(root):
+                        for child in root:
+                            recurse(child)
+
+                recurse(d)
+
+                s = lxml.etree.tostring(
+                    d,
+                    pretty_print=True,
+                    doctype=docinfo.doctype
+                )
+                return "HTML", cv.format_text(s)
+
+    context.plugins.register_view('pig_latin_HTML',
+                                  title='Pig Latin Plugin',
+                                  class_ref=ViewPigLatin)
--- a/libmproxy/contentviews.py
+++ b/libmproxy/contentviews.py
@@ -515,6 +515,14 @@ def get_by_shortcut(c):
             return i
 
 
+def add(obj):
+    for i in views:
+        if i.name == obj.name:
+            raise ContentViewException("Duplicate view: " + obj.name)
+
+    views.append(obj)
+
+
 def get(name):
     for i in views:
         if i.name == name:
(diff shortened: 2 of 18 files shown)


libmproxy/contentviews.py
```diff
@@ -515,6 +515,24 @@ def get_by_shortcut(c):
             return i
 
 
+def add(obj):
+    for i in views:
+        if i.name == obj.name:
+            raise ContentViewException("Duplicate view: " + obj.name)
+
+    for prompt in view_prompts:
```
mhils: This problem sucks. :neutral_face: 
This should be good enough for now, but at some point we need a "automatically find different prompt key" thingie.
Can we add a TODO here please?

libmproxy/contentviews.py
```diff
@@ -515,6 +515,24 @@ def get_by_shortcut(c):
...
+        if prompt[1] == obj.prompt[1]:
+            raise ContentViewException("Duplicate view shortcut: " + obj.prompt[1])
+
+    views.append(obj)
+
+    for ct in obj.content_types:
+        l = content_types_map.setdefault(ct, [])
+        l.append(obj)
+
+    view_prompts.append(obj.prompt)
```
mhils: This function looks useful. Can we use this function to add the "normal" views as well?

examples/custom_contentviews.py
```diff
@@ -0,0 +1,60 @@
+import string
+from libmproxy import script, flow, utils
+import libmproxy.contentviews as cv
+from netlib.http import Headers
+import lxml.html
+import lxml.etree
+
+
+def start(context, argv):
+    class ViewPigLatin(cv.View):
```
mhils: Can we move this out of `start()`?

libmproxy/contentviews.py
```diff
@@ -515,6 +515,24 @@ def get_by_shortcut(c):
...
+
```
mhils: We should have `remove()` right from the start to unregister content views - #820 may result in frequent reloads of inline scripts. Can we add this? :smiley:

libmproxy/flow.py
```diff
@@ -614,6 +619,43 @@ def killall(self, master):
         self.flows.kill_all(master)
 
 
+class Plugins(object):
```
mhils: This looks a bit java-ish - what are your plans here? I think we can just add `ScriptContext.(add|remove)_contentview()` and don't need a Plugin registry thingie?

FWIW, I pondered with the idea of adding "extension"-functionality as well. My line of thought was using setuptools' `entry_points` so that you could `pip install mitmproxy-pig-contenview` to add sort of a permanent inline script that executes like a normal inline script (but is displayed differently in the UI). This is unrelated to this though - happy to discuss that in a separate place. :smiley:

libmproxy/web/app.py
```diff
@@ -176,6 +177,60 @@ def post(self, flow_id):
             raise APIError(400, r)
 
 
+class ViewPluginFlowContent(RequestHandler):
```
mhils: This should just be called `FlowContentView` or something along those lines, no? If we have something like that, we surely want to expose all existing content views! :relaxed:

libmproxy/web/app.py
```diff
@@ -176,6 +177,60 @@ def post(self, flow_id):
...
+                headers=message.headers
+            )
+
+        def get_text(line):
+            for (style, text) in line:
+                yield text
+
+        content = u"\r\n".join(
+            u"".join(get_text(line)) for line in lines
+        )
```
mhils: I think we want to do this concatentation on the client side so that we can add colors later on. :smiley:

libmproxy/web/app.py
```diff
@@ -239,6 +294,32 @@ def put(self):
...
+                            if isinstance(v, dict):
+                                _check_dict_for_callables(v)
+
+                    _check_dict_for_callables(new_dict)
+                    ret_arr.append(new_dict)
+            return ret_arr
+
+        self.write(dict(
+            data=list(_flatten(self.master.plugins))
+        ))
```
mhils: Following the parts above, let's just expose the list of content views in the Settings API endpoint?
(Exposing application state is surely something that's not ideal yet, but that's a different issue)

examples/custom_contentviews.py
zbuc: sorry, which part? the class definition?

libmproxy/contentviews.py
zbuc: What should `remove()` take as input? View name, or an instantiated view object?

libmproxy/flow.py
zbuc: This was a placeholder for the "action plugins" in my earlier pull request... I had both "view plugins" (now custom content views) and "action plugins" grouped together in `Plugins`. ContentViews don't need to be in that object, but registering actions & their options will need storage somewhere.

examples/custom_contentviews.py
mhils: Yes - just move the class to the top level please. It doesn't have to live in the start hook. :smiley:

libmproxy/contentviews.py
mhils: `add()` accepts an instance, so I think we should do the same for `.remove()`. Ideally, the inline script should look like this:
```python

class ViewPigLatin(View):
    # ...

pig_view = ViewPigLatin()

def start(ctx, argv):
    ctx.add_contentview(pig_view)

def stop(ctx):
    ctx.remove_contentview(pig_view)
```

libmproxy/flow.py
mhils: I think it's really superfluous for the content views - can we leave that out for now and just implement `ScriptContext.add_contentview()` to call `contentviews.add()` ? Let's keep the code in the history so that we can use it if we need to, but it seems like overkill for the contentviews.

PR #833, opened 2015-11-13
Contentview scripts

zbuc:
Based on feedback from PR #832, simple custom content view support in scripts for mitmproxy/mitmdump.


mhils:
Thanks! :smiley: :sparkles: :cake:

PR #836, opened 2015-11-16
Fix #835 , can display unicode character correctly

### LINKED ISSUE #835: Can't parse unicode into character when viewing HTML content
### Step to reproduce the problem:
1. Open a HTML webpage in Chinese with mitmproxy.
2. Switch to the *Response* page and view the content.

###  What is the expected behavior?
Should display the content like 
``` html
<meta name="description" content="中文字符"/>
```
instead of
``` html
<meta name="description" content="&#20013;&#25991;&#23383;&#31526;"/>
```
### What went wrong?
I guess ```libmproxy.contentviews.ViewHTML``` should be improved to fix this. 

### Any other comments?
It's really confusing seeing whole screen of these unicodes instead of the correct characters. Hope we can fix this soon!

MatthewShao:
Fix this by myself! :smiley:  
Simply set encoding type to utf8 in ```lxml.etree.tostring``` function.

### DIFF
--- a/libmproxy/contentviews.py
+++ b/libmproxy/contentviews.py
@@ -228,7 +228,8 @@ class ViewHTML(View):
             s = lxml.etree.tostring(
                 d,
                 pretty_print=True,
-                doctype=docinfo.doctype
+                doctype=docinfo.doctype,
+                encoding='utf8'
             )
             return "HTML", format_text(s)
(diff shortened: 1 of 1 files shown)


mhils:
:+1: Thanks! :smiley:

PR #843, opened 2015-11-22
Added new proxy option: invisible. 

vitalyrepin:
If this option is set, mitmproxy does not send any error messages to the clients, stays silent and does not reveal its existence in the network.

It is useful to monitor / reverse engineer applications which are probing for proxy (e.g., by sending malformed requests to analyze the response) and are changing their behavior depending on the results.

### DIFF
--- a/libmproxy/cmdline.py
+++ b/libmproxy/cmdline.py
@@ -362,6 +362,11 @@ def proxy_options(parser):
         action="store", type=int, dest="port", default=8080,
         help="Proxy service port."
     )
+    group.add_argument(
+        "--invisible",
+        action="store_true", dest="invisible_proxy", default=False,
+        help="Make proxy invisible for the clients."
+    )
     http2 = group.add_mutually_exclusive_group()
     http2.add_argument("--http2", action="store_true", dest="http2")
     http2.add_argument("--no-http2", action="store_false", dest="http2",
--- a/libmproxy/protocol/http.py
+++ b/libmproxy/protocol/http.py
@@ -322,7 +322,8 @@ class HttpLayer(Layer):
                 # don't throw an error for disconnects that happen before/between requests.
                 return
             except NetlibException as e:
-                self.send_error_response(400, repr(e))
+		if not self.config.invisible:
+               	    self.send_error_response(400, repr(e))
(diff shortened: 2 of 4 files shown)


mhils:
Thanks for the PR.
I'd love to hear a better justification why this is needed. It will always be very easy to probe for a proxy on the wire in a deterministic way. From my point of view, these changes are only useful to hide mitmproxy from a unsuspecting human user and I can't think of an use case where I actively want to support this. Am I missing something? Do you have a concrete example where this is beneficial?

vitalyrepin:
Thanks for the feedback. I'm sorry for the late response - was busy with other tasks and now came back to mitmproxy-related issues.

In fact, it looks like my patch does not solve the problem it was supposed to address - make the proxy really "transparent"  for the application. In fact, sometimes application works worse if "invisible mode" is active - it sends request and waits for response which never comes. But connection is established from app PoV (because TCP ACK was received from the proxy).

What do you mean by "It will always be very easy to probe for a proxy on the wire in a deterministic way"?  How exactly? By sending commands to the proxy (e.g., CONNECT request)? What if the proxy does not answer to them / answers with HTTP error?

If I am looking for the real transparency (proxy which is invisible by the application) - shall I take a look at tproxy: https://www.balabit.com/support/community/products/tproxy  and combine it with mitm to be able to intercept https?  Or is there any easier way?

Thanks!

mhils:
> What do you mean by "It will always be very easy to probe for a proxy on the wire in a deterministic way"? How exactly?

You can just establish a TLS connection with a remote host and check if the certificate has been substituted. There's really no way to fix this - actively detecting an active MITM will (thankfully) always be possible.

vitalyrepin:
Ah. Yes - with TLS connections it's easy. By design.  

But my focus was plain http in fact. Some apps (e.g. skype) are trying to detect they are running behind the proxy and seem to change  their behavior depending on the environment (proxied / direct connection).

mhils:
Thanks for clarifying this :smiley:.
So, in general, mitmproxy is designed for active interception, which can more or less always be profiled. If you have plain HTTP, the client can send specially crafted messages that we can't parse properly (and indeed it looks like Skype is doing something like this?) - you can both detect the case where we send a error message reply and where we just close the connection. If you really want to stay invisible, you can use use mitmproxy's `--tcp` mode or passively record pcaps. Is there a case that these two don't cover?

PR #853, opened 2015-12-10
TLS verification warnings write to stderr

dweinstein:
I tested this and warnings no longer corrupt the file when I run

`mitmdump -q -w /dev/stdout > out` then `mitmdump -n -r out`

cc @mhils


mhils:
Thanks!

Issue #867, opened 2015-12-24
force send client cert


bazzinotti:
## Premise

Currently, mitmproxy supports client certs by way of the commandline argument `--client-certs [dir]` and expects cert filenames to be pre-named, apparently by target domain name (the documentation uses "example.org.pem" as an example). There are use-cases where it would more suitable to force the sending of a specific client-cert no matter the destination.

### Use Cases
* The correct client cert filename(s) are simply not known. The target(s) may have already been modified for MITM, which can be a lengthy process.
* a TLS session is established to a known destination with the client cert. A new destination for a TLS session is dynamically determined from within the current TLS session. The connection to the new host requires the same client cert. This inability to know the destination prevents the client-cert from being specified for the new connection.

I found this to be important during a reverse engineering experiment in which I knew the domain, but I saw in a packet capture that the client cert was not being sent .. I tried a variety of symlinks that included subdomain information and whatever else to no avail. Since I had to also deal with the dynamic host aspect, I had to modify mitmproxy.

So, I wound up modifying mitmproxy to automatically always send a specific client certificate. My modifications are trivial PoC (proof-of-concept) modifications right now, and are not proper command-line argument invocations. I'm interested in talking with you if you are experienced with mitmproxy codebase and would like to help me implement an official contribution.

The PoC hard-codes a client-cert in `libmproxy/models/connections.py` `establish_ssl` function. To further it into an official update, the cmdline arg will need to be developed. I am thinking of using `--force-client-cert`
`establish_ssl` function must check for the existence of a forced-client-cert file (provided via the cmdline arg), verify it, and use it. Else, it simply refers to its prior implementation.

If you would like to give me some mitmproxy-codebase knowledge regarding how to properly implement this feature, please do.

Thanks, I'm hoping to implement this feature via a pull request sometime soon.

mhils:
I think we can just keep using `--client-certs` and let it accept a filename as well (this of course needs to be documented), in which case we take the file for every connection. Thoughts?

bazzinotti:
I like that idea. I am inspired to develop my own pull request with the following: 

* the new feature -- "keep using --client-certs and let it accept a filename as well"
* Add documentation in `docs/certinstall.rst` (where the current --client-certs behavior is documented)

Are there any additional considerations for the contribution? I understand non-repo contributions will need to be made eventually (online documentation).

If you can, please guide me to code the feature in a style acceptable by the mitmproxy maintainers. I am interested mostly in advice on the following:

[...]

PR #870, opened 2015-12-28
[docs/libmproxy/test] Support single client-side cert file

bazzinotti:
This PR references the discussion made in Issue https://github.com/mitmproxy/mitmproxy/issues/867.

### DIFF
--- a/libmproxy/proxy/config.py
+++ b/libmproxy/proxy/config.py
@@ -133,12 +133,9 @@ def process_proxy_options(parser, options):
 
     if options.clientcerts:
         options.clientcerts = os.path.expanduser(options.clientcerts)
-        if not (os.path.exists(options.clientcerts) or
-                os.path.isdir(options.clientcerts) or
-                os.path.isfile(options.clientcerts)):
+        if not os.path.exists(options.clientcerts):
             return parser.error(
-                "Client certificate argument is not a file or directory, "
-                "or does not exist: %s" % options.clientcerts
+                "Client certificate path does not exist: %s" % options.clientcerts
             )
 
     if options.auth_nonanonymous or options.auth_singleuser or options.auth_htpasswd:
--- a/libmproxy/proxy/config.py
+++ b/libmproxy/proxy/config.py
@@ -133,10 +133,12 @@ def process_proxy_options(parser, options):
 
     if options.clientcerts:
         options.clientcerts = os.path.expanduser(options.clientcerts)
-        if not os.path.exists(options.clientcerts) or not os.path.isdir(options.clientcerts):
+        if not (os.path.exists(options.clientcerts) or
+                os.path.isdir(options.clientcerts) or
+                os.path.isfile(options.clientcerts)):
             return parser.error(
-                "Client certificate directory does not exist or is not a directory: %s" %
-                options.clientcerts
+                "Client certificate argument is not a file or directory, "
+                "or does not exist: %s" % options.clientcerts
             )
 
     if options.auth_nonanonymous or options.auth_singleuser or options.auth_htpasswd:
--- a/test/test_server.py
+++ b/test/test_server.py
@@ -1,3 +1,4 @@
+import os
 import socket
 import time
 from OpenSSL import SSL
@@ -316,6 +317,14 @@ class TestHTTPS(tservers.HTTPProxTest, CommonMixin, TcpMixin):
     clientcerts = True
 
     def test_clientcert(self):
+        self.config.clientcerts = os.path.join(
+            tutils.test_data.path("data/clientcert"), "client.pem")
+        f = self.pathod("304")
+        assert f.status_code == 304
+        assert self.server.last_log()["request"]["clientcert"]["keyinfo"]
+
+    def test_clientcerts(self):
+        self.config.clientcerts = tutils.test_data.path("data/clientcert")
         f = self.pathod("304")
         assert f.status_code == 304
         assert self.server.last_log()["request"]["clientcert"]["keyinfo"]
--- a/test/test_server.py
+++ b/test/test_server.py
@@ -314,21 +314,24 @@ class TestHTTPAuth(tservers.HTTPProxTest):
 class TestHTTPS(tservers.HTTPProxTest, CommonMixin, TcpMixin):
     ssl = True
     ssloptions = pathod.SSLOptions(request_client_cert=True)
-    clientcerts = True
 
-    def test_clientcert(self):
-        self.config.clientcerts = os.path.join(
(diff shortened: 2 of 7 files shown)


bazzinotti:
All Tests pass on my box.. The error in Travis CI build seems to be unrelated. Can someone shed light on this?

libmproxy/proxy/config.py
```diff
@@ -133,10 +133,12 @@ def process_proxy_options(parser, options):
 
     if options.clientcerts:
         options.clientcerts = os.path.expanduser(options.clientcerts)
-        if not os.path.exists(options.clientcerts) or not os.path.isdir(options.clientcerts):
+        if not (os.path.exists(options.clientcerts) or
```
mhils: This condition doesn't make much sense, `isdir/isfile` always imply ´.exists`. I think just checking for `.exists()` is good enough.

libmproxy/proxy/config.py
```diff
@@ -133,10 +133,12 @@ def process_proxy_options(parser, options):
...
+                os.path.isdir(options.clientcerts) or
+                os.path.isfile(options.clientcerts)):
             return parser.error(
-                "Client certificate directory does not exist or is not a directory: %s" %
-                options.clientcerts
+                "Client certificate argument is not a file or directory, "
+                "or does not exist: %s" % options.clientcerts
```
mhils: As above, we can simplify this to "Client certificate argument is neither a file nor a directory".

test/test_server.py
```diff
@@ -316,6 +317,14 @@ class TestHTTPS(tservers.HTTPProxTest, CommonMixin, TcpMixin):
...
 
     def test_clientcert(self):
+        self.config.clientcerts = os.path.join(
+            tutils.test_data.path("data/clientcert"), "client.pem")
+        f = self.pathod("304")
+        assert f.status_code == 304
+        assert self.server.last_log()["request"]["clientcert"]["keyinfo"]
+
+    def test_clientcerts(self):
+        self.config.clientcerts = tutils.test_data.path("data/clientcert")
```
mhils: This is the right way to implement it I think, but we should remove the `clientcerts = True` leftover in line 317 here and at https://github.com/mitmproxy/mitmproxy/blob/a978c6b9cee134d01f84216f3fbcf1ae9303d965/test/tservers.py#L133.
As a minor nitpick, please set `self.config.clientcerts` to `None` once the test is completed (you can use a `try: ... finally:` block).
Another nitpick: More expressive test names would be nice, e.g. `test_clientcert_file`/`test_clientcert_dir`  :-)

mhils:
This looks great, thanks :+1: 
Added some minor remarks above and restarted the tests (which now pass). Let me know once you updated the PR! :smiley:

libmproxy/proxy/config.py
bazzinotti: I was under the impression that using these function calls would help deter from special files being used (device files, fifo, etc.), in which case os.path.exists() could be removed. But since I could not find any mention of special files on https://docs.python.org/2/library/os.path.html I am not sure Python has such support. Thus, I'll go ahead and follow your recommendation and remove the isFile and isDir calls.

mhils: > I was under the impression that using these function calls would help deter from special files being used

Yes, and then we get bug reports like #788 :stuck_out_tongue: . 
Just checking for `.exists` is fine I think, if it's not readable we'll raise an exception anyway. :smile:

test/test_server.py
bazzinotti: You have obviously noticed that I changed the default behavior of the assignment of clientcerts from the setup of ProxTestBase class to the actual test function themselves. I did this for important reasons, that I would like you to analyze and assess.

Since the clientcerts assignment was made at object instantiation, the only way to have changed the setting would be to add additional logic to get_proxy_config(). I did this at first -- ProxTestBase's `clientcerts` class variable able to handle the numbers 0, 1 and 2 as a reference to whether the config should use None, file, or directory. The problem here is that the config is only made at object instantiation. This means another TestHTTPS class must be made with a different clientcert variable value. But, this also meant a re-running of all of the tests that belong to the classes that TestHTTPS inherits from (tservers.HTTPProxtTest, CommonMixin, TcpMixin).

This is what led me to investigate if I could manually change the real clientcerts proxy config from the test itself effectively -- No need to recreate objects just to change the clientcerts configuration.

The reason why I knew (or was convinced) that I could set the clientcerts config manually during the tests is because the class ProxyConfig `__init__` does not perform any special duties with the clientcerts dict argument returned from get_proxy_config() that I could not do outside of the object instantiation phase. A simple search for "clientcerts" in the project will show that it is not widely used and may be safe to control outside of object instantiation.


see https://github.com/mitmproxy/mitmproxy/blob/a978c6b9cee134d01f84216f3fbcf1ae9303d965/test/tservers.py#L127

According to your recommended removal of `clientcerts = True` from class TestHTTPS, it may be recommended to also have me remove the use of `clientcerts` in ProxTestBase. I'm referring to the class variable and the dict entry altogether. This would cause all proxies to use no clientcerts unless specifically configured as done in the new tests that are shown here.

Of course, this approach can only continue to work as long as the clientcerts setting can be so freely modified during the server's existence.

I've checked to make sure this logic is not used elsewhere. So, if you agree, I'll also remove this cruft from the ProxTestBase.

mhils: > But, this also meant a re-running of all of the tests that belong to the classes that TestHTTPS inherits from (tservers.HTTPProxtTest, CommonMixin, TcpMixin).

FWIW, the tests are only in CommonMixin and TcpMixin, so just inherting from tservers.HTTPProxtTest would have worked. However, I like your current approach better, so that doesn't matter in the end. :relaxed: 

> I've checked to make sure this logic is not used elsewhere. So, if you agree, I'll also remove this cruft from the ProxTestBase.

Yes, please! In fact, my previous comment linked to the place in ProxTestBase with the intent for it to be removed as well. Sorry if that was not clear. :smiley:

bazzinotti:
PR has just been updated

mhils:
Great job. Thanks a lot for the quick update! :cake: :smiley:

bazzinotti:
Wait, I cannot accept my cake yet, I just realized my test functions do not have blank lines between them. :dancer: Would you like me to fix that with a quick commit?

mhils:
I somehow need to get rid of the cake now. So, please send a quick commit if you'd like to! :smile:

bazzinotti:
OK well I pushed the commit. Can you receive it without further work on my part?

mhils:
All set, thanks! :cake:

bazzinotti:
*nom nom*

Issue #747, opened 2015-09-02
Add Inline Script Hooks to TCP mode


mhils:
Adding inline script hooks to the raw tcp mode ([libmproxy.protocol.rawtcp](https://github.com/mitmproxy/mitmproxy/blob/master/libmproxy/protocol/rawtcp.py)) would be an excellent addition to mitmproxy. There could be one hook on start and another one per message.

bazzinotti:
@mhils I'd like to help catalyze and motivate the upbringing of this feature request.

If you enjoyed the process we took with my first PR, please consider similarly providing insights on accomplishing this task and I will pursue the implementation.

Consider the following key points from #866:

- [ ] The TCP layer should emit events per message
- [ ] we need some form of container format at some point that also provides access to the client and server connection (see Flow and HTTPFlow).

Pertinent files
```
(Flow)
models/flow.py
models/http.py

protocol/rawtcp.py
```

Do you think it would be too difficult to steer me in the right direction?

mhils:
Great to hear you want to give this a shot! :smiley:

I think we should first focus on getting the events piped to inline scripts. The container/flow thing is a bit more abstract and we should leave that for a later PR (We really haven't wrapped our minds around this either).

You'll get the brief version as I'm on holiday, so please don't refrain from asking further questions:

- Every HTTP/TCP/... connection is handled in an isolated connection thread. You can communicate with the (single-threaded) FlowMaster over a channel. The FlowMaster then calls the inline script and hands a return value over to the channel.
- You can create a "handle_tcpmessage" method in the FlowMaster and then call `self.channel.ask("tcpmessage", obj)` in the TcpLayer to invoke the function on the master. In handle_tcpmessage, you just call the inline script stuff and return the result. https://github.com/mitmproxy/mitmproxy/commit/d002371d30e4b0ab7d1d23023236a9446d4c2396 is pretty similar to what you want to do.
- You can only pass a single object to `.ask`, but we probably want to transmit more than just the actual message. For now, I think it's perfectly okay if we polyfill this and just create a data class that holds everything we want:
```python
class TcpMessage(object): 
    def __init__(self, layer, message):
        self.layer = layer
        self.message = message
```
We can then just transmit this class thingie. Later on, we'd use the actual flow for this (as we do for HTTP).

if there's anything unclear, let me know. :smile: By the way, you're more than welcome to join us on Slack if you'd like - this usually shortens the turnaround times considerably.

[...]

PR #872, opened 2016-01-02
Add "tcp_message" Inline Script Hook

bazzinotti:
:dancer:  This early PR is the beginning of advancement on #747

@mhils Specifically, regarding the removal of the `layer` attribute from TcpMessage, I took out the `layer` from class `TcpMessage` and now send the `client_conn` and `server_conn` explicitly to the initializer... Let me know if you had something else in mind, as far as getting the `client_conn` and `server_conn` into the TcpMessage. 

### Inline Script Example
This is not yet included, but I am working on an inline script using the new feature that might be able to serve as an example script. More details to come on this later.

### Questions
Do new inline script hooks need unit tests??

Do any docs need updating?

### Documentation

The following is a key point of `TcpMessage`:

`sender` and `receiver` are mirrors to `client_conn` or `server_conn`. Together, the four are used to identify the origin of the message. For example,

```
if sender == client_conn:
    # we now know the client sent this message
```

I didn't put this information in the code, but I feel that maybe it belongs there :question: 
Or perhaps, only in more formal documentation, or both!

Direction and encouragement are always welcome! :sweat_smile:

### DIFF
--- a/test/test_server.py
+++ b/test/test_server.py
@@ -502,6 +502,18 @@ class TestHttps2Http(tservers.ReverseProxTest):
 class TestTransparent(tservers.TransparentProxTest, CommonMixin, TcpMixin):
     ssl = False
 
+    def test_tcp_stream_modify(self):
+        self.master.load_script(
+            tutils.test_data.path("scripts/tcp_stream_modify.py"))
+
+        self._tcpproxy_on()
+        d = self.pathod('200:b"foo"')
+        self._tcpproxy_off()
+
+        assert d.content == "bar"
+
+        self.master.unload_scripts()
+
 
 class TestTransparentSSL(tservers.TransparentProxTest, CommonMixin, TcpMixin):
     ssl = True
--- a/docs/scripting/inlinescripts.rst
+++ b/docs/scripting/inlinescripts.rst
@@ -36,14 +36,13 @@ We encourage you to either browse them locally or on `GitHub`_.
 Events
 ------
 
-.. TODO: Split this into Connection, HTTP and TCP events once we have TCP events.
-
 The ``context`` argument passed to each event method is always a
 :py:class:`~libmproxy.script.ScriptContext` instance. It is guaranteed to be the same object
 for the scripts lifetime and is not shared between multiple inline scripts. You can safely use it
 to store any form of state you require.
 
-Events are listed in the order they usually occur.
+Script Lifecycle Events
+^^^^^^^^^^^^^^^^^^^^^^^
 
 .. py:function:: start(context, argv)
 
@@ -51,6 +50,13 @@ Events are listed in the order they usually occur.
 
     :param List[str] argv: The inline scripts' arguments.
         For example, ``mitmproxy -s 'example.py --foo 42'`` sets argv to ``["--foo", "42"]``.
+
+.. py:function:: done(context)
+
+    Called once on script shutdown, after any other events.
+
+Connection Events
+^^^^^^^^^^^^^^^^^
 
 .. py:function:: clientconnect(context, root_layer)
 
@@ -64,14 +70,13 @@ Events are listed in the order they usually occur.
         :py:class:`~libmproxy.proxy.RootContext`. For example, ``root_layer.client_conn.address``
         gives the remote address of the connecting client.
 
+.. py:function:: clientdisconnect(context, root_layer)
 
-.. py:function:: request(context, flow)
+    Called when a client disconnects from the proxy.
 
-    Called when a client request has been received. The ``flow`` object is
-    guaranteed to have a non-None ``request`` attribute.
+    .. versionchanged:: 0.14
 
-    :param HTTPFlow flow: The flow containing the request which has been received.
-        The object is guaranteed to have a non-None ``request`` attribute.
+    :param Layer root_layer: see :py:func:`clientconnect`
 
 .. py:function:: serverconnect(context, server_conn)
 
@@ -80,6 +85,25 @@ Events are listed in the order they usually occur.
 
     :param ServerConnection server_conn: The server connection object. It is guaranteed to have a
         non-None ``address`` attribute.
+
+.. py:function:: serverdisconnect(context, server_conn)
+
+    Called when the proxy has closed the server connection.
+
+    .. versionadded:: 0.14
+
+    :param ServerConnection server_conn: see :py:func:`serverconnect`
+
+HTTP Events
+^^^^^^^^^^^
+
+.. py:function:: request(context, flow)
+
+    Called when a client request has been received. The ``flow`` object is
+    guaranteed to have a non-None ``request`` attribute.
+
+    :param HTTPFlow flow: The flow containing the request which has been received.
+        The object is guaranteed to have a non-None ``request`` attribute.
 
 .. py:function:: responseheaders(context, flow)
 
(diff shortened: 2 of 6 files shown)


bazzinotti:
All tests pass on my local box, using `py.test -n 4 --cov libmproxy`

mhils:
Thanks! You're too damn quick. :smile: 

This looks very good so far. It'd be great if we can enable the modification of streams right from the start. It should be enough if we just send the value of the `.message` attribute after all scripts have been run instead of sending `buf[:size]`. :smiley:

**Docs:** It would be nice to add the `tcp_message` to the event list at [docs/scripting/inlinescripts.rst](https://github.com/mitmproxy/mitmproxy/blob/master/docs/scripting/inlinescripts.rst), but it needs to include some form of warning that the API is subject to change. Other than that, an example script is enough documentation for now I think.

**Tests:** Yes, please. It should be enough to test the modification use case: Force tcp mode for the connection, then modify the server reponse in an inline script and validate in the test that this works. You can take a look at 
https://github.com/mitmproxy/mitmproxy/blob/f8d8a80bd07fe1e40577a2090ba47e3cc5abd3d3/test/test_server.py#L284 for something vaguely similar.

bazzinotti:
my latest commit passed with TravisCI but not with appveyor. I'm not sure how to identify the cause of failure. I take note that appveyor test is on win32 and python 2.7.11. however, one of the TravisCI tests passes using 2.7.11, so that's not the issue. All tests also pass locally as well (Linux 64bit / python 2.7.10).

mhils:
We have some really really ugly race conditions in the tests (not in mitmproxy, just in the tests) at the moment, apologies for this.

test/test_server.py
```diff
@@ -502,6 +502,20 @@ def test_http(self):
...
+        self.master.load_script(
+            tutils.test_data.path("scripts/tcp_stream_modify.py"))
+
+        self._tcpproxy_on()
+        d = self.pathod('200:b"foo"')
+        self._tcpproxy_off()
+
+        self.master.masterq.join()
+
+        assert any("bar" in m for m in self.master.log)
```
mhils: We should check if the response (`d`) contains bar as response, screw the log. :smile:

mhils:
LGTM except for the comment above!

bazzinotti:
**Status Update**: Documentation has been progressing just fine. In fact I should probably work on splitting my progress into separate smaller commits :wink: I hit a snag when writing the example inline script -- which has the following line (it should be obvious to decipher):

```
filehandle = (
   		ctx.client_filehandle if tcp_msg.sender == tcp_msg.client_conn
   		else ctx.server_filehandle
   	)
```

You can even begin to imagine what kind of example script I am developing :smile: 
But, I found myself mulling over the `tcp_msg.sender == tcp_msg.client_conn`. I thought to myself, "Wow, what a burden not having a helper function" -- and came up with the following addition to `TcpMessage`

```
    def sender_is_client(self):
        return self.sender == self.client_conn

    def receiver_is_client(self):
        return self.receiver == self.client_conn

    def sender_is_server(self):
        return self.sender == self.server_conn

    def receiver_is_server(self):
        return self.receiver == self.server_conn
```

Using these aids would simplify the writing of inline script logic, to:

```
filehandle = (
   		ctx.client_filehandle if tcp_msg.sender_is_client()
   		else ctx.server_filehandle
   	)
```

and it also prevents the user from needing to know how to compare the inner workings of the `TcpMessage`

Tell me, do you recommend I add these helper functions to the class? I became curious if there's some sort of downside to having all these TcpMessage instances now have this "added baggage" -- but I do tend to overthink things. I am curious to know whether external functions that accept the class instance provide a performance boost over putting the methods directly into the class definition and calling them directly from the class instances!

**TLDR**
To elaborate, I come from a C/CPP background in programming. I always had this tendency to believe that creating an external helper function that accepted a class object reference was better than putting the method into the class itself. I believe this because I imagine that the one external function would require less memory than all the class instances and their copies of the same method. In fact, this may be true, or it might not - I've never followed the path to proving it. but I'm just not sure it's worth racking my brain over. I imagine that Python works similarly -- a class with more function definitions will have a bigger instance than one without one -- so perhaps my C/CPP idiom is identical. Therefore, I hope you can take this consideration under your own wing and teach me a proper way to implement the helper functions, if they should be implemented at all.

Thanks!

bazzinotti:
I would appreciate some feedback on this doc snippet. 

* I took the burden of splitting the Events into Connection, TCP, and HTTP
image

Do you like the exclamation point image!? I added it into the project myself :smile: Though, only currently in the file's scope. It's really cool! I can call upon it by using `|warning|`

In the next image,  the 2 high level events -- `start` and `done` are not placed under a specific subsection. It looks like it could use one, right?! But I'm not sure of a good title other than "High Level Events"... :question: 

image

I could use some advice on that.

bazzinotti:
You'll see I just made 6 commits total. Unknowingly, I broke tests at the 3rd commit (The test itself breaks at commit 5 :x:  because I pushed all 5 together). I fixed the problem with my 6th commit. Now I'm wondering if I can safely integrate the 6th commit into the 3rd somehow -- so that testing isn't broken for the 2 commits in between..

bazzinotti:
@mhils I'll avoid rebasing the pull request until I receive further guidance and feedback.

At this point, I've made all my intended contributions. They are in your hands for review now.

Might I make some suggestions for your feedback session?
* Keep the feedback organized to [doc, example-script, rawtcp.py / TcpMessage]
* make the docs to see the product yourself

mhils:
#### docs

> I took the burden of splitting the Events into Connection, TCP, and HTTP.

Excellent, thanks - this looks good. You can just put `start` / `done` under a `Script Lifecycle` subsection or something like that.

------------

> Do you like the exclamation point image!?

I like it, but to be honest, I'd like to see a standard [`.. note:`](http://sphinx-doc.org/markup/para.html#directive-note) or `.. warning:` even more. :wink:

#### git

> I don't know how to do such a radical reworking of commits, especially during a pull request.

I know there are some projects which are pretty religious about a clean/linear Git history, but we (or at least I) don't care too much. If you like, you can squash them in a single commit, but I'm fine with it as-is.

#### code

> "Wow, what a burden not having a helper function"

Once we migrate to a flow object, we will have multiple messages per flow, so these methods won't make too much sense anymore. I think `tcp_msg.sender == tcp_msg.client_conn` is quite expressive in itself. Leave them if you'd like, but I won't make any promises how long they will survive :wink: 

-----------

> I am curious to know whether external functions that accept the class instance provide a performance boost over putting the methods directly into the class definition and calling them directly from the class instances!

is anwered by

> I do tend to overthink things.

This really sounds like premature optimization. :wink: 

-----------

> I always had this tendency to believe that creating an external helper function that accepted a class object reference was better than putting the method into the class itself. I believe this because I imagine that the one external function would require less memory than all the class instances and their copies of the same method.

FWIW, Class methods are not copied per instance/object, so there's no significant difference here. :smiley: 

-------------

The example script is nice (a bit too verbose for an example maybe), but pretty much breaks apart if we have multiple connections. How about just logging some info for each packet and maybe replacing `foo` with `bar` instead? At the very least, the sole point of the script is to demonstrate the API. You can also reduce the documentation for it in this case - a few print statements really don't need much explanation.

Thanks so far! Please take your time with updates to this, I think I won't be able to take a second look before this Saturday/Sunday. :smiley:

bazzinotti:
In the docs -> Inline Scripts page, in the section on `tcp_message`, I make reference to the `TCP Proxy` doc. This is provided to explain the `--tcp` flag that must be present to put the proxy into TCP mode.

`--raw-tcp` is undocumented, but it also conditionally puts the proxy into TCP mode. I have made no official reference to it. 

When the time comes to document `--raw-tcp`, I have the following recommendations. 

* Document it on the same page as `--tcp`
* Document it on different pages, but ensure that each page makes reference to each other

The following code block I drafted as a summary. I am not planning to incorporate it into the official doc (couldn't find a suitable place for it), but I like it so much I want to archive it here at least. Maybe you will suggest I use it somewhere.

```
Putting the Proxy into TCP Mode
-------------------------------

tcp_message events will only fire when the proxy is functioning in TCP mode.
This is only enabled via the following cmdline invocation flags:

--tcp [HOST-regexp] will unconditionally put the proxy into TCP mode if the
target host matches the regexp.

--raw-tcp puts the proxy into TCP mode when it cannot identify the
traffic. This currently equates to non HTTP/TLS traffic.
```

bazzinotti:
I definitely want to squash my commits when we are finished. Although I'm mildly concerned that squashing the commits will make the PR conversation incoherent. I've never squashed commits mid-PR before, but I'm confident in the following self-made solution: 

1. `git diff` to create the squashed commit patch
2. `git reset` to go back in time to redo my commit history
3. `git apply` to apply the patch
4. `git commit`
5.  `git push -f` to finalize it.

I don't know if there's a more popularly approved approach.

mhils:
> `--raw-tcp` is undocumented

Yes. We haven't properly introduced support for other protocols yet and will do this all at once. Thanks for pointing this out and the draft!

Everything else looks good to merge. For squasing, take a look at https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git.

Thanks! :smile: 
Max

bazzinotti:
New squashed commit has been pushed.

bazzinotti:
Thank you for working on this PR with me. I found it a very enjoyable programming experience.

mhils:
Thanks! :tada: :cake:

Issue #877, opened 2016-01-08
Bind address is not respected when contacting upstream servers


snemes:
##### Steps to reproduce the problem:
1. You need to have a configuration with at least 3 network interfaces. Let's say that the three interfaces are eth0 (with IP 10.0.0.1 - has no internet access), eth1 (with IP 10.1.0.1 - has access to the internet) and vboxnet0 (a VirtualBox host-only network with address 192.168.1.254) - the last interface does not need to be a VirtualBox host-only interface at all, it was just easier to use that as an example and it reflects my current configuration.

2. Have an iptables rule that forwards traffic from the VirtualBox interface to eth1 (the interface with access to the Internet):
iptables -t nat -A POSTROUTING -s 192.168.1.0/24 -j SNAT --to-source 10.1.0.1
Also have the required iptables rules for mitmproxy:
iptables -t nat -A PREROUTING -i vboxnet0 -p tcp -m tcp --dport 443 -j REDIRECT --to-ports 8080
iptables -t nat -A PREROUTING -i vboxnet0 -p tcp -m tcp --dport 80 -j REDIRECT --to-ports 8080

3. Launch mitmproxy/mitmdump with the arguments "-T -b 192.168.1.254"

The netstat command shows that the python script is listening on 192.168.1.254, which is fine.
However a Wireshark/tcpdump session quickly shows that the bind address is not respected when mitmproxy tries to contact the upstream HTTPS server, because the TCP packets use the source IP 10.0.0.1 (eth0 - the 1st interface, which is a network with no Internet access) - instead of the IP address 192.168.1.254 (which would be then forwarded by iptables to eth1 - the interface having internet access).

##### What is the expected behavior?
When the --bind-address or -b command line argument is used, mitmproxy should use that address as the source address when contacting the upstream servers.

---
mitmproxy version: 0.15
Operating System: Ubuntu 15.10

snemes:
Submitted pull request #879 that fixes this.

PR #879, opened 2016-01-08
Fixed a problem with the bind address not being used as the source address on outgoing TCP packets

snemes:


snemes:
Fixes https://github.com/mitmproxy/mitmproxy/issues/877

snemes:
Note: merging https://github.com/mitmproxy/netlib/pull/110 will fix the integration checks.

libmproxy/protocol/base.py
```diff
@@ -109,9 +109,9 @@ def __call__(self):
                         self.disconnect()
     """
 
-    def __init__(self, server_address=None):
+    def __init__(self, server_address=None, source_address=None):
         super(ServerConnectionMixin, self).__init__()
-        self.server_conn = ServerConnection(server_address)
+        self.server_conn = ServerConnection(server_address, source_address)
```
mhils: Can we take this out of the constructor here and just use the config value?
(This would also remove the other proxy mode cases)

libmproxy/models/connections.py
```diff
@@ -88,8 +88,10 @@ def finish(self):
 
 
 class ServerConnection(tcp.TCPClient, stateobject.StateObject):
-    def __init__(self, address):
-        tcp.TCPClient.__init__(self, address)
+    def __init__(self, address, source_address=None):
+        if source_address:
+            source_address = (source_address, 0)
```
mhils: This is a bit confusing (i.e. inconsistent) to be honest. Can we just pass it through as-is?

libmproxy/protocol/base.py
```diff
@@ -157,10 +157,11 @@ def disconnect(self):
         """
         self.log("serverdisconnect", "debug", [repr(self.server_conn.address)])
         address = self.server_conn.address
+        source_address = self.server_conn.source_address()[0]
```
mhils: Very minor nitpick: please use `source_address.host` instead.

mhils:
Thanks for the PR and sorry for the late reply - holidays! :smile: 
I added a few minor comments above. Please let me know once things are in place.

snemes:
I fixed them. I was not sure if I understood your request regarding the constructor (which constructor were you referring to, the ServerConnection or the ServerConnectionMixin constructor?), so please check if that's how you pictured it.

mhils:
Thanks! :smiley: My point was to use 
```python
self.server_conn = ServerConnection(server_address, (self.config.host, 0))
```
in the ServerConnectionMixin, so that we can keep all the modes as-is. Once that's done, LGTM! 
For the record, binding to two differents interfaces is another feature we may want to support at some point, but I'll leave that for a further PR.

snemes:
I am afraid that will not work. All the proxy modes inherit from the `Layer` and `ServerConnectionMixin` classes, and then use:

    super(<classname>, self).__init__(ctx, server_address=server_address)

in the constructor. This will, however never call ServerConnectionMixin's constructor.

mhils:
Yes, this is a tricky part of Python's inheritance system. `ServerConnectionMixin` will get called [here](
https://github.com/mitmproxy/mitmproxy/blob/7eb9bf818c6ab0b4fa341d741ead0a4199805473/libmproxy/protocol/base.py#L63), no need to worry. :smiley: 


Here's how that translates to your example (py3 notation):
```python
class A:
	def __init__(self):
		print("A")
		super().__init__()
		
class B:
	def __init__(self):
		print("B")
		super().__init__()

class C(A,B):
	def __init__(self):
		print("C")
		super().__init__()

snemes:
Ah, ok, I see what you mean. Sorry, I totally missed that call to super() in `ServerConnectionMixin`. I have made the necessary changes now.

mhils:
No problem, thanks!

PR #881, opened 2016-01-12
Fix #846

### LINKED ISSUE #846: Script Reloader Tests fail on OSX
The script reloader tests currently fail on OSX and are thereby disabled. The feature is awesome though, so we should get them running. Maybe just forcing the polling reloader [on OSX] could do the job.

MatthewShao:
Finally the script reloader works on both Linux and OS X!:tada:
1. I found that the reloader will crash if we load two scripts in same directory. Because we simply use `PatternMatchingEventHandler` to watch the all `*.py` file, when, say, two observer were set in the same, even the modification on the second script will hit the first observer. Then, the first observer will callback to reload the first script. So I pass the `filename` to `_ScriptModificationHandler` to make sure the modification will hit the corresponding observer.
2. The `Observer()` is kind of tricky. 
+ On OS X, `event.src_path` will only prints the path to the folder containing the file, not the path to the file itself. So I added some code to find out the script's real path. See: [Watchdog library in Python on OS X — not showing full event path](http://stackoverflow.com/questions/17537291/watchdog-library-in-python-on-os-x-not-showing-full-event-path)
+ For some reasons, multiple observers seem not work on OS X, typically will throw an exception when we stop them. So it need to be changed to `watchdog.observers.polling.PollingObserver()`, which will work on any platform.
+ On Linux, however, the `PollingObserver()` behave a little bit different than `Observer()`, it recognize all `modify` events as `creat` events. Therefore a `on_created` method need to be added to the `_ScriptModificationHandler`.

Already tested this on my Ubuntu 14.04 as well as OS X 10.10.

### DIFF
--- a/libmproxy/script/reloader.py
+++ b/libmproxy/script/reloader.py
@@ -38,7 +38,6 @@ class _ScriptModificationHandler(RegexMatchingEventHandler):
             regexes=['.*'+filename]
         )
         self.callback = callback
-        self.filename = filename 
 
     def on_modified(self, event):
         self.callback()
--- a/libmproxy/script/reloader.py
+++ b/libmproxy/script/reloader.py
@@ -5,7 +5,7 @@ if sys.platform == 'darwin':
     from watchdog.observers.polling import PollingObserver as Observer
 else:
     from watchdog.observers import Observer
-# The OSX reloader is watchdog 0.8.3 breaks when unobserving paths. 
+# The OSX reloader in watchdog 0.8.3 breaks when unobserving paths. 
 # We use the PollingObserver instead.
 
 _observers = {}
--- a/libmproxy/script/reloader.py
+++ b/libmproxy/script/reloader.py
@@ -10,7 +10,8 @@ def watch(script, callback):
     if script in _observers:
         raise RuntimeError("Script already observed")
     script_dir = os.path.dirname(os.path.abspath(script.args[0]))
-    event_handler = _ScriptModificationHandler(callback)
+    script_name = os.path.basename(script.args[0])
+    event_handler = _ScriptModificationHandler(callback, filename=script_name)
     observer = Observer()
     observer.schedule(event_handler, script_dir)
     observer.start()
@@ -25,7 +26,7 @@ def unwatch(script):
 
 
 class _ScriptModificationHandler(PatternMatchingEventHandler):
-    def __init__(self, callback, pattern='*.py'):
+    def __init__(self, callback, filename='*'):
         # We could enumerate all relevant *.py files (as werkzeug does it),
         # but our case looks like it isn't as simple as enumerating sys.modules.
         # This should be good enough for now.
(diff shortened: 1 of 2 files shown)


mhils:
Thanks for looking at this again! :smiley:
I wanted to give it a stab by myself, but didn't really find the time to do so - sorry.

> On OS X, [...]

This is fixed by using the PollingObserver, right?

> For some reasons, multiple observers seem not work on OS X

Yes, that's the reason we had it disabled until now. However, I think we want to use the native observers on Linux/Windows, so how about only using the PollingObserver on OSX?

>  two observer were set in the same, even the modification on the second script will hit the first observer.

The advantage is that we pick up changes in imported modules, the problem is, as you say, the double-reloads. I'm open to what tradeoff we want to favor. In the latter case, why don't we just pass the `os.path.abspath(script.args[0])` to the Observer?

MatthewShao:
> This is fixed by using the PollingObserver, right?

Yes! PollingObserver works fine on OS X! I also have considered to only use PollingObsever on OS X, which need to judge the system first, but I haven't done this before, any suggestion? If we use native observer on Linux, we don't need to consider `on_created` case any more.

> the problem is, as you say, the double-reloads.

I use `os.path.basename()` to extract the filename and use `fnmatch.fnmatsh` to judge if it is the script file correspond  to the observer. For the time being, most of scripts were written in single `.py` file, and I think most of users will store them in same directory, just like the `mitmproxy/examples` directory. So if any modification on `*.py` in the same directory will hit the observer, it will be unreasonable.
If we want to detect the changes in imported modules, I can't find any way other than parse the script file and add a observer to the imported modules. But it sounds tricky, right?

mhils:
> I also have considered to only use PollingObsever on OS X, which need to judge the system first, but I haven't done this before, any suggestion? 

Let's try this. The native Linux/Windows observer should be way faster than the polling one. :smiley: 

> I use os.path.basename() to extract the filename and use fnmatch.fnmatsh to judge if it is the script file correspond to the observer.

Maybe a dumb question: Can't we just pass the full file path (i.e. dir+filename) to an observer and thereby only get events for that file?

> If we want to detect the changes in imported modules, I can't find any way other than parse the script file and add a observer to the imported modules. But it sounds tricky, right?

Yes. This is tricky - I think we should skip this for now.

MatthewShao:
> Maybe a dumb question: Can't we just pass the full file path (i.e. dir+filename) to an observer and thereby only get events for that file?

I'm afraid we can only pass the **directory path** to `observer.schedule` according to the documents.:pensive: Actually, I try this by my self, it turns out to be a exception like `OSError: Path is not a directory`.

As you said, we use `PollingObserver` on OS X and native `Observer` on Linux and Windows now!:smile:

libmproxy/script/reloader.py
```diff
@@ -1,6 +1,12 @@
 import os
+import fnmatch
+import sys
 from watchdog.events import PatternMatchingEventHandler
-from watchdog.observers import Observer
+if sys.platform == 'darwin':
+    from watchdog.observers.polling import PollingObserver as Observer
+else:
+    from watchdog.observers import Observer
+# Use PollingObserver on OS X, and natvie Observer on Linux and Windows
```
mhils: This comment is not very helpful for the reader. How about "The OSX reloader is watchdog 0.8.3 breaks when unobserving paths. We use the PollingObserver instead."?

libmproxy/script/reloader.py
```diff
@@ -24,17 +31,27 @@ def unwatch(script):
...
 class _ScriptModificationHandler(PatternMatchingEventHandler):
-    def __init__(self, callback):
-        # We could enumerate all relevant *.py files (as werkzeug does it),
-        # but our case looks like it isn't as simple as enumerating sys.modules.
-        # This should be good enough for now.
+    def __init__(self, callback, filename='*'):
+
         super(_ScriptModificationHandler, self).__init__(
             ignore_directories=True,
-            patterns=["*.py"]
```
mhils: Can't we just pass the filename here and then avoid all the stuff below?

MatthewShao:
Yes, you are right! We can pass the filename directly to `_ScriptModificationHandler`, but should use ` RegexMatchingEventHandler` instead. We can cut down the dummy code in `on_modified` by doing this.
Thanks Max!

libmproxy/script/reloader.py
```diff
@@ -23,18 +30,18 @@ def unwatch(script):
...
+class _ScriptModificationHandler(RegexMatchingEventHandler):
+    def __init__(self, callback, filename='.*'):
+
         super(_ScriptModificationHandler, self).__init__(
             ignore_directories=True,
-            patterns=["*.py"]
+            regexes=['.*'+filename]
         )
         self.callback = callback
+        self.filename = filename
```
mhils: Last minor nitpick: Is this even necessary?

MatthewShao: Yeah, we don't need that any more, sorry for my carelessness. :flushed:

mhils: No problem. Thanks for the PR!

PR #882, opened 2016-01-12
added sslstrip to inline script examples

tinius:
Added sslstrip.py to the inline script examples. The script mimics the functionality of Moxie Marlinspike's popular standalone tool.

### DIFF
--- a/examples/sslstrip.py
+++ b/examples/sslstrip.py
@@ -9,7 +9,6 @@ def start(context, argv) :
 
 def request(context, flow) :
 
-    flow.request.headers.pop('Accept-Encoding', None)
     flow.request.headers.pop('If-Modified-Since', None)
     flow.request.headers.pop('Cache-Control', None)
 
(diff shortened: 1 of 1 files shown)


examples/sslstrip.py
```diff
@@ -0,0 +1,41 @@
...
+from six.moves import urllib
+
+def start(context, argv) :
+
+    #set of SSL/TLS capable hosts
+    context.secure_hosts = set()
+
+def request(context, flow) :
+
+    flow.request.headers.pop('Accept-Encoding', None)
```
mhils: Why do we strip off the Accept-Encoding header here?

mhils:
Thanks for the awesome PR! :smiley: :tada: 
Just a very minor remark above - everything else looks gread and good to merge!

examples/sslstrip.py
tinius: Now that you mention it, I don't think stripping that is necessary. It's part of Moxie's script but probably just because his proxy can't handle gzipped responses. mitmproxy should be just fine.

mhils:
Thanks! :cake:

PR #883, opened 2016-01-12
HTTP/2: Implementation using hyper-h2

Kriechi:
mitmproxy now has a fully working HTTP/2 implementation based on `hyper-h2` and `hyperframe`!

* Streaming is working
* Multiple concurrent streams are working
* Large file downloads are working
* Stream resets are propagated to the other side
* Setting changes are propagated to the other side

TODOs:
  - [x] check for memory leaks and do some cleaning of zombie streams
  - [x] discuss if we should move some parts to `netlib` and/or `pathod` (no we shouldn't)
  - [x] should we propagate `GoAway` to the other side? (idle timeout, error, ...)
  - [x] add script hooks (@mhils: all hooks already in place)
  - [x] flow control is still YOLO<sup>tm</sup>: Throttling of outgoing data is necessary (to client, to server)
  - [x] add timestamps to request and response objects
  - [x] implement PushPromise and map it to new flows
  - [x] check tests and coverage
  - [x] wait for `hyper-h2` and `hyperframe` releases (and pending pull requests)


Lukasa:
\o/ This is super exciting! I may need to download and play with this a bit today.

mhils:
:+1: 

This is definitely better than what we had before.

I don't feel that we need to move stuff to netlib. For pathod, we probably just want to implement fuzzing with hyperframe (without a proper state engine as hyper-h2) and some syntactic sugar.

Script Hooks: What do you have in mind? As I understand it, we have the normal HTTP events from the HTTP layer. If that's the case, I'd be fine for now.

One problem I'm still seeing is that the Http2SingleStreamLayer does not expose a full Layer API. For example, if we receive a HTTP request that signals an upgrade to WebSockets, we'll put a tcp/passthrough layer on our protocol stack. The TCP Layer will then just read from the underlying sockets, which is not going to end well. Re-using the code here is tricky. I tried to tackle this in https://github.com/mitmproxy/mitmproxy/compare/http, but the result is overly complex. The simple workaround would be to overwite `next_layer()` to always raise an Exception.

I'm still a bit worried that we're pointing a gun on our foots with one thread per stream. Let's see :innocent: 

Great work!

Lukasa:
If you're worried about one-thread-per-stream, consider setting SETTINGS_MAX_CONCURRENT_STREAMS to a reasonable value (e.g. 50, 100, whatever). That will bound the state commitment mitmproxy has to make, and hyper-h2 will enforce the value you choose.

Lukasa:
(Do note that you probably want to set that value to the *same* value on both sides of mitmproxy, to avoid the risk of one side pushing you over the limit you can bear on the other!)

mhils:
> If you're worried about one-thread-per-stream, consider setting SETTINGS_MAX_CONCURRENT_STREAMS to a reasonable value (e.g. 50, 100, whatever). That will bound the state commitment mitmproxy has to make, and hyper-h2 will enforce the value you choose.

I'm more concerned about race conditions which will have the potential to be horrible to debug. But hey, I can leave that to @Kriechi :smile: :wink: 

In any case, this is a good suggestion and we should do it.

Kriechi:
Yes, I'm mirroring all settings from the client<->mitmproxy to mitmproxy<->server. So Ideally mitmproxy can handle as many concurrent streams as the client & server announce in their settings. Only Push is disabled for now (needs some more implementation regarding mitmproxy GUI and flows)

Script hooks: ok, I wasn't sure if there are any missing, but if you say everything is in place, then :+1: 

I still believe that one-thread-per-stream approach is the proper way to do it. So far it seems to work very well.

libmproxy/protocol/http.py
```diff
@@ -130,102 +131,333 @@ def __call__(self):
...
+        self.streams = dict()
+        self.server_to_client_stream_ids = dict([(0, 0)])
+        self.client_conn.h2 = SafeH2Connection(self.client_conn, client_side=False)
+
+        # make sure that we only pass actual SSL.Connection objects in here,
+        # because otherwise ssl_read_select fails!
+        self.active_conns = [self.client_conn.connection]
+
+        if self.server_conn:
+            self._initiate_server_conn()
```
mhils: We shouldn't alter the connection state in the constructor. Can we do this in `__call__`?

Rmohid:
Great work !
I'm able to proxy h2 traffic but unable to "view flow", is this still a wip or is there more required than what's detailed in the markdown with the hyper-h2 branch ?

Kriechi:
Thanks @Rmohid! I spotted the problem and will work on it in the next days. Stay tuned!

Rmohid:
Awesome !

Kriechi:
@Rmohid I think 4fab60f will solve your problem - let me please know!

Rmohid:
I can see the flows now, thanks. I was hoping server pushes would be picked up in the flow, but they don't seem to show up :(

Kriechi:
@Rmohid sorry, PushPromise is not implemented yet - and therefor disabled on server & client-side connections. We are happy to accept new PRs though! :smile:

Kriechi:
@Rmohid that was easier than expected - could you give it a try? I'm not sure if it works properly, how do you test it? Firefox, Chrome, or something else?

Rmohid:
I'm using chrome and  nghttp  with a hacked version of http2.golang.org that statically replies with server pushes (can put on github if you want) 

Right now I see the push promises going through mitm to the browser unscathed, but the push promise frames don't show up in mitmdump or mitmproxy views (or on chrome for that matter) but I do see the get requests for the pushed content come back to the origin, so I can confirm that it is working... just transparently ...

Kriechi:
Yes, I think it is supposed to work transparently.
The only thing I add is a header called `x-mitmproxy-push` with `true`.
If you see that in your request, then this request originated as PushPromise from the server.

libmproxy/models/http.py
```diff
@@ -11,9 +11,10 @@
 from .. import version, stateobject
 from .flow import Flow
 
+from collections import OrderedDict
 
 class MessageMixin(stateobject.StateObject):
-    _stateobject_attributes = dict(
+    _stateobject_attributes = OrderedDict(
```
mhils: Is that to fix the pypy serialization?

libmproxy/protocol/http.py
```diff
@@ -400,7 +235,7 @@ def send_error_response(self, code, message):
         try:
             response = make_error_response(code, message)
             self.send_response(response)
-        except NetlibException:
+        except NetlibException as H2Error:
```
mhils: ?

test/test_protocol_http2.py
```diff
@@ -0,0 +1,389 @@
...
+from libmproxy.proxy.server import ProxyServer
+from libmproxy.cmdline import APP_HOST, APP_PORT
+
+import logging
+logging.getLogger("hyper.packages.hpack.hpack").setLevel(logging.WARNING)
+logging.getLogger("requests.packages.urllib3.connectionpool").setLevel(logging.WARNING)
+logging.getLogger("passlib.utils.compat").setLevel(logging.WARNING)
+logging.getLogger("passlib.registry").setLevel(logging.WARNING)
+logging.getLogger("PIL.Image").setLevel(logging.WARNING)
+logging.getLogger("PIL.PngImagePlugin").setLevel(logging.WARNING)
```
mhils: Maybe move that to __init__?

libmproxy/script/script.py
```diff
@@ -88,6 +88,8 @@ def run(self, name, *args, **kwargs):
             Raises:
                 ScriptException if there was an exception.
         """
+        if not self.ns:
```
mhils: This may hide errors at other places - do we want to raise a ScriptError here?

libmproxy/models/http.py
Kriechi: Yes, you are correct. 
I checked what `dict.iteritems()` returns on python2.7 and pypy. And the order in which the items are yielded is different. This results in a data dependency between `headers` and `content`, since setting the content also adds a header, which might be overwritten later.

test/test_protocol_http2.py
Kriechi: We don't have a **init** (yet)?

libmproxy/script/script.py
Kriechi: It is already crashing here in some testcases. This was the simplest fix for me. Maybe you can investigate where this comes from?

```
Exception in thread ProxyThread (0.0.0.0:49486):
  File ".../2.7/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File ".../mitmproxy/test/tservers.py", line 76, in run
    self.tmaster.run()
  File ".../mitmproxy/libmproxy/controller.py", line 123, in run
    self.tick(self.masterq, 0.1)
  File ".../mitmproxy/libmproxy/flow.py", line 809, in tick
    return super(FlowMaster, self).tick(q, timeout)
  File ".../mitmproxy/libmproxy/controller.py", line 108, in tick
    self.handle(*msg)
  File ".../mitmproxy/libmproxy/controller.py", line 130, in handle
    m(obj)
  File ".../mitmproxy/libmproxy/flow.py", line 958, in handle_serverdisconnect
    self.run_script_hook("serverdisconnect", server_conn)
  File ".../mitmproxy/libmproxy/flow.py", line 691, in run_script_hook
    self._run_single_script_hook(script_obj, name, *args, **kwargs)
  File ".../mitmproxy/libmproxy/flow.py", line 685, in _run_single_script_hook
    script_obj.run(name, *args, **kwargs)
  File ".../mitmproxy/libmproxy/script/script.py", line 90, in run
    f = self.ns.get(name)
AttributeError: 'NoneType' object has no attribute 'get'
```

libmproxy/models/http.py
mhils: kk. Let's add a quick comment explaining that so we can fix that nicely at a later point.

teerupkant:
Please help me 

Any documents about *Mitmproxy*

How to read /add/ update and response realtime using python ?

I need to check how you can read headers and write in to headers on request and responses 


Thank you

Issue #808, opened 2015-10-27
Export/copy request as code


areichow:
It would be great if mitmproxy allowed you to generate a snippet of code, presumably in Python, that would issue the same request. 

Related to the following two requests:
* [Export/copy request as raw text](https://github.com/mitmproxy/mitmproxy/issues/807)
* [Export as curl command](https://github.com/mitmproxy/mitmproxy/issues/596)

mhils:
Thanks for the suggestion. We'd be super happy to accept a PR for this!

nurav:
@mhils Hi, I'd like to work on this. Could you give me a few pointers so that I know how to go about it?

mhils:
@nurav: Great to hear that! @gato has openend PR #799 for curl export (which I still didn't get to review properly, sorry :see_no_evil: ). Maybe it's best if you set up a dev environment (see the README), take a look at Marcelos PR and work on based on that. :smiley:

nurav:
I'll do that. Thanks!

dufferzafar:
#884 only added support for copying [GET requests](https://github.com/mitmproxy/mitmproxy/pull/884/files#diff-9641efb691a86154d6143b6ec654f8f7R307) as Python code. I am working on adding others.

Kriechi:
closed by #916

Buff-r:
Downloaded the latest mitmproxy version can't find a way to export as curl command! its annoying! was it removed?

ujjwal96:
@Buff-r It can be done by pressing <kbd>e</kbd>

Issue #807, opened 2015-10-27
Export/copy request as raw text


areichow:
A feature I've been using in Fiddler is the ability to export the request as raw text. Once saved, the raw request easily replayed (with or without manual or programmatic modification) with netcat with a command like this:

```
nc api.host.com 80 < raw-request.txt > response.txt
```

mhils:
refs #799

gato:
Will add this to #799 in the next couple of days

mhils:
:+1:

dufferzafar:
@mhils I think this can be closed, since #884 was merged!

PR #884, opened 2016-01-13
Export/Copy flow as...

### LINKED ISSUE #596: Export as curl command

dufferzafar:
This pull supersedes #799 & #619. I've taken the code from pull #799 by @gato, rebased it and made my own changes.

Fixes #596 (export as curl) and #808 (export as python code)

~~I will also add support for exporting as raw http request - #807.~~ I've added code for copying raw requests (but am not sure if it is the right way.)

For the python code part, I've used the [requests](https://readthedocs.org/projects/requests) module as it has become quite a standard.
Currently, only GET methods are supported, but if you feel that the general implementation is correct then I can extend it to work for other methods as well.

While testing out export as curl, I felt having an option for exporting as [httpie](https://github.com/jkbrzt/httpie) might be nice too. Thoughts?

### DIFF
--- a/libmproxy/console/common.py
+++ b/libmproxy/console/common.py
@@ -276,6 +276,10 @@ def copy_flow_format_data(part, scope, flow):
                     raise ValueError("Unknown part: {}".format(part))
     return data, False
 
+def export_prompt(k, flow):
+    if k == "c":
+        copy_as_curl_command(flow)
+
 def copy_as_curl_command(flow):
 
     if flow.request.content is None or flow.request.content == CONTENT_MISSING:
--- a/libmproxy/console/common.py
+++ b/libmproxy/console/common.py
@@ -242,7 +242,7 @@ def ask_save_path(prompt, data):
     signals.status_prompt_path.send(
         prompt = prompt,
         callback = ask_save_overwrite,
-        args = (data)
+        args =  (data, ) 
     )
 
 
--- a/libmproxy/console/common.py
+++ b/libmproxy/console/common.py
@@ -276,9 +276,13 @@ def copy_flow_format_data(part, scope, flow):
                     raise ValueError("Unknown part: {}".format(part))
     return data, False
 
+
 def export_prompt(k, flow):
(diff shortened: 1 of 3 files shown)


dufferzafar:
Hey! Is there anything amiss from the pull and I need to fix? or is it just on hold for now? :)

libmproxy/console/common.py
```diff
@@ -277,26 +277,73 @@ def copy_flow_format_data(part, scope, flow):
...
+    elif k == "p":
+        copy_as_python_code(flow)
+    elif k == "r":
+        copy_as_raw_request(flow)
 
-    if err:
-        signals.status_message.send(message=err)
+
+def copy_as_curl_command(flow):
+    if flow.request.content is None or flow.request.content == CONTENT_MISSING:
```
mhils: This restriction seems unnecessary - any reason why we want to do this?

mhils:
Thanks for the great PR and sorry for the super slow response! :smiley:
This looks really good so far and is certainly an improvement over what we had before. 
I think we can optimize both the curl and Python export a bit, but that's something we can do in a later PR, let's first make sure that we finally land this!

libmproxy/console/common.py
dufferzafar: Um, to be quite honest, I don't totally get this line. It was already a part of pull #799 and since I couldn't think of a reason to remove it either. I just kept it. [Line](https://github.com/mitmproxy/mitmproxy/pull/799/files#diff-9641efb691a86154d6143b6ec654f8f7R281)

mhils: Ok. Let's just remove it here and at the other occurences. :smiley:

mhils:
Thanks @dufferzafar, @Fuzion24  and @gato!
Super sorry that this took so long, super happy that we finally have a great export functionality! :smiley:

Fuzion24:
:+1:

Issue #889, opened 2016-01-21
Add CLI option for "following" mode


Kidlike:
The one when you press Shift+f :)

Couldn't find anything on the docs or --help. I assume it's not there.

PR #891, opened 2016-01-22
Added CLI argument to enable "follow flow list" mode

ganguera:
This pull request adds to mitmproxy (cli) an argument to enable the "follow flow list" as requested in #889.


mhils:
Great work, thanks for the PR! :smiley: :tada: :cake:

PR #893, opened 2016-01-24
Add ignore based on TLS ClientHello SNI

dweinstein:
cc @mhils

### DIFF
--- a/libmproxy/protocol/tls.py
+++ b/libmproxy/protocol/tls.py
@@ -221,6 +221,75 @@ def is_tls_record_magic(d):
         d[2] in ('\x00', '\x01', '\x02', '\x03')
     )
 
+def get_client_hello(client_conn):
+    """
+    Peek into the socket and read all records that contain the initial client hello message.
+
+    client_conn:
+        The :py:class:`client connection <libmproxy.models.ClientConnection>`.
+
+    Returns:
+        The raw handshake packet bytes, without TLS record header(s).
+    """
+    client_hello = ""
+    client_hello_size = 1
+    offset = 0
+    while len(client_hello) < client_hello_size:
+        record_header = client_conn.rfile.peek(offset + 5)[offset:]
+        if not is_tls_record_magic(record_header) or len(record_header) != 5:
+            raise TlsProtocolException('Expected TLS record, got "%s" instead.' % record_header)
+        record_size = struct.unpack("!H", record_header[3:])[0] + 5
+        record_body = client_conn.rfile.peek(offset + record_size)[offset + 5:]
+        if len(record_body) != record_size - 5:
+            raise TlsProtocolException("Unexpected EOF in TLS handshake: %s" % record_body)
+        client_hello += record_body
+        offset += record_size
+        client_hello_size = struct.unpack("!I", '\x00' + client_hello[1:4])[0] + 4
+    return client_hello
+
+class TlsClientHello(object):
+    def __init__(self, raw_client_hello):
+        self._client_hello = ClientHello.parse(raw_client_hello)
+
+    def raw(self):
+        return self._client_hello
+
+    @property
+    def client_cipher_suites(self):
+        return self._client_hello.cipher_suites.cipher_suites
+
+    @property
+    def client_sni(self):
+        for extension in self._client_hello.extensions:
+            if (extension.type == 0x00 and len(extension.server_names) == 1
+                    and extension.server_names[0].type == 0):
+                return extension.server_names[0].name
+
+    @property
+    def client_alpn_protocols(self):
+        for extension in self._client_hello.extensions:
+            if extension.type == 0x10:
+                return list(extension.alpn_protocols)
+
+    @classmethod
+    def from_client_conn(cls, client_conn):
+        """
+        Peek into the connection, read the initial client hello and parse it to obtain ALPN values.
+        client_conn:
+            The :py:class:`client connection <libmproxy.models.ClientConnection>`.
+        Returns:
+            :py:class:`client hello <libmproxy.protocol.tls.TlsClientHello>`.
+        """
+        try:
+            raw_client_hello = get_client_hello(client_conn)[4:]  # exclude handshake header.
+        except ProtocolException as e:
+            raise TlsProtocolException('Cannot parse Client Hello: %s' % repr(e))
+
+        try:
+            return cls(raw_client_hello)
+        except ConstructError as e:
+            #self.log("Raw Client Hello: %s" % raw_client_hello.encode("hex"), "debug")
+            raise TlsProtocolException('Cannot parse Client Hello: %s' % repr(e))
 
 class TlsLayer(Layer):
     def __init__(self, ctx, client_tls, server_tls):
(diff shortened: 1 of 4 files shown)


libmproxy/proxy/root_context.py
```diff
@@ -33,6 +59,7 @@ def __init__(self, client_conn, config, channel):
         self.client_conn = client_conn
         self.channel = channel
         self.config = config
+        self._client_tls = False
```
mhils: Is this just for caching the magic check? I think we should avoid additional state as much as possible.

libmproxy/proxy/root_context.py
```diff
@@ -47,6 +74,7 @@ def next_layer(self, top_layer):
         layer = self._next_layer(top_layer)
         return self.channel.ask("next_layer", layer)
 
+    @tls_sni_check_ignore
```
mhils: This decorator-approach feels a bit weird. 
Can we just do this explicitly in the `_next_layer` function? We can move the `is_tls_record_magic` part to the top, then check if `self.config.check_ignore or self.config.check_tcp` is True and then parse the ClientHello if so.

mhils:
This looks very good so far, thanks @dweinstein!

libmproxy/proxy/root_context.py
dweinstein: I'll take that out

mhils: :+1:

libmproxy/proxy/root_context.py
dweinstein: I was doing it because there were a few exit points where a TlsLayer was being created (e.g., for scripts) and I wasn't sure if they might be upgraded with SNI (and subsequently ignored if desired). But of course I can remove the decorator and move the check directly into _next_layer.

If I move the `check_tcp` up will that impact any of the later cases? before it was happening at step 4, after CONNECT etc.

I'll have a patch incoming soon so you can review.

mhils: > I was doing it because there were a few exit points where a TlsLayer was being created (e.g., for scripts) and I wasn't sure if they might be upgraded with SNI (and subsequently ignored if desired). 

Not sure if I understand your point. How does the decorator fix this?

> If I move the `check_tcp` up will that impact any of the earlier cases? before it was happening at step 4, after CONNECT etc.

Sorry, not sure what I thought when writing my previous comment :smile:.
Please keep the order as-is. We can start with checking for `is_tls_record_magic`, then parse ClientHello if check_ignore is not None, then continue with check_ignore and the rest as-is.

dweinstein:
@mhils take a look. If you like what you see I can squash these down into one commit. I'm also doing some more testing on my end while you review.

libmproxy/proxy/root_context.py
```diff
@@ -58,6 +57,16 @@ def _next_layer(self, top_layer):
...
 
+        # 1A. check for --ignore with SNI host
+        if client_tls:
+            try:
+                client_hello = TlsClientHello.from_client_conn(self.client_conn)
+                if (client_hello and
+                        self.config.check_ignore((client_hello.client_sni, 443))):
+                    return RawTCPLayer(top_layer, logging=False)
+            except TlsProtocolException as e:
+                six.reraise(ProtocolException, ProtocolException(str(e)), sys.exc_info()[2])
```
dweinstein: @mhils I wasn't positive about what to do with the exception here...

libmproxy/protocol/tls.py
```diff
@@ -359,6 +388,10 @@ def sni_for_server_connection(self):
     def alpn_for_client_connection(self):
         return self.server_conn.get_alpn_proto_negotiated()
 
+    @property
```
mhils: This is a leftover?

libmproxy/proxy/root_context.py
mhils: I'm afraid we need to mute that - we had some edge cases which we didn't parse properly before, so I don't want to screw up the whole thing because of that.

dweinstein: so what would you like me to do here, comment it out? pass the exception? please be specific

libmproxy/protocol/tls.py
dweinstein: removing

libmproxy/protocol/tls.py
```diff
@@ -281,60 +355,15 @@ def __repr__(self):
...
-                    self.log("Unknown Server Name Indication: %s" % extension.server_names, "error")
-                self.client_sni = extension.server_names[0].name
-            elif extension.type == 0x10:
-                self.client_alpn_protocols = list(extension.alpn_protocols)
-
-        self.log(
-            "Parsed Client Hello: sni=%s, alpn=%s" % (self.client_sni, self.client_alpn_protocols),
-            "debug"
-        )
+        parsed = TlsClientHello.from_client_conn(self.client_conn)
```
mhils: We need to catch the exception here and log it (see L320 previously)

mhils:
We still have some confusion how `_next_layer` should look like - sorry if I haven't been clear on that. Here's what I propose:

```python
     def _next_layer(self, top_layer):
         try:
             d = top_layer.client_conn.rfile.peek(3)
         except TcpException as e:
             six.reraise(ProtocolException, ProtocolException(str(e)), sys.exc_info()[2])
         client_tls = is_tls_record_magic(d)

         # 1. Check for --ignore.
         if self.config.check_ignore:
              address = top_layer.server_conn.address
              if client_tls:
                  try:
                      client_hello = TlsClientHello.from_client_conn(self.client_conn)
                  except TlsProtocolException as e:
                      self.log(...) # see L320 of protocol.tls
                  else:
                      address = (client_hello.client_sni, 443) # TODO: may need to wrap that in tcp.Address?
             if self.config.check_ignore(address):
                 return RawTCPLayer(top_layer, logging=False)

          # 2. Always
```

libmproxy/protocol/tls.py
dweinstein: done in 9fad2e6941c3fff587d833e17c60d3f643f7454a

dweinstein:
@mhils ok let me know how things are looking now.

libmproxy/protocol/tls.py
```diff
@@ -281,60 +355,18 @@ def __repr__(self):
...
-        try:
-            client_hello = ClientHello.parse(raw_client_hello)
-        except ConstructError as e:
+            parsed = TlsClientHello.from_client_conn(self.client_conn)
+            self.client_sni = parsed.client_sni
+            self.client_alpn_protocols = parsed.client_alpn_protocols
+            self.client_ciphers  = parsed.client_cipher_suites
+        except TlsProtocolException as e:
             self.log("Cannot parse Client Hello: %s" % repr(e), "error")
-            self.log("Raw Client Hello: %s" % raw_client_hello.encode("hex"), "debug")
```
mhils: Minor nitpick: Let's restore the debug statement here, otherwise I'll have a hard time figuring out ClientHello bug reports :wink:

mhils:
This looks good to merge now. Last two minor issues are (1) the comment above and (2) the documentation at http://docs.mitmproxy.org/en/stable/features/passthrough.html#how-it-works. It'd be super awesome if you could quickly adjust that as well. :relaxed: :smiley:

libmproxy/protocol/tls.py
dweinstein: I took care of this with 8f8c2efccd52f9791cc5990f9863cdd02617bc0d

mhils:
Thanks, @dweinstein! :+1: :tada: :cake:

Issue #630, opened 2015-06-12
Clarify procedure for setting a local transparent proxy


justincy:
The docs are pretty good but they don't point out that setting up a transparent proxy on the same machine where the traffic originates is a bit different as [noted here](http://askubuntu.com/a/281645/91159). More specifically, the iptables commands are different.

```
sudo iptables -t nat -A OUTPUT -p tcp -m owner ! --uid-owner root --dport 443 -j REDIRECT --to-port 8080
sudo iptables -t nat -A OUTPUT -p tcp -m owner ! --uid-owner root --dport 80 -j REDIRECT --to-port 8080
```

You have to run mitmproxy as a different user than the traffic you're trying to capture, otherwise that won't work. If you didn't have an expection for mitmproxy traffic then you'd end up with a loop (I ran out of open file descriptors when I did that). I'm not familiar enough with iptables to know whether there's any other match expression that could be used other than the process owner.

cortesi:
Yes, I'd definitely support a patch that points out that both pf and iptables can filter based on UID for same-host setups.

fxfactorial:
I just tried the iptables commands listed and had mitmproxy running as another user but I still wasn't able to get traffic through. curl on anything plain http would end with "transfer closed with xx bytes remaining to read". This is on Ubuntu 14.04 in a VM.

PR #894, opened 2016-01-25
Update tunneling docs

fxfactorial:
Tested on OS X machine running Ubuntu 14.04 over VMware, both 80/443 ports were correctly tunneled. 

Great tool, thank you for mitmproxy.

### DIFF
--- a/docs/tutorials/transparent-dhcp.rst
+++ b/docs/tutorials/transparent-dhcp.rst
@@ -70,12 +70,15 @@ DHCP and TFTP) services to a small-scale network.
 3. Redirect traffic to mitmproxy
 ------------------------------------------
 
-To redirect traffic to mitmproxy, we need to add two iptables rules:
+To redirect traffic to mitmproxy, we need to add two iptables
+rules. Note the ``mitm_account``, this is the account that you will
+run ``mitmproxy`` with, hence the account doing these iptables must be
+different than ``mitm_account``.
 
 .. code-block:: none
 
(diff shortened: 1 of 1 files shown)


mhils:
Thanks. Is this a bug in the VM tutorial or is this PR just intended as a fix for #630?

fxfactorial:
Isn't it both? The tutorial was wrong, which was referenced in https://github.com/mitmproxy/mitmproxy/issues/630 I guess this kills two birds with one stone.

mhils:
The tutorial works fine for me - it is not intended to proxy traffic from the same machine. I think it makes more sense to add the "proxy your own machine" part to the transparent proxy linux page.
Thoughts?

fxfactorial:
I agree. What do you think ought to be added, exactly?

fxfactorial:
@mhils Ah, I misread you. You're suggesting an entirely separate proxy your own machine section, correct?

mhils:
Sorry for the super slow reply this week - yes, exactly. a short "you can proxy your own machine with these iptables rules" section in the linux transparent mode docs would be best I think, unless you have a better suggestion of course. :smiley:

mhils:
Any update on this? :smiley:

fxfactorial:
Sorry, I definitely will get to this soon, got a little confused by yet another markdown language to fudge around.

mhils:
Let me know once you are ready here - we'll close this for now. :smiley:

PR #898, opened 2016-01-26
[console] always show connection info in Details tab

### LINKED ISSUE #886: Client Address is not displayed for flows from dumpfiles
I have two clients connecting to mitmproxy now, and I'd like to see the client IP for each request in the "Detail" tab.

Maybe show the server's IP too, since it can differ when connections come from different local IPs (e.g. 127.0.0.1 and 192.168.1.101).

lilydjwg:
Connections become false values when it's `finished`. Check if it's
`None` when trying to show address info.

Fixes #886.


mhils:
:+1: :tada: :cake:

Issue #802, opened 2015-10-15
mitmweb - possibility to add username / password for web interface


fliespl:
Hi everyone,
I am not sure if it's possible or not. I went through docs and couldn't find it. Is there an easy way to protect web port access (not proxy).

mhils:
Not yet, contributions are welcome!

ganguera:
We can close this issue as this feature is included in 0.16.

joar:
This feature seems to have been removed in mitmproxy 3.0.3.

Kriechi:
This feature wasn't available since at least v1.0
https://github.com/mitmproxy/mitmproxy/commit/9b08279c7c3384f716b66329fefbe97a368189a2
https://github.com/mitmproxy/mitmproxy/commit/40f0193dda992960b9836482425cd45329f240af

PR #900, opened 2016-01-26
Added Basic Auth support for mitmweb interface

ganguera:
Added Basic Auth support for mitmweb interface as requested in #802. It supports both single user format ("username:password") or .htpasswd file.


libmproxy/web/app.py
```diff
@@ -240,10 +277,14 @@ def put(self):
...
-    def __init__(self, master, debug):
+    def __init__(self, master, debug, singleuser, htpasswd):
         self.master = master
+        self.additional_args = dict(
+            singleuser=singleuser,
+            htpasswd=htpasswd,
+        )
         handlers = [
-            (r"/", IndexHandler),
+            (r"/", IndexHandler, self.additional_args),
```
mhils: So, this looks like it only implements basic auth for the index page, but not the API?

libmproxy/web/app.py
```diff
@@ -25,6 +27,41 @@ def set_default_headers(self):
...
+                        self.set_auth_headers()
+                        raise APIError(401, "Invalid username or password.")
+                elif self.htpasswd:
+                    try:
+                        password_manager = authentication.PassManHtpasswd(self.htpasswd)
+                    except ValueError as v:
+                        raise APIError(500, "%s. Please contact your systems administrator." % v.message)
+                    if not password_manager.test(self.username, self.password):
+                        self.set_auth_headers()
+                        raise APIError(401, "Invalid username or password.")
```
mhils: This looks like there's quite a bit of duplication here. It'd be useful if we could just pass a password_manager to whatever authentication thingie we're using. The `password_manager` could be initialized when we parse the command line parameters.

mhils:
Any progress on this?

ganguera:
Hi @mhils, sorry for the delay. Tonight I'll pull request the refactor.

libmproxy/web/app.py
ganguera: You're right. I need to think about it.

mhils:
This still looks like it only protects the index page?

ganguera:
You're right @mhils , so far this only protects the index page. I think it's easy to enable the basic auth for the rest of the URLs except "/updates", that are being processed with the "tornado.websocket.WebSocketHandler". I need to check whether is possible to enable basic auth using websockets.

ganguera:
Hi @mhils !

As requested, now I'm parsing the specific web options in the web module. I think I also find a way to authorize the usage of websockets by setting / getting a secure cookie.

Looking forward to hearing from you.

mhils:
Thanks, this looks better now. Let me know once we have authentication for the WebSocket channel :)

ganguera:
Sorry guys, I need help to merge this...

Issue #714, opened 2015-08-06
flow.intercept() isn't working in inline scripts


ojab:
Not sure if it's a bug or desired behavior, but it looks like an issue for me. Simple script like
```python
def request(context, flow):
    flow.intercept(context._master)
```
correctly sets "'intercepted': True", but flow is actually accepted by mitmproxy (I assume that if interception pattern is not set, intercepted is set to False somewhere).

mhils:
This is more or less by design currently - you can only intercept flows programatically by setting an interception pattern. I realize there's a use case, so if you're interested, feel free to give it a spin (you'd need to suppress the `.reply` somehow) and send a PR! 

- You'd need to suppress `.reply()`, which gives control back to the main channel: https://github.com/mitmproxy/mitmproxy/blob/adf2877020c563fa9dc9a4df26bfd8b243e8e41a/libmproxy/console/__init__.py#L703
- Related: https://mitmproxy.org/doc/dev/architecture.html

dufferzafar:
I guess this can be closed too.

PR #906, opened 2016-01-30
Added feature to intercept flows from within inline scripts

anantk17:
Adds feature requested in issue #714


mhils:
Thanks. Could you please add a quick comment that explains why we check for `if f.intercepted:` here?

anantk17:
Calling `flow.intercept()` from within an inline script sets `f.intercepted : True`. Checking for `f.intercepted` allows us to suppress the call to `flow.reply()` and intercept the flow successfully.

mhils:
Thanks! :tada: :smiley:

PR #909, opened 2016-02-02
fix crash with viewing unpacked date objects

chrishamant:
Was getting a crash viewing AMF requests... tracked it down to inflated python date objects that can't deserialize directly to JSON. This worked for me - might be a better way to do this...


mhils:
Good catch - thanks! :smiley: :cake: :tada:

PR #914, opened 2016-02-04
Modern setuptools

mhils:
This PR simplifies our setup.py to use enviroment markers to specify dependencies. This allows us to build a wheel for mitmproxy, which should improve installation speed. Also, this should make requires.io work with mitmproxy. The downside is that we don't support old setuptools versions anymore, but to my understanding, we will have this "issue" with `h2` as a dependency anyway. Ubuntu 14.04 LTS+ onwards should work.


Kriechi:
LGTM! :shipit:

PR #916, opened 2016-02-05
Export any flow request as Python code

dufferzafar:
#884 added support for copying GET requests only.

This PR removes that restriction.

Generic code idea taken from [Postman](https://www.getpostman.com/), which has support for exporting a request as code in many different languages (C#, Ruby, etc.)

I tested the code with these requests: https://gist.github.com/096254f2dbece1d221ab


mhils:
Thanks. Now that this is becoming more complex, it would be nice if we could move it out of the console part and add tests for it. Thoughts?

dufferzafar:
Yeah, I felt the need of having some tests for this too. I'll look into that. 

About moving it from the console part, I am not really familiar with the codebase yet. So can you please recommend which file would make the most sense for this code?

mhils:
I think moving it into utils is just fine, if that works for you? :smiley:

test/test_export_flow.py
```diff
@@ -0,0 +1,105 @@
...
+              """headers = {\n"""
+              """    'header': 'qvalue',\n"""
+              """    'content-length': '7',\n"""
+              """}\n\n"""
+              """response = requests.request(\n"""
+              """    method='GET',\n"""
+              """    url=url,\n"""
+              """    headers=headers,\n"""
+              """)\n\n"""
+              """print(response.text)""")
```
Kriechi: Maybe using something like this makes it more readable:
```
from textwrap import dedent

...

print dedent("""
    import requests

    url = 'http://address/path'
    foo
      bar
""").strip()
```

mhils: (You can also just use single quotes in every line) :smiley:

libmproxy/flow_export.py
```diff
@@ -0,0 +1,70 @@
...
+    code = """import requests
+
+url = '{url}'
+{headers}{params}{data}
+response = requests.request(
+    method='{method}',
+    url=url,{args}
+)
+
+print(response.text)"""
```
Kriechi: Can you use [textwrap.dedent](https://docs.python.org/2/library/textwrap.html#textwrap.dedent) here to make it more readable?

Kriechi:
LGTM! @mhils?

mhils:
:+1:  Thanks! :smiley:

PR #917, opened 2016-02-05
remove compatible version specifiers

mhils:
This PR removes the compatible version specifiers again. This allows us to `pip install mitmproxy` on Ubuntu 14.04 out of the box. Also, I added a note for Ubuntu 12.04 that users need to update pip first.

FWIW, Ubuntu 12.04 is currently responsible for 2.5% of our PyPi installs (thx @dstufft!):

`null` means PyPi doesn't have that data available because of the version of the installer they used didn't include it in the information they send.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,55 +50,55 @@ setup(
     # https://packaging.python.org/en/latest/requirements/#install-requires
     # It is not considered best practice to use install_requires to pin dependencies to specific versions.
     install_requires=[
-        "netlib~={}".format(version.VERSION),
-        "h2~=2.1.0",
-        "tornado~=4.3.0",
-        "configargparse~=0.10.0",
-        "pyperclip~=1.5.22",
-        "blinker~=1.4.0",
-        "pyparsing~=2.0.5",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -50,55 +50,55 @@
...
-        "blinker~=1.4.0",
-        "pyparsing~=2.0.5",
-        "html2text~=2016.1.8",
-        "construct~=2.5.2",
-        "six~=1.10.0",
-        "Pillow~=3.0.0",
-        "watchdog~=0.8.3",
-        "click~=6.2",
-        "urwid~=1.3.1",
+        "netlib>={}, <{}".format(version.MINORVERSION, version.NEXT_MINORVERSION),
```
Kriechi: do we actually want to do that? I thought mitmproxy/pathod/netlib share the same version number at all times?

mhils: This enforces the same minor version, but allows us to push 0.15.x bugfix release for a single project - e.g. netlib is currently at 0.15.1, whereas mitmproxy is at 0.15.

PR #929, opened 2016-02-10
Added Basic Auth support for MITMWeb interface

ganguera:
Added Basic Auth support for MITMWeb interface as requested in #802. It supports both single user format ("username:password") or .htpasswd file.

### DIFF
--- a/libmproxy/web/app.py
+++ b/libmproxy/web/app.py
@@ -48,21 +48,18 @@ class BasicAuth(object):
         self._transforms = []
         self.finish()
 
-    def initialize(self, **kwargs):
-        self.wauthenticator = kwargs.get("wauthenticator")
-
     def prepare(self):
-        if self.wauthenticator:
+        wauthenticator = self.application.settings['wauthenticator']
+        if wauthenticator:
             auth_header = self.request.headers.get('Authorization')
             if auth_header is None or not auth_header.startswith('Basic '):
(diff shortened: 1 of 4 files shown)


libmproxy/web/app.py
```diff
@@ -284,8 +309,11 @@ def put(self):
 
 class Application(tornado.web.Application):
 
-    def __init__(self, master, debug):
+    def __init__(self, master, debug, wauthenticator):
         self.master = master
+        self.additional_args = dict(
```
mhils: Do we have to set this here?

libmproxy/web/app.py
```diff
@@ -40,7 +41,31 @@ class APIError(tornado.web.HTTPError):
     pass
 
 
-class RequestHandler(tornado.web.RequestHandler):
+class BasicAuth(object):
```
mhils: This isolation looks good :+1:

libmproxy/web/app.py
mhils: Can we use http://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.settings instead of passing that manually to each RequestHandler?

mhils:
Excellent - thanks! :smiley: :cake: :tada:

ganguera:
You're welcome!

PR #933, opened 2016-02-14
Print HTTP2 frames to debug log

mhils:
This may become handy when getting bug reports. Also, it may be useful to peek into internals for some users.

### DIFF
--- a/libmproxy/protocol/http2.py
+++ b/libmproxy/protocol/http2.py
@@ -131,6 +131,12 @@ class Http2Layer(Layer):
         raise NotImplementedError()
 
     def _handle_event(self, event, source_conn, other_conn, is_server):
+        self.log(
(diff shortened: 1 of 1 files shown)


libmproxy/protocol/http2.py
```diff
@@ -131,6 +131,12 @@ def next_layer(self):  # pragma: no cover
         raise NotImplementedError()
 
     def _handle_event(self, event, source_conn, other_conn, is_server):
+        self.log(
+            "HTTP2 Frame from {}".format("server" if is_server else "client"),
```
Kriechi: This function receives events, which are NOT HTTP/2 frames.
A single frame could trigger multiple events.
Do you want to log frames or events?

There are probably (too?) many frames to log - and without logging sent frames as well, they usually don't contain much information.

PR #935, opened 2016-02-14
Improve Coding Style

Kriechi:
Using autopep8 and autoflake tools.
Only minor changes.

### DIFF
--- a/test/test_filt.py
+++ b/test/test_filt.py
@@ -209,7 +209,7 @@ class TestMatching:
 
     def test_method(self):
         q = self.req()
-        s = self.resp()
+        self.resp()
         assert self.q("~m get", q)
         assert not self.q("~m post", q)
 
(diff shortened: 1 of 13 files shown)


test/test_filt.py
```diff
@@ -209,7 +209,7 @@ def test_body(self):
 
     def test_method(self):
         q = self.req()
-        s = self.resp()
+        self.resp()
```
mhils: remove entirely?
(same below)

Kriechi: `s` is not used - that's the reason why autopep8 decided to remove it.
Do you want to keep it? For what reason?

mhils: `self.resp()` just creates a response object without side-effects. There is no point for keeping the whole line. (I suppose someone originally just copied `q=...; s=...` to each test)

PR #937, opened 2016-02-15
Combine mitmproxy, pathod and netlib in a single repo.

mhils:
This is an experiment to combine our projects into a single repo. I hope that this brings benefits on a variety of fronts:

- No more syncing issues between repositories.
   - No more PRs that are split over three repos.
   - No more release builds that depend on race conditions
   - No more repo syncing when switching branches
- Test runs for all projects when we change parts in netlib.
- A more straightforward project structure - the `dev` script is much more intuitive
- Shared metadata. We had some parts go out of sync repeatedly (.landscape.yml, .env, ...)

Drawbacks:
- We will only have a single coverage percentage shared by all projects.
- This breaks requires.io - I talked to them in January and they promised a fix for this anyway in February.

I moved mitmproxy into `./mitmproxy`, netlib into `./netlib`, pathod into `./pathod`, release into `./release` and then started moving the metadata files that can be unified one layer up. The best way to inspect how things are looking is browsing through the repo at https://github.com/mhils/mitmproxy/tree/single-repo

TODO:
- [x] Adjust Travis
- [x] Adjust AppVeyor
- [ ] Adjust release script
- [x] Figure out how coverage collection should work
- [ ] Figure out what to do with .dockerignore (mitmproxy)
- [ ] Figure out what to do with .jsbeautifyrc (pathod)
- [x] <s>Figure out what to do with</s> Remove check_coding_style.sh (netlib)
- [x] <s>Maybe move CONTRIBUTORS file one level up again? (`git shortlog -n -s -- mitmproxy` works)</s>
- [ ] Figure out how to include LICENSE/CONTRIBUTING.md into source packages - maybe have one copy per repo again?
- [ ] Figure out what to do with old tags/branches - just discard them?
- [x] Write central README
- [ ] Check if contributor count on GitHub actually updates.
- [X] Rename `dev` to `dev.sh`
- [ ] Mark netlib and pathod repos as outdated.
- [ ] Migrate issues?
- [ ] Fix pathod tests

Ideas that we may want to implement after this has landed:
 - Move all test folders together.


mhils:
> Figure out how to include LICENSE/CONTRIBUTING.md/CONTRIBUTORS into source packages - maybe have one copy per repo again?

The problem here is that we cannot specify files beyond the top-level directory in MANIFEST.in (which determines the files included in our source distribution). I see three ways to solve this:

1. Keep them in the repo root and let the release script copy and delete them
2. Keep them in the repo root  and symlink them into subdirectories. Does not work on Windows, but should not affect Windows developers (except for building releases)
3. Keep regular copies.

I currently think that approach 1 is probably best, but I'm open for suggestions.

PR #948, opened 2016-02-16
Preserve host header when modifying request path

fimad:
Currently the path_components and query setters of the Request object
use the url setter under the hood. The url setter updates all parts of
the URL including the host. If the host header and the host in the
request URL are different (as is common when making HTTPS requests)
then the host header will be updated to the value in the URL as a
result of modifying the path.

This change fixes this problem by modifying the query and
path_components setters to not use the url setter and instead directly
update the path field.


mhils:
Thanks! :smiley: :cake:

Issue #947, opened 2016-02-15
Finish repository merge


mhils:
From https://github.com/mitmproxy/mitmproxy/pull/937:

- [x] Mark netlib and pathod repos as outdated.
- [x] Migrate issues
- [x] Adjust release script
- [x] Figure out how to include LICENSE/CONTRIBUTING.md into source packages - maybe have one copy per repo again?
- [x] Figure out what to do with Dockerfile and .dockerignore (mitmproxy)
- [x] Figure out what to do with .jsbeautifyrc (pathod)
- [x] Figure out what to do with old tags/branches
- [ ] Fix pathod tests

mhils:
@dweinstein, what do we need to do with .dockerignore? Just move it to the repo root? :-)

mhils:
@cortesi, .jsbeautifyrc stays where it is?

PR #950, opened 2016-02-16
Rename lib folders

Kriechi:
refs #947


pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+testpaths = test
```
mhils: Do we still need to specify `./test` for the runs then?

mhils:
As I mentioned before, I like this change. It seems much more intuitive that mitmproxy provides a package that is actually called mitmproxy, and the same applies to pathod. Technically, both are tools, not libraries (although we obviously want to design the nicely nonetheless).

pytest.ini
Kriechi: No, we do not. But since you use this fancy trick to get netlib to test on py3 I'm not sure how to handle it...

mhils: Ok, that is excellent. So we can just keep it for travis but remove it everywhere else?

cortesi: Not if the py.test command is being run from the top of the repo.

cortesi:
This looks good to me. Way back, the reason for the "libmproxy" naming was to distinguish between mitmproxy the tool and mitmproxy the library. Things have evolved over time and this makes less sense nowadays.

PR #952, opened 2016-02-17
Add a content view for query parameters

fimad:
The query content view uses format_dict to display a table of query
parameters and is made the default content view for requests with
query parameters and no request body.

To facilitate this the query parameter dictionary of HTTPRequests is
added to the metadata content view parameter under the "query" key.

Additionally, the logic for handling "no content" messages is moved
from contentviews.get_content_view to ViewAuto. This is necessary as
it allows the query content view to be displayed when there is no
request body.


cortesi:
Looks good, thanks!

PR #960, opened 2016-02-17
enable HTTP/2 by default if available

Kriechi:


mitmproxy/mitmproxy/proxy/config.py
```diff
@@ -181,7 +182,9 @@ def process_proxy_options(parser, options):
         certs.append(parts)
 
     if options.http2 and not tcp.HAS_ALPN:
-        raise RuntimeError("HTTP2 support requires OpenSSL 1.0.2 or above.")
```
mhils: How does that play out with mitmproxy? Users won't be able to see this, will they?

Kriechi: Once they exit mitmproxy the users will see the message. Yes this is a bit counter-intuitive, but the easiest solution.

Issue #651, opened 2015-06-24
problem with port in host


Nerudifus:
RFC 7230 (HTTP 1.1) declares in 5.4 Host that it is possible to add the port to the Host header field. For example ```Host = this.is.some.example.host:8080```. This is especially true if not the default port is used.
This is not properly handled by mitmdump. In my first, pre 0.12.1, tests with HTTPS mitmdump generates a bad gateway as response. In my current test with HTTP and port 8080 it seems to work but the output of mitmdump is still incorrect. It uses this.is.some.example.host:8080:8080 in its output.

I am not sure if this is only a display error or if mitmdump internally use the duplicated port and that the old HTTPS server could not handle it but the current HTTP could.

mhils:
Thanks for the report. This probably needs to be fixed in `HTTPRequest.pretty_host`. I'd be happy to accept a patch!

MatthewShao:
Hi there!
I want to have a try on this issue, could you guys please tell me how to reproduce the error?

dufferzafar:
I am having a hard time reproducing this too!

I ran `mitmdump --host`, and then ran this command to add a request to mitmdump, but I don't see repeated port numbers. 

```bash
printf "GET http://example.com HTTP/1.1\r\nhost: this.is.some.example.host:8090" | nc localhost 8080 >/dev/null
```

mhils:
You can reproduce it with `GET http://example.com:8084/ HTTP/1.1\r\nHost: example.org:8090`.

Cheers,
Max

dufferzafar:
@mhils Thanks. 

This is being caused because [`netlib.http.request.pretty_host`](https://github.com/mitmproxy/netlib/blob/master/netlib/http/request.py#L162-L169) just returns the `host` header as it is.

So, this should be fixed in netlib (rather than mitmproxy's `HTTPRequest.pretty_host`) right?

mhils:
Yes, please.

https://github.com/mitmproxy/mitmproxy/blob/master/examples/dns_spoofing.py#L29 is also relevant here.

dufferzafar:
@mhils I am a bit confused :/

What should actually happen in the case of two port numbers (like the one you mentioned: `GET http://example.com:8084/ HTTP/1.1\r\nHost: example.org:8090`)

Which port does the actual connection opens to?

Nerudifus:
[...]

PR #961, opened 2016-02-18
Handle port numbers in host header

dufferzafar:
From: https://github.com/mitmproxy/netlib/pull/121

Also incorporates the [comment](https://github.com/mitmproxy/mitmproxy/issues/651#issuecomment-184007109) by @mhils

### DIFF
--- a/netlib/netlib/http/request.py
+++ b/netlib/netlib/http/request.py
@@ -183,7 +183,12 @@ class Request(Message):
         This is useful in transparent mode where :py:attr:`host` is only an IP address,
         but may not reflect the actual destination as the Host header could be spoofed.
         """
-        return self._parse_host_header()[0] or self.host
+        host, port = self._parse_host_header()
+        if not host:
+            return self.host
+        if not port:
+            port = 443 if self.scheme == 'https' else 80
+        return host if port == self.port else self.host
 
     @property
     def pretty_url(self):
--- a/netlib/netlib/http/request.py
+++ b/netlib/netlib/http/request.py
@@ -188,6 +188,7 @@ class Request(Message):
             return self.host
         if not port:
             port = 443 if self.scheme == 'https' else 80
+        # Prefer the original address if host header has an unexpected form
         return host if port == self.port else self.host
 
     @property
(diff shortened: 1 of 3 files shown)


netlib/netlib/http/request.py
```diff
@@ -166,16 +183,24 @@ def pretty_host(self):
...
     @property
     def pretty_url(self):
         """
         Like :py:attr:`url`, but using :py:attr:`pretty_host` instead of :py:attr:`host`.
         """
+        host, port = self._parse_host_header()
+        host = host or self.host
+        port = port or self.port
         if self.first_line_format == "authority":
-            return "%s:%d" % (self.pretty_host, self.port)
```
mhils: Why can't we continue using `pretty_host` here?

netlib/netlib/http/request.py
```diff
@@ -166,16 +183,24 @@ def pretty_host(self):
         This is useful in transparent mode where :py:attr:`host` is only an IP address,
         but may not reflect the actual destination as the Host header could be spoofed.
         """
-        return self.headers.get("host", self.host)
+        host, port = self._parse_host_header()
+        if not host:
+            return self.host
+        if not port:
+            port = 443 if self.scheme == 'https' else 80
+        return host if port == self.port else self.host
```
mhils: Looks good. Can you add a comment that explains why we do this check here?

Kriechi:
Does this affect this part in any way? Or should it?
https://github.com/mitmproxy/mitmproxy/blob/be02dd105b7803b7b2b6942f9d254539dfd6ba26/mitmproxy/mitmproxy/protocol/http2.py#L297-L320

mhils:
I don't think so. There's no host header in HTTP2, so this should not affect that part. Looking at the code though, `authority.partition(':')` may fail for IPv6 (see how we parse the header up here), but that is an unrelated issue.

netlib/netlib/http/request.py
dufferzafar: This change was made *before* I added new checks to `pretty_host`. I've fixed it!

mhils:
Tests are failing :(

mhils:
Thanks! :smiley:

PR #969, opened 2016-02-21
Indent JSON body while exporting it as code

### LINKED ISSUE #968: Indent JSON data while exporting it as Python code
I was testing out a web API and used the "Export flow as Python code" feature for the first time as user, and noticed an improvement.

Currently we just export the `flow.request.body` as is (independent of it's content type) but mitmproxy's interface is smart and renders different bodies differently (for eg. it indents JSON)

I think we could add this indent behaviour while exporting things as code too.

dufferzafar:
Fixes #968

### DIFF
--- a/mitmproxy/flow_export.py
+++ b/mitmproxy/flow_export.py
@@ -1,8 +1,9 @@
+import json
 import urllib
 from textwrap import dedent
 
 import netlib.http
-from . import contentviews
+from netlib.utils import parse_content_type
 
 
 def curl_command(flow):
@@ -55,14 +56,9 @@ def python_code(flow):
 
     data = ""
     if flow.request.body:
-        cv = contentviews.get_content_view(
-            viewmode=contentviews.get("Auto"),
-            data=flow.request.body,
-            headers=flow.request.headers,
-        )
-
-        if cv[0] == "JSON":
-            data = "\njson = %s\n" % "\n".join(l[0][1] for l in cv[1])
+        if is_json(flow.request.headers):
+            data = json.dumps(json.loads(flow.request.body), indent=4)
+            data = "\njson = %s\n" % data
             args += "\n    json=json,"
         else:
             data = "\ndata = '''%s'''\n" % flow.request.body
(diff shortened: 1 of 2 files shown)


mitmproxy/flow_export.py
```diff
@@ -1,7 +1,25 @@
...
+        return data
+
+    cv = contentviews.get_content_view(
+        contentviews.get("Auto"),
+        data,
+        headers=headers,
+    )
+
+    if cv[0] == "JSON":
+        return "\n".join(l[0][1] for l in cv[1])
```
dufferzafar: Currently, I am just hardcoding this for JSON only as I couldn't figure out a way to make it work for all the cases. Help?

mhils:
The problem with the content views is that they often do more than just adjust whitespace, e.g., decode protobufs. Maybe we just want to consider the JSON case here and use `requests.post(url, json=payload)` ?

dufferzafar:
Well to be honest, I'd be happy with only the JSON case too, but do we really need to use the `json` parameter in `requests.post` ? Since we already have valid headers etc. we could make do with the `data` param too, right?

mhils:
My understanding is that you can pass a dict/list/json thing into `json=`, whereas you pass a string otherwise?

dufferzafar:
@mhils I guess I wasn't quite familiar with the `json` parameter, I 've updated the code to use it now!

mhils:
Hm.. Maybe we just want to check the content type and then use `json.dumps` here? This looks like a weird "let's make a content view and then unfiddle the content view" dance here. The less cross-dependencies we have, the better :smiley:

dufferzafar:
@mhils I've updated the code to use `json.dumps` and have also written a test. Please review :)

mitmproxy/flow_export.py
```diff
@@ -53,8 +56,16 @@ def python_code(flow):
 
     data = ""
     if flow.request.body:
-        data = "\ndata = '''%s'''\n" % flow.request.body
-        args += "\n    data=data,"
+        json_obj = is_json(flow.request.headers, flow.request.body)
+        if json_obj:
+            # Without the separators field json.dumps() produces
+            # trailing white spaces: https://bugs.python.org/issue16333
```
Kriechi: Is this still a thing? As far I see it from quickly browsing the bug tracker, this should be already be fixed a long time ago...?

dufferzafar: The 'fix' for that issue was to update the documentation that suggests using `separators=(',', ': '))` - https://hg.python.org/cpython/rev/78bad589f205

Kriechi:
Thanks! :tada:

Issue #972, opened 2016-02-23
Error in setting dev environment


anchitjain1234:
##### Steps to reproduce the problem:

1. Did steps as described in ``README``, cloned the project and then used ``./dev.sh``. When ``dev.sh`` script executes it gives error ``TypeError: find_packages() got an unexpected keyword argument 'include'``

##### What is the expected behavior?
Virtualenv should set up.

##### What went wrong?
Unexpected error while installing ``TypeError: find_packages() got an unexpected keyword argument 'include'``

##### Any other comments?
Full error log
```
New python executable in ./venv/bin/python
Installing setuptools, pip...done.
Obtaining file:///home/ubuntu/workspace/mitmproxy (from -r requirements.txt (line 1))
  Running setup.py (path:/home/ubuntu/workspace/mitmproxy/setup.py) egg_info for package from file:///home/ubuntu/workspace/mitmproxy
    Traceback (most recent call last):
      File "<string>", line 17, in <module>
      File "/home/ubuntu/workspace/mitmproxy/setup.py", line 47, in <module>
        "netlib", "netlib.*"
    TypeError: find_packages() got an unexpected keyword argument 'include'
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):

  File "<string>", line 17, in <module>

  File "/home/ubuntu/workspace/mitmproxy/setup.py", line 47, in <module>

    "netlib", "netlib.*"

TypeError: find_packages() got an unexpected keyword argument 'include'

----------------------------------------
Cleaning up...
Command python setup.py egg_info failed with error code 1 in /home/ubuntu/workspace/mitmproxy
Storing debug log for failure in /home/ubuntu/.pip/pip.log
```

---

Mitmproxy Version:
Operating System: Ubuntu 14.04

Kriechi:
@anchitjain1234 thanks for the nice bug report - a single search on the internet points to an outdated pip/setuptools.
See https://stackoverflow.com/questions/29477456/find-package-errors-during-installing-package-via-pip

anchitjain1234:
@Kriechi Sorry if I sounded silly. I already have uptodate setuptools and pip but than also I am getting same error. Thats why I opened this issue.

Kriechi:
Can you show us your pip version? Run `pip -V` and `python -V`

anchitjain1234:
Pip version ``pip 1.5.4 from /usr/lib/python2.7/dist-packages (python 2.7)``
[...]

PR #974, opened 2016-02-23
add pip version check

### LINKED ISSUE #973: Check pip version before setting up dev environment
Currently we require a rather new version of pip and setuptools - which is not available on old Ubuntu releases. We should add a check in the `dev.sh` script to display an error message if the pip version is outdated.

ref #972

ParthGanatra:
Fixes #973

Kept mimimum version required to be **6**.

And error message:

Error. Outdated pip.
To upgrade pip run: `pip install -U pip`

### DIFF
--- a/dev.sh
+++ b/dev.sh
@@ -4,13 +4,7 @@ VENV=./venv
 
 python -m virtualenv $VENV --always-copy
 . $VENV/bin/activate
-pip_version=$(pip -V | cut -d' ' -f 2 | cut -d'.' -f 1)
-if [ "$pip_version" -lt 6 ];
-then
-	echo "Error. Outdated pip"
-	echo "To upgrade pip run: \`pip install -U pip\`r"
-	exit 1
-fi
+pip install -U pip
 pip install -r requirements.txt
 
 echo ""
--- a/dev.sh
+++ b/dev.sh
@@ -4,6 +4,13 @@ VENV=./venv
 
 python -m virtualenv $VENV --always-copy
 . $VENV/bin/activate
+pip_version=$(pip -V | cut -d' ' -f 2 | cut -d'.' -f 1)
+if [ "$pip_version" -lt 6 ];
+then
+	echo "Error. Outdated pip"
+	echo "To upgrade pip run: \`pip install -U pip\`r"
+	exit 1
+fi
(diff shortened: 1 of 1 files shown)


dev.sh
```diff
@@ -4,6 +4,13 @@ VENV=./venv
 
 python -m virtualenv $VENV --always-copy
 . $VENV/bin/activate
+pip_version=$(pip -V | cut -d' ' -f 2 | cut -d'.' -f 1)
+if [ "$pip_version" -lt 6 ];
+then
+	echo "Error. Outdated pip"
+	echo "To upgrade pip run: \`pip install -U pip\`r"
```
Kriechi: What is the `r` for at the end of this line?

dev.sh
```diff
@@ -4,6 +4,13 @@ VENV=./venv
 
 python -m virtualenv $VENV --always-copy
 . $VENV/bin/activate
+pip_version=$(pip -V | cut -d' ' -f 2 | cut -d'.' -f 1)
+if [ "$pip_version" -lt 6 ];
```
Kriechi: Did you check if `6.0` is actually enough?
Maybe add a comment that mentions the reason (because Ubuntu 12.04 and 14.04 ship something older)

mhils:
I think we probably just want to run pip install -U pip? in the virtualenv?

dev.sh
mhils: 3.x is enough, see my comment in #972. Nonetheless, let's just upgrade to the latest thing available.

dufferzafar:
Yeah, since the only way to 'fix' things is to update pip. Why not just do it unconditionally.

dev.sh
ParthGanatra: Sorry that was a typo.

dev.sh
ParthGanatra: I checked it for version 6.0 it worked.
screenshot-1
Only downgrade option available less than 6.0 was 1.5.6, it fails for this version.
Anyways upgrading pip unconditionally in dev script.

Kriechi:
fixed by 1b7b96852632ed9e89c6d3dc1ae54248f683ac2f
Thanks!

PR #978, opened 2016-02-24
use dev-scripts on CI

Kriechi:
This should help us keep the install routine clean and tested on all platforms.


dev.sh
```diff
@@ -2,6 +2,8 @@
 set -e
 VENV=./venv
 
+pip install -U virtualenv || sudo pip install -U virtualenv
```
mhils: I don't like that we upgrade system Python things here. Why is that necessary?

Kriechi: upgrade is not necessary - but why not?
install is necessary - travis does not ship with virtualenv.

Kriechi:
@mhils please take another look, I fixed the things you mentioned.

PR #980, opened 2016-02-26
[web] Fix layout for container and main-view

gzzhanghao:
This PR aims to fix a dislocation in main-view:

image

The following steps can use to reproduce the bug (verified in `Chrome 47.0.2526.106` on `Ubuntu 14.04.4`):

- Open the flow detail panel
- Scroll the flow table

### DIFF
--- a/mitmproxy/web/static/app.css
+++ b/mitmproxy/web/static/app.css
@@ -1,11 +1,3 @@
-html {
-  box-sizing: border-box;
-}
-*,
-*:before,
-*:after {
-  box-sizing: inherit;
-}
 .resource-icon {
   width: 32px;
   height: 32px;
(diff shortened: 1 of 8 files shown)


mhils:
Thanks for the PR! 
Things work just fine for me (Chrome 48.0.2564.116 and Firefox 43, Win 10).
What's the motivation for changing the box-sizing? Why do you modify the event log flex?

gzzhanghao:
@mhils Things works again with the latest version of Chrome and Firefox, likely to be an issue for flex in specific version of Chrome.

And the event log, again, was broken on my browser and it works now :smile: 

As for the box-sizing, I found it confusing when I debugging the issue. Though it should not be included it here, I think.

Seems good to be closed.

mhils:
There are some good reasons for border-box: http://www.paulirish.com/2012/box-sizing-border-box-ftw/ - I personally like it a lot and I feel it makes dev easier.

Issue #644, opened 2015-06-23
Upstream proxy authentication


cortesi:
Add a command-line flag to specify the upstream proxy authentication credentials, which are then set on both HTTP and CONNECT requests (but not HTTPS requests). 

See discussion in #643 for the use case.

PR #988, opened 2016-03-01
Add upstream proxy authentication

xhy940801:
Finished #644

### DIFF
--- a/test/mitmproxy/test_cmdline.py
+++ b/test/mitmproxy/test_cmdline.py
@@ -56,8 +56,12 @@ def test_parse_server_spec():
 
 def test_parse_upstream_auth():
     tutils.raises("Invalid upstream auth specification", cmdline.parse_upstream_auth, "")
+    tutils.raises("Invalid upstream auth specification", cmdline.parse_upstream_auth, ":")
+    tutils.raises("Invalid upstream auth specification", cmdline.parse_upstream_auth, ":test")
     assert cmdline.parse_upstream_auth(
         "test:test") == "Basic" + " " + base64.b64encode("test:test")
+    assert cmdline.parse_upstream_auth(
+        "test:") == "Basic" + " " + base64.b64encode("test:")
 
 
 def test_parse_setheaders():
--- a/test/mitmproxy/test_cmdline.py
+++ b/test/mitmproxy/test_cmdline.py
@@ -1,4 +1,5 @@
 import argparse
+import base64
 from mitmproxy import cmdline
 from . import tutils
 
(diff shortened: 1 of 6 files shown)


mhils:
This looks good, thanks. Can you please add a test that verifies that the feature working? :smiley:

xhy940801:
I've added some test cases, but I'm not sure it's enough:disappointed_relieved:

test/mitmproxy/test_cmdline.py
```diff
@@ -53,6 +54,12 @@ def test_parse_server_spec():
         "http://")
 
 
+def test_parse_upstream_auth():
+    tutils.raises("Invalid upstream auth specification", cmdline.parse_upstream_auth, "")
+    assert cmdline.parse_upstream_auth(
+        "test:test") == "Basic" + " " + base64.b64encode("test:test")
```
Kriechi: I'm not sure what the RFC says, but are these allowed?
  * `test:` (empty password, maybe valid)
  * `:test` (empty username, probably invalid)
  * `:` (both empty - is it then a username by default, with empty password? or the other way around)

We should make sure to cover these cases: in the tests as well as the cli-parsing part.

Kriechi:
Thanks! :tada:

JakeAustwick:
I'm trying to use this right now, but hitting some issues.

```
mitmproxy -U http://xxxx.com:23333 --upstream-auth=xxx:yyy
```

Then when I try and use the proxy:
```
>>> proxies = {'http': 'http://127.0.0.1:8080'}
>>> url = 'http://geoip.nekudo.com/api'
>>> r = requests.get(url, proxies=proxies)
```

The mitmproxy console shows:
```
mitmproxy has crashed! call last):
Please lodge a bug report at: https://github.com/mitmproxy/mitmproxy, line 121, in handle
    root_layer()
  File "/Library/Python/2.7/site-packages/mitmproxy/proxy/modes/http_proxy.py", line 25, in __call__
    layer()
  File "/Library/Python/2.7/site-packages/mitmproxy/protocol/tls.py", line 358, in __call__
    layer()
  File "/Library/Python/2.7/site-packages/mitmproxy/protocol/http1.py", line 67, in __call__
    layer()
  File "/Library/Python/2.7/site-packages/mitmproxy/protocol/http.py", line 187, in __call__
    self.data.headers["Proxy-Authorization"] = self.config.upstream_auth
  File "/Library/Python/2.7/site-packages/mitmproxy/protocol/base.py", line 81, in __getattr__
    return getattr(self.ctx, name)
  File "/Library/Python/2.7/site-packages/mitmproxy/protocol/base.py", line 81, in __getattr__
    return getattr(self.ctx, name)
  File "/Library/Python/2.7/site-packages/mitmproxy/protocol/base.py", line 81, in __getattr__
    return getattr(self.ctx, name)
  File "/Library/Python/2.7/site-packages/mitmproxy/protocol/base.py", line 81, in __getattr__
    return getattr(self.ctx, name)
AttributeError: 'RootContext' object has no attribute 'data'
```

It's not the proxy because using the proxy directly with `requests` works. Any ideas?

mhils:
@JakeAustwick: fixed, thanks.

JakeAustwick:
Awesome. What's the easiest way for me to install this patch locally, so that I can test? I'm guessing it won't be pushed to PyPI for a while?

PR #989, opened 2016-03-01
Improve HAR Extractor Script

### LINKED ISSUE #987: har_extractor.py is broken
There are python problems with the har_extractor.py. It doesn't currently work on the stable release. Unfortunately I'm not good enough with python to provide a lot of additional details.

This was discussed in IRC, but I mostly wanted to file this so that others are aware of it's current state.

dufferzafar:
Fixes #987.

I've got the `har_extractor` script to run :smile:

I just ran `mitmdump -n -r small_dumpfile -s "examples/har_extractor.py /tmp/out.har"` and kept on fixing errors that came up but I am not too sure about the script right now!

There's room for a lot of improvements (maybe not in this same PR):

- [x] Clean up the code - remove duplication, make pythonic etc.

- [x] Write unit tests

~~- [ ] Test with real world stuff~~

@colemickens maybe you could test this out with your set of flows?

~~- [ ] Add some usage documentation~~

I had to look here and there to figure out that har_extractor works with mitmdump & not mitmproxy (is that really true though?)

### DIFF
--- a/examples/har_extractor.py
+++ b/examples/har_extractor.py
@@ -16,7 +16,7 @@ class _HARLog(HAR.log):
     __page_count__ = 0
     __page_ref__ = {}
 
-    def __init__(self, page_list):
+    def __init__(self, page_list=[]):
         self.__page_list__ = page_list
         self.__page_count__ = 0
         self.__page_ref__ = {}
@@ -66,7 +66,7 @@ def start(context, argv):
             '(- will output to stdout, filenames ending with .zhar '
             'will result in compressed har)'
         )
-    context.HARLog = _HARLog(['https://github.com'])
+    context.HARLog = _HARLog()
     context.seen_server = set()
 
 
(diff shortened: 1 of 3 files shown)


mhils:
Awesome - thanks for tackling this. Everything I see so far looks good to merge, let us know once you think stuff is ready for a final review :smiley:

dufferzafar:
@mhils @Kriechi I've added the tests for `har_extractor`. Please have a look :)

examples/har_extractor.py
```diff
@@ -239,6 +216,18 @@ def done(context):
...
+    return ""
+
+
+def format_headers(obj):
+    if obj:
+        return [{"name": k, "value": v} for k, v in obj.fields]
+    return ""
+
+
 def print_attributes(obj, filter_string=None, hide_privates=False):
```
dufferzafar: This function is is not used anywhere in the code. Should I remove it?

Kriechi: Yes - or could you add some code to show how to use it (example)?

dufferzafar: Um, I could come with an example, but it'll probably be something obtuse since I've never actually used this function. I guess it was added by the original author to quickly debug some objects.

Kriechi:
Thanks! :tada:

PR #990, opened 2016-03-02
[web] fix StoreView.index

gzzhanghao:
Use `_.sortedIndexBy` instead of `_.sortedIndex`, see https://lodash.com/docs#sortedIndex


mhils:
Apparently that was changed in [lodash 4.0.0](https://github.com/lodash/lodash/wiki/Changelog#v400) - thanks for the great find!

PR #991, opened 2016-03-02
[web] fix FlowTable.scrollIntoView

gzzhanghao:
`FlowTable.scrollIntoView` doesn't work correctly when there are multiple flows with same value for sorting function.

Reproduce by:

- select a flow whose method is `GET`
- sort flows by method field

### DIFF
--- a/mitmproxy/web/static/app.js
+++ b/mitmproxy/web/static/app.js
@@ -1580,7 +1580,7 @@ var FlowTable = _react2.default.createClass({
         this.forceUpdate();
     },
     scrollIntoView: function scrollIntoView(flow) {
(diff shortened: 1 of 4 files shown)


mhils:
Couldn't we fix that by using `_.sortedIndexOf` in `StoreView.index`?
(That would also require updating `StoreView.add` to manually use `_.sortedIndexBy`)

gzzhanghao:
@mhils We can also use `_.sortedIndexOf` in `FlowTable.scrollIntoView`. So the problem becomes what `StoreView.index` actually stands for :confused:

mhils:
Right - let's just rename that to indexOf then?

gzzhanghao:
@mhils Sounds good, pushed.

PR #992, opened 2016-03-02
[web] gulp: log errors from buildScript/bundle

gzzhanghao:
Gulp eats errors from watchify, which makes `scripts-app-dev` never finish when start up gulp with an error script.


mhils:
:+1: :smiley:

PR #994, opened 2016-03-02
[web] remove common.ChildFocus

gzzhanghao:
This PR eliminates `common.ChildFocus` mixin by extracting `returnFocus` method to `ProxyAppMain` component.

### DIFF
--- a/web/src/js/components/proxyapp.js
+++ b/web/src/js/components/proxyapp.js
@@ -55,6 +55,8 @@ var ProxyAppMain = React.createClass({
         };
     },
     focus: function () {
+        document.activeElement.blur();
+        window.getSelection().removeAllRanges();
         ReactDOM.findDOMNode(this).focus();
     },
     getMainComponent: function () {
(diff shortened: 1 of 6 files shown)


mhils:
Maybe I'm missing someone, but I don't see that `ProxyAppMain` implements `returnFocus`?

web/src/js/components/proxyapp.js
```diff
@@ -55,6 +55,8 @@ var ProxyAppMain = React.createClass({
         };
     },
     focus: function () {
+        document.activeElement.blur();
+        window.getSelection().removeAllRanges();
```
mhils: What's the point of these two added lines here?

gzzhanghao: @mhils They comes from `ChildFocus.returnFocus`. Not sure if  `activeElement.blur()` is necessary. But for convenience I just moved everything here.

See https://github.com/mitmproxy/mitmproxy/blob/master/web/src/js/components/common.js#L59

gzzhanghao:
@mhils It passes `focus` method as `returnFocus` in context. `ChildFocus.returnFocus` can safely replace the `ProxyAppMain.focus` method. So I think it's not necessary to add a new method which implements the same thing right there. :confused:

mhils:
:+1:

Issue #939, opened 2016-02-15
Check for expired CA certificate


mhils:
_From @mhils on November 4, 2015 10:37_

The mitmproxy CA certificate is generated on the first start of mitmproxy. Due to mitmproxy/mitmproxy/issues/815, the certificate is only valid for three years. `CertStore.from_store` should check if the CA certificate is still valid and raise an error if that is not the case (which we can then act on in mitmproxy).

_Copied from original issue: mitmproxy/netlib#101_

mhils:
_From @tinius on January 14, 2016 1:47_

Seems to me that this would do it:

    ca = OpenSSL.crypto.load_certificate(
                OpenSSL.crypto.FILETYPE_PEM,
                raw)
    if ca.has_expired():
        raise CertExpiredException("mitmproxy's certificate has expired.")

That being said, is there a reason why we shouldn't just call create_store on the spot?

mhils:
> That being said, is there a reason why we shouldn't just call create_store on the spot?

I'm not sure what's the best way, but silently replacing the CA (without user notification) is not really optimal. At the very least, we need some kind of notification that a new certificate has been generated.

anchitjain1234:
@mhils I am a new contributor and would like to work on this 
> At the very least, we need some kind of notification that a new certificate has been generated.

What kind of notification? Like printing on screen that new certificate has been generated or taking user response to whether or not generate a new certificate?

mhils:
Not sure yet. Let's start by adding a function in netlib that determines if the CA is outdated + tests? :smiley:

anchitjain1234:
@mhils please review #995.

dwfreed:
From reading the chromium issue, validity length checks only apply to leaf certificates, not CAs.  This means your CA can still be valid for 20 years (mine is), so long as mitmproxy's generated leaf certificates are only valid for 39 months or less.  Since mitmproxy always generates certs on the fly, and never caches them to disk, checking for expiration of them only matters in really long running instances (usually mitmdump).

baptx:
@mhils This issue seems to be still present on the latest version of mitmproxy (6.0.2), my mitmproxy certificate is expired since 16 December and I got this error on Firefox when visiting a website with HSTS preload like Facebook, which does not allow exceptions:
```
Did Not Connect: Potential Security Issue

Firefox detected an issue and did not continue to www.facebook.com. The website is either misconfigured or your computer clock is set to the wrong time.

What can you do about it?

[...]

PR #995, opened 2016-03-02
Added method and test for checking outdated CA

anchitjain1234:
For #939.

### DIFF
--- a/netlib/certutils.py
+++ b/netlib/certutils.py
@@ -116,6 +116,20 @@ def dummy_cert(privkey, cacert, commonname, sans):
     return SSLCert(cert)
 
 
+def check_outdated_ca(ca):
+    """
+        Checks whether certificate is outdated or not.
+
+        ca: cert
+
+        Returns True if expired, else False.
+    """
+    if ca.has_expired():
+        return True
+    else:
+        return False
+
+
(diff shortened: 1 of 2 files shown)


Kriechi:
Well, that seems a bit too easy, don't you think?

The `check_outdated_ca` function is not really necessary, why don't we just use `has_expired` directly instead of wrapping it again?

anchitjain1234:
> Well, that seems a bit too easy, don't you think?

Yes. What else could be added?

> The check_outdated_ca function is not really necessary, why don't we just use has_expired directly instead of wrapping it again?

I would remove that function. 
Also original issue requires notifying the user about the expired certificate, how this should be achieved?
Like new certificate generation or exception?

mhils:
My last comment in the original issue (https://github.com/mitmproxy/mitmproxy/issues/939#issuecomment-191230712) is bollocks, sorry. I did not know that `has_expired` is a thing and thought that'd need to be implemented.

> Like new certificate generation or exception?

So, any automated response is a bit difficult - users need to react anyway by updating their client to accept the new CA certificate. I think we should just display a warning that instructs the user to delete `~/.mitmproxy` if he wants to generate a new certificate.

So, sorry for the initially bad task description: I think implementing a check for mitmdump first (as we do for HTTP2 ALPN - see `mitmproxy.dump`) would be best.

PR #999, opened 2016-03-04
Add Icon for Binaries

### LINKED ISSUE #981: Add Icon for Binaries
The OSX and Windows binaries we currently generate have the default PyInstaller Icon:
image

It would be super cool if we could have our mitmproxy icon there. The icon is in the [mitmproxy.org repo](https://github.com/mitmproxy/mitmproxy.org) (favicon version in ./src, raw logo in ./logo).

adityadivekar03:
The original icon for the binaries was a default PyInstaller icon. 
This commit will instead create the mitmproxy icon. 
fixes #981


mhils:
Thanks. Can you please use forward slashes (`/` instead of `\`) so that things work on OSX as well? Also, I think it should be a relative path?

adityadivekar03:
@mhils  Yes, you are right. 

I am sorry, I somehow mixed up the Windows backslash here!

mhils:
Look at your PR again - do you think it is correct?

adityadivekar03:
Hi @mhils . Sorry for the absence, I was away on some coursework.
I tried testing the issue locally. When I executed the dev script and tried to check the virtual environment for the executables, I found the scripts. without any icons.
Could you point out where I am going wrong? 
I am running Ubuntu 14.04 on my machine.

mhils:
Maybe look at your PR again - why did you use `\\` in the first place (and not just `\`) ?

adityadivekar03:
The linux and osx system will actually ignore the leading second slash and treat it as a ```/```. I included the ```//``` while I was looking up a few spec file syntax examples. 
I executed the dev script again after replacing the ```//``` with ```/```but still I can only see scripts. No icons.

mhils:
Relative paths are relative to the execution directory. Did you experiment with just using `../docs/favicon.ico`, or  `docs/favicon.ico`?

adityadivekar03:
Yes, I tried both of the above methods. 
The command I use to execute the dev script is - 
```sudo -EH ./dev.sh```

After this as given in the instructions, I activated the virtual environment, and tried executing the binaries. There is no problem in executing them, but they look like plain scripts when I check on them in the ```/venv/bin``` folder

mhils:
I added an icon in `./release/specs` for you - this works just fine for me. :smiley:

adityadivekar03:
Yes, thanks for that!
I tried the process again, but I still see only scripts. Not able to visually verify them.
Do you think I should push the changes as they are, now that you've verified them?

mhils:
Yes, please. Can you please squash the commits as well?

adityadivekar03:
Sorry for the inconvenience, I wasn't familiar with the squashing of commits, and seems my attempt at it failed. 

I will open a new MR.

PR #1004, opened 2016-03-06
[web] VirtualScroll and AutoScroll helper

gzzhanghao:
Implement virtual scroll and auto scroll with helper functions.


web/.babelrc
```diff
@@ -1,3 +1,4 @@
 {
-  "presets": ["es2015", "react"]
+  "presets": ["es2015", "react"],
+  "plugins": ["transform-class-properties"]
```
mhils: I'm ok with this particular addition (as long as we restrict ourselves to static properties), but we should define a general strategy how we want to handle ES7+ additions. 
I think we enable ES7 in the very few cases that meet the following criteria:

- There is an obvious benefit in readability.
- We can replace the code easily with a corresponding ES6 version.
- The resulting code is relatively self-explaining for someone who knows a tiny bit of Python/Java/C++/.... For instance, I'd probably need to look up what `::this.foo()` does when I see it, whereas static properties are easy to grasp.

@cortesi, is this ok for you?

web/src/js/components/eventlog.js
```diff
@@ -1,115 +1,151 @@
...
-            className += "label-default";
+    onViewportUpdate() {
+        const viewport = ReactDOM.findDOMNode(this);
+
+        const vScroll = calcVScroll({
+            itemCount: this.state.entries.length,
+            rowHeight: this.props.rowHeight,
+            viewportTop: viewport.scrollTop,
+            viewportHeight: viewport.offsetHeight,
+            itemHeights: this.state.entries.map(entry => this.heights[entry.id]),
```
mhils: This feels like it's overengineered (besides the performance impact). Were there any issues with just using rowHeight and minRowHeight?

web/src/js/components/eventlog.js
```diff
@@ -1,115 +1,151 @@
...
-            <a
-                href="#"
-                className={className}
-                onClick={this.toggle}>
-                {this.props.name}
-            </a>
+            <pre onScroll={this.onViewportUpdate}>
+                <div style={{ height: vScroll.paddingTop }}></div>
+                {entries.map((entry, index) => (
+                    <div key={entry.id} ref={this.setHeight.bind(this, entry.id)}>
```
mhils: Can we break out the log message to a simple React render function?

web/src/js/components/flowtable.js
```diff
@@ -1,188 +1,217 @@
...
+function FlowRow(props) {
+    const flow = props.flow;
+
+    const className = classNames({
+        "selected": props.selected,
+        "highlighted": props.highlighted,
+        "intercepted": flow.intercepted,
+        "has-request": flow.request,
+        "has-response": flow.response,
+    });
```
mhils: This classNames thing is a good addition :+1:

web/src/js/components/helpers/VirtualScroll.js
```diff
@@ -0,0 +1,69 @@
...
+                paddingBottom += height;
+            }
+
+            pos += height;
+        }
+
+    } else {
+
+        start = Math.max(0, Math.floor(viewportTop / rowHeight) - 1);
+        end = Math.min(
```
mhils: How about adding
```javascript
// Make sure that we start at an even row so that CSS `:nth-child(even)` is preserved
start = Math.floor(start/2)*2;
```
here and then get rid of the second placeholder element altogether?

mhils:
Thanks for the high quality PR again. 
I hate to say it, but it's at the border of being reviewable - we should really try to make PRs as small as possible.

web/src/js/components/eventlog.js
gzzhanghao: @mhils The component behaves weird if there are some large chunks in it. I think this mechanism is necessary for a better user expeirence. As for the performance issue, it works smoothly on my laptop, and it can be faster if we do some optimization on it. Anyway, I'll remove this if you still think it's unnecessary, but I think it's really great for UX.

web/src/js/components/eventlog.js
gzzhanghao: @mhils Do you mean a stateless component? We cannot have refs on stateless components. So if we do that we should either abandon the variable height mechanism or add some stuffs to pass refs back to the parent. :confused:

web/src/js/components/eventlog.js
mhils: > The component behaves weird if there are some large chunks in it.

Did you test that with or without minRowHeight?

web/src/js/components/helpers/VirtualScroll.js
gzzhanghao: @mhils Good idea :smile:

web/src/js/components/eventlog.js
gzzhanghao: I tested in the master branch, there are some tracebacks in the eventlog and the scrollbar jumps when I'm scrolling.

cortesi:
This looks good to me - huge... but good. I think the ES7 transform is OK. Let's rebase and merge.

gzzhanghao:
@mhils @cortesi Alright, done. :wink:

mhils:
:+1: :cake:

PR #1006, opened 2016-03-06
Add upstream socks mode

xhy940801:
Add socks mode of upstream
Used:[username:password@]host[:port]
upstream socks mode and other proxy mode can be used simultaneously
eg:
mitmproxy --socks -US xxxx
mitmproxy -U http://xxxx -US xxxx
mitmproxy -US xxxx
they are all ok

### DIFF
--- a/mitmproxy/cmdline.py
+++ b/mitmproxy/cmdline.py
@@ -127,6 +127,37 @@ def parse_upstream_auth(auth):
     return "Basic" + " " + base64.b64encode(auth)
 
 
+def parse_upstream_socks_spec(url):
+    address, port, username, password = None, None, None, None
+    pattern = re.compile("^.+:.+@")
+    str_usr_psd = pattern.search(url)
+    if str_usr_psd is not None:
+        str_usr_psd = str_usr_psd.group()
+        usr_psd = str_usr_psd.split(":", 2)
+        username = usr_psd[0]
+        password = usr_psd[1]
+        password = password[:-1]
+        url = url[len(str_usr_psd):]
+    pattern = re.compile(":\\d+$")
+    str_port = pattern.search(url)
+    if str_port is not None:
+        str_port = str_port.group()
+        port = int(str_port[1:])
+        address = url[:-len(str_port)]
+    else:
+        address = url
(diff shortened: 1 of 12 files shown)


netlib/socksocket.py
```diff
@@ -0,0 +1,387 @@
+"""SocksiPy - Python SOCKS module.
```
Kriechi: How about replacing this with https://github.com/Anorov/PySocks ?

Kriechi:
This PR adds lots of uncovered code - would you mind adding more tests?
So far only the CLI parsing is tested, but the actual main code not.

mhils:
Thanks for the PR!
This is a great contribution, but there is way too much duplication in the code. We need to think about ways how we can reduce the size of this PR.

xhy940801:
Why the auto test all failed?

xhy940801:
@mhils

mhils:
Thanks for the great PR in #1030. We now need to discuss how we can intergrate SOCKS client functionality into mitmproxy. To be upfront honest, I don't think that we should subclass `socket.socket`,  `tcp.TCPClient` and `ServerConnection` for this feature. For example, code that currently uses `TCPClient` will not expect `tcp_client.connect` to raise a `SocksError`. Put simply, I think we touch too many of the core classes here to make this work.

I would prefer to see this functionality implemented in a self-contained fashion as a protocol layer, e.g. as we do for the `--socks` mode (see [here](https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/proxy/modes/socks_proxy.py)). You can find some documentation of the concept here: http://docs.mitmproxy.org/en/stable/dev/protocols.html
In other words, we'd create a `SocksClientLayer` in mitmproxy.protocol that mainly does two things:

1. Hook `set_server()`: Store the target destination and call the parent layer with the SOCKS proxy IP.
2. Hook `connect()`: Instruct the parent layer to `connect()`, then do a SOCKS handshake over the connection.

If that makes sense to you, I'd love to see a PR with a  first implementation of this concept without tests and with a hardcoded proxy server. We can then iterate on this and fill the remaining gaps if the core is in place. If not, just hit me up on Slack or comment here and I'm happy to clarify things :smiley:

Thanks!

PR #1009, opened 2016-03-07
[web] eliminate SettingsState

gzzhanghao:
Eliminate `common.SettingsState` mixin with props.

### DIFF
--- a/web/src/js/components/footer.js
+++ b/web/src/js/components/footer.js
@@ -1,19 +1,18 @@
 import React from "react";
 import {SettingsState} from "./common.js";
 
-var Footer = React.createClass({
-    mixins: [SettingsState],
-    render: function () {
-        var mode = this.state.settings.mode;
-        var intercept = this.state.settings.intercept;
-        return (
-            <footer>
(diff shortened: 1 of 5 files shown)


web/src/js/components/footer.js
```diff
@@ -1,19 +1,18 @@
...
-    }
-});
+Footer.propTypes = {
+    settings: React.PropTypes.object.isRequired,
+};
 
-export default Footer;
\ No newline at end of file
+export default function Footer({ settings }) {
+    const mode = settings.mode;
```
mhils: Let's try to have a code style that is as uniform as possible. Do you prefer
```javascript
const mode = settings.mode;
const intercept = settings.intercept;
```
or 
```javascript
const {mode, intercept} = settings;
```
?

mhils:
Other than my super minor nitpick above, LGTM! Thanks! :smiley:

web/src/js/components/footer.js
gzzhanghao: @mhils The second one looks better :smiley:

PR #1012, opened 2016-03-08
[web] eliminate common.Router.getParams

gzzhanghao:
Use `props.routeParams` instead of `this.getParams()`. I prefer to remove react-router and use history directly though. :confused:


mhils:
Right -  the props thing just feels weird.
We may want to take a look at redux at some point and use that for params as well, but I haven't played around with it properly.

PR #1013, opened 2016-03-08
[web] Fix label margin in footer

gzzhanghao:
It looks pretty weird since #1009 

image


mhils:
Thanks! :smiley: :tada:

PR #1014, opened 2016-03-08
New option: Add server certs to client chain

ikoz:
If enabled, append all server certificates to the certificate chain
served to the client, as extras. Can be used to test/bypass certain
certificate pinning impementations.

### DIFF
--- a/pathod/pathoc.py
+++ b/pathod/pathoc.py
@@ -42,7 +42,8 @@ class SSLInfo(object):
             "Cipher: %s, %s bit, %s" % self.cipher,
             "SSL certificate chain:"
         ]
-        for i in self.certchain:
+        for n,i in enumerate(self.certchain):
+            parts.append("  Certificate [%s]" % n)
             parts.append("\tSubject: ")
             for cn in i.get_subject().get_components():
                 parts.append("\t\t%s=%s" % cn)
@@ -69,7 +70,7 @@ class SSLInfo(object):
             s = certutils.SSLCert(i)
             if s.altnames:
                 parts.append("\tSANs: %s" % " ".join(s.altnames))
-            return "\n".join(parts)
+        return "\n".join(parts)
 
 
 
--- a/test/mitmproxy/test_server.py
+++ b/test/mitmproxy/test_server.py
@@ -999,3 +999,63 @@ class TestProxyChainingSSLReconnect(tservers.HTTPUpstreamProxyTest):
         # (both terminated)
         # nothing happened here
         assert self.chain[1].tmaster.state.flow_count() == 2
+
+
+class TestHTTPSAddServerCertsToClientChainTrue(tservers.HTTPProxyTest):
+    ssl = True
+    add_server_certs_to_client_chain = True
+    servercert = tutils.test_data.path("data/trusted-server.crt")
+    ssloptions = pathod.SSLOptions(
+            cn="trusted-cert",
+            certs=[
+                ("trusted-cert", servercert)
+            ]
+    )
+
+    def test_add_server_certs_to_client_chain_true(self):
+        """
+        If --add-server-certs-to-client-chain is True, then the client should receive the server's certificates
+        """
+        with open(self.servercert, "rb") as f:
+            d = f.read()
+        c1 = SSLCert.from_pem(d)
+        p = self.pathoc()
+        print("digest of p.cert[1]: %s"%p.server_certs[1].digest('sha256'))
+        print("digest of c1.cert[1]: %s"%c1.digest('sha256'))
+        server_cert_found_in_client_chain = False
+
+        for cert in p.server_certs:
+            if cert.digest('sha256') == c1.digest('sha256'):
+                server_cert_found_in_client_chain = True
+                break
+
+        assert(server_cert_found_in_client_chain == True)
+
+
+class TestHTTPSAddServerCertsToClientChainFalse(tservers.HTTPProxyTest):
+    ssl = True
+    add_server_certs_to_client_chain = False
+    servercert = tutils.test_data.path("data/trusted-server.crt")
+    ssloptions = pathod.SSLOptions(
+            cn="trusted-cert",
+            certs=[
+                ("trusted-cert", servercert)
+            ]
+    )
+
+    def test_add_server_certs_to_client_chain_false(self):
+        """
+        If --add-server-certs-to-client-chain is False, then the client should not receive the server's certificates
+        """
+        with open(self.servercert, "rb") as f:
+            d = f.read()
+        c1 = SSLCert.from_pem(d)
+        p = self.pathoc()
+        server_cert_found_in_client_chain = False
+
+        for cert in p.server_certs:
+            if cert.digest('sha256') == c1.digest('sha256'):
+                server_cert_found_in_client_chain = True
+                break
+
+        assert(server_cert_found_in_client_chain == False)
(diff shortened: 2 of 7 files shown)


mhils:
Thanks for the great PR! :smiley:
Could you add a test which verifies that the feature is working as intended?

ikoz:
Some more context on this was just made public: 

https://www.cigital.com/blog/ineffective-certificate-pinning-implementations/
related: CVE-2016-2402

@mhils What we need for a test is to implement the following scenario:

1. start a web server (443, certificate signed by some self-signed CA)
2. start mitmproxy using the new option
3. start a 'client' and configure it to use the mitmproxy
4. use the 'client' to connect to the web server through mitmproxy
5. use the 'client' to retrieve the certificates mitmproxy sends 
6. if this list includes the certificates the real server is using, then the test succeeds.

Unfortunately I'm not very familiar with the mitmproxy test suite, so I am not sure where to start to implement this. Is the a similar test case already?

mhils:
Really nice find, congratulations! :smiley:

[test/mitmproxy/test_server.py#L419](https://github.com/mitmproxy/mitmproxy/blob/448d28e2cbe5020f97bccc80e9a7e1e5a3df9990/test/mitmproxy/test_server.py#L419) is the closest we have right now I think. `ssloptions` configures the server, `f.sslinfo` is what we get on the client. You can pass additional arguments to the proxy by implementing `get_proxy_config` (just grep for it in the tests). Let me know if you run into any issues.

ikoz:
I finally got around to create two tests for this, please review.

Also found a bug in the (unrelated) cert chain printing code which I was using for debugging, fixed it as well. Could have been a different pull but got mixed up in here.

The only thing missing, which I noticed fairly late, is that this option doesn't work when verify-upstream-cert is also in use. I'm not sure how to make these two options mutually exclusive - argparse groups?

pathod/pathoc.py
```diff
@@ -69,7 +70,7 @@ def __str__(self):
             s = certutils.SSLCert(i)
             if s.altnames:
                 parts.append("\tSANs: %s" % " ".join(s.altnames))
-            return "\n".join(parts)
+        return "\n".join(parts)
```
mhils: Very good find, thanks! :smiley:

test/mitmproxy/test_server.py
```diff
@@ -999,3 +999,63 @@ def handler(f):
...
+    def test_add_server_certs_to_client_chain_true(self):
+        """
+        If --add-server-certs-to-client-chain is True, then the client should receive the server's certificates
+        """
+        with open(self.servercert, "rb") as f:
+            d = f.read()
+        c1 = SSLCert.from_pem(d)
+        p = self.pathoc()
+        print("digest of p.cert[1]: %s"%p.server_certs[1].digest('sha256'))
+        print("digest of c1.cert[1]: %s"%c1.digest('sha256'))
```
mhils: Can we remove the print statements from the tests please?

test/mitmproxy/test_server.py
```diff
@@ -999,3 +999,63 @@ def handler(f):
...
+        c1 = SSLCert.from_pem(d)
+        p = self.pathoc()
+        server_cert_found_in_client_chain = False
+
+        for cert in p.server_certs:
+            if cert.digest('sha256') == c1.digest('sha256'):
+                server_cert_found_in_client_chain = True
+                break
+
+        assert(server_cert_found_in_client_chain == False)
```
mhils: These look really good - thanks. How about replacing the last assert with `== self.add_server_certs_to_client_chain` and then moving all the common code in a mixin? (happy to help here if you are not sure how to do this)

mhils:
Good job - thanks for the high-quality PR!

> The only thing missing, which I noticed fairly late, is that this option doesn't work when verify-upstream-cert is also in use. 

Why is that the case? Any idea?

> Finally, I now also realize that I could rename 'server' and 'client' to 'upstream' and 'downstream' if that's better for consistency, let me know.

I think it's just fine, thanks. :relaxed:

test/mitmproxy/test_server.py
ikoz: not sure how I missed that, will fix.

test/mitmproxy/test_server.py
ikoz: That seems like a good idea. Not sure what how mixin works, will try to understand.

ikoz:
> Why is that the case? Any idea?

Have to test this further, but the reasoning seems to be this: If the chain is verified, then the API that retrieves the server's cert chain only retrieves the 'validated trusted path' (e.g. 1 or 2 certificates) - instead of 'all' the certificates that the server really sends down to the proxy (which what we need for this feature).

test/mitmproxy/test_server.py
```diff
@@ -999,3 +999,50 @@ def handler(f):
...
+    """
+    add_server_certs_to_client_chain = False
+    ssl = True
+    servercert = tutils.test_data.path("data/trusted-server.crt")
+    ssloptions = pathod.SSLOptions(
+        cn="trusted-cert",
+        certs=[
+            ("trusted-cert", servercert)
+        ]
+    )
```
mhils: Excellent. I think we can even move `ssl`, `servercert` and `ssloptions` into `AddServerCertsToClientChainMixin` ? :smiley:

mhils:
> Have to test this further, but the reasoning seems to be this: If the upstream chain is verified, then the API that retrieves the server's cert chain only retrieves the 'validated trusted path' (e.g. 1 or 2 certificates) - instead of 'all' the certificates that the server really sends down to the proxy (which what we need for this feature).
View changes

OpenSSL ¯\_(ツ)_/¯

Let's just make it a [mutually exclusive group](https://github.com/mitmproxy/mitmproxy/blob/f118d9abb16b608d28cbd4a78356791802286df1/mitmproxy/cmdline.py#L392) then.

ikoz:
Here we go. 

One last thing: The no-upstream-cert feature is also obviously exclusive to this new feature, since the  [_establish_tls_with_client_and_server](https://github.com/mitmproxy/mitmproxy/blob/4604c25c6055a37e5f25a238d2a089759bd5d98a/mitmproxy/protocol/tls.py#L345) little dance doesn't happen if it's enabled. 

Do you think no-upstream-cert should be moved into the mutually exclusive group? 

Is no-upstream-cert mutually exclusive to verify-upstream-cert ? 

It sounds like it is, but the underlying verify_options inside [_establish_tls_with_server](https://github.com/mitmproxy/mitmproxy/blob/4604c25c6055a37e5f25a238d2a089759bd5d98a/mitmproxy/protocol/tls.py#L462) is reachable even if no-upstream-cert is enabled.

mhils:
> Is no-upstream-cert mutually exclusive to verify-upstream-cert ?

I don't think it is or should be. no-upstream-cert merely denotes that the upstream cert/connection is not used/needed for generating the fake certificate, but we still want to retain the ability to verify it if there is an upstream connection. So, we have 
```
add_upstream_certs_to_client_chain xor no_upstream_cert
add_upstream_certs_to_client_chain xor verify_upstream_cert
```
but no interdependency between `no_upstream_cert` and `verify_upstream_cert`. Now that we can't express that with exclusive groups, do we just want to do ditch these and do manual checks in [mitmproxy.proxy.config](https://github.com/mitmproxy/mitmproxy/blob/f118d9abb16b608d28cbd4a78356791802286df1/mitmproxy/proxy/config.py#L140)?

ikoz:
Good point. Let's see if that works.

mhils:
Excellent. Thanks for the awesome PR! :smiley: :cake: :tada:

PR #1015, opened 2016-03-09
[web] dont mutate props

gzzhanghao:
Props are frozen since React 0.14.

https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#breaking-changes


mhils:
:+1: 

This is why big PRs are bad. I fixed this in https://github.com/mitmproxy/mitmproxy/compare/react-higher-order-functions a few days ago as well. I just should have done what you have done here - thanks! :smiley:

PR #1019, opened 2016-03-10
add resolved IP address in "Details" tab

lilydjwg:
It's useful to know the peer address when the domain has several IP addresses and you need to know where the request has gone.

### DIFF
--- a/mitmproxy/console/flowdetailview.py
+++ b/mitmproxy/console/flowdetailview.py
@@ -23,6 +23,7 @@ def flowdetails(state, flow):
         text.append(urwid.Text([("head", "Server Connection:")]))
         parts = [
             ["Address", repr(sc.address)],
+            ["Resolved", repr(sc.sock_address)],
         ]
 
         text.extend(
--- a/mitmproxy/console/flowdetailview.py
+++ b/mitmproxy/console/flowdetailview.py
@@ -23,7 +23,7 @@ def flowdetails(state, flow):
         text.append(urwid.Text([("head", "Server Connection:")]))
         parts = [
             ["Address", repr(sc.address)],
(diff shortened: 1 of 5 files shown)


Kriechi:
Please don't increment our version numbers - it messes with too many things.
You can just use `convert_016_017` since 0.17 isn't release yet.

lilydjwg:
Oh, that's good! I've updated the commit.

mhils:
This is a nice addition - thanks!
I think `sock_address` is a fairly nondescriptive attribute name. I don't have better suggestions right now, but can anybody else come up with a decent alternative?

Kriechi:
@lilydjwg can you try and fix the tests?
If no other name is proposed, I'm just merging it the way it is (once the tests pass).

lilydjwg:
Hi, I've fixed the tests :-)

lilydjwg:
And for the name, what about "peer_address"? It's the address of the peer mitmproxy connects to.

mhils:
Thanks for fixing the tests! How do you feel about `peername`? That would match the `getpeername()` syscall at least.

lilydjwg:
When I see something called "xxx name", I will think it as a string. But `getpeername` is somewhat special; for TCP sockets I'll get a tuple of `(ip_address, port)`. But what we have here is an `Address`. I prefer to have "address" in it, matching other `*address` attributes.

mhils:
Ok, let's go for peer_address then! :smiley:

mhils:
Resurrecting an old thread - would anyone object against `ip_address`?

lilydjwg:
@mhils can mitmproxy connect to a UNIX socket address of a proxy (now or in the future)? If not, I'll like it.

PR #1023, opened 2016-03-11
Add mitmproxy icon for binaries

adityadivekar03:
The executables created earlier used the default PyInstaller icon.
Now, the mitmproxy icon is used.


mhils:
Thanks! :smiley: :tada:

PR #1027, opened 2016-03-12
[web] fix virtual scroll range

gzzhanghao:
There will be an empty row at the end of the flow table when an odd row was preserved.


mhils:
:+1:

PR #1030, opened 2016-03-15
Add 'UsernamePasswordAuth' 'UsernamePasswordAuthResponse' to SOCKS

xhy940801:
Add `UsernamePasswordAuth` and `UsernamePasswordAuthResponse` to `netlib.socks`

### DIFF
--- a/netlib/socks.py
+++ b/netlib/socks.py
@@ -10,7 +10,6 @@ class SocksError(Exception):
         super(SocksError, self).__init__(message)
         self.code = code
 
-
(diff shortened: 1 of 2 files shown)


Kriechi:
I guess this only makes sense together with #1006 - or how and where would you actually use this code?

mhils:
Yes - I asked @xhy940801 to factor out some smaller bits from #1006 so that we can work on it more iteratively. This PR looks great, thanks! :smiley:

PR #1034, opened 2016-03-16
Add 'SocksSocket' to SOCKS5

xhy940801:
A part of  #1006


mhils:
Thanks. Did you take a look at my last comment in #1006? I don't think we want/need to subclass `socket.socket`.

PR #1041, opened 2016-03-18

mhils:


setup.py
```diff
@@ -100,7 +100,7 @@
         'dev': [
             "coveralls>=1.1, <1.2",
             "mock>=1.3.0, <1.4",
-            "pytest>=2.8.7, <2.9",
+            "pytest>=2.9,<2.10",
```
Kriechi: are we using any features from 2.9 yet?
I would like to keep the minimum version as low as possible (for all dependencies!)

mhils: No. I reverted this to `>=2.8.7,< 2.10`

PR #1042, opened 2016-03-18
implement SOCKS as a layer (All hardcoded)

xhy940801:
@mhils First version completed:smile:


mhils:
Thanks! This looks pretty good so far.
Few remarks: It would be nice to populate the ServerConnection's `via` attribute. Here's how we do it for HTTP CONNECT: [ConnectServerConnection](https://github.com/mitmproxy/mitmproxy/blob/b90579fe4513c1794bb8159880479a05da02cbae/mitmproxy/protocol/http.py#L68) It might be worthwhile to rename that to ProxyServerConnection, move it into mitmproxy.protocol.base and re-use it?

mitmproxy/protocol/socks_client.py
```diff
@@ -0,0 +1,82 @@
...
+from netlib.exceptions import TcpException
+from netlib import tcp
+
+from .base import Layer
+from ..exceptions import ProtocolException
+
+class SocksClientLayer(Layer):
+
+    def __init__(self, ctx, server_tls):
+        Layer.__init__(self, ctx)
```
mhils: Please use `super` for calling parent classes.

mitmproxy/protocol/socks_client.py
```diff
@@ -0,0 +1,82 @@
...
+        self.socks_addr = tcp.Address(("127.0.0.1", 1080))
+        self.socks_username = "test"
+        self.socks_password = "test"
+
+    def __call__(self):
+        layer = self.ctx.next_layer(self)
+        layer()
+
+    def _socks_username_password_auth(self, rfile, wfile):
+        if self.socks_username is None or self.socks_password is None:
```
mhils: This check is duplicate, so we can remove it? Also, we don't need to pass an extra rfile and wfile to this function I think. :smiley:

mitmproxy/protocol/socks_client.py
```diff
@@ -0,0 +1,82 @@
...
+            raise socks.SocksError(0, "authentication failed")
+        cauth = socks.UsernamePasswordAuth(socks.USERNAME_PASSWORD_VERSION.DEFAULT, self.socks_username, self.socks_password)
+        cauth.to_file(wfile)
+        wfile.flush()
+        sauth = socks.UsernamePasswordAuthResponse.from_file(rfile)
+        if sauth.status != 0:
+            raise socks.SocksError(0, "authentication failed")
+
+    def _get_atype(self, address):
```
mhils: This is ugly, but unavoidable as it seems. Seems to be generally useful though, so let's move it into netlib.socks?

mitmproxy/protocol/socks_client.py
```diff
@@ -0,0 +1,82 @@
...
+        try:
+            if address.use_ipv6:
+                taddr = socket.inet_pton(socket.AF_INET6, address.host)
```
mhils: Did you check if that works without false-positives? We use `ipaddress.IPv4Address(self.addr.host)` in `netlib.socks`, this could be a nasty sources of inconsistencies.

mitmproxy/protocol/socks_client.py
```diff
@@ -0,0 +1,82 @@
...
+    def connect(self):
+        if not self.target_address:
+            raise ProtocolException("Cannot connect to server, no server address given.")
+        self.server_conn.address = self.socks_addr
+        try:
+            self.server_conn.connect()
+            rfile, wfile = self.server_conn.rfile, self.server_conn.wfile
+            method = (socks.METHOD.NO_AUTHENTICATION_REQUIRED,)
+            if self.socks_username is not None and self.socks_password is not None:
+                method = (socks.METHOD.NO_AUTHENTICATION_REQUIRED, socks.METHOD.USERNAME_PASSWORD)
```
mhils: Can't we just use a list here and `.append`?

mitmproxy/protocol/socks_client.py
```diff
@@ -0,0 +1,82 @@
+import socket
+
+from netlib import socks
```
mhils: Very minor nitpick: `from netlib.socks import METHOD, VERSION, CMD, ...` should reduce a lot of text noise further down.

mhils:
Overall, this looks really great - thanks! :smiley: :tada:

mitmproxy/protocol/socks_client.py
xhy940801: This check is not duplicate.When the username and password for MITMProxy not set and also the SOCKS server tell us we can only accept METHOD.USERNAME_PASSWORD,  it can throw an exception. :wink:

mhils: Good catch. Looking closely, we may just want to do a  
```python
if sgreeting.method not in methods:
    raise socks.SocksError(None, "No acceptable authentication methods")
``` 
check in `connect` to cover all cases and skip that one here.

mhils:
Great work. Can you move `ProxyServerConnection` to `protocol/base` and remove all changes to `root_context`? Other than that, this looks ready to merge. :smiley:

xhy940801:
@mhils Finished :smile:

Kriechi:
@xhy940801 looks like the test could be improved to cover all new code paths - maybe you could try to fix that?

xhy940801:
Could I add the test in a new PR? #1006

xhy940801:
@mhils @Kriechi I get a merge conflict in mitmproxy/protocol/http.py

test/netlib/tservers.py
```diff
@@ -107,3 +108,67 @@ def teardown_class(cls):
     @property
     def last_handler(self):
         return self.server.server.last_handler
+
+
+class SocksHandler(tcp.BaseHandler):
```
Kriechi: This part duplicates quite a bit of code from the above implementation... This means the test is almost as complex as the actual code - so we don't gain very much from it.
Can we maybe test it without re-implementing it, but maybe using `urllib2` or `socksipy`?

xhy940801: socksipy is just a socks client, it's not a socks server. May I use mitmproxy.proxy.socks_proxy to test this?

Kriechi: Did you take a look at this? https://github.com/Anorov/PySocks

mitmproxy/protocol/__init__.py
```diff
@@ -34,6 +34,7 @@
 from .http1 import Http1Layer
 from .http2 import Http2Layer
 from .rawtcp import RawTCPLayer
+from .socks_client import SocksClientLayer
```
mhils: Needs to be mentioned in `__all__` below.

mitmproxy/protocol/socks_client.py
```diff
@@ -0,0 +1,86 @@
...
+    def _socks_username_password_auth(self):
+        rfile, wfile = self.server_conn.rfile, self.server_conn.wfile
+        cauth = UsernamePasswordAuth(USERNAME_PASSWORD_VERSION.DEFAULT, self.socks_username, self.socks_password)
+        cauth.to_file(wfile)
+        wfile.flush()
+        sauth = UsernamePasswordAuthResponse.from_file(rfile)
+        if sauth.status != 0:
+            raise SocksError(0, "authentication failed")
+
+    def change_socks_proxy_server(self, address, username = None, password = None):
```
mhils: This is a great idea - how about we call it `change_proxy_server` for both `http` and `socks`?

mhils:
Sorry for getting back to this so late. :-(

This looks very nice so far, good job. All we need is a bit more [coverage](https://coveralls.io/builds/5602663/source?filename=mitmproxy%2Fprotocol%2Fsocks_client.py) and then we can add it to our protocol layer detection.

cortesi:
Closing due to inactivity/staleness. Please rebase to master and re-submit a PR if you'd still like to work on this.

PR #1043, opened 2016-03-18
Better scripts

mhils:
This PR slightly improves the semantics of `mitmproxy.script.Script` (Scripts must now be loaded explicitly or using a `with` statement) and cleans up the tests.

### DIFF
--- a/test/mitmproxy/script/test_reloader.py
+++ b/test/mitmproxy/script/test_reloader.py
@@ -1,0 +1,29 @@
+import mock
+from mitmproxy.script.reloader import watch, unwatch
+from test.mitmproxy import tutils
+from threading import Event
+
+
+def test_simple():
+    with tutils.tmpdir():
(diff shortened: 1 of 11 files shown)


test/mitmproxy/script/test_reloader.py
```diff
@@ -0,0 +1,34 @@
...
+
+        watch(script, _onchange)
+        with tutils.raises("already observed"):
+            watch(script, _onchange)
+
+        # Some reloaders don't register a change directly after watching, because they first need to initialize.
+        # To test if watching works at all, we do repeated writes every 100ms.
+        for _ in range(100):
+            with open("foo.py", "a") as f:
+                f.write(".")
```
Kriechi: maybe we should also call `flush` - if that is even a thing?

mhils: `with` closes files on exit, which should also flush.

PR #1045, opened 2016-03-19
Port mitmproxy.script to Python 3

mhils:
Let's start porting mitmproxy to Python 3 :tada: 

This PR makes `mitmproxy.script` PY2/3 cross-compatible and adds the script tests to travis.


Kriechi:
Yeah! LGTM!

dufferzafar:
@mhils What strategy do you follow to do the port? Currently, I'm just running a particular test file with the python 3 interpreter and am fixing errors as they come up, one at a time. I was wondering if there's a better/different way to do things?

mhils:
@dufferzafar: I do that module-wise.

1. Read through the module, fix all simple stuff (six.moves, print, ...)
2. Check if there are any more complex issues (bytes/str)
3. Clean up tests - make sure that there's a corresponding test file with ideally 100% coverage.
4. Run tests (`py.test -f modulename`), fix everything that comes up.

1,3 and 4 are usually straightforward. The pain part is when it comes to str/bytes problems. Fixing stuff like [`netlib.utils.clean_bin`](https://github.com/mitmproxy/mitmproxy/blob/96782407833a7f827e79fa7aa5f2576f1bee3547/netlib/utils.py#L91) is what takes time.

Issue #976, opened 2016-02-23
Flow Export Improvements


mhils:
There are currently three ways to export flows in mitmproxy:

- `b` "save request/response body"
- `E` "export"
- `P` "copy request/response (content/headers) to clipboard"

This is kind of confusing - in particular, there's a lot of overlap between `E` and `P`. Having `b` separately isn't too bad I think, this probably the most-popular action you want to do. I'd propose we do something like this:

![image](https://cloud.githubusercontent.com/assets/1019198/13254233/cda4effa-da40-11e5-96e9-830d9ab55c24.png)

Thoughts?

dufferzafar:
This is certainly better than what we currently have. Apart from the `b` key duplicating things a little bit, I can't see anything wrong with it. 

I'll start working on this as soon as possible! :smile_cat: 

P.S: What tool did you use to create that neat chart?

mhils:
Great, happy to hear that! :smiley:

Chart is made in 5 minutes with Microsoft Visio. :stuck_out_tongue:

Dutchy-:
Do you think it's possible to account for marked flows? I have the occasional use case where I export multiple raw requests/responses to include them in a report. For example, to document a login or redirect sequence.

mhils:
I think we want to limit this export feature to single flows for now. The semantics of the whole feature get too complex otherwise I'm afraid.

dufferzafar:
These hotkeys should be available in both `flowlist` and `flowview` right?

In `flowview`, the key `p` is already bound to previous flow!

dufferzafar:
Also, [`Headers+Content`](https://github.com/dufferzafar/mitmproxy/blob/flow-export/mitmproxy/console/common.py#L263-265) is the same as [`Raw Request`](https://github.com/dufferzafar/mitmproxy/blob/flow-export/mitmproxy/flow/export.py#L97-100)

mhils:
> These hotkeys should be available in both flowlist and flowview right?

Yes! :smiley:

> In flowview, the key p is already bound to previous flow!

Hm - unless there are better options, we can probably remove this?

> So, having them both doesn't make much sense.

Agreed - let's just keep one of these :+1:

dufferzafar:
The `Up/Down` arrow keys seem to free in flowview. I could bind Prev/Next to them?

PR #1049, opened 2016-03-23
Export flows to locust.io (http://locust.io)

zlorb:
Add options for exporting flows to be used with locust load test tool (https://github.com/locustio/locust):
- Export locust:  complete code for running locust test on a specific flow.
- Export locust task: additional code snippet, to be added as an extra flow to a pre-existing locust code (as above).


mitmproxy/flow_export.py
```diff
@@ -2,6 +2,12 @@
 import netlib.http
 from textwrap import dedent
 
+import sys, re
+if (sys.version_info > (3, 0)):
+    from urllib.parse import urlparse
+else:
+    import urlparse
```
Kriechi: @mhils started using `six` for py2/3 compatible imports in recent commits to master.
Maybe you could try to adopt it here as well?

Kriechi:
Did you take a look at #976?
Maybe we can try to improve the situation (if possible) and plan for future refactorings...

test/mitmproxy/test_flow_export.py
```diff
@@ -1,361 +1,361 @@
-from textwrap import dedent
```
Kriechi: Your branch looks a bit messy with a lot of diff.
Could you please try to rebase to current master and keep the file modes intact (0644 vs. 0755).
Ideally Github should only list very few files in the "Files changed" tab - and not 106!

zlorb:
Github client on Windows seems to mess up file permissions on mounted drives.
Will clean up and create a new pull request.

PR #1050, opened 2016-03-23
Flow export to locust.io load test tool [http://locust.io]

zlorb:
Add options for exporting flows to be used with [Locust load test tool](https://github.com/locustio/locust):
- Export locust: complete code for running locust test on a specific flow.
- Export locust task: additional code snippet, to be added as an extra flow to a pre-existing locust code (as above).

### DIFF
--- a/docs/features/responsestreaming.rst
+++ b/docs/features/responsestreaming.rst
@@ -48,8 +48,7 @@ Implementation Details
 ----------------------
 
 When response streaming is enabled, portions of the code which would have otherwise performed
-changes on the response body will see an empty response body instead
-(:py:data:`netlib.http.CONTENT_MISSING`). Any modifications will be ignored.
+changes on the response body will see an empty response body. Any modifications will be ignored.
 
 Streamed responses are usually sent in chunks of 4096 bytes. If the response is sent with a
 ``Transfer-Encoding: chunked`` header, the response will be streamed one chunk at a time.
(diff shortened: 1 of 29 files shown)


zlorb:
[#976](https://github.com/mitmproxy/mitmproxy/issues/976) is beyond the scope of this change.

mhils:
Thanks for the PR! I'm not very familiar with locust:

1. Is there any reason why we need to export two different syntaxes?
2. Don't you want to integrate it in the UI somehow?

zlorb:
Thanks for looking into this PR. 
Below are answers to @mhils questions.

**1. Is there any reason why we need to export two different syntaxes?**

Having two exports saves quite a bit of editing. Expected user workflow is something like:
- Export first flow as a locust snippet of code. This snippet is saved as a python file and can be used as-is to run load tests.
  The code in this snippet contains few imports and two class definitions.
- If needed, user exports additional flows, and then adds them as to the snippet above.
  These are going to be only function declarations inside one of the existing classes.

With the above one can take flows captured by mitmproxy and replay them as a load test within seconds.

**2. Don't you want to integrate it in the UI somehow?**

Mitmproxy already has options for exporting to curl and to python code. The use-case here is the same - take a flow and "re-play" it using external tools. Therefore it makes sense to keep locust playback options together with the curl/python exports.


It would definitely be nice to have these show up in the UI and the web GUI as well. In that case, we would probably want to move other export options as well.

mhils:
Thanks for the clarification. Could we please make `locust_code` re-use `locust_task`?

zlorb:
Done: re-use `locust_code` in `locust_task` code, and merge with latest master.

mhils:
Thanks! :smiley:

PR #1055, opened 2016-03-26
Simplify '.content' by removing CONTENT_MISSING

### LINKED ISSUE #963: Simplify `.content`
Our HTTPRequest/HTTPResponse's `.content` attribute can currently be either

- `None`, if we have not read the body yet (e.g. in the `reponseheaders` hook)
- `CONTENT_MISSING=0`, if we deliberately deleted the body
- `b""`, if we read an (empty) body.
- `b"some-content"`, if we actually read some content.

While the first three values intuitively evaluate to `False`, I feel the bytes/int/None heterogenity a bit confusing. I'd like to combine the first two cases, so that the type of this attribute can accurately be described as `Optional[bytes]`. This would fix #692 and make the code IMO easier to understand. We could also first deprecate `CONTENT_MISSING` gradually by setting it to `None`, and then remove it later on. @cortesi, would you be okay with this approach?

MatthewShao:
fixes #963
As @mhils said, I did this in two step:

- [x] setting ```None``` to ```CONTENT_MISSING``` (to see if anything go wrong)
- [x] replacing ```CONTENT_MISSING``` with ```None```

The relative comments and document have been updated too, hope I didn't miss anything.

### DIFF
--- a/mitmproxy/console/common.py
+++ b/mitmproxy/console/common.py
@@ -4,7 +4,6 @@ import urwid
 import urwid.util
 import os
 
-from netlib.http import CONTENT_MISSING
 import netlib.utils
 
 from .. import utils
@@ -256,7 +255,7 @@ def copy_flow_format_data(part, scope, flow):
     else:
         data = ""
         if scope in ("q", "a"):
-            if flow.request.content is None or flow.request.content == CONTENT_MISSING:
+            if flow.request.content is None:
                 return None, "Request content is missing"
             with decoded(flow.request):
                 if part == "h":
@@ -269,7 +268,7 @@ def copy_flow_format_data(part, scope, flow):
             # Add padding between request and response
             data += "\r\n" * 2
         if scope in ("s", "a") and flow.response:
-            if flow.response.content is None or flow.response.content == CONTENT_MISSING:
+            if flow.response.content is None:
                 return None, "Response content is missing"
             with decoded(flow.response):
                 if part == "h":
@@ -418,7 +417,7 @@ def format_flow(f, focus, extended=False, hostheader=False, marked=False):
     if f.response:
         if f.response.content:
             contentdesc = netlib.utils.pretty_size(len(f.response.content))
-        elif f.response.content == CONTENT_MISSING:
+        elif f.response.content is None:
             contentdesc = "[content missing]"
         else:
             contentdesc = "[no content]"
--- a/mitmproxy/models/http.py
+++ b/mitmproxy/models/http.py
@@ -84,9 +84,8 @@ class HTTPRequest(MessageMixin, Request):
 
         headers: Headers object
 
-        content: Content of the request, None, or CONTENT_MISSING if there
-        is content associated, but not present. CONTENT_MISSING evaluates
-        to False to make checking for the presence of content natural.
+        content: Content of the request, the value is None if there is content
+        associated, but not present.
 
         form_in: The request form which mitmproxy has received. The following
         values are possible:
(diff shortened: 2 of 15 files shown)


mitmproxy/models/http.py
```diff
@@ -226,9 +225,8 @@ class HTTPResponse(MessageMixin, Response):
 
         headers: Headers object
 
-        content: Content of the request, None, or CONTENT_MISSING if there
-        is content associated, but not present. CONTENT_MISSING evaluates
-        to False to make checking for the presence of content natural.
+        content: Content of the request, the value is None if there is content
```
Kriechi: ...Content of the **response**...

mitmproxy/protocol/http.py
```diff
@@ -50,8 +50,8 @@ def read_response_body(self, request, response):
         yield "this is a generator"  # pragma: no cover
 
     def send_response(self, response):
-        if response.content == CONTENT_MISSING:
-            raise HttpException("Cannot assemble flow with CONTENT_MISSING")
+        if response.content == None:
+            raise HttpException("Cannot assemble flow with None content")
```
Kriechi: ...Cannot assemble flow with **missing content**

netlib/http/http1/assemble.py
```diff
@@ -3,12 +3,10 @@
...
 import itertools
 from ...exceptions import HttpException
-from .. import CONTENT_MISSING
-
 
 def assemble_request(request):
-    if request.content == CONTENT_MISSING:
-        raise HttpException("Cannot assemble flow with CONTENT_MISSING")
+    if request.content == None:
+        raise HttpException("Cannot assemble flow with None content")
```
Kriechi: ...**missing content**

netlib/http/http1/assemble.py
```diff
@@ -21,8 +19,8 @@ def assemble_request_head(request):
 
 
 def assemble_response(response):
-    if response.content == CONTENT_MISSING:
-        raise HttpException("Cannot assemble flow with CONTENT_MISSING")
+    if response.content == None:
+        raise HttpException("Cannot assemble flow with None content")
```
Kriechi: ...**missing content**

mitmproxy/console/common.py
```diff
@@ -418,7 +417,7 @@ def format_flow(f, focus, extended=False, hostheader=False, marked=False):
     if f.response:
         if f.response.content:
             contentdesc = netlib.utils.pretty_size(len(f.response.content))
-        elif f.response.content == CONTENT_MISSING:
+        elif f.response.content is None:
             contentdesc = "[content missing]"
         else:
             contentdesc = "[no content]"
```
Kriechi: Just for clarification:
The else-block is only reached if the content is basically an empty string?
Or is this code non-reachable now?

MatthewShao: I think it is only reached if the content is ```""``` (empty string) or it is ```False```.

mhils: This is reached if the content is `""`. The confusion is justified though - can we please do an explicit `content == ""` check?

mitmproxy/models/http.py
mhils: @Kriechi: ?

mitmproxy/web/app.py
```diff
@@ -26,7 +25,7 @@ def _strip_content(flow_state):
                 continue
             if message["content"]:
                 message["contentLength"] = len(message["content"])
-            elif message["content"] == CONTENT_MISSING:
+            elif message["content"] is None:
```
mhils: This `if elif else` can be simplified to `if else` by combining the cases where content is not None.

mhils:
Other than my minor comments above, LGTM! :smiley:

mitmproxy/models/http.py
Kriechi: Ah yes - it's ok.
For some reason the Github diff view indicated a different class...
screen shot 2016-03-26 at 18 07 15

Kriechi:
Thanks! :tada:

PR #1057, opened 2016-03-27
Safeguard

### LINKED ISSUE #907: On Shutdown: flow is None and crashes
The Version is 0.15. Crash info as follow:

```
./mitmdump.run.sh: line 2: 14504 Segmentation fault      mitmdump -p 59873 > network
  File "/usr/local/lib/python2.7/site-packages/libmproxy/proxy/server.py", line 120, in handle
    root_layer()
  File "/usr/local/lib/python2.7/site-packages/libmproxy/proxy/modes/http_proxy.py", line 10, in __call__
    layer()
  File "/usr/local/lib/python2.7/site-packages/libmproxy/protocol/tls.py", line 272, in __call__
    layer()
  File "/usr/local/lib/python2.7/site-packages/libmproxy/protocol/http.py", line 130, in __call__
    layer()
  File "/usr/local/lib/python2.7/site-packages/libmproxy/protocol/http.py", line 323, in __call__
    self.handle_regular_mode_connect(request)
  File "/usr/local/lib/python2.7/site-packages/libmproxy/protocol/http.py", line 409, in handle_regular_mode_connect
    layer()
  File "/usr/local/lib/python2.7/site-packages/libmproxy/protocol/tls.py", line 272, in __call__
    layer()
  File "/usr/local/lib/python2.7/site-packages/libmproxy/protocol/http.py", line 130, in __call__
    layer()
  File "/usr/local/lib/python2.7/site-packages/libmproxy/protocol/http.py", line 340, in __call__
    self.get_response_from_server(flow)
  File "/usr/local/lib/python2.7/site-packages/libmproxy/protocol/http.py", line 471, in get_response_from_server
    if flow.response.stream:
AttributeError: 'NoneType' object has no attribute 'response'
```

Kriechi:
* cleanup h2 code a bit
  * this might fix #907

### DIFF
--- a/mitmproxy/protocol/http.py
+++ b/mitmproxy/protocol/http.py
@@ -187,8 +187,7 @@ class HttpLayer(Layer):
                     flow.request.headers["Proxy-Authorization"] = self.config.upstream_auth
                 self.process_request_hook(flow)
 
-                if not flow.response:
-                    self.establish_server_connection(flow)
+                if not flow.response and self.establish_server_connection(flow):
                     self.get_response_from_server(flow)
                 else:
                     # response was set by an inline script.
@@ -197,6 +196,9 @@ class HttpLayer(Layer):
 
                 self.log("response", "debug", [repr(flow.response)])
                 flow = self.channel.ask("response", flow)
+
+                if not flow.response:
+                    return
                 self.send_response_to_client(flow)
 
                 if self.check_close_connection(flow):
@@ -304,9 +306,12 @@ class HttpLayer(Layer):
             # > server detects timeout, disconnects
             # > read (100-n)% of large request
             # > send large request upstream
-            self.disconnect()
-            self.connect()
-            get_response()
+            try:
+                self.disconnect()
+                self.connect()
+                get_response()
+            except NotImplementedError:
+                return
 
         # call the appropriate script hook - this is an opportunity for an
         # inline script to set flow.stream = True
@@ -359,7 +364,10 @@ class HttpLayer(Layer):
                 self.set_server(address, tls, address.host)
             # Establish connection is neccessary.
             if not self.server_conn:
-                self.connect()
+                try:
+                    self.connect()
+                except:
+                    return False
         else:
             if not self.server_conn:
                 self.connect()
@@ -379,6 +387,7 @@ class HttpLayer(Layer):
                 tls_layer = TlsLayer(self, False, True)
                 tls_layer._establish_tls_with_server()
             """
+        return True
 
     def validate_request(self, request):
         if request.first_line_format == "absolute" and request.scheme != "http":
--- a/mitmproxy/protocol/http2.py
+++ b/mitmproxy/protocol/http2.py
@@ -26,11 +26,6 @@ class SafeH2Connection(H2Connection):
         self.conn = conn
         self.lock = threading.RLock()
 
-    def safe_close_connection(self, error_code):
-        with self.lock:
-            self.close_connection(error_code)
-            self.conn.send(self.data_to_send())
-
     def safe_increment_flow_control(self, stream_id, length):
         if length == 0:
             return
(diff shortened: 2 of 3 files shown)


mitmproxy/protocol/http.py
```diff
@@ -187,8 +187,7 @@ def __call__(self):
                     flow.request.set_auth(self.config.upstream_auth)
                 self.process_request_hook(flow)
 
-                if not flow.response:
-                    self.establish_server_connection(flow)
+                if not flow.response and self.establish_server_connection(flow):
```
mhils: This looks wrong? Why do we do this here?

mitmproxy/protocol/http.py
```diff
@@ -196,11 +195,16 @@ def __call__(self):
...
                         raise Kill()
+                    if flow is None:
+                        return
 
                 self.log("response", "debug", [repr(flow.response)])
                 flow = self.channel.ask("response", flow)
                 if flow == Kill:
                     raise Kill()
+                if flow is None or flow.response is None:
+                    return
```
mhils: How can `flow.response` be None here?

mitmproxy/protocol/http.py
```diff
@@ -308,9 +312,12 @@ def get_response():
...
             # > read (100-n)% of large request
             # > send large request upstream
-            self.disconnect()
-            self.connect()
-            get_response()
+            try:
+                self.disconnect()
+                self.connect()
+                get_response()
+            except NotImplementedError:
```
mhils: WTF?

mitmproxy/protocol/http.py
```diff
@@ -368,7 +375,10 @@ def establish_server_connection(self, flow):
                 self.set_server(address, tls, address.host)
             # Establish connection is neccessary.
             if not self.server_conn:
-                self.connect()
+                try:
+                    self.connect()
+                except:
```
mhils: This looks like a very bad idea as well?

mitmproxy/protocol/http.py
Kriechi: If the http2 layer gets terminated (due to a closed connection), `establish_server_connection` would go into a weird state in the h2 protocol - so it is easier just to cancel the whole stack of layers, and begin a fresh connection.

mitmproxy/protocol/http.py
Kriechi: Similar to above, if h2 gets canceled there might be no response, or even no flow at all.
Yes, this is all very strange and dirty - but apparently happens (although not deterministically reproducible)

mitmproxy/protocol/http.py
Kriechi: Yes - this needs to be improved - same reasoning as above.
If h2 fails once, it does not make sense to reconnect, since the h2 protocol is already messed up. just tear down everything and make a new connection.
I'm open for suggestions.

mitmproxy/protocol/http.py
Kriechi: If the connect fails - we need some way of passing this up the call stack.
Returning True and False seems like a clean solution.

mitmproxy/protocol/http.py
mhils: Why don't we just raise a proper exception in `establish_server_connection` then?

mitmproxy/protocol/http2.py
```diff
@@ -410,9 +398,6 @@ def check_close_connection(self, flow):
         # RFC 7540 8.1: An HTTP request/response exchange fully consumes a single stream.
         return True
 
-    def connect(self):  # pragma: no cover
```
mhils: What's wrong with this line?

mitmproxy/protocol/http.py
Kriechi: We can do that. Raising or returning False will have the same effect - it needs to be handled. Right now, nobody cares if it fails.
Do you prefer an exception in this case?

mitmproxy/protocol/http2.py
Kriechi: Duplicated code - already implemented here: https://github.com/Kriechi/mitmproxy/blob/0148380209ef4f4ad0c6f03f385e03c36d90c2f3/mitmproxy/protocol/http2.py#L107

mhils: Ok, good catch. :+1:

PR #1066, opened 2016-03-31
Fix XSS vulnerability in HTTP errors

fimad:
The make_error_response method does not properly escape characters
that end up in the response body. Since the error code can contain
user supplied values this leads to a potential XSS vulnerability.

Example:

    echo '<script>alert(1)</script>' | nc localhost 8888


mhils:
Thanks. Do you see away how that could be exploited in practice?

fimad:
I'm not aware of a way to get this to trigger in the browser, but I also haven't looked super hard \\\_(ツ)\_/

PR #1069, opened 2016-04-02
Fix being unable to export binary data in requests

Uberi:
On Ubuntu 15.10 and Python 2 with PyGTK, when exporting a large amount of binary data (with the E shortcut), `pyperclip` will fail with a `Gtk.Clipboard.set_text() must be string without null bytes, not str`.

The failure then proceeds to crash mitmproxy, losing your session. This patch allows us to recover from this.


mhils:
Thanks and sorry for the slowish response! :smiley:

PR #1073, opened 2016-04-03
form_(in|out) -> first_line_format

mhils:
This PR replaces all (already deprecated) occurences of `Request.form_(in|out)` with `Request.first_line_format`.


Kriechi:
I'm not sure if the name `first_line_format` is a good fit - might be confusing for HTTP/2 use cases.

PR #1075, opened 2016-04-03
update dependencies

mhils:
AppVeyor builds are currently failing on master as the Pillow guys removed the 3.1 wheels from PyPi. :confused:


setup.py
```diff
@@ -69,9 +69,10 @@
         "Flask>=0.10.1, <0.11",
         "h2>=2.1.2, <3.0",
         "hpack>=2.1.0, <3.0",
-        "html2text==2016.1.8",
+        "html2text==2016.4.2",
```
Kriechi: Is this necessary? Can we loosen the restriction here?

PR #1079, opened 2016-04-11
Add installation instructions for Arch Linux

felixonmars:


mhils:
Thanks you so much for always keeping the arch packages in sync, @felixonmars! :smiley: 
I hope things got a bit easier for you with the combined mitmproxy/netlib/pathod package?

I'm happy to merge this PR. Just to be very clear right from the start, we don't want to provide installation instructions that result in outdated mitmproxy versions in the future. If you get hit by a bus and Arch versions fall a year behind, we may remove this. In other words, Ubuntu 12.04's mitmproxy package is terrible enough.

Thanks!

felixonmars:
Thank you, the combined package is indeed much easier :)

I'll always keep it up-to-date.

PR #1091, opened 2016-04-21
support for setting/sending multiple cookies

StevenVanAcker:
This patch adds support for using multiple cookies set by the same server, when using stickycookies.
The original code only remembered the last cookie set.

Testcase through: mitmdump -t "."
visit https://httpbin.org/cookies/set?helloWWW=worldZZZ&xxx=yyy
then flush cookies from your browser
and finally visit https://httpbin.org/cookies

### DIFF
--- a/mitmproxy/flow.py
+++ b/mitmproxy/flow.py
@@ -22,6 +22,7 @@ from .proxy.config import HostMatcher
 from .protocol.http_replay import RequestReplayThread
 from .exceptions import Kill
 from .models import ClientConnection, ServerConnection, HTTPFlow, HTTPRequest
+from collections import defaultdict
 
 
 class AppRegistry:
@@ -309,7 +310,7 @@ class StickyCookieState:
         """
             flt: Compiled filter.
         """
-        self.jar = {}
+        self.jar = defaultdict(dict)
         self.flt = flt
 
     def ckey(self, m, f):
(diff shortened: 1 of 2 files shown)


mitmproxy/flow.py
```diff
@@ -337,7 +337,9 @@ def handle_response(self, f):
             for m in c.values():
                 k = self.ckey(m, f)
                 if self.domain_match(f.request.host, k[0]):
-                    self.jar[k] = m
+                    if not k in self.jar:
+                        self.jar[k] = {}
```
mhils: Could we use `collections.defaultdict` instead? :smiley:

mitmproxy/flow.py
```diff
@@ -349,10 +351,10 @@ def handle_request(self, f):
                     f.request.path.startswith(i[2])
                 ]
                 if all(match):
-                    l.append(self.jar[i].output(header="").strip())
+                    l.extend([m.output(header="").strip() for (_, m) in self.jar[i].items()])
```
mhils: Very minor nitpick: Why not just `.values()`?

mhils:
Thanks for the excellent PR! :smiley:
It'd be great if you could add a small test at [test/mitmproxy/test_flow.py](https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/test_flow.py#L45), but we can take that over if you are busy with other things.

StevenVanAcker:
I've switched to defaultdict, simplified the list comprehension with .values() and squashed the new commits into one. I'll have a look at the test as well later today

mhils:
Awesome. Thanks so much for the great PR! :smiley: :cake: :tada:

PR #1097, opened 2016-04-26
Chrome/Chromium and certificate pinning

victorhooi:
Add note about Chrome/Chromium's behaviour around certificate pinning and the mitmproxy CA.

### DIFF
--- a/docs/certinstall.rst
+++ b/docs/certinstall.rst
@@ -113,6 +113,8 @@ accepted by these applications without modifying them. It is recommended to use 
 :ref:`passthrough` feature in order to prevent **mitmproxy** and **mitmdump** from intercepting
 traffic to these specific domains. If you want to intercept the pinned connections, you need to patch the application manually. For Android and (jailbroken) iOS devices, various tools exist to accomplish this.
 
+For Chrome/Chromium, as long as the mitmproxy CA is added to the *system* (or _public_) trust store, then pin validation on the certificate will *not* be performed (i.e. certificate pinning will not be enforced and mitmproxy will work as expected). Note that if you add the certificate to your personal user store, then it will be enforced, and mitmproxy will not work. There is more information on this on the `Chromium Security FAQ`_ (tl;dr this is intended behaviour).
+
 
 CA and cert files
 -----------------
@@ -193,3 +195,4 @@ directory and uses this as the client cert.
 
 
 .. _Certificate Pinning: http://security.stackexchange.com/questions/29988/what-is-certificate-pinning/
+.. _Chromium Security FAQ: https://www.chromium.org/Home/chromium-security/security-faq#TOC-How-does-key-pinning-interact-with-local-proxies-and-filters-


mhils:
Thanks for the PR. Which OS is your addition referring to?

victorhooi:
It should apply to all Chromium platforms (i.e. OSX, Windows, Android etc.), assuming you install the cert into the system trust store (and not a user trust store).

mhils:
Hm. I really appreciate the effort, but I think the current version is not really correct. From the page you linked (emphasis added by me):

> One trust anchor store is the system or public trust anchor store, and the other other is the local or private trust anchor store. The public store is provided as part of the operating system, and intended to authenticate public internet servers. The private store contains certificates installed by the user or the administrator of the client machine.
> [...]
> Chrome does not perform pin validation when the certificate chain chains up to a **private** trust anchor.

For mitmproxy, I think it would be sufficient to note that CA certificates which have been manually added are exempt from pinning, both in Firefox and Chrome. There is no way to install the mitmproxy CA so that we would still have pinning, is there?

victorhooi:
No, it is possible to install it wrongly.

That's why I filed this issue - because I was talking somebody through doing it on Windows, and it didn't work for them.

For example, on Windows you will see the following screen:

capture

By default, the first option "Automatically select the certificate store based on the type of certificate" is ticked.

If you go with the defaults - `mitmproxy` will not work - Chrome will outright block you from accessing any sites that use pinning (e.g. google.com, gmail.com etc.)

However, the correct approach is to select the second option, "Place all certificates in the following store", hit Browse, and then select "Trusted Root Certification Authorities".

I think at the very least we need to make it clear that if you install it in the wrong trust, then pinning will break mitmproxy - whether we choose to include instructions, I'm not sure, since it's obviously OS specific (although we do have OS specific steps in other parts of the doc, I believe).

mhils:
Thanks for the clarification! So, the problem you are describing is installing the certificate properly at all. My understanding is that it never works if you install it into a different store, is that correct?

victorhooi:
Yup - the issue is with ensuring the certificate is installed correctly.

If you follow the defaults, mitmproxy won't work.

Note that this is just how Chrome works - it's a intentional feature of Chrome, to allow things like filtering proxies to work.

Other browsers may operate differently, or may not have such a loophole available/exposed.

Pinning is applied in a number of ways - sometimes the signatures are shipped with the browser, sometimes via HSTS.

mhils:
According to https://stackoverflow.com/questions/30153690/windows-detect-certificate-as-root-certificate-authority, we cannot fix the defaults, so we should provide better guidance for users. I think we should just mention this at http://docs.mitmproxy.org/en/latest/certinstall.html#windows-automated and ideally on the onboarding app (when somebody clicks Windows) as well.

Unrelated to that, it would be interesting to pick up the original idea from this PR and note in the docs that manually installed CAs override certificate pinning in both Chrome and Firefox. I'm not very happy with the current wording of this PR, because it (1) contradicts what the Chrome page says (public vs private) and (2) Windows's concept of User/System trust stores has nothing to do with that. The certificate needs to be installed in the root store, both the user and the system root store work. I think we should rather just state that pinning is overridden and link to the corresponding docs.

mhils:
I'm closing this for now. If you would like to help us clarifying the certpinning as outlined above, please open a new PR! :smiley:

PR #1103, opened 2016-04-30
Include correct documentation URL in error message

snoblenet:


mhils:
Thanks! :smiley: :cake: :tada:

Issue #1108, opened 2016-05-01
No way to prevent hostname-from-client from being auto-added to X.509 SANs


javabrett:
As at current master (d84fc64), `mitmproxy` will in certain situations generate X.509 certificates for remote sites which are accepted without error by a web-browser, whereas a direct-connection to the same site would trigger TLS errors such as `SSL_ERROR_BAD_CERT_DOMAIN`, `NET::ERR_CERT_COMMON_NAME_INVALID`.  This certificate "enrichment", whilst nicely removing a browser-error seen by a direct-connection, makes `mitmproxy` less transparent, and may prevent or interfere with certain types of testing.  Currently this behaviour cannot be configured; the behaviour should be absent or at least configurable, and preferably non-default.

The error-nilling occurs when accessing a domain via a name which is not the `Subject` `CN`, and also does not match a `subjectAltName` (SAN).  `mitmproxy` currently adds the client-request hostname to the SANs, even when upstream certificate-sniffing is enabled (default), so `mitmproxy` can see that the client's requested hostname does not match, but it is added to the certificate regardless.  This occurs [here](https://github.com/javabrett/mitmproxy/blob/master/mitmproxy/protocol/tls.py#L569).

A good test-case is https://mitm.it/ , which is not-intended to be accessed over HTTPs, and when it is (directly), it serves the main `mitmproxy.org` website, but not before throwing a browser-error, because it is using `mitmproxy.org`'s certificate, which has no SAN for `mitm.it`.  If you `mitmproxy` this connection, the resulting generated certificate will have `mitm.it` as a SAN, and your browser-connection will be error-less.  Normally perhaps a nice thing in removing the error, but I didn't ask `mitmproxy` to do that, and it is sniffing upstream certs by-default, so I expect the cert to be more-or-less identical, other than the `Issuer` of course.

##### Steps to reproduce the problem:

1. Access https://mitm.it/ directly from your browser (no `mitmproxy`), observe CN mismatch warning and not mitm.it in CN or SAN.
2. Access https://mitm.it/ via `mitmproxy` and observe that `mitm.it` appears in the certificate SAN.

##### What is the expected behavior?

* When upstream cert sniffing is enabled, just pass-through the `Subject` `CN` and the `subjectAltName` unmodified by-default.  Consider some _non-default_ runtime options to:
[...]

PR #1109, opened 2016-05-02
Speed up TLS handshake if SNI is present

mhils:
This is an experimental PR that speeds up mitmproxy's TLS handshake. The assumption is as follows: If the client sends a SNI value in the handshake, we can be reasonably sure that this is actually the server the client is requesting. Thus, we can just mirror the SNI value in the cert and don't need to an upstream connection.

I really don't see a case where this could be failing. @cortesi, any concerns with this? If we all agree that this should be implemented, I'll adjust the tests.

### DIFF
--- a/mitmproxy/protocol/base.py
+++ b/mitmproxy/protocol/base.py
@@ -133,24 +133,15 @@ class ServerConnectionMixin(object):
                     "The proxy shall not connect to itself.".format(repr(address))
                 )
 
-    def set_server(self, address, server_tls=None, sni=None):
+    def set_server(self, address):
         """
         Sets a new server address. If there is an existing connection, it will be closed.
-
-        Raises:
-            ~mitmproxy.exceptions.ProtocolException:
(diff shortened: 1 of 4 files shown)


cortesi:
Assuming s/server/client in the second sentence of the description above, I think this is a useful speedup.

This does change the behaviour for server cert mismatches, in ways related to #1108. We should make a consistent choice here (I'm happy for this to be in favor of user convenience), and log a warning if there's a mismatch that can be detected later. 

We also need to tighten up the code in tls.py - I find it quite indirect and difficult to follow. There are many small methods with cascading side-effects that really should be side-effect-free.

mhils:
Ok, this is ready for review now.

cortesi:
Give me a couple of days on this one. There's an interesting possibility of using Netograph (which now has a constant stream of data flowing through it) as a test bed for mitmproxy. This change is a great candidate for that.

PR #1119, opened 2016-05-07
Python 3 - pathod.utils

dufferzafar:
Tests for the file `test/pathod/test_utils.py` pass with both Py2 & 3 interpreters.

### DIFF
--- a/pathod/language/base.py
+++ b/pathod/language/base.py
@@ -4,6 +4,7 @@ import abc
 import pyparsing as pp
 
 from six.moves import reduce
+from netlib.utils import escaped_str_to_bytes, bytes_to_escaped_str
 
 from .. import utils
 from . import generators, exceptions
@@ -108,7 +109,7 @@ class Token(object):
 class _TokValueLiteral(Token):
 
     def __init__(self, val):
-        self.val = val
+        self.val = escaped_str_to_bytes(str(val))
 
     def get_generator(self, settings_):
         return self.val
@@ -133,7 +134,7 @@ class TokValueLiteral(_TokValueLiteral):
         return v
 
     def spec(self):
-        inner = self.val.encode("string_escape")
+        inner = bytes_to_escaped_str(self.val)
         inner = inner.replace(r"\'", r"\x27")
         return "'" + inner + "'"
 
@@ -146,7 +147,7 @@ class TokValueNakedLiteral(_TokValueLiteral):
         return e.setParseAction(lambda x: cls(*x))
 
     def spec(self):
-        return self.val.encode("string_escape")
+        return bytes_to_escaped_str(self.val)
 
 
 class TokValueGenerate(Token):
@@ -164,7 +165,7 @@ class TokValueGenerate(Token):
 
     def freeze(self, settings):
         g = self.get_generator(settings)
-        return TokValueLiteral(g[:].encode("string_escape"))
+        return TokValueLiteral(bytes_to_escaped_str(g[:]))
 
     @classmethod
     def expr(cls):
@@ -224,7 +225,7 @@ class TokValueFile(Token):
         return generators.FileGenerator(s)
 
     def spec(self):
-        return "<'%s'" % self.path.encode("string_escape")
+        return "<'%s'" % bytes_to_escaped_str(self.path)
 
 
 TokValue = pp.MatchFirst(
@@ -576,4 +577,4 @@ class NestedMessage(Token):
 
     def freeze(self, settings):
         f = self.parsed.freeze(settings).spec()
-        return self.__class__(TokValueLiteral(f.encode("string_escape")))
+        return self.__class__(TokValueLiteral(bytes_to_escaped_str(f)))
--- a/pathod/language/base.py
+++ b/pathod/language/base.py
@@ -3,8 +3,11 @@ import os
 import abc
 import pyparsing as pp
 
+from six.moves import reduce
+
 from .. import utils
 from . import generators, exceptions
+
 
 class Settings(object):
 
@@ -105,7 +108,7 @@ class Token(object):
 class _TokValueLiteral(Token):
 
     def __init__(self, val):
-        self.val = val.decode("string_escape")
+        self.val = val
 
     def get_generator(self, settings_):
         return self.val
--- a/pathod/language/base.py
+++ b/pathod/language/base.py
@@ -109,7 +109,7 @@ class Token(object):
 class _TokValueLiteral(Token):
 
     def __init__(self, val):
-        self.val = escaped_str_to_bytes(str(val))
+        self.val = escaped_str_to_bytes(val)
 
     def get_generator(self, settings_):
         return self.val
--- a/pathod/utils.py
+++ b/pathod/utils.py
@@ -2,6 +2,8 @@ import os
 import sys
 import netlib.utils
 
+import six
+
 
 SIZE_UNITS = dict(
     b=1024 ** 0,
@@ -58,7 +60,7 @@ def inner_repr(s):
         Returns the inner portion of a string or unicode repr (i.e. without the
         quotes)
     """
-    if isinstance(s, unicode):
+    if six.PY2 and isinstance(s, unicode):
         return repr(s)[2:-1]
     else:
         return repr(s)[1:-1]
@@ -70,7 +72,10 @@ def escape_unprintables(s):
     """
     s = s.replace("\r\n", "PATHOD_MARKER_RN")
     s = s.replace("\n", "PATHOD_MARKER_N")
-    s = inner_repr(s)
+    if six.PY2:
+        s = inner_repr(s)
+    else:
+        s = s.encode('unicode_escape').decode('ascii')
     s = s.replace("PATHOD_MARKER_RN", "\n")
     s = s.replace("PATHOD_MARKER_N", "\n")
     return s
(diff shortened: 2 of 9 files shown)


pathod/language/base.py
```diff
@@ -105,7 +108,7 @@ def __repr__(self):
 class _TokValueLiteral(Token):
 
     def __init__(self, val):
-        self.val = val.decode("string_escape")
```
dufferzafar: Not too sure about this :/

pathod/utils.py
```diff
@@ -69,7 +71,10 @@ def escape_unprintables(s):
     """
     s = s.replace("\r\n", "PATHOD_MARKER_RN")
     s = s.replace("\n", "PATHOD_MARKER_N")
-    s = inner_repr(s)
+    if six.PY2:
+        s = inner_repr(s)
+    else:
+        s = s.encode('unicode_escape').decode('ascii')
```
dufferzafar: ...and this

pathod/language/base.py
mhils: We use string escape here so that users can supply escaped strings which are then treated as bytes - I'm afraid we can't just leave it out.

(I believe we can do that in a later PR though, this is not pathod.utils)

pathod/utils.py
```diff
@@ -57,7 +59,7 @@ def inner_repr(s):
         Returns the inner portion of a string or unicode repr (i.e. without the
         quotes)
     """
-    if isinstance(s, unicode):
+    if six.PY2 and isinstance(s, unicode):
```
mhils: Assuming this function accepts both bytes and unicode, this case should also cover the "we supplied bytes here in py3" case, which prefixes with `b`.

pathod/utils.py
mhils: Where is this used? Is the input bytes or unicode?

dufferzafar: `escape_unprintables` is used [here](https://github.com/dufferzafar/mitmproxy/blob/pathod-port/pathod/app.py#L169).

The input to it is bytes in Py2, and unicode in Py3.

mhils: So, in the code you linked, `s` is the thing pathod sends out on the wire,
which we escape here to preview it to the user. On the wire = bytes, so I
think `s` should always be bytes/BytesIO in the first place. Let's make
this bytes-only (and add a docstring indicating so)!

pathod/utils.py
```diff
@@ -59,7 +59,8 @@ def inner_repr(s):
         Returns the inner portion of a string or unicode repr (i.e. without the
         quotes)
     """
-    if six.PY2 and isinstance(s, unicode):
+    if (six.PY2 and isinstance(s, unicode)) or \
+            (six.PY3 and isinstance(s, bytes)):
```
Kriechi: using `six.PY3` is usually not a good idea, because this might break if Python 4 gets released. We should assume future-compatibility and only check for legacy-incompatibilities.

dufferzafar:
I've made the changes suggested. Please have a look again.

Kriechi:
@dufferzafar would be great if the tests still pass on python2:
https://travis-ci.org/mitmproxy/mitmproxy/jobs/129256183#L456

pathod/utils.py
```diff
@@ -57,7 +59,7 @@ def inner_repr(s):
         Returns the inner portion of a string or unicode repr (i.e. without the
         quotes)
     """
-    if isinstance(s, unicode):
+    if (six.PY2 and isinstance(s, unicode)) or isinstance(s, bytes):
```
mhils: on Python 2, `bytes` == `str`, so this will always cut off the first character.

mhils:
Thanks @dufferzafar! :-)

https://github.com/mitmproxy/mitmproxy/pull/1119/files#r62433803 <- can we fix this as well?

mhils:
Update on this: I think you can remove `inner_repr` entirely and replace it with `netlib.bytes_to_escaped_str` which we needed in mitmproxy. :smiley:

dufferzafar:
I'm sorry this is taking so much time. I'm having a hard time figuring out this string encoding business and as I am in the middle of my semester exams, I can't devote this as much time as I'd like to. :confused:

With the `inner_repr` thing, I did remove it, but couldn't figure out how to make the `escape_unprintables` handle newline characters correctly. 

I've made updates to the `TokValueLiteral` stuff though, please have a look!

mhils:
No problem - finals come first. :smiley:
Shall we just have a short chat once you have more time available?

Other than that, let's make sure that this PR does not get out of hand, let's go for pathod.utils only. :wink:

pathod/language/base.py
```diff
@@ -105,7 +109,7 @@ def __repr__(self):
...
+        self.val = escaped_str_to_bytes(str(val))
```
mhils: find out what happens here?

pathod/utils.py
```diff
@@ -69,7 +71,10 @@ def escape_unprintables(s):
     """
     s = s.replace("\r\n", "PATHOD_MARKER_RN")
     s = s.replace("\n", "PATHOD_MARKER_N")
```
mhils: use `b"...l"`

pathod/utils.py
```diff
@@ -57,7 +59,7 @@ def inner_repr(s):
         Returns the inner portion of a string or unicode repr (i.e. without the
         quotes)
     """
-    if isinstance(s, unicode):
```
mhils: remove unused function

pathod/language/base.py
dufferzafar: These sort of errors come up: https://travis-ci.org/mitmproxy/mitmproxy/jobs/132823626#L460-L516

The function receives `unicode` values as input, but I can't figure out the source.

dufferzafar:
The `test_roundtrip` function is failing too. https://travis-ci.org/mitmproxy/mitmproxy/jobs/132823626#L6336-L6356

I think the failure is realted to escaping of quotes, as [this](https://github.com/dufferzafar/mitmproxy/blob/pathod-port/test/pathod/test_language_base.py#L71) test case pass, but the others don't.

dufferzafar:
The failure in `test_nested_response_freeze` is possibly related to the one in roundtrip. https://travis-ci.org/mitmproxy/mitmproxy/jobs/132823626#L7299-L7332

That's all I could figure out about it :/

Kriechi:
This is a bit tricky. Take a look here about the different str/unicode/byte data types:
https://pythonhosted.org/six/#six.string_types
https://pythonhosted.org/six/#six.text_type
https://pythonhosted.org/six/#six.binary_type

I propose this change, which fixes almost all test failures (except 2):
```diff
diff --git a/netlib/utils.py b/netlib/utils.py
index 7499f71..7b6eead 100644
--- a/netlib/utils.py
+++ b/netlib/utils.py
@@ -430,8 +430,8 @@ def bytes_to_escaped_str(data):
     # One way to do would be calling .decode("utf8", "backslashreplace") first
     # and then escaping UTF8 control chars (see clean_bin).

-    if not isinstance(data, bytes):
-        raise ValueError("data must be bytes")
+    if not isinstance(data, six.binary_type):
+        raise ValueError("data must be six.binary_type, but is {}".format(data.__class__.__name__))
     return repr(data).lstrip("b")[1:-1]


@@ -439,8 +439,8 @@ def escaped_str_to_bytes(data):
     """
     Take an escaped string and return the unescaped bytes equivalent.
     """
-    if not isinstance(data, str):
-        raise ValueError("data must be str")
+    if not isinstance(data, six.string_types):
+        raise ValueError("data must be six.text_type, but is {}".format(data.__class__.__name__))

     if six.PY2:
         return data.decode("string-escape")
```
@mhils what do you think about this?

test/pathod/test_utils.py
```diff
@@ -27,13 +29,8 @@ def test_data_path():
...
 
-def test_inner_repr():
-    assert utils.inner_repr("\x66") == "\x66"
-    assert utils.inner_repr(u"foo") == "foo"
-
-
 def test_escape_unprintables():
     s = "".join([chr(i) for i in range(255)])
-    e = utils.escape_unprintables(s)
+    e = utils.escape_unprintables(six.b(s))
```
mhils: This probably works well, but we should adopt a strict "Python 3 by default" policy. That is, once we have full Python 3 compatibility, we should just be able to delete everything releated to `six`. How about this?

```python
s = bytes(range(256))
if six.PY2:
    s = "".join([chr(i) for i in range(255)])
```

mhils:
@Kriechi: Regarding the first change,

Python 2:
```python
>>> import six
>>> six.binary_type == bytes
True
>>> 
```

Python 3:
```python
>>> import six
>>> six.binary_type == bytes
True
>>> 
```

We can just use `bytes`, which IIRC is only unsupported on Python 2.6, which we don't support anyway.

mhils:
@Kriechi: Regarding the second change, 7a8da48a306dfc8e43239d7f2a141c465e40ab77 introduces support for unicode input on Python 2. I restarted the tests, let's see where this goes. :smiley:

mhils:
Alright, d3477e27fa4ffbcfa9c1b9aa937d1d54448cc597 makes the test lights go green. :relaxed: 

- [ ] Fix https://github.com/mitmproxy/mitmproxy/pull/1119#discussion_r64668942
- [ ] Add pathod.utils to `.travis.yml`
- [ ] Continue on a separate branch so that we can merge this :smiley:

dufferzafar:
The failing test seems unrelated: https://travis-ci.org/mitmproxy/mitmproxy/jobs/133181308#L359

mhils:
:tada: :cake:

Kriechi:
Yeah! Good start! 🎉 
Looking forward to your next PR already!

PR #1123, opened 2016-05-09
Fix #1120

### LINKED ISSUE #1120: Wrong URL displayed for OPTIONS * requests
Launch `mitmproxy`, then use the following Python code to send an OPTIONS * request through the tunnel:

```
import socket, ssl

sock = socket.create_connection(('localhost', 8080))
sock.send(b'CONNECT http2bin.org:443 HTTP/1.1\r\n'
          b'Host: http2bin.org:443\r\n'
          b'User-Agent: demo\r\n'
          b'\r\n')
print(sock.recv(4096))

context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
sock = context.wrap_socket(sock, server_hostname='http2bin.org')
sock.send(b'OPTIONS * HTTP/1.1\r\n'
          b'Host: http2bin.org\r\n'
          b'User-Agent: demo\r\n'
          b'Content-Length: 0\r\n'
          b'\r\n')
print(sock.recv(4096))
```

mitmproxy displays the URL of the request as: `https://http2bin.org*`

The correct URL of the request is `https://http2bin.org` ([RFC 7230 § 5.5](https://tools.ietf.org/html/rfc7230#section-5.5)).

Also, in the inline scripts API, `flow.request.path` returns `*`, although the [docs say](http://docs.mitmproxy.org/en/latest/dev/models.html#netlib.http.Request.path) that it is “guaranteed to start with a slash.” But I think this is a documentation problem.

mitmproxy master, Linux.

mhils:
Fixing #1120 turned out to be surprisingly difficult from a UI/UX perspective. The key problem is that if we just display `OPTIONS *`, a user cannot associate the request with a host in the flow list:

```
λ mitmdump
127.0.0.1:52407: clientconnect
127.0.0.1 OPTIONS *                   <- this goes where?
 << 200 OK 0B
```

To fix this, I'd like to display `OPTIONS * (example.com)` in the UI. There are two ways to accomplish this:

1. Let `request.url` return the stuff in the parantheses as well.(https://github.com/mitmproxy/mitmproxy/compare/issue-1120)
2. Manually add them in the UI part. (https://github.com/mitmproxy/mitmproxy/compare/issue-1120-alt)

I personally prefer the second solution even though it is way more verbose code-wise. Thoughts?

### DIFF
--- a/mitmproxy/console/common.py
+++ b/mitmproxy/console/common.py
@@ -401,6 +401,15 @@ flowcache = utils.LRUCache(800)
 
 
 def format_flow(f, focus, extended=False, hostheader=False, marked=False):
+    if hostheader:
+        url = f.request.pretty_url
+        if url == "*":
+            url = "* ({})".format(f.request.pretty_host)
+    else:
+        url = f.request.url
+        if url == "*":
+            url = "* ({})".format(f.request.host)
+
     d = dict(
         intercepted = f.intercepted,
         acked = f.reply.acked,
@@ -408,7 +417,7 @@ def format_flow(f, focus, extended=False, hostheader=False, marked=False):
         req_timestamp = f.request.timestamp_start,
         req_is_replay = f.request.is_replay,
         req_method = f.request.method,
-        req_url = f.request.pretty_url if hostheader else f.request.url,
+        req_url = url,
         req_http_version = f.request.http_version,
 
         err_msg = f.error.msg if f.error else None,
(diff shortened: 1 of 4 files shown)


vfaronov:
I think `request.url` should ideally be `http://example.com` because that is the actual request URI of such a request.

mhils:
Looking at https://tools.ietf.org/html/rfc7230#section-5.3.4: You are absolutely right! Thanks for that pointer.

mhils:
Ok, that was by all means the superior solution. Fix is on master, thanks!

PR #1125, opened 2016-05-10
sanitize strings with shell control characters

Kriechi:
@mhils: anything else we should sanitize?

### DIFF
--- a/mitmproxy/dump.py
+++ b/mitmproxy/dump.py
@@ -174,8 +174,8 @@ class DumpMaster(flow.FlowMaster):
         if self.o.flow_detail >= 2:
             headers = "\r\n".join(
                 "{}: {}".format(
-                    click.style(k, fg="blue", bold=True),
-                    click.style(v, fg="blue"))
+                    click.style(netlib.utils.bin_safe(k), fg="blue", bold=True),
+                    click.style(netlib.utils.bin_safe(v), fg="blue"))
                 for k, v in message.headers.fields
             )
             self.echo(headers, indent=4)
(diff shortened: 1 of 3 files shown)


mitmproxy/dump.py
```diff
@@ -295,7 +296,7 @@ def _echo_response_line(self, flow):
             replay=replay,
             arrows=arrows,
             code=code,
-            reason=reason,
+            reason=shlex_quote(reason),
```
mhils: we need to apply this before `click.style`

Kriechi: I changed it - please take another look.

netlib/utils.py
```diff
@@ -431,3 +430,16 @@ def safe_subn(pattern, repl, target, *args, **kwargs):
         need a better solution that is aware of the actual content ecoding.
     """
     return re.subn(str(pattern), str(repl), target, *args, **kwargs)
+
+def dequote(s):
```
mhils: This is very similar to pathod's `inner_repr`. Can we combine these?
(cc @dufferzafar)

Also, tests please.

netlib/utils.py
```diff
@@ -431,3 +431,6 @@ def safe_subn(pattern, repl, target, *args, **kwargs):
         need a better solution that is aware of the actual content ecoding.
     """
     return re.subn(str(pattern), str(repl), target, *args, **kwargs)
+
+def bin_safe(s):
+    return ''.join(["\\x{:02x}".format(ord(i)) if ord(i) < 32 else i for i in s])
```
mhils: This looks very much like `repr` now (we miss 127 though) - maybe just do `repr(x)` + strip outer parantheses?

Kriechi: repr also might add `u` and quotes and stuff. I want keep the string as much as possible to the original version. Not sure how this all changes with bytes vs. str vs. unicode...

mhils: We need something similar for pathod as well. I'll work on it.

PR #1126, opened 2016-05-10
Safeguard

### LINKED ISSUE #1084: HTTP/2 does not handle killed requests

1. `mitmdump -S test_replay --kill`
2. Open a website that is not contained in the replay file.

##### What went wrong?

```
127.0.0.1:50907: clientconnect
127.0.0.1:50906: clientconnect
127.0.0.1 GET https://news.ycombinator.com/ HTTP/2.0
 << Connection killed
Exception in thread Thread-2:
  File "c:\python27\Lib\threading.py", line 810, in __bootstrap_inner
    self.run()
  File "c:\users\user\git\mitmproxy\mitmproxy\protocol\http2.py", line 422, in run
    layer()
  File "c:\users\user\git\mitmproxy\mitmproxy\protocol\http.py", line 188, in __call__
  File "c:\users\user\git\mitmproxy\mitmproxy\protocol\http.py", line 356, in process_request_hook
    raise Kill()
Kill
```

Kriechi:
prevent backtrace with lost http2 connections
might fix #1084

### DIFF
--- a/mitmproxy/protocol/http.py
+++ b/mitmproxy/protocol/http.py
@@ -11,7 +11,7 @@ from netlib.http import Headers
 from h2.exceptions import H2Error
 
 from .. import utils
-from ..exceptions import HttpProtocolException, ProtocolException
+from ..exceptions import HttpProtocolException, Http2ProtocolException, ProtocolException
 from ..models import (
     HTTPFlow,
     HTTPResponse,
@@ -243,7 +243,7 @@ class HttpLayer(Layer):
         try:
             response = make_error_response(code, message)
             self.send_response(response)
-        except (NetlibException, H2Error):
+        except (NetlibException, H2Error, Http2ProtocolException):
             self.log(traceback.format_exc(), "debug")
 
     def change_upstream_proxy_server(self, address):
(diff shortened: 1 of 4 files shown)


mitmproxy/protocol/http.py
```diff
@@ -359,7 +364,10 @@ def establish_server_connection(self, flow):
                 self.set_server(address, tls, address.host)
             # Establish connection is neccessary.
             if not self.server_conn:
-                self.connect()
```
mhils: We need to fail if there is an error with the server connection. Just returning False is not an option I think.

Kriechi: How hard would you want it to fail? Throwing an exception?

mhils: Yes. We should talk about this later or tomorrow, I actually disagree with all changes to `http.py`. :innocent:

Kriechi:
ping @mhils

mitmproxy/protocol/http.py
```diff
@@ -304,6 +304,11 @@ def get_response():
             # > server detects timeout, disconnects
             # > read (100-n)% of large request
             # > send large request upstream
+
+            if "Http2Layer" in [repr(l) for l in self.layers]:
```
mhils: Should we rather check for Http2Exception here?

mhils:
Other than my comment above (and the merge conflict), this looks awesome. :+1: :smiley:

Kriechi:
Done.

mhils:
:tada:

PR #1130, opened 2016-05-11
added UI for options menu

cle1000:


web/src/js/components/common.js
```diff
@@ -131,4 +131,26 @@ export var Splitter = React.createClass({
...
+
+export class ToggleComponent extends React.Component{
+  render(){
+    let toggleIconClass = this.props.checked ? "fa-check-square-o" : "fa-square-o";
+    let toggleButtonClass = this.props.checked ? "btn-primary" : "btn-default";
+    return (
+    <div
+      className={"btn-option btn " + toggleButtonClass}
+      onClick={this.props.onToggleChanged}>
+      <i className={"fa " + toggleIconClass}>
```
mhils: `<i>` should not contain the test.

web/src/js/components/common.js
```diff
@@ -131,4 +131,26 @@ export var Splitter = React.createClass({
             </div>
         );
     }
-});
\ No newline at end of file
+});
+
+export class ToggleComponent extends React.Component{
```
mhils: This component is great becase it does not have any state. We can just make it a pure component: https://facebook.github.io/react/docs/reusable-components.html#stateless-functions :smiley:

web/src/js/components/header.js
```diff
@@ -227,32 +227,74 @@ var ViewMenu = React.createClass({
...
+        <div>
+          {this.state.options.map(function(entry, i) {
+            return (
+              <ToggleComponent
+                key={i}
+                checked={entry.checked}
+                name = {entry.name}
+                onToggleChanged={() => this.setOption(entry)}/>
+            );
+          }.bind(this))}
```
mhils: Just make this an arrow function ( ` () => { ... } `, then we don't need to bind here.

web/src/js/components/common.js
```diff
@@ -131,4 +131,26 @@ export var Splitter = React.createClass({
...
+          {this.props.name}
+        </span>
+      </i>
+      {(this.props.icon) ? <i className={this.props.icon}/> : false}
+    </div>);}
+}
+ToggleComponent.propTypes = {
+    name: React.PropTypes.string.isRequired,
+    onToggleChanged: React.PropTypes.func.isRequired,
+    icon: React.PropTypes.string
```
mhils: I think it makes sense to just remove the `icon`, what do you think? :-)

mitmproxy/web/static/app.css
```diff
@@ -163,6 +163,12 @@ header .menu {
   max-height: 500px;
   overflow-y: auto;
 }
+.btn-option {
+  margin: 2px 2px 2px 2px;
+}
+.text-padding {
+  padding: 0px 5px 0px 5px;
+}
```
mhils: We should make sure that we don't have those super-generic CSS classes. I think it's just fine to have this specific, e.g. `.menu .btn`.

web/src/js/components/header.js
```diff
@@ -227,32 +227,74 @@ var ViewMenu = React.createClass({
...
+          {name: "--no-upstream-cert", checked: false},
+          {name: "--http2", checked: false},
+          {name: "--anticache", checked: false},
+          {name: "--anticomp", checked: false},
+          {name: "--stickycookie", checked: true},
+          {name: "--stickyauth", checked: false},
+          {name: "--stream", checked: false}
+        ]
+      }
+    }
```
mhils: I think the set of options we have is quite fixed. No need to put that in the state. :-)

web/src/js/components/common.js
cle1000: i did this because then the text is in one line with the icon, otherwise there is a little ugly space gap

web/src/js/components/common.js
cle1000: yes i added it because the event log has an icon but i remove it :+1:

web/src/js/components/header.js
cle1000: i dont get it :) can you give me more details here?

web/src/js/components/header.js
cle1000: i put options into the state because then react js handle the propagation of the state change to all children and ensures the rerendering.

web/src/js/components/common.js
mhils: This should just work fine?
http://jsbin.com/saruxaxoru/edit?html,output

web/src/js/components/header.js
mhils: Instead of 
```javascript
function(foo){ /* ... */ }.bind(this)
```
use ES6 arrow-style functions:
```javascript
(foo) => { /* ... */ }

```

web/src/js/components/header.js
mhils: ah yes - sorry, was too quick here. So, the "checked" state will ultimately come from the outside (currently there's the settings store, ultimately that'll be redux). In it's final form, this thing should not have state, right now that's fine. Mabye add a comment that we want to remove state here. :-)

cle1000:
i have problems with the conflicts. 
Step 1: From your project repository, check out a new branch and test the changes.

git checkout -b cle1000-master master
git pull https://github.com/cle1000/mitmproxy.git master
Step 2: Merge the changes and update on GitHub.

git checkout master
git merge --no-ff cle1000-master
git push origin master

i tried that but doesn't work, any ideas?

mitmproxy/web/static/app.css
```diff
@@ -163,6 +163,9 @@ header .menu {
   max-height: 500px;
   overflow-y: auto;
 }
+.btn {
+  margin: 2px 2px 2px 2px;
```
mhils: `.btn` is a generic bootstrap class - we should not override that.

mhils:
Uhm - I would just create a patch file, check out mitmproxy/master and then
apply your patch file.

cle1000:
i cannot found a patch file :(

PR #1133, opened 2016-05-11
update and move version information

Kriechi:
With this mitmproxy sets the correct headers: e.g. `Server: mitmproxy 0.18`, instead of using the netlib identifier.

### DIFF
--- a/pathod/version.py
+++ b/pathod/version.py
@@ -1,3 +1,3 @@
 from __future__ import (absolute_import, print_function, division)
 
-from netlib.version import *
+from mitmproxy.version import *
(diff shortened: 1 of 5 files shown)


pathod/version.py
```diff
@@ -1,3 +1,3 @@
 from __future__ import (absolute_import, print_function, division)
 
-from netlib.version import *
\ No newline at end of file
+from mitmproxy.version import *
```
mhils: I like the idea, but we don't want a dependency on mitmproxy in pathod.

mhils: (The pathod binaries don't include mitmproxy and its dependencies)

Kriechi: What is the release plan for pathod? I though it got integrated into mitmproxy and will be release as part of the mitmproxy wheel?

PR #1135, opened 2016-05-11
Sanitize Print

mhils:
This is a follow-up on #1125.

### DIFF
--- a/netlib/utils.py
+++ b/netlib/utils.py
@@ -431,3 +431,6 @@ def safe_subn(pattern, repl, target, *args, **kwargs):
         need a better solution that is aware of the actual content ecoding.
     """
     return re.subn(str(pattern), str(repl), target, *args, **kwargs)
+
+def bin_safe(s):
+    return ''.join(["\\x{:02x}".format(ord(i)) if ord(i) < 32 else i for i in s])
--- a/netlib/utils.py
+++ b/netlib/utils.py
@@ -432,5 +432,26 @@ def safe_subn(pattern, repl, target, *args, **kwargs):
     """
     return re.subn(str(pattern), str(repl), target, *args, **kwargs)
 
-def bin_safe(s):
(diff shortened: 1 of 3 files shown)


netlib/utils.py
```diff
@@ -432,5 +432,26 @@ def safe_subn(pattern, repl, target, *args, **kwargs):
...
 
-def bin_safe(s):
-    return ''.join(["\\x{:02x}".format(ord(i)) if ord(i) < 32 else i for i in s])
+
+def bytes_to_escaped_str(data):
+    """
+    Take bytes and return a safe string that can be displayed to the user.
+    """
+    if not isinstance(data, bytes):
+        raise ValueError("data must be bytes")
```
Kriechi: I don't think this helps us, since header keys and values are strings (most likely UTF8 in the case of HTTP/2 coming from hpack).

netlib/utils.py
```diff
@@ -432,5 +432,26 @@ def safe_subn(pattern, repl, target, *args, **kwargs):
...
+    return repr(data).lstrip("b")[1:-1]
```
Kriechi: This also encodes UTF8 multi-byte characters into hex. But we actually want those to remain untouched so the terminal renders them properly as what they are (`üäö` instead of some weird hex characters).
We only want to sanitise control characters (<32).

netlib/utils.py
mhils: Did you take a look at the latest changes in hyper/hpack? We'll get bytes from there. Whatsoever, our internal headers should *always* be `bytes`.

Kriechi: They are **not** bytes. At least not in `dump.py` when they get printed (and wrapped with color etc.).

Kriechi: hyper/hpack still defaults to UTF8 - so no change there.

PR #1136, opened 2016-05-11
Adding example which turns every reponse into an HTTP 500.

cellularmitosis:
I love using mitmproxy to easily reproduce network failures against API's.  I think this example will help raise awareness of how easy it can be to use mitmproxy in this way.


mhils:
I like the idea, thanks! :smiley:

LGTM once `content` is `b""` - `None` as a different semantic meaning to us (content still unknown or stripped off to save space - mitmproxy may fail in some cases when it should print a 500.

cellularmitosis:
Thanks @mhils.  I pushed another commit -- I'm fairly new to the pull-request workflow.  Is that all I need to do?

mhils:
All good, thanks! :smiley:

PR #1141, opened 2016-05-14
Send URL to stderr when starting mitmweb

jjones646:
Nothing major, just makes it a bit easier for opening it up in the browser.


mhils:
Thanks!

We'll certainly improve that even more when mitmweb gets officially supported, but this message is probably very helpful for now. :smiley:

PR #1166, opened 2016-05-23
Flow module

mhils:
`mitmproxy/flow.py` currently clocks at 1,200 lines and I'm afraid that I'll get lost in it at some point and won't find a way out... :smirk: 

This is an initial low-effort PR to start a discussion on how we could reorganize things. My initial proposal would be to split flow.py into a `flow` module:

1. Split flow.py into `mitmproxy.flow.`...
   - `io` (`FlowWriter, FilteredFlowWriter, FlowReader, read_flows_from_paths`) 
   - `master` (`FlowMaster`)
   - `modules/extensions/???` (`AppRegistry, ReplaceHooks, SetHeaders, StreamLargeBodies, ClientPlaybackState, ServerPlaybackState, StickyCookieState, StickyAuthState`)
   - `state` (`State/FlowState, FlowStore, FlowView`)
2. Move `mitmproxy.flow_export` into `mitmproxy.flow.export`
3. Move `mitmproxy.flow_format_compat` into `mitmproxy.flow.compat` (or `mitmproxy.flow.io`?)
4. Use `mitmproxy/flow/__init__.py` to provide a reasonable interface (as with [`netlib.http`](https://github.com/mitmproxy/mitmproxy/blob/master/netlib/http/__init__.py))

@mitmproxy/devs, what do you think?


Kriechi:
Good idea!
I'm a bit afraid git might loose track of the history if we move things around in such big chunks. But I'm willing to pay the price for better maintainability.

cortesi:
This looks good to me. One comment on this particular patch that goes for the codebase generally is that we way, way over use ```from X import Y`` constructions, which is now beginning to impact code legibility.

PR #1167, opened 2016-05-24
implement a toggle for viewing marked flows only in console

xntrik:
Hi All, I've been using this capability a bit when wrangling a bit of traffic in mitmproxy. It's not implemented as a filter, as the 'marking' of flows is handled in the ConsoleState. The 'M' command will quickly change the view to either marked flows or not. If there was an existing filter in place, this will be temporarily stored, and then replaced.

I'm unsure if this is the 'best' way to do this (don't hack much in Python), but it seems to be working for me fine.

### DIFF
--- a/mitmproxy/console/__init__.py
+++ b/mitmproxy/console/__init__.py
@@ -33,6 +33,8 @@ class ConsoleState(flow.State):
         self.default_body_view = contentviews.get("Auto")
         self.flowsettings = weakref.WeakKeyDictionary()
         self.last_search = None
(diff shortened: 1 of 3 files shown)


cortesi:
Very handy! At some point we'll have to have a more systematic measure to save and store state like the last filter, but for now this is sympathetic with the rest of the codebase. Apologies for the long time it took to merge this - we've had a bunch of stuff happen simultaneously in the codebase in the last while. :)

PR #1168, opened 2016-05-25
add a py3 venv in dev environment

Kriechi:
This allows us to run e.g. tests in a different python version:
$ venv3/bin/py.test test/netlib/test_tcp.py


cortesi:
Nice!

PR #1169, opened 2016-05-27
[web] Eliminate Router mixin and RawMixin

gzzhanghao:
This PR aims to eliminate Router mixin and RawMixin with props and higher order function.

Router.updateLocation method was temporarily moved to ProxyApp component and will be moved to actions once react-redux is ready.


mhils:
Great job! :smiley: :tada: :cake:

PR #1170, opened 2016-05-27
Python 3 - pathod.app

dufferzafar:
Fixing errors that came up when trying to make the `test/pathod/test_app.py` tests pass with Py3.

### DIFF
--- a/pathod/log.py
+++ b/pathod/log.py
@@ -63,7 +63,7 @@ class LogCtx(object):
             for line in strutils.hexdump(data):
                 self("\t%s %s %s" % line)
         else:
-            for i in strutils.clean_bin(data).split("\n"):
+            for i in strutils.clean_bin(data).split(b"\n"):
                 self("\t%s" % i)
 
     def __call__(self, line):
--- a/pathod/pathod.py
+++ b/pathod/pathod.py
@@ -5,7 +5,7 @@ import os
 import sys
 import threading
 
-from six.moves.urllib.parse import unquote
+from six.moves import urllib
 
 from netlib import tcp, certutils, websockets
 from netlib.exceptions import HttpException, HttpReadDisconnect, TcpTimeout, TcpDisconnect, \
@@ -184,7 +184,7 @@ class PathodHandler(tcp.BaseHandler):
                     break
             else:
                 if m(path.startswith(self.server.craftanchor)):
-                    spec = unquote(path)[len(self.server.craftanchor):]
+                    spec = urllib.parse.unquote(path)[len(self.server.craftanchor):]
                     if spec:
                         try:
                             anchor_gen = language.parse_pathod(spec, self.use_http2)
--- a/pathod/pathod.py
+++ b/pathod/pathod.py
@@ -204,7 +204,7 @@ class PathodHandler(tcp.BaseHandler):
                         )])
 
             if anchor_gen:
-                spec = anchor_gen.next()
+                spec = next(anchor_gen)
 
                 if self.use_http2 and isinstance(spec, language.http2.Response):
                     spec.stream_id = req.stream_id
--- a/pathod/pathod.py
+++ b/pathod/pathod.py
@@ -4,7 +4,8 @@ import logging
 import os
 import sys
 import threading
-import urllib
+
+from six.moves.urllib.parse import unquote
 
 from netlib import tcp, certutils, websockets
 from netlib.exceptions import HttpException, HttpReadDisconnect, TcpTimeout, TcpDisconnect, \
@@ -183,7 +184,7 @@ class PathodHandler(tcp.BaseHandler):
                     break
             else:
                 if m(path.startswith(self.server.craftanchor)):
-                    spec = urllib.unquote(path)[len(self.server.craftanchor):]
+                    spec = unquote(path)[len(self.server.craftanchor):]
                     if spec:
                         try:
                             anchor_gen = language.parse_pathod(spec, self.use_http2)
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,9 @@ matrix:
       git:
         depth: 9999999
     - python: 3.5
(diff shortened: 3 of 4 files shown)


pathod/language/__init__.py
```diff
@@ -26,7 +26,7 @@ def parse_pathod(s, use_http2=False):
         May raise ParseException
     """
     try:
-        s = s.decode("ascii")
+        s.encode("ascii")
```
dufferzafar: We're only checking whether the string is valid ASCII here, so `encode` would be better since it works on both Python versions. Right?

dufferzafar:
After making these changes, the tests in files `test_patho{c,d}_cmdline.py` pass too. Should I just enable Travis for them in this pull only?

pathod/language/__init__.py
mhils: :+1:

mhils:
Go for it! :smiley:

dufferzafar:
I'm stuck on the currently failing tests so can you please review and provide some hints as to how I can fix them?

Kriechi:
@dufferzafar please rebase to current master - there are a few changes there which might help debug this. Thanks.

Kriechi:
Looks like you need to properly encode `PathodHandler.settings`.
These settings get passed as `str` to `crafted.freeze(self.settings)`
and from there on it bubbles through multiple function, to finally explode in `pathod/language/base.py:169` at `return TokValueLiteral(bytes_to_escaped_str(g[:]))`.

dufferzafar:
Correct me if I'm wrong, but I think the problem is in the `freeze` functions not returning correct type of data. The same types of errors also seem to be occurring in [`pathod.language.base`](https://github.com/mitmproxy/mitmproxy/pull/1180).

Kriechi:
Yes - looks reasonable. Can you try and make the changes? If this is to complex, we might need to explore other approaches.

mhils:
@dufferzafar: `.encode()` works for all the `string.*` values, for `bytes` we can use https://github.com/mitmproxy/mitmproxy/blob/a7abf8b731658b4e7ed8705f7a94a6a62f08d51d/test/pathod/test_utils.py#L26-L28 again.

pathod/language/generators.py
```diff
@@ -55,12 +55,12 @@ def __len__(self):
         return self.length
 
     def __getitem__(self, x):
-        return random.choice(DATATYPES[self.dtype])
+        return random.choice(DATATYPES[self.dtype]).encode()
```
dufferzafar: @mhils I did this, instead of encoding values in the DATAYPES dict?

dufferzafar:
Still can't make sense of the failing tests :'(

Kriechi:
So I guess the test failures happen due to the tracebacks where one of the threads just silently fails (without raising a real error, but just printing the traceback and exiting cleanly).

I think this is the core problem:
```
  File "/home/travis/build/mitmproxy/mitmproxy/pathod/pathod.py", line 110, in http_serve_crafted
    self.settings
  File "/home/travis/build/mitmproxy/mitmproxy/pathod/language/__init__.py", line 119, in serve
    ret.update(msg.log(settings))
  File "/home/travis/build/mitmproxy/mitmproxy/pathod/language/message.py", line 82, in log
    v = [x[:LOG_TRUNCATE] for x in v.values(settings)]
  File "/home/travis/build/mitmproxy/mitmproxy/pathod/language/generators.py", line 63, in __getslice__
    return "".join(random.choice(chars).encode() for x in range(a, b))
  File "/home/travis/build/mitmproxy/mitmproxy/pathod/language/generators.py", line 63, in <genexpr>
    return "".join(random.choice(chars).encode() for x in range(a, b))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xf3 in position 0: ordinal not in range(128)
```

pathod/language/generators.py
mhils: We could, but I think we should tackle the problem at the source and fix DATATYPES directly.

pathod/pathod.py
```diff
@@ -4,7 +4,8 @@
 import os
 import sys
 import threading
-import urllib
+
+from six.moves.urllib.parse import unquote
```
mhils: @cortesi and I discussed yesterday that we do not want to import individual functions (as per https://google.github.io/styleguide/pyguide.html). Can we change this to `from six.moves import urllib`?

pathod/language/generators.py
```diff
@@ -55,12 +55,12 @@ def __len__(self):
...
 
     def __getslice__(self, a, b):
         b = min(b, self.length)
         chars = DATATYPES[self.dtype]
-        return "".join(random.choice(chars) for x in range(a, b))
+        return "".join(random.choice(chars).encode() for x in range(a, b))
```
mhils: I think you want `b"".join`

pathod/log.py
```diff
@@ -65,7 +65,7 @@ def dump(self, data, hexdump):
             for line in netlib.utils.hexdump(data):
                 self("\t%s %s %s" % line)
         else:
-            for i in netlib.utils.clean_bin(data).split("\n"):
+            for i in netlib.utils.clean_bin(data).split(b"\n"):
                 self("\t%s" % i)
```
mhils: `b"\t%s"` ?

dufferzafar: But we decided that we'll be logging `unicode` and not `bytes` right?

mhils: Oh damn, you're right. Thanks for catching this! We then need to decode `i` somehow?

dufferzafar:
Will have to be rebase (to remove some conflicts) once #1195 gets merged.

Kriechi:
@dufferzafar #1195 is merged! Great work so far - I hope we don't loose track of all those PRs you are sending 👍

dufferzafar:
Stalled because it depends on a lot of pathod stuff that hasn't been ported yet.

dufferzafar:
Closing this, since #1211 will remove the `pathod.app` module.

PR #1171, opened 2016-05-28
Python 3 - pathod.log

dufferzafar:


test/pathod/test_log.py
```diff
@@ -1,10 +1,10 @@
-import StringIO
 from pathod import log
 from netlib.exceptions import TcpDisconnect
-import netlib.tcp
 
+from six.moves import cStringIO
```
mhils: I think we can just use `io.BytesIO` here.

dufferzafar: Then `pathod.log` will have to be modified to work with `bytes`

mhils:
Uh, I think my last comment was moot. We probably want to log unicode?

dufferzafar:
:confused:

mhils:
I merged cdc4acc689673141a928e2ee3e5e2b753a28a8f7 into master and fixed the tests. We're set here! :smiley:

PR #1172, opened 2016-05-29
First steps to solidifying the core

cortesi:
This is a hefty patch that takes some first steps towards solidifying our core comms mechanism. Highlights include:

- Introduce a @controller.handler decorator that automatically replys() to messages
- Introduce a reply.take() method to take control of a message and prevent automatic reply
- Handers are no longer named handle_message. The decorator makes clear what is a handler, so it's just "message" now.
- Add an explicit list of permitted messages. This clarifies our mechanisms, and will be the start of better message documentation and more sophisticated handling mechanisms down the track.
- Add a bunch of sanity checks for messages that are acked twice, or never, or are given invalid arguments. Use this to catch numerous occurrences throughout the codebase.
- Simplify the class hierarchy in controller by merging Master and ServerMaster.

### DIFF
--- a/mitmproxy/controller.py
+++ b/mitmproxy/controller.py
@@ -1,8 +1,14 @@
 from __future__ import absolute_import
 from six.moves import queue
 import threading
+import functools
+import sys
 
-from .exceptions import Kill
+from . import exceptions
+
+
+class ControlError(Exception):
+    pass
 
 
 class Master(object):
@@ -36,6 +42,12 @@ class Master(object):
             while True:
                 mtype, obj = self.event_queue.get(timeout=timeout)
                 handle_func = getattr(self, "handle_" + mtype)
+                # if not handle_func.func_dict.get("handler"):
+                #     raise ControlError(
+                #         "Handler function %s is not decorated with controller.handler"%(
+                #             handle_func
+                #         )
+                #     )
                 handle_func(obj)
                 self.event_queue.task_done()
                 changed = True
@@ -100,7 +112,7 @@ class Channel(object):
         master. Then wait for a response.
 
         Raises:
-            Kill: All connections should be closed immediately.
+            exceptions.Kill: All connections should be closed immediately.
         """
         m.reply = Reply(m)
         self.q.put((mtype, m))
@@ -110,11 +122,11 @@ class Channel(object):
                 g = m.reply.q.get(timeout=0.5)
             except queue.Empty:  # pragma: no cover
                 continue
-            if g == Kill:
-                raise Kill()
+            if g == exceptions.Kill:
+                raise exceptions.Kill()
             return g
 
-        raise Kill()
+        raise exceptions.Kill()
 
     def tell(self, mtype, m):
         """
@@ -133,6 +145,10 @@ class DummyReply(object):
 
     def __init__(self):
         self.acked = False
+        self.taken = False
+
+    def take(self):
+        self.taken = True
 
     def __call__(self, msg=False):
         self.acked = True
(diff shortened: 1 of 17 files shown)


mitmproxy/controller.py
```diff
@@ -1,22 +1,58 @@
...
+    "responseheaders",
+
+    "next_layer",
+
+    "error",
+    "log",
+])
 
+
+class ControlError(Exception):
```
mhils: This should be `class ControlException(ProxyException): pass` in `mitmproxy.exceptions`.

mitmproxy/controller.py
```diff
@@ -45,38 +91,12 @@ def tick(self, timeout):
         return changed
 
     def shutdown(self):
-        self.should_exit.set()
-
-
-class ServerMaster(Master):
```
mhils: What was your motivation to merge these two again?

mitmproxy/controller.py
```diff
@@ -147,23 +167,63 @@ def __call__(self, msg=False):
 NO_REPLY = object()
 
 
-class Reply(object):
+def handler(f):
+    @functools.wraps(f)
+    def wrapper(*args, **kwargs):
+        if len(args) == 1:
```
mhils: What funky magic are we doing here?
(From reading the diff sequentially until here I suspect that this deals with inheritance, but we should add a comment whatsoever)

mitmproxy/controller.py
```diff
@@ -147,23 +167,63 @@ def __call__(self, msg=False):
...
+
+        handling = False
+        # We're the first handler - ack responsibility is ours
+        if not message.reply.handled:
+            handling = True
+            message.reply.handled = True
+
+        ret = f(*args, **kwargs)
 
+        if handling and not message.reply.acked and not message.reply.taken:
```
mhils: :+1: :+1: :+1:

mitmproxy/controller.py
```diff
@@ -147,23 +167,63 @@ def __call__(self, msg=False):
...
+            message.reply()
+        return ret
+    wrapper.func_dict["__handler"] = True
```
mhils: Why can't we just use `wrapper.__handler = True`?

mitmproxy/controller.py
cortesi: There's no need for multiple layers of abstraction here - the very simple event loop is never used, and in almost all circumstances the server will be running in a different thread. We have many such unused layers in the codebase at the moment that I'd like to cut out.

mitmproxy/controller.py
cortesi: Yes, this deals with inheritance, and making sure we can over-ride handlers without ack-ing twice. I'll add a comment.

netlib/tcp.py
```diff
@@ -901,7 +901,7 @@ def handle_error(self, connection_, client_address, fp=sys.stderr):
         """
         # If a thread has persisted after interpreter exit, the module might be
         # none.
-        if traceback:
+        if traceback and six:
```
mhils: ?

cortesi: That code can execute after modules have been unloaded (interpreter exit with a persistent thread). If the six module or the traceback module is gone, there's nothing we can do, so we ignore rather than crashing.

mitmproxy/controller.py
cortesi: Using func_dict is mildly more explicit about what we're doing. I don't have strong feelings here.

mhils:
This looks great! :tada: 
Do we need to make any adjustements to [`ReplyProxy`](https://github.com/mitmproxy/mitmproxy/blob/b70663f434897c3830ac0ca4c972face484ce064/mitmproxy/script/concurrent.py#L9)? (It looks like we get away without implementing `__setattr__` there as `take()` is a function?)

mitmproxy/proxy/root_context.py
```diff
@@ -132,7 +132,6 @@ def __repr__(self):
 
 
 class Log(object):
-
```
Kriechi: Please keep this blank line. PEP8 says: 
> Method definitions inside a class are surrounded by a single blank line.

cortesi: Lets fix this once and for all with a style enforcer. The specific choices don't matter as much as consistency.

PR #1173, opened 2016-05-29
Format examples

mhils:
This PR makes `examples/` fully PEP8-compilant.


examples/mitmproxywrapper.py
```diff
@@ -16,7 +16,6 @@
 
 
 class Wrapper(object):
-
```
Kriechi: flake8 seems to like it - not sure why...

examples/mitmproxywrapper.py
```diff
@@ -142,7 +141,7 @@ def main(cls):
             '--toggle',
             action='store_true',
             help='just toggle the proxy configuration')
-#         parser.add_argument('--honeyproxy', action='store_true', help='run honeyproxy instead of mitmproxy')
+        #         parser.add_argument('--honeyproxy', action='store_true', help='run honeyproxy instead of mitmproxy')
```
Kriechi: This messes things up a bit - can you fix that manually?

examples/mitmproxywrapper.py
```diff
@@ -155,8 +154,8 @@ def main(cls):
 
         if args.toggle:
             wrapper.toggle_proxy()
-#         elif args.honeyproxy:
-#             wrapper.wrap_honeyproxy()
+        #         elif args.honeyproxy:
+        #             wrapper.wrap_honeyproxy()
```
Kriechi: Same here

PR #1175, opened 2016-05-29
add tox

Kriechi:
@dufferzafar: this might be very useful to run tests in py2 and py3 at the same time.
Just run `tox` and it tests both. I think you can even specify files and stuff.
Feel free to adapt it to fit your needs.

I'm using pyenv to install the required python version before running tox the first time.


cortesi:
Yes, this is great. As a next step, we should replace our dev virtualenv creation scripts with this too. E.g.:

```
tox
source ./.tox/py27/bin/activate
```

PR #1178, opened 2016-05-31
Improve handling of HTTP2 pseudo-headers

cortesi:
- The canonical source for :method, :scheme and :path are the .method, .scheme
and .path attributes on the request object.
- These pseudo-headers are stripped after reading the request, and re-inserted
just before sending.
- The :authority header remains, and should be handled analagously to the Host
header in HTTP1 with respect to display and user interaction.

### DIFF
--- a/netlib/multidict.py
+++ b/netlib/multidict.py
@@ -171,6 +171,14 @@ class _MultiDict(MutableMapping, Serializable):
         else:
             return super(_MultiDict, self).items()
 
+    def clear(self, key):
+        """
+            Removes all items with the specified key, and does not raise an
+            exception if the key does not exist.
+        """
+        if key in self:
+            del self[key]
+
     def to_dict(self):
         """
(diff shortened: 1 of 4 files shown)


netlib/multidict.py
```diff
@@ -171,6 +171,14 @@ def items(self, multi=False):
         else:
             return super(_MultiDict, self).items()
 
+    def clear(self, key):
+        """
```
mhils: I think we should stick to the standard API and use `.pop(key, None)`

mhils:
Other than that, I fully agree on the changes here, but this is more of @Kriechi's domain than mine.

netlib/multidict.py
cortesi: Let me ponder that. I've always found the .pop interface messy and am in the habit of adding a .clear to my implementations of the interface.

PR #1179, opened 2016-05-31
Start reorganising */utils.py

cortesi:
- Move functions to more natural places, where appropriate
- Extract humanization functions to netlib.human, remove reduntant implementations and clean up
- Remove unused code


netlib/http/request.py
```diff
@@ -6,7 +6,9 @@
 from six.moves import urllib
 
 from netlib import utils
-from netlib.http import cookies
+import netlib.http.url
```
Kriechi: we are in `netlib.http` - why not just `from . import url`?

cortesi: This is to avoid stepping on a number of naked occurrences of the "url" name in the file itself. Renaming those would be ok, but seemed more invasive than I want for this patch (which should be an almost 0 behaviour changed refactoring).

netlib/http/http1/read.py
```diff
@@ -6,6 +6,19 @@
 from ... import utils
 from ...exceptions import HttpReadDisconnect, HttpSyntaxException, HttpException, TcpDisconnect
 from .. import Request, Response, Headers
+from .. import url
+
+
+def get_header_tokens(headers, key):
```
mhils: Move this to `Headers.get_tokens`?

netlib/http/url.py
```diff
@@ -0,0 +1,96 @@
+import six
```
mhils: Maybe I'm a bit pendantic here, but shouldn't it be `netlib/url.py`? It's not really HTTP-specific.

PR #1180, opened 2016-05-31
Python 3 - pathod.language.base

dufferzafar:


mhils:
I'll merge this right away because it's obviously correct, but we should keep adding travis coverage :smiley:

dufferzafar:
Um, but not all tests in the `test_language_base` were fixed :/

PR #1183, opened 2016-06-01
use module-level imports only

mhils:
This PR rewrites all mitmproxy modules to use module-level imports only.

My personal opinion is that this change is unnecessary and decreases legibility. Prefixing every exception with `exceptions.*` is just worse than Java.


mitmproxy/platform/windows.py
```diff
@@ -1,18 +1,19 @@
...
-from six.moves import socketserver
 import struct
 import threading
 import time
-from collections import OrderedDict
 
+import configargparse
+from pydivert.enum import Direction
+from pydivert.enum import Flag
+from pydivert.enum import Layer
```
cortesi: import pydivert.enum?

mitmproxy/contentviews.py
```diff
@@ -12,27 +12,32 @@
...
+from netlib import http
+from netlib import odict
 from netlib.http import url
-from netlib.utils import clean_bin, hexdump
-from . import utils
-from .exceptions import ContentViewException
-from .contrib import jsbeautifier
-from .contrib.wbxml.ASCommandResponse import ASCommandResponse
+from netlib.utils import clean_bin
+from netlib.utils import hexdump
```
cortesi: import netlib.utils

PR #1189, opened 2016-06-01
[web] WIP: eliminate states

gzzhanghao:
This PR eliminates states from components. It move states to ProxyApp component and pass data to children with props.

Roadmap:

- [x] FlowTable
- [ ] Header
- [ ] MainView
- [ ] ContentView
- [ ] FlowView
- [ ] EventLog

### DIFF
--- a/web/src/js/components/flowtable.js
+++ b/web/src/js/components/flowtable.js
@@ -41,6 +41,9 @@ class FlowTableHead extends React.Component {
     static propTypes = {
         setSortKeyFun: React.PropTypes.func.isRequired,
         columns: React.PropTypes.array.isRequired,
+        onChangeSortMethod: React.PropTypes.array.isRequired,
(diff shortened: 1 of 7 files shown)


web/src/js/components/flowtable.js
```diff
@@ -51,13 +54,13 @@ class FlowTableHead extends React.Component {
     onClick(Column) {
         const hasSort = Column.sortKeyFun;
 
-        let sortDesc = this.state.sortDesc;
+        var sortDesc = this.props.sortDesc;
```
cle1000: i think it is better to use `let` instead of `var` with es6

web/src/js/components/flowview/contentview.js
```diff
@@ -236,7 +237,7 @@ var ContentView = React.createClass({
     },
     render: function () {
         var { flow, message } = this.props
```
cle1000: again use let

PR #1191, opened 2016-06-02
Utils reorganisation

cortesi:
We currently have 3 (primary) utils.py modules in the project. They've become a messy grab-bag of unorganised stuff. I'm teasing them out into something more structured. 

- Extract string-relate utility functions to netlib.strutils


mhils:
Can we replace xrepr with the new stuff for that right away?

cortesi:
Which new stuff? It seems like a super-simple thing used once that shouldn't even be its own function.

cortesi:
OK, when CI passes this can be merged.

PR #1192, opened 2016-06-02
WIP: Solidify pathod test suite

cortesi:
This PR tightens up the Pathod test suite, and adds some tools that will help us everhwere:

- Implement a service thread counter that lets us wait until all current service threads have exited
- Make TCPClient.connect return a context handler that reliably closes connections on exit
- Move away from our reliance on requests for tests. It's a nice library, but its behaviour is just too opaque with respect to things like thread pools for us to use it when we need precise control
- Make the Pathod test library access logs directly with a thread lock, rather than using an HTTP API (also make log thread locks not be a nop. heh.)
- Fix a number of bugs exposed by the above

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ install:
     if [[ $TRAVIS_OS_NAME == "osx" ]]
     then
       brew update || brew update # try again if it fails
-      brew outdated openssl || brew upgrade openssl
+      # brew outdated openssl || brew upgrade openssl
       brew install python
     fi
   - pip install -U virtualenv
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,9 @@ matrix:
       git:
         depth: 9999999
     - python: 3.5
-      env: SCOPE="netlib ./test/mitmproxy/script ./test/pathod/test_utils.py ./test/pathod/test_log.py"
+      env: SCOPE="netlib ./test/mitmproxy/script ./test/pathod/test_utils.py ./test/pathod/test_log.py ./test/pathod/test_language_generators.py"
     - python: 3.5
-      env: SCOPE="netlib ./test/mitmproxy/script ./test/pathod/test_utils.py ./test/pathod/test_log.py" NO_ALPN=1
+      env: SCOPE="netlib ./test/mitmproxy/script ./test/pathod/test_utils.py ./test/pathod/test_log.py ./test/pathod/test_language_generators.py" NO_ALPN=1
     - python: 2.7
       env: DOCS=1
       script: 'cd docs && make html'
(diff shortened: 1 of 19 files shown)


.travis.yml
```diff
@@ -36,7 +36,7 @@ install:
     if [[ $TRAVIS_OS_NAME == "osx" ]]
     then
       brew update || brew update # try again if it fails
-      brew outdated openssl || brew upgrade openssl
+      # brew outdated openssl || brew upgrade openssl
```
mhils: Why?

cortesi: Keep watching the Travis build logs, and lets discuss in Slack.

netlib/tcp.py
```diff
@@ -708,6 +715,7 @@ def connect(self):
         self.connection = connection
         self.ip_address = Address(connection.getpeername())
         self._makefile()
+        return _closer(self)
```
mhils: This looks weird. How about `__enter__` and `__exit__` for `TCPClient` ?

```python
with TCPClient(..) as t:
    t.write(b"foo")

Comment (cortesi) [inline netlib/tcp.py]: That's the wrong place to put the handler. We're only connected after we call .connect().

Comment (mhils) [inline netlib/tcp.py]: Well, the idea is that `__enter__` `.connect()`s.

Comment (mhils) [inline netlib/tcp.py]: (no super strong opinion on this though)

Comment (cortesi) [inline netlib/tcp.py]: Well, there's a reason why we don't just connect() int he __init__ of TCPClient. It's just a more flexible API. This is especially the case with derived classes like Pathoc where reuse is quite often useful.

File: netlib/tcp.py [hunk 3]
```diff
@@ -845,15 +872,17 @@ def __init__(self, address):
         self.socket.bind(self.address())
         self.address = Address.wrap(self.socket.getsockname())
         self.socket.listen(self.request_queue_size)
+        self.counter = Counter()
```
mhils: We should give this a descriptive name.

pathod/test.py
```diff
@@ -39,39 +41,51 @@ def p(self, spec):
...
+        while 1:
+            if time.time() - start >= timeout:
+                raise TimeoutError(
+                    "%s service threads still alive" %
+                    self.thread.server.counter.count
+                )
+            if self.thread.server.counter.count == 0:
+                return
+
+    def expect_log(self, n, timeout=5):
```
mhils: Do we still need this now?

mhils:
Other than that, LGTM!

pathod/test.py
cortesi: Yes, I think we still do. First, this is our synchronisation mechanism to ensure that tests don't bleed into each other. Second, some bug classes, like the websocket one fixed in this PR, exhibit themselves through failure to properly exit.

netlib/tcp.py
cortesi: I'll make this handler_counter then. Not elegant, but at least explicit.

PR #1195, opened 2016-06-02
Python 3 - pathod.language.generators

dufferzafar:
Received a huge amount of help from the awesome folks at #python IRC :)

### DIFF
--- a/pathod/language/generators.py
+++ b/pathod/language/generators.py
@@ -37,10 +37,6 @@ class TransformGenerator(object):
         d = self.gen.__getitem__(x)
         return self.transform(x, d)
 
-    def __getslice__(self, a, b):
-        d = self.gen.__getslice__(a, b)
-        return self.transform(a, d)
-
     def __repr__(self):
         return "'transform(%s)'" % self.gen
 
--- a/pathod/language/generators.py
+++ b/pathod/language/generators.py
@@ -2,6 +2,8 @@ import string
 import random
 import mmap
 
+import six
+
 DATATYPES = dict(
     ascii_letters=string.ascii_letters.encode(),
     ascii_lowercase=string.ascii_lowercase.encode(),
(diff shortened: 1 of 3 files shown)


pathod/language/generators.py
```diff
@@ -37,14 +37,19 @@ def __getitem__(self, x):
...
         return "'transform(%s)'" % self.gen
 
 
+def rand_byte(chars):
+    """
+        Return a random character as byte from a charset.
+    """
+    # bytearray has consistent behaviour on both Python 2 and 3
+    # while bytes does not
+    return bytes(bytearray([random.choice(chars)]))
```
mhils: I really appreciate the work that went into this, nonetheless let me propose to simplify it by keeping in mind that we want to ditch Python 2 completely once your awesome job is done here:
```
if six.PY2:
    return random.choice(chars)
return bytes([chars])
```

pathod/language/generators.py
```diff
@@ -70,17 +73,17 @@ class FileGenerator(object):
...
 
     def __getitem__(self, x):
-        return self.map.__getitem__(x)
-
-    def __getslice__(self, a, b):
-        return self.map.__getslice__(a, b)
+        if isinstance(x, slice):
+            return self.map.__getitem__(x)
+        # A slice of length 1 returns a byte object (not an integer)
+        return self.map.__getitem__(slice(x, x+1 or self.map.size()))
```
mhils: I don't understand the `or self.map.size()` part here - what is this about?

dufferzafar: When `x = -1`, and we want the last character.

mhils: Ok. But this won't work for `x = -2`, will it? Shall we just use `bytes([chars])` (+ six.PY2) here as well?

dufferzafar: Um, it does work for -2. Just ran the checks with a bunch of negative numbers.

mhils: Oh wow, you're absolutely right, we only need to work around x=0. Scratch my last comment!

mhils:
Let's fix https://github.com/mitmproxy/mitmproxy/pull/1195#discussion_r65497564 + the linter and then merge this ASAP! :smiley:

mhils:
I fixed the lint error and merged this. Really great work! :smiley: :cake:

Issue #1111, opened 2016-05-03
Verify upstream certificates by default.


mhils:
Per discussion in #1108 and some other places, I would like to turn on `--verify-upstream-cert` by default, with an easy option to disable it.

Basic Features:
- [ ] Add setting to the mitmproxy options menu.
- [ ] Add `--insecure` command line switch to disable certificate verification. Do we also want to have `-k` for this, to mirror curl's API?

Bonus features:
- [ ] `~/.mitmproxy/trusted.pem` should contain additional certs that are trusted by default.
- [ ] When sending an error to the client, send a detailed error page giving the client the option to add the cert to `~/.mitmproxy/trusted.pem`.
- [ ] Display rejects connections as flows in the UI with an easy option to add certs to `~/.mitmproxy/trusted.pem`. (this requires substantial work at other places)

Caveats:
- In a [user] -- TLS -- [mitmproxy] -- TCP -- [server] configuration, we cannot signal to the client that there's essentially no security. As I see it, this should only affect reverse proxy mode, manually changed requests and mitm.it. As such a configuration is a deliberate choice, we should be good.

If anyone wants to work on this, the first task would be to add this setting to mitmproxy's Options screen and send a separate PR for this! :smiley:

s4chin:
I would like to take this. For the basic features, I think it would go like this:
Remove the `--verify-upstream-cert` and add the `--insecure` option. In the `config.py`, remove the `ssl_verify_upstream_cert` variable and add an `ssl_insecure` variable instead and make changes accordingly.
That way the `ssl_insecure` default value is `False` and we set an `--insecure` option to set it to `True`. Is this the correct approach or you prefer I keep the `ssl_verify_upstream_cert` variable and make changes accordingly?

mhils:
@s4chin: This sounds like an excellent first step. Please go ahead! :smiley:

PR #1197, opened 2016-06-02
 Verify upstream certificates by default. #1111

s4chin:
This adds the `--insecure` option and enables certificate verification by default. The tests fail and I need help in knowing how to fix them.

### DIFF
--- a/mitmproxy/console/options.py
+++ b/mitmproxy/console/options.py
@@ -93,7 +93,7 @@ class Options(urwid.WidgetWrap):
                 ),
                 select.Option(
                     "Don't Verify Upstream Certs",
-                    "K",
+                    "V",
                     lambda: master.server.config.ssl_insecure,
                     self.toggle_ssl_insecure
                 ),
@@ -198,7 +198,7 @@ class Options(urwid.WidgetWrap):
         signals.update_settings.send(self)
 
     def toggle_ssl_insecure(self):
-        self.master.server.config.ssl_insecure = not self.master.server.config.ssl_insecure
+        self.master.server.config.openssl_verification_mode_server = VERIFY_PEER if self.master.server.config.openssl_verification_mode_server is VERIFY_NONE else VERIFY_NONE
         signals.update_settings.send(self)
 
     def setheaders(self):
--- a/mitmproxy/console/options.py
+++ b/mitmproxy/console/options.py
@@ -198,7 +198,8 @@ class Options(urwid.WidgetWrap):
         signals.update_settings.send(self)
 
     def toggle_ssl_insecure(self):
-        self.master.server.config.openssl_verification_mode_server = VERIFY_PEER if self.master.server.config.openssl_verification_mode_server is VERIFY_NONE else VERIFY_NONE
+        self.master.server.config.ssl_insecure = not self.master.server.config.ssl_insecure
+        self.master.server.config.openssl_verification_mode_server = VERIFY_NONE if self.master.server.config.ssl_insecure else VERIFY_PEER
         signals.update_settings.send(self)
 
     def setheaders(self):
--- a/mitmproxy/console/options.py
+++ b/mitmproxy/console/options.py
@@ -8,6 +8,7 @@ from mitmproxy.console import grideditor
 from mitmproxy.console import palettes
 from mitmproxy.console import select
 from mitmproxy.console import signals
+from OpenSSL import SSL
 
 footer = [
     ('heading_key', "enter/space"), ":toggle ",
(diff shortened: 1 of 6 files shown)


s4chin:
Adds the basic features as per #1111

mhils:
This looks great, but we're still failing on some systems. @Kriechi, any idea?

@s4chin: I'd love to have a toggle for this in the mitmproxy options (press `o` in mitmproxy). Do you think you could get that going? :smiley:

s4chin:
Which key do I assign for the toggle option?

mhils:
`k` is good, go ahead! :-)

mitmproxy/console/options.py
```diff
@@ -191,6 +197,10 @@ def toggle_upstream_cert(self):
         self.master.server.config.no_upstream_cert = not self.master.server.config.no_upstream_cert
         signals.update_settings.send(self)
 
+    def toggle_ssl_insecure(self):
+        self.master.server.config.ssl_insecure = not self.master.server.config.ssl_insecure
```
cortesi: This fails, because ProxyConfig doesn't have an ssl_insecure attribute

mitmproxy/proxy/config.py
```diff
@@ -116,10 +116,10 @@ def __init__(
         self.openssl_method_server, self.openssl_options_server = \
             tcp.sslversion_choices[ssl_version_server]
 
-        if ssl_verify_upstream_cert:
-            self.openssl_verification_mode_server = SSL.VERIFY_PEER
-        else:
+        if ssl_insecure:
             self.openssl_verification_mode_server = SSL.VERIFY_NONE
+        else:
+            self.openssl_verification_mode_server = SSL.VERIFY_PEER
```
cortesi: Once we've added the attribute to the object here, we also need to make sure that the openssl_verification_mode_server attribute changes correctly when ssl_insecure changes. If we set this only in __init__, changing it in the console options will be a NOP.

mitmproxy/console/options.py
```diff
@@ -91,6 +91,12 @@ def __init__(self, master):
                     lambda: master.server.config.check_tcp,
                     self.tcp_proxy
                 ),
+                select.Option(
+                    "Don't Verify Upstream Certs",
+                    "K",
```
cortesi: The auto-highlighting feature for options looks for the given shortcut in the description string. Since "K" is no in "Don't Verify Upstream Certs" this will fail. Pick a mnemonic that's in the string.

cortesi:
Sachin, thanks for working on this. It's a valuable feature we definitely need. Please take a look at the comments I've added inline in the code - there's more work needed before we can merge.

s4chin:
I made changes according to all your comments. Should I also squash all the commits?

mhils:
Looks way better now! :smiley:

The `ProxyConfig.openssl_verification_mode_server` part looks like it's going to fail at some point where somebody forgets to update one of the attributes. Can we implement `openssl_verification_mode_server` as a `@property` ?

We should also add a test (class) that overrides `get_proxy_config` to remove the `insecure=True` default in the test suite and verifies that we actually fail for untrusted CAs.

mhils:
What about the remaining failing tests? @Kriechi, any idea?

tdickers:
I may be late to the party, but you may find that this site helps with testing: https://badssl.com/

test/mitmproxy/tservers.py
```diff
@@ -122,6 +122,7 @@ def get_proxy_config(cls):
             cadir = cls.cadir,
             authenticator = cls.authenticator,
             add_upstream_certs_to_client_chain = cls.add_upstream_certs_to_client_chain,
+            ssl_insecure = True,
```
Kriechi: I think this changes the default for all HTTP/2 tests - maybe we can only set this to true in the tests we actually want it to change?

mhils: It's fine if all tests are run using `--insecure`. We should add a tests that assures that we actually verify the certificate if `ssl_insecure` is not passed.

Kriechi:
About the failing HTTP/2 tests, the error message is: 

`TlsException('Cannot validate certificate hostname without SNI',)",)`

Kriechi:
I just pushed an improved version of the HTTP/2 tests in 5d0de16f57e8af971fd282660e3c32d08e7e1317. This should make it easier to debug and track down the underlying cause.

test/mitmproxy/tservers.py
Kriechi: Actually, HTTP/2 misses this value - because it has its own TestBase.
@s4chin please add the same option here: https://github.com/mitmproxy/mitmproxy/blob/5d0de16f57e8af971fd282660e3c32d08e7e1317/test/mitmproxy/test_protocol_http2.py#L97-L101

mhils:
Thanks for fixing the tests, I will take a look over the weekend! :smiley:

PR #1201, opened 2016-06-03
Redux ducks

mhils:
This is my third attempt (#1194, #1200) at introducing a reasonable redux architecture. This time I used https://github.com/erikras/ducks-modular-redux to structure our redux code, and I'm _very_ happy with that approach. This very much fall into place magically. If noone objects, I think we should merge and continue based on this.


cortesi:
LGTM

PR #1203, opened 2016-06-03
Force a build to confirm that the new cryptography wheels work for you. [DO NOT MERGE]

Lukasa:
Please don't merge this: I just wanted to trigger a travis run to make sure that the builds work with the new cryptography.


Lukasa:
Awesome, this works great. Thanks folks!

PR #1209, opened 2016-06-04
Python 3 - pathod.language.base

dufferzafar:


pathod/language/base.py
```diff
@@ -226,7 +226,7 @@ def get_generator(self, settings):
         return generators.FileGenerator(s)
 
     def spec(self):
-        return "<'%s'" % strutils.bytes_to_escaped_str(self.path)
+        return "<'%s'" % self.path
```
dufferzafar: I'm not sure if we need to store path as bytes, and then decode/escape it when printing or just store it as a unicode string.

Kriechi:
LGTM!

One thing before I merge this: could you please add the test file also to `tox.ini`?

dufferzafar:
Done!

PR #1210, opened 2016-06-04
Python 3 - pathod.language.http

dufferzafar:


pathod/language/base.py
```diff
@@ -261,7 +261,7 @@ class _Component(Token):
 
     """
         A value component of the primary specification of an message.
-        Components produce byte values desribe the bytes of the message.
+        Components produce byte values desribing the bytes of the message.
```
Kriechi: *describing*

pathod/language/base.py
```diff
@@ -376,6 +376,30 @@ def freeze(self, settings):
         return self.__class__(self.value.freeze(settings))
 
 
+class BytesWrapper():
```
dufferzafar: @mhils Please have a look at this! (some related discussion is on slack.)

pathod/language/http.py
```diff
@@ -191,7 +191,7 @@ def resolve(self, settings, msg=None):
         if self.ws:
             if not settings.websocket_key:
                 raise exceptions.RenderError(
-                    "No websocket key - have we seen a client handshake?"
+                    b"No websocket key - have we seen a client handshake?"
```
Kriechi: I think the exception doesn't have to be bytes.

dufferzafar: Yeah, you're correct. It doesn't have to be. I was just blindly converting literals to bytes. Will fix this!

pathod/language/base.py
mhils: I agree that the byte handling is a bit fiddly, but I'm not sure if introducing a second kind of `bytes` with improved/different semantics is really worth it. For example, you cannot call `.decode()` on this anymore, etc. etc..

I'd rather have us do the `[i:i+1]` dance where required - I believe/hope we have tackled most problem instances of this anway.

pathod/language/http.py
```diff
@@ -171,17 +171,17 @@ def reason(self):
...
+        l.append(b" ")
         if self.reason:
             l.extend(self.reason.values(settings))
         else:
             l.append(
                 status_codes.RESPONSES.get(
                     status_code,
-                    "Unknown code"
+                    b"Unknown code"
                 )
```
mhils: that doesn't encode `status_codes.RESPONSES`, does it?

dufferzafar: I've done that here: https://github.com/mitmproxy/mitmproxy/pull/1210/commits/b3763fd850f13ed8616c454910d29d00c371cc65

dufferzafar:
Removed the BytesWrapper. 3 Tests still fail, but I'll need some help with them.

netlib/http/user_agents.py
```diff
@@ -48,5 +48,5 @@ def get_by_shortcut(s):
         Retrieve a user agent entry by shortcut.
     """
     for i in UASTRINGS:
-        if s == i[1]:
+        if s == i[1].encode():
```
mhils: I think this function should still accept `str`, we should rather `.decode()` in pathod. Otherwise we'll have way more work in mitmproxy with that.

netlib/http/status_codes.py
```diff
@@ -104,3 +106,5 @@
     INSUFFICIENT_STORAGE_SPACE: "Insufficient Storage Space",
     NOT_EXTENDED: "Not Extended"
 }
+
+RESPONSES = {k: v.encode() for k, v in six.iteritems(RESPONSES)}
```
mhils: Similar situation here: we may want to `.encode()` in pathod - the fewer we have to deal with bytes, the happier we'll be.

mhils:
@dufferzafar: I took a look at the remaining failing tests and fixed all of them: https://github.com/mitmproxy/mitmproxy/commits/pathod-lang-http - this is merged! :smiley:

PR #1211, opened 2016-06-05
WIP: Radical webectomy of pathod

cortesi:
This pull request removes the API and web app from Pathod entirely. The plans I had for these interfaces when I first wrote Pathod are now dated, and we now have better ways to interface with the Pathod logs for testing (thread locks). 

All that remains to be done is to move the documentation from the "templates" directory into the mitmproxy docs proper.


Kriechi:
Nice! I think this will make things a bit easier for @dufferzafar as well.

dufferzafar:
Yeah, one less file to port. :smile:

PR #1212, opened 2016-06-05
web: completely move flow state to redux

mhils:
This PR removes the last bits of `FlowStore` and moves all state into redux. There are a few features broken now (e.g. sorting), but overall the transition to redux turns out to be great.


cortesi:
So much nicer.

PR #1213, opened 2016-06-05
Allow cryptography 1.4

felixonmars:
All tests passed and it works correctly here.


mhils:
Fast as always, thanks! :smiley:

Will merge as soon as the tests went through.

Kriechi:
Thanks! 🎉

PR #1215, opened 2016-06-05
Now keeps header capitalization

### LINKED ISSUE #1214: Headers aren't always case insensitive
Headers like `Content-Length` should preserve case after going trough mitmproxy, because some versions of PHP don't like lowercase headers (this is a server-side misimplementation of [RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2), but it would be nice if mitmproxy was able to handle this.).  

Steps to reproduce:
Capture some traffic, replay it with `mitmdump -nc capture_file`. Everything works as intended.
Now replay it, but use a script that replaces the request content. The `Content-Length` header get's changed to `content-lenght`.  

The problem probably lies at [netlib/http/message.py](https://github.com/mitmproxy/mitmproxy/blob/master/netlib/http/message.py#L104)
Mitmproxy Version: latest

redfast00:
Fixes #1214
Some servers don't fully respect RFC 2616 (Headers should be case-insensitive). This makes it so that the capitalization of the header fields stay the same, even if (for example) the `Content-Length` changes.

### DIFF
--- a/netlib/multidict.py
+++ b/netlib/multidict.py
@@ -105,7 +105,7 @@ class _MultiDict(MutableMapping, basetypes.Serializable):
             if self._kconv(field[0]) == key_kconv:
                 if values:
                     new_fields.append(
-                        (key, values.pop(0))
+                        (field[0], values.pop(0))
                     )
             else:
                 new_fields.append(field)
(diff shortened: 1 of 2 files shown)


mhils:
Thanks. Could you please adjust the failing multidict tests as well?

PR #1219, opened 2016-06-06
Show tracebacks on mitmproxy error pages

dwfreed:
Instead of the repr of the exception, which is basically useless, show
the full traceback on the error page.

### DIFF
--- a/mitmproxy/models/http.py
+++ b/mitmproxy/models/http.py
@@ -251,10 +251,11 @@ def make_error_response(status_code, message, headers=None):
             <head>
                 <title>%d %s</title>
             </head>
-            <body>%s</body>
+            <body><pre>%s</pre></body>
         </html>
     """.strip() % (status_code, response, cgi.escape(message))
-    body = body.encode("utf8", "replace")
+    # Remove the .decode('unicode_escape') when converting to py3
+    body = body.decode('unicode_escape').encode("utf8", "replace")
 
     if not headers:
         headers = Headers(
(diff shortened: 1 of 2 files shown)


Kriechi:
Wouldn't it be just easier to get rid of the html stuff and render plain text?

dwfreed:
But then you wouldn't have the HTML title with the HTTP status

mhils:
I'm not sure if this is a great idea, this introduces a potential information leak in some contexts. @cortesi, thoughts?

dwfreed:
Not really much more than printing the repr(exception) already does

dwfreed:
It needs some fixing anyway: https://travis-ci.org/mitmproxy/mitmproxy/jobs/135699606#L421

dwfreed:
Oh, even better, that test passes locally, but not on travis

dwfreed:
I see, it's because the data is random, so test failure there is non-deterministic

cortesi:
This is an idea worth considering, but I'm afraid that having considered it, I'm not in favour of it. ;) We should expose as little as possible of our internal workings to the proxy clients (which may be hostile, or we may be trying to be covert, etc.). In fact, the exception repr itself should be removed altogether. Instead, we should expose error information to the controller of the proxy - that is, in the eventlog for mitmproxy console and mitmweb, and printed to screen in mitmdump.

dwfreed:
From the discussion on Slack, this likely isn't going to be merged, so I'll close it.  I'll put together some improved logging handling next week.

PR #1220, opened 2016-06-06
Add sorting with redux

cle1000:
moved sort state to redux state, and start with sorting the flow

### DIFF
--- a/web/src/js/components/flowtable.js
+++ b/web/src/js/components/flowtable.js
@@ -94,7 +94,7 @@ FlowTableHead.propTypes = {
 };
 
 const FlowTableHeadContainer = connect(
-    (state, ownProps) => ({
+    (state) => ({
         sort: state.flows.sort
     })
 )(FlowTableHead)
@@ -215,7 +215,8 @@ const parseFilter = _.memoize(Filt.parse)
 
 const FlowTableContainer = connect(
     state => ({
-        flows: state.flows.view,
+        // first idea to sort here, but i think thats not good enough ( and not working yet)...
+        flows: state.flows.view.sort((a,b) => state.flows.sort.sortColumn ?  a.response.status_code > b.response.status_code : 0),
     })
 )(FlowTable)
 
--- a/web/src/js/components/flowtable.js
+++ b/web/src/js/components/flowtable.js
@@ -9,6 +9,7 @@ import AutoScroll from "./helpers/AutoScroll";
 import {calcVScroll} from "./helpers/VirtualScroll";
 import flowtable_columns from "./flowtable-columns.js";
 import Filt from "../filt/filt";
+import {setSort} from "../ducks/flows";
 
 
 FlowRow.propTypes = {
@@ -94,8 +95,11 @@ FlowTableHead.propTypes = {
 };
 
 const FlowTableHeadContainer = connect(
-    (state) => ({
+    state => ({
         sort: state.flows.sort
+    }),
+    dispatch => ({
+        loc: (sort) => dispatch(setSort(sort)),
     })
 )(FlowTableHead)
 
--- a/web/src/js/components/flowtable.js
+++ b/web/src/js/components/flowtable.js
@@ -2,7 +2,6 @@ import React from "react";
 import ReactDOM from "react-dom";
 import {connect} from 'react-redux'
 import classNames from "classnames";
-import {reverseString} from "../utils.js";
 import _ from "lodash";
 import shallowEqual from "shallowequal";
 import AutoScroll from "./helpers/AutoScroll";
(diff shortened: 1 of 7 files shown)


web/src/js/components/flowtable.js
```diff
@@ -215,7 +215,8 @@ const parseFilter = _.memoize(Filt.parse)
 
 const FlowTableContainer = connect(
     state => ({
-        flows: state.flows.view,
+        // first idea to sort here, but i think thats not good enough ( and not working yet)...
+        flows: state.flows.view.sort((a,b) => state.flows.sort.sortColumn ?  a.response.status_code > b.response.status_code : 0),
```
cle1000: maybe some hints how and where to apply the sorting ?

web/src/js/components/mainview.js
```diff
@@ -173,12 +170,14 @@ const MainViewContainer = connect(
...
         flows: state.flows.view,
         filter: state.flows.filter,
+        sort: state.flows.sort,
         highlight: state.flows.highlight,
         selectedFlow: state.flows.all.byId[state.flows.selected[0]]
     }),
     dispatch => ({
         selectFlow: flowId => dispatch(selectFlow(flowId)),
         setFilter: filter => dispatch(setFilter(filter)),
+        setSort: (sort) => dispatch(setSort(sort)),
```
mhils: We can inject this into the `FlowTableHeadContainer` directly, MainView  doesn't need it.

web/src/js/ducks/flows.js
```diff
@@ -20,6 +21,7 @@ const defaultState = {
     view: [],
     filter: undefined,
     highlight: undefined,
+    sort: {sortColumn: undefined, sortDesc: false},
 }
```
mhils: 1) Add `makeSortFn` here. Without testing, I hope we can implement it by exporting all Columns in `flowtable-columns.js` and then just using this:

```javascript
import Columns from "../components/flowtable-columns"

function makeSortFn(sort){
    return Columns[sort.sortColumn].sortKeyFun
}
```

(all_columns in that file is also an relict of the past where we didn't have ES6 modules and could be replaced similarly)

web/src/js/ducks/flows.js
```diff
@@ -46,6 +48,11 @@ export default function reducer(state = defaultState, action) {
                 ...state,
                 highlight: action.highlight
             }
+        case SET_SORT:
```
mhils: The idea is that the view is already sorted in the state. The only actions that modify `view` are `UPDATE_FLOWS`, `SET_FILTER` and `SET_SORT`. 
- `For UPDATE_FLOWS` we can just pass `makeSortFn(...)` as an additional argument to `updateViewList` (this is already implemented, see [here](https://github.com/mitmproxy/mitmproxy/blob/7cb7d9ad32c40cff9ceb0e28a5123960fed3638e/web/src/js/ducks/utils/view.js#L77))
- Same for `SET_FILTER`/`updateViewFilter`
- We probably just want to add `updateViewSort()` in `utils/view` for `SET_SORT` which does exactly the same as `updateViewFilter`, except that it creates a copy of the existing list in the first line instead of calling `.filter(...)` on all flows.

web/src/js/components/flowtable.js
mhils: 3) Just pass the view here as-is :tada:

mhils:
Relevant to this: 54ee8ff4d4b1458880554bf3a28c92b8c069cb53
You probably want to rebase onto master or merge master into your branch. :smiley:

web/src/js/components/mainview.js
```diff
@@ -173,6 +169,7 @@ const MainViewContainer = connect(
     state => ({
         flows: state.flows.view,
         filter: state.flows.filter,
+        sort: state.flows.sort,
```
mhils: If I'm not mistaken, we can skip this? :smiley:

web/src/js/ducks/flows.js
```diff
@@ -20,32 +23,54 @@ const defaultState = {
...
+    sort: {sortColumn: undefined, sortDesc: false},
 }
 
 function makeFilterFn(filter) {
     return filter ? Filt.parse(filter) : () => true;
 }
 
+
+function makeSortFn(sort){
+    let column = _.find(flowtable_columns, c => c.name == sort.sortColumn);
```
mhils: If we `export` all the columns in flowtable-columns.js, we can just `import columns from "../components/flowtable-columns"` here. `columns` is a `name -> function` dict then, so no need for find :smiley:

web/src/js/ducks/flows.js
```diff
@@ -20,32 +23,54 @@ const defaultState = {
...
+    if (!column) return;
+
+    let sortKeyFun = column.sortKeyFun;
+    if (sort.sortDesc) {
+        sortKeyFun = sortKeyFun && function () {
+            const k = column.sortKeyFun.apply(this, arguments);
```
mhils: I believe we can just simplify this to `column.sortKeyFun(flow)`

web/src/js/ducks/flows.js
```diff
@@ -20,32 +23,54 @@ const defaultState = {
...
         case SET_HIGHLIGHT:
             return {
                 ...state,
                 highlight: action.highlight
             }
+        case SET_SORT:
+            return {
+                ...state,
+                sort: action.sort,
+                view: updateViewFilterSort(state.all, makeFilterFn(state.filter), makeSortFn(action.sort))
```
mhils: I think we want two distinct functions `updateViewFilter` and `updateViewSort`, so that we don't have to `.filter()` again when we change the sort order. (see [previous comment](https://github.com/mitmproxy/mitmproxy/pull/1220/files#r65981944))

mhils:
This looks fantastic so far! :smiley: :cake:

web/src/js/ducks/flows.js
cle1000: i cannot find a dict in ../components/flowtable-columns should i create one ? i found an array 
`var all_columns = [
    TLSColumn,
    IconColumn,
    PathColumn,
    MethodColumn,
    StatusColumn,
    SizeColumn,
    TimeColumn
]`

and if a just use `    let c1 = columns[sort.sortColumn];` i got undefined.

web/src/js/ducks/flows.js
cle1000: i copied the old implemetation, if i skip this part: 
`if (sort.sortDesc) {
        sortKeyFun = sortKeyFun && function () {
            const k = column.sortKeyFun.apply(this, arguments);
            return _.isString(k) ? reverseString("" + k) : -k;
        };
    }`
only asc sorting is possible.  what exactly shoud i replace with column.sortKeyFun(flow) ?

mhils:
Excellent!

PR #1230, opened 2016-06-08
fix after cleanup of netlib.utils in mitmproxy.console.statusbar

redfast00:


redfast00:
Can someone please tell me what the codecov/patch tool does and why it failed?

mhils:
We recently enabled coverage reporting for the console, codecov is complaining that there is none. Whatsoever, Thanks for the PR! :smiley:

PR #1231, opened 2016-06-08
Python 3 - pathod.pathoc

dufferzafar:


pathod/pathod.py
```diff
@@ -284,10 +285,10 @@ def addlog(self, log):
         # moment because JSON encoding can't handle binary data, and I don't
         # want to base64 everything.
         if self.server.logreq:
-            encoded_bytes = self.rfile.get_log().encode("string_escape")
+            encoded_bytes = strutils.bytes_to_escaped_str(self.rfile.get_log())
```
dufferzafar: I'm not sure about these, since `bytes_to_escaped_str` returns a unicode string, and I think we need `bytes` here.

dufferzafar:
@mhils I'm a bit confused with what should be bytes (and what should be str) wrt pathod (like the `rfile`, `wfile` etc.)

pathod/pathod.py
cortesi: The comment about JSON there is out of date, since I've just ripped out the web interface entirely, and that was the only portion that required the logs to be JSON serializable. We now have a much simpler thread lock over the log and retrieve it directly. So, strip the out of date comment, and no need to escape the log data.

pathod/pathoc.py
```diff
@@ -73,7 +73,7 @@ def __str__(self):
             parts.append("\tPubkey: %s bit %s" % (pk.bits(), t))
             s = certutils.SSLCert(i)
             if s.altnames:
-                parts.append("\tSANs: %s" % " ".join(s.altnames))
+                parts.append("\tSANs: %s" % b" ".join(s.altnames))
```
Kriechi: Shouldn't the format string also be bytes?
Or what happens if you format `str` with `bytes` arguments?

dufferzafar: Yeah, you're right, the format string should be `bytes` too, and if that's the case then everything in the parts array should be bytes.

Something like [this](https://travis-ci.org/mitmproxy/mitmproxy/jobs/136891877#L617) happens when we mix bytes/str.

dufferzafar: I printed out the parts. Here's how it looks currently:

```
Application Layer Protocol: b'http/1.1'
Cipher: AES256-GCM-SHA384, 256 bit, TLSv1/SSLv3
SSL certificate chain:
  Certificate [0]
        Subject: 
                b'CN'=b'pathod.net'
        Issuer: 
                b'CN'=b'mitmproxy'
                b'O'=b'mitmproxy'
        Version: 2
        Validity: b'20160611160814Z' - b'20190613160814Z'
        Serial: 14658340949850
        Algorithm: b'sha256WithRSAEncryption'
        Pubkey: 2048 bit RSA
        SANs: b'test1.com test2.com'
```

Weird mix of bytes & strings.

Kriechi: Yes, so this is basically used only for visualization and debugging purposes, we can convert/keep everything to `str` (i.e. utf8).

dufferzafar:
Ready for review!

Kriechi:
Merged! Thanks! 🎉

PR #1233, opened 2016-06-09
[web] ProxyApp and MainView with ES6

gzzhanghao:
There are lots of things we need to deal with, but let's start with components first :)


web/src/js/components/flowview/index.js
```diff
@@ -89,7 +89,6 @@ var FlowView = React.createClass({
             } else {
                 active = tabs[0];
             }
-            this.selectTab(active);
```
mhils: Was this change intentional?

mhils:
Looks great, happy to see you pushing forward!

Any explanation why your compiled JS and CSS files are so much smaller?

gzzhanghao:
@mhils I used `gulp prod` to minimize script files, we should do this everytime we push a commit, but I didn't realize that before :P

web/src/js/components/flowview/index.js
gzzhanghao: I'd like to keep `render` pure.
Also, it leads to an issue when `selectedFlow` is not sync with `routerParams.flowId`

mhils: Ah yes - saw the purity issue as well, but didn't  get back to it. Good find.

PR #1235, opened 2016-06-09
virtualenv installation link fixed

bemre:


cortesi:
Hi there. Thanks for submitting this tweak - helping us keep on top of this sort of cleanliness issue is much appreciated.

I see we already have the virtualenv link listed as an external link at the bottom of the README - and we should keep consistent with that. It looks like the issue has to do with the "here" identifier being previously used for a link. Good old ReStructuredText. Could I perhaps ask you to change this PR to use an external link definition with a more sensible name than "here" - maybe "virtualenvdocs" or some such.

PR #1238, opened 2016-06-10
Distinguish between CI and interactive versions of tox

cortesi:
- This patch changes tox so that local invocation runs the tests
with paralellism, coverage disabled, and without attempting to report coverage
upstream.

- You can now run specific tests for the py27 environment:

    tox -e py27 ./path/to/my/test

We can't do this for py35 just yet, because of our junk filter, and the fact
that we specify which tests to run directly in the conf file. Once these are
cleaned up, we can extend.


tox.ini
```diff
@@ -9,11 +9,23 @@ passenv = CI TRAVIS_BUILD_ID TRAVIS TRAVIS_BRANCH TRAVIS_JOB_NUMBER TRAVIS_PULL_
...
+[testenv:py27-ci]
+basepython = python2.7
+commands =
   py.test --cov netlib --cov mitmproxy --cov pathod --color=yes --timeout 60 ./test
   codecov -e TOXENV
 
-[testenv:py35]
+[testenv:py35-ci]
 # remove bash & pipe & grep hack after cryptography ships with openssl 1.1.0
+basepython = python3.5
```
Kriechi: Why did you set `basepython` here? Isn't this auto-detected (at least on my system)?

Kriechi:
I think it might be cleaner to use `{posargs:default}` instead of duplicating all the environments.
Meaning: For `py.test` we add a posargs which defaults to `-n 8` and other settings we want locally.
And on travis we simply overwrite this with `''` or something.
http://tox.readthedocs.io/en/1.0/example/general.html#interactively-passing-positional-arguments

tox.ini
cortesi: Hmm... ok, that does work - I guess Travis picks up the environment prefix and matches it against its default environments. I'd be ok with removing this line.

cortesi:
Using postargs doesn't help with things like the coverage submissions, and I'd like to use the postargs to specify specific tests to run without having to re-type all the options (i.e. tox -e py27 ./my/test/foo.py::bar). I think the interactive and CI variants of the commands are different enough to justify two entries.

Kriechi:
One last thing - should we extract the py3-test-enabled files into a ENV variable at the top?
Right now we have to touch two lines to add a new file...

cortesi:
The Appveyor failure here is unrelated, as far as I can see - merging anyway.

PR #1239, opened 2016-06-11
Beef up our debugging capabilities

cortesi:
Improve debugging of thread and other leaks

- Add basethread.BaseThread that all threads outside of test suites should use
- Add a signal handler to mitmproxy, mitmdump and mitmweb that dumps resource information to screen when SIGUSR1 is received.
- Improve thread naming throughout to make thread dumps understandable
- Add a --sysinfo flag to all daemons. This dumps all the platform information and mitmproxy version data we'd normally need to troubleshoot an issue.

### DIFF
--- a/netlib/debug.py
+++ b/netlib/debug.py
@@ -1,0 +1,26 @@
+import platform
+from netlib import version
+
+"""
+    Some utilities to help with debugging.
+"""
+
+def sysinfo():
+    data = [
+        "Mitmproxy verison: %s"%version.VERSION,
+        "Python version: %s"%platform.python_version(),
+        "Platform: %s"%platform.platform(),
+    ]
+    d = platform.linux_distribution()
+    if d[0]:
+        data.append("Linux distro: %s %s %s"%d)
+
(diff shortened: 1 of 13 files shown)


netlib/debug.py
```diff
@@ -0,0 +1,76 @@
...
+import platform
+
+import psutil
+
+from netlib import version
+
+
+def sysinfo():
+    data = [
+        "Mitmproxy verison: %s" % version.VERSION,
```
Kriechi: **version**

netlib/debug.py
```diff
@@ -0,0 +1,76 @@
...
+        "Python version: %s" % platform.python_version(),
+        "Platform: %s" % platform.platform(),
```
Kriechi: Can we add the openssl version as well?
`from OpenSSL import SSL; print(SSL.SSLeay_version(SSL.SSLEAY_VERSION))`

cortesi:
Closing to re-create, and hopefully un-confuse Travis.

PR #1242, opened 2016-06-11
fix #1240

### LINKED ISSUE #1240: Http2SingleStreamLayer threads leak

1. run a mitimdump with the new debug functionality that may or may not be merged by the time you look at this
2. Fire up a browser going through mitmdump, and visit our old nemesis: https://http2.akamai.com/demo
3. Exit the browser
4. Send a SIGUSR1 to mitmdump to trigger debug output `pkill -SIGUSR1 -f mitmdump`

I expect no HTTP service threads to remain.

##### What went wrong?

Many HTTP service threads remain. 

This may be related to #1148

---

> mitmdump --sysinfo
Mitmproxy verison: 0.18
Python version: 2.7.11
Platform: Darwin-15.5.0-x86_64-i386-64bit
Mac version: 10.11.5 ('', '', '') x86_64

Kriechi:


cortesi:
I still see leaking Http2SingleStreamLayer threads with this patch applied. I _may_ be seeing fewer, but they're still there.

Kriechi:
After applying this path, all stale threads vanish within a couple of seconds on my system.
Seems to be a Heisenbug? Or do you have any idea to make it reproducible?
Do you think it's possible to write a (generic) test - to check if all threads exited after a test is done?

cortesi:
The work I've been doing in netlib.tcp and pathod do _exactly_ that. Have a look at this branch on my personal repo:

https://github.com/cortesi/mitmproxy/tree/netlibrace

My current plan is:

- Extend the thread leak detection I've been painstakingly rolling out in pathod to the rest of the project. In this case, I've tackled netlib, and the http2 tests are the next thing that's failing.
- Use context handlers, bulit on what I've added to netlib.tcp, to restructure our code and make sure we rigorously release resources everywhere.

I've picked up a ton of bugs doing this so far, so things are looking good. I'm going to be super busy for the next few days, so I might not have time to turn back to this for a bit. Feel free to push on with this, or to wait for me to get back :)

Kriechi:
Let's merge this for now - since the biggest bugs should be fixed with this PR.
Then we can hunt down the rest!

PR #1250, opened 2016-06-13
Add click integration for inline scripts

mhils:
Some developers in this community tend to write reasonably complex inline scripts (*hides*), which can be configured using additional arguments. This PR adds `click` support for the `start` script hook, so that we can use `click` to parse arguments:

```python
import click

@click.command()
@click.argument("outfile", type=click.File("wb"))
def start(outfile):
    # ...
```

### DIFF
--- a/docs/scripting/inlinescripts.rst
+++ b/docs/scripting/inlinescripts.rst
@@ -207,6 +207,8 @@ The arguments are then exposed in the start event:
    :caption: examples/modify_response_body.py
    :language: python
 
+Mitmproxy also supports `integration`_ with `click`_ for command line interfaces.
+
 Running scripts on saved flows
 ------------------------------
 
@@ -229,3 +231,5 @@ By default, spaces are interpreted as a separator between the inline script and 
 quotes if it contains spaces: ``-s '\'./foo bar/baz.py\' 42'``.
 
 .. _GitHub: https://github.com/mitmproxy/mitmproxy
+.. _integration: https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/data/scripts/click_simple.py
+.. _click: http://click.pocoo.org/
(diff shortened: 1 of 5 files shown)


cortesi:
I don't think this pays for itself when you can just to this:

```python
import click

@click.command()
@click.argument("outfile", type=click.File("wb"))
def cli(outfile):
    print(outfile)

def start(ctx, args):
    cli.main(args=args[1:])
```

I do think we need to have a broader discussion about how we want to handle script arguments in the GUIs. One could make an argument for integrating Click if, for instance, you could inspect the decorator on start() and generate an options pane for scripts in mitmproxy console and on the web automatically.

mhils:
Your comment is actually a good illustration why I think this 10 line addition is useful enough to have it. Just passing `args[1:]` isn't enough, because click will `sys.exit()` unless `standalone_mode=False` is passed. Error messages are pretty terrible if you don't specify `prog_name`. There's also no way to access `context` in the code you posted, so you'd need to define `cli` in the start hook an then invoke it.

Of course, a stronger integration of click (or something similar) would be even more favorable. I have no strong opinions on what would work best, although click's pretty friendly when it comes to introspection (e.g. `cli.params[0].name` works as you'd expect it to do).

cortesi:
These are good points, but I'd still like to push back a bit. ;)

- See #1252 for a proposal that cleans up the invocation, and addresses the script name problem.
- The fact that sys.exit() is called by click and that you need the standalone_mode argument is probably easier to document than explicit click support is. 
- We can pass the context in an environment variable on invocation. Passing complex objects like this has precedents in things like wsgi, and this would mean that we can support any argument parsing library rather than just click. 

In sum, I think all these problems have generic solutions that don't tie us to a specific parser. I'm still happy to talk about blessing click officially as the One True Way if we can use it to construct our GUIs, though.

mhils:
Superseded by https://github.com/mitmproxy/mitmproxy/pull/1254

PR #1254, opened 2016-06-14
Inline Scripts: use sys.argv instead of args argument.

mhils:
Continuation of #1252.


mhils:
On top of this, I'd also like to add https://github.com/mitmproxy/mitmproxy/commit/81a28c2f2062a7818b7ec15bb29a77b7c8ddb045, so that click integration is 100% seamless:

```python
@click.command()
@click.argument("outfile", type=click.File("wb"))
def start(outfile):
    # ...
```

PR #1255, opened 2016-06-14
Inline Scripts: Run click Scripts in Standalone Mode

mhils:
As threatened in https://github.com/mitmproxy/mitmproxy/pull/1254#issuecomment-225755577, I'd still like to have a minimal `click` integration for the script module. I regularly run into click's `sys.exit()ing` behaviour while livereloading scripts, and this should make everything work out of the box. 

I'm happy to make this an undocumented feature so that we can remove it if it ever causes any issues (which I doubt).

### DIFF
--- a/mitmproxy/script/script.py
+++ b/mitmproxy/script/script.py
@@ -15,6 +15,11 @@ import warnings
 
 import six
 
+try:
+    import click
+except ImportError:
+    click = False
+
 from mitmproxy import exceptions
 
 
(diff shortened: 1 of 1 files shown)


cortesi:
I really don't think we should do this, Max - especially since I'm in the process of refactoring scripts entirely to unify things with handlers. Handlers will grow an analogous "start" handler that this code isn't applicable to and it's just not worth catering this specifically for a single mis-remembered option in click.

Let's defer this until after the context refactoring, and discuss it along with the idea of really integrating click as our preferred option handler.

mhils:
Ok, let's continue here once #1265 is done.

PR #1256, opened 2016-06-14
added flowMenu

cle1000:


web/src/js/components/Header.js
```diff
@@ -1,56 +1,66 @@
...
+        const { settings, updateLocation, query, selectedFlow, active_menu} = this.props
         return (
             <header>
                 <nav className="nav-tabs nav-tabs-lg">
                     <FileMenu/>
                     {Header.entries.map(Entry => (
                         <a key={Entry.title}
                            href="#"
-                           className={classnames({ active: Entry === Active })}
+                           className={classnames({ active: Entry.title === active_menu, hidden: !selectedFlow && Entry === FlowMenu })}
```
mhils: This looks overly complicated. How about we just do this?
```javascript
let entries = [...Header.entries]
if(selectedFlow)
    entries.push(FlowMenu)
```

web/src/js/components/Header.js
```diff
@@ -1,56 +1,66 @@
...
                     ))}
                 </nav>
                 <div className="menu">
-                    <Active
-                        ref="active"
-                        settings={settings}
-                        updateLocation={updateLocation}
-                        query={query}
-                    />
+                    {Header.entries.map(Entry => (
```
mhils: What happened here?

web/src/js/ducks/view.js
```diff
@@ -0,0 +1,31 @@
...
+    switch (action.type) {
+        case ACTIVE_MENU:
+            return {
+                ...state,
+                active_menu: action.active_menu
+            }
+        case SELECT_FLOW:
+            return{
+                ...state,
+                active_menu: action.flowId ? 'Flow' : 'Start'
```
mhils: So, we probably don't want to change the tab when we switch between flows. For example, when I select a flow, then select the start tab, then press arrow up to select the flow above, I don't want to have the "Flow" tab again.
Unfortunately, this brings up the problem that we'd actually need the previous flow state here, but we only have the view state. This is tricky: http://redux.js.org/docs/FAQ.html#reducers-share-state

Possible workarounds:
1. Custom rootReducer
2. Transfer previous state as part of the action (we can access the full state in the actionCreater using redux-thunk's `getState`.
3. Keep the UI state in React and use `componentWillReceiveProps` (http://redux.js.org/docs/FAQ.html#organizing-state-only-redux-state)

My gut feeling is that 2 or 3 should be the way to go. Thoughts?

web/src/js/components/Header.js
cle1000: it shows the content for each menu item,` settings, updateLocation, query` is used by `MainMenu`, `settings` is used by `OptionsMenu`

mhils: It looks like each menu should be `connect`ed to the store directly?
Nonwithstanding, my comment was rather related to the fact that  `Active` went missing according to the diff`? :stuck_out_tongue:

cle1000: yes each menu should be connected directly. :)

mhils: Right - let's try to keep that in a separate PR though :)

web/src/js/ducks/view.js
cle1000: yes i understand that problem, i think i just use the privious state and go for 2.

web/src/js/ducks/view.js
```diff
@@ -0,0 +1,36 @@
...
+}
+export default function reducer(state = defaultState, action) {
+    switch (action.type) {
+        case ACTIVE_MENU:
+            return {
+                ...state,
+                active_menu: action.active_menu
+            }
+        case SELECT_FLOW:
+            if (!action.currentSelection != !action.flowId){
```
mhils: What's the point of this check?

mhils: Nevermind - didn't see the `!`. We can make this code more clear by introducing variables for the conditions that describe our intent, e.g.  
```
let isNewSelection = (action.flowId && !action.currentSelection)
if(isNewSelection){
  // ...
}
```

Besides, we don't have a DESELECT event yet. I'm wondering if it's nicer to have `SELECT_FLOW` with `flowId=undefined` or a dedicated event for that. I'll leave that up to you. :smiley:

web/src/js/ducks/view.js
```diff
@@ -0,0 +1,36 @@
+import ReduxThunk from 'redux-thunk'
+
+import { SELECT_FLOW } from './flows'
+const ACTIVE_MENU = 'ACTIVE_MENU'
```
mhils: This is an action, so it should be `SET_*`

web/src/js/ducks/index.js
```diff
@@ -2,11 +2,13 @@ import {combineReducers} from 'redux'
 import eventLog from './eventLog'
 import websocket from './websocket'
 import flows from './flows'
+import view from './view'
 
 const rootReducer = combineReducers({
     eventLog,
     websocket,
     flows,
+    view
```
mhils: Can we call this `ui` ? `view` kind of overlaps with the flow view.

web/src/js/ducks/index.js
```diff
@@ -2,11 +2,13 @@ import {combineReducers} from 'redux'
 import eventLog from './eventLog'
 import websocket from './websocket'
 import flows from './flows'
+import ui from './ui.js'
```
mhils: Minor nitpick: We can/should skip the extension here.

web/src/js/ducks/ui.js
```diff
@@ -0,0 +1,37 @@
...
+export default function reducer(state = defaultState, action) {
+    switch (action.type) {
+        case SET_ACTIVE_MENU:
+            return {
+                ...state,
+                active_menu: action.active_menu
+            }
+        case SELECT_FLOW:
+            let isNewSelection = (action.flowId && !action.currentSelection)
+            let isSelectAction = action.flowId
```
mhils: What's the point of `isSelectAction` when `isNewSelection` already assures that `action.flowId` is true? The `false` case can never get triggered currently.

mhils:
I fixed the minor isues above (94e2929b3fd6dc335d1c787167956df18f05494e), merged! :smiley:

PR #1257, opened 2016-06-14
Make har_extractor.py output HAR 1.2 spec-compliant

tdickers:
HAR files were failing to load in harviewer (http://www.softwareishard.com/har/viewer/) due to:
- ISO 8601 dates for startedDateTime missing timezone (http://www.softwareishard.com/blog/har-12-spec/#pages)
-- Used UTC but could add detection of default system timezone if desired
- pages object missing pageTimings (http://www.softwareishard.com/blog/har-12-spec/#pageTimings)
-- Used null because all child fields are optional

### DIFF
--- a/examples/har_extractor.py
+++ b/examples/har_extractor.py
@@ -4,6 +4,7 @@
 """
 import six
 import sys
+import pytz
 from harparser import HAR
 
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


tdickers:
Working on conflicts & getting py.test passing.

mhils:
Thanks! :smiley: :cake: :tada: 
The HAR extractor is not really in a good shape at the moment, any improvements are welcome. :relaxed:

Issue #1101, opened 2016-04-29
clarify error message on SSLv3 failure


romainfun:
openssl SSLv3 support has been disabled on ubuntu 16.04 ([debian/patches/no-sslv3.patch](https://launchpad.net/ubuntu/+source/openssl/1.0.2g-1ubuntu1)),<br />and probably debian unstable + other distros.

`mitmproxy --ssl-version-server SSLv3 -R https://google.fr:443`

results in the following failure :

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/libmproxy/proxy/server.py", line 120, in handle
    root_layer()
  File "/usr/lib/python2.7/dist-packages/libmproxy/proxy/modes/reverse_proxy.py", line 14, in __call__
    layer()
  File "/usr/lib/python2.7/dist-packages/libmproxy/protocol/tls.py", line 267, in __call__
    self._establish_tls_with_client_and_server()
  File "/usr/lib/python2.7/dist-packages/libmproxy/protocol/tls.py", line 392, in _establish_tls_with_client_and_server
    six.reraise(*sys.exc_info())
  File "/usr/lib/python2.7/dist-packages/libmproxy/protocol/tls.py", line 386, in _establish_tls_with_client_and_server
    self._establish_tls_with_server()
  File "/usr/lib/python2.7/dist-packages/libmproxy/protocol/tls.py", line 458, in _establish_tls_with_server
    alpn_protos=alpn,
  File "/usr/lib/python2.7/dist-packages/libmproxy/models/connections.py", line 183, in establish_ssl
    self.convert_to_ssl(cert=clientcert, sni=sni, **kwargs)
  File "/usr/lib/python2.7/dist-packages/netlib/tcp.py", line 605, in convert_to_ssl
    **sslctx_kwargs
  File "/usr/lib/python2.7/dist-packages/netlib/tcp.py", line 580, in create_ssl_context
    **sslctx_kwargs)
  File "/usr/lib/python2.7/dist-packages/netlib/tcp.py", line 488, in _create_ssl_context
    context = SSL.Context(method)
  File "/usr/local/lib/python2.7/dist-packages/OpenSSL/SSL.py", line 478, in __init__
    _raise_current_error()
  File "/usr/local/lib/python2.7/dist-packages/OpenSSL/_util.py", line 48, in exception_from_error_queue
    raise exception_type(errors)
Error: [('SSL routines', 'SSL_CTX_new', 'null ssl method passed')]

[...]

PR #1258, opened 2016-06-14
Solving the SSL version exception

Fronchetti:
Trying to solve #1101


Kriechi:
Thanks for the PR! We are always happy about contributions!

Does this really affect every `TlsException` possible? I think this might be to generic.
We should check for the pyOpenSSL exception and only target this if possible.

Fronchetti:
Thank you for the reply and the receptivity@Kriechi 
Reading the traceback this was the best idea I had, but thinking about it this way, a general exception would be the best solution.
What do you suggest?

PR #1264, opened 2016-06-16
added tests for ui

cle1000:


web/src/js/__tests__/ducks/ui.js
```diff
@@ -0,0 +1,34 @@
...
+
+
+describe("ui reducer", () => {
+    it("should return the initial state", () => {
+        expect(reducer(undefined, {})).toEqual({ activeMenu: 'Start'})
+    }),
+    it("should return the state for view", () => {
+        expect(reducer(undefined, setActiveMenu('View'))).toEqual({ activeMenu: 'View'})
+    }),
+    it("should change the state to Start", () => {
```
mhils: Let's be a bit more explicit about this: "should change the state to Start when deselecting a flow and we are currently on the flow tab"

(same below)

web/src/js/ducks/ui.js
```diff
@@ -1,5 +1,5 @@
-import { SELECT_FLOW } from './flows'
-const SET_ACTIVE_MENU = 'SET_ACTIVE_MENU'
+export const SELECT_FLOW = "SELECT_FLOW"
```
mhils: This looks wrong, I think we don't want to duplicate the constant?

cle1000: i changed it in the new commit, but i got an error with i cannot solve:
`Error: Cannot find module '../components/FlowTable/FlowColumns' from 'flows.js'`

if i change this line in js/ducks/flows.js 
`import * as columns from "../components/FlowTable/FlowColumns";
`to
`import * as columns from "../components/FlowTable/FlowColumns.jsx";
`
i got this error:
```
  - TypeError: Cannot read property 'reduceList' of undefined
        at Object.<anonymous> (src/js/ducks/flows.js:103:1542)
```

mhils:
LGTM, except for the conflicts. Let's see if we can put app.js and vendor.js into `.gitignore` (https://mitmproxy.slack.com/archives/general/p1466217848000398)

cle1000:
i changed it in the new commit, but i got an error with i cannot solve:
Error: Cannot find module '../components/FlowTable/FlowColumns' from 'flows.js'

if i change this line in js/ducks/flows.js 
import * as columns from "../components/FlowTable/FlowColumns"; to
import * as columns from "../components/FlowTable/FlowColumns.jsx"; 
i got this error:

  - TypeError: Cannot read property 'reduceList' of undefined
        at Object.<anonymous> (src/js/ducks/flows.js:103:1542)

mhils:
Cherry-picked and merged this.

PR #1266, opened 2016-06-17
[web] Settings ducks

gzzhanghao:
Add settings ducks 🐥


web/src/js/components/ProxyApp.jsx
```diff
@@ -60,30 +52,16 @@ class ProxyAppMain extends Component {
         return _.clone(this.props.location.query)
     }
 
+    componentWillMount() {
+        this.props.fetchSettings();
+    }
```
mhils: This is kind of the same "where to put it" question as with the websockets - let's discuss this in #1268 first.

web/src/js/ducks/settings.js
```diff
@@ -0,0 +1,92 @@
...
+export const FETCH_ERROR = 'SETTINGS_FETCH_ERROR'
+export const RECV_WS_MSG = 'SETTINGS_RECV_WS_MSG'
+
+const defaultState = { settings: {}, pendings: null, req: null }
+
+export default function reduce(state = defaultState, action) {
+    switch (action.type) {
+
+        case BEGIN_FETCH:
+            return { ...state, pendings: [], req: action.req }
```
mhils: Same issue as with the websockets: req is not serializable nor immutable.

web/src/js/connection.js
```diff
@@ -28,6 +30,8 @@ export default function Connection(url, dispatch) {
                 return dispatch(eventLogActions.updateLogEntries(message))
             case flowActions.UPDATE_FLOWS:
                 return dispatch(flowActions.updateFlows(message))
+            case settingsActions.WS_MSG_TYPE:
```
mhils: Don't we just want to call this `UPDATE` or `UPDATE_SETTINGS` ?

web/src/js/ducks/settings.js
```diff
@@ -0,0 +1,92 @@
...
+            return _.merge({}, data, action.data)
+
+        default:
+            return data
+    }
+}
+
+export function fetch() {
+    return dispatch => {
+        const req = $.getJSON('/settings')
```
mhils: let's use `fetch/fetchApi` - I want to remove the jQuery dependency at some point. :relaxed:

web/src/js/ducks/settings.js
```diff
@@ -0,0 +1,92 @@
+import { StoreCmds } from '../actions'
+import { addLogEntry } from './eventLog'
+
+export const WS_MSG_TYPE = 'settings'
+export const WS_MSG_CMD_RESET = 'reset'
+export const WS_MSG_CMD_UPDATE = 'update'
```
mhils: This seems duplicate. Can we share the same constants for both events emitted on the python end and events emitted in JS? Happy to help you with the Python part. :smiley:

web/src/js/ducks/settings.js
```diff
@@ -0,0 +1,92 @@
...
+}
+
+function reduceData(data, action) {
+    switch (action.cmd) {
+
+        case WS_MSG_CMD_RESET:
+            return action.data || {}
+
+        case WS_MSG_CMD_UPDATE:
+            return _.merge({}, data, action.data)
```
mhils: We can use `{...data, ...action.data}` :smiley:

web/src/js/ducks/settings.js
```diff
@@ -0,0 +1,92 @@
...
+                return { ...state, pendings: state.pendings.concat(action) }
+            }
+            return { ...state, settings: reduceData(state.settings, action) }
+
+        default:
+            return state
+    }
+}
+
+function reduceData(data, action) {
```
mhils: Can we combine this function with the reducer? The reducer can just call itself for the pending actions, see the list reducer. :smiley:

web/src/js/ducks/settings.js
```diff
@@ -0,0 +1,92 @@
...
+
+        dispatch({ type: BEGIN_FETCH, req })
+
+        return req
+    }
+}
+
+export function handleWsMsg(msg) {
+    return (dispatch, getState) => {
+        if (msg.cmd !== StoreCmds.RESET) {
```
mhils: AFAIK, we're currently never calling `RESET` on the settings, so we can simplify this a lot! :tada:

web/src/js/ducks/settings.js
gzzhanghao: @mhils I'd love to, but fetch API is not cancelable so we'll need to have a workaround for that

web/src/js/ducks/settings.js
gzzhanghao: I think it's better to have these constants somewhere, it's not a good idea to have magic strings in logic codes.

mhils: I agree on this. My point was to merge `WS_MSG_CMD_RESET` and `RESET` and `WS_MSG_CMD_UPDATE` and `UPDATE`.

web/src/js/ducks/settings.js
gzzhanghao: But that does not merge deep properties :pensive:

mhils: That's true, the settings are flat though! :smiley:

web/src/js/ducks/settings.js
gzzhanghao: But we should keep the reference somewhere, I prefer to save it in state instead of a global variable.

web/src/js/connection.js
gzzhanghao: It's a config entry but not an action type. So I think it's better to have a MSG_TYPE prefix for it.

web/src/js/ducks/settings.js
gzzhanghao: They are two different things. `RESET` is an action type and `WS_MSG_CMD_RESET` is a config option. I think keeping them separate can improve maintainability.

web/src/js/ducks/settings.js
gzzhanghao: Won't it make things more complicated? I prefer to separate functions AMAP.

web/src/js/ducks/settings.js
gzzhanghao: That's a good news :tada:

web/src/js/ducks/settings.js
gzzhanghao: :tada:

PR #1267, opened 2016-06-17
[web] Working on components

gzzhanghao:
Now all components are in es6 :tada:


web/src/js/components/ContentView.jsx
```diff
@@ -0,0 +1,78 @@
...
+    isContentTooLarge(msg) {
+        return msg.contentLength > 1024 * 1024 * (ViewImage.matches(msg) ? 10 : 0.2)
+    }
+
+    render() {
+        const { flow, message } = this.props
+        const { displayLarge, View } = this.state
+
+        if (message.contentLength === 0) {
+            return <ContentErrors.ContentEmpty {...this.props}/>
```
mhils: Technically these are not errors, but we can change that later on.

web/src/js/components/EventLog.jsx
```diff
@@ -1,31 +1,70 @@
...
+    static defaultProps = {
+        defaultHeight: 200,
+    }
+
+    constructor(props, context) {
+        super(props, context)
+
+        this.state = { height: this.props.defaultHeight }
+
+        this.onDragStart = this.onDragStart.bind(this)
```
mhils: What is this about?

web/src/js/components/FlowView/Details.jsx
```diff
@@ -0,0 +1,133 @@
...
+export function ConnectionInfo({ conn }) {
+    return (
+        <table className="connection-table">
+            <tbody>
+                <tr key="address">
+                    <td>Address:</td>
+                    <td>{conn.address.address.join(':')}</td>
+                </tr>
+                {conn.sni ? (
+                    <tr key="sni"></tr>
```
mhils: We removed this for TimeStamp but not for ConnectionInfo. If the react bug is actually gone, we should do the same here.

web/src/js/components/FlowView/Details.jsx
```diff
@@ -0,0 +1,133 @@
...
+    // @todo We should fetch human-readable certificate representation from the server
+    return (
+        <div>
+            {flow.client_conn.cert && [
+                <h4 key="name">Client Certificate</h4>,
+                <pre key="value" style={{ maxHeight: 100 }}>{flow.client_conn.cert}</pre>
+            ]}
+
+            {flow.server_conn.cert && [
+                <h4 key="name">Server Certificate</h4>,
```
mhils: Does this yield duplicate keys?

web/src/js/components/FlowView/Details.jsx
```diff
@@ -0,0 +1,133 @@
...
+        }, {
+            title: "First request byte",
+            t: req.timestamp_start,
+        }, {
+            title: "Request complete",
+            t: req.timestamp_end,
+            deltaTo: req.timestamp_start
+        }, res && {
+            title: "First response byte",
+            t: res.timestamp_start,
```
mhils: Did you test this? It looks like this will fail if res is `undefined`.

web/src/js/components/Prompt.jsx
```diff
@@ -0,0 +1,71 @@
...
+                        {prefix}<strong className="text-primary">{opt.key}</strong>
+                        {idx !== -1 ? opt.text.substring(idx + 1) : ')'}
+                    </span>
+                )
+            })}
+            </div>
+        </div>
+    )
+
+    function onKeyDown(event) {
```
mhils: It's impressive that this works, but we probably want to move it before the return statement :wink:

web/src/js/components/EventLog.jsx
mhils: Ok, I saw that you removed the splitter and added it here instead. What's the motivation for this? I think we want to keep that separate and maybe make the Splitter nicer at a later point in time.

gzzhanghao: @mhils Its UX is much better now. I agree that we need to keep splitter separate but I'd like to have it here before we find a better solution for this.

mhils: Ok, a temporary move is fine. I'll create new issues for all other comments at mitmproxy/mitmweb, the most important part is that we get this monster merged.

web/src/js/components/FlowView/Details.jsx
gzzhanghao: Not sure about it, I think react just asks for unique keys for children in the same array

web/src/js/components/Prompt.jsx
gzzhanghao: But then you will have to walk through the entire function to get the DOM structure. Personally I'd like to promote the DOM structure AMAP :)

mhils: Not sure if I get your point - we can just move the function block above the return statement and nothing should change? :wink:

PR #1268, opened 2016-06-17
[web] WebSocket ducks

gzzhanghao:
Move connection stuffs into websocket ducks :hatched_chick:


web/src/js/components/ProxyApp.jsx
```diff
@@ -37,6 +38,10 @@ class ProxyAppMain extends Component {
         this.onSettingsChange = this.onSettingsChange.bind(this)
     }
 
+    componentWillMount() {
+        this.props.wsConnect()
```
mhils: I'm not entirely convinced - what's the benefit of moving the WebSocket connection into React?

web/src/js/ducks/websocket.js
```diff
@@ -1,30 +1,103 @@
...
 
-const defaultState = {
-    connected: false,
-    /* we may want to have an error message attribute here at some point */
-}
 export default function reducer(state = defaultState, action) {
     switch (action.type) {
+
+        case CONNECT:
+            return { ...state, socket: action.socket }
```
mhils: We talked about this before: State should be serializable and immutable (see http://redux.js.org/docs/introduction/ThreePrinciples.html).
What's your argument for moving the WebSocket into state?

web/src/js/ducks/websocket.js
```diff
@@ -1,30 +1,103 @@
...
             return state
     }
 }
 
+export function connect() {
+    return dispatch => {
+        const socket = new WebSocket(location.origin.replace('http', 'ws') + '/updates')
+
+        // @todo remove this
+        window.ws = socket
```
mhils: (fwiw, this just exists for debugging and is not used by any parts of the code)

web/src/js/ducks/websocket.js
```diff
@@ -1,30 +1,103 @@
...
+                return dispatch(eventLogActions.updateLogEntries(data))
+
+            case flowActions.UPDATE_FLOWS:
+                return dispatch(flowActions.updateFlows(data))
+
+            default:
+                console.warn('unknown message', data)
+        }
+
+        dispatch({ type: MESSAGE, msg })
```
mhils: The problem I see with this generic approach is that the server may send messages which do no modify any redux state, but we won't notice that, as the action just passes by everything and does not cause any change.
Is there anything you did not like about the `flowActions.updateFlows` etc approach, which provided one dedicated actionCreator for each type of websocket message?

web/src/js/components/ProxyApp.jsx
gzzhanghao: I'm just not comfort with initializing things out of the react - redux architecture. No convincing reason for this :pensive:

web/src/js/ducks/websocket.js
gzzhanghao: Again, I think we should place the ws object somewhere, rather than making it unmaintainable.

How about putting the ws object into a Symbol entry so the state is serializable and we can keep the ws object across action creators? We just treat the ws object a Symbol and don't read / write properties of it, so we don't need to worry about immutability.

web/src/js/ducks/websocket.js
gzzhanghao: There's a `console.warn` and a `MESSAGE` action for that case :smile:
I don't like to packaging action creators and reducers into utils. I'll have to walk through a 166 lines function to find out how flows or eventLog works. IMO they should be much simpler. :confused:

mhils:
As I mentioned yesterday, this needs tests. The state transitions aren't complete yet, e.g. an errored connection won't set `connected` to `false`. We also agreed on using a Symbol for the WebSocket.

gzzhanghao:
@mhils if I understand the "errored connection" correctly, the websocket will emits a close event after connection errors, and that event will set `connected` to `false` correctly.

mhils:
As I understand it, a close event triggers `onDisconnect`, which only emits `DISCONNECTED`. `DISCONNECTED` does not set `connected` to `false` - only `DISCONNECT` does.

gzzhanghao:
Alright, fixed :wink:

PR #1270, opened 2016-06-18
[web] Settings Ducks 2

mhils:
This builds on top of #1266. :hatched_chick: 

@gzzhanghao, what do you think?


web/src/js/ducks/settings.js
```diff
@@ -1,64 +1,72 @@
...
+        case RECEIVE_SETTINGS:
+            let s = {
+                settings: action.settings,
+                isFetching: false,
+                actionsDuringFetch: [],
+            }
+            for (action of state.actionsDuringFetch) {
+                s = reducer(s, action)
+            }
+            return s
```
gzzhanghao: maybe something like:
```javascript
state.actionsDuringFetch.reduce(reducer, { settings: action.settings, isFetching: false, actionsDuringFetch: [] })
```

web/src/js/ducks/settings.js
```diff
@@ -1,64 +1,72 @@
...
-
-const defaultState = { settings: {}, pendings: null }
+const defaultState = {
+    settings: {},
+    isFetching: false,
+    actionsDuringFetch: [],
+}
 
-export default function reduce(state = defaultState, action) {
+export default function reducer(state = defaultState, action) {
```
gzzhanghao: I prefer verbs for functions and nouns for classes :)

web/src/js/ducks/settings.js
```diff
@@ -1,64 +1,72 @@
-import { addLogEntry } from './eventLog'
+import {fetchApi} from "../utils";
 
-export const WS_MSG_TYPE = 'settings'
-export const WS_MSG_CMD_UPDATE = 'update'
```
gzzhanghao: As I said, these constants are options but not action type, so I think we should keep them here

web/src/js/ducks/settings.js
mhils: I considered that as well, but I believe the for loop is more readable (see: http://stackoverflow.com/a/181706)

web/src/js/ducks/settings.js
mhils: Agree, I'm just doing this for https://github.com/erikras/ducks-modular-redux#rules

web/src/js/ducks/settings.js
mhils: `updateSettings` is the only occasion where we'd ever want to use them, so I think this is fine. If we ever come up with a second usage, we should put it in a constant ASAP.

PR #1272, opened 2016-06-18
mitmproxy/models: missed comma in __all__

dwfreed:
Spotted by Landscape (thanks @Kriechi for pointing out the site).  Add a
comma to the last item too, to prevent this from happening in the future
and reduce messing up blame later.


mhils:
Thanks! :smiley:

PR #1274, opened 2016-06-19
add Discourse to README

Kriechi:


mhils:
I like the general idea.

- Let's put Slack below Discourse, because people just click on the first thing they see.
- We can have a badge for discourse, e.g. this: . `shields.mitmproxy.org` is our on-premise version of `shields.io` (which fails all the time). The API is the same, except that it's /api/ for custom badges on our host (don't ask me why).
- I personally prefer the more spaced version (with badges in its own line) over the condensed one in this PR as the badges don't line up with text :stuck_out_tongue:

Kriechi:
done.

PR #1275, opened 2016-06-19
py3++

Kriechi:
/cc @dufferzafar


dufferzafar:
This looks good. I don't get why the codecov thing fails though.

PR #1276, opened 2016-06-19
mitmproxy.console tests - PathEdit

dufferzafar:
I've not used mock/patch before so I'm not sure if this is the correct
way to test the things.

### DIFF
--- a/test/mitmproxy/console/test_pathedit.py
+++ b/test/mitmproxy/console/test_pathedit.py
@@ -2,6 +2,8 @@ import os
 from os.path import normpath
 from mitmproxy.console import pathedit
 
+from mock import patch
+
 from .. import tutils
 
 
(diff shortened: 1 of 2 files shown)


mhils:
This looks good. You can also use `@patch()` decorators which (IMO) make the code a tiny bit more readable, but both approaches are fine!

test/mitmproxy/console/test_pathedit.py
```diff
@@ -47,3 +49,25 @@ def test_completion(self):
         s = "thisisatotallynonexistantpathforsure"
         assert c.complete(s) == s
         assert c.final == s
+
+
+class TestPathEdit():
+
+    def test_keypress(self):
+
+        pe = pathedit.PathEdit()
```
dufferzafar: Actually, I couldn't get the `@patch()` decorators to work. If you'd notice, i create a `PathEdit` object *before* doing the patching or else weird errors were showing up. :confused:

PR #1281, opened 2016-06-21
WIP: Refactor mitmproxy core

cortesi:
I'm a few days away from having this ready for comment. I'm putting this in a PR to get CI going and to get the coming changes onto people's radar, especially since we're moving so quickly with the Python3 migration. I'm busy for the next few days, but will continue and expand this PR description as I go.

# The grand plan

- Add a unified API for addons and scripts
- Move nearly all complicated functionality out of the flow core, into a suite of built-in addon modules
- Add options.Options, a "single source of truth" to manage options state

# The current status

The basic schematic for the new addon and script interfaces is done, and part of the migration to builtin addons has been done. You can see the result [here](https://github.com/cortesi/mitmproxy/tree/core-refactor/mitmproxy/builtins). As you can see, the difference is huge - all code for a particular concern is in one place, and testing can be completely isolated from everything else. 
 
# Todo

- [x] First-pass design of the new context-based API
- [x] Add options.Options
- [x] Migrate anticache to addon
- [x] Migrate anticomp to addon
- [x] Migrate external scripts to addon
- [x] Migrate sticky auth to addon
- [x] Migrate sticky cookie to addon
- [x] Migrate streaming to file to addon
- [ ] Move server playback to addon
- [ ] Move replacehooks to addon
- [ ] Move setheaders to addon
- [ ] Move web apps to addon 
- [ ] Formalise addon testing, and move it to mitmproxy so our users can use it too
- [ ] Documentation and examples


mitmproxy/ctxmanager.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import absolute_import, print_function, division
+import six
+
+
+class Facade(object):
+    def __init__(self, getobj):
+        self._getobj = getobj
+
+    if six.PY3:
```
Kriechi: Using `six.PY3` is generally not a good idea - once Py4 is here, this will break...

mhils:
There are some great pieces in here, but I'm not happy with some design decisions:

1. Why is `flow` not an argument to `reponse()` etc. anymore? I think we both agreed on `def response(flow):`, which I still strongly favor.
2. We also agreed to provide backwards-compatibility for inline scripts which still accept `context, flow` as args. What happened to this?
3. What happend to the `@concurrent` decorator?
4. Why is `start` renamed to `configure`?

This PR touches way too many critical parts for me to be comfortable with. We should start splitting out parts ASAP.

mitmproxy/builtins/script.py
```diff
@@ -0,0 +1,156 @@
...
+def load_script(path, args):
+    ns = {'__file__': os.path.abspath(path)}
+    with scriptenv(path, args):
+        with open(path, "rb") as f:
+            code = compile(f.read(), path, 'exec')
+            exec(code, ns, ns)
+    return ns
+
+
+class ReloadHandler(watchdog.events.FileSystemEventHandler):
```
mhils: Why do we mix concerns here?
Reloading functionality was completely separated from script processing, this really looks like an unececssary interdependency.

mitmproxy/builtins/script.py
```diff
@@ -0,0 +1,156 @@
...
+        yield
+    except Exception:
+        _, _, tb = sys.exc_info()
+        scriptdir = os.path.dirname(os.path.abspath(path))
+        for i, s in enumerate(reversed(traceback.extract_tb(tb))):
+            if not os.path.abspath(s[0]).startswith(scriptdir):
+                break
+            else:
+                tb = tb.tb_next
+        ctx.log.warn("".join(traceback.format_tb(tb)))
```
mhils: Why do we supress exceptions here? This feels like an accident waiting to happen.

cortesi: Because scripts are never allowed to crash the core. Instead, script exceptions are turned into a log.

mitmproxy/builtins/script.py
cortesi: All script functionality is now isolated in one single module. Pulling out reloading functionality resulted in a clumsy and unclear API. This is way, way simpler.

cortesi:
And just relaying our discussion on slack to here:

1) The plan is indeed to move flow from context to an argument on the events that have flows
2) Backwards compatibility is not a tenable goal for a refactoring of this sort. For instance, then entire ScriptContext API is going away.
3) The concurrent decorator needs to be re-thought, and will come back in some form.
4) Start is renamed to configure because it's called whenever there is a configuration change, not just on start.

mitmproxy/ctx.py
```diff
@@ -0,0 +1,12 @@
...
+from mitmproxy import ctxmanager
+
+_flow = None
+_log = None
+_master = None
+
+
+flow = ctxmanager.Facade(lambda: _flow)
+log = ctxmanager.Facade(lambda: _log)
+master = ctxmanager.Facade(lambda: _master)
```
mhils: Can we avoid the whole Facade dance and just assign to `ctx.flow` etc. in `handlecontext`?

cortesi: No, because this allows users to do "from mitmproxy.ctx import master", and not be surprised. What we can do is change this interface to make this clearer by making the context objects callables, i.e. "master = mitmproxy.ctx.master()", but I'm unsure if this is worth it.

mitmproxy/builtins/script.py
mhils: This introduces an ugly race condition, as reloading events are not synced over the master queue anymore.

cortesi: Describe the race? I'm sure we can cope with it in a generic fashion, while keeping everything isolated in the script addon. There's no reason for the core to know about reload events.

mhils: The race is that `Script.run` gets executed while `Script.reload` is running.

mitmproxy/builtins/script.py
```diff
@@ -0,0 +1,156 @@
...
+        if self.ns is not None:
+            func = self.ns.get(name)
+            if func:
+                with scriptenv(self.path, self.args):
+                    func(*args, **kwargs)
+
+    def reload(self, master, options):
+        with master.handlecontext(None):
+            self.ns = None
+            self.configure(options)
```
mhils: A reload can also trigger new (syntax)errors, which are not caught anymore here.

cortesi: Errors should just be caught and written to log with log level warn.

dufferzafar:
Pulled your branch. Rebased with master. Ran already ported tests.

```diff
diff --git mitmproxy/controller.py mitmproxy/controller.py
index 1f2983f..eb7f40c 100644
--- mitmproxy/controller.py
+++ mitmproxy/controller.py
@@ -120,7 +120,7 @@ class Master(object):
                 handle_func = getattr(self, mtype)
                 if not callable(handle_func):
                     raise exceptions.ControlException("Handler %s not callable" % mtype)
-                if not handle_func.func_dict.get("__handler"):
+                if not handle_func.__dict__.get("__handler"):
                     raise exceptions.ControlException(
                         "Handler function %s is not decorated with controller.handler" % (
                             handle_func
@@ -209,7 +209,7 @@ def handler(f):
         with master.handlecontext(message):
             f(master, message)
             if handling:
-                master.addons(f.func_name)
+                master.addons(f.__name__)
 
         if handling and not message.reply.acked and not message.reply.taken:
             message.reply.ack()
```

cortesi:
Closing - I've been pulled away for a week or so by other events. Max is pushing forward with this elsewhere.

PR #1282, opened 2016-06-21
Python 3 - mitmproxy.{protocol.http1, proxy, platform.pf}

dufferzafar:


mitmproxy/models/http.py
```diff
@@ -22,8 +23,8 @@ def get_decoded_content(self):
         """
         ce = self.headers.get("content-encoding")
         if not self.content or ce not in encoding.ENCODINGS:
-            return self.content
-        return encoding.decode(ce, self.content)
+            return strutils.native(self.content)
```
dufferzafar: @mhils I couldn't decide whether `get_decoded_content` should return `bytes` or `str`, so instead of this we could keep returning bytes here, and I could make the change where the function is called!

Also, do i `strutils.native` or just plain `decode`?

mitmproxy/platform/pf.py
```diff
@@ -8,6 +8,7 @@ def lookup(address, port, s):
 
         Returns an (address, port) tuple, or None.
     """
+    s = s.decode()
```
dufferzafar: No harm in just decoding pfctl's output right?

mitmproxy/filt.py
```diff
@@ -155,9 +155,9 @@ class FHead(_Rex):
     flags = re.MULTILINE
 
     def __call__(self, f):
-        if f.request and self.re.search(bytes(f.request.headers)):
+        if f.request and self.re.search(bytes(f.request.headers).decode()):
```
mhils: mitmproxy.filt is unfortunately not as trivial to port as this may look.
I propose we open a separate issue for that to not clog up things here. :smiley:

mitmproxy/models/http.py
mhils: We have two encoding passes with HTTP:

1. The raw/gzipped content as bytes, as it is transmitted on the wire.
2. The decoded/ungzipped content as bytes: `decoded = raw.decode(content_encoding_header)`
3. The message text as str: `text = decoded.decode(content_type_header_charset)`

Mapping this to mitmproxy, 1 is `message.content`, 2 is `message.get_decoded_content()`, and 3 is `message.text`. 

Long story short, even though the function name states *decoded*, this function should return bytes (please add a type annotation!).
In the longer term, we probably want to replace all uses of this function with `decoded` (which itself should get a LRU cache).

mitmproxy/filt.py
dufferzafar: Okay, so I've removed `filt` related code from here! I think this can be merged now.

PR #1287, opened 2016-06-22
[web] WIP: Refactoring ducks/utils

gzzhanghao:
Roadmap:

- [x] eventLog
- [x] utils/list
- [x] utils/view
- [x] flows
- [x] components
- [ ] FlowActions


web/src/js/ducks/eventLog.js
```diff
@@ -1,80 +1,138 @@
-import makeList from "./utils/list"
-import {updateViewFilter, updateViewList} from "./utils/view"
+import { fetchApi as fetch } from '../utils'
```
mhils: Can we please not rename it so that it's always clear which function is used?

web/src/js/ducks/flows.js
```diff
@@ -1,114 +1,252 @@
...
+            return {
+                ...state,
+                highlight: action.highlight,
+            }
 
-export default function reducer(state = defaultState, action) {
-    switch (action.type) {
-        case UPDATE_FLOWS:
-            let all = reduceList(state.all, action)
+        case UPDATE_SORTER:
```
mhils: We want to keep the state serializable, so we don't want to pass functions here. Can we please go back to the previous implementation?

web/src/js/ducks/utils/list.js
```diff
@@ -1,166 +1,151 @@
...
-    isFetching: false,
-    actionsDuringFetch: [],
+    raw: [],
+    data: [],
     byId: {},
-    indexOf: {},
+    isFetching: false,
+    [SYM_FILTER]: () => true,
+    [SYM_SORTER]: () => 0,
+    [SYM_PENDING]: [],
```
mhils: Why is pending a Symbol?

mhils:
I'm sorry, but this PR goes in the wrong direction.

1. As we discussed previously, state should be serializable. This is not a random rule I'm making up: http://redux.js.org/docs/FAQ.html#organizing-state-non-serializable. I do not understand why you are thwarting this after we discussed it extensively.
2. The reason for splitting `list` and `view` is that a `list` can/will have multiple `views`. If you want to make architectural changes like that, it's best to quickly ping me or @cortesi and discuss it.
3. `utils/view` contains some performance optimizations optimizations that are required once we reach a certain number of flows and events - we hit issues here with previous mitmweb versions, and the current optimizations are really necessary (the `sortedInsert/Remove` methods actually need to be optimized, but I left that for further work)

Let's have a quick chat how we proceed, just ping me on Slack please.

web/src/js/ducks/utils/list.js
```diff
@@ -1,166 +1,91 @@
...
-                indexOf: {}
-            }
-            for (let i = 0; i < action.list.length; i++) {
-                let item = action.list[i]
-                s.byId[item.id] = item
-                s.indexOf[item.id] = i
-            }
-            for (action of state.actionsDuringFetch) {
-                s = reduceList(s, action)
+        case SET:
```
mhils: Why do we have `SET` and `CLEAR` here now? What is `list` supposed to be doing now?

gzzhanghao: `Object.keys` are in order, but you are right, we need a list for this :+1:

web/src/js/components/FlowView/Messages.jsx
```diff
@@ -153,16 +152,3 @@ export class Response extends Component {
         }
     }
 }
-
-export function Error({ flow }) {
```
mhils: Why did you remove this?

PR #1292, opened 2016-06-22
py3++: test/mitmproxy/test_server

dufferzafar:


test/mitmproxy/tservers.py
```diff
@@ -148,7 +150,7 @@ def pathod(self, spec, sni=None):
             Constructs a pathod GET request, with the appropriate base and proxy.
         """
         p = self.pathoc(sni=sni)
-        spec = spec.encode("string_escape")
```
dufferzafar: ^

test/mitmproxy/tservers.py
```diff
@@ -43,7 +45,7 @@ def clear_log(self):
         self.tlog = []
 
     def add_event(self, message, level=None):
-        self.tlog.append(message)
+        self.tlog.append(strutils.always_bytes(message))
```
dufferzafar: ^

dufferzafar:
Not all tests have yet passed. So this is WIP!

test/mitmproxy/tservers.py
mhils: I think log events should always be text, not bytes.

test/mitmproxy/tservers.py
mhils: If it does not cause any issues, let's remove the line entirely.

netlib/http/request.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, first_line_format, method, scheme, host, port, path, http_ver
 
         self.first_line_format = first_line_format
         self.method = method
-        self.scheme = scheme
+        self.scheme = scheme or b''
```
dufferzafar: Made this change to fix these errors: https://travis-ci.org/mitmproxy/mitmproxy/jobs/140021906#L342-L365

Kriechi: How does this relate? Why/where is `scheme` not set?

dufferzafar: I think I linked the wrong lines. [These](https://travis-ci.org/mitmproxy/mitmproxy/jobs/140021906#L367-L384) are the ones!

test/mitmproxy/tservers.py
```diff
@@ -43,7 +45,7 @@ def clear_log(self):
         self.tlog = []
 
     def add_event(self, message, level=None):
-        self.tlog.append(message)
+        self.tlog.append(strutils.native(message, "utf8"))
```
mhils: Looks like this fixes symptoms, not causes. Where do we get binary events from?

mhils:
Fixed & Merged! :smiley:

PR #1294, opened 2016-06-26
update appveyor with tox

Kriechi:


mhils:
Looks good so far. I pushed the wheel to https://snapshots.mitmproxy.org/lxml-3.6.0-cp35-cp35m-win32.whl. :smiley:

dev.bat
```diff
@@ -5,6 +5,8 @@ virtualenv %VENV% --always-copy
 if %errorlevel% neq 0 exit /b %errorlevel%
 call %VENV%\Scripts\activate.bat
 if %errorlevel% neq 0 exit /b %errorlevel%
+pip install -U pip
```
mhils: That's a bad idea on Windows, you want `python -m pip install -U pip`. Otherwise, it can't overwrite the `pip.exe` executable as it's currently in use.

tox.ini
```diff
@@ -7,28 +7,37 @@ deps =
...
 # remove bash & pipe & grep hack after cryptography ships with openssl 1.1.0
 whitelist_externals = bash
 commands =
-  bash -c 'set -o pipefail ; py.test --cov netlib --cov mitmproxy --cov pathod --color=yes --timeout 60 {env:PY3TESTS:} 2>&1 | grep -v Cryptography_locking_cb'
+  bash -c 'set -o pipefail ; py.test --cov netlib --cov mitmproxy --cov pathod --timeout 60 {env:PY3TESTS:} 2>&1 | grep -v Cryptography_locking_cb'
+  codecov -e TOXENV
+
+[testenv:py35-ci-win]
+deps =
+  https://dl.dropboxusercontent.com/u/964866/lxml-3.6.0-cp35-cp35m-win32.whl
```
mhils: https://snapshots.mitmproxy.org/misc/lxml-3.6.0-cp35-cp35m-win32.whl

mhils:
From http://lists.idyll.org/pipermail/testing-in-python/2015-June/006468.html:

> This breaks a lot of basic features, in a way that is fairly complicated to debug. E.g. with tox 2.0, os.path.expanduser("~") is now broken on windows because e.g. USERNAME is not propagated.

Kriechi:
Not sure what you have in mind - maybe you could add it as commit to this PR?
Or is there a more general solution - some ENV variable we could pass?

PR #1297, opened 2016-06-28
Python 3 - test_flow_export

dufferzafar:


mitmproxy/flow/export.py
```diff
@@ -4,23 +4,34 @@
...
 from textwrap import dedent
 
-from six.moves.urllib.parse import quote, quote_plus
+import six
+from six.moves import urllib
 
 import netlib.http
 
 
+def _native(s):
```
dufferzafar: This function is exactly the same as [`strutils.native`](https://github.com/mitmproxy/mitmproxy/blob/master/netlib/strutils.py#L14-L29) except for [this part](https://github.com/mitmproxy/mitmproxy/blob/master/netlib/strutils.py#L21-L22), as we need this here because s *could* be lists etc.

mitmproxy/flow/export.py
```diff
@@ -4,23 +4,34 @@
...
+    if six.PY2:
+        if isinstance(s, six.text_type):
+            return s.encode()
+    else:
+        if isinstance(s, six.binary_type):
+            return s.decode()
```
dufferzafar: Plain ASCII decoding won't work here I guess, but I'm not sure what will!

mitmproxy/flow/export.py
```diff
@@ -4,23 +4,34 @@
...
     for k, v in items:
-        lines.append(indent + "%s: %s,\n" % (repr(k), repr(v)))
+        lines.append(indent + "%s: %s,\n" % (repr(_native(k)), repr(_native(v))))
     return "{\n%s}\n" % "".join(lines)
 
 
 def curl_command(flow):
     data = "curl "
 
     for k, v in flow.request.headers.fields:
```
dufferzafar: @mhils There's `flow.request.data.*` attributes for accessing `bytes`, while `flow.request.*` for `str`. Shouldn't there be something similar for headers, or is decoding the tuples of headers the right way to handle this?

mhils: You can use
```python
for k, v in flow.request.headers.items():
```

It's not 100% exact as headers are folded, but if we follow the RFCs that does not change anything.

mitmproxy/flow/export.py
```diff
@@ -29,7 +40,7 @@ def curl_command(flow):
     data += "'%s'" % full_url
 
     if flow.request.content:
-        data += " --data-binary '%s'" % flow.request.content
+        data += " --data-binary '%s'" % _native(flow.request.content)
```
mhils: I'm all but an expert on that, but doesn't that look like a `bytes_to_escaped_str` use case?

mitmproxy/flow/export.py
mhils: > because s could be lists etc.

When does this apply?

netlib/strutils.py
```diff
@@ -20,12 +20,12 @@ def native(s, *encoding_opts):
...
-    if six.PY3:
-        if isinstance(s, six.binary_type):
-            return s.decode(*encoding_opts)
-    else:
+    if six.PY2:
         if isinstance(s, six.text_type):
             return s.encode(*encoding_opts)
+    else:
+        if isinstance(s, six.binary_type):
+            return s.decode(*encoding_opts)
```
mhils: :+1: :+1: :+1:

mitmproxy/flow/export.py
```diff
@@ -149,7 +164,7 @@ class WebsiteUser(HttpLocust):
 
     data = ""
     if flow.request.body:
```
mhils: Can we update the check as well please? :relaxed:

mitmproxy/flow/export.py
```diff
@@ -126,17 +137,21 @@ class WebsiteUser(HttpLocust):
...
-    file_name = "_".join(components)
-    name = re.sub('\W|^(?=\d)', '_', file_name)
-    url = flow.request.scheme + "://" + flow.request.host + "/" + "/".join(components)
+    components = [urllib.parse.quote(c, safe="") for c in flow.request.path_components]
+    name = re.sub('\W|^(?=\d)', '_', "_".join(components))
     if name == "" or name is None:
         new_name = "_".join([str(flow.request.host), str(flow.request.timestamp_start)])
         name = re.sub('\W|^(?=\d)', '_', new_name)
+
+    url = flow.request.scheme + "://" + flow.request.host + "/" + "/".join(components)
```
mhils: Why don't we use `flow.request.url` here?

mitmproxy/flow/export.py
dufferzafar: The `_native` function is used in the `dicstr` function to convert a dictionary to string (sort of like json.dumps), so for eg. [this params dict](https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/data/test_flow_export/python_get.py#L11) has a list as value.

mitmproxy/flow/export.py
dufferzafar: I honestly have no idea how Locust works, but this may have been written if only the path is required in the url and not the query parameters. That doesn't make much sense though :/

mitmproxy/flow/export.py
dufferzafar: Because we're using that data as a part of a string? - that makes sense.

Does that mean, we don't even need `_native`? What will happen in [these](https://github.com/dufferzafar/mitmproxy/blob/py3-flow-export/mitmproxy/flow/export.py#L151-L155) [cases](https://github.com/dufferzafar/mitmproxy/blob/py3-flow-export/mitmproxy/flow/export.py#L99)?

mitmproxy/flow/export.py
dufferzafar: `dictstr` feels like a bit hacky to me. Should we replace it with a `json.load` & `json.dumps` pair?

mhils: :+1:

PR #1299, opened 2016-06-28
Python 3 - test_dump

dufferzafar:
- [ ] Squash Commits etc.

### DIFF
--- a/mitmproxy/contentviews.py
+++ b/mitmproxy/contentviews.py
@@ -172,7 +172,7 @@ class ViewHex(View):
     def _format(data):
         for offset, hexa, s in strutils.hexdump(data):
             yield [
-                ("offset", offset + " "),
+                ("offset", offset + b" "),
                 ("text", hexa + "   "),
                 ("text", s)
             ]
--- a/mitmproxy/contentviews.py
+++ b/mitmproxy/contentviews.py
@@ -172,7 +172,7 @@ class ViewHex(View):
     def _format(data):
         for offset, hexa, s in strutils.hexdump(data):
             yield [
-                ("offset", offset + b" "),
+                ("offset", offset + " "),
                 ("text", hexa + "   "),
                 ("text", s)
             ]
--- a/mitmproxy/dump.py
+++ b/mitmproxy/dump.py
@@ -244,21 +244,21 @@ class DumpMaster(flow.FlowMaster):
             stickycookie = ""
 
         if flow.client_conn:
-            client = click.style(strutils.bytes_to_escaped_str(flow.client_conn.address.host), bold=True)
+            client = click.style(strutils.escape_control_characters(flow.client_conn.address.host), bold=True)
         else:
             client = click.style("[replay]", fg="yellow", bold=True)
 
-        method = flow.request.data.method
+        method = flow.request.method
         method_color = dict(
             GET="green",
             DELETE="red"
         ).get(method.upper(), "magenta")
-        method = click.style(strutils.bytes_to_escaped_str(method), fg=method_color, bold=True)
+        method = click.style(strutils.escape_control_characters(method), fg=method_color, bold=True)
         if self.showhost:
             url = flow.request.pretty_url
         else:
             url = flow.request.url
-        url = click.style(url, bold=True)
+        url = click.style(strutils.escape_control_characters(url), bold=True)
 
         httpversion = ""
         if flow.request.http_version not in ("HTTP/1.1", "HTTP/1.0"):
@@ -288,7 +288,7 @@ class DumpMaster(flow.FlowMaster):
         elif 400 <= code < 600:
             code_color = "red"
         code = click.style(str(code), fg=code_color, bold=True, blink=(code == 418))
-        reason = click.style(strutils.bytes_to_escaped_str(flow.response.data.reason), fg=code_color, bold=True)
+        reason = click.style(strutils.escape_control_characters(flow.response.reason), fg=code_color, bold=True)
 
         if flow.response.content is None:
             size = "(content missing)"
--- a/mitmproxy/dump.py
+++ b/mitmproxy/dump.py
@@ -248,7 +248,7 @@ class DumpMaster(flow.FlowMaster):
         else:
             client = click.style("[replay]", fg="yellow", bold=True)
 
-        method = flow.request.method
+        method = flow.request.data.method
         method_color = dict(
             GET="green",
             DELETE="red"
@@ -258,7 +258,7 @@ class DumpMaster(flow.FlowMaster):
             url = flow.request.pretty_url
         else:
             url = flow.request.url
-        url = click.style(strutils.bytes_to_escaped_str(url), bold=True)
+        url = click.style(url, bold=True)
 
         httpversion = ""
         if flow.request.http_version not in ("HTTP/1.1", "HTTP/1.0"):
@@ -288,7 +288,7 @@ class DumpMaster(flow.FlowMaster):
         elif 400 <= code < 600:
             code_color = "red"
         code = click.style(str(code), fg=code_color, bold=True, blink=(code == 418))
-        reason = click.style(strutils.bytes_to_escaped_str(flow.response.reason), fg=code_color, bold=True)
+        reason = click.style(strutils.bytes_to_escaped_str(flow.response.data.reason), fg=code_color, bold=True)
 
         if flow.response.content is None:
             size = "(content missing)"
--- a/test/mitmproxy/test_dump.py
+++ b/test/mitmproxy/test_dump.py
@@ -65,6 +65,8 @@ class TestDumpMaster(mastertest.MasterTest):
         cs = StringIO()
         if "verbosity" not in options:
             options["verbosity"] = 0
+        if "flow_detail" not in options:
+            options["flow_detail"] = 0
         o = dump.Options(filtstr=filt, **options)
         return dump.DumpMaster(None, o, outfile=cs)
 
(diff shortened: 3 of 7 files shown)


mitmproxy/dump.py
```diff
@@ -258,7 +258,7 @@ def _echo_request_line(self, flow):
             url = flow.request.pretty_url
         else:
             url = flow.request.url
-        url = click.style(strutils.bytes_to_escaped_str(url), bold=True)
```
dufferzafar: @mhils The `url` is not a `str` here and not `bytes`, so is removing this okay?

mitmproxy/contrib/tnetstring.py
```diff
@@ -158,6 +158,8 @@ def _rdumpq(q, size, value):
         write(b':')
         write(span)
         return size + 2 + len(span) + lvalue
+    elif isinstance(value, str):
+        return _rdumpq(q, size, value.encode())
```
dufferzafar: encoding issues?

dufferzafar:
Now that `tnetstring` deals with `bytes` a lot of changes need to be made that look like [this](https://gist.github.com/dufferzafar/9dc92fbfeff863d944617e5008c8de0d)

mitmproxy/contrib/tnetstring.py
mhils: The underlying issue seems to be that we have unicode/str in `.get_state()`. I feel we should rather fix that rather than hoping that we can encode here. Which part triggers this?

mitmproxy/dump.py
```diff
@@ -158,7 +158,7 @@ def _readflow(self, paths):
 
     def add_event(self, e, level="info"):
         needed = dict(error=0, info=1, debug=2).get(level, 1)
-        if self.o.verbosity >= needed:
+        if self.o.verbosity or -1 >= needed:
```
mhils: Can you explan this? :smile:

mitmproxy/dump.py
mhils: You mean the `url` is a `str`? Yes, then let's remove this!

mitmproxy/dump.py
```diff
@@ -248,7 +248,7 @@ def _echo_request_line(self, flow):
         else:
             client = click.style("[replay]", fg="yellow", bold=True)
 
-        method = flow.request.method
+        method = flow.request.data.method
```
mhils: The point of all the `strutils.bytes_to_escaped_str` calls here is that we don't want to beep on the terminal.
As we do have unicode, we should use `clean_bin` here, not `bytes_to_escaped_str`. If that still beeps (can you test that?), we should fix `clean_bin` to use the symbols from the control_pictures group for this: http://www.fileformat.info/info/unicode/block/control_pictures/images.htm

Ping me on Slack if you need help :smiley:

mitmproxy/dump.py
mhils: (see https://github.com/mitmproxy/mitmproxy/pull/1299/files#r68885432)

mitmproxy/dump.py
dufferzafar: `verbosity` can be `None` which would result in comparing `NoneType` with `Int`. This used to work in Py2 but [not in Py3](https://docs.python.org/release/3.0/whatsnew/3.0.html#ordering-comparisons). So, This was the easiest 'fix' I could come up with.

mitmproxy/dump.py
dufferzafar: I tested out the `clean_bin` like so:

```python
strutils.clean_bin(u'\x07')
```
and it doesn't produce the beep (which `print(u'\x07'` does.) This means all is good.

[This commit](https://github.com/mitmproxy/mitmproxy/pull/1299/commits/de96e64732dd26d32aee0de0a1bed35fe4732e51) will handle control pictures, I hope.

mitmproxy/dump.py
mhils: I see. verbosity should never be 0 though, where is that happening?

mitmproxy/contrib/tnetstring.py
dufferzafar: `flow.get_state`. I printed the state dict on this line. This is how it looks (with a weird mix of strings and bytes):

```

{
    "id": "f559982b-c57f-4093-a0d6-def70c111154", 
    "response": {
        "reason": b"OK", 
        "timestamp_start": 1467552427.8577867, 
        "content": "", 
        "status_code": 200, 
        "headers": (
            (b"header-response", b"svalue"), 
            (b"content-length", b"7")
        ), 
        "timestamp_end": 1467552427.8577871, 
        "http_version": b"HTTP/1.1"
    }, 
    "intercepted": False, 
    "request": {
        "stickycookie": False, 
        "host": b"address", 
        "first_line_format": "relative", 
        "port": 22, 
        "is_replay": False, 
        "stickyauth": False, 
        "http_version": b"HTTP/1.1", 
        "path": b"/path", 
        "timestamp_start": None, 
        "scheme": b"http", 
        "method": b"GET", 
        "headers": (
            (b"header", b"qvalue"), 
            (b"content-length", b"7")
        ), 
        "timestamp_end": None, 
        "content": ""
    }, 
    "type": "http", 
    "error": None, 
    "server_conn": {
        "cert": None, 
        "source_address": {
            "address": ("address", 22), 
            "use_ipv6": True
        }, 
        "timestamp_start": 1.0, 
        "timestamp_ssl_setup": 3.0, 
        "sni": "address", 
        "ip_address": None, 
        "ssl_established": False, 
        "timestamp_end": 4.0, 
        "via": None, 
        "timestamp_tcp_setup": 2.0, 
        "address": {
            "address": ("address", 22), 
            "use_ipv6": True
        }
    }, 
    "client_conn": {
        "timestamp_start": 1.0, 
        "timestamp_ssl_setup": 2.0, 
        "ssl_established": False, 
        "timestamp_end": 3.0, 
        "clientcert": None, 
        "address": {
            "address": ("address", 22), 
            "use_ipv6": True
        }
    }, 
    b"version": (0, 18)
}
```

dufferzafar:
@mhils Now that `tnetstring` deals with `bytes` a lot of changes need to be made that look like [this](https://gist.github.com/dufferzafar/9dc92fbfeff863d944617e5008c8de0d). Is there an alternative to doing this?

mhils:
@dufferzafar: I think we have three ways to tackle this:

1. Add a new data type to our tnetstring implementation for utf8-encoded bytes.
2. Do the changes you sketched out. Hopefully it's not too bad.
3. convert dict keys by default and keep all values as bytes. Fixes most of the changes for the price of  inconsistency.

An implementation for 1 is at https://github.com/mitmproxy/mitmproxy/compare/master...mhils:tnetstring3-unicode-type?expand=1. For 2, we'd ideally make everything serializable subclass StateObject and handle the dict key conversion 3. 3 is relatively simple to implement, but that would force all dict keys to be str, which is a deviation from "normal" tnetstrings as well, so we can probably just go for 1 immediately.

@cortesi, thoughts?

mhils:
After playing around with this for a bit longer, #1318 should be the way to go.

dufferzafar:
Only one test fails now (and that too seems easy to fix.)

mitmproxy/dump.py
mhils: ^ This still needs to be addressed.

mitmproxy/dump.py
```diff
@@ -253,12 +253,12 @@ def _echo_request_line(self, flow):
...
             DELETE="red"
         ).get(method.upper(), "magenta")
-        method = click.style(strutils.bytes_to_escaped_str(method), fg=method_color, bold=True)
+        method = click.style(strutils.escape_control_characters(method), fg=method_color, bold=True)
         if self.showhost:
             url = flow.request.pretty_url
         else:
             url = flow.request.url
-        url = click.style(strutils.bytes_to_escaped_str(url), bold=True)
+        url = click.style(url, bold=True)
```
mhils: This wants a `escape_control_characters` I believe.

mitmproxy/dump.py
```diff
@@ -288,7 +288,7 @@ def _echo_response_line(self, flow):
         elif 400 <= code < 600:
             code_color = "red"
         code = click.style(str(code), fg=code_color, bold=True, blink=(code == 418))
-        reason = click.style(strutils.bytes_to_escaped_str(flow.response.reason), fg=code_color, bold=True)
+        reason = click.style(strutils.bytes_to_escaped_str(flow.response.data.reason), fg=code_color, bold=True)
```
mhils:
```
+ escape_control_characters
- data
```

tox.ini
```diff
@@ -16,7 +16,7 @@ commands =
...
+  TESTS = test/netlib test/pathod/ test/mitmproxy/script test/mitmproxy/test_contentview.py test/mitmproxy/test_custom_contentview.py test/mitmproxy/test_app.py test/mitmproxy/test_controller.py test/mitmproxy/test_fuzzing.py test/mitmproxy/test_script.py test/mitmproxy/test_web_app.py test/mitmproxy/test_utils.py test/mitmproxy/test_stateobject.py test/mitmproxy/test_cmdline.py test/mitmproxy/test_contrib_tnetstring.py test/mitmproxy/test_proxy.py test/mitmproxy/test_protocol_http1.py test/mitmproxy/test_platform_pf.py test/mitmproxy/test_server.py test/mitmproxy/test_filt.py test/mitmproxy/test_flow_export.py test/mitmproxy/test_web_master.py test/mitmproxy/test_flow_format_compat.py test/mitmproxy/test_examples.py test/mitmproxy/test_dump.py
   HOME = {envtmpdir}
```
mhils: This is the last file. I think we can remove the whole subsection. :tada:

mitmproxy/contentviews.py
```diff
@@ -172,7 +172,7 @@ class ViewHex(View):
     def _format(data):
         for offset, hexa, s in strutils.hexdump(data):
             yield [
-                ("offset", offset + " "),
+                ("offset", offset + b" "),
```
mhils: This should just be `" "`?
strutils.hexdump returns `native()`.

test/mitmproxy/test_dump.py
```diff
@@ -65,20 +65,27 @@ def mkmaster(self, filt, **options):
...
         return dump.DumpMaster(None, o, outfile=cs)
 
     def test_basic(self):
         for i in (1, 2, 3):
-            assert "GET" in self.dummy_cycle(self.mkmaster("~s", flow_detail=i), 1, "")
             assert "GET" in self.dummy_cycle(
                 self.mkmaster("~s", flow_detail=i),
                 1,
-                "\x00\x00\x00"
+                b""
```
mhils: Why this change?

mitmproxy/contentviews.py
dufferzafar: The offset is actually `bytes`: https://github.com/mitmproxy/mitmproxy/pull/1299/commits/c7cbbf5f0de9bceccc37b73586ed3baf9421f63a

I didn't know whether to fix that or this!

mhils: Ugh. Fix the offset then!

mhils:
\o/

Kriechi:
Yeah!!!! 🎉🎉🎉 👍👍👍

PR #1306, opened 2016-07-02
Improve Message Body Encoding

mhils:
This PR improves our handling of HTTP message body encodings:

- The unaltered message body is now accessible as `.raw_content` instead of `.content`. It's still `.content` for MessageData.
- The "content-encoding"-decoded content (i.e. gzip removed) content
  is now `.content`, as this is what we want in 99% of the cases. It's still bytes though.
- `.text` now provides the "content-encoding"-decoded and then
  "content-type charset"-decoded message body. This is always unicode.
- The decoded values for `.content` and `.text` are cached,
  so that repeated access and `x.text = x.text` is cheap.
- The `decoded()` decorator is now deprecated, as 
  `.content` does just this. Similarly, `HTTPMessage.get_decoded_content()` is
  deprecated.


Kriechi:
LGTM! Would you mind resolving the conflicts - then we can merge it right away! 👍

netlib/http/message.py
```diff
@@ -137,56 +198,81 @@ def timestamp_end(self):
...
+            except ValueError:
+                # Do we have an unknown content-type charset?
+                # If so, we want to replace it with utf8.
+                if text and enc:
+                    self.headers["content-type"] = re.sub(
+                        "charset=[^;]+",
+                        "charset=utf-8",
+                        self.headers["content-type"]
+                    )
+                encoded = text.encode("utf8", "replace" if six.PY2 else "surrogateescape")
```
cortesi: This feels like a dangerous convenience. We have two different use cases here - "programmatic text" and "display text". Programmatic text is what we want if we're going to parse the resulting text in some way (json, html, etc.). Display text is what we want if we're just displaying the data to the user (print to screen, display in mitmproxy console, etc.). Guessing is OK for the latter but definitely not for the former.

I'd rather return None or raise here, and add a .display property that guesses. I'd be OK with the .display property ALSO guessing about the content-encoding, which may address the convenience issue for .content in many use cases. 

In summary, we'd have:

.content() (raises if there's an issue)
.text() (raises if there's an issue)
.display (never raises, so can be a property)

netlib/http/message.py
```diff
@@ -52,7 +52,22 @@ def from_state(cls, state):
...
 
 
+class CachedDecode(object):
+    __slots__ = ["encoded", "encoding", "decoded"]
+
+    def __init__(self, object, encoding, decoded):
+        self.encoded = object
+        self.encoding = encoding
+        self.decoded = decoded
+
```
cortesi: The cache needs to be bounded in some way. This data can be huge. We should extend LRUCache to do what we want here, and bound it to X megabytes, for some reasonable X.

mhils:
Still needs some more docs + a bound for the decoding cache, but I'd like to see this merged quickly to avoid further merge conflicts.

PR #1307, opened 2016-07-02
py3++: test_protocol_http2

dufferzafar:
Only 1 test passes for now!

### DIFF
--- a/test/mitmproxy/test_protocol_http2.py
+++ b/test/mitmproxy/test_protocol_http2.py
@@ -126,7 +126,7 @@ class _Http2TestBase(object):
         client.wfile.flush()
 
         # read CONNECT response
-        while client.rfile.readline() != "\r\n":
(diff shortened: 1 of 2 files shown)


test/mitmproxy/test_protocol_http2.py
```diff
@@ -227,8 +232,8 @@ def test_simple(self):
 
         assert len(self.master.state.flows) == 1
         assert self.master.state.flows[0].response.status_code == 200
-        assert self.master.state.flows[0].response.headers['server-foo'] == 'server-bar'
-        assert self.master.state.flows[0].response.headers['föo'] == 'bär'
+        assert self.master.state.flows[0].response.headers['server-foo'] == b'server-bar'
+        assert self.master.state.flows[0].response.headers['föo'] == 'bär'.encode('utf8')
```
mhils: This looks weird? `response.headers` should always return a native `str`.

PR #1313, opened 2016-07-04
[RFC] Raise ValueError if content-encoding is invalid

mhils:
@cortesi: As we discussed, this makes `.content` raise a ValueError if decoding the Content-Encoding fails instead of returning our best guess. Unfortunately, after implementing this I'm back at my opinion that we probably should not do this. `Request` is supposed to provide an API that favors simplicity, and wrapping every call to `.content` with
```
try:
    content = request.content
except ValueError:
    content = request.raw_content  # well, let's do a best guess...
```
really does not improve things :confused: . 

I believe we should document our graceful degradation approach, and add a proper UI indicator for mitmproxy if we ever encounter an unknown/invalid content encoding. Alternatively, I suggest we implement `get_content(strict=False)` and make `.content` a shorthand for this. Anyone who should ever be interested in content-encoding correctness (I still can't come up with a convincing use case), is free to use `.data` directly. Again - `Request` should put simplicity first. We're not failing on weird stuff in HTTP headers either.

I know that we agreed on the ValueError approach, but after implementing these changes I feel this puts such a dent in the API usability that we should not go forward with it.

-----------

To provide some context for others: #1306 changes `Request.content` to automatically remove gzip/deflate content encoding, which was previously realized using `with decoded(request)`. This discussion is about what should happen in the rare event where we fail to decode the content, either because we don't know the content-encoding or because the content is malformed.
The original implementation in #1306 tries to decode the content, but returns the raw content if that fails. The advantage of that approach is that it is guaranteed to return a valid value, the disadvantage is that it makes those 0.01% of cases where the content-encoding is invalid harder to debug. The alternative to this is raising an error when accessing the content, which however means that `.content` is not safe to use anymore and we must expect it to raise a ValueError. In a nutshell, the question is whether it's worth sacrificing usability for this 0.01% error case or if we would rather document our graceful degradation behaviour.

Comments are welcome. :-)

### DIFF
--- a/mitmproxy/console/common.py
+++ b/mitmproxy/console/common.py
@@ -256,24 +256,34 @@ def copy_flow_format_data(part, scope, flow):
     else:
         data = ""
         if scope in ("q", "a"):
-            if flow.request.content is None:
+            request = flow.request.copy()
+            try:
+                request.decode()
+            except ValueError:
+                pass
+            if request.raw_content is None:
                 return None, "Request content is missing"
             if part == "h":
-                data += netlib.http.http1.assemble_request(flow.request)
+                data += netlib.http.http1.assemble_request(request)
             elif part == "c":
-                data += flow.request.content
+                data += request.raw_content
             else:
                 raise ValueError("Unknown part: {}".format(part))
-        if scope == "a" and flow.request.content and flow.response:
+        if scope == "a" and flow.request.raw_content and flow.response:
             # Add padding between request and response
             data += "\r\n" * 2
         if scope in ("s", "a") and flow.response:
-            if flow.response.content is None:
+            response = flow.response.copy()
+            try:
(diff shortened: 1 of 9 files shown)


cortesi:
I don't believe we can get away with this strategy when content-encoding is invalid. If this happens, something is very seriously wrong. There is no circumstance in which the right thing to do here is to return the raw, un-decoded content without making the user decide explicitly what to do. This matters more for mitmproxy than for most projects, because we're used in environments where invalid content encodings may well be encountered from malicious actors. Attempts to evade intrusion detection systems spring to mind here.

I suggest  .get_content(strict=True), with the user explicitly opting in to the lie by passing strict=False as an option. This covers most of the convenience use cases within mitmproxy core. Our default advice to users should be to just use .get_content().

PR #1317, opened 2016-07-06
Flow Export Improvements

### LINKED ISSUE #976: Flow Export Improvements

dufferzafar:
Fixes #976

Things still left to do:

- [x] Test workflow for the `b` key
- [x] Integrate Contents/Headers+Content/Url
- [x] Ask for scope (q/s/a)
- [x] Refactor; so that less number of changes have to be made when adding a new exporter
- [x] Test out things from the `flowview`

### DIFF
--- a/mitmproxy/console/common.py
+++ b/mitmproxy/console/common.py
@@ -284,12 +284,6 @@ def copy_flow_format_data(part, scope, flow):
     return data, False
 
 
-def export_prompt(k, f):
-    for exporter in flow.export.EXPORTERS:
-        if k == exporter[1]:
-            copy_to_clipboard_or_prompt(exporter[2](f))
-
-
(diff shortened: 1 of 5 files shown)


mhils:
Looks excellent so far :+1:

dufferzafar:
@mhils: This is ready for review!

cortesi:
This looks good to me. We have to retain "p" for previous flow, though. I like E for export better, and it's free in both the list and flow views.

PR #1318, opened 2016-07-06
[tests actually pass] Add unicode to tnetstring3

mhils:
This PR...

- rewrites the tnetstring module for Python 2/3 cross-compatibility.
- adds a unicode datatype to our tnetstrings implementation
- adjusts `flow.io_compat` to transparently accept dumpfiles and convert them to the native Python str encoding.


mhils:
Ok, this is ready for a final review :smiley:
Needed to include the SNI changes to fix some failing tests.

PR #1322, opened 2016-07-07
Detect JSON API media type as a json content view

linrock:
In practice, Ember.js applications support JSON API out the box, so it would be awesome if HTTP requests with a `Content-Type: application/vnd.api+json` header (as defined in the spec) are detected and parsed as JSON 😃 

IANA assignment here:
https://www.iana.org/assignments/media-types/application/vnd.api+json

More about JSON API:
https://github.com/json-api/json-api


mhils:
Awesome, thanks for the great PR! :smiley: :cake: 
Will merge once tests pass.

PR #1324, opened 2016-07-07
Delete stickycookies when told by the server

### LINKED ISSUE #1096: deleting cookies (stickycookies) does not work

1. start mitmdump with stickycookies: mitmdump -t "." and point a browser through it
2. visit https://httpbin.org/cookies/set?aaa=1 to set a cookie
3. visit https://httpbin.org/cookies/delete?aaa= to delete the cookie
4. visit https://httpbin.org/cookies to display the cookies

The expected behavior, according to RFC 6265, is that the cookie is deleted

##### What went wrong?

The cookie is not deleted

##### Any other comments?

The cookie library used by mitmproxy is Cookie.py from python, which is woefully outdated.
It implements RFC 2109 from 1997 and does not (yet?) support RFC 6265 from 2011.

RFC 6265 standardises both "expires" and "max-age" attributes which can be used
for expiration control of cookies. A cookie can be deleted by setting "expires" to a date in the past,
or "max-age" to 0.

A more up-to-date cookie library is probably beneficial here. I'm not sure if there are any
special reasons for depending on Cookie.py?
Is it better to attempt to fix cookie.py, or switch mitmproxy to another cookie library?

---

Mitmproxy Version: latest
Operating System:

dufferzafar:
Fixes #1096

### DIFF
--- a/mitmproxy/flow/modules.py
+++ b/mitmproxy/flow/modules.py
@@ -1,10 +1,8 @@
 from __future__ import absolute_import, print_function, division
 
 import collections
-import email.utils
 import hashlib
 import re
-import time
 
 from six.moves import http_cookiejar
 from six.moves import urllib
@@ -325,20 +323,7 @@ class StickyCookieState:
             dom_port_path = self.ckey(attrs, f)
 
             if self.domain_match(f.request.host, dom_port_path[0]):
-
-                # See if 'expires' time is in the past
-                expired = False
-                if 'expires' in attrs:
-                    e = email.utils.parsedate_tz(attrs["expires"])
-                    if e:
-                        exp_ts = email.utils.mktime_tz(e)
-                        now_ts = time.time()
-                        expired = exp_ts < now_ts
-
-                # or if Max-Age is 0
-                expired = expired or (int(attrs.get('Max-Age', 1)) == 0)
-
-                if expired:
+                if cookies.is_expired(attrs):
                     # Remove the cookie from jar
                     self.jar[dom_port_path].pop(name, None)
 
--- a/mitmproxy/flow/modules.py
+++ b/mitmproxy/flow/modules.py
@@ -1,8 +1,10 @@
 from __future__ import absolute_import, print_function, division
 
 import collections
+import email.utils
 import hashlib
 import re
+import time
 
 from six.moves import http_cookiejar
 from six.moves import urllib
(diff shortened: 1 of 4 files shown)


mitmproxy/flow/modules.py
```diff
@@ -320,10 +322,33 @@ def handle_response(self, f):
...
             # valid RFC 822/1123 datetime specifications for expiry. Sigh.
-            a = self.ckey(attrs, f)
-            if self.domain_match(f.request.host, a[0]):
-                b = attrs.with_insert(0, name, value)
-                self.jar[a][name] = b
+            dom_port_path = self.ckey(attrs, f)
+
+            if self.domain_match(f.request.host, dom_port_path[0]):
+
+                # See if 'expires' time is in the past
```
mhils: Let's move that into a separate `is_expired` routine in `netlib.http.cookies` ?

dufferzafar: Done!

netlib/http/cookies.py
```diff
@@ -260,3 +261,24 @@ def refresh_set_cookie_header(c, delta):
...
+    # See if 'expires' time is in the past
+    if 'expires' in cookie_attrs:
+        e = email.utils.parsedate_tz(cookie_attrs["expires"])
+        if e:
+            exp_ts = email.utils.mktime_tz(e)
+            now_ts = time.time()
+            expired = exp_ts < now_ts
+
+    # or if Max-Age is 0
+    expired = expired or (int(cookie_attrs.get('Max-Age', 1)) == 0)
```
mhils: This can fail if Max-Age is not an integer, we should catch this.

test/mitmproxy/test_flow.py
```diff
@@ -101,6 +103,28 @@ def test_response(self):
         assert len(s.jar[googlekey]) == 1
         assert list(s.jar[googlekey]["somecookie"].values())[0] == "newvalue"
 
+    def test_response_delete(self):
+        c = "duffer=zafar; Path=/", "www.google.com"
+
+        # Test that a cookie is be deleted
+        # by setting the expire time in the past
+        s, f = self._response(*c)
```
mhils: Can we have separate tests for deletion and `is_expired`?

dufferzafar: @mhils Have a look now.

mhils:
:+1:

PR #1325, opened 2016-07-07
HTTP/2: Python 3-compatibility

Kriechi:
py3++

closes #1307


mitmproxy/protocol/http2.py
```diff
@@ -62,6 +62,9 @@ def safe_send_headers(self, is_zombie, stream_id, headers, **kwargs):
             self.conn.send(self.data_to_send())
 
     def safe_send_body(self, is_zombie, stream_id, chunks):
+        if isinstance(chunks, bytes):
+            chunks = [chunks]
+
```
Kriechi: @dufferzafar Not sure about this change - couldn't thing of something more elegant...

Kriechi: ah, nevermind.

PR #1327, opened 2016-07-08
Remove script context

mhils:
This PR removes the ScriptContext from mitmproxy. Instead, we expose `master` and `log` objects that can be imported. In a nutshell, what previously has been

```python
def request(context, flow):
    context.log("hello there")
```
is now
```python
import mitmproxy

def request(flow):
    mitmproxy.log("hello there")
```

This is based on @cortesi's refactoring branch, but comes with a few changes:

- `flow` is passed explicitly (I think this one is uncontroversial)
- We directly use `mitmproxy`, not `mitmproxy.ctx`. IMO this just looks way nicer. If there are convincing reasons for not doing this, changing this is still easy.
- We don't use the `Facade` thing added in the refactor branch. The current implementation just sets the attribute directly whenever loading or running scripts, which supports both  
`from mitmproxy import log; log("foo")`  
and  
`import mitmproxy; mitmproxy.log("foo")`  
as long as `master` and `log` do not change over the lifetime of a script. As we do not want to transfer/share scripts between masters anyway, we can cut off some complexity without cost I believe? Concurrent masters are obscure and not addressed by either implementation anyway.
- Everything else is left in place. In particular, script reloading and script objects remain separate, so that reloading remains a part of the event loop to avoid race conditions.


dwfreed:
What if one wants different scripts to have different loggers for different logging configuration?  Eg, using python's logging module, it's convention to do getLogger(__name__) to get the correctly nested logger for yourself, which will allow you to have unique logging configuration for each script if you want (or it'll just bubble up parent loggers until it reaches one with the relevant configuration set).

mhils:
There are various ways how we can improve our logging mechanisms - this PR just copies what we do right now, I'd be happy to change that at a later point.

cortesi:
Comments, as they occurred to me:

- Terminology: we should start distinguishing between addons (builtins + external scripts) and scripts (external scripts added and manged by the user, which are handled by the Script addon).
- mitmproxy.ctx or something similar should stay for clarity. This is explicitly the addon context, and is not useful when mitmproxy is used as a library and extended or in the rest of our codebase. Putting these magic objects at the top of the namespace muddies things up. 
- Multiple masters will be worth catering for, but if flow is an argument the facade is not necessary for now. 
- Leaving scripts where they are for this patch is fine, but they should certainly come out into an addon as in my branch. It's way, way cleaner and there's no intrinsic reason for them to be so intimately intertwined with the rest of our codebase.

PR #1330, opened 2016-07-08
Added flow converters for versions 0.11 and 0.12

jpkrause:
I had some really old flows I needed to use. Thankfully, this small change solved my problem. Figured I'd contribute it back in case anyone else has old flow recordings.


Kriechi:
Thanks! 

I'm not sure why, but we have a test which explicitly tests that conversion to 0.12 should fail.
https://github.com/mitmproxy/mitmproxy/blob/819e737f0d0f787215d3648481524f99816b2a71/test/mitmproxy/test_flow_format_compat.py#L14-L18
@mhils is there a reason - or why does this test exist? Or is this simply a lowest value we have, and now we should decrement?

(the tests currently fail, because our coding style requires a few more blank lines between functions 😉 ...)

jpkrause:
@Kriechi Apologies, I neglected to run the tests. I went ahead and created dump files from v0.11 and v0.10 to use in the test. Also updated the styles to have more spaces 😄

mhils:
> Or is this simply a lowest value we have, and now we should decrement?

Exactly.

@jpkrause: Thanks for the great PR! :smiley: :cake:

dufferzafar:
@jpkrause Is the `dumpfile-011` really of v0.11? because I looked at the file and the version info inside it shows v0.18!

mhils:
The dumpfile was indeed 0.18, fixed it.

jpkrause:
Thanks, my apologies about that mistake.

PR #1331, opened 2016-07-09
Test Implementation of Immutable Options

mhils:
I had a bit of a headache with the blinker Signals in the refactor branch. This is an experimental PR which brings immutable options. The idea is that options can only be changed  by calling `Master.update_options(new_options)`, which then calls `configure(new_options)` for all Addons. If any of these calls raises an exeception, we call `configure(old_options)` to roll back the existing changes. Compared to the refactor approach, we force a stricter chain of command here - not anybody can subscribe to updates anymore.
One direct advantage of this is that we can compare the current options with `new_options` in each configure call. This may also be very useful if we want to include the ProxyConfig in the options, as we can ensure that ProxyConfig does not change for existing connections.

@cortesi: This is still very exploratory here. Thoughts?

### DIFF
--- a/mitmproxy/cmdline.py
+++ b/mitmproxy/cmdline.py
@@ -184,9 +184,9 @@ def get_common_options(options):
             )
 
     return dict(
-        app=options.app,
-        app_host=options.app_host,
-        app_port=options.app_port,
+        onboarding_app=options.onboarding_app,
+        onboarding_app_host=options.onboarding_app_host,
+        onboarding_app_port=options.onboarding_app_port,
 
         anticache=options.anticache,
         anticomp=options.anticomp,
@@ -502,12 +502,12 @@ def onboarding_app(parser):
     group = parser.add_argument_group("Onboarding App")
     group.add_argument(
         "--noapp",
-        action="store_false", dest="app", default=True,
+        action="store_false", dest="onboarding_app", default=True,
         help="Disable the mitmproxy onboarding app."
     )
     group.add_argument(
         "--app-host",
-        action="store", dest="app_host", default=APP_HOST, metavar="host",
+        action="store", dest="onboarding_app_host", default=APP_HOST, metavar="host",
         help="""
             Domain to serve the onboarding app from. For transparent mode, use
             an IP when a DNS entry for the app domain is not present. Default:
@@ -517,7 +517,7 @@ def onboarding_app(parser):
     group.add_argument(
         "--app-port",
         action="store",
-        dest="app_port",
+        dest="onboarding_app_port",
         default=APP_PORT,
         type=int,
         metavar="80",
--- a/mitmproxy/console/master.py
+++ b/mitmproxy/console/master.py
@@ -222,9 +222,9 @@ class ConsoleMaster(flow.FlowMaster):
     palette = []
 
     def __init__(self, server, options):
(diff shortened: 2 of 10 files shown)


cortesi:
I have two reservations here. First, under the new scheme, options are updated all the time by our interactive programs. I think some syntax sugar (i.e. being able to directly reassign an attribute when only one thing is changing) is worth having. Second, I like the blinker signals. There's no reason to have a tight dependency between Master and Options, and using blinker lets us untie things. I want to do more of that, not less, to try to tease out the warren of cross-dependenies we've accreted. Basically, I still prefer the design here:

https://github.com/cortesi/mitmproxy/blob/t1/mitmproxy/options.py

What were the problems you had with blinker?

mhils:
>  I think some syntax sugar (i.e. being able to directly reassign an attribute when only one thing is changing) is worth having.

The approach here provides `master.update_options(intercept="foo")`, which I think isn't too bad. In particular, we can return a value (or raise an exception) that indicates if the update has succeeded. Maybe I'm missing an easy way, but the property approach sounds like a giant dance in e.g. the mitmweb API handler (subscribe for error, perform update, check if error callback has been called, unsubscribe)?
Also, it's kind of nice that you can compare the current options with the options you are about to receive without caching values yourself.

> Second, I like the blinker signals. There's no reason to have a tight dependency between Master and Options, and using blinker lets us untie things.

I like blinker in general, but I feel it introduces some unecessary spaghetti here when we try to uncouple option update from its error handling. In the refactor branch, we have
```
ConsoleMaster.edit_scripts -> Options.__setattr__ -> Options.errored.send -> ConsoleMaster.options_error
```
which kind of feels to complex for what we are doing.

Is there any party except callee/master which needs to handle errors? I really like blinker for the console parts, but here it feels like it makes updates intransparent - the only way to somehow follow the chain of execution is grepping for `changed.connect`.

cortesi:
> Maybe I'm missing an easy way, but the property approach sounds like a giant dance in e.g. the mitmweb API handler (subscribe for error, perform update, check if error callback has been called, unsubscribe)?

There's no extra complexity - a single attribute assignment (maybe the most common case) is just an update with only a single item. I think it's sugar worth having, given the new role Options takes - that is, being the one true source for our options state that gets manipulated often from the interactive programs. I think a look at the console changes for Options should convince you that this is a worthwhile quality of life convenience. 

> Is there any party except callee/master which needs to handle errors?

The blinker event notification system is really light-weight, and buys us modularity at little extra cost. The console handler for options errors is a really good example of that - we just subscribe to notifications and display them in one spot, and there's no need to couple the code at the call site with the notification mechanism at all. This is miles easier to reason about than either exception propagation, or adding a guard at every single update site that does the notification.

There's one thing worth chatting about here - I do think that addons and other points listening for options changes will need a way to ask "what's changed" for an option. As it happens, I hadn't encountered this just yet in my integration branch, so I haven't come up with a solution yet. Making addons do a diff between a stored "previous options" and the new options object is definitely the wrong way to go, though. The keys in options can be nested, complex structures, and a deep comparison is just too brittle here. We already have a complete list of all keys that have changed by virtue of the update mechanism. I'd rather just expose that on the change event somehow. Again, we can leave the details for this until we have the first use case.

cortesi:
A cleaner Options implementation is now in base. ;)

PR #1335, opened 2016-07-10
move OSX Travis job to py3 for snapshots

### LINKED ISSUE #1332: Update OSX builds to Python 3
We can move the whole travis OSX tests to Python 3 (only), which will then also create Python 3 binaries.

Kriechi:
@mhils is this everything need to fix #1332?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,12 +20,10 @@ matrix:
   include:
     - python: 3.5
       env: TOXENV=lint
-    - language: generic
-      os: osx
-      osx_image: xcode7.1
-      git:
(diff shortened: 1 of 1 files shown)


Kriechi:
I tested the generated snapshot for OSX:

```
$ mitmdump --sysinfo
Mitmproxy version: 0.18
Python version: 3.5.2
Platform: Darwin-15.5.0-x86_64-i386-64bit
SSL version: b'OpenSSL 1.0.2h  3 May 2016'
Mac version: 10.11.5 ('', '', '') x86_64
```

Looks good!

PR #1341, opened 2016-07-11
modify flow content

cle1000:


mitmproxy/web/app.py
```diff
@@ -245,6 +245,9 @@ def put(self, flow_id):
                         request.port = int(v)
                     elif k == "headers":
                         request.headers.set_state(v)
+                    elif k == "content":
+                        print(v)
+                        response.content = str(v)
```
mhils: I like the general idea, but this will break:

- JSON is not binary-safe, so anything binary will break this.
- JSON is super slow for multiple-MB things
- We want to have drag-and-drop upload - the easiest way to implement this is FormData upload, so we should have a multipart/formdata endpoint.

Can we put to /flow/content?

web/package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "scripts": {
     "test": "jest",
-    "build": "gulp prod",
+    "build": "gulp dev",
```
mhils: same as above

web/src/js/components/ContentView/ContentViews.jsx
```diff
@@ -22,13 +24,19 @@ export function ViewImage({ flow, message }) {
 
 ViewRaw.textView = true
 ViewRaw.matches = () => true
+ViewRaw.input = {}
```
mhils: Class-level, not instance-level?

web/src/js/components/common/MonacoEditor.jsx
```diff
@@ -0,0 +1,34 @@
+//not working
```
mhils: @gzzhanghao: can you get the monaco editor working in react?

web/src/js/components/ContentView/ContentViews.jsx
cle1000: yes but ViewRaw is a pure function not a class ?

web/src/js/components/common/MonacoEditor.jsx
gzzhanghao: @mhils Yes, but I think we can use `react-ace` or `react-codemirror` instead, any reason for using monaco here?

web/src/js/components/ContentView/ContentViews.jsx
gzzhanghao: So we should make `ViewRaw` a class :wink:

web/src/js/components/common/MonacoEditor.jsx
cle1000: hmm, it was a recommendation from max to use monaco. but `react-ace` looks great. if we use react-ace i can do this :)

mitmproxy/web/app.py
```diff
@@ -298,6 +292,22 @@ def post(self, flow_id):
...
+        # handle request later now just change response content
+
+        flow = self.flow
+        flow.backup()
+        content = ''
+        if (len(self.request.files.values()) > 0):
+            content = self.request.files.values()[0][0]["body"]
+        elif (len(self.request.arguments) > 0):
+            content = self.request.arguments['file']
+
```
cle1000: is not working as i want, do you have any hints how to set the content, I have problems with gzip, when i upload something i need to disable gzip to view the content

mitmproxy/web/app.py
```diff
@@ -298,6 +292,22 @@ def post(self, flow_id):
...
+        flow.response.content = str(content)
+        self.state.update_flow(flow)
```
cle1000: this update call wont reload the content of the flow :(

web/src/js/components/ContentView/ContentViews.jsx
```diff
@@ -1,10 +1,12 @@
...
 import ContentLoader from './ContentLoader'
 import { MessageUtils } from '../../flow/utils.js'
-import Button from '../common/Button'
+import CodeEditor from '../common/CodeEditor'
+import {formatSize} from '../../utils.js'
 
 
-const views = [ViewAuto, ViewImage, ViewJSON, ViewRaw]
+
+const views = [ViewAuto, ViewImage, ViewJSON, ViewRaw, ViewFile]
```
cle1000: should i add different flows, or should i just implement the ViewRaw mode and we add new views in an extra PR

web/src/js/ducks/flows.js
```diff
@@ -117,6 +117,14 @@ export function update(flow, data) {
     return { type: REQUEST_ACTION }
 }
 
+export function update_content(flow, file) {
+    const body = new FormData()
+    body.append('file', file)
+    fetchApi(`/flows/${flow.id}/response/content`, {method: 'post',  body} )
```
cle1000: this fileupload should be used with a string or a file... i think it is better to create a file out of the string and then upload it as a file ? What do you think ?

web/src/js/components/common/CodeEditor.jsx
```diff
@@ -0,0 +1,41 @@
+import React, { Component, PropTypes } from 'react'
+import { render } from 'react-dom';
+import brace from 'brace';
+import AceEditor from 'react-ace';
```
cle1000: i replaced the textbox with the `ace-react` editor

web/src/js/ducks/flows.js
```diff
@@ -123,6 +123,7 @@ export function update_content(flow, file) {
         file = new Blob([file], {type: 'plain/text'})
     body.append('file', file)
     fetchApi(`/flows/${flow.id}/response/content`, {method: 'post',  body} )
+    update(flow, {response: {headers: [['Content-Encoding', '']]} } )
```
cle1000: thats the problem with gzip, i have to disable it, otherwise i got an error. should i gzip on my own? any hints ?

web/src/js/ducks/flows.js
```diff
@@ -117,6 +117,17 @@ export function update(flow, data) {
     return { type: REQUEST_ACTION }
 }
 
+export function update_content(flow, file) {
```
mhils: Let's stick to JS's camel-case and call this `updateContent`. Looks good, but should get an additional "request"/"response" argument :wink:

cle1000: yes of course, i just handled every where the response case, when everything is working i will add the stuff for response

web/src/js/components/ContentView.jsx
```diff
@@ -59,18 +69,28 @@ export default class ContentView extends Component {
...
+                    <View flow={flow} update_content={content => flowsActions.update_content(this.props.flow, content)}  message={message} />
                 )}
                 <div className="view-options text-center">
                     <ViewSelector onSelectView={this.selectView} active={View} message={message}/>
                     &nbsp;
                     <a className="btn btn-default btn-xs" href={MessageUtils.getContentURL(flow, message)}>
                         <i className="fa fa-download"/>
                     </a>
+                    &nbsp;
+                    <a className="btn btn-default btn-xs" href="#" onClick={e => {this.fileInput.click(); e.preventDefault();}}>
```
mhils: Can we add a title attribute for both buttons?

web/src/js/components/common/Button.jsx
```diff
@@ -10,8 +10,10 @@ export default function Button({ onClick, text, icon, disabled }) {
         <div className={"btn btn-default"}
              onClick={onClick}
              disabled={disabled}>
-            <i className={"fa fa-fw " + icon}/>
-            &nbsp;
+            <span hidden={!icon}>
```
mhils: `{icon && <i ... />}` ?

web/src/js/components/FlowView/Messages.jsx
```diff
@@ -89,7 +89,7 @@ export class Request extends Component {
                     onChange={headers => updateFlow({ request: { headers } })}
                 />
                 <hr/>
-                <ContentView flow={flow} message={flow.request}/>
+                <ContentView flow={flow} message={flow.request} onChange={content => updateFlow({request: {content} })}/>
```
mhils: remove

cle1000:
i changed the way to edit the flow content, what do you think? i had problems with dependencies when i tried to replace rea with react-codemirror or react-code-mirror. i will try this tomorrow.

mitmproxy/web/app.py
```diff
@@ -292,6 +292,20 @@ def post(self, flow_id):
 
 class FlowContent(RequestHandler):
 
+    def post (self, flow_id, message):
+        # handle request later now just change response content
+
+        flow = self.flow
+        flow.backup()
+        content = self.request.files.values()[0][0].body
+        if (message == "response"):
```
mhils: We can use `message = getattr(self.flow, message)` and then do the rest in a generic fashion.

web/src/css/codemirror.less
```diff
@@ -0,0 +1,5 @@
+.ReactCodeMirror {
+	border: 1px solid #ccc;
+}
+
```
mhils: How about adding the `height: auto` part from https://codemirror.net/demo/resize.html, so that we can get rid of one of the scrollbars? :smiley:

web/src/css/flowview.less
```diff
@@ -6,4 +6,28 @@
...
         max-height: 100%;
     }
-}
\ No newline at end of file
+}
+
+.edit-flow  {
+    position: absolute;
+    right: 25px;
+    top: 140px;
```
mhils: If you wrap `.edit-flow` in a positioned div, we can put in useful values here. These `140px` are dependend on the height of the header etc, that's very fragile.

web/src/js/components/ContentView/ContentEditor.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+    constructor(props){
+        super(props)
+        this.state = {content: this.props.content}
+    }
+
+    render() {
+        return (
+            <div>
+                {this.props.isClosed ?
```
mhils: This Component is a weird mixup of both the buttons and the code editor. These should be in different components.

web/src/js/components/ContentView.jsx
```diff
@@ -58,20 +67,50 @@ export default class ContentView extends Component {
...
+                        <ContentLoader  flow={flow} message={message}>
+                            <this.state.View content="" />
+                        </ContentLoader>
+                    ) : (
+                        <View flow={flow}  message={message} />
+                    )}
+
+
+
+                    <div className="view-options text-center">
```
mhils: The buttons should be moved into their own components.

PR #1342, opened 2016-07-11
fixed bug

cle1000:
bug is fixed

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -3,7 +3,7 @@
   "private": true,
   "scripts": {
     "test": "jest",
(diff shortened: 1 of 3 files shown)


web/package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "scripts": {
     "test": "jest",
-    "build": "gulp prod",
+    "build": "gulp dev",
```
mhils: Can you just run `gulp dev` to compile stuff and leave this as-is?

mhils:
:+1:

PR #1343, opened 2016-07-12
TCPFlow Filters

dufferzafar:
Filters that make sense for TCPFlows:

- [x] body: `~b`
- [x] client body:`~bq`
- [x] server body: `~bs`
- [x] src / dest address: `~src` `~dst`
- [x] error: `~e`
- [x] operators: `! & |`

Filters that will just return `False` for TCPFlows:

- [x] request/response: `~q`, `~s`
- [x] headers: `~h`, `~hq`, `hs`
- [x] content-type: `~t`, `~tq`, `~ts`
- [x] response code: `~c`
- [x] domain: `~d`
- [x] method: `~m`
- [x] url: `~u`

Other Tasks:

- [x] `@only` decorator
- [x] Add a [`match`](https://github.com/dufferzafar/mitmproxy/blob/15c0ccb306fa25f76ddcd17daa8d4f617dc68904/mitmproxy/models/http.py#L216-L233) function.
- [x] `~http` & `~tcp` filters
- [x] Tests for `TCPFlow.match`
- [x] Tests for a new type of flow: `TestFlow`

### DIFF
--- a/mitmproxy/filt.py
+++ b/mitmproxy/filt.py
@@ -81,6 +81,24 @@ class FErr(_Action):
         return True if f.error else False
 
 
+class FHTTP(_Action):
+    code = "http"
+    help = "Match HTTP flows"
+
+    @only(HTTPFlow)
+    def __call__(self, f):
+        return True
+
+
+class FTCP(_Action):
+    code = "tcp"
+    help = "Match TCP flows"
+
+    @only(TCPFlow)
+    def __call__(self, f):
+        return True
+
+
 class FReq(_Action):
     code = "q"
     help = "Match request with no response"
@@ -384,7 +402,9 @@ filt_unary = [
     FReq,
     FResp,
     FAsset,
-    FErr
+    FErr,
+    FHTTP,
+    FTCP,
 ]
 filt_rex = [
     FHeadRequest,
--- a/mitmproxy/filt.py
+++ b/mitmproxy/filt.py
@@ -76,7 +76,6 @@ class FErr(_Action):
     code = "e"
     help = "Match error"
 
-    @only(HTTPFlow, TCPFlow)
     def __call__(self, f):
         return True if f.error else False
 
@@ -326,7 +325,6 @@ class FSrc(_Rex):
     help = "Match source address"
     is_binary = False
 
-    @only(HTTPFlow, TCPFlow)
     def __call__(self, f):
         return f.client_conn.address and self.re.search(repr(f.client_conn.address))
 
@@ -336,7 +334,6 @@ class FDst(_Rex):
     help = "Match destination address"
     is_binary = False
 
-    @only(HTTPFlow, TCPFlow)
     def __call__(self, f):
         return f.server_conn.address and self.re.search(repr(f.server_conn.address))
 
--- a/mitmproxy/filt.py
+++ b/mitmproxy/filt.py
@@ -193,12 +193,22 @@ class FBod(_Rex):
     help = "Body"
 
     def __call__(self, f):
-        if f.request and f.request.content:
-            if self.re.search(f.request.get_decoded_content()):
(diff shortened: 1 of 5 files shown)


mitmproxy/filt.py
```diff
@@ -193,12 +193,22 @@ class FBod(_Rex):
...
     def __call__(self, f):
-        if f.request and f.request.content:
-            if self.re.search(f.request.get_decoded_content()):
-                return True
-        if f.response and f.response.content:
-            if self.re.search(f.response.get_decoded_content()):
-                return True
+
+        # HTTPFlow
+        if hasattr(f, 'request'):
```
dufferzafar: We could also use:

```python
from mitmproxy.models.http import HTTPFlow
from mitmproxy.models.tcp import TCPFlow

if isinstance(f, HTTPFlow):
    pass
elif isinstance(f, TCPFlow)
    pass
```

Is there a better way to distinguish between HTTPFlow and TCPFlow?

dufferzafar:
[Duplicate Line?](https://github.com/dufferzafar/mitmproxy/blob/tcpflow-filter/test/mitmproxy/test_filt.py#L228)

mitmproxy/filt.py
mhils: How about this?

```python
def http(fn):
    @functools.wraps(fn)
    def filter_http_only(flow):
        return isinstance(flow, HTTPFlow) and fn(flow)
    return filter_http_only
```

```python
class FBod(_Rex):
    code = "b"
    help = "Body"
    
    @http
    def __call__(self, f):
```

mhils: You can then define your own TestFlow(Flow) class in the tests (which has no additional attributes) and use that to test all filters.

mhils: We can also support decorator chaining, e.g.
```python
@http
@tcp
def __call__(self, f):
    # this filter only works for http and tcp flows
```
This is a bit funky to implement, but certainly doable - I'll leave the challenge for you. :blush:

mitmproxy/models/tcp.py
```diff
@@ -53,3 +55,22 @@ def __init__(self, client_conn, server_conn, live=None):
 
     def __repr__(self):
         return "<TCPFlow ({} messages)>".format(len(self.messages))
+
+    def match(self, f):
```
dufferzafar: Where should the corresponding test for this go? I couldn't find the tests for TCPFlow.

dufferzafar:
Ready for a final review!

mitmproxy/filt.py
```diff
@@ -61,14 +76,34 @@ class FErr(_Action):
     code = "e"
     help = "Match error"
 
+    @only(HTTPFlow, TCPFlow)
```
mhils: I think all flows have an `.error` attribute, so we don't need to limit this.

mitmproxy/filt.py
```diff
@@ -261,6 +326,7 @@ class FSrc(_Rex):
     help = "Match source address"
     is_binary = False
 
+    @only(HTTPFlow, TCPFlow)
```
mhils: Same over here. We currently only have HTTPFlow and TCPFlow, but `.client_conn` is already an attribute of `Flow`, so this should extend to all flows we'll ever encounter. Let's just strip off the decorator.

mitmproxy/filt.py
```diff
@@ -270,6 +336,7 @@ class FDst(_Rex):
     help = "Match destination address"
     is_binary = False
 
+    @only(HTTPFlow, TCPFlow)
```
mhils: Same

mhils:
Other than that, LGTM :+1:

dufferzafar:
@mhils Made the changes!

PR #1344, opened 2016-07-12
Fix is_ascii check when determining next layer

strohu:
An ASCII character can't be both upper case and lower case. Fixing this, otherwise --raw-tcp option will assume innermost layer is always raw TCP.


Kriechi:
Thanks! 🎉

PR #1345, opened 2016-07-12
Make sudo pfctl error check Python 3 compatible

strohu:
In Python 3, subprocess.check_output() returns a sequence of bytes. This change ensures that it will be converted to a string, so the substring test for the sudo error message does not raise a TypeError. This fixes the code in Python 3 while remaining compatible with Python 2.


Kriechi:
Thanks! 🎉

mhils:
Thanks for testing mitmproxy on Python 3! We're still in a very early phase here, so having testers that early is amazing! :smiley:

strohu:
It's all thanks to those hacking instructions in the readme that suggest Python 3, but happy to help :)

PR #1348, opened 2016-07-13
Addons

cortesi:


mitmproxy/controller.py
```diff
@@ -191,6 +196,10 @@ def wrapper(master, message):
 
         with master.handlecontext():
             ret = f(master, message)
+            if handling:
+                # Python2/3 compatibility hack
+                fn = getattr(f, "func_name", None) or getattr(f, "__name__")
```
Kriechi: I think `__name__` also works on Python 2.7.

dwfreed: It does.  `f.__name__` would work just fine here.

PR #1352, opened 2016-07-14
StickyCookies to addon

cortesi:

PR #1355, opened 2016-07-14
Re-add deletion of stickycookies

dufferzafar:
This was originally added in #1324 (fixing issue #1096)
but got lost during "the big refactor" (#1352)


mhils:
Thanks! :smiley:

PR #1356, opened 2016-07-14
Scripts to addon

cortesi:


mitmproxy/builtins/script.py
```diff
@@ -0,0 +1,162 @@
...
+from mitmproxy import ctx
+
+
+import watchdog.events
+# The OSX reloader in watchdog 0.8.3 breaks when unobserving paths.
+# We use the PollingObserver instead.
+if sys.platform == 'darwin':  # pragma: no cover
+    from watchdog.observers.polling import PollingObserver as Observer
+else:
+    from watchdog.observers import Observer
```
mhils: We need to backport the changes from master here.

mitmproxy/builtins/script.py
```diff
@@ -0,0 +1,162 @@
...
+            self.ns = load_script(self.path, self.args)
+        self.run("configure", options)
+
+
+class ScriptLoader():
+    """
+        An addon that manages loading scripts from options.
+    """
+    def configure(self, options):
+        for s in options.scripts or []:
```
mhils: Per https://github.com/mitmproxy/mitmproxy/blob/e051928f26afbc58c01d7854b62f8bfb0d4ed5ee/mitmproxy/flow/options.py#L24, `.scripts` is never None and always an empty sequence.

mitmproxy/builtins/script.py
```diff
@@ -0,0 +1,162 @@
...
+        if not self.ns:
+            self.ns = load_script(self.path, self.args)
+        self.run("configure", options)
+
+
+class ScriptLoader():
+    """
+        An addon that manages loading scripts from options.
+    """
+    def configure(self, options):
```
mhils: Edge-case: Changing the script order does not change the addon order.

mitmproxy/builtins/script.py
```diff
@@ -0,0 +1,162 @@
...
+def load_script(path, args):
+    ns = {'__file__': os.path.abspath(path)}
+    with scriptenv(path, args):
+        with open(path, "rb") as f:
+            code = compile(f.read(), path, 'exec')
+            exec(code, ns, ns)
+    return ns
+
+
+class ReloadHandler(watchdog.events.FileSystemEventHandler):
```
mhils: Any reason for changing that? 
This picks up any logfiles generated by scripts, which triggers a lot of reloads.

test/mitmproxy/builtins/test_script.py
```diff
@@ -0,0 +1,128 @@
...
+            assert script.parse_command("data\\scripts\\a.py 'foo \\ bar'") == ("data\\scripts\\a.py", 'foo \\ bar', [])
+
+
+def test_load_script():
+    ns = script.load_script(
+        tutils.test_data.path(
+            "data/addonscripts/recorder.py"
+        ), []
+    )
+    assert ns["configure"]
```
mhils: We should test the error case here.

mhils:
The following should exit during startup:
```
echo "foo" > syntaxerr.py
mitmdump -s syntaxerr.py
```

mitmproxy/builtins/script.py
```diff
@@ -0,0 +1,162 @@
...
+    def configure(self, options):
+        for s in options.scripts or []:
+            if not ctx.master.addons.has_addon(s):
+                ctx.log.info("Loading script: %s" % s)
+                sc = Script(s)
+                ctx.master.addons.add(sc)
+        for a in ctx.master.addons.chain:
+            if isinstance(a, Script):
+                if a.name not in options.scripts or []:
+                    ctx.master.addons.remove(a)
```
mhils: I think this needs to invoke `done()` ?

mitmproxy/builtins/script.py
```diff
@@ -0,0 +1,162 @@
...
+                break
+            else:
+                tb = tb.tb_next
+        ctx.log.warn("".join(traceback.format_tb(tb)))
+    finally:
+        sys.argv = oldargs
+        sys.path.pop()
+
+
+def load_script(path, args):
```
mhils: It feels very weird that `load_script` always succeeds. Have you considered moving the scriptenv part into `Script.configure` ?

mitmproxy/builtins/script.py
```diff
@@ -0,0 +1,162 @@
...
+                evt = i
+
+                def prox(*args, **kwargs):
+                    self.run(evt, *args, **kwargs)
+                return prox
+            setattr(self, i, mkprox())
+
+    def run(self, name, *args, **kwargs):
+        # It's possible for ns to be un-initialised if we failed during
+        # configure
```
mhils: I don't think so, because `load_script` never fails?

mitmproxy/builtins/script.py
cortesi: OK, switching to always using PollingObserver

mitmproxy/builtins/script.py
cortesi: I think this comment is for the part where we monitor the directory the script is in, rather than the script. I have use-cases where the specified script is part of a module and imports other stuff. Most of the modifications I care about are in the other stuff.  This also keeps an eye here on a more sophisticated addon system where we specify an entry  point into a module. 

Frankly, my experience with modd/devd tells me that the right thing to do here would be to detach the monitoring into an explicit option (i.e. if path X changes reload addon Y), but we can think about that down the track.

mitmproxy/builtins/script.py
cortesi: load_script can raise, e.g. if the script contains a syntax error. We can make this a bit more user friendly from an error perspective, but load_script definitely doesn't always succeed.

mitmproxy/builtins/script.py
cortesi: See previous comment re load_script

mhils:
Looking at it with a clearer mind: Scripts now have `configure` instead of `start`. However their own arguments are not really part of `options`. This is really confusing for everyone who just uses scripts.

mitmproxy/builtins/script.py
cortesi: addons.remove invokes .done()

PR #1380, opened 2016-07-18
Marked property and filter

dufferzafar:
This tries to replace the marked flow view (activated by `M` key) with a `~marked` filter.

`marked` property in flows:

- [x] Add a `marked` property to flow
- [x] Update `save_marked_flows`
- [x] Update action on pressing `m` (toggle marked attribute)
- [x] Update action on pressing `U` (unmark all flows)
- [x] Ensure proper serialization
- [x] Make necessary changes in `io_compat`

`marked` filter:

- [x] Add a `~marked` filter
- [x] Remove code that currently runs on pressing the `M` key
- [x] Re-add `M` key
    * which will just toggle view filter between `~marked` and what was before
- [x] Ensure that fancy stuff like `~marked & (existing filter)` works

focus flows:

- [x] If I have a marked flow selected, I should have the same flow selected when I toggle marked view on
- [x] If I have an un-marked flow selected, I should ideally have the nearest marked flow selected when toggling marked view on.
- [x] If I have a marked flow selected in marked view, the same flow should be selected when I toggle marked view off.
- [x] I don't know what the current behaviour is re updating the flow list, but if it doesn't currently update, we might as well handle that now.

save flows:

- [x] Change save flows prompt to: "Listed, This"
- [x] Ensure that there are no special functions for `marked` flows - everything should be for `matching`

### DIFF
--- a/mitmproxy/console/common.py
+++ b/mitmproxy/console/common.py
@@ -410,7 +410,7 @@ def raw_format_flow(f, focus, extended):
     return urwid.Pile(pile)
 
 
-def format_flow(f, focus, extended=False, hostheader=False, marked=False):
+def format_flow(f, focus, extended=False, hostheader=False):
     d = dict(
         intercepted = f.intercepted,
         acked = f.reply.acked,
@@ -423,7 +423,7 @@ def format_flow(f, focus, extended=False, hostheader=False, marked=False):
 
         err_msg = f.error.msg if f.error else None,
 
-        marked = marked,
+        marked = f.marked,
     )
     if f.response:
         if f.response.raw_content:
--- a/mitmproxy/console/flowlist.py
+++ b/mitmproxy/console/flowlist.py
@@ -120,7 +120,6 @@ class ConnectionItem(urwid.WidgetWrap):
             self.flow,
             self.f,
             hostheader = self.master.options.showhost,
-            marked=self.state.flow_marked(self.flow)
         )
 
     def selectable(self):
--- a/mitmproxy/console/flowlist.py
+++ b/mitmproxy/console/flowlist.py
@@ -230,7 +230,7 @@ class ConnectionItem(urwid.WidgetWrap):
                 )
         elif key == "U":
             for f in self.state.flows:
-                self.state.set_flow_marked(f, False)
+                f.marked = False
             signals.flowlist_change.send(self)
         elif key == "V":
             if not self.flow.modified():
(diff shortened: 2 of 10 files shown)


dufferzafar:
Add 3 flows. Mark 1st two. Then set limit pattern to "~marked". Only those 2 flows will be visible now.

Now, in this "marked view", what should happen a user unmarks a flow (by pressing the `m` key`) ?

Should we hide the unmarked flows from view now? (in essence refreshing the limit view)

cortesi:
This is the kind of tricky corner-case that makes or breaks UX. If we're viewing a list of marked flows and we un-mark a flow, having it disappear immediately will be confusing and frustrating if we did this accidentally or want to change our mind. I think we should keep the flow in the view, and just remove the mark indicator so it's clear what happened. The user can always toggle marked view on and off to "clean" things. 

Now, how about the opposite case. We're viewing marked flows, and an addon tags a new inbound flow as marked. Here, I do think the marked view should be updated with the new flow.

dufferzafar:
> I think we should keep the flow in the view, and just remove the mark indicator so it's clear what happened.

This is the current behaviour.

> ...We're viewing marked flows, and an addon tags a new inbound flow as marked. Here, I do think the marked view should be updated with the new flow.

Since there are no addons currently that mark flows, do we have to worry about this now?

dufferzafar:
I haven't yet figured out what changes are required in `io_compat` but apart from that, this is ready for a review.

cortesi:
This looks good to me. A few more tweaks:

- If I have a marked flow selected, I should have the same flow selected when I toggle marked view on
- If I have an un-marked flow selected, I should ideally have the nearest marked flow selected when toggling marked view on.
- If I have a marked flow selected in marked view, the same flow should be selected when I toggle marked view off.
- I don't know what the current behaviour is re updating the flow list, but if it doesn't currently update, we might as well handle that now. I remember working on this years ago for the view mechanism, so it may "just work".

cortesi:
Good work on the position retention. I see a crash when toggling mark on when a filter is already set. I think the simplest thing to do there is just to clear the current filter.

dufferzafar:
The Travis test still fails because I suspect that the `dumpfile-011` is actually of version 0.18, but I'm reasonably sure that 562b8c4 is the only change required.

mhils:
I restarted the tests. LGTM if everything passes! :+1:

dufferzafar:
I've changed the save flows prompt to "Listed flows, This flows" and with that this ready for final review!

PR #1388, opened 2016-07-21
Flow editing v2

cle1000:


web/src/js/components/FlowView/Messages.jsx
```diff
@@ -68,17 +68,18 @@ function ResponseLine({ flow, readonly, updateFlow }) {
 
 const Message = connect(
     state => ({
-        flow: state.ui.flow.modifiedFlow || state.flows.byId[state.flows.selected[0]],
+        flow: state.flows.byId[state.flows.selected[0]],
```
cle1000: i removed modifiedFlow, because otherwise react js propagates the changes down to the contentloader and then the content will be reloaded so i cannot edit anything

web/src/js/components/FlowView/ToggleEdit.jsx
```diff
@@ -10,11 +10,11 @@ ToggleEdit.propTypes = {
     stopEdit: PropTypes.func.isRequired,
 }
 
-function ToggleEdit({ isEdit, startEdit, stopEdit, flow }) {
+function ToggleEdit({ isEdit, startEdit, stopEdit, flow, old_flow }) {
     return (
         <div className="edit-flow-container">
             {isEdit ?
-                <a className="edit-flow" onClick={() => stopEdit(flow)}>
+                <a className="edit-flow" onClick={() => stopEdit(flow, old_flow)}>
```
mhils: We should stay consistent and call these `modifiedFlow` and `flow`

web/src/js/ducks/ui/flow.js
```diff
@@ -87,9 +88,31 @@ export function updateEdit(update) {
     return { type: UPDATE_EDIT, update }
 }
 
-export function stopEdit(flow) {
+export function uploadContent(flow, content, type){
     return (dispatch) => {
-        dispatch(flowsActions.update(flow, flow)).then(() => {
+            dispatch(flowsActions.updateContent(flow, content, type)).then( () => {
```
mhils: Why does uploadContent update the flow as well?

web/src/js/ducks/ui/flow.js
```diff
@@ -87,9 +88,31 @@ export function updateEdit(update) {
...
+}
+
+export function stopEdit(modified_flow, old_flow) {
+    //make diff of modified_flow and old_flow
+    return (dispatch) => {
+        let flow = {...modified_flow}
+
+        if (flow.response.content) {
+            dispatch(flowsActions.updateContent(flow, flow.response.content, "response"))
+            flow.response = _.omit(flow.response, "content")
```
mhils: If this is a copy anyway (L103), we  can just call `delete flow.response.content`.

web/src/js/components/FlowView/Messages.jsx
mhils: Can you explain this a bit further?

cle1000: `flow: state.ui.flow.modifiedFlow || state.flows.byId[state.flows.selected[0]],` changes the flow prop, and then `<ContentLoader ..> <View.... > </ContentLoader> will be rerendered and the content will be reloaded.

web/src/js/ducks/ui/flow.js
cle1000: STOP_EDIT: because i thought that after a content upload it would be good to close the edit mode if it is open. but it would be better when the editor is refreshed with the new content

mhils: We probably just want to listen to the update flow action in the reducer to disable editing mode once the flow is updated by something?

web/src/js/components/FlowView/Messages.jsx
mhils: I think that providing the modified flow is essential here, does Header editing work with this change!? The ContentView/ContentLoader needs to be fixed so that it doesn't rerender.

cle1000: in case that we just edit either the response or the request, this will work. if we want (and i think we want that) we can edit repsonse and request simultaneously than this is not possible

cle1000:
flow updates only differences between modified flow and old flow, it is working but if i update the content and the header together i got an error (dont know why i will check it again tomorrow) if you update the content or the header everything is fine.

web/src/js/ducks/ui/flow.js
```diff
@@ -1,18 +1,22 @@
...
-             STOP_EDIT        = 'UI_FLOWVIEW_STOP_EDIT'
+             STOP_EDIT        = 'UI_FLOWVIEW_STOP_EDIT',
+             UPLOAD_CONTENT   = 'UI_FLOWVIEW_UPLOAD_CONTENT'
 
 
 const defaultState = {
     displayLarge: false,
     modifiedFlow: false,
     contentView: 'ViewAuto',
+    lastFileUpload: false,
```
mhils: I don't fully understand why we need this. If I upload a new content, that should trigger a flow update, which should update the message? What am I missing?

web/src/js/utils.js
```diff
@@ -105,3 +105,17 @@ fetchApi.put = (url, json, options) => fetchApi(
         ...options
     }
 )
+
+
+export function getDiff(obj1, obj2) {
+    let result = {...obj2};
+    _.forIn(obj1, (value, key) => {
```
mhils: `for(i in obj1) {` ?

web/src/js/utils.js
```diff
@@ -105,3 +105,17 @@ fetchApi.put = (url, json, options) => fetchApi(
...
+    	  if(_.isEqual(obj2[key], obj1[key]))
+            result[key] = undefined;
+        else if(typeof Array.isArray(obj2[key]) && Array.isArray(obj2[key]))
+            result[key] = {...obj2[key]};
```
mhils: This looks wrong?

web/src/js/utils.js
```diff
@@ -105,3 +105,17 @@ fetchApi.put = (url, json, options) => fetchApi(
         ...options
     }
 )
+
+
+export function getDiff(obj1, obj2) {
```
mhils: This should have tests right away. :smiley:

web/src/js/ducks/ui/flow.js
```diff
@@ -87,10 +98,27 @@ export function updateEdit(update) {
...
+            dispatch(flowsActions.updateContent(flow, diff.response.content, "response"))
+            delete diff.response.content
+        }
+        if (diff.request && diff.request.content) {
+            dispatch(flowsActions.updateContent(flow, diff.request.content, "request"))
+            delete diff.request.content
+        }
+
+        dispatch(flowsActions.update(flow, diff)).then(() => {
+            dispatch(flowsActions.updateFlow(modified_flow))
```
mhils: This silentlys persist the `.content` property on the flow.

mhils: We have multiple dispatch promises here. We should wait for them all (`Promise.all`) before emitting updates :smiley:

web/src/js/ducks/ui/flow.js
```diff
@@ -87,10 +98,27 @@ export function updateEdit(update) {
     return { type: UPDATE_EDIT, update }
 }
 
-export function stopEdit(flow) {
+export function uploadContent(flow, content, type){
```
mhils: This looks like a second action creator that does exactly the same thing as `updateContent` but emits an additional event. I don't think that separation is useful.

mhils:
This is a good start, but it looks like there are quite a few band-aids in this at the moment. In particular, we should isolate the upload content button from the rest of the edit UI - let's just `connect()` it and be its own free-standing thing. The rest of the editing functionality should already be able to get information about content changes by listening to FLOWS_UPDATE and checking if the flow id matches the currently edited flow.

web/src/js/ducks/ui/flow.js
cle1000: my problem here was that the contentloader compontent did not rerender because the flow does not really change because the content will not be updated.

web/src/js/utils.js
cle1000: i know, but the problem was that flow.response.headers is an array with lots of (tuple) arrays. if i generally replace all values which are equal we are not able to update the headers. e.g if you update `["host",  "abc.def"]` to  `["host",  "testf"]` will result into  `[undefined,  "testf"]`. thats the reason for the strange array handling

cle1000:
` listening to FLOWS_UPDATE` do you have an example how a component can listen to an event ?

mhils:
> do you have an example how a component can listen to an event ?

The component should not listen, the reducer should. If the see an update to the current active flow in the reducer, we probably just want to stop editing (set modifiedFlow to false) to display the updated version. Does that make sense?

web/src/js/ducks/ui/flow.js
mhils: Generally speaking, after receiving an update for the currently displayed flow, the contentloader should rerender. Most extreme example: If I upload a new flow content in window a, it should rerender the flow in window b. if that's not the case, we should fix this issue and not add band-aids.

web/src/js/utils.js
mhils: My understanding is that we would like to send the whole headers array on update, not just parts of it. So if the headers are not equal, we should just send the whole array. Otherwise, updating on the server-side would be relatively tricky.

cle1000: yes, i kown. its was just an example. in the getDiff methode i just use the array of the secound obj if they are not equal. --> headers: [...] is an array and so if anything changes in the header, the header will be updated (completly)

cle1000:
hmm the problem here was that content is not always part of the flow. so i just select a flow, i do not edit it and upload a new file for the content. after that i the update event reloads the flow but the flow doesnt change -> no update. 
stop editing:  i thought if you are in the edit mode, then you can upload a file for the content an change it directly in mitmweb. if i set modifiedFlow to false after an update, than this solves one case of the problem above -> if i am in edit mode and upload a new file than the ui will be refreshed.

web/src/js/utils.js
mhils: I think arrays do not need this special case here at all. First, that assignment should have been `[...obj2[key]]`, not `{...obj2[key]}`. Second, `result = {...obj2};` already copies over the value as-is. As everything is immutable, we don't need to create a copy of the array.

I think this should clear up itself when writing tests :smiley:

mhils:
> hmm the problem here was that content is not always part of the flow. so i just select a flow, i do not edit it and upload a new file for the content. after that i the update event reloads the flow but the flow doesnt change -> no update. 

This is actually in interesting issue. We probably should transfer some `contentHash` attribute next to `contentLength` so that content modifications can be detected on the client unambiguously.

Other than that, if we receieve a flow update while we are editing, this means that someone else has just updated the flow. In this case, I think it's fair to just exit the editing mode and display the updated flow.

Maybe let's skype on Monday if things are still unclear :smiley:

web/src/js/ducks/ui/flow.js
```diff
@@ -94,15 +94,6 @@ export function updateEdit(update) {
 export function stopEdit(flow, modified_flow) {
     let diff = getDiff(flow, modified_flow)
     return (dispatch) => {
-        if (diff.response && diff.response.content) {
-            dispatch(flowsActions.updateContent(flow, diff.response.content, "response"))
-            delete diff.response.content
-        }
-        if (diff.request && diff.request.content) {
-            dispatch(flowsActions.updateContent(flow, diff.request.content, "request"))
```
cle1000: we should remove this line right ? because the sever should tell the client that the flow is changed ?  but thats not working :( i try to find the bug, but i did not found it

PR #1404, opened 2016-07-22
 [web] introduce FocusHelper

gzzhanghao:


web/src/js/components/FlowView.jsx
```diff
@@ -78,9 +60,12 @@ export default class FlowView extends Component {
                     active={active}
                     onSelectTab={this.props.selectTab}
                 />
-                <Tab ref={ tab => this.tabComponent = tab } flow={flow} updateFlow={updateFlow} />
```
cle1000: just a hint, if you don't know
<input ref="myInput" />
usage: this.refs.myInput

https://facebook.github.io/react/docs/more-about-refs.html

PR #1406, opened 2016-07-22
[web] Update view with sortedUpdate

gzzhanghao:
Fix https://github.com/mitmproxy/mitmweb/issues/36


mhils:
Awesome. Can we add a test for this and not add the compiled assets in the PR?

web/src/js/__tests__/ducks/tutils.js
```diff
@@ -10,3 +10,9 @@ export function createStore(parts) {
         applyMiddleware(...[thunk])
     )
 }
+
+describe('tutils', () => {
```
mhils: What's the point of this? :smiley:

web/src/js/ducks/utils/view.js
```diff
@@ -54,21 +54,30 @@ export default function reduce(state = defaultState, action) {
...
-                ...nextState,
-                ...sortedInsert(nextState, action.item, action.sort)
+            if (hasNewItem && hasOldItem) {
+                return {
+                    ...state,
+                    ...sortedUpdate(state, action.item, action.sort),
+                }
             }
-
+            break;
```
mhils: We should never hit this, right?

mhils:
Thanks for the changes, but why is ever testfile called *Spec now? How about we make a negative RegEx just excluding the tutils?

gzzhanghao:
I think it's better to have a general method to distinguish test files and utilities, we may have `__tests__/utils/*.js` sometimes later. :)

PR #1410, opened 2016-07-23
Keep maturing scripts and addons

cortesi:
- Script .start() can now return a class to handle all hooks from that point in. This is the last step in completely unifying scripts and addons, and gives us a nicer way to handle stateful data. Refactor some examples accordingly.
- Tweak change detection to make things a little better. We're never going to be completely happy with how this works unless we more or less recreate modd in mitmproxy.

### DIFF
--- a/mitmproxy/builtins/script.py
+++ b/mitmproxy/builtins/script.py
@@ -16,6 +16,19 @@ import watchdog.events
 from watchdog.observers import polling
 
 
+class NS:
+    def __init__(self, ns):
+        self.__dict__["ns"] = ns
+
+    def __getattr__(self, key):
+        if key not in self.ns:
+            raise AttributeError("No such element: %s", key)
+        return self.ns[key]
+
+    def __setattr__(self, key, value):
+        self.__dict__["ns"][key] = value
+
+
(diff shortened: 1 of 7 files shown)


mitmproxy/builtins/script.py
```diff
@@ -16,6 +16,19 @@
 from watchdog.observers import polling
 
 
+class NS:
```
mhils: What's the point of this class?

cortesi: It lets us access the event objects in the namespace dictionary as attributes. This unifies the interface with Addons, which means that ns can just be a class if it's been replaced by the script .start.

mhils: I see. Can't we do a simple
```python
class Namespace:
    def __init__(self, entries): 
        self.__dict__.update(entries)
```
?

cortesi: That would be more brittle. The namespace that returns from the script can contain arbitrary names. What happens if "__getattribute__" is in there? I mean, I could sit and analyse this to figure it out, but why bother if we can do something simple that's safer?

PR #1413, opened 2016-07-23
http2: fix kill

### LINKED ISSUE #1395: Error when killing HTTP/2 Flow

1. `mitmproxy`
2. Add catch-all interception filter
2. Visit HTTP/2 URL
3. Press `X`

Flow should be killed.

##### What went wrong?
image

When fixing this, let's make sure that killing responses works as well.

---

Mitmproxy Version: master
Operating System: Ubuntu 14.04 x64

Kriechi:
fixes #1395

Not sure if this will do the trick...


mhils:
This brings up an interesting point: If I kill a single request in a HTTP2 connection, should that kill just that request or the whole connection? Looking at server replay with `--kill`, the former option seems right, so I'll just merge this for now.

PR #1436, opened 2016-07-26
Use jsbeautifier from pip

### LINKED ISSUE #1431: console: Javascript content view broken

1. Use Python3 to view application/javascript data

No crash

##### What went wrong?

```
Content viewer failed:
  File "/Users/cortesi/mitmproxy/mitmproxy/mitmproxy/contentviews.py", line 634, in get_content_view
    ret = viewmode(data, **metadata)
  File "/Users/cortesi/mitmproxy/mitmproxy/mitmproxy/contentviews.py", line 151, in __call__
    return content_types_map[ct][0](data, **metadata)
  File "/Users/cortesi/mitmproxy/mitmproxy/mitmproxy/contentviews.py", line 401, in __call__
    res = jsbeautifier.beautify(data, opts)
  File "/Users/cortesi/mitmproxy/mitmproxy/mitmproxy/contrib/jsbeautifier/__init__.py", line 93, in beautify
    return b.beautify(string, opts)
  File "/Users/cortesi/mitmproxy/mitmproxy/mitmproxy/contrib/jsbeautifier/__init__.py", line 218, in beautify
    token_text, token_type = self.get_next_token()
  File "/Users/cortesi/mitmproxy/mitmproxy/mitmproxy/contrib/jsbeautifier/__init__.py", line 443, in
get_next_token
    if c in self.wordchar:
TypeError: 'in <string>' requires string as left operand, not int
```

---

Mitmproxy version: 0.18
Python version: 3.5.1
Platform: Darwin-15.6.0-x86_64-i386-64bit
SSL version: OpenSSL 1.0.2h  3 May 2016
Mac version: 10.11.6 ('', '', '') x86_64

dufferzafar:
This will fix #1431

Things still left to do:
- [x] Use proper `decode` or `_native` etc. which will fix the Travis failure.
- [x] Delete jsbeautifier files from repo

- [ ] Use cssbeautifier instead of cssutils?


mhils:
The contentviews dependencies are optional - we should either make them all mandatory or make sure that mitmproxy still runs without jsbeautifier being installed.

dufferzafar:
I'd be in favor of just installing the dependencies so that things work out of the box. 

`jsbeautifier` package actually also has a `cssbeautfier` so we could use it directly instead of `cssutils`?

mitmproxy/contentviews.py
```diff
@@ -398,7 +389,7 @@ class ViewJavaScript(View):
     def __call__(self, data, **metadata):
         opts = jsbeautifier.default_options()
         opts.indent_size = 2
-        res = jsbeautifier.beautify(data, opts)
+        res = jsbeautifier.beautify(strutils.native(data), opts)
```
mhils: This is going to fail for any non-valid utf8 :wink:

setup.py
```diff
@@ -67,10 +67,12 @@
         "configargparse>=0.10, <0.11",
         "construct>=2.5.2, <2.6",
         "cryptography>=1.3, <1.5",
+        "cssutils>=1.0.1, <1.1",
         "Flask>=0.10.1, <0.12",
         "h2>=2.4.0, <3",
         "html2text>=2016.1.8, <=2016.5.29",
         "hyperframe>=4.0.1, <5",
+        "jsbeautifier>=1.6.3"
```
mhils: We should set a 1.7 upper bound for this as well. :smiley:

mhils:
**+14 −1,583**

Way to go :+1:  :smile:

mhils:
Can we have a very simple test for this view as well? This is a bit more important now that this is an external dependency which may change without us noticing.

dufferzafar:
We already have [tests for javascript contentview](https://github.com/mitmproxy/mitmproxy/pull/1436/commits/1cffa5f46b84928f1dd211806122d01700345b40#diff-03a6481eb84a368bad5be0547f7f4286R108) but what I don't get is why they didn't fail before.

PR #1441, opened 2016-07-27
Integrate mitmproxy contentviews

cle1000:
first PR for this issue, can you tell me more about the content generator?


mitmproxy/web/app.py
```diff
@@ -364,7 +367,8 @@ def get(self):
                 anticomp=self.master.options.anticomp,
                 stickyauth=self.master.options.stickyauth,
                 stickycookie=self.master.options.stickycookie,
-                stream= self.master.options.stream_large_bodies
+                stream= self.master.options.stream_large_bodies,
+                contentViews= map(lambda v : v.name, contentviews.views)
```
mhils: `[v.name for v in contentviews.views]`
(http://www.artima.com/weblogs/viewpost.jsp?thread=98196)

web/src/js/components/ContentView/ContentViews.jsx
```diff
@@ -57,4 +57,14 @@ function ViewAuto({ message, flow, readonly, onChange }) {
     return <View message={message} flow={flow} readonly={readonly} onChange={onChange}/>
 }
 
-export { ViewImage, ViewRaw, ViewAuto, ViewJSON }
+function ViewServer({contentView, content}){
```
mhils: I think we have a possible misunderstanding here. We want one button per server-side contentview - for the user, there should not be a difference between client-side and server-side views.

web/src/js/components/ContentView.jsx
```diff
@@ -33,10 +33,10 @@ function ContentView(props) {
         return <MetaViews.ContentTooLarge {...props} onClick={displayLarge}/>
     }
 
-    const View = ContentViews[contentView]
+    const View = ContentViews[contentView] || ContentViews['ViewServer']
```
mhils: contentView should never be undefined - there's  a default in the flow ducks.

web/src/js/flow/utils.js
```diff
@@ -43,13 +43,13 @@ export var MessageUtils = {
...
     },
-    getContentURL: function (flow, message) {
+    getContentURL: function (flow, message, view) {
         if (message === flow.request) {
             message = "request";
         } else if (message === flow.response) {
             message = "response";
         }
-        return "/flows/" + flow.id + "/" + message + "/content";
+        return "/flows/" + flow.id + "/" + message + "/content" +  (view ? "?content_view="+view : "");
```
mhils:
```javascript
(view ? `?content_view=${view}` : '')

```

mhils:
Content views:

https://github.com/mitmproxy/mitmproxy/blob/d97fe767dc7b8ea47f0e170c6f002c506f606d57/mitmproxy/contentviews.py#L621-L632
https://github.com/mitmproxy/mitmproxy/blob/d97fe767dc7b8ea47f0e170c6f002c506f606d57/mitmproxy/contentviews.py#L117-L135

> The content generator yields lists of (style, text) tuples, where each list represents a single line. 

In a nutshell, a generator is a sequence you can iterate over. The main difference compared to a list is that generators are computed lazily. See: https://jeffknupp.com/blog/2013/04/07/improve-your-python-yield-and-generators-explained/

```python
>>> import pprint
>>> from mitmproxy import contentviews
>>> view = contentviews.ViewURLEncoded()
>>> description, generator = view(b"foo=42&bar=43&baz=44")
>>> description
'URLEncoded form'
>>> generator
<generator object format_dict at 0x041DE1E0>
>>> lines = list(generator)
>>> pprint.pprint(lines)
[[('header', 'foo: '), ('text', '42')],
 [('header', 'bar: '), ('text', '43')],
 [('header', 'baz: '), ('text', '44')]]
```

The style is normally just "text", but it can also be "highlight", "offset" or "header". This is probably best solved by giving the span a matching className.

web/src/js/components/ContentView.jsx
cle1000: ContentViews[contentView] is undefined if that content view is an client contentview (like img, auto, or raw with the editor option) then i will use the "server content view". "server content view" is a view which uses the name of the active content view form the redux state and applies an request with the corresponding parameter.

mhils: Ok, that makes a ton of sense (I assume you mean server, not client, in the first sentence). Let's keep it as-is!

mitmproxy/web/app.py
```diff
@@ -303,6 +306,36 @@ def post(self, flow_id):
 
 
 class FlowContent(RequestHandler):
+    def _get_content_view(self, message, viewmode):
```
mhils: We have more or less the same function three times now (mitmproxy/console/flowview, mitmproxy/builtins/dumper and here). I think this should be centralized.

mitmproxy/web/app.py
```diff
@@ -336,7 +369,21 @@ def get(self, flow_id, message):
         self.set_header("Content-Type", "application/text")
         self.set_header("X-Content-Type-Options", "nosniff")
         self.set_header("X-Frame-Options", "DENY")
-        self.write(message.raw_content)
+
+        cv = self.get_argument("cv", None)
+        if cv:
+            self.set_header("Content-Encoding", "")
+            viewmode = next(v for v in contentviews.views if v.name == cv)
```
mhils: I believe we have `contentviews.get` or something along those lines.

mitmproxy/web/app.py
```diff
@@ -336,7 +369,21 @@ def get(self, flow_id, message):
...
+            description, lines  = self._get_content_view(
+                 message, viewmode
+            )
+
+            self.write(dict(
```
mhils: This may need a JSON content type - at this point, I believe it makes sense to move the normal content and contentviews into two distinct endpoints.

web/src/css/contentview.less
```diff
@@ -0,0 +1,10 @@
+.contentview {
+     .header {
+         font-weight: bold;
+     }
+    .highlight{
+        color: pink;
```
mhils: :heart_eyes:

mitmproxy/builtins/dumper.py
```diff
@@ -73,20 +73,10 @@ def _echo_message(self, message):
             elif content:
                 self.echo("")
 
-                try:
```
mhils: You're missing the part in line 66+.

cle1000:
i am struggeling with the sticky contentview options, currently its not working, but the rest should be fine

mhils:
Can you resolve the merge conflicts on this please?

mitmproxy/web/app.py
```diff
@@ -338,6 +340,40 @@ def get(self, flow_id, message):
...
+            filename = self.flow.request.path.split("?")[0].split("/")[-1]
+
+        filename = re.sub(r"[^\w\" \.\-\(\)]", "", filename)
+        cd = "attachment; filename={}".format(filename)
+        self.set_header("Content-Disposition", cd)
+        self.set_header("Content-Type", "application/json")
+        self.set_header("X-Content-Type-Options", "nosniff")
+        self.set_header("X-Frame-Options", "DENY")
+
+        self.set_header("Content-Encoding", "")
```
mhils: This endpoint is just emitting JSON, so I believe we don't need any of the stuff above. For example, the Content-Disposition is useful for download filenames. The view will not be downloaded, only the original content.

web/src/css/flowdetail.less
```diff
@@ -41,6 +42,24 @@
 
 }
 
+.flowview-container{
+    display: flex;
+    flex-direction: column;
+    padding: 0 !important;
```
mhils: How about we just adjust `.flow-detail section`? :wink:

web/src/css/flowdetail.less
```diff
@@ -41,6 +42,24 @@
...
+}
+
+.flowview-body{
```
mhils: It feels like this could just be `section.request, section.response` inserted at line 14/15.

web/src/css/flowdetail.less
```diff
@@ -41,6 +42,24 @@
...
+    overflow: auto;
+    padding: 5px 12px 0;
+}
+
+.flowview-footer{
```
mhils: And this just `> footer` in there.

web/src/js/components/ContentView.jsx
```diff
@@ -33,18 +32,11 @@ function ContentView(props) {
...
-            <div className="view-options text-center">
-                <ViewSelector message={message}/>
-                &nbsp;
-                <DownloadContentButton flow={flow} message={message}/>
-                &nbsp;
-                <UploadContentButton uploadContent={uploadContent}/>
-            </div>
+        <div className="contentview">
+            <View flow={flow} message={message} contentView={contentView} readonly={readonly} onChange={onContentChange}/>
+            <ShowFullContentButton/>
```
mhils: :+1:

web/src/js/components/common/Button.jsx
```diff
@@ -1,14 +1,15 @@
...
     onClick: PropTypes.func.isRequired,
     text: PropTypes.string,
     icon: PropTypes.string
 }
 
-export default function Button({ onClick, text, icon, disabled }) {
+export default function Button({ onClick, text, icon, disabled, isXs, className }) {
     return (
-        <div className={"btn btn-default"}
+        <div className={classnames(className, 'btn btn-default', { 'btn-xs': isXs})}
```
mhils: isXs seems superfluous - we can just pass `btn-xs` as className ?

web/src/js/components/ContentView/ContentViews.jsx
```diff
@@ -18,43 +20,69 @@ function ViewImage({ flow, message }) {
...
-ViewJSON.propTypes = {
-    content: React.PropTypes.string.isRequired,
-}
-function ViewJSON({ content }) {
-    let json = content
-    try {
-        json = JSON.stringify(JSON.parse(content), null, 2);
-    } catch (e) {
-        // @noop
+    componentWillMount(){
```
mhils: This should be done in the constructor, otherwise we'll have to render twice if I understand things correctly.

web/src/js/components/ContentView/ContentViews.jsx
```diff
@@ -18,43 +20,69 @@ function ViewImage({ flow, message }) {
...
-    return <pre>{json}</pre>
-}
-ViewJSON = ContentLoader(ViewJSON)
+    componentWillReceiveProps(nextProps){
+        this.setContentView(nextProps)
+    }
+    setContentView(props){
+        try {
+            this.data = JSON.parse(props.content)
+        }catch(err) {
```
mhils: Why should this ever fail?

web/src/js/components/ContentView/ContentViews.jsx
```diff
@@ -18,43 +20,69 @@ function ViewImage({ flow, message }) {
...
+            this.data= {lines: [], description: err.message}
+        }
 
+        props.setContentViewDescription(props.contentView != this.data.description ? this.data.description : '')
+
+        let isFullContentShown = this.data.lines.length < props.maxLines
+        if (isFullContentShown) props.setShowFullContent(true)
```
mhils: This really looks like some spaghetti - it looks like we should replace the ContentLoader with something living in redux, but that's not in the scope of this PR.

web/src/js/components/ContentView/ViewSelector.jsx
```diff
@@ -1,47 +1,82 @@
...
+    }
+    close() {
+        this.props.setContentViewSelectorOpen(false)
+        document.removeEventListener('click', this.close)
+    }
 
-            <ViewButton name="ViewAuto">auto: {autoViewName}</ViewButton>
+    onDropdown(e){
+        e.preventDefault()
+        this.props.setContentViewSelectorOpen(!this.props.isContentViewSelectorOpen)
```
mhils: Ugh. I appreciate the effort, but let's keep that in the plain react state - there's no way that we ever need this outside of this component.

web/src/js/components/FlowView/Messages.jsx
```diff
@@ -81,27 +82,35 @@ const Message = connect(
...
+                    />
+
+                    <hr/>
+                    <ContentView
+                        readonly={!isEdit}
+                        flow={flow}
+                        onContentChange={content => updateFlow({ request: {content}})}
+                        message={flow.request}/>
+                </div>
+                <div hidden={noContent} className="flowview-footer">
```
mhils: `{ !noContent && /* dom here */ }` ?

mhils:
This looks good. Let's get this in a mergable state ASAP so that we can move that into master and do smaller iterations there! :smiley:

web/src/js/components/ContentView/ContentViews.jsx
cle1000: ```warning.js:44 Warning: setState(...): Cannot update during an existing state transition (such as within `render` or another component's constructor). Render methods should be a pure function of props and state; constructor side-effects are an anti-pattern, but can be moved to `componentWillMount`.```

web/src/js/components/ContentView/ContentViews.jsx
cle1000: e.g. if you edit, a request with no content, and than switch to a different viewer, in this case there will be an error with can not be transformed into json

PR #1447, opened 2016-07-28
Verify upstream certificates by default

### LINKED ISSUE #1111: Verify upstream certificates by default.

mhils:
closes #1197
fixes #1111

I finally got around to take a closer look at @s4chin's excellent certificate verification PR! :smiley:
This PR supersedes #1197. The core functionality is the same, but I did a bit of work on the UX.

Here's the mitmdump output:
```
λ curl -x localhost:8080 -k https://expired.badssl.com/

127.0.0.1:54216: clientconnect
127.0.0.1:54216: Certificate Verification Error for expired.badssl.com: certificate has expired (errno: 10, depth: 0)
127.0.0.1:54216: Invalid certificate, closing connection. Pass --insecure to disable validation.
127.0.0.1:54216: clientdisconnect

λ curl -x localhost:8080 -k https://wrong.host.badssl.com/

127.0.0.1:54221: clientconnect
127.0.0.1:54221: Certificate Verification Error for wrong.host.badssl.com: hostname 'wrong.host.badssl.com' doesn't match either of '*.badssl.com', 'badssl.com'
127.0.0.1:54221: Invalid certificate, closing connection. Pass --insecure to disable validation.
127.0.0.1:54221: clientdisconnect

λ curl -x localhost:8080 -k https://self-signed.badssl.com/

127.0.0.1:54224: clientconnect
127.0.0.1:54224: Certificate Verification Error for self-signed.badssl.com: unable to get local issuer certificate (errno: 20, depth: 0)
127.0.0.1:54224: Invalid certificate, closing connection. Pass --insecure to disable validation.
127.0.0.1:54224: clientdisconnect

λ curl -x localhost:8080 https://example.com/

127.0.0.1:54227: clientconnect
127.0.0.1:54227: Client Handshake failed. The client may not trust the proxy's certificate for example.com.
127.0.0.1:54227: clientdisconnect
```

Here's how it looks like in mitmproxy:

image

Here's how it looks like in the client:

image

@cortesi, what do you think?


mhils:
So, turns out that codecov really helped with that PR - I mistakenly believed we'd trigger the "invalid certificate" path in mitmproxy, but we actually just triggered a "no sni" exception before that. :tada:

mitmproxy/proxy/config.py
```diff
@@ -83,24 +83,18 @@ def __init__(self, options):
...
-        )
-        if conflict:
+        # type: (mitmproxy.options.Options, Any) -> None
+        if options.add_upstream_certs_to_client_chain and not options.ssl_insecure:
             raise exceptions.OptionsError(
-                "The verify-upstream-cert and add-upstream-certs-to-client-chain "
-                "options are mutually exclusive. If upstream certificates are verified "
-                "then extra upstream certificates are not available for inclusion "
-                "to the client chain."
+                "The verify-upstream-cert requires certificate verification to be disabled. "
```
cortesi: "verify-upstream-cert option"

PR #1450, opened 2016-07-30
Fix platform import, substitute "~" with user's home

vhaupert:
Fixes two small bugs:

1. Fix platform import on Linux using python3 

```sh
$ mitmproxy -T --host
mitmproxy: Transparent mode not supported on this platform.
```

Using Python3 `sys.platform` returns `linux` instead of `linux2` (Python2). This causes the platform import to fail on python3.

2. Substitute tilde with user's home.

Delivering the cert `~/.mitmproxy/mitmproxy-ca-cert.pem` using `mitm.it` fails because the tilde is not replaced with the user's home directory.


mhils:
Thanks, this is super helpful! :smiley: :cake:

PR #1451, opened 2016-07-30
Integrated encode/decoder for brotli

amenuor:


mhils:
Thanks for the PR! :smiley:

This looks great. My only concern is that brotlipy is a binary dependency - @Lukasa, do you have any practical experience how brotlipy works out with PyInstaller?

mhils:
Seems to work just fine, so let's go ahead with this. :smiley: 

Thanks again for your work! :blush: :cake:

amenuor:
Super :)

Lukasa:
Ah, sorry about the delay! My answer was going to be: no, but it's a binary package that is built entirely using setuptools and the setuptools CFFI integration. So as long as pyinstaller does sensible things, there should be no trouble. Which reminds me, I should push out a new release of brotlipy!

PR #1452, opened 2016-07-30
Support Comma Separated Cookies

### LINKED ISSUE #1434: cookies: support multi-cookie, comma-separated header variant
At the moment, we only support strictly compliant cookies, where each cookie is in a separate header. The justification for this we give in the module docstring is that "violations of RFC6265 that makes the parsing problem indeterminate are much more common than genuine occurences of the multi-cookie variants". However, a recent sample from Netograph shows old-style cookies in ~9% of sites, with many high-profile sources like DoubleClick. There's no way around it - we simply must correctly parse this variant.

dufferzafar:
Fixes #1434

Todo:

- [x] Rebased with master
- [x] The `_read_pairs` change should only affect `SetCookie` and not `Cookie`
- [x] Update `format_set_cookie_header` to take in a list of `SetCookie`s & return the comma separated list.
- [x] Add new tests for multi-cookies

### DIFF
--- a/netlib/http/cookies.py
+++ b/netlib/http/cookies.py
@@ -185,8 +185,7 @@ def parse_set_cookie_headers(headers):
     for header in headers:
         cookies = parse_set_cookie_header(header)
         if cookies:
-            for cookie in cookies:
-                name, value, attrs = cookie
+            for name, value, attrs in cookies:
                 ret.append((name, SetCookie(value, attrs)))
     return ret
 
@@ -221,6 +220,8 @@ def parse_set_cookie_header(line):
 
     if cookies:
         return cookies
+    else:
+        return None
 
 
 def format_set_cookie_header(name, value, attrs):
--- a/netlib/http/cookies.py
+++ b/netlib/http/cookies.py
@@ -128,6 +128,15 @@ def _read_cookie_pairs(s, off=0):
             break
 
     return pairs, off
+
+
+def _read_set_cookie_pairs(s, off=0):
+    """
+        Read pairs of lhs=rhs values from SetCookie headers while handling multiple cookies.
+
+        off: start offset
+        specials: attributes that are treated specially
+    """
     cookies = []
     pairs = []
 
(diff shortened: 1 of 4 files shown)


netlib/http/cookies.py
```diff
@@ -161,10 +175,11 @@ def _parse_set_cookie_pairs(s):
...
     ret = []
     for header in headers:
-        v = parse_set_cookie_header(header)
-        if v:
-            name, value, attrs = v
-            ret.append((name, SetCookie(value, attrs)))
+        cookies = parse_set_cookie_header(header)
+        if cookies:
+            for cookie in cookies:
+                name, value, attrs = cookie
```
Kriechi: I think you can directly use this as loop args: `for name, value, attrs in cookies:`

netlib/http/cookies.py
```diff
@@ -187,13 +202,17 @@ def parse_set_cookie_header(line):
...
-    pairs = _parse_set_cookie_pairs(line)
-    if pairs:
-        return pairs[0][0], pairs[0][1], CookieAttrs(tuple(x) for x in pairs[1:])
+    cookies = [
+        (pairs[0][0], pairs[0][1], CookieAttrs(tuple(x) for x in pairs[1:]))
+        for pairs in _parse_set_cookie_pairs(line) if pairs
+    ]
+
+    if cookies:
+        return cookies
```
Kriechi: This implicitly return None - which a neat way to do it - but I fear a bit too much 😄 
I think we should make it explicit in that case - what do you think?

Kriechi:
Maybe the py.test parametrize feature would be a good fit to cleanup the cookie tests?
http://docs.pytest.org/en/latest/parametrize.html

dufferzafar:
@cortesi @resam I understand that we should separate cookies based on comma in the `SetCookie` header.

But what should happen if there is a comma in the `Cookie` header.

For eg: in `Cookie: name=value, key=value"

cortesi:
The RFC says that cookie values can't contain commas, so let's assume that's correct for now. So for a first pass, we can assume that the only place we need to deal with comma ambiguity is in the expires attribute. The caveat here is that browsers might do something different, and if they do, we should support it. We can get to that later. 

I'm super excited about this PR because I have a very real use for this. If you could rebase and push  ahead with the fixes you've outlined, I'll test on the Netograph corpus and merge.

mhils:
@cortesi @dufferzafar: What's the status here? 😃

dufferzafar:
I fixed all the current tests this week. Now I need to add some new tests specific to comma separated cookies, and this will be done.

mhils:
Sounds great, thanks for the update! :smiley:

dufferzafar:
New tests added! Ready for review.

PR #1453, opened 2016-07-30
Rename "Limit" feature to "Filter View"

### LINKED ISSUE #542: Rename "Limit" Feature
In a nutshell, I feel that "filter" describes the limit/filter feature more accurately than "limit".  Also, `f` is not taken as a keyboard shortcut. There were a few users in #mitmproxy asking for said feature, not realizing that limit does provide this functionality.

dufferzafar:
Fixes #542.

Two filter view related functionalities:

- [ ] Ability to save filtered flows from the flowview is [missing](https://github.com/dufferzafar/mitmproxy/blob/master/mitmproxy/console/flowview.py#L83).

- [ ] A hotkey that toggles between the last set filter. (similar to `M` for marked flows)

### DIFF
--- a/mitmproxy/console/flowlist.py
+++ b/mitmproxy/console/flowlist.py
@@ -18,8 +18,8 @@ def _mkhelp():
         ("d", "delete flow"),
         ("D", "duplicate flow"),
         ("e", "toggle eventlog"),
+        ("f", "filter view"),
         ("F", "toggle follow flow list"),
-        ("l", "set limit filter pattern"),
         ("L", "load saved flows"),
         ("m", "toggle flow mark"),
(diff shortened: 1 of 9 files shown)


mitmproxy/console/flowlist.py
```diff
@@ -18,8 +18,8 @@ def _mkhelp():
         ("d", "delete flow"),
         ("D", "duplicate flow"),
         ("e", "toggle eventlog"),
+        ("f", "set filter view pattern"),
```
mhils: If I recall the discussion about this correctly, this should just be "filter view".

mhils:
Looks good. :+1:  I think we can keep your second bullet point out of this PR.

PR #1457, opened 2016-08-01
integer division for python 3 compatibility

mkagenius:
* Regarding color coding the response codes

The colors were not working in python 3. As 404/100 was coming out to be 4.04 instead of 4.


dufferzafar:
Nice catch! :+1:

Kriechi:
Thanks! :tada:

PR #1459, opened 2016-08-01
fixed client update event

cle1000:
if a flow is modified and sent to the server, the server doesnt not send a client broadcast to update all clients, in the PR this issue is solved

### DIFF
--- a/mitmproxy/web/master.py
+++ b/mitmproxy/web/master.py
@@ -82,6 +82,9 @@ class WebState(flow.State):
             data=entry
         )
 
+    def update_flow(self, flow):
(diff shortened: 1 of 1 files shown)


mhils:
> if a flow is modified and sent to the server, the server doesnt not send a client broadcast

How can I reproduce this? Calling `self.view._update(flow)` is a weird thing to do, the store should always update its views, so I wonder where the root cause is.

cle1000:
to reproduce this: just select a flow, modify the header and then click the save button. i never get the update_flows event from the server

mhils:
This is fixed now! ☺️

PR #1460, opened 2016-08-02
fixing queryString (python 2 and 3) and cookies value for latest mitm

mkagenius:
* Regarding har_extractor

* queryString for param `id=aflon` was coming as 

```json
"queryString": [
            {
              "name": "i",
              "value": "d"
            }
          ],
```
(due to missing iteritems()  / items() .. i guess this was never working even in python 2)

* Cookies were coming as 

```json
"request": {
          "cookies": [
            {
              "name": "__cfduid",
              "value": "d"
            },
            {
              "name": "user",
              "value": "m"
            }
          ],
```

values have only one letter due to `v[0]` .. probably because it earlier used to an array? 

This fixes those two issues. Rest of the fields look good in python 3 unless someone finds a problem.

### DIFF
--- a/examples/har_extractor.py
+++ b/examples/har_extractor.py
@@ -137,7 +137,7 @@ def response(flow):
         flow.request.timestamp_start).replace(tzinfo=pytz.timezone("UTC")).isoformat()
 
     request_query_string = [{"name": k, "value": v}
-                            for k, v in flow.request.query or {}]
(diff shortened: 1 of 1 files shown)


mkagenius:
There are issues with some other fields too. Will create a new pull request if I ever come around to fix those too.  (line 31 in harparser.py has iteritems() )

Will close this pull request. Anyone is welcome to proceed with the fixes.

Issue #649, opened 2015-06-24
crash on 32 bit systems


Nerudifus:
This problem was viewed in version 0.11.1, I think. I am currently not able to test it with the current version but the corresponding source code did not change in 0.12.1 thus I think it is still a problem.

I am using mitmdump as transparent proxy. After changing the system mitmdump was running on, it crashes after some time with
```
  File "/usr/local/lib/python2.7/dist-packages/libmproxy/protocol/http.py", line 813, in _refresh_cookie
    i["expires"] = formatdate(d)
  File "/usr/lib/python2.7/email/utils.py", line 161, in formatdate
    now = time.gmtime(timeval)
ValueError: timestamp out of range for platform time_t
```
After some time I realised that the problem occurs after the change from 64 bit to 32 bit and disappeared after another change to 64 bit again.
Looking at libmproxy/protocol/http.py:810 or in 0.12.1 at libmproxy/protocol/http.py:842
```
            d = parsedate_tz(i["expires"])
                if d:
                    d = mktime_tz(d) + delta
                    i["expires"] = formatdate(d)
```
The documentation for email.utils.formatdate mentioned for the first parameter ```Optional timeval if given is a floating point time value as accepted by time.gmtime() and time.localtime(), otherwise the current time is used.``` 
Internally time.gmtime is used. The documentation for time mentioned ```The functions in this module do not handle dates and times before the epoch or far in the future. The cut-off point in the future is determined by the C library; for Unix, it is typically in 2038.```
Under 32 bit Ubuntu python crashes with ```time.gmtime(sys.maxint+1)``` but with ```time.gmtime(sys.maxint)``` you will get 2038-01-19 and some hours. 
Thus under 32 bit this code will only work with expiration dates previous to 2038-01-19. But unfortunately there are site setting cookies with expiration dates far in the future. Different advertiser set it 30 years in future thus in 2045. But I also see dates after year 2080.

mhils:
I'm rarely convinced by bug reports from old versions, but this sounds like a very good find. I'd be super happy if someone could step up and provide a fix + test case!

bcopos:
[...]

PR #1464, opened 2016-08-02
HAR

### LINKED ISSUE #1320: HAR Export
We used to have a [har_extractor.py](https://github.com/mitmproxy/mitmproxy/blob/811b72cd304a8c75efaf706fd57cfbe9494cd3d9/examples/har_extractor.py) inline script which converted a mitmproxy dump to a har file. Unfortunately, that script depends on the `harparser` library, which is Python 2 only. It'd be great to add support for HAR export again, ideally without depending on `harparser`.

dufferzafar:
This adds a script that can dump flows as HAR but does not depend on harparser.

## Tasks:

- [ ] Tests based on comparison of entire HAR files?
- [x] Unit tests for formatting functions
- [x] `.zhar` Compression
- [x] SSL and Connect Timings
- [x] Logging via `ctx.log`
- [x] Basic Test
- [x] Proper Cookies Format
- [x] Remove harparser dependency in setup.py
- [x] <s>Append to file on every request, rather than writing at end</s> (forgot that this is json!)
- [x] <s>Read har file?</s> (Separate PR if at all)

**New Fields:**

- [x] `text` in response content
- [x] `postData` for `POST`, `PUT`, `PATCH`
- [x] `serverIPAddress`
- [x] <s>`connection`</s> (we have no unique ids for connections)
- [x] <s>`pages`</s> (mitmproxy has no concept of pages)

Fixes #1320


mkagenius:
The post Data field is missing in case of POST method (was missing in original har_extractor too). Would it possible to add that? Something like this, maybe:

```python

    if flow.request.method == "POST":
        entry["request"]["postData"] = format_post_data(flow.request)


and 

def format_params(obj):
    if obj:
        return [{"name": k, "value": unicode(v, errors='ignore')} for k, v in obj.items()]
    return ""

def format_post_data(obj):
    content_type = obj.headers.get("content-type", "")
    post_data = {"mimeType" : content_type}
    is_valid_content_type = "application/x-www-form-urlencoded" in content_type.lower()
    if is_valid_content_type:
        post_data["params"] = format_params(obj.urlencoded_form)

    post_data["text"] = unicode(obj.content, errors='ignore')
    return post_data
```

mkagenius:
One more thing, regarding cookies (in response), they are coming like (same behavior in har_extractor too):

```
  "cookies": [
            {
              "name": "__cfduid",
              "value": "SetCookie(value='d4554e2d8e2c1406a252163f6f693300a1470146048', attrs=CookieAttrs[('expires', 'Wed, 02-Aug-17 13:54:08 GMT'), ('path', '/'), ('domain', '.fallible.co'), ('HttpOnly', None)])"
            }
          ],
```

While each one should be a separate field as in the specification:

```
"cookies": [
    {
        "name": "TestCookie",
        "value": "Cookie Value",
        "path": "/",
        "domain": "www.janodvarko.cz",
        "expires": "2009-07-24T19:20:30.123+02:00",
        "httpOnly": false,
        "secure": false,
        "comment": ""
    }
]
```

mhils:
This looks great. So much clearer than what we had before :+1:

examples/har_dump.py
```diff
@@ -0,0 +1,168 @@
...
+        mitmproxy.ctx.log(pprint.pformat(HAR))
+    # TODO: .zhar compression
+    else:
+        with open(dump_file, "wb") as f:
+            f.write(json.dumps(HAR, indent=2))
+
+    # TODO: Log results via mitmproxy.ctx.log
+
+
+def format_cookies(cookies):
```
dufferzafar: How can I unit test these functions? `examples` directory has no `__init__.py` so these can't directly be imported!

examples/har_dump.py
```diff
@@ -0,0 +1,173 @@
...
+            f.write(json.dumps(HAR, indent=2))
+
+    # TODO: Log results via mitmproxy.ctx.log
+
+
+def format_datetime(dt):
+    return dt.replace(tzinfo=pytz.timezone("UTC")).isoformat()
+
+
+def format_cookies(cookies):
```
dufferzafar: How can I unit test these functions? `examples` directory has no `__init__.py` so `har_dump` can't directly be imported!

mhils: Based on the stuff in test_examples:
```python
m, sc = tscript("har_dump.py", six.moves.shlex_quote(path))
format_datetime = sc.ns["format_datetime"]
# work with format_datetime
```
Not really ideal, but I feel that's still better than making the examples directory a module.

dufferzafar:
@mkagenius Should the `postData` field be included in `PUT`, `PATCH` requests too?

@mhils This is ready for review!

mkagenius:
@dufferzafar Good question. I am also not sure as the specification is not too clear about that. But I checked what chrome does when we do `copy ALL as HAR` from inspect element window and found that in case of `PUT` / `PATCH` too it creates the post body.

dufferzafar:
@mkagenius If chrome does this, then I think we should too. It seems like the correct thing to do anyways.

netlib/http/cookies.py
```diff
@@ -263,27 +269,69 @@ def refresh_set_cookie_header(c, delta):
...
+    """
     if 'expires' in cookie_attrs:
         e = email.utils.parsedate_tz(cookie_attrs["expires"])
         if e:
-            exp_ts = email.utils.mktime_tz(e)
+            return email.utils.mktime_tz(e)
+
+    elif 'max-age' in cookie_attrs:
+        try:
+            max_age = int(cookie_attrs['Max-Age'])
```
Kriechi: Do we have a policy about cookie attribute capitalisation?
Sometimes we use only-lowercase, sometimes mixed...

test/mitmproxy/test_examples.py
```diff
@@ -98,30 +102,66 @@ def test_redirect_requests(self):
...
+        assert f['value'] == "v"
+        assert not f['httpOnly']
+        assert not f['secure']
+
+        f = format_cookies([("n", "v", CA([("httponly", None), ("secure", None)]))])[0]
+        assert f['httpOnly']
+        assert f['secure']
+
+        f = format_cookies([("n", "v", CA([("expires", "Fri, 24-Aug-2063 00:00:00 GMT")]))])[0]
+        assert f['expires'] == "2063-08-24T05:30:00+00:00"
```
Kriechi: Does this test still work on a 32-bit system, considering `time.time()` will be limited to the year 2038?

test/netlib/http/test_cookies.py
```diff
@@ -260,9 +280,29 @@ def test_is_expired():
     # or both
     assert cookies.is_expired(CA([("Expires", "Thu, 01-Jan-1970 00:00:00 GMT"), ("Max-Age", "0")]))
 
-    assert not cookies.is_expired(CA([("Expires", "Thu, 24-Aug-2063 00:00:00 GMT")]))
+    assert not cookies.is_expired(CA([("Expires", "Fri, 24-Aug-2063 00:00:00 GMT")]))
     assert not cookies.is_expired(CA([("Max-Age", "1")]))
-    assert not cookies.is_expired(CA([("Expires", "Thu, 15-Jul-2068 00:00:00 GMT"), ("Max-Age", "1")]))
+    assert not cookies.is_expired(CA([("Expires", "Sun, 15-Jul-2068 00:00:00 GMT"), ("Max-Age", "1")]))
```
Kriechi: Why did the weekdays change?

test/netlib/http/test_cookies.py
```diff
@@ -260,9 +280,29 @@ def test_is_expired():
...
+    assert len(group(c)) == 3
+
+    c = "one=uno; Path=/; foo=bar; Max-Age=0; foo=baz; expires=24-08-1993"
+    assert len(group(c)) == 3
+
+    c = "one=uno;"
+    assert len(group(c)) == 1
+
+    c = "one=uno; Path=/; Max-Age=0; Expires=24-08-1993"
+    assert len(group(c)) == 1
```
Kriechi: I think we should assert the groups explicitly - not only the length of the result.

netlib/http/cookies.py
```diff
@@ -263,27 +269,69 @@ def refresh_set_cookie_header(c, delta):
     return ret
 
 
-def is_expired(cookie_attrs):
+def get_expiration_ts(cookie_attrs):
     """
-        Determines whether a cookie has expired.
+        Determines the time when the cookie will be expired.
```
Kriechi: ... will be **expiring**.

netlib/http/cookies.py
dufferzafar: There's no policy, i guess because the [`_kconv`](https://github.com/dufferzafar/mitmproxy/blob/har/netlib/http/cookies.py#L179) function converts these keys to lowercase.

IMO, we should keep using mixed cases just to show that they don't matter.

test/mitmproxy/test_examples.py
dufferzafar: Nice catch! I don't really think this will work on 32bit machines. Should I change to something < 2038?

See also: #649

test/netlib/http/test_cookies.py
dufferzafar: Because 15th July 2068 is a Sunday not Thursday :stuck_out_tongue_winking_eye:

PR #1466, opened 2016-08-03
Simplify contentview logic

mhils:
This PR unifies the (1) "oh the contentview failed let's log an exception and just do raw mode" and (2) "let's remove the content decoding" logic we have for mitmproxy, mitmdump and (soon) mitmweb.


mhils:
@cle1000: Can you adjust for this in the contentviews PR?

PR #1474, opened 2016-08-09
Improve controller.Reply semantics

### LINKED ISSUE #1468: controller.handler race condition

1. Call `reply.send()`

##### What went wrong?

1. In a handler, we send a reply to the connection thread using `reply.send()`. The handler thread pauses.
2. The proxy thread does some stuff and asks for new info from the master.
3. In the main thread, we are still in the handler from step 1. The handler now calls a child handler (e.g. by calling super() in one of our masters).
4. `controller.handler` determines if it is the topmost handler by checking `not message.reply.handled`. However, `message.reply` has already been updated to the new reply object which has been set by the connection thread in 2, so the sub handler believes nobody is handling and starts handling the completely different event.

@cortesi: How do we solve this?

(For reference, I reproduced this by swapping the calls to `master.addons()` and `f(master,message)` in `controller.handler` while having an inline script that kills requests)

---

Mitmproxy Version: master
Operating System: Win10 x64

mhils:
This PR tries to make the semantics of controller.Reply more explicit by representing a Reply's state as a single attribute and making all unexpected state transitions raise.

- [x] Adjust rest of the codebase
- [x] Tests

Fixes #1468.

### DIFF
--- a/mitmproxy/controller.py
+++ b/mitmproxy/controller.py
@@ -222,12 +222,17 @@ def handler(f):
         # Reset the handled flag - it's common for us to feed the same object
         # through handlers repeatedly, so we don't want this to persist across
         # calls.
-        if message.reply.handled:
+        if handling:
+            if not message.reply.taken:
+                message.reply.commit()
             message.reply.handled = False
         return ret
     # Mark this function as a handler wrapper
     wrapper.__dict__["__handler"] = True
     return wrapper
+
+
+NO_REPLY = object()
 
 
 class Reply(object):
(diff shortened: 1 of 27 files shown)


mitmproxy/controller.py
```diff
@@ -239,52 +241,70 @@ class Reply(object):
...
         self.obj = obj
         self.q = queue.Queue()
-        # Has this message been acked?
-        self.acked = False
-        # Has the user taken responsibility for ack-ing?
-        self.taken = False
-        # Has a handler taken responsibility for ack-ing?
-        self.handled = False
+
+        self.state = "unhandled"  # "unhandled" -> "handled" -> "taken" -> "committed"
```
Kriechi: Should we make a setter for `state` to verify the value? 
I don't like depending on some (possible random) string.

mhils: I actually considered using Enum initially, but that requires an import of the enum class whereever we use this and things just get needlessly verbose. I changed `.state` to be readonly now, with modification only happening by the state transition functions below.

PR #1479, opened 2016-08-13
fix #1430

### LINKED ISSUE #1430: HTTP2: Cannot change host in-flight

1. Start mitmdump with the following inline script:

```python
def request(flow):
	if flow.request.host == "google.com":
		flow.request.host = "example.com"
```

2. Use a HTTP2-capable browser to open https://google.com

Requests should be redirected to example.com.

##### What went wrong?
Requests are answered by google.com, even though the UI indicates example.com.

This issue can be mitigated by passing `--no-http2`.
Implementing host change for HTTP2 is super tricky, so we may not go for this right now. 
However, the user should get a visible notification when changing hosts and the request should probably be aborted.

---

Mitmproxy Version: master
Operating System: Win10 x64

Kriechi:
fixes #1430 by raising a `SetServerNotAllowedException` and writing an entry to the eventlog.

### DIFF
--- a/mitmproxy/exceptions.py
+++ b/mitmproxy/exceptions.py
@@ -103,6 +103,10 @@ class ControlException(ProxyException):
     pass
 
 
+class SetServerNotAllowedException(ProxyException):
+    pass
+
+
 class OptionsError(Exception):
     pass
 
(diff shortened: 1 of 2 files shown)


mhils:
:+1:

PR #1480, opened 2016-08-13
Add basic auth support for reverse proxy mode

mhils:
This PR makes it possible to have basic auth in reverse proxy mode:
```
mitmdump -R http://example.com --singleuser foo:bar
```

### DIFF
--- a/mitmproxy/protocol/http.py
+++ b/mitmproxy/protocol/http.py
@@ -132,6 +132,8 @@ class HttpLayer(base.Layer):
         # We cannot rely on server_conn.tls_established,
         # see https://github.com/mitmproxy/mitmproxy/issues/925
         self.__initial_server_tls = None
+        # Requests happening after CONNECT do not need Proxy-Authorization headers.
+        self.http_authenticated = False
 
     def __call__(self):
         if self.mode == "transparent":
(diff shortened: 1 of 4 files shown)


mitmproxy/protocol/http.py
```diff
@@ -397,10 +400,17 @@ def authenticate(self, request):
...
+                        401,
+                        "Authentication Required",
+                        http.Headers(**self.config.authenticator.auth_challenge_headers())
+                    ))
+                else:
+                    self.send_response(models.make_error_response(
+                        407,
+                        "Proxy Authentication Required",
+                        http.Headers(**self.config.authenticator.auth_challenge_headers())
+                    ))
```
Kriechi: If the only difference is the message, how do you feel about extracting it out to remove the branching?
```python
message = "Authentication Required" if self.mode == "transparent" else "Proxy Authentication Required"
self.send_response(models.make_error_response(
    407,
    message,
    http.Headers(**self.config.authenticator.auth_challenge_headers())
))
```

Kriechi:
LGTM! Except the linter is not happy yet... (even after rebasing onto master)
😉

mitmproxy/protocol/http.py
mhils: Status Code differs as well. Had it inline before, didn't look better.

PR #1482, opened 2016-08-15
update docs: how mitmproxy works

### LINKED ISSUE #859: Update "How mitmproxy works"
We now peek into the traffic to identify if it's SSL, there are not hardcoded ports anymore.

Kriechi:
fixes #859
use TLS instead of SSL when revering to HTTPS encryption

### DIFF
--- a/docs/howmitmproxy.rst
+++ b/docs/howmitmproxy.rst
@@ -6,17 +6,17 @@ process works will help you deploy it creatively, and take into account its
 fundamental assumptions and how to work around them. This document explains
 mitmproxy's proxy mechanism in detail, starting with the simplest unencrypted
 explicit proxying, and working up to the most complicated interaction -
-transparent proxying of SSL-protected traffic [#ssl]_ in the presence of `Server Name Indication`_.
+transparent proxying of TLS-protected traffic [#tls]_ in the presence of `Server
+Name Indication`_.
 
 Explicit HTTP
 -------------
 
-Configuring the client to use mitmproxy as an explicit proxy is the simplest
-and most reliable way to intercept traffic. The proxy protocol is codified in the
-`HTTP RFC`_, so the behaviour of both
-the client and the server is well defined, and usually reliable. In the
-simplest possible interaction with mitmproxy, a client connects directly to the
-proxy, and makes a request that looks like this:
+Configuring the client to use mitmproxy as an explicit proxy is the simplest and
+most reliable way to intercept traffic. The proxy protocol is codified in the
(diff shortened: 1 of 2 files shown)


mhils:
Replacing SSL with TLS is fine with me, but we should consider that the use of "SSL" is still prevalent and people are searching for it. So maybe make it "SSL/TLS" in the introduction?
I doubtful that we should list "Full support for HTTP/1.0, HTTP/1.1, and HTTP/2." as the first bullet point for the features in the introduction. That section focuses on stuff people can do with mitmproxy. In this specific context, I see the HTTP versions supported by mitmproxy more as an implementation detail. This certainly becomes more interesting once we add WebSockets et. al, but even then I would probably start with the simple "Intercept HTTP requests and responses and modify them on the fly." as the main message.

Kriechi:
I updated the docs.

However, I'm feeling strongly about sticking to "TLS" only. 
The WebSockets RFC from 2011 doesn't mention "SSL" even once, but "TLS" 27 times.
The new HTTP/1.1 RFCs from 2014 don't mention "SSL" even once, but "TLS" 14 times.
The HTTP/2 RFCs from 2015 don't mention "SSL" even once, but "TLS" 369 times.

mhils:
Fair point. I agree on the more detailed docs pages, but I'm indifferent on the slogan/title in introduction.rst. The problem here is that the set of users which understand the term "TLS" is more or less a strict subset of the users which understand what SSL means, and our main concern should be making it simple to understand what mitmproxy does.

I'd be fine with both variants, but given that a project's slogan is a pretty central part of the project, let's hear @cortesi before we merge this.

cortesi:
So, this is an argument that I've had myself when writing about these topics. When I picked SSL, my thought process was as follows:

- We support both SSL and TLS. Using TLS/SSL everywhere is ugly, so we need one term to do the work of both.
- SSL is sometimes used generically to refer to both protocols generically, but TLS never is.

None the less, technically correct is the best kind of correct and real SSL is getting rarer, so I'm going to merge this as-is and we can work on it down the track. 

Regarding the slogan, we really need something better here. We support way more than http and https and listing all our protocols in the slogan is not attractive. An alternative would be to make the slogan simple:

>    mitmproxy: a man-in-the-middle proxy

And then list our protocol support prominently on the webpage, readme, etc:

>  Supports HTTP, HTTP2, Websockets and generic TCP connections. Intercepts traffic encrypted 
>  with SSLv3, TLS 1.0, TLS 1.1, and TLS 1.2.

PR #1488, opened 2016-08-18
add WebSockets support

Kriechi:
Work for future PRs:
  - frame continuation: abstract frames to messages and provide them to the user
    * respect `--body_size_limit SIZE` and `--stream SIZE`
  - implement compression extensions
  - add hooks for inline scripts


mitmproxy/protocol/http.py
```diff
@@ -170,13 +172,33 @@ def __call__(self):
...
                 flow = models.HTTPFlow(self.client_conn, self.server_conn, live=self)
                 flow.request = request
+
                 # set upstream auth
                 if self.mode == "upstream" and self.config.upstream_auth is not None:
                     flow.request.headers["Proxy-Authorization"] = self.config.upstream_auth
                 self.process_request_hook(flow)
 
+                # WebSockets
+                if websockets.check_handshake(request.headers):
```
mhils: I think this logic should sit at https://github.com/mitmproxy/mitmproxy/blob/cacee3871c6a9f0be7127f3c790e09a1daaf8490/mitmproxy/protocol/http.py#L193 - we don't know yet if the remote actually speaks websocket. Having it down there also enables us to fall back to TCP mode for versions != 13.

mitmproxy/protocol/http2.py
```diff
@@ -96,15 +96,16 @@ def __init__(self, ctx, mode):
         self.server_to_client_stream_ids = dict([(0, 0)])
         self.client_conn.h2 = SafeH2Connection(self.client_conn, client_side=False, header_encoding=False)
 
-        # make sure that we only pass actual SSL.Connection objects in here,
-        # because otherwise ssl_read_select fails!
-        self.active_conns = [self.client_conn.connection]
-
```
mhils: Everything in protocol/http2.py could/should probably be a separate PR?
On a first sight, this looks like it breaks replay with an unreachable host. Ok, maybe that was even broken before, but whatsoever there should be a test for this.

mitmproxy/protocol/websockets.py
```diff
@@ -0,0 +1,140 @@
...
+        self.log(
+            "WebSockets Frame received from {}".format("server" if is_server else "client"),
+            "debug",
+            [repr(frame)]
+        )
+
+        if frame.header.opcode & 0x8 == 0:
+            # forward the data frame to the other side
+            other_conn.send(bytes(frame))
+            # TODO: make some hooks for inline scripts to modify data
```
mhils: This should actually be very similar to the TCPFlow/TCPMessage implementation. I have something similar for SSH if you are interested.

netlib/websockets/frame.py
```diff
@@ -128,17 +141,16 @@ def __bytes__(self):
             # '!Q' = pack as 64 bit unsigned long long
             # add 8 bytes extended payload length
             b += struct.pack('!Q', self.payload_length)
+        else:
+            raise ValueError("Payload length exceeds 64bit integer")
```
mhils: Should probably be RuntimeError (or something better), because __bytes__ does not really accept a parameter/value?

netlib/websockets/utils.py
```diff
@@ -0,0 +1,100 @@
+"""
+Collectiton of utility functions that implement small portions of the RFC6455
```
mhils: Just to proof that I'm actually reading this: typo :stuck_out_tongue:

netlib/websockets/utils.py
```diff
@@ -0,0 +1,100 @@
...
+WebSockets Protocol Useful for building WebSocket clients and servers.
+
+Emphassis is on readabilty, simplicity and modularity, not performance or
+completeness
+
+This is a work in progress and does not yet contain all the utilites need to
+create fully complient client/servers #
+Spec: https://tools.ietf.org/html/rfc6455
+
+The magic sha that websocket servers must know to prove they understand
```
mhils: This went away during reformat?

mhils:
This looks fantastic overall :+1:

netlib/websockets/frame.py
Kriechi: No, not `bytes`, but the only argument is invalid. I wasn't sure what to use - RuntimeError seemed too generic.

> exception ValueError
Raised when a built-in operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError.

vs. 
> exception RuntimeError
Raised when an error is detected that doesn’t fall in any of the other categories. The associated value is a string indicating what precisely went wrong.

mitmproxy/protocol/http2.py
Kriechi: Ehm - there shouldn't be any real changes, just cosmetics to make use of the same "layout" as WebSockets and RawTCP. That's why I included it in this PR, because it is only style-related to make it look the same as websockets.

mhils: `__call__` started with `if self.server_conn`, which is now missing. Truth to be told, we never hit different code paths as long as `protocol.tls` always connects upstream because of ALPN (ugh). Whatsoever, that is/was the reason why we have that distinction into `active_conns`. So, the whole story is pretty bad and we probably need to do something about it at some point. I'm a bit in a disagreement with the changes here, so maybe let's leave that out so that we don't get sidetracked?

mhils: After discussing this on Slack: We won't support this soon anyway due to the ALPN requirements (which would require a major rewrite), so the changes here are fine for now.

netlib/websockets/frame.py
mhils: Scratch my previous comment here. `bytes(invalid_frameheader)` should (as you said) raise a ValueError. Did not consider that it was called that way. 
My reasoning was that `foo.bar()` should never raise one, as we don't have any arguments here. This does not apply here, as the object itself is an argument to the `bytes()` builtin.

mitmproxy/protocol/websockets.py
Kriechi: Yes - I already took a look.
However, emitting raw frames is mostly useless for the user since data transfer might be chunked (CONTINUATION) or compressed (gzip/deflate/...)
So this work should probably be done in a separate PR next.

mitmproxy/protocol/http.py
Kriechi: What if I want to intercept and modify the websockets handshake?
Moving it down basically only gives me a completed handshake - without the possibility to do anything afterwards?

mitmproxy/protocol/websockets.py
mhils: The same is basically true for TCP. Separate PR sounds fine, let's just do _something_ (I think of just logging) here for now.

mitmproxy/protocol/http.py
```diff
@@ -198,7 +205,7 @@ def __call__(self):
                 # It may be useful to pass additional args (such as the upgrade header)
                 # to next_layer in the future
                 if flow.response.status_code == 101:
-                    layer = self.ctx.next_layer(self)
+                    layer = self.ctx.next_layer(self, flow)
```
mhils: It feels like hardcoding WebSockets here would just be easier?

mhils:
Other than that, this is good to merge. Sorry for getting back so late. :)

mitmproxy/protocol/http.py
Kriechi: Maybe - the code was already there - I figured the person who placed it here had a reason... 😄 
However, I think we should keep it like that.
We can use this mechanism to add h2c support (HTTP/2 over plain TCP without TLS) in the future.

PR #1489, opened 2016-08-18
Web refactor

cle1000:
refactoring of my code

### DIFF
--- a/web/src/css/dropdown.less
+++ b/web/src/css/dropdown.less
@@ -1,0 +1,4 @@
+hr .divider {
+    margin-top: 5px;
+    margin-bottom: 5px;
+}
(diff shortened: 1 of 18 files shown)


web/src/css/dropdown.less
```diff
@@ -0,0 +1,4 @@
+hr .divider {
```
mhils: This probably does not work unless it's `hr.divider` ?

web/src/js/ducks/flows.js
```diff
@@ -210,5 +210,8 @@ export function updateFlow(item) {
  * @private
  */
 export function removeFlow(id) {
-    return { type: REMOVE, id }
+    return (dispatch) => {
+        dispatch(select())
```
mhils: I'd prefer if removing a flow would select the next one, so that I can remove a bunch of flows by pressing `d` repeatedly.

mhils:
This is a really nice cleanup, great job :+1: . Let's fix the two minor things mentioned above and we are good to go. :smiley:

PR #1493, opened 2016-08-20
Truncate URLs in console

### LINKED ISSUE #1485: View only a part of the URL
I'm currently dealing with an app that has very looooong urls (see screenshots.)

This makes the UI look crappy. I understand that there's not much we can do in cases in like these.

But having an option to only view part of the URL, such as everything but the query string, or only the first 100 characters etc. in the `flowlist`, and then showing the entire URL when user presses a key (in the `flowview`) would be nice.

flow list

flow view

dufferzafar:
An attempt to fix #1485

Currently, URLs are shortened in the `flowlist`, but are displayed as-is in the `flowview`. 

I'd actually prefer to shorten them in both the views, but then we need to come up with a way to view the full URL.

It could be via an option in the UI like we have `Show Host` or a hotkey that toggles the behaviour in both the views.

### DIFF
--- a/mitmproxy/console/common.py
+++ b/mitmproxy/console/common.py
@@ -32,6 +32,8 @@ METHOD_OPTIONS = [
     ("edit raw", "e"),
 ]
 
+MAX_URL_LEN = 200
+
 
 def is_keypress(k):
     """
@@ -328,7 +330,7 @@ def export_to_clip_or_file(key, scope, flow, writer):
 flowcache = utils.LRUCache(800)
 
 
-def raw_format_flow(f, focus, extended):
+def raw_format_flow(f, focus, extended, short_urls):
     f = dict(f)
     pile = []
     req = []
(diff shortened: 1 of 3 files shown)


mitmproxy/console/common.py
```diff
@@ -359,6 +359,11 @@ def raw_format_flow(f, focus, extended):
         uc = "title"
 
     url = f["req_url"]
+
+    # TODO: Add a configuration setting that disables this behaviour?
```
mhils: :-1: on this, I can hardly imagine a use case for this. No strong opinion though.

mitmproxy/console/common.py
```diff
@@ -359,6 +359,11 @@ def raw_format_flow(f, focus, extended):
...
+    if truncate_urls and len(url) > truncate_urls:
+        url = url[:truncate_urls] + "..."
```
mhils: Let's use http://www.fileformat.info/info/unicode/char/2026/index.htm ? :smiley:

mitmproxy/console/common.py
```diff
@@ -451,5 +456,8 @@ def format_flow(f, focus, extended=False, hostheader=False):
             d["resp_ctype"] = ""
     return flowcache.get(
         raw_format_flow,
-        tuple(sorted(d.items())), focus, extended
+        tuple(sorted(d.items())),
+        focus,
+        extended,
+        truncate_urls,
```
mhils: I think these should be part of `d` ?

dufferzafar:
This is ready for review & merge.

PR #1494, opened 2016-08-20
[web] Add context menu for flow row

gzzhanghao:
ContextMenuLayer always creates an extra layer when rendering [react-contextmenu/contextmenu-layer.js#L94](https://github.com/vkbansal/react-contextmenu/blob/master/src/contextmenu-layer.js#L94), so we need to put it into FlowColumn instead of FlowRow.

### DIFF
--- a/web/src/css/app.less
+++ b/web/src/css/app.less
@@ -19,3 +19,4 @@ html {
 @import (less) "footer.less";
 @import (less) "codemirror.less";
 @import (less) "contentview.less";
+@import (less) "contextmenu.less";
--- a/web/src/css/contextmenu.less
+++ b/web/src/css/contextmenu.less
@@ -1,0 +1,26 @@
+.react-context-menu {
+    outline: none;
+    background: white;
+    border: 1px solid #a0a0a0;
+    min-width: 120px;
+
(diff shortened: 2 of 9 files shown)


mhils:
Hm - this clutters things up a bit. Can't we just have a single manual HTML5 `<menu type="context">` for the whole table? As I see it, the only tricky part might be letting the menu know which flow we are on, but this could be solved by listening to https://developer.mozilla.org/de/docs/Web/Events/contextmenu ?

gzzhanghao:
AFAIK, `<menu type="context">` is behind a flag in latest version of chrome and lack of compatibility in most of other browsers. So we still need `react-contextmenu` here.
Maybe we can have a single `ContextMenuWrapper` instead of `ContextMenu`?

mhils:
Looks ok so far. Three issues:

1. Right-clicking a flow does not work if another context menu is present.
2. We should not show things like "Accept" if the flow is accepted.
3. The context menu looks broken to me:

image

gzzhanghao:
1st. It seems to be a feature of `react-contextmenu`, I wonder if there are any method to gracefully hide the backdrop layer :(
Other than that, fixed :tada:

PR #1496, opened 2016-08-21
[web] multiple selection with ctrl key

gzzhanghao:
Multiple selection using shift key is far more complicated than the one which use ctrl key, so let's start with the easy one :)

### DIFF
--- a/web/src/js/ducks/flows.js
+++ b/web/src/js/ducks/flows.js
@@ -53,7 +53,7 @@ export default function reduce(state = defaultState, action) {
         case SELECT:
             return {
                 ...state,
-                selected: action.flowIds
+                selected: action.selectedFlowIds,
             }
 
         default:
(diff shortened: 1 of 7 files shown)


mhils:
Why don't we use the click event's .ctrlKey/.shiftKey attributes? :smiley:

web/src/js/ducks/flows.js
```diff
@@ -144,15 +144,21 @@ export function upload(file) {
...
 
-
-export function select(id) {
-    return {
-        type: SELECT,
-        flowIds: id ? [id] : []
+/**
+ * @public
+ */
+export function select(id, ctrlkey, selectedFlowIds) {
```
mhils: I think we could make a separate "ctrlClick" action creator.

web/src/js/ducks/flows.js
```diff
@@ -144,15 +144,21 @@ export function upload(file) {
...
 
-
-export function select(id) {
-    return {
-        type: SELECT,
-        flowIds: id ? [id] : []
+/**
+ * @public
+ */
+export function select(id, selectedFlowIds, ctrl) {
```
mhils: Can we split this into `select` and `ctrlSelect`/ `shiftSelect` so that other parts can still call `select` without suppling all the other arguments?
(There's also a merge conflict from @cle1000's work I think)

web/src/js/ducks/ui/flow.js
```diff
@@ -23,10 +23,10 @@ const defaultState = {
     tab: 'request',
     content: [],
     maxContentLines: 80,
+    ctrlkey: false,
```
mhils: Can we remove the whole ctrlkey state? I don't think we need to keep track of this in redux.

web/src/js/ducks/flows.js
gzzhanghao: @mhils `selectedFlowIds` and `ctrl` are optional here, you can ignore them if you don't want to select multiple flows ;)

cortesi:
Closing due to inactivity. Please do feel free to rebase to master and re-open!

PR #1506, opened 2016-08-24
Stream "requests" (upload) in the same way as "responses" (download)

fviard:
...when stream mode is enabled.

When using `--stream` option with mitmdump, only "download" (downlink/responses) were "streamed" in chunks. The goal was to have a constant memory usage irrespective of the size of the response.

"upload" (uplink/requests) were not streamed, and so, let's say that you are "uploading" a 1GB file, mitmproxy/dump would temporarily use 1GB RAM.

This patch adds the stream option for "upload" in the same way that it is done for "download".

Note: I first created the patch against the last release and tested it. Then i rebased it. I was not able to test it after the rebase on master as I have some issues running the git master version on my setup.

The "http2" part of the code is completely untested, and just based on code logic.

### DIFF
--- a/mitmproxy/controller.py
+++ b/mitmproxy/controller.py
@@ -25,6 +25,7 @@ Events = frozenset([
     "tcp_close",
 
     "request",
+    "requestheaders",
     "response",
     "responseheaders",
 
--- a/mitmproxy/flow/master.py
+++ b/mitmproxy/flow/master.py
@@ -318,6 +318,16 @@ class FlowMaster(controller.Master):
         return f
 
     @controller.handler
+    def requestheaders(self, f):
+        try:
+            if self.stream_large_bodies:
+                self.stream_large_bodies.run(f, False)
+        except netlib.exceptions.HttpException:
+            f.reply.kill()
+            return
+        return f
+
+    @controller.handler
     def responseheaders(self, f):
         try:
             if self.stream_large_bodies:
--- a/mitmproxy/models/http.py
+++ b/mitmproxy/models/http.py
@@ -70,6 +70,7 @@ class HTTPRequest(MessageMixin, Request):
 
         # Is this request replayed?
         self.is_replay = is_replay
+        self.stream = False
 
     def get_state(self):
         state = super(HTTPRequest, self).get_state()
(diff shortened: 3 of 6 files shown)


Kriechi:
This is a really great feature contribution - I had that on my list for a long time now 👍 

We will try to review this in the next days (and weeks).
However, since it is summer time, this might take a bit 😄 

What are your issues with making the tests work again? 

I will try to look at the http2-related parts. 
Maybe we can work together on writing tests and making everything work!

Thanks! 🎉

fviard:
Thank you for your reply.
No worry for the delays, I took myself a quit long time to push this PR.
I just understood what was preventing me to run the last master of mitmdump.
I use it with a "script" and my script was like that:
```
def start(context, argv):
    context.log("start")
def done(context):
    context.log("end")
```
It looks like that context and args were removed of "start" without notice by the following commit:
https://github.com/mitmproxy/mitmproxy/commit/7c67faa8da39f428d1860bccae806137943b66a6

It is a little bad move in my opinion, is it not replaced by anything?

dufferzafar:
@fviard Yes, removal of the `context` parameter is breaking change but I think the notice will be provided upon a new release.

To use the context:

```python
import mitmproxy

def start():
    mitmproxy.ctx.log("start")
```

Have a look at the `examples/stub.py` to see what the new signature of various functions is!

fviard:
Thanks, that did it to be able to run mitmdump correctly.

Kriechi:
I would love to see tests for the http1 parts. Then I can gladly help you on the http2 portions and getting it tested properly!
Do you need any more help rebasing your changes to the latest git-master code?

mhils:
What's the current status here? 😃 
@fviard, are you still working on this? ☺️

cortesi:
Please see #1624 for some further developments here. If you return to this patch, it would be great if this could also address the design wart where the .stream attribute has no clearly defined type - it can be True, False or an object depending on which phase we're in, which is a bit horrible.

fviard:
@mhils Sadly I was busy and not yet had the time to do the http1 tests.
@cortesi Your changes look nice. I will try to manage to find the time to update the PR with your changes.

cortesi:
@fviard Great! :)

Kriechi:
Sorry, but there it seems bigger changes in mitmproxy happened over the last weeks, so I'll close this PR for now, until we have decided how to integrate this feature into the new architecture.

Thanks @fviard! We will try to get something new going soon!
Your work is much appreciated! 🎉

Issue #1469, opened 2016-08-04
Replace body only


dufferzafar:
Currently, the replacements work on both Headers and Body.

I think we should support body only replacements.

---

User story: https://discourse.mitmproxy.org/t/emulate-charles-proxys-map-local/116

dufferzafar:
See also: #1471

Prinzhorn:
Fixed via #4046 ? I think this can be closed.

PR #1509, opened 2016-08-26
Fixes #1471 - Change .* to match entire body

### LINKED ISSUE #1471: Make `.*` match the whole body for replacements.
See https://discourse.mitmproxy.org/t/emulate-charles-proxys-map-local/116 for context - specifying `.*` as a replacement pattern currently replaces every single line in the message body. It feels more natural if that would replace the whole body altogether. We can accomplish this behaviour by passing `re.DOTALL` as a flag for `FBod/FBodRequest/FBodResponse` in `mitmproxy/filt.py`.

This issue should be pretty straightforward to implement, so I'll leave this for newcomers who are interested in contributing to mitmproxy. Feel free to ask here or on Slack if you have any further questions! :relaxed:

arjun23496:


dufferzafar:
@Kriechi: What we discussed was more related to #1469 I think.

This change is alright, but it makes all replacements to be with `DOTALL`. In #1471 @mhils says that the fix is to be made in `mitmproxy/filt.py`, but I'm not sure how that'll work.

arjun23496:
@dufferzafar 
 

> but it makes all replacements to be with `DOTALL`

isnt this what we are trying to achieve?

dufferzafar:
I think we wanted "body" related replacements to be with `DOTALL`, and the rest stay as-is. But as I said, I'm not too sure about that.

arjun23496:
In that case it will be more useful if the user has the option to choose which to replace.. header or body.....

arjun23496:
Hey I am new here... any suggestions on how to proceed further ?

Kriechi:
Thanks for your input - we really want to fix this once and for all!
Let's see what our core-dev guys have to say @mhils @cortesi.

(They are probably somewhere on the beach enjoying the hot summer 😎 )

dufferzafar:
If you want to have a look at other issues to fix, you could [start here](https://github.com/mitmproxy/mitmproxy/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+contribution%22)

Or [join us at Slack](http://slack.mitmproxy.org/), for random "discussions".

cortesi:
Hi @arjun23496 - welcome to the mitmproxy project. A few comments here:

- We should use the same flavor of regex across all portions of the requests and responses just to be consistent. Because newlines don't really occur elsewhere than the body, switching to re.DOTALL won't functionally affect most usecases other than body replacements. 
- The match function that we use needs to be symmetric with the replacement function. So if we make regex replacement have re.DOTALL by default, the match patterns in filt.py must as well.  This is what @mhils was mentioning in #1471.

Arjun, if you could make the matching changes in fillt.py, I can merge.

arjun23496:
Sorry for the delay, was busy for some time, hope the new commit completes the PR.

mhils:
Thanks for making the adjustments! 😄 

Sorry for getting back so slowly as well, we've all been on vacation or extremely busy. ☺️

PR #1510, opened 2016-08-26
fixed #1501 : Divide by zero error came when string was empty, also a test

### LINKED ISSUE #1501: Divide by Zero error
Its in `netlib/strutils.py`

This line # around 126 :
```python
    for i in six.iterbytes(s[:100])
) / len(s[:100]) > 0.3
```

if s is empty, it gives this error in the mitmproxy, (doesn't crash though due to recent improvements in mitmproxy i guess..)

mkagenius:
Fixed https://github.com/mitmproxy/mitmproxy/issues/1501

### DIFF
--- a/netlib/strutils.py
+++ b/netlib/strutils.py
@@ -121,6 +121,9 @@ def escaped_str_to_bytes(data):
 
 def is_mostly_bin(s):
     # type: (bytes) -> bool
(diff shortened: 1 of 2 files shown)


Kriechi:
Thanks for fixing this!
Would you mind fixing the linting errors? You can run `tox -e lint`.

PR #1511, opened 2016-08-26
Fixes #1495 - Added count argument for replacing contents in body

### LINKED ISSUE #1495: Add max replacements/count argument to flow.replace

```python
 regex1 = r'\d+\#\d+\#\d+\#\d+\#(\d+)\#\d+\#\d+\#\d+\#\d+\#'
 re.compile(regex1)
 flow.response.replace(regex1, r'444#5555#66#77#\1#0#0#0#0#', 10)
```

notepad _2016-08-20_11-17-57

`flow.response.replace(regex1, r'444#5555#66#77#\1#0#0#0#0#', 10)`

The **10** used should be Max Count or Max Replace

##### What went wrong?

In  `flow.response.replace(old, new[, maxreplace])`

is no longer working, all my older scripts are now broken; this used to work fine.

I have tried changing my replace to a variable, changing the long regex and replace with something very small and simple.

Work Great on my other servers that run MITMProxy 0.15 and i think 0.16 was working too.

*When i updated to 0.17 or 0.18 it then no longer works and my scripts are broken.

---

Mitmproxy Version:  0.17 and 0.18
Operating System:  Ubuntu 14.04 and 16.04

arjun23496:


dufferzafar:
@arjun23496 There are some more replaces in the codebase. Can you please make similar changes for them too? `netlib/http/headers.py:161`, `netlib/http/request.py:71`

arjun23496:
@dufferzafar Thanks for the reply, I am having doubt about how the actual implementation of count happens for header. There are 2 possible ways, one is that we can replace by count each value of header key value pairs or we can replace count times for the whole header, which one do you think is more intuitive ?

arjun23496:
I have made the necessary changes for this PR.. Is anything more required ?

Kriechi:
This looks very good already! Thanks! 👍 
Would you mind adding a test for the `count` attribute, so we can verify everything works properly?

Kriechi:
@arjun23496 thanks for adding some tests!
Unfortunately some of the existing automated tests are failing. 
Maybe you could look into that and try to fix the bugs? 
Please pay attention to our linter as well - it is very strict! 😜

Kriechi:
Thanks! Great work getting this PR in shape! 🎉

Issue #302, opened 2014-07-06
Add threading.Event() to indicate when mitmproxy is ready to receive traffic


mhils:
https://groups.google.com/forum/#!topic/mitmproxy/4YynkCSFUoM
> I am writing an app using libmproxy. The example in the libmproxy document shows functions for run, handle_request, and handle_response. I'm wondering if a similar function is available that would get called when the proxy is running and ready to receive traffic. I would like to have a function called at that time so it could start a browser that would drive the traffic through the proxy. Anyone know if libmproxy has something like this? Thanks.
> I have tried starting the browser process in the run callback, just before calling controller.Master.run(self). It works in my limited testing so far. However, I'm not sure if this will always work or if I am just getting lucky with the timing (i.e., the browser process takes a while to get going, and that gives the call to controller.Master.run(self) time to get the proxy ready to handle traffic).

Rather than relying on race conditions here, we should have a threading.Event() (like should_exit) that signals that mitmproxy has started.

albertzaharovits:
IMHO starting the proxied app after the controller.Master instantion is no race.
controller.Master.init receives a netlib.tcp.TCPServer instance listening on the TCP socket. controller.Master.run will start the Slave thread which will accept any pending connection requests on the socket.
Am I missing something here?

PS I assume one is using libmproxy as in examples/flowbasic

cortesi:
We now have the ``start()`` event, which serves this purpose. It's called on instantiation of the master on all addons. This is after the server has been bound to the socket, so I don't believe there's a race there.

PR #1522, opened 2016-09-03
Notify when proxy server started

### LINKED ISSUE #1505: mitmdump should display "ready" message after start.

2. `mitmdump`

mitmdump should print some notification that it is running and ready to receive traffic.

##### What went wrong?

mitmdump just displays nothing, it looks like it hangs.

Ideally we'd only show that if mitmdump is invoked in a tty?

---

Mitmproxy Version: 0.18
Operating System: Win10 x64

chhsiao90:
Resolved #1505 

I think use the channel for communication between server and master is a good practice.
But I need your advise for
1. Is the notification message is OK?
2. Is override the serve_forever from TCPServer ok? Or we have some better way to do this.

Thanks!

### DIFF
--- a/mitmproxy/dump.py
+++ b/mitmproxy/dump.py
@@ -48,6 +48,10 @@ class DumpMaster(flow.FlowMaster):
         self.options = self.options  # type: Options
         self.set_stream_large_bodies(options.stream_large_bodies)
 
+        print("Proxy server listening at http://%s:%d" % (
+            (options.listen_host or "0.0.0.0"),
+            options.listen_port))
+
         if self.server and self.options.http2 and not tcp.HAS_ALPN:  # pragma: no cover
             print("ALPN support missing (OpenSSL 1.0.2+ required)!\n"
                   "HTTP/2 is disabled. Use --no-http2 to silence this warning.",


mhils:
Thanks for the PR!

This seems to be a bit more complicated than it needs to be. Can we just add a simple print statement [here](https://github.com/mitmproxy/mitmproxy/blob/1e5a5b03f8d56df62a04a368bd5eb2d59cb7582a/mitmproxy/dump.py#L51) that states `"Proxy server listening at http://$options.listen_host:$options.listen_port"`? :-)

chhsiao90:
No problem and done!
Currently, the mitmdump message show as
```
>mitmdump -b 192.168.56.1
Proxy server listening at http://192.168.56.1:8080

>mitmdump
Proxy server listening at http://:8080
```
Is that empty host acceptable or not? Or we should give it a default value?

mhils:
Can we just make it `(self.config.options.listen_host or "0.0.0.0")`?
Also, we should move that to the DumpMaster constructor to keep things simpler. When that code is called we have already initialized the socket, so we are already listening on the proxy port.

chhsiao90:
Done! Please take a look.
Sorry I misunderstood that you mean use print instead of notify.
And I didn't notice that it need to be moved to DumpMaster.

That was make sense! thanks

mhils:
This looks great now, thanks!

Kriechi:
What about https://github.com/mitmproxy/mitmproxy/issues/302?

PR #1523, opened 2016-09-03
improve error handling in http2

Kriechi:


Kriechi:
This PR should also fix our CI problems with failing tests at the moment.
It seems like cryptography doesn't like it when threads (and sockets) are still open at the end of the test suite. It will segfault and therefore fail the test run.

PR #1531, opened 2016-09-06
Fix decoding when content-encoding: none

Schamper:

```
In [7]: encoding.decode('', 'none')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-7-a83dfad39aec> in <module>()
----> 1 encoding.decode('', 'none')

/home/ames/dev/mitmproxy/netlib/encoding.pyc in decode(encoded, encoding, errors)
     54             type(e).__name__,
     55             repr(encoded)[:10],
---> 56             repr(encoding),
     57         ))
     58 

ValueError: LookupError when decoding '' with 'none'
```

Return the same value.

##### What went wrong?

I encountered a website that sent a content-encoding header of 'none', which `netlib.encoding` doesn't seem to be able to handle. I realize that it's an out of spec header value, but I believe it should be handled gracefully.

---

Mitmproxy Version: master (0.18)

### DIFF
--- a/netlib/encoding.py
+++ b/netlib/encoding.py
@@ -76,7 +76,7 @@ def encode(decoded, encoding, errors='strict'):
     Raises:
         ValueError, if encoding fails.
     """
-    if len(decoded) == 0:
+    if len(decoded) == 0 or encoding == "none":
         return decoded
 
     global _cache
--- a/netlib/encoding.py
+++ b/netlib/encoding.py
@@ -34,7 +34,7 @@ def decode(encoded, encoding, errors='strict'):
     Raises:
         ValueError, if decoding fails.
     """
-    if len(encoded) == 0:
+    if len(encoded) == 0 or encoding == "none":
         return encoded
 
     global _cache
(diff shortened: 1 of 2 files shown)


Kriechi:
Thanks - this looks reasonable to me.
Would you mind adding a test for this as well?

Kriechi:
LGTM! 👍 

If you encountered a website with such a faulty content-encoding header, is the value `'none'` or `b'none'`?

Schamper:
Not sure how I'm supposed to tell, I'm visiting a bunch of websites in automated fashion and currently don't have a way to trace back an exception to a certain URL (but I will fix that right now, so debugging will be easier). The way I discovered this is that I got the following traceback:
```
[...]
  File "~/mitmproxy/netlib/http/message.py", line 115, in get_content
    return encoding.decode(self.raw_content, ce)
  File "~/mitmproxy/netlib/encoding.py", line 56, in decode
    repr(encoding),
ValueError: LookupError when decoding 'GIF89a\x0 with 'none'
```

I'm currently also getting a bunch of decoding errors related to utf-8, but haven't investigated yet. When I can find the issue, I'll submit a different PR.

mhils:
Thanks for the PR. Can we just make none synonymous to identity (see custom_encode/decode in encoding.py) instead of adding a separate if clause? It's OK to just re-use the identity functions.

mhils:
Fantastic, thanks!

PR #1532, opened 2016-09-06
Playback and fix construct breakage

cortesi:
Move playback into an addon, and fix a termination bug that's on the 0.18 hit-list.


Kriechi:
The construct breakage was resolved upstream - so your commit is not needed anymore.
(you might need to clear your cache etc)

Testing only your "server playback" commit gives me one test failure on python3 (reproducibly on travis and my local osx):
https://travis-ci.org/mitmproxy/mitmproxy/jobs/157968327#L561

cortesi:
I've removed the construct fix. Let's see about the weird Python3 error.

PR #1534, opened 2016-09-06
Add API to programmatically create new requests

### LINKED ISSUE #1473: Add API to programatically create new requests.
Original discussion: https://discourse.mitmproxy.org/t/are-there-any-script-examples-for-creating-a-get-post-put/134

It would be great if we'd have a dedicated API to make new requests from scripts, e.g. `master.new_request()` or something along those lines. This would probably just call `new_request` and then `replay_request`, but this seems useful enough to spell out explicitly.

tylerstonge:
Resolves #1473


mhils:
Thanks for the PR!

Can we get a test for this? :smiley:

tylerstonge:
Sorry about leaving that off. I'm still getting my bearing around the code, but hopefully this test is sufficient.

suhanimirza:
Good

PR #1554, opened 2016-09-14
Added Content-Length header to examples/redirect_requests.py

alexeypetrenko:
Current example returns no *Content-Length* header, which causes some warnings. For example:

<pre>
$ curl -x http://localhost:8080  -v example.com
* Rebuilt URL to: example.com/
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET HTTP://example.com/ HTTP/1.1
> Host: example.com
> User-Agent: curl/7.43.0
> Accept: */*
> Proxy-Connection: Keep-Alive
> 
< HTTP/1.1 200 OK
< Content-Type: text/html
<b>* no chunk, no close, no size. Assume close to signal end</b>
< 
* Closing connection 0
Hello World
</pre>

This patch adds missing header.

### DIFF
--- a/examples/redirect_requests.py
+++ b/examples/redirect_requests.py
@@ -12,6 +12,7 @@ def request(flow):
     # Method 1: Answer with a locally generated response
     if flow.request.pretty_host.endswith("example.com"):
         resp = HTTPResponse.make(200, b"Hello World", {"Content-Type": "text/html"})
+        resp.headers["Content-Length"] = str(len(resp.body))
         flow.reply.send(resp)
 
     # Method 2: Redirect the request to a different server


mhils:
Thanks for the PR, but
https://github.com/mitmproxy/mitmproxy/blob/master/netlib/http/response.py#L87
should already set a content-length header. Are you testing on 0.17 or on
master?

alexeypetrenko:
Yeah, I've noticed this comment while trying to find anything in a source code that is supposed to set aforementioned header. Unfortunately, it doesn't work for me.

Testing have been done with corresponding `redirect_request.py` example scripts on both `master` and `0.17` branches. 

I've been using *dockerhub* `mitmproxy/mitmproxy:0.17` and `mitmproxy/mitmproxy:dev` images. They seem to be up to date:

```
$ sudo docker images 
REPOSITORY                        TAG                 IMAGE ID            CREATED             SIZE
docker.io/mitmproxy/mitmproxy     dev                 1f3d8a86d359        3 days ago          112.7 MB
docker.io/mitmproxy/mitmproxy     0.17                b8d85f6c84a6        3 months ago        87.47 MB
```

```
$ sudo docker run -it mitmproxy/mitmproxy:dev mitmproxy --version
mitmproxy 0.18
```

```
$ sudo docker run -it mitmproxy/mitmproxy:0.17 mitmproxy --version
mitmproxy 0.17
```

mhils:
Thanks for the quick follow up! I just tried to reproduce this locally and found the culprit: We overwrite the headers after setting the content in [Response.make()](https://github.com/mitmproxy/mitmproxy/blob/f17c0fdac636f7269f4885294e2a8d2c52c23590/netlib/http/response.py#L87-L105).

Could you update your PR to fix this and add a test in [test/netlib/http/test_response.py](https://github.com/mitmproxy/mitmproxy/blob/f17c0fdac636f7269f4885294e2a8d2c52c23590/test/netlib/http/test_response.py#L32)? :smiley:

mhils:
Fixed in #1564. Thanks again for the report!

Issue #1555, opened 2016-09-15
sslstrip.py processes flow.request.content instead of flow.response.content


phackt:
##### Steps to reproduce the problem:

1. https://github.com/mitmproxy/mitmproxy/blob/v0.17.1/examples/sslstrip.py
2. https://github.com/mitmproxy/mitmproxy/blob/master/examples/sslstrip.py

##### What is the expected behavior?

Should pop headers from response (flow.response.headers) instead of request.

##### What went wrong?

def response(flow):
    flow.request.headers.pop('Strict-Transport-Security', None)
    flow.request.headers.pop('Public-Key-Pins', None)

##### Any other comments? What have you tried so far?

Issue created in order to track discussion https://discourse.mitmproxy.org/t/inline-script-sslstrip-py-with-flow-request-headers-in-response-function/169
---

Mitmproxy Version: 0.17.1
Operating System: #1 SMP Debian 4.6.4-1kali1 x86_64 GNU/Linux

phackt:
PR merged ;)

PR #1556, opened 2016-09-15
Fixes - #1555 sslstrip.py flow.response.headers

phackt:


examples/sslstrip.py
```diff
@@ -9,19 +9,26 @@ def request(flow):
...
-    flow.request.headers.pop('Strict-Transport-Security', None)
-    flow.request.headers.pop('Public-Key-Pins', None)
+    flow.response.headers.pop('Strict-Transport-Security', None)
+    flow.response.headers.pop('Public-Key-Pins', None)
 
     # strip links in response body
     flow.response.content = flow.response.content.replace('https://', 'http://')
 
+    # strip meta tag upgrade-insecure-requests in response body
+    csp_meta_tag_pattern = '<meta.*http-equiv=["\']Content-Security-Policy[\'"].*upgrade-insecure-requests.*?>'
```
mhils: This will not work on Python 3 (to which we are transitioning) because `.content` is bytes, not a str. Can you make the pattern a bytes object as well (like so: `b"pattern"`)?

Also, feel free to simplify that regex if you wish. I'd be fine if it's just `b"upgrade-insecure-requests[;\s]*(?=[\s\S]+</head>)"` or something along those lines.

phackt: Hi Maximilian, thanks for your review. I made the changes, the fact is do you still want me to simplify the csp_meta_tag_pattern (csp_meta_tag_pattern = b'<meta.*upgrade-insecure-requests.*?>' should take the risk to match every meta tag with this string, even meta description tag.).

Thanks,

mhils: @phackt: Do the regex as you wish - I think the current version is good enough, but a more greedy one would work for me as well. I'd merge right now, but I don't see the changes on the PR branch yet :)

phackt: @mhils : Hello, done. Regards!

mhils:
Fantastic, thanks!

PR #1566, opened 2016-09-22
add websocket on/off switch, improve logging (fix #1546, fix #1547)

### LINKED ISSUE #1546: WebSocket Support needs an on/off switch.

1. Run mitmproxy
2. Visit https://www.websocket.org/echo.html
3. Send some messages

The Websocket activity should be reflected in the UI.

##### What went wrong?

There is no UI indication that any activity is ongoing after the initial 101 Upgrade.

Similar to our TCP mode, we should have a `--websocket/--no-websocket` switch to control if the WS protocol is enabled. As we don't have a UI integration for this either, I would suggest to keep it turned off by default.

Setting this as a release blocker because the current master version swallows WS contents without showing something.

---

Mitmproxy Version: master
Operating System: Win10 x64

### LINKED ISSUE #1547: WebSocket Messages should be shown to the user.

1. Run mitmproxy
2. Visit https://www.websocket.org/echo.html
3. Send some messages

The Websocket activity should be reflected in the UI.

##### What went wrong?

There is no UI indication that any activity is ongoing after the initial 101 Upgrade.

Similar to our TCP mode, the easiest thing to do for now is just to display messages in the event log. Do we just want to do the same for now in builtins/dumper and console/master for now?

---

Mitmproxy Version: master
Operating System: Win10 x64

mhils:
@Kriechi: Can I get a quick review from you for this one?

- We now have a `--websockets/--no-websockets` switch. Off by default (like `--raw-tcp`), but there's an info log message if a client attempts a websocket connection.
- Move the websocket upgrade code into protocol/http as it is entirely HTTP specific. Before that it would have broken on non-HTTP Flows.
- Set logging verbosity of websocket data frames from debug to info (this is what #1547 was referring to, where I did not have the time to log into the code)
- Move websocket logging before calling `.send()` so that we still log if sending fails.

### DIFF
--- a/mitmproxy/options.py
+++ b/mitmproxy/options.py
@@ -70,6 +70,7 @@ class Options(optmanager.OptManager):
             mode = "regular",  # type: str
             no_upstream_cert = False,  # type: bool
             rawtcp = False,  # type: bool
+            websockets = False,  # type: bool
             spoof_source_address = False,  # type: bool
             upstream_server = "",  # type: str
             upstream_auth = "",  # type: str
@@ -129,6 +130,7 @@ class Options(optmanager.OptManager):
         self.mode = mode
         self.no_upstream_cert = no_upstream_cert
         self.rawtcp = rawtcp
+        self.websockets = websockets
         self.spoof_source_address = spoof_source_address
         self.upstream_server = upstream_server
         self.upstream_auth = upstream_auth
--- a/mitmproxy/protocol/http.py
+++ b/mitmproxy/protocol/http.py
@@ -1,23 +1,20 @@
 from __future__ import absolute_import, print_function, division
 
-import time
-import sys
-import traceback
-
 import h2.exceptions
+import netlib.exceptions
 import six
-
+import sys
+import time
+import traceback
 from mitmproxy import exceptions
 from mitmproxy import models
-from mitmproxy.protocol import base
-
-import netlib.exceptions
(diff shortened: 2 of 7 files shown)


mitmproxy/options.py
```diff
@@ -70,6 +70,7 @@ def __init__(
             mode = "regular",  # type: str
             no_upstream_cert = False,  # type: bool
             rawtcp = False,  # type: bool
+            websockets=False,  # type: bool
```
Kriechi: All the other lines in this block use spaces around the `=`. Would you mind fixing it please to keep consistency?

Kriechi: I think we should enable it - so people can start using it right with the next release?
Or do you think it might be buggy?

mitmproxy/protocol/http.py
```diff
@@ -1,17 +1,15 @@
...
-
+import sys
+import time
+import traceback
 from mitmproxy import exceptions
 from mitmproxy import models
 from mitmproxy.protocol import base
-
-import netlib.exceptions
+from mitmproxy.protocol import websockets as pwebsockets
```
Kriechi: Just a note: I usually keep all built-in imports at the top, then 3rd party imports, and lastly our own stuff.
And I don't like the weirdly abbreviated import names like `pwebsockets` or `moptions`.

PR #1571, opened 2016-09-23
Add command line arg to force specific SNI

rofreg:
This fixes an SNI issue in reverse proxy mode.

1. Start a reverse proxy instance that connects to an HTTPS endpoint using SNI:
`mitmproxy -R https://www.example.com --setheader :~q:Host:www.example.com`
3. Open `https://localhost:8080/` in a web browser

mitmproxy should use the domain `www.example.com` to negotiate SNI with the server.

##### What went wrong?

mitmproxy instead sends no SNI value to the server, and the request fails with the following error: `Cannot establish TLS with www.example.com:443 (sni: None): TlsException('Cannot validate certificate hostname without SNI',)`

##### How does this PR fix the issue?

This PR adds a `--server-sni` flag that allows you to specify an SNI value that should always be used. Thus, you can fix the original problem above by adding the new flag: `mitmproxy -R https://www.example.com --setheader :~q:Host:www.example.com --server-sni www.example.com`.

---

Hope this is helpful! Happy to make adjustments if you think there's a better way to address this problem :)

### DIFF
--- a/mitmproxy/protocol/tls.py
+++ b/mitmproxy/protocol/tls.py
@@ -326,12 +326,12 @@ class TlsLayer(base.Layer):
           the server connection.
     """
 
-    def __init__(self, ctx, client_tls, server_tls):
+    def __init__(self, ctx, client_tls, server_tls, custom_server_sni = None):
         super(TlsLayer, self).__init__(ctx)
         self._client_tls = client_tls
         self._server_tls = server_tls
 
-        self._custom_server_sni = None
+        self._custom_server_sni = custom_server_sni
         self._client_hello = None  # type: TlsClientHello
 
     def __call__(self):
--- a/test/mitmproxy/test_server.py
+++ b/test/mitmproxy/test_server.py
@@ -101,10 +101,16 @@ class CommonMixin:
         if not self.ssl:
             return
 
+        if getattr(self, 'reverse', False):
+            # In reverse proxy mode, we expect to use the upstream host as our SNI value
+            expected_sni = "127.0.0.1"
+        else:
+            expected_sni = "testserver.com"
+
         f = self.pathod("304", sni="testserver.com")
         assert f.status_code == 304
         log = self.server.last_log()
-        assert log["request"]["sni"] == "testserver.com"
+        assert log["request"]["sni"] == expected_sni
 
 
 class TcpMixin:
--- a/mitmproxy/proxy/root_context.py
+++ b/mitmproxy/proxy/root_context.py
@@ -69,7 +69,7 @@ class RootContext(object):
         # An inline script may upgrade from http to https,
         # in which case we need some form of TLS layer.
         if isinstance(top_layer, modes.ReverseProxy):
-            return protocol.TlsLayer(top_layer, client_tls, top_layer.server_tls)
+            return protocol.TlsLayer(top_layer, client_tls, top_layer.server_tls, top_layer.server_conn.address.host)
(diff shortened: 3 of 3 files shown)


mhils:
Thanks for the fantastic PR! :smiley:

Now this looks like we pretty much always just want to send "www.example.com" as the SNI in reverse proxy mode. Do you think it would make sense to just make that the default behaviour in reverse proxy mode, no matter what the client sends (we could add a flag to disable that _if_ we ever need to)? It'd be great if things would just work out of the box. :)

rofreg:
@mhils Makes sense to me! My only concern is that that *might* be a breaking change for existing users, if for some reason they were depending on the original SNI value passed via the original request. The [docs for overwriting the Host header](http://docs.mitmproxy.org/en/stable/features/reverseproxy.html#host-header) imply that there are use cases where you'd want to pass through the original Host header, and I'm unclear if the same is true for SNI.

All that said, if we're not worried about a breaking change, then I think the ideal behavior for reverse proxy mode would actually be to use "www.example.com" as both the default SNI value AND the default Host header. I imagine that's the desired behavior for 99% of users, but I'm new to this, so it's always possible that I'm missing some important use cases. Thoughts?

mhils:
> All that said, if we're not worried about a breaking change, then I think the ideal behavior for reverse proxy mode would actually be to use "www.example.com" as both the default SNI value AND the default Host header. I imagine that's the desired behavior for 99% of users, but I'm new to this, so it's always possible that I'm missing some important use cases. Thoughts?

I completely agree on this - in fact, I was about to mention that in the review, but then decided to keep this PR simple for now. :-)
I'd be more than happy to merge a PR for SNI and/or the Host header (ideally separate PRs, there shouldn't be an overlap). Let me know if there's anything I can help you with, you're also more than welcome to join us on Slack! :smiley:

rofreg:
@mhils Updated! I believe this should behave as expected, though if I've misunderstood how `self.server_conn` works, just let me know. I'll look into making a separate PR to handle the Host header in reverse proxy mode as well 👍

mitmproxy/protocol/tls.py
```diff
@@ -433,7 +433,9 @@ def server_sni(self):
         """
         The Server Name Indication we want to send with the next server TLS handshake.
         """
-        if self._custom_server_sni is False:
+        if self.config.options.mode == "reverse":
+            return self.server_conn.address.host
+        elif self._custom_server_sni is False:
```
mhils: Can we just extend the constructor to accept a `server_sni` argument which sets `self._custom_server_sni` and pass that [here](https://github.com/mitmproxy/mitmproxy/blob/8021427ab919e5301d9cc0e5d6ef28f19dee048b/mitmproxy/proxy/root_context.py#L72)?

test/mitmproxy/test_server.py
```diff
@@ -101,10 +101,16 @@ def test_sni(self):
         if not self.ssl:
             return
 
+        if hasattr(self, 'reverse') and self.reverse:
```
mhils: Minor nitpick: `if getattr(self, 'reverse'):`

rofreg: Ha, that's definitely better :) I'm usually a Ruby guy, gotta brush up on my Python skills!

mitmproxy/protocol/tls.py
rofreg: Sounds good! On it now.

rofreg:
Updated again! Thanks for the suggestions.

mhils:
Awesome. Thanks again for the great PR and the fast turnaround time! 🍰 🙌 😊

PR #1573, opened 2016-09-23
Include port number when exporting

zweifisch:


mhils:
Thanks for your contribution! Looks like we have a couple of occasions in export.py where we do a bad job with this.

Can we just use `request.url` whereever possible and `netlib.http.url.hostport()` for the other cases?

mhils:
Fixed this in #1656, thanks again!

PR #1581, opened 2016-09-26
Set Host header automatically in reverse proxy mode

rofreg:
Follow-up to #1571. This changes the default behavior of reverse proxy mode so that the Host header automatically matches the upstream host. This solution should still be backwards compatible with existing use of the `--setheader` flag. When this and #1571 are merged, that should make it very easy to run a reverse proxy for endpoints on the open web (e.g. `mitmproxy -R https://example.com`).

I spent a bit of time digging through the codebase, and this seemed like a relatively clean place to insert this special handling, but let me know if you think there's a better location for this logic.

### DIFF
--- a/docs/features/reverseproxy.rst
+++ b/docs/features/reverseproxy.rst
@@ -29,29 +29,14 @@ them to what the server expects.
 Host Header
 -----------
 
-In reverse proxy mode, mitmproxy does not rewrite the host header. While often useful, this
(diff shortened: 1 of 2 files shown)


mhils:
Thank you so much for also updating the docs right away, highly appreciated! 😃 🍰 

It's probably better to have the header updating directly in the protocol part, I changed that in #1582.

PR #1582, opened 2016-09-27
Set Host header automatically in reverse proxy mode 

mhils:
Follow-up to #1581 using a slightly different technique to override the header. I don't want to rely on addons for this, this should always happen.

### DIFF
--- a/docs/features/reverseproxy.rst
+++ b/docs/features/reverseproxy.rst
@@ -29,29 +29,14 @@ them to what the server expects.
 Host Header
 -----------
 
-In reverse proxy mode, mitmproxy does not rewrite the host header. While often useful, this
-may lead to issues with public web servers. For example, consider the following scenario:
+In reverse proxy mode, mitmproxy automatically rewrites the Host header to match the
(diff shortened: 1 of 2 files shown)


rofreg:
I don't believe this is working as intended:

```
>>> mitmdump -d -R https://www.example.com/		
>>> curl http://localhost:8080/

>> GET https://www.example.com/
   Host: localhost.com:8080
   ...
```

It looks like [`self.mode`](https://github.com/mhils/mitmproxy/blob/998f161ab0f55c992aa39b983bf17652fdbe2521/mitmproxy/protocol/http.py#L188) is actually evaluating as `"transparent"` at this point, though I'm not immediately sure why. Thus the new code block doesn't execute, and the old Host behavior persists.

mhils:
Uh - Looks like I screwed up when testing this, thanks for catching that! We need to check `self.config.options.mode` here, the HTTP Layer mode is more restricted and can only be transparent/regular/upstream. I'll fix it ASAP.

rofreg:
I went ahead and made the PR. Looks like everything behaves as expected after that small change :)

PR #1568, opened 2016-09-22
Support socket manipulation through the server_conn hook.

omfgroflbbq:
The checks for self.[rfile|wfile] have been introduced to avoid the following exceptions:
```
Traceback (most recent call last):
  File "/tmp/mitmproxy/mitmproxy/proxy/server.py", line 119, in handle
    root_layer()
  File "/tmp/mitmproxy/mitmproxy/proxy/modes/transparent_proxy.py", line 25, in __call__
    self.disconnect()
  File "/tmp/mitmproxy/mitmproxy/protocol/base.py", line 162, in disconnect
    self.server_conn.finish()
  File "/tmp/mitmproxy/mitmproxy/models/connections.py", line 226, in finish
    tcp.TCPClient.finish(self)
  File "/tmp/mitmproxy/netlib/tcp.py", line 488, in finish
    self.wfile.flush()
AttributeError: 'NoneType' object has no attribute 'flush'

Traceback (most recent call last):
  File "/tmp/mitmproxy/mitmproxy/proxy/server.py", line 119, in handle
    root_layer()
  File "/tmp/mitmproxy/mitmproxy/proxy/modes/transparent_proxy.py", line 25, in __call__
    self.disconnect()
  File "/tmp/mitmproxy/mitmproxy/protocol/base.py", line 162, in disconnect
    self.server_conn.finish()
  File "/tmp/mitmproxy/mitmproxy/models/connections.py", line 226, in finish
    tcp.TCPClient.finish(self)
  File "/tmp/mitmproxy/netlib/tcp.py", line 494, in finish
    self.rfile.close()
AttributeError: 'NoneType' object has no attribute 'close'
```
These exceptions occur if the box, on which mitmproxy is running on, is unable to set up a connection to an external resource, and thereby violates disconnect's precondition.

https://github.com/mitmproxy/mitmproxy/blob/0dbb7033ee1a6e238752381ce99439ef7d38b208/mitmproxy/protocol/base.py#L148-L151

PR #1583, opened 2016-09-27
allow socket creation monkeypatchs

mhils:
This PR supersedes #1568.


Kriechi:
If we allow somebody to hook this - why do we use the private underscore prefix-notation?

mhils:
Uh, very good point. 🙈 
Fixed!

PR #1584, opened 2016-09-28
Fix Host header in reverse proxy mode

rofreg:
Minor PR to fix #1582.


mhils:
I was at that point earlier, then I saw the test failures on Python 2. Thanks for pushing this, I just fixed this in e0cd33b5863cebb78c38873e32540ee4c495ec15 though. :-)

rofreg:
Aha, totally missed that commit on master :) Thanks!

PR #1587, opened 2016-09-29
rename mitmproxy.filt -> mitmproxy.flowfilter

Kriechi:
There seems to be a bug here, which is now fixed:
https://github.com/mitmproxy/mitmproxy/blob/d068000f16cb44b7277de04ba20bd1fc62e4a956/mitmproxy/builtins/filestreamer.py#L30

Also I removed a bit of code duplication here:
https://github.com/mitmproxy/mitmproxy/pull/1587/files#diff-6aaebdf9a233f01096f39e9db9f3d091L63

Since `flow.match` will be removed, I added a `DeprecationWarning` and moved the actual function to `flowfilter.match`.

### DIFF
--- a/mitmproxy/flowfilter.py
+++ b/mitmproxy/flowfilter.py
@@ -408,7 +408,7 @@ class FNot(_Token):
         return not self.itm(f)
 
 
-filt_unary = [
+filter_unary = [
     FAsset,
     FErr,
     FHTTP,
@@ -417,7 +417,7 @@ filt_unary = [
     FResp,
     FTCP,
 ]
-filt_rex = [
+filter_rex = [
     FBod,
     FBodRequest,
     FBodResponse,
@@ -433,7 +433,7 @@ filt_rex = [
     FSrc,
     FUrl,
 ]
-filt_int = [
+filter_int = [
     FCode
 ]
 
@@ -442,7 +442,7 @@ def _make():
     # Order is important - multi-char expressions need to come before narrow
     # ones.
     parts = []
-    for klass in filt_unary:
+    for klass in filter_unary:
         f = pp.Literal("~%s" % klass.code) + pp.WordEnd()
         f.setParseAction(klass.make)
         parts.append(f)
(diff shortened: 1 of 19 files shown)


Kriechi:
meh - lazy Python excuse IMHO :D
`filt` is just super weird - should we rename the import to e.g. `flow_filter` or something?

Either way, there is a bug in the above linked code, that needs fixing.

cortesi:
Max is right - I used `filt` because `filter` is a builtin. Some of the most infuriating Python bugs I've ever had to track down have been due to builtin name clashes (cue Max with "but if you used PyCharm..."). 

I agree that `filt` is, as you say, super weird and should be changed, but I also don't like `flow_filter`, which seems verbose. Maybe some sort of half-hearted compromise like "filtr"? Any other suggestions?

Kriechi:
We are not trying to make a hipster web-startup company... `filtr`... *SCNR* 😄 
Other ideas are welcome!

Kriechi:
@mhils @cortesi: I updated this PR according to our Slack discussion yesterday.

mitmproxy/flowfilter.py
```diff
@@ -492,25 +493,42 @@ def _make():
...
+        filter = bnf.parseString(s, parseAll=True)[0]
+        filter.pattern = s
+        return filter
     except pp.ParseException:
         return None
     except ValueError:
         return None
 
 
+def match(flow, flt):
```
mhils: I would prefer a `match(flt, flow):` signature, as that would be in line with `re.match`. Thoughts?

mitmproxy/flowfilter.py
```diff
@@ -492,25 +493,42 @@ def _make():
 def parse(s):
     # type: (str) -> TFilter
     try:
-        filt = bnf.parseString(s, parseAll=True)[0]
-        filt.pattern = s
-        return filt
+        filter = bnf.parseString(s, parseAll=True)[0]
```
mhils: As tempting as it is, we don't want to override builtins :wink: 
filtexpr, filter_function, flowfilter, ... ?

mitmproxy/flowfilter.py
Kriechi: Makes perfect sense! Done.

mitmproxy/flowfilter.py
Kriechi: meh - I fixed it.

Issue #1585, opened 2016-09-29
Issue w/ trusted SSL intermediate certs in Safari


rofreg:
##### Steps to reproduce the problem:

1. Get a trusted SSL certificate issued by Comodo (let's say it's for `www.rofreg.com`), and convert it into a .pem file that includes all intermediate certs
2. Update your `/etc/hosts` file to redirect `www.rofreg.com` to `127.0.0.1`
3. Run `mitmproxy --cert=your_cert.pem -R https://www.example.com`
4. Open `https://localhost:8080` in Safari

##### What is the expected behavior?

The page should load without any SSL issues.

##### What went wrong?

Safari does not trust the intermediate certificates, and the request is blocked.

##### Any other comments? What have you tried so far?

This one is baffling me pretty good, but I've confirmed that the exact same .pem file DOES work in Safari when served up by a plain Rails server. My hunch is that something is going wrong with the intermediate certs in the .pem file, such that they don't get properly delivered during the handshake, but I'm not yet familiar enough with mitmproxy to know where I should start looking. Any thoughts? I'd be happy to dig into this further myself (especially since I have a valid .pem file that can reproduce the problem), just hoping for some guidance :)

---

Mitmproxy Version: 0.18
Operating System: OS X 10.11.6

mhils:
Potentially stupid question, but did you put the intermediary certs at the end of the PEM file?

(See: http://docs.mitmproxy.org/en/stable/certinstall.html#using-a-custom-certificate)

rofreg:
Yep! Private key first, followed by `www.rofreg.com` cert, followed by intermediate certs in ascending order (e.g. from leaf to root).

mhils:
Ok, not sure why this is happening then, sorry.
I would try and set [SSLKEYLOGFILE](http://docs.mitmproxy.org/en/stable/dev/sslkeylogfile.html), capture the traffic with WireShark and see what mitmproxy actually sends. We don't do much besides calling pyOpenSSL's [use_certificate_chain_file](https://github.com/mitmproxy/mitmproxy/blob/d068000f16cb44b7277de04ba20bd1fc62e4a956/netlib/tcp.py#L826).

rofreg:
Cool – thanks for the tips! Hopefully that'll help me dig into this further over the next few days. I'll let you know if I find anything!

mhils:
You're welcome - feel free to drop by on Slack if you have any ad hoc questions!

rofreg:
[...]

PR #1590, opened 2016-09-30
Fix SSL error on OS X

rofreg:
I've found a fix for #1585, but I'm not sure it's complete yet, or if it might break something that I'm unaware of. I'd love some feedback as to whether this seems like a proper solution.

This fixes an issue that occurs when a user supplies a custom SSL cert w/ intermediate certs that *contradict* the default certifi set of root certificates. In particular, this addresses an issue where the "COMODO RSA Certification Authority" cert in certifi is NOT trusted on OS X by default as of OS X 10.11.6. Even when the user manually supplied a different valid "COMODO RSA Certification Authority" cert in their custom SSL cert .pem file, that cert would be overridden by certifi's default cert.

Now, when the user specifies a custom certificate, mitmproxy will load ONLY that custom certificate when creating an SSL context for the client. That way, we guarantee that the manually specified intermediate certs will be returned to the client. This means the client will definitely receive the set of certs specified in the manually specified .pem file.

### DIFF
--- a/netlib/tcp.py
+++ b/netlib/tcp.py
@@ -817,7 +817,7 @@ class BaseHandler(_Connection):
             until then we're conservative.
         """
 
-        context = self._create_ssl_context(**sslctx_kwargs)
+        context = self._create_ssl_context(ca_pemfile=chain_file, **sslctx_kwargs)
 
         context.use_privatekey(key)
         if isinstance(cert, certutils.SSLCert):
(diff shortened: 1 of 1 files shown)


mhils:
> This fixes an issue that occurs when a user supplies a custom SSL cert w/ intermediate certs that contradict the default certifi set of root certificates. In particular, this addresses an issue where the "COMODO RSA Certification Authority" cert in certifi is NOT trusted on OS X by default as of OS X 10.11.6. Even when the user manually supplied a different valid "COMODO RSA Certification Authority" cert in their custom SSL cert .pem file, that cert would be overridden by certifi's default cert.

WTF. Thanks for digging into this, dealing with OpenSSL is a PITA sometimes. The PR looks right and all tests that make sure that we verify certs properly do pass, so let's go with this!

PR #1591, opened 2016-10-02
http2: add hooks

### LINKED ISSUE #1561: Identify Pushed H2-Flows
In previous versions, one could identify pushed flows by checking for the custom `x-mitmproxy-pushed` header 

However, in https://github.com/mitmproxy/mitmproxy/commit/1c4e1e049cbe5357b1f0191c616ea92e09c5e36b it was removed. Is there a way to classify in a script whether or not a flow has been pushed by the server?

Kriechi:
fixes #1561

### DIFF
--- a/examples/h2_push_promises.py
+++ b/examples/h2_push_promises.py
@@ -1,0 +1,8 @@
+import netlib
+
+
+def h2_push_promise(event):
+    headers = netlib.http.Headers([[k, v] for k, v in event.headers])
+    method = headers[':method']
+    path = headers[':path']
+    print("PUSH_PROMISE received: {} {}".format(method, path))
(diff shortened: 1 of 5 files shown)


cortesi:
We should do something with HTTP2 frames that's sympathetic with whatever we end up doing for websockets. I wonder if a more flexible scheme here might not be an interface like this:

```http2_frame(f, frame, sender)```

Here f is the associated flow, which will have the HTTP response/request attributes, frame is the HTTP2 frame, and sender is "server" or "client". The scripter can then look at the frame type and decide how to handle it, and we don't need a different event for each frame type (or worse, events for some frame types but not others). I think it's pretty important to be able to associate frames with flows, if possible. If we can't, the frame argument can be None.

We can do something similar for websockets:

```websocket_frame(f, frame, sender)```

Within the websocket flow, we'll accumulate all text messages (like TCP and like HTTP). 

Thoughts?

Kriechi:
superseded by #1686

PR #1592, opened 2016-10-02
docs and API-related cleanups

cortesi:


Kriechi:
LGTM!

Should we get rid of the websockets plural everywhere? I though I had a good reason for using the plural, but apparently even the RFC uses the singular version 99% of the time...

cortesi:
@Kriechi There are a few other places where the singular would be a bit better, but nothing that seems totally glaring. Let's schedule this as a cleanup down the road.

PR #1601, opened 2016-10-04
certutils: cap the cert store size at 100 by default

cortesi:
This should be enough to give us reuse without growing infinitely. This is part of fixing the memory situation in mitmdump.

TODO: There's an opportunity here for a better algorithm, that expires certs based on least-recently-accessed time, rather than oldest generated time.

### DIFF
--- a/netlib/certutils.py
+++ b/netlib/certutils.py
@@ -169,6 +169,7 @@ class CertStore(object):
     """
         Implements an in-memory certificate store.
     """
+    STORE_CAP = 100
 
     def __init__(
             self,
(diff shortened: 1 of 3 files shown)


mhils:
> There's an opportunity here for a better algorithm, that expires certs based on least-recently-accessed time, rather than oldest generated time.

For the record, we cannot blindly delete from `.certs`, as the store does not only contain automatically generated certs.

cortesi:
Yes - the expiry queue here is ONLY for generated certs, and does not affect certs added through add_cert. 

The whole certstore mechanism is in dire need of clarification, but that's down the track.

Kriechi:
Not sure how common it is for people, but making `STORE_CAP` an cmd option might be useful to prevent cache trashing in very specific use cases.

mhils:
I'm 👎 on command line options that are irrelevant for 99% of our users. In that very obscure edge case you can always monkeypatch or modify the code directly.

PR #1603, opened 2016-10-06
Simplify EOF detection

mhils:
This PR simplifies EOF handling for tnetstrings. This also allows reading Python's `zipfile.ZipExtFile`, which would previously not meet the EOF condition.

### DIFF
--- a/mitmproxy/flow/io.py
+++ b/mitmproxy/flow/io.py
@@ -26,16 +26,6 @@ class FlowReader:
         """
             Yields Flow objects from the dump.
         """
-
-        # There is a weird mingw bug that breaks .tell() when reading from stdin.
-        try:
-            self.fo.tell()
-        except IOError:  # pragma: no cover
-            can_tell = False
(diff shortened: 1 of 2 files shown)


mitmproxy/flow/io.py
```diff
@@ -26,32 +26,19 @@ def stream(self):
...
-                    off = self.fo.tell()
                 if data["type"] not in models.FLOW_TYPES:
                     raise exceptions.FlowReadException("Unknown flow type: {}".format(data["type"]))
                 yield models.FLOW_TYPES[data["type"]].from_state(data)
-        except ValueError:
-            # Error is due to EOF
-            if can_tell and self.fo.tell() == off and self.fo.read() == b'':
-                return
+        except ValueError as e:
+            if str(e) == "not a tnetstring: empty file":
```
Kriechi: Doing a branch based on an exception message is not very safe.
We should use a specific Exception class for this.

PR #1610, opened 2016-10-14
Fix #1605

### LINKED ISSUE #1605: Interception does not work

1. `mitmproxy -i '.*'`
2. `curl -x localhost:8080 example.com`

Request should be intercepted

##### What went wrong?

Request is not intercepted.

https://discourse.mitmproxy.org/t/is-intercept-working-in-the-head-of-master/186

---

Mitmproxy Version: master
Operating System: Win10 x64

mhils:


Kriechi:
Do we need a test for that?

mhils:
You are right, of course. :)

PR #1621, opened 2016-10-17
Dumper: limit url length

mhils:
This PR limits the URL length echoed by the dumper addon. We already do the same in the console ui, and I find it kind of useful when facing screen-filling ad tracker URLs.

Ok for everyone?

### DIFF
--- a/mitmproxy/builtins/dumper.py
+++ b/mitmproxy/builtins/dumper.py
@@ -130,6 +130,8 @@ class Dumper:
             url = flow.request.pretty_url
         else:
             url = flow.request.url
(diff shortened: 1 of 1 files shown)


cortesi:
I see the use case, but don't like this patch in the current form. Being able to see the full URL in the terminal is too valuable to lose entirely. A better version of this would:

- Add a command-line flag to turn on full-width output (a la ``ps --ww`).
- Use the terminal width to determine the cutoff. See ``click.get_terminal_size()``. 
- Maybe make the cutoff marker clearer by making it a different colour.

PR #1644, opened 2016-10-20
Clean up multidict a bit

cortesi:


mhils:
What's the motivation for this?

cortesi:
The module is over-complicated and and contains code we don't use and don't need. These two commits are no-brainers, but the core needs to be looked at again as well.

mhils:
Just yesterday I started to fix code to use `.to_dict`: https://github.com/mhils/mitmproxy/commit/a2d5491e7c21788dca16d8dab3b726438e286072
Taking out the immutability for cookie attributes is a risky thing because that allows assignments which look valid but are actually no-ops. What's your plan to fix this?

cortesi:
So, a few things:

- The code you have for headers in that patch is incorrect. Requests will blow up if you pass it a header value that's a list.
- The code you have for params that uses to_dict is not optimal. You can pass a list of (key, value) pairs instead. Just use .fields.
- We can think about cookie attributes, maybe by deriving a class that errors when invalid attributes are set. But this is a corner case, and immutable multidicts are simply the wrong solution.

mhils:
> maybe by deriving a class that errors when invalid attributes are set.

This is not why we had the attributes immutable. With this PR merged,
```python
del flow.request.cookies["foo"].attrs["HttpOnly"]
```
or
```python
flow.request.cookies["foo"].attrs["Max-Age"] = 2592000
```
are now both no-ops, which is all but obvious.

cortesi:
The interface here is bad either way, and I'm convinced immutability is not the answer. 

The problem is not with attrs - it's a few levels up the hierarchy. I'm not suggesting we do this exact thing, but consider that the problem disappears if the interface is designed to accurately reflect what's happening to the user:

```python
c = flow.request.cookies()
c["foo"].attrs["HttpOnly"] = True
flow.request.set_cookies(c)
```

Now you can have mutable cookies, mutable attrs and a mutable cookie collection, and just reflect it to the headers when they're set. It's obvious to the user what this does, and there are no surprises. This also avoids the weird unexpected effects through MultiDictView we have at the moment:

```python
c = flow1.request.cookies
# Oops, you've just changed flow1
c["bar"] = "voing"
flow2.request.cookies = c
```
Yes, being able to say ```flow.request.cookies["foo"] = ["bar"]`` is nice, but our interfaces sometimes over-use properties and container interfaces. When we do that, simple things are nice and trivial, but the mental model is incoherent and anything slightly harder becomes very non-obvious. Interfaces like this are "glib" not expressive. 

Anyway, this stuff is hard. Lets talk about this in a separate ticket or on Slack.

Issue #1651, opened 2016-10-22
README is not rendered properly on PyPI


mhils:
#### Steps to reproduce the problem:

1. Open https://pypi.python.org/pypi/mitmproxy
2. Notice that README is broken.

#### Any other comments? What have you tried so far?

PyPI is super restrictive about restructuredText, we probably have some validation errors. Might make sense to have that checked on CI if possible without too much effort.

RitwikGupta:
With this commit, the README passes a reST validator.

PR #1654, opened 2016-10-23
Change duplicate link to anonymous link

RitwikGupta:
Issue #1651, `virtualenv` and `here` point to the same explicit URL. Making `here` anonymous fixes this.


mhils:
Thanks a lot for the PR - I'm sorry to say that I just fixed this an hour ago in #1653 after we discussed our plans for a quick 0.18 patch release on Slack. Sorry for the race condition, I really appreciate the effort! :smiley:

Issue #582, opened 2015-05-13
View used SSL/TLS Version and Cipher Suite


raymontag:
Hey,

is it possible to view the Cipher Suite used to encrypt the traffic? I think this could be interesting while testing some application without the need to set specific ciphers. E.g. this information could be saved as an attribute in the HTTPMessage class. Moreover this could be helpful to check if the something was really encrypted, e.g. by setting the attribute to None (or is this information already provided by the scheme attribute?).

mhils:
> (or is this information already provided by the scheme attribute?)

yes. We also have the flow detail view (`X` in 0.11.3, more prominently on the current github master) to view the certificate.
Displaying the current cipher suite is useful, but we should keep in mind that the cipher suite may change at any point in time of the TLS connection, which makes displaying/storing it not entirely trivial.

mhils:
From #1589: It would also be very useful to display the TLS version that has been used for the server and client connections. If anyone picks this up, let me know and I'm happy to help!

PR #1664, opened 2016-10-25
Resolved #1639: display sni on ClientConnection

### LINKED ISSUE #1639: Display Server Name Indication in the UI

1. `mitmproxy`
2. `curl -x localhost:8080 https://example.com/`
3. View Details Tab

The Server Name Indication sent by the client should be displayed in the UI.

##### What went wrong?

The SNI is not displayed in the "Client Connection" section.

The first step here is to make a `.sni` attribute in ClientConnection which also gets serialized. I'm happy to help if anyone wants to pick this issue up! :smiley:

---

Mitmproxy Version: master
Operating System: Win10 x64

chhsiao90:
There is still one test cannot pass that I need your advise how could I solved that.

The test is ```test/mitmproxy/test_flow_format_compat.py``` that will load a dump file ```data/dumpfile-011```
The dumpfile didn't contains sni info so that when the program try deserialize from dumpfile,
the state of ClientConnection didn't contains the sni field.

Is that I need add sni in the dumpfile manually? Or there is another way?
Thanks!

```
    def test_load():
        with open(tutils.test_data.path("data/dumpfile-011"), "rb") as f:
            flow_reader = io.FlowReader(f)
>           flows = list(flow_reader.stream())

test/mitmproxy/test_flow_format_compat.py:9: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
mitmproxy/io.py:43: in stream
    yield FLOW_TYPES[data["type"]].from_state(data)
mitmproxy/flow.py:112: in from_state
    f.set_state(state)
mitmproxy/flow.py:107: in set_state
    super().set_state(state)
mitmproxy/stateobject.py:60: in set_state
    obj = cls.from_state(val)
mitmproxy/connections.py:81: in from_state
    f.set_state(state)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <ClientConnection: [ssl] None>, state = {'address': {'address': [b'127.0.0.1', 60786], 'use_ipv6': False}, 'timestamp_end': None, 'timestamp_start': 1469208633.936}

    def set_state(self, state):
        """
            Load object state from data returned by a get_state call.
            """
        state = state.copy()
        for attr, cls in self._stateobject_attributes.items():
>           val = state.pop(attr)
E           KeyError: 'sni'

mitmproxy/stateobject.py:52: KeyError
```

### DIFF
--- a/mitmproxy/connections.py
+++ b/mitmproxy/connections.py
@@ -20,6 +20,7 @@ class ClientConnection(tcp.BaseHandler, stateobject.StateObject):
         timestamp_start: Connection start timestamp
         timestamp_ssl_setup: TLS established timestamp
         timestamp_end: Connection end timestamp
+        sni: Server Name Indication sent by client during the TLS handshake
     """
 
     def __init__(self, client_connection, address, server):
@@ -40,6 +41,7 @@ class ClientConnection(tcp.BaseHandler, stateobject.StateObject):
         self.timestamp_end = None
         self.timestamp_ssl_setup = None
         self.protocol = None
+        self.sni = None
 
     def __bool__(self):
         return bool(self.connection) and not self.finished
@@ -61,6 +63,7 @@ class ClientConnection(tcp.BaseHandler, stateobject.StateObject):
         timestamp_start=float,
         timestamp_ssl_setup=float,
         timestamp_end=float,
+        sni=str,
     )
 
     def copy(self):
@@ -86,12 +89,14 @@ class ClientConnection(tcp.BaseHandler, stateobject.StateObject):
             ssl_established=False,
             timestamp_start=None,
             timestamp_end=None,
-            timestamp_ssl_setup=None
+            timestamp_ssl_setup=None,
+            sni=None
         ))
 
     def convert_to_ssl(self, *args, **kwargs):
         super().convert_to_ssl(*args, **kwargs)
         self.timestamp_ssl_setup = time.time()
+        self.sni = self.connection.get_servername()
 
     def finish(self):
         super().finish()
--- a/mitmproxy/io_compat.py
+++ b/mitmproxy/io_compat.py
@@ -66,6 +66,7 @@ def convert_017_018(data):
 
 def convert_018_019(data):
     data["version"] = (0, 19)
+    data["client_conn"]["sni"] = None
     return data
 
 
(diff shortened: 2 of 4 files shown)


chhsiao90:
refs #1639

mitmproxy/proxy/protocol/tls.py
```diff
@@ -336,6 +336,7 @@ def __call__(self):
             # Peek into the connection, read the initial client hello and parse it to obtain SNI and ALPN values.
             try:
                 self._client_hello = TlsClientHello.from_client_conn(self.client_conn)
+                self.client_conn.sni = self._client_hello.sni
```
mhils: This is using our own TlsClientHello parser, which may not work under some circumstances. Can we use [`.get_servername()`](http://www.pyopenssl.org/en/stable/api/ssl.html?highlight=get_servername#OpenSSL.SSL.Connection.get_servername) [here](https://github.com/chhsiao90/mitmproxy/blob/7274b3a58f8eb36e68bc4d758c5e191b684d16d9/mitmproxy/proxy/protocol/tls.py#L485) instead?

mhils:
(Note to whoever merges this: please add https://github.com/mitmproxy/mitmproxy/blob/21f133fae9faf16f5b4322fab3e92eaa8a2bb719/mitmproxy/io_compat.py#L58 to the 0.19 converter afterwards. I refrain from doing that right away to avoid unecessary merge conflicts)

mitmproxy/proxy/protocol/tls.py
chhsiao90: Got it, I will update it.

chhsiao90:
Had fixed about how we get the sni, and io_compat.py for versions compatibility of dumpfile.
Please help review it, thanks!
I will help on #582 when I'm free and if there anyone else pick it up~

mhils:
That was a quick turnaround - look great, thank you very much! 😊

PR #1668, opened 2016-10-26
Make Options keyword-only

mhils:
If noone objects, I would like to make our option classes keyword-only as proposed in [PEP 3102](https://www.python.org/dev/peps/pep-3102/):  Arguments can only be supplied by keyword and not as a positional argument.


cortesi:
LGTM

PR #1667, opened 2016-10-25
[MINOR] fix client log_replay

navinpai:
Resolves #1666 

This is the output for mitmdump -nc stats 

running mitmdump -nc stats

PR #1669, opened 2016-10-26
Typecheck options

mhils:
This PR adds strict run-time type checks to the options manager. Implementing this uncovered a whole bunch of inconsistencies regarding which type we specified and what got actually passed around, so this change should improve stability by quite a bit 😄 

This also fixes #1667 and #1668.


setup.py
```diff
@@ -67,7 +67,7 @@
         "cryptography>=1.3, <1.6",
         "cssutils>=1.0.1, <1.1",
         "Flask>=0.10.1, <0.12",
-        "h2>=2.4.1, <3",
+        "h2>=2.4.1, <2.5",
```
cortesi: This change relevant to the patch?

cortesi:
Excellent. This is a nice improvement.

setup.py
Kriechi: python-hyper is very strong on semver. Using `<3` as upper bound is ok.

test/mitmproxy/utils/test_data.py
```diff
@@ -1,7 +1,8 @@
+import pytest
 from mitmproxy.utils import data
-from . import tutils
 
 
 def test_pkg_data():
     assert data.pkg_data.path("tools/console")
-    tutils.raises("does not exist", data.pkg_data.path, "nonexistent")
```
Kriechi: Do we still have a use for our custom `tutils.raises` handler?
Maybe we should ditch it everywhere and move to `pytest.raises`?

setup.py
mhils: I had test failures for 2.5 locally, which was released today. This is to have a clean CI build here. We can bump this again in a separate PR and let CI decide.

Kriechi: If you can pinpoint the failures, please let @python-hyper know!

test/mitmproxy/utils/test_data.py
cortesi: I'm -0.001 on this. The string-based exception stuff can be nice, so moving to pytest.raises would be a slight net negative. If we can make an argument that the context handler pytest.raises offers would be useful, I might be +0.001.

PR #1670, opened 2016-10-26
Resolved #582: display ClientConnection select cipher of TLS

### LINKED ISSUE #582: View used SSL/TLS Version and Cipher Suite

chhsiao90:
Resolved #582 
Display the select cipher of TLS on the tui

### DIFF
--- a/mitmproxy/connections.py
+++ b/mitmproxy/connections.py
@@ -22,6 +22,7 @@ class ClientConnection(tcp.BaseHandler, stateobject.StateObject):
         timestamp_end: Connection end timestamp
         sni: Server Name Indication sent by client during the TLS handshake
         cipher_name: The current used cipher
+        tls_version: TLS version
     """
 
     def __init__(self, client_connection, address, server):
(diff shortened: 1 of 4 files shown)


mhils:
Fantastic, thanks! This was requested for a loong time 😊 🎉 🍰 

Next stop: ServerConnection 😉

Tushar-Rastogi:
Guys, any update on the server connection details?
@mhils @chhsiao90

mhils:
@Tushar-Rastogi: all that information is present in the data (https://docs.mitmproxy.org/stable/api/mitmproxy/connection.html), if it's not exposed in the UI I'd be more than happy to accept a pull request! 😃

PR #1671, opened 2016-10-26
Bump h2 version

mhils:


mhils:
cc @Kriechi/@python-hyper. I don't have time to look into this right now, but I'll push a mitmproxy 0.18.2 release that limits the h2 version tonight.

Lukasa:
You should note that, in addition to the changes that immediately broke mitmproxy, we added *loads* more header validation in 2.5.0. You'll need to [turn it off](https://python-hyper.org/h2/en/stable/api.html#configuration). Probably worth disabling normalisation too. 😁

test/mitmproxy/protocol/test_http2.py
```diff
@@ -298,9 +298,9 @@ def handle_server_event(cls, event, h2_conn, rfile, wfile):
 
                 headers = [(':status', '200')]
                 if event.priority_updated:
-                    headers.append(('priority_exclusive', event.priority_updated.exclusive))
-                    headers.append(('priority_depends_on', event.priority_updated.depends_on))
-                    headers.append(('priority_weight', event.priority_updated.weight))
+                    headers.append(('priority_exclusive', str(event.priority_updated.exclusive)))
+                    headers.append(('priority_depends_on', str(event.priority_updated.depends_on)))
+                    headers.append(('priority_weight', str(event.priority_updated.weight)))
```
mhils: should be bytes, not str?

PR #1677, opened 2016-10-26
Add mypy to lint

cortesi:
An extremely modest beginning, but we'll roll this out over the whole codebase
in time.

Also move flake8 and mypy dependencies into setup[dev].

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -18,9 +18,7 @@ changedir = docs
 commands = sphinx-build -W -b html -d {envtmpdir}/doctrees . {envtmpdir}/html
 
 [testenv:lint]
-deps =
-  flake8>=2.6.2, <3.1
-  rstcheck>=2.2, <3.0
 commands =
   flake8 --jobs 8 --count mitmproxy pathod examples test
   rstcheck README.rst
+  mypy -s ./mitmproxy/addonmanager.py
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -18,9 +18,7 @@ changedir = docs
 commands = sphinx-build -W -b html -d {envtmpdir}/doctrees . {envtmpdir}/html
 
 [testenv:lint]
-deps =
```
mhils: This should stay here, so that the lint env does not download all of mitmproxy's dependencies/installs mitmproxy itself.

cortesi: Eventually, we're going to extend mypy to check interfaces for upstream packages that have type specifications for their APIs. To do this, those packages will need to be installed. In the short term, I'm not sure there's a significant impact for us to installing all of mitmproxy here?

cortesi: OK, I don't feel strongly either - let's keep this version for now and see if we feel adding the dependencies in again is worth it. The Linux Travis builds usually happen in parallel, which means that as long as our lint job is < our longest job, there should be no practical impact for CI. Looking at this latest build, it seems lint goes from about 1.5 minutes to 5.5 minutes with this patch. For local testing the tox environment is cached so there's no impact (apart from first build).

PR #1663, opened 2016-10-24
Replace mitmproxy.utils.lrucache with functools.lru_cache

mhils:
This supersedes @SahilC's #1659, which I acidentally closed by force-pushing the wrong branch to his master branch. Apologies 🙈 

Long story short, 7d9f1d2 fixes a bug where messages needed to be hashable.

PR #1679, opened 2016-10-27
Add tests for mitmproxy.utils.lrucache

alexwlchan:
Spotted there was a gap in the coverage here.


mhils:
Hey,

Sorry for the confusion - we were actually just replacing that with `functools.lru_cache` in #1663 and it looks like we left the implementation around. Sorry! 😳

PR #1686, opened 2016-10-28
http2: add an indicator for pushed streams

Kriechi:
This PR adds an indicator to mitmproxy and mitmdump to indicate pushed streams.

screen shot 2016-10-27 at 19 44 27

screen shot 2016-10-27 at 19 42 50

To keep track of this information, the `HTTPLayer` now gets to keep track of its own `flow` as attribute.

Furthermore, this PR introduces a `flow.metadata` dict, which can be used to store additional information about flows.
This might be useful for scripts/addons to store external data:
  * correlating SQL query times from an external DB system
  * computed statistics
  * a GUID to keep track of each flow
  * etc.

### DIFF
--- a/mitmproxy/addons/dumper.py
+++ b/mitmproxy/addons/dumper.py
@@ -116,7 +116,8 @@ class Dumper:
         else:
             client = ""
 
-        method = flow.request.method
+        pushed = ' PUSH_PROMISE' if 'h2-pushed-stream' in flow.metadata else ''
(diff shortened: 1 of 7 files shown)


cortesi:
An aside regarding the GUID idea - this is what flow.id is supposed to be already. I'm beginning to use this more actively (see the View stuff I'm working on). +1 on a general metadata bucket for flows, though.

PR #1697, opened 2016-10-31
Update for Ubuntu 16.04

xorrbit:
Update with service restart commands for Ubuntu 16.04+.
Add sudo prefix to iptables commands.


cortesi:
Thanks for this! 

I'm merging this patch, but I think I would actually be happy to just keep documentation for the latest LTS and subsequent releases in our docs. "Over documenting" platform minutiae is a real maintenance burden, and I think most users installing mitmproxy will be on reasonably recent versions of Ubuntu.

PR #1698, opened 2016-10-31
Wsgiapp: Only raise exception if an error occurred.

vhaupert:
Currently, mitmproxy always raises an exception when serving a wsgiapp even if everything worked as expected. This patch causes mitmproxy to only raise an exception if an error actually occurred.


cortesi:
Good catch, thanks.

PR #1712, opened 2016-11-03
[MINOR] fix link in documentation

navinpai:
Fix Nose link in pathod test documentation.

Current: http://docs.mitmproxy.org/en/stable/pathod/test.html
Fixed: http://imgur.com/a/aRDZS


cortesi:
Thanks!

PR #1720, opened 2016-11-04
proxy.protocol.http-related refactoring

cortesi:
- Flatten proxy core so we can see what's up. This will un-flatten again once we've cleaned things up.
- Extract upstream auth into an addon
- Extract proxy auth into an addon
- Disable auth in web app for now, since it re-used the proxy auth mechanism. This sharing is not worth it - the web app should probably just use the built-in Tornado auth stuff.


mitmproxy/proxy/protocol/http.py
```diff
@@ -120,35 +112,72 @@ def set_server(self, address):
...
+
+
 class HttpLayer(base.Layer):
 
     def __init__(self, ctx, mode):
         super().__init__(ctx)
+        if mode not in MODES:
+            raise exceptions.ServerException("Invalid mode: %s"%mode)
         self.mode = mode
-        self.flow = None  # type: http.HTTPFlow
```
cortesi: At every step here I'm making sure that the tests don't break. If there's a behaviour you see me introducing here, it would be super helpful just to add a test for it. I'm proceeding on the assumption that anything not covered in the test suite we can fix up after the refactoring.

Issue #1638, opened 2016-10-20
Certificate Verification Error with -T, not in regular mode


phackt:
##### Steps to reproduce the problem:

Mitmproxy v0.18.1

Hello Team, i'm creating a transparent proxy with a custom sslstrip.py script (keeping if possible clear connection between proxy and victim, and on the other side mitmproxy is replaying the https one on its side with the destination server).

On the routing machine with the mitmproxy running:
1. sysctl -w net.ipv4.ip_forward=1
2. echo 0 | sudo tee /proc/sys/net/ipv4/conf/*/send_redirects
3. iptables -t nat -A PREROUTING -i wlan0 -p tcp --dport 80 -j REDIRECT --to-port 8080
4. mitmproxy -T --anticache --host --anticomp --noapp --script ./sslstrip.py --eventlog

Victim is configured to route on the routing machine with mitmproxy running.
##### What is the expected behavior?

With regular proxy tests, everything is working fine, no certificate problem for mitmproxy when it's establishing the httpS connection.
##### What went wrong?

But in transparent mode:

GET https://assure.ameli.fr/
    ← Certificate Verification Error for 93.174.145.36: hostname '93.174.145.36' doesn't match u'assure.ameli.fr'

It seems that it's expecting the hostname, not the ip address of the destination server. Don't know what could be the root cause of this problem. I will be glad if you have an idea, i will be able to investigate further.

If needed can find all the stuff for my test here, mainly the screenshots done: 
https://github.com/phackt/share/tree/master/mitmproxy

Thanks again for your work guyz,
## Phackt.

Mitmproxy Version: v0.18.1
Operating System: Kali2 (Debian Jessie)

mhils:
What kind of client are you using? Does it send a Server Name Indication as part of the TLS handshake?

phackt:
I'm accessing via web browser (Chrome) on the victim's machine, but everything stripped is requested as classic HTTP connection from web browser, intercepted and detected as secured host with mitmproxy (previously stripped) so scheme is changed to https and port to 443. I add that i do not want to capture TLS initiated by victim's web browser (in case of HSTS and so on), just let it flow. 

Just wondering what can be changed between transparent and regular mode with the TLS connection initiating by mitmproxy (secured web server <--- HTTPS ---> mitmproxy <-- HTTP --> victim).

[...]

PR #1724, opened 2016-11-08
fix SNI for transparent mode - #1638

phackt:
In transparent mode host is set with the target server ip.
Attribute flow.request.host is used as SNI while mitmproxy is
initiating TLS handshake, so it should be set with the pretty_host.

### DIFF
--- a/examples/sslstrip.py
+++ b/examples/sslstrip.py
@@ -16,6 +16,10 @@ def request(flow):
     if flow.request.pretty_host in secure_hosts:
         flow.request.scheme = 'https'
         flow.request.port = 443
+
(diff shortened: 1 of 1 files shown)


examples/sslstrip.py
```diff
@@ -16,6 +16,7 @@ def request(flow):
     if flow.request.pretty_host in secure_hosts:
         flow.request.scheme = 'https'
         flow.request.port = 443
+        flow.request.host = flow.request.pretty_host
```
mhils: Can we add a comment here explaining why this is necessary?

phackt: yes of course!

mhils:
Thanks!

PR #1732, opened 2016-11-12
Added proxy_username to HTTPFlow object

martyn-w:
Hi,

This PR adds a variable `proxy_username` to the HTTPFlow class, which is used to store the username authenticating with MITMProxy.

I'm quite new to the codebase (and also to Python) so this may not be structured in the best way possible - please could you review the code and feel free to suggest improvements or additional tests?

Thanks,

Martyn.

### DIFF
--- a/mitmproxy/http.py
+++ b/mitmproxy/http.py
@@ -163,6 +163,8 @@ class HTTPFlow(flow.Flow):
         """:py:class:`ClientConnection` object """
         self.intercepted = False  # type: bool
         """ Is this flow currently being intercepted? """
+        self.proxy_username = None # type: string
+        """ :py:class:`String` object """
 
     _stateobject_attributes = flow.Flow._stateobject_attributes.copy()
     _stateobject_attributes.update(
@@ -172,7 +174,7 @@ class HTTPFlow(flow.Flow):
 
     def __repr__(self):
         s = "<HTTPFlow"
-        for a in ("request", "response", "error", "client_conn", "server_conn"):
+        for a in ("request", "response", "error", "client_conn", "server_conn", "proxy_username"):
             if getattr(self, a, False):
                 s += "\r\n  %s = {flow.%s}" % (a, a)
         s += ">"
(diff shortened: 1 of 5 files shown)


cortesi:
Hi there, and thanks for working on mitmproxy. I'm rewriting this portion of the codebase right now (proxy and upstream authentication). Could you fill me in on the use-case here?

martyn-w:
Hi @cortesi,

Thanks for the comment. Just to clarify the use case: we have a system with a number of different users all connecting to the same instance of MITMProxy, each authenticated with their own username and password. We want to be able to match browsing patterns to different users, hence the need to know the username authenticating with the proxy when inline scripts run.

Thanks,

Martyn.

cortesi:
Hi Martyn,

Great, that makes sense. Let's see if we can reconcile the direction I'm moving auth in with your use case.

Authentication is moving out into an addon that hooks into the proxy core through the usual addon events. I'm working on this right now, and should have the first version of this merged later today or tomorrow. I want to minimise the number of attributes we have on the core objects, especially if they are related to things entirely dealt with in an addon. We have some other attributes that aren't "clean" in this way, but those are going to go away in the longer term.

I see that you can't pull this from the request headers, because our authentication strips the headers before forwarding and serialising them. However, the headers are still present at the point where the *requestheaders* event is called. Now, we have recently added a *metadata* attribute to flows that get serialized, and can contain arbitrary annotations from addons. A better approach to your problem might simply be to hook in to the *requestheaders* event and then to add an annotation to the metadata. Since I'm working on this code now, adding that annotation might even just be a standard feature of the authentication addon.

Would you mind giving me a wee bit of time to get the new structure down, after which we can discuss how to make things work for you?


Cheers,


Aldo

martyn-w:
Hi Also,

Yes, that sounds like a much cleaner approach. Please go ahead and work out the structure, then feel free to ping me. Thanks again for looking at this.

Best,

Martyn.

cortesi:
@martyn-w Excellent. I'll close this PR for now - I've added #1734 to track this. Please also consider joining our developer Slack where we can discuss this sort of thing with lower latency.

PR #1736, opened 2016-11-12
test/netlib/test_tcp: Update ciphers

dwfreed:
Try updating ciphers to see if this fixes Travis builds.


dwfreed:
Docs failure only.  Look's like Travis's installed openssl has started refusing to allow RC4-SHA.

PR #1738, opened 2016-11-13
restrict ALPN offers if host changes

### LINKED ISSUE #1737: Crash when connecting to HTTP/1.1 server via mitmdump and changing host to a server that supports h2

1. Run mitmdump with a simple inline script that changes the host to one capable of h2. 
2. Open https://holmusk.com (any website that only supports http1.1) via Chrome connected to proxy server
3. Observe mitmdump stdout for error like below - 
observed error

1) Here is the script I am using to change the host - 
sample-script
2) Changing host to a server that supports HTTP1.1 only helps to bypass the issue. So does passing in the disable http2 flag when starting mitmdump. Also force setting scheme and port to http and 80 helps to mask the issue (since h2 works on https only).
3) I have already created an AWS CloudFront distribution that is capable of h2. `flow.request.host = 'dvmu1lzlfqifb.cloudfront.net'`. Please pass in the original website hostname as a request header `flow.request.headers['Dex-Client-Domain'] = 'https://holmusk.com/'` if you want to use my distribution.
##### System information

Mitmproxy version: 0.19
Python version: 3.5.2
Platform: Linux-4.4.0-45-generic-x86_64-with-Ubuntu-16.04-xenial

Kriechi:
fixes #1737

If the client has already negotiated an ALP, then force the server to use the same. 
Otherwise we get a protocol mismatch between client <-> server.
But in order for HTTP/2 to work properly we have to relay some messages directly between client and server.

This can only happen if the host gets changed after the initial connection was established. 

E.g.:
 * Client tries to open an HTTPS url,
 * the client offers http/1.1 and h2,
 * the initial host is only capable of http/1.1,
 * then the first server connection negotiates http/1.1,
 * but after the server_conn change, the new host offers h2
 * which results in garbage because the layers don' match.

@muralisr can you please test if this patch fixes your problem?

### DIFF
--- a/mitmproxy/proxy/protocol/tls.py
+++ b/mitmproxy/proxy/protocol/tls.py
@@ -432,7 +432,6 @@ class TlsLayer(base.Layer):
     def __alpn_select_callback(self, conn_, options):
         # This gets triggered if we haven't established an upstream connection yet.
         default_alpn = b'http/1.1'
-        # alpn_preference = b'h2'
 
         if self.alpn_for_client_connection in options:
             choice = bytes(self.alpn_for_client_connection)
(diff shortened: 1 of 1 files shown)


muralisr:
@Kriechi Yes, the issue is solved with this patch. Thank you very much.

It forces the second host to use the original protocol and in my case, I see HTTP1.1 being the final protocol used to deliver my assets.
Just wondering, Is it technically possible to renegotiate the protocol when switching hosts in the future?

Kriechi:
Great to hear that this fixes your issue!

A clean renegotiation would mean to renegotiate a different TLS session with the client - which is not possible AFAIK. 

The next best thing would be to terminate the client connection, and hope that it tries to reconnect, then mitmproxy would have to remember that it *must not* offer the unwanted ALPs...

To sum up, with this PR the following restrictions apply:
  * If the client offers `http/1.1` and `h2`,
  * and the server selects `http/1.1`,
  * then mitmproxy builds up an HTTP/1-layer stack
  * if a script now changes the server to on that is `h2`-capable,
  * we force the new server connection to the same ALP as the client (==`http/1.1`)
  * and everything works.

--------------

  * If the client offers `http/1.1` and `h2`,
  * and the server selects `h2`,
  * then mitmproxy builds up an HTTP/2-layer stack
  * if a script now tries to change the server it fails, because of
https://github.com/mitmproxy/mitmproxy/blob/afa124a9f65364031080b81b04400be4bd05e418/mitmproxy/proxy/protocol/http2.py#L417

Issue #1518, opened 2016-09-01
Improve WebSockets


Kriechi:
Take from #1488:
- ~~frame continuation: abstract frames to messages and provide them to the user~~
- ~~respect `--body_size_limit SIZE` and `--stream SIZE`~~
- implement compression extensions https://tools.ietf.org/html/rfc7692
- ~~add hooks for inline scripts~~

Kriechi:
fixed by #2545

PR #1739, opened 2016-11-13
add WebSocket flows

Kriechi:
I tried to make it as similar as possible to @mhils's `TCPFlows`.
This partially solves #1518.

ToDo:
  - [x] documentation for new events
  - [x] testing if all flows and messages are correct
  - [x] use frame type (text/binary) for content views and decoding

### DIFF
--- a/docs/features/tcpproxy.rst
+++ b/docs/features/tcpproxy.rst
@@ -3,7 +3,7 @@
 TCP Proxy
 =========
 
-WebSockets or other non-HTTP protocols are not supported by mitmproxy yet. However, you can exempt
+Non-HTTP protocols are not supported by mitmproxy yet. However, you can exempt
 hostnames from processing, so that mitmproxy acts as a generic TCP forwarder.
 This feature is closely related to the :ref:`passthrough` functionality,
 but differs in two important aspects:
--- a/docs/features/tcpproxy.rst
+++ b/docs/features/tcpproxy.rst
@@ -3,7 +3,7 @@
 TCP Proxy
 =========
 
-Non-HTTP protocols are not supported by mitmproxy yet. However, you can exempt
+In case mitmproxy does not handle a specific protocol, you can exempt
 hostnames from processing, so that mitmproxy acts as a generic TCP forwarder.
 This feature is closely related to the :ref:`passthrough` functionality,
 but differs in two important aspects:
--- a/docs/scripting/events.rst
+++ b/docs/scripting/events.rst
@@ -158,12 +158,15 @@ HTTP Events
 WebSocket Events
 -----------------
 
+These events are called only after a connection made an HTTP upgrade with
+"101 Switching Protocols". No further HTTP-related events after the handshake
+are issued, only new WebSocket messages are called.
+
 .. list-table::
     :widths: 40 60
     :header-rows: 0
 
     *   - .. py:function:: websocket_handshake(flow)
-
         - Called when a client wants to establish a WebSocket connection. The
           WebSocket-specific headers can be manipulated to manipulate the
           handshake. The ``flow`` object is guaranteed to have a non-None
@@ -172,6 +175,36 @@ WebSocket Events
           *flow*
             The flow containing the HTTP WebSocket handshake request. The
             object is guaranteed to have a non-None ``request`` attribute.
+
+    *   - .. py:function:: websocket_start(flow)
+        - Called when WebSocket connection is established after a successful
+          handshake.
+
+          *flow*
+            A ``models.WebSocketFlow`` object.
+
+    *   - .. py:function:: websocket_message(flow)
+
+        - Called when a WebSocket message is received from the client or server. The
+          sender and receiver are identifiable. The most recent message will be
+          ``flow.messages[-1]``. The message is user-modifiable. Currently there are
+          two types of messages, corresponding to the BINARY and TEXT frame types.
+
+          *flow*
+            A ``models.WebSocketFlow`` object.
+
+    *   - .. py:function:: websocket_end(flow)
+        - Called when WebSocket connection ends.
+
+          *flow*
+            A ``models.WebSocketFlow`` object.
+
+    *   - .. py:function:: websocket_error(flow)
+        - Called when a WebSocket error occurs - e.g. the connection closing
+          unexpectedly.
+
+          *flow*
+            A ``models.WebSocketFlow`` object.
 
 
 TCP Events
@@ -185,6 +218,22 @@ connections.
     :widths: 40 60
     :header-rows: 0
 
+
+    *   - .. py:function:: tcp_start(flow)
+        - Called when TCP streaming starts.
+
+          *flow*
+            A ``models.TCPFlow`` object.
+
+    *   - .. py:function:: tcp_message(flow)
+
+        - Called when a TCP payload is received from the client or server. The
+          sender and receiver are identifiable. The most recent message will be
+          ``flow.messages[-1]``. The message is user-modifiable.
+
+          *flow*
+            A ``models.TCPFlow`` object.
+
     *   - .. py:function:: tcp_end(flow)
         - Called when TCP streaming ends.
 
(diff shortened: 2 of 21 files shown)


mitmproxy/websocket.py
```diff
@@ -0,0 +1,51 @@
...
+        self.timestamp = state.pop("timestamp")
+
+    def __repr__(self):
+        return "{direction} {content}".format(
+            direction="->" if self.from_client else "<-",
+            content=repr(self.content)
+        )
+
+
+class WebSocketFlow(flow.Flow):
```
cortesi: The WebsocketFlow class should also contain a record of the handshake that initiated the websocket connection. That way, we can re-establish the connection if, for instance, we're replaying a websocket message in a websocket connection that has already been disconnected. The URL for the initial handshake is also the "end point" that will be shown in places like the flow list in the GUIs.

mitmproxy/websocket.py
```diff
@@ -0,0 +1,51 @@
+import time
+
+from typing import List
+
+from mitmproxy import flow
+from mitmproxy.types import serializable
+
+
+class WebSocketMessage(serializable.Serializable):
```
cortesi: Here we should retain whether the original data arrived as text or a binary frames. This changes the semantics of the content - the standard says that text frames when assembled always has to result in valid UTF-8, binary can obviously be anything. There will be differences in how the servers deal with this, and in how we should expose the content. Maybe even two classes - WebSocketBinaryMessage, WebSocketTextMessage.

mitmproxy/tools/console/master.py
```diff
@@ -438,6 +438,18 @@ def clear_events(self):
 
     # Handlers
     @controller.handler
+    def websocket_message(self, f):
+        super().websocket_message(f)
+        message = f.messages[-1]
+        direction = "->" if message.from_client else "<-"
+        self.add_log("{client} {direction} websocket {direction} {server}".format(
```
cortesi: After we add the handshake to the flow object, this should show the endpoint URL used to establish the websocket connection.

cortesi:
Great stuff. Looking forward to seeing this in the GUIs!

mitmproxy/websocket.py
Kriechi: Yes - thats one of the todos.

mitmproxy/websocket.py
Kriechi: Right now this information is available in the `WebSocketLayer`. Do you want me to move it into the flow? Or is it fine if it stays in the layer?

cortesi: I think we should pull it up to the flow as .handshake or similar, make sure it's serialized, and so forth.

mitmproxy/websocket.py
Kriechi: Done. ✅

mitmproxy/websocket.py
Kriechi: Done. ✅

mitmproxy/websocket.py
```diff
@@ -0,0 +1,83 @@
...
+
+class WebSocketTextMessage(WebSocketMessage):
+
+    type = 'text'
+
+    def __repr__(self):
+        return "text message: {}".format(repr(self.content))
+
+
+class WebSocketFlow(flow.Flow):
```
cortesi: I was thinking about this overnight (dangerous, I know) and I wonder if we don't want this to be a subclass of HTTPFlow. That is, convert the handshake flow into a WebSocketFlow after the handshake succeeds. That means that a complete WebSocketFlow will have .request, .response and .messages attributes. 

My rationale is:

- It seems uncomfortable to keep a separate copy of the handshake flow. The handshake and websockets messages that follow are just part of the same thing, and we really only have one entity here. Let's remove the duplication.
- This is also true in the interface, where I'd argue that we don't want the handshake flow, and then a separate websocket connection flow immediately following. Instead, we'd prefer to see the handshake + some status indicator to show that it's succesfully connected. Then we can expand the handshake to see the flows.

cortesi: Expanding on this a bit, to make this work we want some way to change the type of a flow without changing its identity. We have a lot of code that unfortunately relies on flow identity in the core. This tells me that we want to move towards a Flow that relies on composition rather than inheritance. I.e. one Flow class that gets decorated over time with specific protocol information.

Kriechi: Thinking about it for a couple of nights:

I think if we have the handshake flow as the main thing, and the websocket messages **only** as additional members, it might be too hidden to work with. For me the fully established websocket is the main entity - which should contain the handshake information.

If one would want to manipulate the handshake and force other options, there is the `websocket_handshake` hook to do just that. After establishing the websocket connection, we follow the same procedure as RawTCP already does - this gives us a very nice and clean symmetry (in terms of how the protocols work).

Promoting the handshake flow to become a WebSocketFlow (after a successful handshake only) sounds nice. However, this entails quite a lot of other refactoring, which I don't want to touch right now in this PR.

This will be especially useful in the following cases:
  * HTTP/1.1 upgrade to WebSocket
  * HTTP/1.1 upgrade to HTTP/2
  * HTTP/1.1 change to RawTCP (e.g. byte-by-byte streaming, or to ignore individual paths, large file up-/downloads)

I haven't made up my mind yet if this will be of any real benefit - but it sounds nice 😄 

The mitmproxy core is in a state of flux at the moment - so I don't want to open up another major architectural change (at least not in this PR by myself). IMHO we should try to get this PR merged (in some form or another thats closer to what we already have), and then iterate to get a new flow architecture that allows us to promote / decorate / whatever.

docs/features/tcpproxy.rst
```diff
@@ -3,7 +3,7 @@
 TCP Proxy
 =========
 
-WebSockets or other non-HTTP protocols are not supported by mitmproxy yet. However, you can exempt
+Non-HTTP protocols are not supported by mitmproxy yet. However, you can exempt
```
mhils: Technically websockets is not a HTTP protocol, so you may want to weaken that statement even further? ☺️

docs/scripting/events.rst
```diff
@@ -158,21 +158,54 @@ HTTP Events
...
     :widths: 40 60
     :header-rows: 0
 
-    *   - .. py:function:: websockets_handshake(flow)
-
-        - Called when a client wants to establish a WebSockets connection. The
-          WebSockets-specific headers can be manipulated to manipulate the
+    *   - .. py:function:: websocket_handshake(flow)
+        - Called when a client wants to establish a WebSocket connection. The
+          WebSocket-specific headers can be manipulated to manipulate the
```
mhils: nitpick: "manipulated to alter"?

mitmproxy/tools/cmdline.py
```diff
@@ -468,20 +474,15 @@ def proxy_options(parser):
...
     )
+
     rawtcp = group.add_mutually_exclusive_group()
     rawtcp.add_argument("--raw-tcp", action="store_true", dest="rawtcp")
     rawtcp.add_argument("--no-raw-tcp", action="store_false", dest="rawtcp",
                         help="Explicitly enable/disable experimental raw tcp support. "
                         "Disabled by default. "
                         "Default value will change in a future version."
                         )
-    websockets = group.add_mutually_exclusive_group()
```
mhils: As long as we don't have any UI indication in mitmproxy (not even in the event log), we should keep the current disabled-by-default behaviour. I'd love to have the TCP mode enabled by default as well, but I think we really can't justify it as long as a user is unable to see what's going on.

In order to get this merged quickly, let's revert the changes here and then follow up with another PR that provides a very simple UI integration and enables it by default.

Kriechi:
Weird traceback - any ideas how to fix it? I just want plain UTF8 or encoded binary text to be displayed.

mhils:
I think the problem is that we implicitly pass `headers=None` [here](https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/contentviews.py#L630-L633) instead of not passing headers, so `headers in metadata` is technically true. :smiley:

Kriechi:
All fixed and ready for final review.

mhils:
Merged! 🎉 🍰 😃

PR #1735, opened 2016-11-12
Exempt localhost from upstream binding introduced in #879

dwfreed:
Fixes #1718.

PR #1742, opened 2016-11-14
Make the upstream bind address a separate option

dwfreed:
This forward ports the changes in #1735 to master, and makes the upstream bind address its own option entirely.  I'm going to build on this to allow combining upstream bind address and spoof source address options to spoof a fixed source address for all clients in another PR.

### DIFF
--- a/mitmproxy/tools/cmdline.py
+++ b/mitmproxy/tools/cmdline.py
@@ -251,6 +251,7 @@ def get_common_options(args):
         ignore_hosts = args.ignore_hosts,
         listen_host = args.addr,
         listen_port = args.port,
+        upstream_bind_address = args.upstream_bind_address,
         mode = mode,
         no_upstream_cert = args.no_upstream_cert,
         spoof_source_address = args.spoof_source_address,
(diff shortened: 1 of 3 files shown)


dwfreed:
There's no way that test failure is my fault :P

mitmproxy/tools/cmdline.py
```diff
@@ -486,6 +487,11 @@ def proxy_options(parser):
         action="store_true", dest="spoof_source_address",
         help="Use the client's IP for server-side connections"
     )
+    group.add_argument(
+        "--upstream-bind-address",
+        action="store", type=str, dest="upstream_bind_address", default='',
+        help="Address to bind upstream requests to (defaults to none)"
+    )
```
mhils: Can/should we make this mutually excluse with spoof_source_address?
(see `websockets` just above this for how)

dwfreed: No, because as mentioned above, I'm next going to make it possible to have a fixed spoof address, instead of using the client address.

PR #1748, opened 2016-11-16
add macOS/OSX installation instructions to onboarding page

Kriechi:
Ideally somebody would contribute installation instructions for other systems as well.
Maybe even make it fancy by only showing the correct block after clicking the icon...


mhils:
👍 

We had instructions in the docs a while ago and decided to remove those because they were a burden to maintain. After seeing a few more users setting up mitmproxy, I think this is definitely worth it, especially for Windows and OSX.

Kriechi:
@mhils: could you maybe a few lines about Windows in a new PR?

PR #1753, opened 2016-11-16
Update Dependencies

mhils:


Kriechi:
Where do the whitespace changes come from?
I tried merging in master to get rid of them, but they still show up in the diff...

PR #1757, opened 2016-11-18
Remove dead run_scripthooks example reference.

fotinakis:
It seems that `run_scripthooks` is dead, so this should fix this example.


Kriechi:
Thanks!

PR #1760, opened 2016-11-19
handle SIGINT with a quit prompt

eboda:
Added a SIGINT handler to mitmproxy to catch CTRL-C key combinations and prompt the user to confirm that he indeed wants to quit. It will prevent accidentally closing down mitmproxy when using CTRL-C.

**Motivation**
 
For example, if you are viewing a flow and click on  `e` to edit it, but then actually reconsider and don't want to edit it after all, instinctively I would hit CTRL-C to cancel the current command. But this kills the whole process and I lose all the work I have done, which is quite frustrating.

### DIFF
--- a/mitmproxy/tools/console/master.py
+++ b/mitmproxy/tools/console/master.py
@@ -3,6 +3,7 @@ import mimetypes
 import os
 import os.path
 import shlex
+import signal
 import stat
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


mhils:
Thanks for the PR! This sounds like a reasonable use case to me. :smiley:

Could you please adjust your code so that the linter does not complain anymore? Other than that, this looks good to me!

eboda:
Lint should be fine now, not sure why the other test failed though.

Thanks!

PR #1781, opened 2016-11-23
Simplify console logging, fix #1780

### LINKED ISSUE #1780: Inline script errors not thrown in eventlog

1. edit any inline script and add an error (example sslstrip-test.py attached with function nothing())
[sslstrip-test.txt](https://github.com/mitmproxy/mitmproxy/files/608581/sslstrip-test.txt)
2. mitmproxy --script sslstrip-test.py --eventlog
3. nothing in eventlog (see screenshot) 
mitmproxy-error-not-shown

Remember that it was working with v0.17.1, Maximilian said it was fixed from v0.18 to v0.18.2, so am i missing something here (a debug mode?)? Mitmproxy is still working, ignoring the script.

Thanks a lot,

##### System information
Mitmproxy version: 0.19
Python version: 3.5.2+
Platform: Linux-4.8.0-kali1-amd64-x86_64-with-Kali-kali-rolling-kali-rolling
Linux distro: Kali kali-rolling kali-rolling

mhils:


cortesi:
The point of using a signal for logging here is to decouple things, and move away from the current situation where we have to pass master up and down our hierarchy for everything. I'd like to retain the signal mechanism, and in fact expand this slowly as we refactor console.

mhils:
How shall we handle signals that are also events?

The problem we have right now is that some events are logged by one of add_log/log/sig_addlog (this is confusing!), but not by the others.

eeeb32b reintroduces the log signal so that `signals.add_log` is just a synonym for `master.add_log`. Subscribing to it will not give you the full log (which is available in `.log` for addons). We can flip it the other way around (`master.log` sends an event to the signal), but then we can't use the signal for sending anymore.

cortesi:
I feel we can do something a lot simpler here. See #1822, which uses a tiny addon to capture .log events. That means we now have two types of events, both of which are handled by sig_add_log: those generated by mitmproxy and its addons, and those generated by the console app itself. The latter aren't (and shouldn't be) passed through the normal event mechanism.

cortesi:
Let's close this one for now, and move ahead as per our Slack discussion.

PR #1799, opened 2016-12-01
Make Travis Great Again (Master Edition)

dwfreed:
Working on improving the speed and accuracy of Travis's testing.

PR #1801, opened 2016-12-01
simplify ALPN and OpenSSL on macOS

Kriechi:
supersedes and closes #1799

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,6 @@
 sudo: false
 language: python
 
-addons:
-  apt:
-    sources:
-      # Debian sid currently holds OpenSSL 1.0.2
-      # change this with future releases!
-      - debian-sid
-    packages:
-      - libssl-dev
-
 env:
   global:
     - CI_DEPS=codecov>=2.0.5
@@ -25,9 +16,21 @@ matrix:
       language: generic
       env: TOXENV=py35 BDIST=1
     - python: 3.5
-      env: TOXENV=py35 BDIST=1
+      env: TOXENV=py35 OPENSSL_OLD
+      addons:
+        apt:
+          packages:
+            - libssl-dev
     - python: 3.5
-      env: TOXENV=py35 NO_ALPN=1
+      env: TOXENV=py35 BDIST=1 OPENSSL_ALPN
+      addons:
+        apt:
+          sources:
+            # Debian sid currently holds OpenSSL 1.1.0
+            # change this with future releases!
+            - debian-sid
+          packages:
+            - libssl-dev
     - python: 3.5
       env: TOXENV=docs
       git:
@@ -39,10 +42,8 @@ install:
   - |
     if [[ $TRAVIS_OS_NAME == "osx" ]]
     then
-      brew update || brew update # try again if it fails
-      brew upgrade
-      brew reinstall openssl
-      brew reinstall pyenv
+      brew update || brew update
+      brew outdated pyenv || brew upgrade pyenv
       eval "$(pyenv init -)"
       env PYTHON_CONFIGURE_OPTS="--enable-framework" pyenv install --skip-existing 3.5.2
       pyenv global 3.5.2
@@ -52,8 +53,8 @@ install:
   - pip install tox
 
 script:
+  - tox -- --cov mitmproxy --cov pathod -v
   - |
-    tox -- --cov mitmproxy --cov pathod -v
     if [[ $BDIST == "1" ]]
     then
         git fetch --unshallow --tags
@@ -80,3 +81,4 @@ cache:
   directories:
     - $HOME/.pyenv
     - $HOME/.cache/pip
+    # - $HOME/build/mitmproxy/mitmproxy/.tox
--- a/test/mitmproxy/net/test_tcp.py
+++ b/test/mitmproxy/net/test_tcp.py
@@ -6,7 +6,7 @@ import random
 import os
 import threading
 import mock
-
+import pytest
 from OpenSSL import SSL
 
 from mitmproxy import certs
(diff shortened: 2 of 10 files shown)


.travis.yml
```diff
@@ -24,11 +15,17 @@ matrix:
       osx_image: xcode7.3
       language: generic
       env: TOXENV=py35 BDIST=1
+      addons:
```
dwfreed: Wrong section, this does nothing.

.travis.yml
```diff
@@ -39,11 +36,10 @@ install:
   - |
     if [[ $TRAVIS_OS_NAME == "osx" ]]
     then
-      brew update || brew update # try again if it fails
-      brew upgrade
-      brew reinstall openssl
-      brew reinstall pyenv
+      pyenv -v
```
dwfreed: We deal with brew here because pyenv is already installed by brew, so running pyenv's provided installer is duplicating things.  Also, curl | bash is awful, and pyenv's installer is meh anyway.

.travis.yml
```diff
@@ -24,11 +15,17 @@ matrix:
...
+          sources:
+            # Debian sid currently holds OpenSSL 1.1.0
+            # change this with future releases!
+            - debian-sid
+          packages:
+            - libssl-dev
     - python: 3.5
       env: TOXENV=py35 BDIST=1
     - python: 3.5
-      env: TOXENV=py35 NO_ALPN=1
```
dwfreed: Don't delete the section that wouldn't have functioning ALPN! Otherwise we'll have the problem I reported all over again later.

.travis.yml
```diff
@@ -80,3 +72,4 @@ cache:
   directories:
     - $HOME/.pyenv
     - $HOME/.cache/pip
+    - $HOME/build/mitmproxy/.tox
```
dwfreed: I'm not sure we should cache .tox; having a clean environment every time will help to catch issues we might not otherwise notice.  99% of what's installed are wheels anyway (already cached by pip), with cryptography being one of the few notable exceptions.  Also, if we change OpenSSL versions, we'd have to dump the caches, otherwise cryptography won't get rebuilt, and then things get fun.

Kriechi: tox does a very good job of recreating environments if necessary.
I actually want to get rid of running the whole test suite twice just because a handful of tests might be different. Finding those tests and expanding them to cover both OpenSSL versions should be much cleaner and a lot easier to maintain.

test/pathod/test_pathoc.py
```diff
@@ -211,45 +210,43 @@ class TestDaemonHTTP2(PathocTestDaemon):
...
-            )
-            with c.connect():
-                resp = c.request("get:/p/200")
-            assert resp.status_code == 200
+    def test_http2(self):
+        c = pathoc.Pathoc(
+            ("127.0.0.1", self.d.port),
+            fp=None,
+            ssl=True,
+            use_http2=True,
```
Kriechi: Yes! Thats the whole point of it :-D
I updated the PR to reflect this more clearly.

dwfreed:
2 things:

1. If you run this edition of the test suite without your ALPN fixes you cherry-picked to master, does it fail?
2. I'm not very familiar with pytest; why do some tests use a decorator, and some seem to take a magic parameter?

test/mitmproxy/net/test_tcp.py
```diff
@@ -532,34 +532,28 @@ class TestALPNClient(tservers.ServerTestBase):
...
+    @pytest.mark.parametrize('has_alpn,alpn_protos, expected_negotiated, expected_response', [
+        (True, [b"foo", b"bar", b"fasel"], b'bar', b'bar'),
+        (True, [], b'', b'NONE'),
+        (True, None, b'', b'NONE'),
+        (False, [b"foo", b"bar", b"fasel"], b'', b'NONE'),
+        (False, [], b'', b'NONE'),
+        (False, None, b'', b'NONE'),
+    ])
+    def test_alpn(self, requires_alpn, monkeypatch, has_alpn, alpn_protos, expected_negotiated, expected_response):
+        monkeypatch.setattr(tcp, 'HAS_ALPN', has_alpn)
```
dwfreed: This should also set the Cryptography_HAS_ALPN attribute, like your disable_alpn function.

Kriechi: Thanks - done.

Kriechi:
It doesn't fail - it simply doesn't run those tests.
So there are two things to consider:

Running tests with an old OpenSSL without ALPN:
just skip all tests that depend on ALPN being available.

Running tests on newer OpenSSL with ALPN:
run certain tests with ALPN and some other tests with disabled/blocked ALPN.

dwfreed:
The test suite as a whole, without 1fcf79f applied, should fail **somewhere**, because you were calling pyopenssl ALPN methods in a manner that was not protected by a check for ALPN availability.  I don't consider this able to be merged (not that I really have much weight in that department) until that is true.

.travis.yml
```diff
@@ -25,9 +16,17 @@ matrix:
       language: generic
       env: TOXENV=py35 BDIST=1
     - python: 3.5
-      env: TOXENV=py35 BDIST=1
+      env: TOXENV=py35 BDIST=1 OPENSSL_OLD
```
dwfreed: This should not have BDIST, because then you'll have two different matrix items running pyinstaller, and thus conflicting on the snapshot server, plus one of them won't have functioning ALPN.

.travis.yml
```diff
@@ -39,21 +38,18 @@ install:
...
-      brew upgrade
-      brew reinstall openssl
-      brew reinstall pyenv
+      brew update || brew update
+      brew outdated pyenv || brew upgrade pyenv
       eval "$(pyenv init -)"
       env PYTHON_CONFIGURE_OPTS="--enable-framework" pyenv install --skip-existing 3.5.2
-      pyenv global 3.5.2
-      pyenv shell 3.5.2
+      pyenv global 3.5.2 && pyenv shell 3.5.2
```
dwfreed: Leave these split, it's cleaner that way.

Issue #953, opened 2016-02-17
Scripts set in config files are cleared when passing scripts on the command line


fimad:
In my config file I have several scripts that provide content views and do other miscellaneous things that I would like to apply to all my sessions. If I invoke `mitmproxy` with any scripts on the command line then all the scripts I defined in my config file are overridden and only the scripts on the command line are used.

Is this the intended behavior? If it is, would it be possible to add a new flag for scripts that are not intended to be overridden by scripts passed on the command line?

mhils:
Happy to see what kind of power users we have :smile: 
This looks more like a silly limitation of ConfigArgParse than intended behavior.

dufferzafar:
I feel this is a legitimate feature request mitmproxy and have created an upstream issue over at ConfigArgParse. If they are willing to add this to the library, I'd be happy to work on it and send them a pull. :smile:

cortesi:
See #1699

PR #1806, opened 2016-12-02
Options

cortesi:
This PR does the following:

- Adds YAML serialization for options, including the ability to rewrite an options file in-place, retaining user comments and formating.
- Ditches configargparse in favor of argparse
- Converts all command-line tools to use a single config file, which is ~/.mitmproxy/config.yaml by default. 
- Amalgamates all the Options classes into one to support a common config file. Having separate files was too awkward. 
- All defaults are now managed in the Options specification rather than being split between options and the command-line parsing functionality. 

Loose ends:

- We lose the ability to set options as environment variables. I think this is rarely used, but we can add it back trivially if we switch to click down the track.
- The *keep_serving* argument used to have a different default in console/web and mitmdump - it's now true by default for all tools. We should re-think this option in a future patch. 

Future work:

- Add config saving to mitmproxy console
- Add a utility method to Options that reads info from parsed arguments and updates the existing config. This would differ from the .update() method in that it would extend sequence types rather than over-writing it to address #953
- Rationalise all the option names. We have a lot of ambiguity, and some overlap between the tools at the moment.


test/mitmproxy/addons/test_dumper.py
```diff
@@ -28,43 +28,40 @@ def test_configure():
 
 
 def test_simple():
-    d = dumper.Dumper()
+    sio = io.StringIO()
+    d = dumper.Dumper(sio)
     with taddons.context(options=dump.Options()) as ctx:
-        sio = io.StringIO()
-        ctx.configure(d, tfile = sio, flow_detail = 0)
+        ctx.configure(d, flow_detail = 0)
```
Kriechi: Please remove the whitespace: `..., kwarg=value, ...`
We only accept whitespace if there is a long list of keyword-args, written one per line.

mitmproxy/options.py
```diff
@@ -30,19 +30,19 @@ def __init__(
             app_port: int = APP_PORT,
             anticache: bool = False,
             anticomp: bool = False,
-            client_replay: Sequence[str] = (),
+            client_replay: Sequence[str] = [],
```
dwfreed: Correct, defaults should never be mutable, because every call gets the same instance if the default is used, so mutating it means all later calls get that mutated version too.

cortesi: This is intended. Options returns a deepcopy of all values to guard against accidental mutation. We need the empty default to compare equal to empty values returned from other parts of our code. This leaves us either with finicky and error-prone list->tuple conversions throughout the code, or trying to special case sequence comparisons. I think just returning copies and specifying mutable default arguments is the least worst option here.

dwfreed: Where are the other comparisons?  Could they just check for falsehood, and then this can just be None?

dwfreed: Alternately, the usual idiom is something like this:

```python
client_replay: Sequence[str] = None,
```

Then in the body of the function:

```python
if not client_replay:
    client_replay = []
```

cortesi: No - we test for equality to check whether an option has changed from the default value or not. We need this in a myriad of places, for instance when we decide whether or not to write a changed option to file.

mitmproxy/optmanager.py
```diff
@@ -1,21 +1,49 @@
 import contextlib
 import blinker
 import pprint
+import inspect
+import copy
+import functools
+import weakref
+import os
+
+import ruamel.yaml
```
cortesi: There's a C component with a pure-Python fallback. I believe this works correctly, but if you have a Windows box could you check for us?

test/mitmproxy/addons/test_dumper.py
cortesi: Getting schooled by the Austrian perspective. ;)

cortesi:
For addon-specific options, the plan is roughly to convert to click first, which gives us the ability to parse the command-line in two phases (allow_extra_args and ignore_unknown_args) - first for built-ins, after which we can load options, then re-look at the command-line arguments for addons. I think the flow for Options will mirror this, and naturally flow from the argument parsing design.

mitmproxy/options.py
mhils: There is an argument to be had that they should be tuples all along to prevent an `options.scripts.append(...)`, which would not trigger the change signals.

cortesi: That would impose a fiddly penalty throughout the codebase without much benefit. It would also not give us a clear consistent way forward if (when) we want to expand config to include maps.

mitmproxy/optmanager.py
```diff
@@ -1,21 +1,49 @@
...
 
 
-class OptManager:
+class _DefaultsMeta(type):
+    def __new__(cls, name, bases, namespace, **kwds):
+        ret = type.__new__(cls, name, bases, dict(namespace))
+        defaults = {}
+        for klass in reversed(inspect.getmro(ret)):
+            for p in inspect.signature(klass.__init__).parameters.values():
+                if p.kind in (p.KEYWORD_ONLY, p.POSITIONAL_OR_KEYWORD):
```
cortesi: The POSITIONAL_OR_KEYWORD type is caught here so we don't fail in mysterious ways when keyword-only forcing was forgotten somewhere, possibly in a base class.

test/mitmproxy/addons/test_dumper.py
cortesi: Done.

mitmproxy/optmanager.py
```diff
@@ -112,6 +168,81 @@ def toggle():
...
+            Has the option changed from the default?
+        """
+        if getattr(self, option) != self._defaults[option]:
+            return True
+
+    def save(self, path, defaults=False):
+        """
+            Save to path. If the destination file exists, modify it in-place.
+        """
+        if os.path.exists(path) and os.path.isfile(path):
```
cortesi: Done

mitmproxy/optmanager.py
```diff
@@ -112,6 +168,81 @@ def toggle():
...
+        return data
+
+    def load(self, text):
+        """
+            Load configuration from text, over-writing options already set in
+            this object. May raise OptionsError if the config file is invalid.
+        """
+        data = self._load(text)
+        for k, v in data.items():
+            setattr(self, k, v)
```
cortesi: We can actually use update here. Good catch and done.

mitmproxy/optmanager.py
```diff
@@ -112,6 +168,81 @@ def toggle():
...
+    def load_paths(self, *paths):
+        """
+            Load paths in order. Each path takes precedence over the previous
+            path. Paths that don't exist are ignored, errors raise an
+            OptionsError.
+        """
+        for p in paths:
+            p = os.path.expanduser(p)
+            if os.path.exists(p) and os.path.isfile(p):
+                txt = open(p, "r").read()
```
cortesi: Done.

mitmproxy/tools/cmdline.py
```diff
@@ -514,56 +513,55 @@ def proxy_ssl_options(parser):
              'as the first entry. Can be passed multiple times.')
     group.add_argument(
         "--ciphers-client", action="store",
-        type=str, dest="ciphers_client", default=options.DEFAULT_CLIENT_CIPHERS,
```
cortesi: Fixed

PR #1807, opened 2016-12-02
Force not verify server certificate if client doesn't send SNI

deadblue:
If _client_hello.sni is None, mitmproxy need get certificate subject from upstream server. But _establish_tls_with_server will be failed if server_sni (whose value gets from _client_hello.sni) is None and config.openssl_verification_mode_server is SSL.VERIFY_PEER.
That causes client who doesn't sent SNI can't connect to server through mitmproxy.

### DIFF
--- a/mitmproxy/proxy/protocol/tls.py
+++ b/mitmproxy/proxy/protocol/tls.py
@@ -2,6 +2,8 @@ import struct
 from typing import Optional  # noqa
 from typing import Union
 
+from OpenSSL import SSL
+
 import construct
 from mitmproxy import exceptions
 from mitmproxy.contrib.tls import _constructs
@@ -522,12 +524,13 @@ class TlsLayer(base.Layer):
                         ciphers_server.append(CIPHER_ID_NAME_MAP[id])
                 ciphers_server = ':'.join(ciphers_server)
 
+            verify_options = self.config.openssl_verification_mode_server if self.server_sni else SSL.VERIFY_NONE
             self.server_conn.establish_ssl(
                 self.config.clientcerts,
                 self.server_sni,
                 method=self.config.openssl_method_server,
                 options=self.config.openssl_options_server,
-                verify_options=self.config.openssl_verification_mode_server,
+                verify_options=verify_options,
                 ca_path=self.config.options.ssl_verify_upstream_trusted_cadir,
                 ca_pemfile=self.config.options.ssl_verify_upstream_trusted_ca,
                 cipher_list=ciphers_server,


Kriechi:
Thanks for the PR @deadblue!
Can you please add a test for this new behaviour?

deadblue:
@mhils Yes, use `--insecure`  can also do that. But I know I can use `--insecure` only after I read the code and find why this issue happened.
On my android device with os version 4.4.4, all app write in Java don't sent SNI, I think that is not an individual case.

@Kriechi Sorry for late reply. 
I find this issue in the case I write above, and I write a simple code to reproduce that.
1. Run `mitmproxy --host --port 9999` to start mitmproxy in normal proxy mode.
2. Run code below:
```
import urllib2
import ssl

class SSLContextWrapper:

    def __init__(self, cafile):
        self.ssl_context = ssl.create_default_context(cafile=cafile)
        self.ssl_context.check_hostname = False

    def wrap_socket(self, sock, server_side=False,
                    do_handshake_on_connect=True,
                    suppress_ragged_eofs=True,
                    server_hostname=None):
        # force not send server_hostname
        return self.ssl_context.wrap_socket(
            sock, server_side,
            do_handshake_on_connect, suppress_ragged_eofs, None
        )

if __name__ == '__main__':
    mitmproxy_cafile = '/Users/deadblue/.mitmproxy/mitmproxy-ca-cert.pem'
    opener = urllib2.build_opener(
        urllib2.ProxyHandler({'https':'127.0.0.1:9999'}),
        urllib2.HTTPSHandler(context=SSLContextWrapper(mitmproxy_cafile))
    )
    resp = opener.open('https://httpbin.org/get')
    print resp.read()
```
3. You will get `HTTP Error 502` which return by mitmproxy.
4. After modify tls.py, you will get correct response.

mhils:
> But I know I can use --insecure only after I read the code and find why this issue happened.

mitmproxy should warn you in case we close a connection due to an invalid certificate. It looks like this:

image

If that is not happening, please file a separate bug report. :)


> I think that is not an individual case.

That's true. Nonetheless, we don't want to have insecure-by-default behaviour.

If you implement this as an optional feature switch, we'd love to have a test similar to this: https://github.com/mitmproxy/mitmproxy/blob/1fcf79fff049859e97559a19264203b89eb42775/test/mitmproxy/test_server.py#L386-L462 :smiley:

mhils:
I'm closing this for now. Feel free to open a new PR for an additional switch, but we don't want this as default behaviour.

Thanks!
Max

PR #1831, opened 2016-12-09
Split contentviews.py into mitmproxy.contentviews

### LINKED ISSUE #845: HTML content view should not "fix" the HTML
I used mitmproxy to analyze an API that an internet radio device interacts with. The server responded with sort-of-XML data, but a `content-type: text/html` header.

mitmproxy used that header to switch to HTML response display mode, lowercasing all XML tags and adding a fsckng doctype header to the output.

I did not know that mitmproxy tampered with the data I saw and wondered for days why my API replication did not work with the device.

Please please please:
1. switch to raw mode by default, or at least provide a parameter for this
2. do not add a DOCTYPE line to the responses of html. please. Do not add additional data that were not in the original response.

### LINKED ISSUE #1444: Wait for lxml release with wheels
For the next mitmproxy release, we want to recommend Python 3.5 by default. While this is not ultimately blocking, it would be ideal to have a lxml release with Python 3.5 wheels on PyPI for this.  I did some work to make that happen in https://github.com/lxml/lxml/pull/200, so this hopefully can be resolved rather soon.

### LINKED ISSUE #1657: Remove lxml dependency
Installing lxml is more or less painful on all the platforms we support. In particular, I tried for months to get the lxml maintainers to provide wheels for Windows, but to no avail. 
We only use lxml for indenting the XML/HTML content views [here](https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/contentviews.py). It would be interesting to see if we can just replace that with a more naive implementation that does indentation, so that we don't have to carry lxml with us.

As an additional benefit, doing this in Python instead of C should make the whole thing considerably safer.

refs #845

mhils:
As a preparation for ditching lxml, this PR splits up mitmproxy/contentviews.py into ~ one file per view. Thinking forward, each view could actually be a separate addon in the future, but let's keep this minimally invasive for now.

The `mitmproxy.contenviews` API remains unchanged.

fixes #845
fixes #1657
fixes #1444


mitmproxy/contentviews/__init__.py
```diff
@@ -0,0 +1,177 @@
...
+add(html.ViewHTMLOutline())
+add(javascript.ViewJavaScript())
+add(css.ViewCSS())
+add(urlencoded.ViewURLEncoded())
+add(multipart.ViewMultipart())
+add(image.ViewImage())
+add(query.ViewQuery())
+
+if protobuf.ViewProtobuf.is_available():
+    add(protobuf.ViewProtobuf())
```
dwfreed: Missing trailing newline

mhils:
This is ready for review! :smiley:

Kriechi:
Does this PR fix / close / invalidate any of: #845, #1657, or #1444?

mhils:
Yes, yes and yes! ☺️

mitmproxy/contentviews/base.py
```diff
@@ -0,0 +1,65 @@
...
+from typing import Iterable, AnyStr, List
+from typing import Mapping
+from typing import Tuple
+
+VIEW_CUTOFF = 512
+
+KEY_MAX = 30
+
+
+class View:
```
cortesi: Presumably not a refactoring you're going to be doing in this patch, but we need a clear interface for Views. I suggest that each view exposes two class methods: 

- ```check(f) -> Optional[View]```, which takes an object (flow, request, response), does whatever verification is needed to see whether it can sensibly render it, and returns an instantiated class.
- ```force(f) -> Optional[View]```, which tries to force the view to render the object, regardless of whether the sanity-check succeeded. This is still optional, because for some things the answer will still be "no". 

This means that each view can now have a natural set of instantiation parameters, rather than the nebulous and poorly defined "metadata" cludge. We can also use the check() method in places like mitmproxy console to only display a list of views that "make sense", cutting down on the very long list of elements we currently have. 

This gets rid of all the arcane mechanisms we currently have like AutoView and centralised content guessing, instead letting each View define their own semantics. When automatically choosing a View, the ViewManager can use the first matching view, and we can use view order to control the priorities.

cortesi:
Let's merge and then think about the refactorings I muse about in the review.

PR #1836, opened 2016-12-10
docs: update installation

Kriechi:


mhils:
Looks great. Minor nitpick: My understanding is that 
```
You can check your system information by running: ``mitmproxy --sysinfo``
```
will never work if mitmproxy hasn't been installed properly and is not required at this point if installation was successful. Long story short, this part seems to be superfluous here. :)

Kriechi:
> will never work if mitmproxy hasn't been installed properly

Thats the idea behind this sentence - to test if mitmproxy works properly. I guess I can reword it to better reflect the actual meaning. Any suggestions?

Kriechi:
I made the requested changes.

PR #1842, opened 2016-12-11
Save options to file from console & related work

cortesi:


mitmproxy/tools/console/master.py
```diff
@@ -386,17 +386,11 @@ def view_flow(self, flow, tab_offset=0):
...
-        path = os.path.expanduser(path)
-        try:
-            f = open(path, "wb")
-            fw = io.FlowWriter(f)
-            for i in flows:
-                fw.add(i)
-            f.close()
-        except IOError as v:
-            signals.status_message.send(message=v.strerror)
+        f = open(path, "wb")
```
Kriechi: We should use `with open(...) as f:` more.

PR #1845, opened 2016-12-11
Mitmweb Improvements

mhils:
This PR brings the mitmweb UI to a state where it actually looks good! As an added bonus, we fix a whole bunch of bugs. :tada: :smiley:

| Before                                                                                  | After                                                                                  |
|-----------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------|
| options before   | options after   |
| flow menu before | flow menu after |


mitmproxy/flow.py
```diff
@@ -169,6 +169,8 @@ def kill(self, master):
             self.reply.take()
         self.reply.kill(force=True)
         self.reply.commit()
+        self.live = False
+        master.addons("kill", self)
```
cortesi: I'm not against adding a kill event, but we need to do it properly, please. That means listing it in events.py, adding docs for it, and so forth.

cortesi: We need to think harder about the event flow here before we move further. This call and the calls to the intercept and resume events below invoke the addon handlers without first establishing the addon context. That's only reasonable if we always invoke these methods from within addon handlers that already have an event context. Looking at calls to, e.g. resume, it doesn't look like this is always the case.

The ideal solution is to have these events triggered from the proxy side just like all other events. That is, after a flow has been killed, resumed or intercepted, the proxy uses a normal .ask() or .tell() call into the addons mechanism. This also resolves some sequencing issues that can arise with nested addon handler calls.

cortesi:
I'm merging this so we can move on, but we need to address the event issue before we release.

PR #1863, opened 2016-12-16
disable h2c upgrades

Kriechi:


cortesi:
Merging, with some reservations about this being an addon. Let's revisit after the release.

PR #1868, opened 2016-12-17
Update send_reply_from_proxy.py

niasand:
Update the code to match the 0.18.2 verision 
old:
```python
from mitmproxy import http
```

new:
```python
from mitmproxy.models import http
```


mhils:
Thanks. If you are looking for the examples for 0.18, please take a look at the 0.18 branch. The examples on master are intended to work with master.

PR #1881, opened 2016-12-20
Always Decode ALPN Where Required

mhils:


mitmproxy/tools/web/app.py
```diff
@@ -34,6 +34,12 @@ def flow_to_json(flow: mitmproxy.flow.Flow) -> dict:
         "type": flow.type,
         "modified": flow.modified(),
     }
+    # .alpn_proto_negotiated is bytes, we need to decode that.
+    for conn in "client_conn", "server_conn":
+        if f[conn]["alpn_proto_negotiated"] is None:
+            continue
+        f[conn]["alpn_proto_negotiated"] = \
+            f[conn]["alpn_proto_negotiated"].decode(errors="backslashreplace")
```
Kriechi: Should we move the decode part directly to the actual first-use of this? Or how does this affect if the value gets decoded and we need to get the bytes back later?

mhils: If you look in the ALPN RFC they state that the protocol identifier may be utf8, but it is only specified as a bytestring. My understanding is that it could be (in theory, at least) a random sequence of bytes, which we cannot necessarily decode.

mhils: (backslashreplace is not bijective)

Kriechi: yes - but here you are overwriting the alpn information with its decoded value. This could potentially be lossy. Or what kind of data structure is used here?

mhils: This is only what is sent to the web UI (which only reads it, never writes it), the actual flow remains untouched. I think that's okay to not be overprecise here? A non-ascii protocol would be suspicious by itself.

PR #1887, opened 2016-12-26
enable HTTP/2 by default

Kriechi:


mhils:
LGTML. Waiting with this until we ship a 1.0.1 with Python 3.6 support.

PR #1902, opened 2016-12-30
fix bug when parse multipart

bestswifter:
When using mitmproxy, I found a bug which may occur when parsing multipart data. Here is my solutioin, as shown below, I changed 3 lines.

In the first line, which leads to the bug, the origin code use `splitlines` method. However, if the data contains data like `\n` or `\r\n`, it will also be discarded. Unfortunately, many protobuf message begins with `\n`.

According to the RFC file, there must be a `\r\n\r\n` between boundary and data body, so I use this to split the string. Then seconde parameter `2` makes sure that data body won't be splitted.

Since this will change the structure of `parts`, now we need to search first element of `parts` to get the key.

Since there is another `\r\n` at the end of `parts[1]`, I just ignore them.

### DIFF
--- a/mitmproxy/net/http/multipart.py
+++ b/mitmproxy/net/http/multipart.py
@@ -21,12 +21,12 @@ def decode(hdrs, content):
         r = []
 
         for i in content.split(b"--" + boundary):
-            parts = i.splitlines()
+            parts = i.split(b'\r\n\r\n', 2)
             if len(parts) > 1 and parts[0][0:2] != b"--":
-                match = rx.search(parts[1])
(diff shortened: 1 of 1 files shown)


mhils:
Hi,

Thanks for the PR! In order for this to be merge, could you please (1) fix the existing tests and (2) add a simple test where the previous implementation failed but this one works as expected?

Thanks!
Max

ujjwal96:
Not able to reproduce this. But I experienced a different issue. When I scrolled through the data randomly help would open and won't go until pressed 'q' a lot of times or kept it pressed :confused:

Issue #1623, opened 2016-10-17
Better control of long output in mitmdump


cortesi:
Comments re #1621. Being able to see the full URL in the terminal is too valuable to lose entirely. A better version of this would:

• Add a command-line flag to turn on full-width output (a la ``ps --ww`).
• Use the terminal width to determine the cutoff. See `click.get_terminal_size()`. 
• Maybe make the cutoff marker clearer by making it a different colour.

mhils:
> Add a command-line flag to turn on full-width output

Would you like to have a separate switch for that or shall we just make `-d` do that as well? I'm not sure if a dedicated switch would be relevant to a large enough share of users to justify it. I'm not particularly strong on this though.

cortesi:
I don't mind, as long as there's a knob for this.  We already have a bit of a case of flag-itis, so just using -d might be the way to go.  Showing full URLs could be detail level #1.

kalyan-kumar:
Shouldn't this issue be closed now?

mhils:
yes, thanks!

PR #1903, opened 2016-12-30
Controls the length of URL displayed using the -d option

lymanZerga11:
Done with reference to #1623. Controls the length of URL displayed using the `-d` or `flow_detail` option. The full length URL is displayed if `flow_detail > 0`, else it is shortened.
Also uses the terminal width obtained from `click.get_terminal_size()` to determine the cutoff length.

### DIFF
--- a/mitmproxy/addons/dumper.py
+++ b/mitmproxy/addons/dumper.py
@@ -125,7 +125,7 @@ class Dumper:
             url = flow.request.pretty_url
         else:
             url = flow.request.url
-        terminalWidthLimit = shutil.get_terminal_size()[0] - 25
+        terminalWidthLimit = max(shutil.get_terminal_size()[0] - 25 , 50)
         if self.flow_detail < 1 and len(url) > terminalWidthLimit:
             url = url[:terminalWidthLimit] + "…"
         url = click.style(strutils.escape_control_characters(url), bold=True)
--- a/mitmproxy/addons/dumper.py
+++ b/mitmproxy/addons/dumper.py
@@ -2,6 +2,7 @@ import itertools
 import sys
 
 import click
+import shutil
 
 import typing  # noqa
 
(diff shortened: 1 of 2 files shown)


mhils:
Thanks!

I was wondering how that behaves when we're not in a terminal, and then found that click internally just uses [`shutil.get_terminal_size`](https://docs.python.org/3.6/library/shutil.html#shutil.get_terminal_size) on Python 3.5+, which has an explicit fallback option. Can we please just use that so that the behaviour becomes clear?

Also, as mundane as it is, this needs a test. 😃 
https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/addons/test_dumper.py

mitmproxy/addons/dumper.py
```diff
@@ -124,6 +125,9 @@ def _echo_request_line(self, flow):
             url = flow.request.pretty_url
         else:
             url = flow.request.url
+        terminalWidthLimit = shutil.get_terminal_size()[0] - 25
```
mhils: Rare case, but what happens if `get_terminal_size()[0]` is < 25? Why do we substract 25 here at all?

lymanZerga11: Oh yes. I missed that case. I subtracted 25 because I thought the idea was to fit the output into one line. We already have `GET`, `HTTP/x.x` and ` PUSH_PROMISE`, so in the worst case we will need 25 places to fit them all into the line.  

I think it would be better to remove the `- 25` and allow it to occupy two lines in some cases. 
Or maybe `terminalWidthLimit = shutil.get_terminal_size()[0] // 2`

mhils: How about a simple `max(shutil.get_terminal_size()[0] - 25, 50)` ? This way we can assure that we have at least 50 characters of the URL. 😃

lymanZerga11: Yeah that makes sense. Haha

mhils:
Thanks! 😃

PR #1905, opened 2016-12-30
Allow underscore in hostname

### LINKED ISSUE #1529: Gracefully handle underscores in hostnames
We need to handle invalid hostnames more gracefully. See the discussion here:

https://discourse.mitmproxy.org/t/mitmproxy-not-working-with-urls-which-have-under-scores/
1. mitmdump -vv
2. env http_proxy=http://localhost:8080 curl "http://foo_bar.com"

My inclination here is to make the host matching more liberal, and pass this sort of error upstream. Basically, if I'm using mitmproxy in an analysis situation, and the client is exhibiting this behaviour, I want to see what the server does with it rather than see an error.

Either way, we certainly shouldn't be spewing a user-visible traceback to the logs for this kind of protocol error.

chhsiao90:
Resolved #1529 

Update the regular expression to allow underscore in hostname.
The user-visible traceback mentioned at the issue seems to be fixed.

Though underscore is not recommended be part of hostname,
not sure that to display some warning message about it will help or make it better?

### DIFF
--- a/mitmproxy/net/check.py
+++ b/mitmproxy/net/check.py
@@ -1,6 +1,7 @@
 import re
 
-_label_valid = re.compile(b"(?!-)[A-Z\d-]{1,63}(?<!-)$", re.IGNORECASE)
+# Allow underscore in host name
+_label_valid = re.compile(b"(?!-)[A-Z\d\-_]{1,63}(?<!-)$", re.IGNORECASE)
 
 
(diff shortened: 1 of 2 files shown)


mhils:
Thanks! 😄 
There are too many sites out there with underscores in the hostname (e.g. nytimes.com has a bunch of subdomains), so I think just accepting what is common in practice is fine.

Issue #1828, opened 2016-12-09
How to configure file system watchdog (away) for attached scripts


RobbBienert:
Hi folks,

I upgraded my local mitmproxy using pip from 0.17 to 0.18.2 and now my script does not work anymore. The mitmproxy is used to connect a local machine with a corporate proxy and measure the traffic in between. The mitmproxy (to be exact, the mitmdump program) is called like this:

    /usr/local/bin/mitmdump -p 9090 -U 'http://proxy:8080' -s 'tracer.py argument'

The `tracer.py` has three functions implemented as described in the [Events Documentation](http://docs.mitmproxy.org/en/latest/scripting/events.html):

    def start():
        # should be called once on startup
        # so do some initialization here
    
    def response(flow):
        # track the response information
    
    def done():
        # called at script shutdown

In the stdout I see a lot of lines telling me `Reloading script: tracer.py` and if I add a print('hello from start()') to the `start()` function I also see this message in stdout, which means that `start()` gets executed more than once. My search for `Reloading script` pointed me to [the function `tick` in addons/script.py](https://github.com/mitmproxy/mitmproxy/blob/6792cc1de90b02b89da9ae80704ea237c5486e66/mitmproxy/addons/script.py#L185), which seems to be running in a certain interval. The problem I can address so far here is that what happens when `self.should_reload.is_set()` returns `True` while the proxy is running.

`should_reload`s internal state is set to `False` by `clear()` in line 187, but it could be set to `True` again by `reload()` in line 175. `reload` is passed as callback to a `ReloadHandler` in line 204, which is connected to a file system watchdog. This sounds similar to the [Scripting Overview](http://docs.mitmproxy.org/en/latest/scripting/overview.html#developing-scripts),

> Mitmprxoy monitors scripts for modifications, and reloads them on change. When this happens, the script is shut down (the done event is called), and the new instance is started up as if the script had just been loaded (the start and configure events are called).

with the exception, that the `ReloadHandler` reloads the script if _anything_ in the observed directory is changed. So I propose not only to filter directories and files starting with a `.`, but to watch _only_ for files ending with `.py`.

##### System information

The output of "mitmdump --sysinfo":

Mitmproxy version: 0.18.2
Python version: 2.7.12
Platform: Linux-4.4.0-42-generic-x86_64-with-LinuxMint-18-sarah
SSL version: OpenSSL 1.0.2g  1 Mar 2016
Linux distro: LinuxMint 18 sarah

[...]

PR #1911, opened 2017-01-04
fixing issue #1828

RobbBienert:


mhils:
Thanks! Could you adjust the tests as well please? 😃

RobbBienert:
If you can tell me what the errors mean we can talk about it ...

mhils:
If you look at https://ci.appveyor.com/project/mhils/mitmproxy/build/3791#L1278, it should be clear where the errors come from. The offending test is `test/mitmproxy/addons/test_script.py::test_reloadhandler` 😃

mhils:
Any update on this? 😃

RobbBienert:
Should the method return `True` by default? The build systems messages are hardly to understand for a newbie to it.

mhils:
This is the test in question: https://github.com/mitmproxy/mitmproxy/blob/5d0f7e5c41f3c553b993162abaef32c44ac22781/test/mitmproxy/addons/test_script.py#L43-L55

We previously reloaded on all files (and this has been tested). We now need to adjust the test so that it only expects reloads for files ending with `.py`.

RobbBienert:
Hope this commit works, because neither my fork/checkout nor the local python installation contains any py.test file (Linux Mint Sarah).

mhils:
Sorry for the late response - I think we can keep the `on_created` handler just like the `on_changed` one, i.e. trigger on new `.py` files. 😃

PR #1918, opened 2017-01-06
Update instructions in README

s4chin:
I'm getting started on this project again and could not install with the instructions in the "Hacking" section. This PR updates the instructions which worked for me on Ubuntu 16.04.


mhils:
Thanks! We need to merge that with the install docs at some point, but this should be good enough for now.

PR #1926, opened 2017-01-13
reSt verbatim formatting error

tripleee:
The rendered code snippets would be rewrapped as if it were prose, which wreaks havoc with the code.


mhils:
Thanks! :smiley:

Issue #1909, opened 2017-01-03
crashed using invalid script path


wbwkx:
##### Steps to reproduce the problem:

1.start proxy from cli
2.using "|" command to load a script and set an invalid path to it. like "~/Desktop/"
3.hit Enter it will crash

##### System information

```
demo » mitmproxy --version
mitmproxy 2.0.0
```

```
demo » mitmproxy --sysinfo
Mitmproxy version: 2.0.0
Python version: 3.5.0
Platform: Darwin-16.3.0-x86_64-i386-64bit
SSL version: OpenSSL 1.0.2j  26 Sep 2016
Mac version: 10.12.2 ('', '', '') x86_64
```

##### Crash-Log trace

```
   Traceback (most recent call last):
  File "/usr/local/var/pyenv/versions/3.5.0/lib/python3.5/site-packages/mitmproxy-2.0.0-py3.5.egg/mitmproxy/tools/console/master.py", line 292, in run
    self.loop.run()
  File "/usr/local/var/pyenv/versions/3.5.0/lib/python3.5/site-packages/urwid/main_loop.py", line 278, in run
    self._run()
  File "/usr/local/var/pyenv/versions/3.5.0/lib/python3.5/site-packages/urwid/main_loop.py", line 376, in _run
    self.event_loop.run()
  File "/usr/local/var/pyenv/versions/3.5.0/lib/python3.5/site-packages/urwid/main_loop.py", line 682, in run
    self._loop()
  File "/usr/local/var/pyenv/versions/3.5.0/lib/python3.5/site-packages/urwid/main_loop.py", line 719, in _loop
    self._watch_files[fd]()
  File "/usr/local/var/pyenv/versions/3.5.0/lib/python3.5/site-packages/urwid/raw_display.py", line 393, in <lambda>
    event_loop, callback, self.get_available_raw_input())
  File "/usr/local/var/pyenv/versions/3.5.0/lib/python3.5/site-packages/urwid/raw_display.py", line 493, in parse_input
    callback(processed, processed_codes)
  File "/usr/local/var/pyenv/versions/3.5.0/lib/python3.5/site-packages/urwid/main_loop.py", line 403, in _update
    self.process_input(keys)
  File "/usr/local/var/pyenv/versions/3.5.0/lib/python3.5/site-packages/urwid/main_loop.py", line 503, in process_input
    k = self._topmost_widget.keypress(self.screen_size, k)
[...]

PR #1929, opened 2017-01-15
Add except clause to catch script parsing errors

lymanZerga11:
Done in reference to #1909 Add an except clause in `run_once` function when calling the `Script` class constructor. Returning an empty `Script` object seems more of a hassle than returning `None`, given that the returned object is never used.

### DIFF
--- a/mitmproxy/addons/script.py
+++ b/mitmproxy/addons/script.py
@@ -252,7 +252,7 @@ class ScriptLoader:
                     try:
                         sc = Script(s)
                     except ValueError as e:
-                        raise ValueError(str(e))
+                        raise exceptions.OptionsError(str(e))
                     ordered.append(sc)
                     newscripts.append(sc)
 
--- a/mitmproxy/addons/script.py
+++ b/mitmproxy/addons/script.py
@@ -208,7 +208,7 @@ class ScriptLoader:
         try:
             sc = Script(command)
         except ValueError as e:
-            raise exceptions.OptionsError(str(e))
+            raise ValueError(str(e))
         sc.load_script()
         for f in flows:
             for evt, o in events.event_sequence(f):
@@ -252,7 +252,7 @@ class ScriptLoader:
                     try:
                         sc = Script(s)
                     except ValueError as e:
-                        raise exceptions.OptionsError(str(e))
+                        raise ValueError(str(e))
                     ordered.append(sc)
                     newscripts.append(sc)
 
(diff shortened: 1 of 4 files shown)


mhils:
Thanks! I appreciate the effort. Looking at the semantics of this, I think an invalid script path is not a script error, but rather a user input error. Can we 
- make `parse_command` raise ValueErrors instead of OptionErrors and add this to the docstring.
- Use `try: sc = Script(s); except ValueError as e: raise OptionError(str(e))` [here](https://github.com/mitmproxy/mitmproxy/blob/727abdba44fa72321f51402b03e9d39c6489dc95/mitmproxy/addons/script.py#L249)
- Document that `run_once` may raise a ValueError.
- Catch said ValueError in the console master [here](https://github.com/mitmproxy/mitmproxy/blob/b39380b00fe324a67c107ac2c4a611cbd71b7839/mitmproxy/tools/console/master.py#L146)?
- Have a test for `ConsoleMaster.run_script_once` - I believe the `except mitmproxy.exceptions.AddonError` in there can never happen essentially and can be removed?

Thanks!
Max

lymanZerga11:
@mhils Thanks for the review. I'll look into the changes over the weekend. Been a bit busy the past few days.

lymanZerga11:
@mhils I have made the changes and added the test using `mock`. Is there a better way to write the tests? Currently, only the xcode build on Travis is failing. Can you help me check why?

mitmproxy/addons/script.py
```diff
@@ -205,7 +205,10 @@ class ScriptLoader:
         An addon that manages loading scripts from options.
     """
     def run_once(self, command, flows):
-        sc = Script(command)
+        try:
+            sc = Script(command)
+        except ValueError as e:
```
mhils: Why do you cast this to an OptionsError? Options are not involved here, we just should let the ValueError propagate and document that.

test/mitmproxy/console/test_master.py
```diff
@@ -37,6 +46,13 @@ def test_basic(self):
             self.dummy_cycle(m, 1, b"")
             assert len(m.view) == i
 
+    @mock.patch('mitmproxy.tools.console.signals.add_log', side_effect=mock_add_log)
+    def test_run_script_once(self, test_func):
+        m = self.mkmaster()
+        f = tflow.tflow(resp=True)
+        with mitmproxy.test.tutils.raises(ScriptError):
```
mhils: Have you considered just running `m.run_script_once("nonexistent", [f])` and then checking `m.logbuffer` ?

lymanZerga11: Done!

mitmproxy/addons/script.py
lymanZerga11: Sorry! Got a bit confused.

PR #1934, opened 2017-01-17
Added servercachereplay addon

Rastii:
This addon will automatically replay known server responses
without having to save/load flows.  It uses the same
whitelisting/blacklisting options of serverplayback addon,
i.e, filtering out payload params, including headers, etc...

This addon can be used either through the URWID GUI or
through CLI arguments.  The argument to enable
the addon is "--server-cache-replay" and the argument
to load saved files is "--server-cache-replay-load".
Note that --server-cache-replay and --server-replay are
mutually exclusive options.

For the GUI, an additional menu is presented when the
user presses the "S" key where he/she will decide
whether they want to use the server replay addon or the
new servercachereplay addon.

### DIFF
--- a/mitmproxy/addons/__init__.py
+++ b/mitmproxy/addons/__init__.py
@@ -8,6 +8,7 @@ from mitmproxy.addons import replace
 from mitmproxy.addons import script
 from mitmproxy.addons import setheaders
 from mitmproxy.addons import serverplayback
+from mitmproxy.addons import servercacheplayback
 from mitmproxy.addons import stickyauth
 from mitmproxy.addons import stickycookie
 from mitmproxy.addons import streambodies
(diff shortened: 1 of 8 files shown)


mhils:
Thanks for the great PR!

I like the idea of what you are doing here, but this introduces a lot of new functionality. Do you think we could integrate this functionality into the existing server replay addon?

Rastii:
Thank you!

Of course, I'll integrate it into the existing serverreplay addon -- perhaps when I get some free time this weekend.

Kriechi:
@Rastii any progress or update?

Rastii:
I'd like to wrap this up and get it merged, sorry for the delay!

One thing though, is the current PR keybinding for the GUI acceptable?
That is S1 -> (S2, C) where S1 is the option for server replay and S2 / C is the selection between caching or loading the available flows.  This would add an extra step to the previous behavior: S -> S -> (A|T).

It's not really backwards compatible, but the only thing we'd be breaking is human muscle memory!

Thanks for the input, and again, apologies for the incredibly long delay!

cortesi:
@rastii We're quite interested in this functionality, but unfortunately this patch has now diverged from master far enough that I'm going to close the PR. On the upside, doing this within the new addon system would be much nicer. Please let us know if you still plan to work on this.

PR #1941, opened 2017-01-19
provide git information with --version

### LINKED ISSUE #1848: mitmproxy sysinfo: show git tag/ref/commit
It would be nice if `mitmproxy --sysinfo` would also display the current git tag/ref/commit when running from source. Currently only `0.19` is displayed, which makes it a bit undeterministic what code is actually running. We can probably reuse some code from rtool here.

Kriechi:
fixes #1848

### DIFF
--- a/mitmproxy/utils/debug.py
+++ b/mitmproxy/utils/debug.py
@@ -5,18 +5,29 @@ import threading
 import signal
 import platform
 import traceback
+import subprocess
 
 from mitmproxy import version
+from mitmproxy import utils
 
 from OpenSSL import SSL
 
 
-def sysinfo():
+def dump_system_info():
+    git_describe = 'release version'
+    with utils.chdir(os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))):
+        try:
+            c = ['git', 'describe', '--tags', '--long']
+            git_describe = subprocess.check_output(c, stderr=subprocess.STDOUT)
+            git_describe = git_describe.decode().strip()
+        except:
+            pass
+
     data = [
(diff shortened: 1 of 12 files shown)


mitmproxy/utils/debug.py
```diff
@@ -5,18 +5,24 @@
...
 
 from mitmproxy import version
 
 from OpenSSL import SSL
 
 
-def sysinfo():
+def dump_system_info():
+    old_dir = os.getcwd()
+    os.chdir(os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
```
mhils: I think we can re-use this: https://github.com/mitmproxy/mitmproxy/blob/490872ebefd87e9a55f2db3dd18c0fbb58dbf206/mitmproxy/test/tutils.py#L22

mitmproxy/utils/debug.py
```diff
@@ -5,18 +5,24 @@
...
+    git_describe = subprocess.check_output(['git', 'describe', '--tags', '--long']).decode().strip()
```
mhils: Let's wrap this in a catch-all try-except block. Users may not have git installed, etc. etc. If it fails, we can still display everything else.

mitmproxy/utils/debug.py
Kriechi: Where would you put that function if we want to move it from `test/` into `mitmproxy/`?

mhils: mitmproxy/utils?

Kriechi: Done.

PR #1944, opened 2017-01-19
Fix SNI invert ignore

yotams:
When using an invert ignore expression on a hostname (e.g.
'?!(example\.com)'), all TLS connections will be ignored as the SNI
hostname is not yet parsed. With this fix, we will always try to match
against the SNI hostname.

### DIFF
--- a/mitmproxy/proxy/root_context.py
+++ b/mitmproxy/proxy/root_context.py
@@ -50,7 +50,7 @@ class RootContext:
         # 1. check for --ignore
         if self.config.check_ignore:
             ignore = self.config.check_ignore(top_layer.server_conn.address)
-            if not ignore and client_tls:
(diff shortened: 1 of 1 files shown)


mhils:
Thanks for the PR!

I do think this is a very good use case, but we need to be careful with changing this. In its current form this PR breaks IP-based filtering. The current logic is a follows:
```
match(ip) or (clienthello_parseable and match(sni))
```
This PR changes it to 
```
match(sni) if clienthello_parseable else match(ip)
```

This has a bunch of issues, namely that `--ignore 17\.178\.96\.59:443` won't match anymore as one would expect it to (this example is currently documented behaviour) and that we ignore that fact that we may have a valid ClientHello, but no SNI. Thoughts?

PR #1945, opened 2017-01-19
fix #939

### LINKED ISSUE #939: Check for expired CA certificate

Kriechi:
- [x] add mitmproxy warning
  - [x] add mitmweb warning

### DIFF
--- a/mitmproxy/addons/__init__.py
+++ b/mitmproxy/addons/__init__.py
@@ -1,35 +1,39 @@
 from mitmproxy.addons import anticache
 from mitmproxy.addons import anticomp
+from mitmproxy.addons import check_alpn
+from mitmproxy.addons import check_ca
 from mitmproxy.addons import clientplayback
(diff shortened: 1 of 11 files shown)


mitmproxy/master.py
```diff
@@ -118,6 +119,37 @@ def shutdown(self):
         self.should_exit.set()
         self.addons.done()
 
+    def is_alpn_missing(self, level='error'):
```
mhils: Can we move this into an addon that just has a `start` function to do the checks and logging?

Issue #1915, opened 2017-01-05
[web]: body filter expressions are not working


wdormann:
##### Steps to reproduce the problem:

1. Run mitmweb
2. In one of the "Start" boxes, start to type each of the filter expressions.   e.g. "~b asdf"

##### Any other comments? What have you tried so far?

I realize that mitmweb is beta, but the documentation indicates "We consider it stable for all features currently exposed in the UI", so maybe this is an oversight.  Filter expressions that cause the error: "SyntaxError: Expected filter expression but "~" found" include:
~b
~bq
~bs
~dst
~http
~marked
~src

##### System information

Mitmproxy version: 1.0.1
Python version: 3.5.2
Platform: Linux-3.13.0-100-generic-i686-with-debian-jessie-sid
SSL version: OpenSSL 1.0.2g  1 Mar 2016
Linux distro: debian jessie/sid

mhils:
Thanks! It seems like these got out of sync, this needs to be fixed.

s4chin:
Adding the missing filters as reference:
- [ ] `~b`
- [ ] `~bq`
- [ ] `~bs`
- [ ] `~dst`
- [ ] `~http`
- [x] `~marked`
- [ ] `~src`
- [ ] `~tcp`

mhils:
Thanks. As mentioned on Slack, let's keep the body filters out for now and do the others first. :)

keymandll:
Gents, so no chance I could get ~bs filter working still?

mhils:
Unfortunately it's not really trivial as we do not transmit body contents to the front end - see https://github.com/mitmproxy/mitmweb/issues/10.

I'm closing this here as this issue is actually tracked in https://github.com/mitmproxy/mitmweb/issues/10!

pravic:
I am not sure whether I should open a new issue here, in mitmproxy/mitmweb or do not open it at all (because of the comment above).

Anyway, right now an attempt to use a body filter will navigate browser to another location with an empty page.

To rephrase, 

1. Open some flow
2. Click on Highlight field, type `~b` (you will get an error about an expected filter expression)
3. Try to type something, e.g. `~b body`. You will immediately be redirected to `http://127.0.0.1:8081/#/flows?h=~b%20s` which is completely empty [*].
4. The back button (in browser) does not work, 

[*] Well, not really empty but it looks empty. The page source is

```html
<!DOCTYPE html>
<html>
<head lang="en">
<meta charset="UTF-8">
[...]

PR #1947, opened 2017-01-20
web: Add ~http and ~tcp filters

s4chin:
As in #1915

### DIFF
--- a/web/src/js/filt/filt.peg
+++ b/web/src/js/filt/filt.peg
@@ -109,6 +109,10 @@ function responseHeader(regex){
     responseHeaderFilter.desc = "resp. header matches " + regex;
     return responseHeaderFilter;
 }
+function httpFilter(flow){
+    return flow.type === "http";
+}
+httpFilter.desc = "is an HTTP Flow";
 function method(regex){
     regex = new RegExp(regex, "i");
     function methodFilter(flow){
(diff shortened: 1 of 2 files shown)


mitmproxy/tools/web/app.py
```diff
@@ -45,6 +46,7 @@ def flow_to_json(flow: mitmproxy.flow.Flow) -> dict:
         f["error"] = flow.error.get_state()
 
     if isinstance(flow, http.HTTPFlow):
+        f["http"] = True
```
mhils: I think we already transmit flow.type, so we can just use that. :)

web/src/js/filt/filt.peg
```diff
@@ -142,6 +146,10 @@ function contentType(regex){
     contentTypeFilter.desc = "content type matches " + regex;
     return contentTypeFilter;
 }
+function tcpFilter(flow){
+    return !!flow.tcp;
+}
+tcpFilter.desc = "is TCP Flow";
```
mhils: "is a TCP flow"

mitmproxy/tools/web/app.py
s4chin: Will do

web/src/js/filt/filt.peg
mhils: No worries. We do reviews for a reason. :)

s4chin:
I did as per review.

mhils:
Thanks!

PR #1948, opened 2017-01-20
Fix for dns_spoofing.py example

amm0nite:
Hi,
The `dns_spoofing.py` example works for HTTPS connections but not for HTTP. The original Host header can not be retrieved from `flow.request.pretty_host` in the `request` event.
Here is a new version of the example based on the solution suggested by @mhils in [this forum thread](https://discourse.mitmproxy.org/t/could-mitmproxy-be-used-for-reverse-proxy-for-multi-sites/253/6).

### DIFF
--- a/examples/complex/dns_spoofing.py
+++ b/examples/complex/dns_spoofing.py
@@ -27,7 +27,7 @@ import re
 # https://bugzilla.mozilla.org/show_bug.cgi?id=45891
 parse_host_header = re.compile(r"^(?P<host>[^:]+|\[.+\])(?::(?P<port>\d+))?$")
 
-class DnsSpoofing:
+class Rerouter:
     def __init__(self):
         self.hostHeader = None
 
@@ -55,4 +55,4 @@ class DnsSpoofing:
         flow.request.port = port
 
 def start():
-    return DnsSpoofing()
+    return Rerouter()
--- a/examples/complex/dns_spoofing.py
+++ b/examples/complex/dns_spoofing.py
@@ -29,15 +29,12 @@ parse_host_header = re.compile(r"^(?P<host>[^:]+|\[.+\])(?::(?P<port>\d+))?$")
 
 
 class Rerouter:
-    def __init__(self):
-        self.host_header = None
-
     def requestheaders(self, flow):
         """
         The original host header is retrieved early
         before flow.request is replaced by mitmproxy new outgoing request
         """
-        self.host_header = flow.request.headers.get('host')
+        flow.metadata["original_host"] = flow.request.headers["Host"]
 
     def request(self, flow):
         if flow.client_conn.ssl_established:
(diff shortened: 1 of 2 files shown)


examples/complex/dns_spoofing.py
```diff
@@ -28,22 +28,33 @@
...
-    if m:
-        host_header = m.group("host").strip("[]")
-        if m.group("port"):
-            port = int(m.group("port"))
-
-    flow.request.host = sni or host_header
-    flow.request.port = port
+class Rerouter:
+    def __init__(self):
+        self.hostHeader = None
```
mhils: We're following [PEP-8](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables) in our codebase, which asks for class attributes to be "lowercase with words separated by underscores". Can we make this `self.host_header` please? 😊

examples/complex/dns_spoofing.py
```diff
@@ -28,22 +28,33 @@
...
+
+    def requestheaders(self, flow):
+        self.hostHeader = flow.request.headers.get('host')
```
mhils: Can we add a comment here explaining why we do this in the requestheaders call and not down below?

amm0nite:
Thanks for your comments!
I made the changes requested, I hope the comment is explicit enough :smile:

mhils:
Awesome, thanks! 😃 

Looking more closely, I think there's a potential race condition here. Take the follwing sequence of events:
```
requestheaders(flow1)
requestheaders(flow2)
request(flow1)
request(flow2)
```
There's only a single `.host_header` attribute here, so this would assign the wrong host header to the first request. Instead of trying to deal with some mapping, can we do everything in `requestheaders`? 😃

amm0nite:
I think it is not possible to do everything in `requestheaders`, as we want to replace the host of the outgoing request made by mitmproxy but it is not yet created (`flow.request` is the initial client request). Am I missing something?

mhils:
>  as we want to replace the host of the outgoing request made by mitmproxy but it is not yet created

We have access to the flow in `requestheaders` and can update its `.request.host` attribute there. Am I missing something? :)

amm0nite:
- In `requestheaders`, `flow.request.host` has already the right value
- We are in reverse proxy mode, the `flow.request.host` is overwritten by the wrong value after `requestheaders`
As I understand it 😄

mhils:
I should have read the code properly - sorry for the confusion 🙈 
So yes, we need to come up with something else then. We can assign `flow.metadata["original_host"] = flow.request.headers["Host"]` in `requestheaders` and then use this later. How does that sound?

amm0nite:
Sounds good!

mhils:
This looks good now, although I believe we also need to update the host header again to the correct destination, not just `request.host`. Do you think you can add a test for this [here](https://github.com/mitmproxy/mitmproxy/blob/695cc23696bcaabf8c5e38d1004bffb167a6dc34/test/mitmproxy/test_examples.py#L41)?

mhils:
Thanks, fantastic! 🎉 🍰 😊

PR #1949, opened 2017-01-21
web: Add ~src and ~dst filters

s4chin:
I have mentioned the problem with `~s` and `~src` in Slack.


mhils:
From Slack:

How do you feel about merging UnaryExpr and NullaryExpr (and BooleanLiteral, while we're at it)?
We can then sort it so that `~src` comes before `~s`, which should fix the problem.

mhils:
:+1:

Issue #1939, opened 2017-01-18
Some critical errors on standard output instead of standard error


tomlabaude:
##### Steps to reproduce the problem:

I'm parsing output of mitmdump in Objective-C and some major errors appears on standard output.

Example:
"Insufficient privileges to access pfctl. " (osx.py:34 - Treated with raise RuntimeError)
-> This appears on standard output, although it's a critical error on a configuration point of view

Instead, example of a correct error output:
"Error starting proxy server: OSError(48, 'Address already in use')" (server.py:51 - Treated with raise exceptions.ServerException)
-> This "correctly" appears on standard error

##### Any other comments? What have you tried so far?

This is just an example, there may be other critical errors which should be treated on standard error

##### System information

Mitmproxy version: 1.0.2
Python version: 3.5.2
Platform: Darwin-16.3.0-x86_64-i386-64bit
SSL version: OpenSSL 1.0.2j  26 Sep 2016
Mac version: 10.12.2 ('', '', '') x86_64

mhils:
I think the root cause for this is here: https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/addons/termlog.py

I'd be happy to accept a PR that modifies the default argument to `None` and then uses either `sys.stdout` or `sys.stderr` depending on the log level.

Kriechi:
fixed by #1952

PR #1952, opened 2017-01-22
Fix errors on stdout

nish21:
Modified the termlog file as suggested in the comments on issue [#1939](https://github.com/mitmproxy/mitmproxy/issues/1939)

### DIFF
--- a/mitmproxy/addons/termlog.py
+++ b/mitmproxy/addons/termlog.py
@@ -18,7 +18,7 @@ class TermLog:
                 self.outfile = sys.stderr
             else:
                 self.outfile = sys.stdout
-        
+
         determine_outfile()
 
         if self.options.verbosity >= log.log_tier(e.level):
--- a/mitmproxy/addons/termlog.py
+++ b/mitmproxy/addons/termlog.py
@@ -18,7 +18,6 @@ class TermLog:
                 self.outfile = sys.stderr
             else:
                 self.outfile = sys.stdout
-
         determine_outfile()
 
         if self.options.verbosity >= log.log_tier(e.level):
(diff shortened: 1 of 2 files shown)


mitmproxy/addons/termlog.py
```diff
@@ -5,14 +5,19 @@
...
+    def __init__(self, outfile=None):
         self.options = None
         self.outfile = outfile
 
     def configure(self, options, updated):
         self.options = options
 
     def log(self, e):
+        if self.outfile == None:
+            if log.log_tier(e.level) == log.log_tier("error"):
```
Kriechi: I think this should be evaluated every time - not just on the first log message - because in the next line you set `self.outfile` to something.

nish21: I'm sorry, I don't think I understood what you want me to do.

mitmproxy/addons/termlog.py
```diff
@@ -5,14 +5,19 @@
...
 
     def configure(self, options, updated):
         self.options = options
 
     def log(self, e):
+        if self.outfile is None:
+            if log.log_tier(e.level) == log.log_tier("error"):
+                self.outfile = sys.stderr
+            else:
+                self.outfile = sys.stdout
```
Kriechi: The purpose of this PR is that all log messages marked as `error` are written to stderr.
However, if the very first message after program start is e.g. `info` you set `self.outfile = sys.stdout`.

This means the whole code block is only executed once, and after that either stdout or stderr is fixed for all following messages. IMHO this is not intended. Every new message should be evaluated again.

Kriechi:
Thanks! 🎉  👏 
I'll make some minor improvements in a separate commit.

PR #1953, opened 2017-01-22
Fix link to Linux binaries

offby1:


Kriechi:
I agree with this change in general. However, you fixed the link for macOS, while the Linux section still points to the too general mitmproxy.org.

Kriechi:
Merged via 6a7eeef0ee1207e92a2c57b630cfdfaf527595a5.

Issue #1925, opened 2017-01-12
Crash JSON error


yash0212:
Traceback (most recent call last):
  File "mitmproxy/tools/console/master.py", line 292, in run
  File "site-packages/urwid/main_loop.py", line 278, in run
  File "site-packages/urwid/main_loop.py", line 376, in _run
  File "site-packages/urwid/main_loop.py", line 682, in run
  File "site-packages/urwid/main_loop.py", line 719, in _loop
  File "site-packages/urwid/raw_display.py", line 393, in <lambda>
  File "site-packages/urwid/raw_display.py", line 493, in parse_input
  File "site-packages/urwid/main_loop.py", line 403, in _update
  File "site-packages/urwid/main_loop.py", line 503, in process_input
  File "mitmproxy/tools/console/window.py", line 84, in keypress
  File "site-packages/urwid/container.py", line 1116, in keypress
  File "mitmproxy/tools/console/statusbar.py", line 155, in keypress
  File "mitmproxy/tools/console/statusbar.py", line 108, in keypress
  File "mitmproxy/tools/console/statusbar.py", line 133, in prompt_execute
  File "mitmproxy/tools/console/statusbar.py", line 31, in __call__
  File "mitmproxy/tools/console/common.py", line 328, in export_to_clip_or_file
  File "mitmproxy/export.py", line 82, in python_code
  File "json/__init__.py", line 312, in loads
TypeError: the JSON object must be str, not 'bytes'

mitmproxy has crashed!

##### System information
Mitmproxy version: 1.0.2
Python version: 3.5.2
Platform: Linux-3.16.0-4-amd64-x86_64-with-debian-8.6
SSL version: OpenSSL 1.1.0c  10 Nov 2016
Linux distro: debian 8.6

mhils:
How can I reproduce this? Can you share the flow that triggers this?

PR #1954, opened 2017-01-23
Fix related to Issue #1925 

jayanthchandra:
Added decode function to resolve TypeError in JSON.


mhils:
Thanks for the PR. When looking at the initial issue I thought this bug be simpler, but the code you patched here is right. `.text` should never be bytes, so the issue is somewhere else. I'm closing this until the root cause is more clear.

PR #1955, opened 2017-01-23
increase addon test coverage

Kriechi:


mhils:
Nice. @cortesi mentioned this before, we should actually hard-fail if mitmproxy/addons reaches <100% test coverage. I would do it, but that sounds like a task you would actually enjoy? 😄

PR #1959, opened 2017-01-25
add test coverage protection

Kriechi:
This PR adds a custom pytest plugin using the hooks API.
New CLI parameter are available: `--full-cov=[path]` and `--no-full-cov=[file]`.
This makes pytest fail with an exitcode if the coverage for one or more file paths is less than 100%.

The plan is to slowly increase coverage for the mitmproxy core and add more and more paths until we can just say: `--full-cov=mitmproxy` which protects everything against dropping coverage metrics.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -28,7 +28,7 @@ install:
   - "pip install -U tox"
 
 test_script:
-  - ps: "tox -- --cov mitmproxy --cov pathod -v"
+  - ps: "tox -- --verbose --cov-report=term"
   - ps: |
       $Env:VERSION = $(python mitmproxy/version.py)
       $Env:SKIP_MITMPROXY = "python -c `"print('skip mitmproxy')`""
--- a/mitmproxy/utils/debug.py
+++ b/mitmproxy/utils/debug.py
@@ -45,17 +45,17 @@ def dump_system_info():
     ]
     d = platform.linux_distribution()
     t = "Linux distro: %s %s %s" % d
-    if d[0]:  # pragma: no-cover
+    if d[0]:  # pragma: no cover
         data.append(t)
 
     d = platform.mac_ver()
     t = "Mac version: %s %s %s" % d
-    if d[0]:  # pragma: no-cover
+    if d[0]:  # pragma: no cover
         data.append(t)
 
     d = platform.win32_ver()
     t = "Windows version: %s %s %s %s" % d
-    if d[0]:  # pragma: no-cover
+    if d[0]:  # pragma: no cover
         data.append(t)
 
     return "\n".join(data)
@@ -135,11 +135,11 @@ def dump_stacks(signal=None, frame=None, file=sys.stdout, testing=False):
             if line:
                 code.append("  %s" % (line.strip()))
     print("\n".join(code), file=file)
-    if not testing:
+    if not testing:  # pragma: no cover
         sys.exit(1)
 
 
 def register_info_dumpers():
-    if os.name != "nt":
+    if os.name != "nt":  # pragma: windows no cover
         signal.signal(signal.SIGUSR1, dump_info)
         signal.signal(signal.SIGUSR2, dump_stacks)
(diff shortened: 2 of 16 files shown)


mitmproxy/script/concurrent.py
```diff
@@ -11,7 +11,7 @@ class ScriptThread(basethread.BaseThread):
     name = "ScriptThread"
 
 
-def concurrent(fn):
+def concurrent(fn):  # pragma: no coverage windows
```
mhils: This should be covered on Windows?

mitmproxy/utils/debug.py
```diff
@@ -34,7 +34,7 @@ def dump_system_info():
                 tag=int(tag_dist),
                 commit=commit,
             )
-        except:
+        except:  # pragma: no cover
```
mhils: I think we could test this?

mitmproxy/utils/typecheck.py
```diff
@@ -26,7 +26,7 @@ def check_type(attr_name: str, value: typing.Any, typeinfo: type) -> None:
     if typename.startswith("typing.Union"):
         try:
             types = typeinfo.__args__
-        except AttributeError:
+        except AttributeError:  # pragma: no cover
```
mhils: Why this?

pathod/protocols/websockets.py
```diff
@@ -53,4 +53,4 @@ def handle_websocket(self, logger):
                         )
                         lg("crafting websocket spec: %s" % frame_log["spec"])
                         self.pathod_handler.addlog(frame_log)
-        return self.handle_websocket, None
+        return self.handle_websocket, None  # pragma: no cover
```
mhils: What's this for?

Kriechi: The `while True` never returns, so this return statement never triggers. I guess it can be removed completely, but it might be required for type deduction.

mitmproxy/utils/typecheck.py
Kriechi: This branch is not executed on Python 3.6 - because the import works (and does not fail as on Python 3.5).

mitmproxy/script/concurrent.py
Kriechi: Apparently it isn't...

pathod/protocols/websockets.py
mhils: This is unreachable code, so let's remove it entirely.

mitmproxy/utils/typecheck.py
mhils: Presuming you mean attribute access and not import: yes - so ideally we'd have a mock object that triggers this specific case instead of taking it out of coverage?

mitmproxy/script/concurrent.py
mhils: Right - there's no excuse why we couldn't test this on Windows, so let's either fix the test for it or exclude this file from `--full-cov` for now. This should be tested on Windows eventually.

mitmproxy/utils/typecheck.py
Kriechi: Oh yes you are right - the pattern just looks too similar to the import clauses we used to have. I will mock it to test it properly.

mitmproxy/script/concurrent.py
Kriechi: I wasn't aware that it SHOULD work on windows. Why do the existing tests not cover it then?

cortesi:
I would love to enforce an even stronger coverage statement here - a module's test suite should result in 100% coverage of the module. That is, this should result in 100% coverage:

> py.test --cov ./mitmproxy/addons ./test/mitmproxy/addons

I don't see an easy way to do that in a single run as a py.test plugin, though 

That said, a great improvement for us, and another step forward for the Kriechi Correctness Empire. Let's do it.

Kriechi:
This PR is now ready. We can evolve the checks and coverage whitelist/blacklist in future PRs.

.appveyor.yml
```diff
@@ -28,7 +28,7 @@ install:
   - "pip install -U tox"
 
 test_script:
-  - ps: "tox -- --cov mitmproxy --cov pathod -v"
+  - ps: "tox -- --verbose --cov-report=term"
```
Kriechi: For my new feature to work, we need `--cov` which by default adds the term-report. However, just running tox, I don't want to display it. Therefore I disabled the report in tox explicitly. On Travis+Appveyor I override it, because I DO want to see it for CI reasons.

PR #1965, opened 2017-01-29
Consolidate option names

cortesi:
- Onboarding app options to onboarding*
- Console-specific options to console_*
- Web-specific options to web_*

### DIFF
--- a/mitmproxy/tools/cmdline.py
+++ b/mitmproxy/tools/cmdline.py
@@ -212,9 +212,9 @@ def get_common_options(args):
         args.verbose = 0
 
     return dict(
-        app=args.app,
-        app_host=args.app_host,
-        app_port=args.app_port,
+        onboarding=args.onboarding,
+        onboarding_host=args.onboarding_host,
+        onboarding_port=args.onboarding_port,
 
         anticache=args.anticache,
         anticomp=args.anticomp,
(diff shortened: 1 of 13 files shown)


Kriechi:
Prefixing looks kinda weird - any thoughts on using a proper class / wrapper object?

mhils:
The bigger question is how we integrate external addons. In any case, this is a great intermediary step to actually prefix everything with what it belongs to.

mitmproxy/tools/cmdline.py
```diff
@@ -568,23 +568,23 @@ def proxy_ssl_options(parser):
 def onboarding_app(parser):
     group = parser.add_argument_group("Onboarding App")
     group.add_argument(
-        "--noapp",
-        action="store_false", dest="app",
+        "--noonboarding",
```
Kriechi: I'd prefer `--no-onboarding` - IMHO this is a more common way to negate options.

cortesi:
We toyed with separate classes for this before, and it introduces a lot of complexity. I think prefixing is fine for now. We could add some sugar to the config file if we want so that a ```console:``` section automatically translates into a ```console_``` prefix to make user configuration a bit nicer.

mitmproxy/tools/cmdline.py
mhils: 🕛🏄 it is!

cortesi: Done.

Kriechi:
Ok - sounds good to me!

mitmproxy/tools/cmdline.py
```diff
@@ -568,23 +568,23 @@ def proxy_ssl_options(parser):
 def onboarding_app(parser):
     group = parser.add_argument_group("Onboarding App")
     group.add_argument(
-        "--noapp",
-        action="store_false", dest="app",
+        "--no-onboarding",
+        action="store_false", dest="onboarding",
         help="Disable the mitmproxy onboarding app."
```
Kriechi: Is everybody familiar with the onboarding app in general?
Maybe We should add a one sentence description of what the purpose is?

Kriechi:
@cortesi please add more tests to increase coverage of optmanager back to 100%!

PR #1967, opened 2017-01-30
Add png parser

s4chin:
From Slack:

`kaitai` has the function from_file(filename), but we have data. I use `imghdr.what(io.BytesIO(data)) == 'png')` and use the parser to return metadata. But it doesn't work. `mitmweb` shows raw data and `Couldn't parse: falling back to Raw`. How do I solve this?


mitmproxy/contentviews/image.py
```diff
@@ -19,6 +21,10 @@ class ViewImage(base.View):
     ]
 
     def __call__(self, data, **metadata):
+        if imghdr.what(io.BytesIO(data)) == 'png':
```
mhils: Take a look at the `h` parameter of `imghdr.what` to simplify this. :)

mitmproxy/contentviews/img/png.ksy
```diff
@@ -0,0 +1,207 @@
+meta:
```
mhils: I don't think we need this in the repo, a comment in the generated file where it comes from is sufficient.

mitmproxy/contentviews/image.py
```diff
@@ -19,6 +21,10 @@ class ViewImage(base.View):
...
+            img = metadata.Metadata.open(io.BytesIO(data))
```
mhils: I feel your pain - the kaitai API documentation is rather shitty on that end. From [here](https://github.com/kaitai-io/kaitai_struct_python_runtime/blob/2a559b66d75d102a19a72a5c6544c850b39dd6f6/kaitaistruct.py#L25):

```
Png(KaitaiStream(data))
```

mitmproxy/contentviews/img/metadata.py
```diff
@@ -0,0 +1,30 @@
+from . import png
```
mhils: Why img/ and not image/?

mitmproxy/contentviews/img/metadata.py
```diff
@@ -0,0 +1,30 @@
...
+
+class Metadata:
+    def __init__():
+        pass
+
+    def get_png_metadata():
```
mhils: How about we just make this a top-level function - I don't think we need a class for this? :)

s4chin: Multiple top level function for multiple image types? Or everything in just one function?

mitmproxy/contentviews/img/metadata.py
s4chin: Because `image.py` is also a file and running `mitmproxy` shows the error `AttributeError: module 'mitmproxy.contentviews.image' has no attribute 'ViewImage'`

mitmproxy/contentviews/img/metadata.py
mhils: Multiple functions please so that they can be easily tested! :smiley:
```python
Metadata = typing.List[typing.Tuple[str, str]]

def parse_png(data: bytes) -> Metadata:
    # extract metadata here
```

mitmproxy/contentviews/img/metadata.py
mhils: Yes, we'd need to move this into `mitmproxy.contentviews.image` as well, e.g. as `mitmproxy/contentviews/image/pillow.py`

mitmproxy/contentviews/image/metadata.py
```diff
@@ -1,21 +1,26 @@
 from . import png
 
 def get_png(img):
-    metadata = [('Format', 'Portable network graphics')]
+    # For debugging
+    try:
+        img = png.Png(img)
+    except Exception as e:
+        print(type(e).__name__, e.args)
```
s4chin: The result of this is `AttributeError ("'bytes' object has no attribute 'read'",)`

mhils: See https://github.com/mitmproxy/mitmproxy/pull/1967#discussion_r98495772

s4chin:
Works! :smiley: 
It had a lot more problems than I think it did.
Now I have to add tests and add `kaitaistruct` in requirements.

s4chin:
So I squashed everything and added a small test. Can you check?
Also, I think `pillow.py` should be renamed to something else.

mitmproxy/contentviews/__init__.py
```diff
@@ -170,7 +171,7 @@ def get_content_view(viewmode: View, data: bytes, **metadata):
 add(css.ViewCSS())
 add(urlencoded.ViewURLEncoded())
 add(multipart.ViewMultipart())
-add(image.ViewImage())
+add(pillow.ViewImage())
```
mhils: We can rename `pillow.py` to `view.py` or something like this and then add this to `__init__.py`:

```
from .view import ViewImage
```

Then all the imports should work just as before and this file can stay untouched.

mitmproxy/contentviews/image/image_parser.py
```diff
@@ -0,0 +1,34 @@
+from kaitaistruct import KaitaiStream
+
+from . import png
+
+def get_png(data):
```
mhils: Can you please add type annotations? Also this function needs a better name, e.g. `extract_metadata_from_png`, `parse_png` or `handle_png`, or .... It does not "get" a png. 😃 


```python
Metadata = typing.List[typing.Tuple[str, str]]

def parse_png(data: bytes) -> Metadata:
```

mitmproxy/contentviews/image/image_parser.py
```diff
@@ -0,0 +1,34 @@
...
+    img = png.Png(KaitaiStream(data))
+    parts = {'format': 'Portable network graphics'}
```
mhils: Making this a dict is intuitively right, but we need to consider that some keys could occur twice in the metadata. I'd propose to make it a list of `(key, value)` tuples.

mitmproxy/contentviews/image/image_parser.py
```diff
@@ -0,0 +1,34 @@
...
+        chunk = img.chunks[i]
+        if chunk.type == 'gAMA':
+            gamma = chunk.gamma_int / 100000
+            parts['gamma'] = gamma
+        elif chunk.type == 'pHYs':
+            aspectx = chunk.pixels_per_unit_x
+            aspecty = chunk.pixels_per_unit_y
+            parts["aspectx"] = aspectx
+            parts["aspecty"] = aspecty
+    return f, parts
```
mhils: What's the point of `f` here? I think it makes sense to move that to `ViewImage.__call__` as well.

mitmproxy/contentviews/image/image_parser.py
```diff
@@ -0,0 +1,34 @@
...
+from . import png
+
+def get_png(data):
+    img = png.Png(KaitaiStream(data))
+    parts = {'format': 'Portable network graphics'}
+    f = 'PNG'
+    width = img.ihdr.width
+    height = img.ihdr.height
+    parts["width"] = width
+    parts["height"] = height
```
mhils: No need to have an additional temporary variable here.

mitmproxy/contentviews/image/image_parser.py
```diff
@@ -0,0 +1,34 @@
...
+    for i in range(0, len(img.chunks)):
+        chunk = img.chunks[i]
```
mhils: See "Loops" at https://jeffknupp.com/blog/2012/10/04/writing-idiomatic-python/

mitmproxy/contentviews/image/image_parser.py
```diff
@@ -0,0 +1,34 @@
...
+        if chunk.type == 'gAMA':
+            gamma = chunk.gamma_int / 100000
+            parts['gamma'] = gamma
+        elif chunk.type == 'pHYs':
+            aspectx = chunk.pixels_per_unit_x
+            aspecty = chunk.pixels_per_unit_y
+            parts["aspectx"] = aspectx
+            parts["aspecty"] = aspecty
```
mhils: This is missing the various text chunks?

mitmproxy/contentviews/image/image_parser.py
```diff
@@ -0,0 +1,34 @@
...
+    return f, parts
+
+def format_contentviews(parts):
```
mhils: What is the purpose of this function?

mitmproxy/contentviews/image/pillow.py
```diff
@@ -19,6 +22,11 @@ class ViewImage(base.View):
     ]
 
     def __call__(self, data, **metadata):
+        if imghdr.what('', h=data) == 'png':
```
mhils: We don't need to pass an empty string here.

mitmproxy/contentviews/image/pillow.py
```diff
@@ -19,6 +22,11 @@ class ViewImage(base.View):
...
+            f, parts = image_parser.get_png(io.BytesIO(data))
+            parts = image_parser.format_contentviews(parts)
+            fmt = base.format_dict(multidict.MultiDict(parts))
```
mhils: dict -> list -> dict -> list?

test/mitmproxy/contentviews/test_image_parser.py
```diff
@@ -0,0 +1,14 @@
+import io
+
+from mitmproxy.contentviews.image import image_parser
+from mitmproxy.test import tutils
+
+def test_png_parser():
```
mhils: This definitely needs more tests. See http://www.schaik.com/pngsuite/#ancillary for test files. We should have those at least:
```
ct0 - no textual data
ct1 - with textual data
ctz - with compressed textual data
cte - UTF-8 international text - english
```

mitmproxy/contentviews/image/image_parser.py
s4chin: I hadn't added them yet. Will add tomorrow.

mitmproxy/contentviews/image/image_parser.py
s4chin: As you said [here](https://mitmproxy.slack.com/archives/general/p1485885141003847) - return values to make testing easy, and then convert it into `contentviews` format

mitmproxy/contentviews/image/pillow.py
s4chin: I removed the empty string and - `TypeError: what() missing 1 required positional argument: 'file'`

test/mitmproxy/contentviews/test_image_parser.py
s4chin: The `png.ksy` given in the `kaitaistruct` formats doesn't deal with `iTXt` and `zTXt` chunks. I'll try to modify `png.ksy` and compile and use that one.

mitmproxy/contentviews/image/image_parser.py
```diff
@@ -0,0 +1,26 @@
...
+
+from kaitaistruct import KaitaiStream
+
+from . import png
+
+Metadata = typing.List[typing.Tuple[str, str]]
+
+def parse_png(data: bytes) -> Metadata:
+    img = png.Png(KaitaiStream(data))
+    parts = [tuple(['Format', 'Portable network graphics'])]
```
mhils: You don't have to write `tuple()`, you can just use
```python
parts = [
    ("Format", "Portable network graphics")
]
```

mitmproxy/contentviews/image/image_parser.py
```diff
@@ -0,0 +1,26 @@
...
+    parts.append(tuple(['Size', str(img.ihdr.width) + " x " + str(img.ihdr.height) + " px"]))
```
mhils: Can we please use `"".format()` here?

mitmproxy/contentviews/image/view.py
```diff
@@ -19,6 +22,11 @@ class ViewImage(base.View):
     ]
 
     def __call__(self, data, **metadata):
+        if imghdr.what('', h=data) == 'png':
+            f = "PNG"
+            parts = image_parser.parse_png(io.BytesIO(data))
+            fmt = base.format_dict(multidict.MultiDict(parts))
+            return "%s image" % f, fmt
```
mhils: Just make this `"PNG image"` and ditch f.

test/mitmproxy/contentviews/test_image_parser.py
```diff
@@ -0,0 +1,88 @@
+import io
+
+from mitmproxy.contentviews.image import image_parser
+from mitmproxy.test import tutils
+
+class TestPngParser:
+    def test_png_size(self):
```
mhils: This looks way better now! We can make @Kriechi happy and parameterize this a bit:


```python
@pytest.mark.parametrize("filename,metadata", {
    "ct1n0g04.png": [
        ('Format', 'Portable network graphics'),
        ('Size', '32 x 32 px'),
        ('gamma', '1.0')
    ],
    "ctzn0g04.png": # ...
}.items())
def test_parse_png(filename, metadata):
    with open(tutils.test_data.path(filename), "rb") as f:
         assert metadata == image_parser.parse_png(f.read())
```

No need to to the `in` checks, let's just specify them entirely in the correct order right away.

mitmproxy/contentviews/image/png.py
```diff
@@ -0,0 +1,289 @@
+# This is a generated file! Please edit source .ksy file and use kaitai-struct-compiler to rebuild
```
mhils: CI complains about missing test coverage in this file, so shall we move this into mitmproxy/contrib/kaitaistruct?

mitmproxy/contentviews/image/view.py
s4chin: I was thinking it'll look like this -
```
if 'png':
    parse_png()
elif 'jpeg':
    parse_jpeg()
fmt = format_dict()
return "%s image" % f, fmt
```
Otherwise I'll have to add a `return` inside every `if`.

s4chin: Or I could just return "PNG image" now and change it on subsequent PRs.

mhils: Ok, makes sense.

test/mitmproxy/contentviews/test_image_parser.py
s4chin: So after this, all the png tests can be done under a single function? No need of class?

setup.py
```diff
@@ -71,6 +71,7 @@
         "html2text>=2016.1.8, <=2016.9.19",
         "hyperframe>=4.0.1, <5",
         "jsbeautifier>=1.6.3, <1.7",
+        "kaitaistruct>=0.5",
```
mhils: We want an upper bound here to guard against upgrade in kaitaistruct that break compatibility. Can we make this `>=0.5, <0.6` please?

test/mitmproxy/contentviews/test_image_parser.py
```diff
@@ -0,0 +1,62 @@
...
+        ('Format', 'Portable network graphics'),
+        ('Size', '1280 x 798 px'),
+        ('aspect', '72 x 72'),
+        ('date:create', '2012-07-11T14:04:52-07:00'),
+        ('date:modify', '2012-07-11T14:04:52-07:00')
+    ],
+}.items())
+def test_parse_png(filename, metadata):
+    with open(tutils.test_data.path(filename), "rb") as f:
+         assert metadata == image_parser.parse_png(io.BytesIO(f.read()))
```
mhils: Why do we wrap with BytesIO here (same above)? We said the function should accept bytes in the type definition, so let's pass bytes! If we need to wrap it into BytesIO before passing it into KaitaiStream, let's do that in `parse_png`. :)

test/mitmproxy/contentviews/test_image_parser.py
```diff
@@ -20,7 +18,9 @@
         ('Title', 'PngSuite'),
         ('Author', 'Willem A.J. van Schaik\n(willem@schaik.com)'),
         ('Copyright', 'Copyright Willem van Schaik, Singapore 1995-96'),
-        ('Description', 'A compilation of a set of images created to test the\nvarious color-types of the PNG format. Included are\nblack&white, color, paletted, with alpha channel, with\ntransparency formats. All bit-depths allowed according\nto the spec are present.'), ('Software', 'Created on a NeXTstation color using "pnmtopng".'), ('Disclaimer', 'Freeware.')
+        ('Description', 'A compilation of a set of images created to test the\nvarious color-types of the PNG format. Included are\nblack&white, color, paletted, with alpha channel, with\ntransparency formats. All bit-depths allowed according\nto the spec are present.'),
```
s4chin: The only lint errors are that these lines are > 140 characters. Doing it like this
```
    A compilation of a \
    set of images ...
```
causes extra spaces in the sentence, and removing the spaces will kill the indentation. So I left it as it is.

mhils: I don't care how you solve this, but CI needs to pass. :)

s4chin: Should I do it like this then?
```
    A compilation of \
a set of images...
```

ujjwal96: @s4chin You can do something like this:
```
('Description', 'A compilation of a set of images created to test the\nvarious color-types of the PNG format. Included are\n'
                'black&white, color, paletted, with alpha channel, with\ntransparency formats. All bit-depths allowed according\n'
                'to the spec are present.')
```

PR #1975, opened 2017-02-01
Catch ValueErrors from url.parse()

lymanZerga11:
Catch `ValueError` from calls to `mitmproxy.net.http.url.parse()`. Otherwise mitmproxy crashes due to uncaught exception.


mitmproxy/tools/console/flowlist.py
```diff
@@ -337,7 +337,11 @@ def get_url(self, method):
...
     def new_request(self, url, method):
-        parts = mitmproxy.net.http.url.parse(str(url))
+        try:
+            parts = mitmproxy.net.http.url.parse(str(url))
+        except ValueError as e:
+            signals.status_message.send(message=str(e))
+            return
         if not parts:
             signals.status_message.send(message="Invalid Url")
             return
```
mhils: `if not parts:` cannot happen anymore I think, let's remove it?

lymanZerga11: okay! I'll remove it once I add the tests

PR #1980, opened 2017-02-02
improve tests

Kriechi:


mitmproxy/test/tflow.py
```diff
@@ -115,6 +115,26 @@ def tflow(client_conn=True, server_conn=True, req=True, resp=None, err=None):
     return f
 
 
+def tdummyflow(client_conn=True, server_conn=True, err=None):
+    class DummyFlow(flow.Flow):
```
mhils: Move class definition out of function?

test/mitmproxy/addons/test_view.py
```diff
@@ -170,8 +171,10 @@ def test_reversed():
     assert v[0].request.timestamp_start == 3
     assert v[-1].request.timestamp_start == 1
     assert v[2].request.timestamp_start == 1
-    tutils.raises(IndexError, v.__getitem__, 5)
-    tutils.raises(IndexError, v.__getitem__, -5)
+    with pytest.raises(IndexError):
+        v.__getitem__(5)
```
mhils: We can just `_ = v[5]` here now, same for __setitem__ below. :-)

test/conftest.py
```diff
@@ -9,6 +10,53 @@
...
+    "APPVEYOR" in os.environ,
+    reason='Skipping due to Appveyor'
+)
+
+
+original_pytest_raises = pytest.raises
+
+
+def raises(exc, *args, **kwargs):
+    functools.wraps(original_pytest_raises)
```
mhils: leftover?

mitmproxy/test/tflow.py
Kriechi: I thought about it, but IMO since this class is only used to create and return the dummyflow, I thought its cleaner to contain everything inside the function. But I don't have strong feelings about it - do you think I should move it out of the function?

test/conftest.py
Kriechi: No, I think this is how it is supposed to look like...

test/mitmproxy/addons/test_view.py
Kriechi: I didn't really look at the overall test code itself - just search&replace.
You can simplify it even further, the assignment is not required either.

test/conftest.py
mhils: It's a decorator: https://docs.python.org/3/library/functools.html#functools.wraps

We can do this if you want:
```python
@functools.wraps(original_pytest_raises)
def raises(exc, *args, **kwargs):
    # ....
```

mitmproxy/test/tflow.py
mhils: Yes, let's move it out.
This creates a new class everytime you call the function, that's a bit weird.

Kriechi: Done.

test/conftest.py
Kriechi: D'oh! Done.

PR #1984, opened 2017-02-03
Fixes #1978

### LINKED ISSUE #1978: Crash of har_dump.py
I sometimes get either `b'' is not JSON serializable` or  `Object of type 'bytes' is not JSON serializable`

I used mitmproxy to extract such a flow and uploaded it [here](https://gist.github.com/Junkern/495d4c7c1b572d7749f0291adfb0dd15)

1.
2.
3.

It seems, that the error mostly occurs for `POST` and `HEAD` requests. Sometimes with an empty response body, sometimes not. Seldomly a `GET` requests triggers this error

##### System information
mitmproxy 1.0.2
macOs Sierra 10.12.3

Edit: Updated the Gist link to point to a working binary gist

ujjwal96:
Handled the case of empty string

### DIFF
--- a/examples/complex/har_dump.py
+++ b/examples/complex/har_dump.py
@@ -7,6 +7,7 @@ import json
 import sys
 import base64
 import zlib
+import os
 
 from datetime import datetime
 import pytz
@@ -166,7 +167,7 @@ def done():
         if dump_file.endswith('.zhar'):
             raw = zlib.compress(raw, 9)
 
-        with open(dump_file, "wb") as f:
+        with open(os.path.expanduser(dump_file), "wb") as f:
             f.write(raw)
 
         mitmproxy.ctx.log("HAR dump finished (wrote %s bytes to file)" % len(json_dump))
--- a/mitmproxy/net/http/encoding.py
+++ b/mitmproxy/net/http/encoding.py
@@ -31,8 +31,8 @@ def decode(encoded: Union[str, bytes], encoding: str, errors: str='strict') -> U
     Raises:
         ValueError, if decoding fails.
     """
-    if len(encoded) == 0:
-        return encoded
+    if encoded is None:
+        return None
 
     global _cache
     cached = (
@@ -72,8 +72,8 @@ def encode(decoded: Union[str, bytes], encoding: str, errors: str='strict') -> U
     Raises:
         ValueError, if encoding fails.
     """
-    if len(decoded) == 0:
-        return decoded
+    if decoded is None:
+        return None
 
     global _cache
     cached = (
@@ -86,10 +86,7 @@ def encode(decoded: Union[str, bytes], encoding: str, errors: str='strict') -> U
         return _cache.encoded
     try:
         try:
-            value = decoded
-            if isinstance(value, str):
-                value = decoded.encode()
-            encoded = custom_encode[encoding](value)
+            encoded = custom_encode[encoding](decoded)
         except KeyError:
             encoded = codecs.encode(decoded, encoding, errors)
         if encoding in ("gzip", "deflate", "br"):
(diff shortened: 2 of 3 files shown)


mhils:
Thanks. This needs to answer some questions first:

- What was the actual problem in #1978?
- How did you solve it?
- Can we create a test for this?

ujjwal96:
> What was the actual problem in #1978?    

There were 2 cases in which I could reproduce the problem. First, if the content size was 0. Second, if in case of POST, PUT, PATCH request there was no postData.  
In both the cases the error was that the "text" key had an empty bytes object, which is not JSON serializable.  
Something like this was happening.
```
>>> d['bytestr'] = b''
>>> type(d['bytestr'])
<class 'bytes'>
>>> json.dumps(d)
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.5/json/__init__.py", line 230, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python3.5/json/encoder.py", line 198, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.5/json/encoder.py", line 256, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.5/json/encoder.py", line 179, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: b'' is not JSON serializable
```
> How did you solve it?

I solved it by assigning an empty string whenever an empty bytes like object was return by `flow.request.get_text(strict=False)` or `flow.response.get_text(strict=False)` 
Also, an empty bytes object OR empty string evaluated to empty string so I used that while assignment
```
>>> b'' or ""
''
```
> Can we create a test for this?

Yes, I think. We can send a response with no content and a POST/PUT/PATCH request with no post data. I'll make them :slightly_smiling_face:

ujjwal96:
Apart from fixing this I also used os.path.expanduser() for opening the file while saving the json. Is that fine? It makes it convenient for the user to specify an output file.

mhils:
Thanks for the explanation.

> I solved it by assigning an empty string whenever an empty bytes like object was return by ?`flow.request.get_text(strict=False)` or `flow.response.get_text(strict=False)`

If you look at the implementation of [`get_text`](https://github.com/mitmproxy/mitmproxy/blob/53f298ac41ff812ca69a8220862746f55c1505b8/mitmproxy/net/http/message.py#L181), you'll see that it should return a `str` or `None`, but never bytes. If `.get_text` returns bytes, we should fix that rather fixing it here in the HAR dump example.

ujjwal96:
So should I close this PR and send another one with that fix?

mhils:
Feel free to force-push the correct changes here. You can also open a new PR if you prefer that.

ujjwal96:
I don't get travis failed :confused:

test/mitmproxy/test_examples.py
```diff
@@ -209,3 +209,21 @@ def test_binary(self):
                 har = json.load(inp)
 
         assert len(har["log"]["entries"]) == 1
+
+    def test_empty_bytes(self):
```
mhils: We should rather have new tests in https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/net/http/test_encoding.py. I just looked at what's there already and that doesn't make much sense to me. gzip/deflate/br should only be tested on bytes (they are byte -> byte transformations), and apparently there are no tests for byte->str encodings like utf8. If you feel comfortable doing that:

- remove str testing for gzip/deflate/br in test_encoders so that it only tests those encodings on bytes.
- copy test_encoders for utf8/latin1 and test byte->str encoding there.
- add docstrings explaining that.
- remove this test here, it tests no special case if the encodings module works as intended.

Issue #1994, opened 2017-02-05
Cannot bind mitmproxy to IPv6 address


ujjwal96:
##### Steps to reproduce the problem:

1. `mitmproxy -b ::1`
`Error starting proxy server: gaierror(-9, 'Address family for hostname not supported')`

##### Any other comments? What have you tried so far?
The relevant code is probably hardcoded to IPv4  
https://github.com/mitmproxy/mitmproxy/blob/2316c0fb74efefd76970c685b8f77f45834f1490/mitmproxy/net/tcp.py#L921

Original Discussion:
https://discourse.mitmproxy.org/t/serving-on-ipv6/331

Kriechi:
Looks like we have a few options how to fix this:
`socket.AF_INET6 if ':' in host and hasattr(socket, 'AF_INET6') else socket.AF_INET`

or probably a more complete solution would be:
```python
try:
    info = socket.getaddrinfo(host, port, socket.AF_UNSPEC, socket.SOCK_STREAM)
except socket.gaierror:
    if ':' in host:
        info = [(socket.AF_INET6, socket.SOCK_STREAM, 0, "", (host, port, 0, 0))]
    else:
        info = [(socket.AF_INET, socket.SOCK_STREAM, 0, "", (host, port))]

family, socktype, proto, canonname, sa = info[0]
s = socket.socket(family, socktype, proto)
s.bind((host, port))
```


(see http://www.programcreek.com/python/example/946/socket.AF_INET6)

mhils:
IIRC there's something in stdlib for this. Will check later.

Thomas Kriechbaumer <notifications@github.com> schrieb am So., 5. Feb.
2017, 14:39:

hedger:
I think I fixed that. https://github.com/mitmproxy/mitmproxy/pull/1995

mhils:
The thing from stdlib I was remembering is [`socket.create_connection`](https://docs.python.org/3/library/socket.html#socket.create_connection), which we actually should use for outgoing connections. We can possibly eliminate `tcp.Address` then, which wasn't a really good idea in hindsight.

For listening, I'll reply in #1995.

PR #1995, opened 2017-02-05
IPv6 binding support

hedger:
Added support for binding to IPv6 interfaces. Now mitmproxy can automatically detect & bind to IPv6 addresses, for example, using `-b ::1` or `-b localhost -6`.
Introduces 2 new options: `-4` (`--bind-ipv4`) and `-6` (`--bind-ipv6`).

Based on my request from https://discourse.mitmproxy.org/t/serving-on-ipv6/331/4.

### DIFF
--- a/mitmproxy/net/tcp.py
+++ b/mitmproxy/net/tcp.py
@@ -333,11 +333,11 @@ class Address(serializable.Serializable):
         return Address(**state)
 
     @classmethod
-    def wrap(cls, t):
+    def wrap(cls, t, **kw):
         if isinstance(t, cls):
             return t
         else:
-            return cls(t)
+            return cls(t, **kw)
 
     def __call__(self):
         return self.address
@@ -923,7 +923,7 @@ class TCPServer:
     request_queue_size = 20
 
     def __init__(self, address, use_ipv6=None):
-        self.address = Address(address, use_ipv6)
+        self.address = Address.wrap(address, use_ipv6=use_ipv6)
         self.__is_shut_down = threading.Event()
         self.__shutdown_request = False
         self.socket = socket.socket(self.address.family, socket.SOCK_STREAM)
--- a/mitmproxy/net/tcp.py
+++ b/mitmproxy/net/tcp.py
@@ -6,6 +6,8 @@ import threading
 import time
 import traceback
 import binascii
+from ipaddress import ip_address, IPv6Address
+
 from ssl import match_hostname
 from ssl import CertificateError
 
@@ -306,9 +308,15 @@ class Address(serializable.Serializable):
         ipv6 information.
     """
 
-    def __init__(self, address, use_ipv6=False):
+    def __init__(self, address, use_ipv6=None):
         self.address = tuple(address)
-        self.use_ipv6 = use_ipv6
+        if use_ipv6 is None:
+            try:
+                self.use_ipv6 = isinstance(ip_address(address[0]), IPv6Address)
+            except ValueError:
+                self.use_ipv6 = False
+        else:
+            self.use_ipv6 = use_ipv6
 
     def get_state(self):
         return {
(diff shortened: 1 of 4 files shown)


hedger:
Fixing for tests to pass.

dwfreed:
This doesn't (and shouldn't) need any options.  I could see providing an option to set IPV6_V6ONLY (option is portable, but note that Windows defaults to on, but most Unixes default to off), but that's it.

mhils:
I agree with @dwfreed. As I see it, a simple `":" in addr` (as proposed by @Kriechi in #1994) should be good enough to distinguish addresses. We can add support for IPV6_V6ONLY option in a later PR if someone has a good use case for it.

hedger:
But `addr` may be a hostname without ":", so you need an option to specify a protocol in that case.
For example, my implementation allows binding with `-b localhost` both to ipv4 and ipv6 interfaces, without need to specify a protocol-specific address.

dwfreed:
Hence @Kriechi's suggestion to use getaddrinfo instead.

mitmproxy/net/tcp.py
```diff
@@ -306,9 +308,15 @@ class Address(serializable.Serializable):
         ipv6 information.
     """
 
-    def __init__(self, address, use_ipv6=False):
+    def __init__(self, address, use_ipv6=None):
         self.address = tuple(address)
-        self.use_ipv6 = use_ipv6
+        if use_ipv6 is None:
+            try:
+                self.use_ipv6 = isinstance(ip_address(address[0]), IPv6Address)
```
hedger: Also here's a place where I try to guess a protocol based on supplied address, so "-b ::1" binds to correct address without need to speficy `-6` option.

hedger: Also here's a place where I try to guess a protocol based on supplied address, so "-b ::1" binds to correct address without need to speficy `-6` option.

mhils:
Thanks for the valuable input. How do you guys feel about just trying to bind to IPv6 first¹ and then falling back to IPv4 if that fails? This should work for both explicit IPs and hostnames and would not require additional configuration switches. I'm not totally opposed to adding those, but I'd prefer to keep things simple. The "I want to specify a hostname, but I want IPv4 only"-use case is kind of exotic and I could live without supporting that (just specifying an IP would be a simple workaround for users).

¹) We may need to enforce IPV6_ONLY=false on Windows (?)

Kriechi:
I guess binding to IPv6-only first works most of the time, even if you don't have real IPv6 connectivity, which might be misleading then.

mhils:
Do you have empirical evidence that supports that claim? For example, do we
know of a scenario where IPv6-binding to `""` causes issues?

I'll check on Windows later.

Thomas Kriechbaumer <notifications@github.com> schrieb am Mi., 8. Feb.
2017, 13:02:

dwfreed:
Binding to the IPv6 any address (`::`), with explicit IPV6_ONLY=false (yes, this is required on Windows) will allow both IPv4 and IPv6 connections to the proxy, regardless of the host's IPv6 connectivity.  Since 1.0, we've had a separate upstream bind address configuration, so this wouldn't break upstream connections.

mhils:
> Binding to the IPv6 any address (::), with explicit IPV6_ONLY=false (yes, this is required on Windows) will allow both IPv4 and IPv6 connections to the proxy, regardless of the host's IPv6 connectivity. 

Right, I just verified this on Windows 10. There's a bit of added fun as that `socket.IPPROTO_IPV6` is undefined on my system (IPV6_V6ONLY exists though), but I see this working for Windows. How are things looking on OSX/Linux?

mhils:
@hedger: Are you interested in pushing this further or shall we close here and take a look ourselves?

hedger:
@mhils yes, short on time lately, unfortunately.

mhils:
No worries. Thanks for the initial PR!

PR #1996, opened 2017-02-05
Closes #1580 Gives warning when explicit proxy configured in transparent mode

### LINKED ISSUE #1580: Provide a better error message for proxy mode misconfigurations
1. `mitmproxy -T`
2. `curl -x localhost:8080 example.com`

mitmproxy should provide a warning that you want to run either transparent mode or configure an explicit proxy.
##### What went wrong?

mitmproxy complains about "Invalid HTTP request form (expected: relative, got: authority)"

We should just detect this very specific case (expected: relative, got: authority) and provide a better error message for it.

See #1579 for a confused user. :stuck_out_tongue:

---

Mitmproxy Version: master
Operating System: Ubuntu 16.04

ujjwal96:
Checks if mode is transparent and request is of authority or absolute form and gives the message accordingly. I tried to find out the best way to check if an explicit proxy is actually set or not but couldn't find anything.  
One thing which I noticed though was, when I did 
`curl -i -v -x localhost:8080 example.com`
`Proxy-Connection: Keep-Alive` header was being sent 
If you want I can check for that instead of just checking whether mode is transparent or not or check for both
Change in line 129 because:
http://stackoverflow.com/a/3511300

### DIFF
--- a/mitmproxy/proxy/protocol/http.py
+++ b/mitmproxy/proxy/protocol/http.py
@@ -145,7 +145,7 @@ def validate_request_form(mode, request):
     if request.first_line_format not in allowed_request_forms:
         if mode == HTTPMode.transparent:
             err_message = (
-                "Mitmproxy received an {} request even though it is not running in regular mode. 
+                "Mitmproxy received an {} request even though it is not running in regular mode. "
                 "This usually indicates a misconfiguration, please see "
                 "http://docs.mitmproxy.org/en/stable/modes.html for details."
             ).format("HTTP CONNECT" if request.first_line_format == "authority" else "absolute-form")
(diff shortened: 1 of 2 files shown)


mhils:
Thanks for the PR and listing alternatives to the method you proposed. 👍 
The current check looks good to me, but I'm not entirely happy with the error message. How about something like this?

> Mitmproxy received an (absolute-form|HTTP CONNECT) request even though it is not running in regular mode. This usually indicates a misconfiguration, please see http://docs.mitmproxy.org/en/stable/modes.html for details.

ujjwal96:
Rather than giving the above URL wouldn't this - http://docs.mitmproxy.org/en/stable/transparent.html be more apt? Since the person has not configured the transparent mode correctly.

mhils:
This usually happens to users who confuse regular and transparent mode.
Most of the times, they just specify `-T` without having to. Let's link to
the modes doc! :)

Ujjwal Verma <notifications@github.com> schrieb am Mi., 8. Feb. 2017, 15:58:

ujjwal96:
Done :slightly_smiling_face:

mhils:
Thanks. Can we please add a test for this? :blush:

ujjwal96:
@mhils Quote not closed :see_no_evil:

mhils:
Here we go, thanks!

PR #2000, opened 2017-02-08
Add jpeg parser

s4chin:
Updated the kaitaistruct version here


Kriechi:
I don't see any jpeg-related changes in this PR...?

s4chin:
There aren't any yet. I'm working on this :)

mitmproxy/contrib/kaitaistruct/exif.py
```diff
@@ -0,0 +1,573 @@
...
+
+import array
+import struct
+import zlib
+from enum import Enum
+
+from kaitaistruct import KaitaiStruct, KaitaiStream, BytesIO
+
+
+class Exif(KaitaiStruct):
```
s4chin: The only variables I get from the `app1` segment are the ones in the `__init__`. The `exif` data is never parsed.

s4chin:
The test fails because the `app13` segment hasn't been specified in the declaration. I'll write the complete declaration later and will try to merge it in `kaitaistruct` before using it here. But my problem with the `app1` segment remains.

Kriechi:
@s4chin looks good! Unfortunately this block of code is not covered any more with tests:
https://github.com/s4chin/mitmproxy/blob/8eb1d34644c454a79c168f6d68fc0ae45bbc4d1a/mitmproxy/contentviews/image/view.py#L55-L62
If it is still needed, please add tests. Otherwise just delete the whole thing.

s4chin:
Done!

test/mitmproxy/contentviews/test_image_parser.py
```diff
@@ -104,3 +104,50 @@ def test_parse_png(filename, metadata):
...
+        ('jfif_version', '(1, 1)'),
+        ('jfif_density', '(72, 72)'),
+        ('jfif_unit', '1'),
+        ('make', 'Canon'),
+        ('model', 'Canon PowerShot A60'),
+        ('modify_date', '2004:07:16 18:46:04'),
+        ('Size', '717 x 558 px')
+    ],
+    # check multiple segments
+    "mitmproxy/data/image_parser/image-err1.jpg": [
```
mhils: This picture is copyrighted (http://www.gettyimages.de/license/93901920), do we have another one which could replace this test case?

s4chin: I just used the image which was being used in `test_image.py`

PR #2001, opened 2017-02-08
fix response body edit issue

cle1000:


mhils:
:+1:

This fixes mitmproxy/mitmweb#40

PR #2002, opened 2017-02-08
Content views edit

cle1000:
there is a bug with the "show full content" button. if you activate the edit mode, the whole content is visible in the code editor, but if you switch between the tab (response, request,...) then the redux state will change to showFullContent=false.

the second commit removes the content  view selector, because if you modify a content without saving it you cannot use the server side content views correctly.


mhils:
Awesome, this looks great. 👍

PR #2003, opened 2017-02-09
test refactoring and coverage++

Kriechi:


mitmproxy/eventsequence.py
```diff
@@ -70,4 +70,4 @@ def event_sequence(f):
             yield "tcp_error", f
         yield "tcp_end", f
     else:
-        raise NotImplementedError
+        raise ValueError
```
mhils: Wouldn't TypeError be more appropriate?
(this should also raise an instance, so let's add parantheses)

mitmproxy/tcp.py
```diff
@@ -47,5 +50,10 @@ def __init__(self, client_conn, server_conn, live=None):
         messages=List[TCPMessage]
     )
 
+    def copy(self):
```
mhils: This should not be required, the [default implementation](https://github.com/mitmproxy/mitmproxy/blob/fa661217c12733da4afe500b51a253f5994f3506/mitmproxy/types/serializable.py#L32) should work?

mitmproxy/version.py
```diff
@@ -3,5 +3,5 @@
 PATHOD = "pathod " + VERSION
 MITMPROXY = "mitmproxy " + VERSION
 
-if __name__ == "__main__":
+if __name__ == "__main__":  # pragma: no cover
```
mhils: We could also test this easily with `runpy.run_module`, what do you think?

mitmproxy/tcp.py
```diff
@@ -9,16 +9,19 @@
...
 
     def __init__(self, from_client, content, timestamp=None):
-        self.content = content
         self.from_client = from_client
+        self.content = content
         self.timestamp = timestamp or time.time()
 
     @classmethod
     def from_state(cls, state):
-        return cls(*state)
```
mhils: Why are we changing this for tcp and websockets? 
The tuple for is more compact on disk, and we have a lot of those objects. Otherwise we could just have used stateobject.

mitmproxy/websocket.py
```diff
@@ -80,8 +72,33 @@ def __init__(self, client_conn, server_conn, handshake_flow, live=None):
...
+        f = super().copy()
+        f.messages = self.messages.copy()
+        f.handshake_flow = self.handshake_flow.copy()
+        return f
+
     def __repr__(self):
-        return "WebSocketFlow ({} messages)".format(len(self.messages))
+        return "<WebSocketFlow ({} messages)>".format(len(self.messages))
+
+    def message_info(self, message):
```
mhils: This would be good candidate to add type declarations. 😇

mitmproxy/websocket.py
```diff
@@ -80,8 +72,33 @@ def __init__(self, client_conn, server_conn, handshake_flow, live=None):
...
         handshake_flow=HTTPFlow,
+        client_key=bytes,
+        client_protocol=bytes,
+        client_extensions=bytes,
+        server_accept=bytes,
+        server_protocol=bytes,
+        server_extensions=bytes,
     )
 
+    def copy(self):
```
mhils: Default copy should work?

mitmproxy/websocket.py
```diff
@@ -11,34 +11,26 @@
...
         self.from_client = from_client
+        self.content = content
         self.timestamp = timestamp or time.time()
 
     @classmethod
     def from_state(cls, state):
-        return cls(*state)
+        f = cls(None, None)
+        f.set_state(state)
+        return f
```
mhils: (see above)

mitmproxy/tcp.py
Kriechi: All other objects return a dict - it seemed inconsistent to return a tuple here.

mitmproxy/websocket.py
Kriechi: No, I would get back the same message object instance - which I don't want. I want a deep-copy.

mitmproxy/tcp.py
Kriechi: No, I would get back the same message object instance - which I don't want. I want a deep-copy.

mitmproxy/eventsequence.py
Kriechi: 👍

mitmproxy/version.py
Kriechi: 👍

mitmproxy/tcp.py
mhils: The default implementation should give a deep copy, so if it doesn't do that, it's a bug.

PR #2005, opened 2017-02-10
Closes #1828 script reloads on py file changes

### LINKED ISSUE #1828: How to configure file system watchdog (away) for attached scripts

ujjwal96:
https://github.com/mitmproxy/mitmproxy/pull/1911 needed love.
Should I change the docs as well?
First line in [this](https://github.com/mitmproxy/mitmproxy/blob/master/docs/scripting/overview.rst#developing-scripts) should be more clear, right?

### DIFF
--- a/mitmproxy/addons/script.py
+++ b/mitmproxy/addons/script.py
@@ -110,11 +110,16 @@ class ReloadHandler(watchdog.events.FileSystemEventHandler):
         self.callback = callback
 
     def filter(self, event):
+        """
+            Returns True only when .py file is changed
+        """
         if event.is_directory:
             return False
         if os.path.basename(event.src_path).startswith("."):
             return False
-        return True
+        if event.src_path.endswith(".py"):
+            return True
+        return False
 
     def on_modified(self, event):
         if self.filter(event):
--- a/test/mitmproxy/addons/test_script.py
+++ b/test/mitmproxy/addons/test_script.py
@@ -44,14 +44,19 @@ def test_reloadhandler():
     rh = script.ReloadHandler(Called())
     assert not rh.filter(watchdog.events.DirCreatedEvent("path"))
     assert not rh.filter(watchdog.events.FileModifiedEvent("/foo/.bar"))
-    assert rh.filter(watchdog.events.FileModifiedEvent("/foo/bar"))
+    assert not rh.filter(watchdog.events.FileModifiedEvent("/foo/bar"))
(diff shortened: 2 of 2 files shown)


mhils:
> "Mitmproxy monitors scripts for modifications, and reloads them on change."

This still is the case, so I think we're good here. Thanks! :smiley:

PR #2006, opened 2017-02-10
coverage++

Kriechi:


mitmproxy/net/http/request.py
```diff
@@ -396,10 +396,7 @@ def multipart_form(self):
     def _get_multipart_form(self):
         is_valid_content_type = "multipart/form-data" in self.headers.get("content-type", "").lower()
         if is_valid_content_type:
-            try:
```
mhils: Are you sure this cannot fail?

PR #2008, opened 2017-02-11
Add transparent proxy support for OpenBSD

yjh0502:
Add transparent proxy support for OpenBSD with pf divert-to.

### DIFF
--- a/docs/transparent/openbsd.rst
+++ b/docs/transparent/openbsd.rst
@@ -1,0 +1,53 @@
+.. _openbsd:
+
+OpenBSD
+=======
+
+ 1. :ref:`Install the mitmproxy certificate on the test device <certinstall>`
+
+ 2. Enable IP forwarding:
+
+    >>> sudo sysctl -w net.inet.ip.forwarding=1
+
+ 3. Place the following two lines in **/etc/pf.conf**:
+
+    .. code-block:: none
+
+        mitm_if = "re2"
+        pass in quick proto tcp from $mitm_if to port { 80, 443 } divert-to 127.0.0.1 port 8080
+
+    These rules tell pf to divert all traffic from ``$mitm_if`` destined for
+    port 80 or 443 to the local mitmproxy instance running on port 8080. You
+    should replace ``$mitm_if`` value with the interface on which your test
+    device will appear.
+
+ 4. Configure pf with the rules:
+
+    >>> doas pfctl -f /etc/pf.conf
+
+ 5. And now enable it:
+
+    >>> doas pfctl -e
+
+ 6. Fire up mitmproxy. You probably want a command like this:
+
+    >>> mitmproxy -T --host
+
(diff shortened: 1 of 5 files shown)


docs/transparent/openbsd.rst
```diff
@@ -0,0 +1,53 @@
...
+    running as the default gateway.
+
+.. note::
+
+    Note that the **divert-to** rules in the pf.conf given above only apply to
+    inbound traffic. **This means that they will NOT redirect traffic coming
+    from the box running pf itself.** We can't distinguish between an outbound
+    connection from a non-mitmproxy app, and an outbound connection from
+    mitmproxy itself - if you want to intercept your traffic, you should use an
+    external host to run mitmproxy. None the less, pf is flexible to cater for
```
dwfreed: "Nonetheless" should be a single word here.

yjh0502: @dwfreed I copy-pasted the text from `osx.rst`, so I fixed both documents and pushed.

yjh0502:
Here's some notes.

### implementation
There is other way to implement transparent proxy on OpenBSD, by imitating macOS/FreeBSD implementation by using rdr-to and nat-to rule. Here's pf rule for rdr/nat implementation.
```
pass out on egress proto tcp from $mitm_if to port { 80, 443 } nat-to 127.0.0.1 rdr-to 127.0.0.1 port 8080
```
The rule first NAT outgoing connection to `127.0.0.1` and then redirect traffic to `127.0.0.1:8080` so that mitmproxy can receive the traffic. mitmproxy can inspect destination host by using `pfctl -s state` command (just like how mitmproxy does on osx and FreeBSD) and connect to remote host. This approach has following problems.
 - Recent OpenBSD replaces `sudo` to `doas`, so `platform/osx.py` should be updated to use `doas` on OpenBSD or user should install `sudo` from ports.
 - `mitmproxy/platform/pf.py` should be changed to deal with OpenBSD pfctl output.
 - `pfctl -s state` will be executed on every new connection, which seems quite inefficient.
 - It uses two ports on `127.0.0.1` for each connections, so port space can be exhausted if there are lots of clients/connections. The problem can be avoided by using multiple address and NAT address pools, but it requires additional configuration. 

### `divert-to`
`divert-to` seems to be supported on FreeBSD PF[1], so FreeBSD implementation/documentation can be changed to use `divert-to` rules, which doesn't require `pfctl` to inspect pf states. Unfortunately `divert-to` is not supported on macOS.

[1] https://www.freebsd.org/cgi/man.cgi?query=pf.conf&sektion=5
[2] https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man5/pf.conf.5.html

mhils:
This looks fantastic, thanks! `divert-to` is indeed way better than `pfctl -s state` when available, so I strongly agree with your thoughts.

Thanks! 🎉 🍰 😃

PR #1991, opened 2017-02-05
Added version in web interface footer

mosajjal:
minimal approach of adding version number in the web interface. 
fixes #1893

PR #2009, opened 2017-02-11
added version info to web interface footer

mosajjal:
I've used `gulp` to create other files after changing the jsx file. Don't know why there's all these changes. Maybe gulp version?

we can close #1991 now.


mhils:
Thanks. We don't update `mitmproxy/tools/web/static` with every commit. Could you remove those files from the PR please?

mosajjal:
I'll create a new clean PR with 1 commit for this one.

PR #2011, opened 2017-02-11
Coverage++

Kriechi:
This PR completes the restructuring and linting checks from earlier PRs.
Mostly moving stuff around.


mitmproxy/contentviews/base.py
```diff
@@ -1,13 +1,23 @@
...
+from mitmproxy.net import http
+from mitmproxy.utils import strutils
+
 
+VIEW_CUTOFF = 512
 KEY_MAX = 30
 
+views = []  # type: List[View]
+content_types_map = {}  # type: Dict[str, List[View]]
+view_prompts = []  # type: List[Tuple[str, str]]
```
mhils: Not sure if this belongs together. `base` previously was stuff that was used by views only, whereas `__init__` contained the external API. Maybe move it to `viewmanager` or so?

mitmproxy/utils/bits.py
```diff
@@ -1,13 +0,0 @@
-def setbit(byte, offset, value):
```
Kriechi: No other parts in the entire code base use these functions.
And we didn't have test for it either...

mitmproxy/contentviews/base.py
Kriechi: Yes this is kinda unclear. My intention was that there should be no code in `__init__.py` files, except for imports and the `__all__` thing. I wasn't aware we had a viewmanage - I will take a look and maybe move it there...

mhils: We don't have `viewmanager` yet,  but it'd make sense to create one for putting the stuff in. I generally agree on the move out of __init__ fwiw. :)

mitmproxy/contentviews/__init__.py
```diff
@@ -173,11 +43,23 @@ def get_content_view(viewmode: View, data: bytes, **metadata):
 add(image.ViewImage())
 add(query.ViewQuery())
 
-if protobuf.ViewProtobuf.is_available():
+if protobuf.ViewProtobuf.is_available():  # pragma: no cover
```
Kriechi: Regardless of the protobuf-view quality, it was never available - I guess travis misses a package or something? Or does this only work if one manually installs the package?

Kriechi:
Ok, I will refactor the individual-coverage check to be similarly configurable as the full-coverage task.

mitmproxy/contentviews/__init__.py
mhils: The view shells out to `protoc`, which needs to be on PATH (this is what this function checks).
Default install does not add it to the PATH. We should use the protobuf library directly, so there's no excuse for hiding missing coverage here.

mitmproxy/utils/bits.py
mhils: I'd still say those are generic enough to be in their own module. 🙂

Kriechi: Feel free to write tests, move them back again, and make it 100% covered. 😃

Kriechi:
Refactor complete. Please take another look at this PR.
If needed, I could try and split out a few things to make the diff more readable...

mhils:
Comments from https://github.com/mitmproxy/mitmproxy/pull/2011#pullrequestreview-21480384 are still unaddressed I think. :)

Kriechi:
Coverage is already enforced - in a previous PR.
I will leave the extraction of a `viewmanager` for a future PR.
Anything still not addressed?

mhils:
Status? :)

PR #2017, opened 2017-02-12
Catch ValueErrors from url.parse()

lymanZerga11:
#1975 had to delete the previous branch

Catch ValueErrors from calls to mitmproxy.net.http.url.parse(). Otherwise mitmproxy crashes due to uncaught exception.

### DIFF
--- a/test/mitmproxy/console/test_flowlist.py
+++ b/test/mitmproxy/console/test_flowlist.py
@@ -2,7 +2,7 @@ import mitmproxy.tools.console.flowlist as flowlist
 from mitmproxy.tools import console
 from mitmproxy import proxy
 from mitmproxy import options
-from .. import mastertest
+from .. import tservers
 import pytest
 from unittest import mock
 
@@ -15,7 +15,7 @@ def mock_add_log(message):
     raise ScriptError(message)
 
 
-class TestFlowlist(mastertest.MasterTest):
+class TestFlowlist(tservers.MasterTest):
     def mkmaster(self, **opts):
         if "verbosity" not in opts:
             opts["verbosity"] = 1
--- a/test/mitmproxy/console/test_flowlist.py
+++ b/test/mitmproxy/console/test_flowlist.py
@@ -1,19 +1,19 @@
 import mitmproxy.tools.console.flowlist as flowlist
-from mitmproxy.test import tutils
-from mitmproxy.test import tflow
 from mitmproxy.tools import console
 from mitmproxy import proxy
 from mitmproxy import options
-from mitmproxy.tools.console import common
 from .. import mastertest
 import pytest
 from unittest import mock
 
+
 class ScriptError(Exception):
     pass
 
+
 def mock_add_log(message):
     raise ScriptError(message)
+
 
 class TestFlowlist(mastertest.MasterTest):
     def mkmaster(self, **opts):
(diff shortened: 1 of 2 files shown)


test/mitmproxy/console/test_flowlist.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class TestFlowlist(tservers.MasterTest):
+    def mkmaster(self, **opts):
+        if "verbosity" not in opts:
+            opts["verbosity"] = 1
+        o = options.Options(**opts)
+        return console.master.ConsoleMaster(o, proxy.DummyServer())
+
+    @mock.patch('mitmproxy.tools.console.signals.status_message.send', side_effect = mock_log)
```
Kriechi: It is probably easier if you directly check that `signals.status_message.send` was called with specific arguments, instead of using a side effect + exception wrapper.
E.g., `Mock.assert_called_once_with` or similar.

lymanZerga11: We learn something new everyday haha  :)

test/mitmproxy/console/test_flowlist.py
```diff
@@ -4,16 +4,7 @@
...
-
-
-class UrlError(Exception):
-    pass
-
-
-def mock_log(message):
-    if "Invalid URL" in message:
-        raise UrlError(message)
+from unittest import mock as Mock
```
Kriechi: You don't have to rename the import - my comment from before was just an example.

test/mitmproxy/console/test_flowlist.py
```diff
@@ -23,9 +14,9 @@ def mkmaster(self, **opts):
...
 
-    @mock.patch('mitmproxy.tools.console.signals.status_message.send', side_effect = mock_log)
-    def test_new_request(self, test_func):
+    def test_new_request(self):
         m = self.mkmaster()
         x = flowlist.FlowListBox(m)
-        with pytest.raises(UrlError):
+        with Mock.patch('mitmproxy.tools.console.signals.status_message.send') as mock:
             x.new_request("nonexistent url", "GET")
+        mock.assert_called_once_with(message = "Invalid URL: No hostname given")
```
Kriechi: Please remove the spaces around the `=`.

test/mitmproxy/console/test_flowlist.py
```diff
@@ -23,9 +14,9 @@ def mkmaster(self, **opts):
         o = options.Options(**opts)
         return console.master.ConsoleMaster(o, proxy.DummyServer())
 
-    @mock.patch('mitmproxy.tools.console.signals.status_message.send', side_effect = mock_log)
-    def test_new_request(self, test_func):
+    def test_new_request(self):
         m = self.mkmaster()
         x = flowlist.FlowListBox(m)
-        with pytest.raises(UrlError):
+        with Mock.patch('mitmproxy.tools.console.signals.status_message.send') as mock:
```
Kriechi: Call it `mock_thing` instead of `mock`, then it doesn't collide with the import above.
This is how the official Python docs call it most of the time.

test/mitmproxy/console/test_flowlist.py
```diff
@@ -0,0 +1,21 @@
+import mitmproxy.tools.console.flowlist as flowlist
+from mitmproxy.tools import console
+from mitmproxy import proxy
+from mitmproxy import options
+from .. import tservers
+from unittest import mock
+
+
+class TestFlowlist(tservers.MasterTest):
```
Kriechi: Do we even need the `tservers.MasterTest` parent? We are building our own master anyway...
If not, please remove it.

PR #2018, opened 2017-02-12
store MITM cert so that TCP/TLS hooks can access them (fixes #1935)

### LINKED ISSUE #1935: How can I get the MITM certificate from a flow in a script hook?
I'm using mitmdump 1.0.2 to intercept a TLS connection and log and modify it with TCP hooks (`--tcp foo.host.com -s script.py`).

I need to find and replace the hex/ASCII digest of the *upstream* server certificate with the digest of mitmproxy's *generated* certificate—because the client expects to receive and check the certificate hash at the application level.

I can easily get the real/upstream server certificate from `flow.server_conn.cert`. But how can I get the generated/MITM cert sent to the client in the TCP hooks?

```python
def tcp_message(flow):
    d = flow.dumpfile
    m = flow.messages[-1]
    c = m.content

    real_server_hash = flow.server_conn.cert.digest('sha1').replace(b':',b'').lower()

    #####
    # FIXME: how can I figure out the hash of the certificate that mitmproxy is sending to the client?
    mitm_server_hash = b'hardcodedvaluehere'
    #####

    nc = m.content
    nc = nc.replace(real_server_hash, mitm_server_hash)
    if nc!=m.content:
        print("%s message: replaced real cert hash (%r) with mitmproxy cert hash (%r)" % ('outgoing' if m.from_client else 'incoming', real_server_hash, mitm_server_hash), file=stderr)
        m.content = nc
```

dlenski:
In #1935, @dlenski wrote:
> @mhils wrote:
> > I'm wondering if there is a general use-case here that warrants adding
> > this to the ServerConnection properly so that it does get persisted.
>
> In my experience reverse engineering VPNs, peers often exchange
> certificates or digests at the application layer.  It's often useful to
> know what MITM certificate we're sending to the client, so that we can
> inject it at the app layer as well.


mitmproxy/connections.py
```diff
@@ -136,6 +136,7 @@ class ServerConnection(tcp.TCPClient, stateobject.StateObject):
         source_address: Local IP address or client's source IP address.
         ssl_established: True if TLS is established, False otherwise
         cert: The certificate presented by the remote during the TLS handshake
+        mimtcert: The certificate presented by mitmproxy to the client
```
Kriechi: `s/mimtcert/mitmcert/`

dlenski:
Several of the failing tests are in `test/mitmproxy/test_flow_format_compat.py`, because the dumpfiles don't include the (now-expected) `server_conn.mitmcert` field.

I'm unsure of how to correctly regenerate the dumpfiles for the tests. Any suggestions?

mhils:
Thanks. Please don't modify the dump files directly, take a look at [io_compat.py#L84](https://github.com/mitmproxy/mitmproxy/blob/75a0a4c09276ef2aa479c9ae9feab97f7803ed6d/mitmproxy/io_compat.py#L84) instead.

dlenski:
Aha, thanks @mhils… does that mean this PR would need to introduce a new flow format version number, because it adds a new `mitmcert` field? 

What should I call it? `(2,1,0)`?

Kriechi:
You can use the latest version 2.0.0 which is not released yet.
https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/io_compat.py#L89-L91

Kriechi:
Are you sure `server_conn` is the right place to put it?
IMO the cert that is generated is between the client and mitmproxy, so I would guess `client_conn` makes more sense? Or which cert are you actually looking for? The name `mitmcert` suggests to me, that this is the on-the-fly generated cert used between client<->mitmproxy...

dlenski:
>The name mitmcert suggests to me, that this is the on-the-fly generated cert used between client<->mitmproxy...

That is correct.

> Are you sure server_conn is the right place to put it?
IMO the cert that is generated is between the client and mitmproxy, so I would guess client_conn makes more sense?

To my mind, `server_conn` seemed like the more natural place to put it, because it is the cert which _replaces_ the real server's cert. Though it's a good point that the server doesn't actually know anything about this certificate.

Should I change it?

Kriechi:
> because it is the cert which replaces the real server's cert
Ok, this makes sense - but I didn't think of it, so I would move it to the `client_conn` and call it `generated_cert` or something.
@mitmproxy/devs any ideas?

dlenski:
> Ok, this makes sense - but I didn't think of it, so I would move it to the client_conn and call it generated_cert or something.
@mitmproxy/devs any ideas?

Okay, I'll move it. In some cases (as with `--cert` on the command line) it won't actually be _generated_ by mitmproxy at all.

Based on my own experience, I thought something like this would be most likely recognized by developers:

```
        mitmcert: The MITM'ed TLS server certificate presented to the client
```

dlenski:
I'm still having trouble understanding why the `test_replay` tests are failing. Any hints on these?

mitmproxy/connections.py
```diff
@@ -38,6 +39,7 @@ def __init__(self, client_connection, address, server):
             self.rfile = None
             self.address = None
             self.clientcert = None
+            self.mitmcert = None
```
Kriechi: Your failing tests a probably because of this here.
Try moving it one indentation level out (and down to the others), then it sets this attribute always - and not only inside this if.

dlenski:
Looks like everything is passing now except for one Travis-CI build that failed on MacOS due to not being able to install build requirements (???).

Kriechi:
I restarted the test job - now it passes!
@mhils good to merge?

mitmproxy/proxy/protocol/tls.py
```diff
@@ -465,6 +465,8 @@ def _establish_tls_with_client(self):
         self.log("Establish TLS with client", "debug")
         cert, key, chain_file = self._find_cert()
 
+        self.client_conn.mitmcert = cert
```
mhils: Can we move this [here](https://github.com/mitmproxy/mitmproxy/blob/4cec88fc7fcf8a5c16c0639961c27792cd73af13/mitmproxy/connections.py#L113)?

Kriechi:
Merged in 391f28f78cf9f4ee2f82c69c7f5ce370b69b77cd.

Thanks! 🎉 👍 👏

Issue #1946, opened 2017-01-20
Use existing urlencoding conventions when re-encoding edited form


mame82:
mitmproxy 1.0.2 crashes trying to edit form data of facebook ajax POST request to facebook.com/ajax/updatestatus.php

##### Steps to reproduce the problem:

1. run mitmproxy with
```mitmproxy -p <listenport> -T --cadir <dir with cert> --host -f "updatestatus | sl/password" -i updatestatus```
2. Post a message on facebook (acces https://www.facebook.com/ajax/updatestatus.php)
3. Try to edit the intercepted flow body (key e) using form data view (key f)
4. results in crash

Crash trace:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/mitmproxy/tools/console/master.py", line 292, in run
    self.loop.run()
  File "/usr/local/lib/python3.5/dist-packages/urwid/main_loop.py", line 278, in run
    self._run()
  File "/usr/local/lib/python3.5/dist-packages/urwid/main_loop.py", line 376, in _run
    self.event_loop.run()
  File "/usr/local/lib/python3.5/dist-packages/urwid/main_loop.py", line 682, in run
    self._loop()
  File "/usr/local/lib/python3.5/dist-packages/urwid/main_loop.py", line 719, in _loop
    self._watch_files[fd]()
  File "/usr/local/lib/python3.5/dist-packages/urwid/raw_display.py", line 393, in <lambda>
    event_loop, callback, self.get_available_raw_input())
  File "/usr/local/lib/python3.5/dist-packages/urwid/raw_display.py", line 493, in parse_input
    callback(processed, processed_codes)
  File "/usr/local/lib/python3.5/dist-packages/urwid/main_loop.py", line 403, in _update
    self.process_input(keys)
  File "/usr/local/lib/python3.5/dist-packages/urwid/main_loop.py", line 503, in process_input
    k = self._topmost_widget.keypress(self.screen_size, k)
  File "/usr/local/lib/python3.5/dist-packages/mitmproxy/tools/console/window.py", line 84, in keypress
    k = super().keypress(size, k)
  File "/usr/local/lib/python3.5/dist-packages/urwid/container.py", line 1116, in keypress
    return self.footer.keypress((maxcol,),key)
  File "/usr/local/lib/python3.5/dist-packages/mitmproxy/tools/console/statusbar.py", line 155, in keypress
    return self.master.ab.keypress(*args, **kwargs)
[...]

PR #2022, opened 2017-02-14
Use existing urlencoding conventions when re-encoding edited form #1946

mame82:
Handles corner case discussed in issue [#1946](https://github.com/mitmproxy/mitmproxy/issues/1946)

Only considers requests (no responses)

Works for "https://www.facebook.com/ajax/updatestatus.php"

### DIFF
--- a/mitmproxy/net/http/request.py
+++ b/mitmproxy/net/http/request.py
@@ -426,8 +426,6 @@ class Request(message.Message):
         self.headers["content-type"] = "application/x-www-form-urlencoded"
         #self.content = mitmproxy.net.http.url.encode(form_data).encode()
         self._sanitized_form_data_replacement(form_data)
-        # only raw_content should be URL encoded, content should be decoded
-
         
     @urlencoded_form.setter
     def urlencoded_form(self, value):
--- a/mitmproxy/net/http/request.py
+++ b/mitmproxy/net/http/request.py
@@ -367,64 +367,13 @@ class Request(message.Message):
                 pass
         return ()
 
-    def _sanitized_form_data_replacement(self, form_data):
-        # check if unmodified content changes with URL encoding
-        url_decoded_content = urllib.parse.parse_qsl(self.content, keep_blank_values=True)
-        url_reencoded_content = urllib.parse.urlencode(url_decoded_content, doseq=True).encode()
-        if self.content == url_reencoded_content:
-            # encoding the unmodified content doesn't change anything, so we should be good to go on as usual
-            self.content = mitmproxy.net.http.url.encode(form_data).encode()
-        else:
-            # if we are here, we have to alter encoding behavior to mimic the one used in unmodified POST request
-
-            # check if content isn't encoded at all
-            if self.content == urllib.parse.unquote(url_reencoded_content.decode()).encode():
-                # if we land here, the new content mustn't be URL encoded thus we decode the result before setting new content
-                # this is the case for "https://www.facebook.com/ajax/updatestatus.php" but additionally the "=" sign is
-                # omitted on SOME empty keys
-                self.content = urllib.parse.unquote(mitmproxy.net.http.url.encode(form_data)).encode()
-            # check if content isn't encoded at all and equal signs are omitted
-            elif self.content == urllib.parse.unquote(url_reencoded_content.decode()).replace("=&", "&").encode():
-                # remove equal sign for keys without value
-                self.content = urllib.parse.unquote(mitmproxy.net.http.url.encode(form_data)).replace("=&", "&").encode()
-            # the next check should address the facebook issue, which is an absolute weird corner case:
-            # here's a real, but trimmed down data section of a POST request to "https://www.facebook.com/ajax/updatestatus.php"
-            # with content-type 'application/x-www-form-urlencoded'
-            #
-            # self.content:
-            # b'attachment&backdated_date[year]&composer_source_surface=newsfeed&multilingual_specified_lang=&__pc=PHASED%3ADEFAULT'
-            #    * the first two keys have no value, but no equal sign appended
-            #    * the second key shows that no URL encoding ist used ("[" and "]")
-            #    * the third key shows a key=value pair
-            #    * the next key "multilingual_specified_lang" again has no value, but this time an equal sign is appended
-            #    * the last key has assigned a value, which is URL encoded ("%3A" == ":")
-            #
-            # To cope with that, a somehow hacky approach is used
-            #    If unmodified and unquoted POST data contains ONE SINGLE key without a value, which hasn't an equal sign appended
-            #    we remove all equal signs on empty keys in modified data
-            elif any("=" not in param for param in urllib.parse.unquote(self.content.decode()).split("&")):
-                # if we are here, there was at least one key without value but no equal sign appended in unquoted POST form data
-                # thus we remove "=" for all keys without value
-
-                # unquote
-                modified_content = mitmproxy.net.http.url.encode(form_data)
-                # remove equal sign
-                modified_content = modified_content.replace("=&", "&").encode()
-                self.content = modified_content
-
-            # additional corner cases could be added to with new "elif" statements
-            else:
-                # if we are here we trigger default behavior again
-                self.content = mitmproxy.net.http.url.encode(form_data).encode()
-
     def _set_urlencoded_form(self, form_data):
         """
         Sets the body to the URL-encoded form data, and adds the appropriate content-type header.
         This will overwrite the existing content if there is one.
         """
         self.headers["content-type"] = "application/x-www-form-urlencoded"
-        # self.content = mitmproxy.net.http.url.encode(form_data).encode()
-        self._sanitized_form_data_replacement(form_data)
+        self.content = mitmproxy.net.http.url.encode(form_data, self.content.decode()).encode()
 
     @urlencoded_form.setter
     def urlencoded_form(self, value):
--- a/mitmproxy/net/http/request.py
+++ b/mitmproxy/net/http/request.py
@@ -366,33 +366,33 @@ class Request(message.Message):
             except ValueError:
                 pass
         return ()
-    
+
     def _sanitized_form_data_replacement(self, form_data):
-        # create backup of unmodified content
-        backup = self.content
         # check if unmodified content changes with URL encoding
-        url_decoded_content = urllib.parse.parse_qsl(self.content,keep_blank_values=True)
+        url_decoded_content = urllib.parse.parse_qsl(self.content, keep_blank_values=True)
         url_reencoded_content = urllib.parse.urlencode(url_decoded_content, doseq=True).encode()
         if self.content == url_reencoded_content:
             # encoding the unmodified content doesn't change anything, so we should be good to go on as usual
             self.content = mitmproxy.net.http.url.encode(form_data).encode()
         else:
             # if we are here, we have to alter encoding behavior to mimic the one used in unmodified POST request
-            
-            #check if content isn't encoded at all
+
+            # check if content isn't encoded at all
             if self.content == urllib.parse.unquote(url_reencoded_content.decode()).encode():
                 # if we land here, the new content mustn't be URL encoded thus we decode the result before setting new content
-                # this is the case for "https://www.facebook.com/ajax/updatestatus.php" but additionally the "=" sign is omitted on SOME empty keys
+                # this is the case for "https://www.facebook.com/ajax/updatestatus.php" but additionally the "=" sign is
+                # omitted on SOME empty keys
                 self.content = urllib.parse.unquote(mitmproxy.net.http.url.encode(form_data)).encode()
             # check if content isn't encoded at all and equal signs are omitted
-            elif self.content == urllib.parse.unquote(url_reencoded_content.decode()).replace("=&","&").encode():
+            elif self.content == urllib.parse.unquote(url_reencoded_content.decode()).replace("=&", "&").encode():
                 # remove equal sign for keys without value
-                self.content = urllib.parse.unquote(mitmproxy.net.http.url.encode(form_data)).replace("=&","&").encode()
+                self.content = urllib.parse.unquote(mitmproxy.net.http.url.encode(form_data)).replace("=&", "&").encode()
             # the next check should address the facebook issue, which is an absolute weird corner case:
             # here's a real, but trimmed down data section of a POST request to "https://www.facebook.com/ajax/updatestatus.php"
             # with content-type 'application/x-www-form-urlencoded'
             #
-            # self.content: b'attachment&backdated_date[year]&composer_source_surface=newsfeed&multilingual_specified_lang=&__pc=PHASED%3ADEFAULT'
+            # self.content:
+            # b'attachment&backdated_date[year]&composer_source_surface=newsfeed&multilingual_specified_lang=&__pc=PHASED%3ADEFAULT'
             #    * the first two keys have no value, but no equal sign appended
             #    * the second key shows that no URL encoding ist used ("[" and "]")
             #    * the third key shows a key=value pair
@@ -402,21 +402,20 @@ class Request(message.Message):
             # To cope with that, a somehow hacky approach is used
             #    If unmodified and unquoted POST data contains ONE SINGLE key without a value, which hasn't an equal sign appended
             #    we remove all equal signs on empty keys in modified data
-            elif any ("=" not in param for param in urllib.parse.unquote(self.content.decode()).split("&")):
+            elif any("=" not in param for param in urllib.parse.unquote(self.content.decode()).split("&")):
                 # if we are here, there was at least one key without value but no equal sign appended in unquoted POST form data
                 # thus we remove "=" for all keys without value
-                
+
                 # unquote
-                modified_content=mitmproxy.net.http.url.encode(form_data)
+                modified_content = mitmproxy.net.http.url.encode(form_data)
                 # remove equal sign
-                modified_content=modified_content.replace("=&","&").encode()
+                modified_content = modified_content.replace("=&", "&").encode()
                 self.content = modified_content
-                
-            # additional corner cases could be added to with new "elif" statements    
+
+            # additional corner cases could be added to with new "elif" statements
             else:
                 # if we are here we trigger default behavior again
                 self.content = mitmproxy.net.http.url.encode(form_data).encode()
-                
 
     def _set_urlencoded_form(self, form_data):
         """
@@ -424,9 +423,9 @@ class Request(message.Message):
         This will overwrite the existing content if there is one.
         """
         self.headers["content-type"] = "application/x-www-form-urlencoded"
-        #self.content = mitmproxy.net.http.url.encode(form_data).encode()
+        # self.content = mitmproxy.net.http.url.encode(form_data).encode()
         self._sanitized_form_data_replacement(form_data)
-        
+
     @urlencoded_form.setter
     def urlencoded_form(self, value):
         self._set_urlencoded_form(value)
--- a/mitmproxy/net/http/request.py
+++ b/mitmproxy/net/http/request.py
@@ -366,6 +366,57 @@ class Request(message.Message):
             except ValueError:
                 pass
         return ()
+    
+    def _sanitized_form_data_replacement(self, form_data):
+        # create backup of unmodified content
+        backup = self.content
+        # check if unmodified content changes with URL encoding
+        url_decoded_content = urllib.parse.parse_qsl(self.content,keep_blank_values=True)
+        url_reencoded_content = urllib.parse.urlencode(url_decoded_content, doseq=True).encode()
+        if self.content == url_reencoded_content:
+            # encoding the unmodified content doesn't change anything, so we should be good to go on as usual
+            self.content = mitmproxy.net.http.url.encode(form_data).encode()
+        else:
+            # if we are here, we have to alter encoding behavior to mimic the one used in unmodified POST request
+            
+            #check if content isn't encoded at all
+            if self.content == urllib.parse.unquote(url_reencoded_content.decode()).encode():
+                # if we land here, the new content mustn't be URL encoded thus we decode the result before setting new content
+                # this is the case for "https://www.facebook.com/ajax/updatestatus.php" but additionally the "=" sign is omitted on SOME empty keys
+                self.content = urllib.parse.unquote(mitmproxy.net.http.url.encode(form_data)).encode()
+            # check if content isn't encoded at all and equal signs are omitted
+            elif self.content == urllib.parse.unquote(url_reencoded_content.decode()).replace("=&","&").encode():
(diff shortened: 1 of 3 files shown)


mame82:
Tests fail due to insufficient code coverage in request.py. @mhils  Could you please add the test cases to reach the needed code coverage?

mitmproxy/net/http/request.py
```diff
@@ -367,13 +367,64 @@ def _get_urlencoded_form(self):
                 pass
         return ()
 
+    def _sanitized_form_data_replacement(self, form_data):
```
mhils: We should move the logic here into a function without side effects so that it's easier to test. I think it would make sense to add a second `similar_to: str=None` argument to `mitmproxy.net.http.url.encode` and then have the logic there.

mitmproxy/net/http/request.py
```diff
@@ -367,13 +367,64 @@ def _get_urlencoded_form(self):
...
+        # check if unmodified content changes with URL encoding
+        url_decoded_content = urllib.parse.parse_qsl(self.content, keep_blank_values=True)
```
mhils: We sometimes use `urllib.parse.urlencode`, sometimes`mitmproxy.net.http.url.encode`. This is too inconsistent.

mitmproxy/net/http/request.py
```diff
@@ -367,13 +367,64 @@ def _get_urlencoded_form(self):
...
+        url_reencoded_content = urllib.parse.urlencode(url_decoded_content, doseq=True).encode()
+        if self.content == url_reencoded_content:
+            # encoding the unmodified content doesn't change anything, so we should be good to go on as usual
+            self.content = mitmproxy.net.http.url.encode(form_data).encode()
+        else:
+            # if we are here, we have to alter encoding behavior to mimic the one used in unmodified POST request
+
+            # check if content isn't encoded at all
```
mhils: This whole thing is called `urlencoded`, so the content being not encoded would be very weird. Have you seen this in practice?

mitmproxy/net/http/request.py
```diff
@@ -367,13 +367,64 @@ def _get_urlencoded_form(self):
...
+            # here's a real, but trimmed down data section of a POST request to "https://www.facebook.com/ajax/updatestatus.php"
+            # with content-type 'application/x-www-form-urlencoded'
+            #
+            # self.content:
+            # b'attachment&backdated_date[year]&composer_source_surface=newsfeed&multilingual_specified_lang=&__pc=PHASED%3ADEFAULT'
+            #    * the first two keys have no value, but no equal sign appended
+            #    * the second key shows that no URL encoding ist used ("[" and "]")
+            #    * the third key shows a key=value pair
+            #    * the next key "multilingual_specified_lang" again has no value, but this time an equal sign is appended
+            #    * the last key has assigned a value, which is URL encoded ("%3A" == ":")
```
mhils: I think we have to live with the fact that we can't provide exact roundtrips in some cases. That's fine, one can still use the raw edit then. This example should go into a test.

mitmproxy/net/http/request.py
```diff
@@ -367,13 +367,64 @@ def _get_urlencoded_form(self):
...
+            # if we are here, we have to alter encoding behavior to mimic the one used in unmodified POST request
+
+            # check if content isn't encoded at all
+            if self.content == urllib.parse.unquote(url_reencoded_content.decode()).encode():
+                # if we land here, the new content mustn't be URL encoded thus we decode the result before setting new content
+                # this is the case for "https://www.facebook.com/ajax/updatestatus.php" but additionally the "=" sign is
+                # omitted on SOME empty keys
+                self.content = urllib.parse.unquote(mitmproxy.net.http.url.encode(form_data)).encode()
+            # check if content isn't encoded at all and equal signs are omitted
+            elif self.content == urllib.parse.unquote(url_reencoded_content.decode()).replace("=&", "&").encode():
```
mhils: This is a very confusing encode-decode dance that does not make too much sense to me.

mitmproxy/net/http/request.py
```diff
@@ -367,13 +367,64 @@ def _get_urlencoded_form(self):
...
+            #    * the first two keys have no value, but no equal sign appended
+            #    * the second key shows that no URL encoding ist used ("[" and "]")
+            #    * the third key shows a key=value pair
+            #    * the next key "multilingual_specified_lang" again has no value, but this time an equal sign is appended
+            #    * the last key has assigned a value, which is URL encoded ("%3A" == ":")
+            #
+            # To cope with that, a somehow hacky approach is used
+            #    If unmodified and unquoted POST data contains ONE SINGLE key without a value, which hasn't an equal sign appended
+            #    we remove all equal signs on empty keys in modified data
+            elif any("=" not in param for param in urllib.parse.unquote(self.content.decode()).split("&")):
```
mhils: Why is this using unquote and not urldecode/parse_qsl? That seems wrong.

mitmproxy/net/http/request.py
```diff
@@ -367,13 +367,64 @@ def _get_urlencoded_form(self):
                 pass
         return ()
 
+    def _sanitized_form_data_replacement(self, form_data):
+        # check if unmodified content changes with URL encoding
+        url_decoded_content = urllib.parse.parse_qsl(self.content, keep_blank_values=True)
+        url_reencoded_content = urllib.parse.urlencode(url_decoded_content, doseq=True).encode()
+        if self.content == url_reencoded_content:
```
mhils: Why do we need to special-case this? None of our heuristics should trigger if it's "normally" urlencoded.

mitmproxy/net/http/request.py
mame82: All corner cases not fulfilling the condition `urlencode(urldecode(self.content) == self.content` should enter this `else` tree after early out. So all special cases are handled here. Examples for not well formed self.content:

**1. Keys without value, but no equal sign appended**
```
>>> test_content = b'key_without_value_and_missing_equal_sign&key_without_value_and_missing_equal_sign2'
>>> reencoded_content=urllib.parse.urlencode(urllib.parse.parse_qsl(test_content, keep_blank_values=True), doseq=True).encode()
>>> reencoded_content
b'key_without_value_and_missing_equal_sign=&key_without_value_and_missing_equal_sign2='
>>> reencoded_content == test_content
False
```

To mimic this behavior, equal signs of key without values have to be removed from the result of `mitmproxy.net.http.url.encode(form_data)`

**2. Key, value or both aren't encoded at all in the client request**
```
>>> test_content = b'date[year]=2017'
>>> reencoded_content=urllib.parse.urlencode(urllib.parse.parse_qsl(test_content, keep_blank_values=True), doseq=True).encode()
>>> reencoded_content
b'date%5Byear%5D=2017'
>>> reencoded_content == test_content
False
```

To mimic this behavior, result of `mitmproxy.net.http.url.encode(form_data)` has to be unquoted again.
As there's no urldecode() and parse_qs() / parse_qsl() don't work on `bytes` urllib.parse.unquote has to be used. This should be perfectly fine, as `urllib.parse.urlencode()` uses `quote()` and `quote_plus()` internally, according to [the docs](https://docs.python.org/3/library/urllib.parse.html).

**Other cases which would hit this code:**

```
# space is encoded with quote() instead of quote_plus() by the client sending the request
self.content = b'full%20name=max%20mustermann'

# mixed usage of equal signs on empty keys
self.content = b'key_with_equal_sign=&key_without_equal_sign'
```

and so on...

I think it is clear where this is getting. Most of these examples occur in the POST requests to facebook endpoints which have already been discussed.  Another example has been given [here](https://github.com/mitmproxy/mitmproxy/issues/1946#issuecomment-277522034) by @Kriechi because curl is used with a content-type header 'application/x-www-form-urlencoded' but data is provided without encoding (`--data-urlencode` should be used instead of `--data`). So yes this is somehow wrong tool usage, but if we call it a special case, that would be exactly what we're talking about. Handling some of these cases would make mitmproxy more robust in the end.

What should become clear at this point: There's no way to handle all the cases, this is why I designed the function to act like this:

```
if urlencode(urldecode(self.content) == self.content:
   # go on as usual
else:
   # handle special cases
   if (special case1):
      # modify result of mitmproxy.net.http.url.encode(form_data) accordingly
   elif (special case2):
      # modify result of mitmproxy.net.http.url.encode(form_data) accordingly
   elif (special case3):
      # modify result of mitmproxy.net.http.url.encode(form_data) accordingly
   else:
      # we have some sort of special case which we haven't considered
      # so this is the FAIL OVER for an unhadled case
      # in current implementation fail over is the default behavior of setting:
      # self.content = mitmproxy.net.http.url.encode(form_data).encode()
```

mitmproxy/net/http/request.py
mame82: `if self.content == urllib.parse.unquote(url_reencoded_content.decode()).encode()`

If this statement is reached, we already know that self.content isn't encoded "well formed".
Whereby `url_reencoded_content.decode()` has the same result as a call to `mitmproxy.net.http.url.encode(form_data).encode()` if form_data hasn't been modified.

By unquoting `url_reencoded_content.decode()` and comparing it to self.content, we basically check if self.content isn't urlencoded at all. The need to use unquote() has already been explained (no urldecode available, parse_qs not applicable).

So this first special case check could be rewritten to:
`if self.content == urldecode(urlencode(urldecode(self.content))):`
Which resolves to:
`if self.content == urldecode(self.content):`

Now the only thing we have to do in order to handle this case is - instead of setting self.content to:
`self.content = mitmproxy.net.http.url.encode(form_data).encode()`
set it to:
`self.content = urldecode(mitmproxy.net.http.url.encode(form_data)).encode()`
as already described, unquote() is used to mimic urldecode(). So we end up with what is seen in line 385 to handle "special case 1".

mitmproxy/net/http/request.py
```diff
@@ -367,13 +367,64 @@ def _get_urlencoded_form(self):
...
+            # check if content isn't encoded at all
+            if self.content == urllib.parse.unquote(url_reencoded_content.decode()).encode():
+                # if we land here, the new content mustn't be URL encoded thus we decode the result before setting new content
+                # this is the case for "https://www.facebook.com/ajax/updatestatus.php" but additionally the "=" sign is
+                # omitted on SOME empty keys
+                self.content = urllib.parse.unquote(mitmproxy.net.http.url.encode(form_data)).encode()
+            # check if content isn't encoded at all and equal signs are omitted
+            elif self.content == urllib.parse.unquote(url_reencoded_content.decode()).replace("=&", "&").encode():
+                # remove equal sign for keys without value
+                self.content = urllib.parse.unquote(mitmproxy.net.http.url.encode(form_data)).replace("=&", "&").encode()
```
mame82: Line 387 is the check for "special case 2" is similar to "special case 1". Beside urldecoding (same principle) our test data again, all occurrences of "=&" are replaced with "&".

So this special case check could be rewritten to:
`if self.content == urldecode(urlencode(urldecode(self.content))).replace("=&", "&"):`
Which resolves to:
`if self.content == urldecode(self.content).replace("=&", "&"):`

In words: if umodified content isn't encoded and empty keys don't have an equal sign appended.

Line 389 fixes the result of `mitmproxy.net.http.url.encode(form_data)`

This code ignores another corner case:
For `self.content = b'key1&key2` without further data modification `mitmproxy.net.http.url.encode(form_data).encode()` would return  `b'key1=&key2=' `
As the check above only touches "=&" the equal sign tailing key2 wouldn't be replaced. Thus the condition check wouldn't hit, because the `elif` statments ends up comparing:
`b'key1=&key2=' == b'key1&key2='`

So this is a more complex case, which I haven't checked for, but could easily be added in with a new `elif` following the principles of the other special cases.

mitmproxy/net/http/request.py
mame82: There should be no need to use raw mode, because another special case could be added in with about 2 lines of code ("elif" condition check + correct value assignment to self.content).
Other cases could be identified by setting a break point to the code of the last else statement, because unhandled special cases end up there.

mitmproxy/net/http/request.py
mame82: This has already been explained: urldecode doesn't exist, as far as I know. Parse_qsl used like this `urllib.parse.parse_qsl(self.content.decode())` would remove every equal sign by nature and thus we would have nothing to check for.

mitmproxy/net/http/request.py
```diff
@@ -367,13 +367,64 @@ def _get_urlencoded_form(self):
...
+
+                # unquote
+                modified_content = mitmproxy.net.http.url.encode(form_data)
+                # remove equal sign
+                modified_content = modified_content.replace("=&", "&").encode()
+                self.content = modified_content
+
+            # additional corner cases could be added to with new "elif" statements
+            else:
+                # if we are here we trigger default behavior again
```
mame82: Exactly here is the FAIL OVER for unconsidered special cases

mame82:
@mhils 
Reviewing the review looks somehow weird now, but anyway. I've commented on your comments except for one thing: The testing.

So my initial issue on facebook has been solved. Although it would be nice to have the code changes reflected in the master, this task became very time consuming. As I'm not able to work on this in my spare time, I have convinced my employer to let me work on this for 2 additional (argumentation: we don't have to do time consuming manual patching on future mitmproxy releases) .

So as nearly one day has passed already, I ask you to comment back on my review and give me a hint on: where to find sources from which I can learn how to implement the needed tests to reach code coverage in less than a day.

Additionally I would like to ask if I'm allowed to contact you directly, to discuss the facts needed to achieve a merge in the next two days.

Btw. I'm not able to run "tox" locally but hadn't have time to dive into the underlying errors.

Regards and Thanks for your great effort

MaMe82

mitmproxy/net/http/request.py
mhils: > If this is what you mean, I fully agree

Yes. The motivation is not necessarily to use this elsewhere, the motivation is to keep encode side-effect free to simplify testing.

mitmproxy/net/http/request.py
mhils: Yes, I understand what his check is for. My point is that none of the of the special cases should trigger, so we would fall through to what you call "FAIL OVER" anyway.

Kriechi:
This seems to be a very complex issue, maybe we should take a step back and perform real TDD.
What is the input we are trying to encode/decode?
What is the expected output for each scenario?
Can we start writing tests first, so we can agree on all pre- and post-conditions?

@mame82 contributing to an open source project means to join a community. Simply dropping a hugely complex patch and asking somebody else to write tests for it is not a very nice way, and it's not helping your cause of getting this issue fixed.

I think if we come up with solid specification of want we want to achieve, we can discuss the necessary changes to mitmproxy - without conflicting and confusing back and forth in this discussion thread.

mitmproxy/net/http/request.py
mhils: the problem with unquote is that it also unquotes equal signs. My initial comment about using "urldecode/parse_qsl" is moot as well (thanks for pointing that out). I think we should not unquote at all here.

mhils:
> Additionally I would like to ask if I'm allowed to contact you directly, to discuss the facts needed to achieve a merge in the next two days.

You are welcome to join our developer chat on Slack.

Other than that, @Kriechi's comment is spot on. We should start by writing test cases [here](https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/net/http/test_url.py#L88-L96) that specify what we actually want to do. Once that is done we can test our implementation with `pytest test_url.py`. I think a lot of your comments above can directly transferred into test cases. For example, here is one test case we can add:

```python
def test_trailing_equal_sign():
    """
    Some web servers do not send a trailing equal sign for parameters, e.g. they send
        foo=bar&baz&qux=quux
    instead of
        foo=bar&baz=&qux=quux
    """
    assert url.encode(
        [('foo', 'bar'), ('baz', '')],
        similar_to="foo=bar&baz&qux=quux"
    ) == "foo=bar&baz"
    assert url.encode(
        [('foo', 'bar'), ('baz', '')],
        similar_to="foo=bar&baz=&qux=quux"
    ) == "foo=bar&baz="
```

mitmproxy/net/http/request.py
mame82: Agree

mitmproxy/net/http/request.py
mame82: There are 3 cases considered:

1. POST data isn't urlencoded at all 

Check: 
`
if self.content == urllib.parse.unquote(url_reencoded_content.decode()).encode():
`

triggering example:
`
self.content
b'date[year]=2017'
`

Tested with: 
`curl -k --proxy localhost:8080 'https://www.example.com/' -H 'Content-Type: application/x-www-form-urlencoded' --data 'date[year]=2017'`

2. POST data isn't urlencoded at all and contains empty key without equal sign

Check: 
`
elif self.content == urllib.parse.unquote(url_reencoded_content.decode()).replace("=&", "&").encode():
`

triggering example:
`
self.content
b'key_without_value&date[year]=2017'
`


Tested with: 
`curl -k --proxy localhost:8080 'https://www.example.com/' -H 'Content-Type: application/x-www-form-urlencoded' --data 'key_without_value&date[year]=2017'`

3. urldecoded POST data contains at least one empty key without equal sign: 

Check:
`
elif any("=" not in param for param in urllib.parse.unquote(self.content.decode()).split("&")):
`

triggers everytime a message is posted on facebook (https://www.facebook.com/ajax/updatestatus.php)

Tested by posting on facebook.


So all triggers work as intended on my side, but i agree that case 1 and case 2 are rare examples, which should be dropped in order to focus on the initial facebook issue.

mitmproxy/net/http/request.py
mame82: I get your point. This would misbehave if self.content contains something like:
`b'comment=word%20%3D%26other%20word'`

unquoting this would result in:
`'comment=word =&other word'`

Although this should be a very rare case, you're right as unquoting isn't needed at all and shouldn't be applied to make the logic work.

mame82:
@Kriechi 

> Simply dropping a hugely complex patch and asking somebody else to write tests for it is not a very nice way, and it's not helping your cause of getting this issue fixed.

I feel a bit misunderstood here, so let me explain:
The initial issue was a potential error in mitmproxy, editing form data which is directed towards facebook.
This is one of the key parts in a "Live Hacking" talk, in which we show how an attacker could not only sniff traffic, but manipulate it. As our target is to achieve "Cyber Awareness" we hide technical complexity from our audience. My task is to create and present these talks. As I personally use mitmproxy for a long time, I put much effort into making it the tool of choice for our talks. I like and trust the project. For exactly that reason I put time into further analysis of the issue, which for somebody who has never looked behind the scenes (inner workings of mitmproxy) isn't an easy task.
I've done the research to come up with a (free) remote debugging  setup to even be able to analyse the code and inner workings, of course without bothering you with that. The result was the insight that the cause of the problem wasn't mitmproxy, but a webservice ignoring existing standards. Along with this came the decision to file the whole issue as enhancement and not as a bug, which I absolutly agree with and again felt the need to contribute in pushing this forward. For exactly this reason I tried to support you with a working solution. Up to this point, I was involved into this "edge case enhancment" for nearly a month and presented a solution solving my initial problem already. For every day I worked on this, somebody else has to do the work I should normally do. This didn't stopped me from convincing my employer to allow me to work another two days on this to do it the way it should been done, because this possibly could lead to an enhancement of mitmproxy. As you surely noticed I'm no professional software developer, but an infosec guy - this means from my perspective Python is a tool to create quick scripts for networking tasks, exploit development or creating small helper tools. Using this for a big and well maintained project like mitmproxy is nothing I'm into normally. I'm impressed of the whole automated testing you applied to the project but I'm far away from understanding all the processes behind this as well as needed. To be precise: only to get an idea how to do a "code review" on github, I had to watch tutorials because I'm new into this. I outlined that I'm not able to work on this in my spare time, which doesn't mean I'm not willing to do so. The fact that I'm sitting at home now, writing this while my wife has to manage three childs + cooking should emphasize this. So my question for a

> hint on: where to find sources from which I can learn how to implement the needed tests to reach code coverage in less than a day.

to get the needed knowledge myself is a bit away from your interpretation of

> dropping a hugely complex patch and asking somebody else to write tests for it

mame82:
@mhils 
Thx for pointing out where to place test cases, your example seems pretty self explanatory. 
So I would suggest the following:
1) Removing everything from the heuristic which doesn't hit the facebook issue
2) moving this logic to mitmproxy.net.http.encode with an additional similar_to argument
3) create a test case to check for the following: If unmodified content contains keys without values without an appended equal sign, modified content shouldn't contain an equal sign on any key without value (neither existing ones, nor newly created ones)

But there seems to be a problem left, if I understand the purpose of self.content right, it is possible that there is no content present before entering "form edit mode" but could be added there. In this case we would miss reference data.

mhils:
Ok, let's do baby steps so that we get somewhere.

1. Take the test case I wrote above to test for the equal sign behaviour.
2. Move logic to `url.encode`. Here's a rough skeleton how we are going to structure `encode`:


```python
from typing import Optional

def encode(s: Sequence[Tuple[str, str]], similar_to: Optional[str]) -> str:
    """
    URL-Encode a sequence of (key, value) tuples.
    If similar_to is passed, the output is formatted similar to the provided urlencoded string.
    """
    remove_trailing_equal = False
    if similar_to:
        remove_trailing_equal = # TODO: check if we can detect a missing trailing equal sign.

    encoded = urllib.parse.urlencode(s, doseq=True, errors="surrogateescape")
    
    if remove_trailing_equal:
        # TODO: remove trailing equal signs here

    return encoded
```

3. Make sure the tests passes. You can test using `pytest`.
4. Adapt _set_urlencoded_form so that it passes `similar_to`.

We can think of implementing further changes afterwards.

mame82:
Sounds like a plan and I'm sure I get this part done tomorrow.
But anyway I want to remind on something I've already pointed out.

The following snippet of POST form data is a real life example directed to a facebook ajax endpoint:
`'attachment&backdated_date[year]&composer_source_surface=newsfeed&multilingual_specified_lang=&__pc=PHASED%3ADEFAULT'`

We have empty keys without equal sign (`attachment`) and an empty key with equal sign appended (`multilingual_specified_lang=`).

Although it is a baby step, to implement "if one empty key has no equal sign, none of the empty output keys should have an equal sign" could have an real life impact. If we'd rather decided to implement  "if one empty key has an equal sign, all of the empty output keys should have an equal sign" things would be different. Even if not obvious these two assumptions compete. Applying the above example as `similar_to` argument would result in all empty keys being decorated with no equal sign if the latter rule is used and no equal sign if the former rule is used.

The impact of this design decision could only be measured in the wild not in a test asserting the logic works.

At least, the chosen approach is known to work on facebook.

mame82:
@mhils  Think the first step is taken, thx for your great support.

mhils:
This looks much better now. Thanks!

mame82:
Learned a ton of things on this: github usage, pytest usage, proper code style, python remote debugging, culprits of urlencoding ...so thanks to you and my respect for dealing with these things in regular manner.

PR #2028, opened 2017-02-14
nuke Pillow

Kriechi:


mitmproxy/contentviews/image/view.py
```diff
@@ -37,19 +33,3 @@ def __call__(self, data, **metadata):
             parts = image_parser.parse_jpeg(data)
             fmt = base.format_dict(multidict.MultiDict(parts))
             return "%s image" % f, fmt
```
mhils: We still need a fallback here. I propose an almost empty view that just states the `image_type`.

Kriechi:
Done.

mhils:
What about https://github.com/mitmproxy/mitmproxy/pull/2028/files#r101171665?

mitmproxy/contentviews/image/view.py
Kriechi: Currently only JPEG, PNG, and GIF are supported and in the content-type filter. No other image types should generate this view.

Kriechi:
Is zlib because of Pillow, I thought that was in general use for compression stuff?

s4chin:
Here, as `imghdr` doesn't return all the image types, I was planning to use this imghdr - https://bugs.python.org/review/21574/diff/12977/Lib/imghdr.py
So `imghdr.what()` can just return the correct image type and that's what will be returned along with say `[("Status", "Unsupported image format")]` as `parts`.
Sorry, I have exams so I'm off for 10 days.

Kriechi:
We could also utilize python-magic.
https://github.com/ahupp/python-magic

Not sure though how @mhils feels about that dependency...

mhils:
> Is zlib because of Pillow, I thought that was in general use for compression stuff?

I think it's there just for Pillow. We should do a test install on a clean VM once this is merged. Easier that way when we have snapshots.

> We could also utilize python-magic.
https://github.com/ahupp/python-magic

A ctypes wrapper around libmagic after we have successfully eliminated Pillow. Not sure if trolling or not 😁 

I added a fallback view and made the image view catch all images by default. I think that's a better experience than showing the raw bytes right away.

Issue #2019, opened 2017-02-13
HTTPS not working with authentication


laurmurclar:
##### Steps to reproduce the problem:

1. `mitmproxy --nonanonymous`
2.  Set a proxy username and password in `System Preferences -> Network -> Wifi -> Advanced -> Proxies` for both HTTP and HTTPS.
![screen shot 2017-02-13 at 19 54 32](https://cloud.githubusercontent.com/assets/9434500/22900943/0b446b4c-f228-11e6-831f-f0bc6989b504.png)
3. Save changes. A dialog appears: 
![Uploading Screen Shot 2017-02-13 at 20.08.40.png…]()
Choose `Always Apply`

3. Open browser and navigate to a HTTPS page eg. `https://www.google.ie/`. This appears:
<img width="1296" alt="screen shot 2017-02-13 at 19 54 53" src="https://cloud.githubusercontent.com/assets/9434500/22900938/04ca0984-f228-11e6-8473-5c8f490b4787.png">

4. Fill in username and password to match those entered earlier. 

5. Get the following error (ERR_SPDY_PROTOCOL_ERROR):
![screen shot 2017-02-13 at 19 55 08](https://cloud.githubusercontent.com/assets/9434500/22901207/d5db43bc-f228-11e6-9384-641aa546d8ba.png)

##### Any other comments? What have you tried so far?

After googling the error, I tried flushing the socket pools at chrome://net-internals/#sockets. I immediately got the same error after trying to access a HTTPS site again.

Previously I had been getting two dialog boxes; one requesting credentials for localhost:8080 and then another requesting credentials for google.ie:443 (or whatever site) directly after the localhost one.

Not sure if it matters, but the proxy credentials in my keychain:
<img width="1393" alt="screen shot 2017-02-13 at 19 58 15" src="https://cloud.githubusercontent.com/assets/9434500/22901465/ba025db4-f229-11e6-8128-6de1305cc3b3.png">

##### System information

Mitmproxy version: 2.0.0 (1.0.1dev0129-0x919385d)
Python version: 3.6.0
Platform: Darwin-15.6.0-x86_64-i386-64bit
SSL version: OpenSSL 1.0.2k  26 Jan 2017
Mac version: 10.11.6 ('', '', '') x86_64

laurmurclar:
Actually, it doesn't seem to be all sites. I can navigate to an app I have on heroku, and then I am asked for credentials for that site (this site does not need credentials): 
[...]

Issue #1734, opened 2016-11-12
Ensure that addons can access auth credentials from clients


cortesi:
See #1732 by @martyn-w for the motivation and discussion.

mhils:
It should now be possible to access `flow.metadata["proxyauth"]` for this.

PR #2029, opened 2017-02-15
Improve ProxyAuth

### LINKED ISSUE #1722: Proxy-Authentication does not work with HTTPS sites

1. `mitmdump --nonanonymous`
2. `curl -x localhost:8080 -k -U foo:bar http://example.com` (works)
3. `curl -x localhost:8080 -k -U foo:bar https://example.com` (doesn't)

The problem is an invalid handling of `http_authenticated`.

mhils:
This PR fixes #1722 and #1734, and hopefully #2019 as well. 
@laurmurclar, can you check if that works for you?

### DIFF
--- a/mitmproxy/addons/proxyauth.py
+++ b/mitmproxy/addons/proxyauth.py
@@ -10,6 +10,7 @@ import mitmproxy.net.http
 from mitmproxy import connections  # noqa
 from mitmproxy import exceptions
 from mitmproxy import http
+from mitmproxy.net.http import status_codes
 
 REALM = "mitmproxy"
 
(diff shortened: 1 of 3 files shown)


laurmurclar:
The dialog requesting credentials for the proxy no longer shows up, but I'm still getting the ERR_SPDY_PROTOCOL_ERROR problem and having to give credentials for HTTPS sites that don't usually need any. 😕 Thanks for working on this so quickly!

laurmurclar:
It does 🎉 I was just being an idiot earlier, sorry 😬 thanks again for fixing this so quickly

mhils:
No problem, thanks for testing that so quickly!

mitmproxy/addons/proxyauth.py
```diff
@@ -1,103 +1,117 @@
...
+    return scheme, user, password
 
 
 class ProxyAuth:
     def __init__(self):
         self.nonanonymous = False
         self.htpasswd = None
         self.singleuser = None
+        self.mode = None
+        self.authenticated = weakref.WeakSet()  # type: Set[connections.ClientConnection]
```
Kriechi: Is this voodoo black magic really necessary?

mitmproxy/addons/proxyauth.py
```diff
@@ -1,103 +1,117 @@
...
+                "Proxy Authentication Required",
+                mitmproxy.net.http.Headers(Proxy_Authenticate='Basic realm="{}"'.format(REALM)),
             )
         else:
             return http.make_error_response(
-                407,
-                "Proxy Authentication Required",
-                headers,
+                401,
+                "Authentication Required",
```
Kriechi: Please use the already defined reasons: https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/net/http/status_codes.py#L82

mitmproxy/addons/proxyauth.py
```diff
@@ -1,103 +1,117 @@
...
+        except ValueError:
+            return None
 
         if self.nonanonymous:
-            pass
+            return username, password
         elif self.singleuser:
-            if [username, password] != self.singleuser:
-                return False
+            if [username, password] == self.singleuser:
```
Kriechi: Is this a [Yoda condition](https://en.wikipedia.org/wiki/Yoda_conditions)?

PR #1791, opened 2016-11-25
use :authority header as pretty_host data source

Kriechi:
fixes #1726

PR #2034, opened 2017-02-17
Avoid adding Host header in reverse mode for HTTP2 streams

yoavweiss:
In "Reverse Proxy" mode, the proxy was adding "Host" headers to all requests sent out, even for ones on HTTP/2 connections.

The spec says this is something senders should not do:
> Clients that generate HTTP/2 requests directly SHOULD use the ":authority" pseudo-header field instead of the Host header field.  An intermediary that converts an HTTP/2 request to HTTP/1.1 MUST create a Host header field if one is not present in a request by copying the value of the ":authority" pseudo-header field.

Apparently, various HTTP/2 servers were not taking it well, and responded to such requests with RST_STREAMs indicating PROTOCOL_ERROR.

This PR changes that behavior.

### DIFF
--- a/mitmproxy/proxy/protocol/http.py
+++ b/mitmproxy/proxy/protocol/http.py
@@ -290,7 +290,7 @@ class HttpLayer(base.Layer):
             request.first_line_format = "relative"
 
         # update host header in reverse proxy mode
-        if self.config.options.mode == "reverse":
+        if self.config.options.mode == "reverse" and not f.request.headers[":authority"]:
             f.request.headers["Host"] = self.config.upstream_server.address.host
 
         # Determine .scheme, .host and .port attributes for inline scripts. For


Lukasa:
Hey @yoavweiss!

So, my reading of RFC 7540 leads me to want to assume the following behaviours:

1. If translating H2 -> H2, mitmproxy should leave the headers exactly as the client generated them. If the client put :authority and Host, mitmproxy should leave those there.
2. If translating H1 -> H2, the `:authority` header *MUST* be omitted, per Section 8.1.2.3, with the `Host` header left in place. mitmproxy is not a client in this context, it's an intermediary, so if it is doing translation then it is bound by that requirement.
3. If translating H2 -> H1, obviously the `:authority` header needs to be translated into a `Host` header.

So I'm not sure about the decision here which avoids adding the `Host` header. I think the `Host` header should almost never be removed, per my reading of RFC 7540.

Kriechi:
Thanks @Lukasa for sharing your opinion here!

About 3., the RFC states:
> An intermediary that converts an HTTP/2 request to HTTP/1.1 MUST
> create a Host header field if one is not present in a request by
> copying the value of the ":authority" pseudo-header field.

One could argue now that, **"copying"** implies that the `:authority` header remains. Or do you think it should be translated, i.e., replaced by a `host` header?

I think we have a special case, in that we are trying to manipulate an h2-request and send it off to a different host (reverse proxy mode). The question here is now, is it valid to have `:authority` AND `host` headers, and should h2-servers accept that.

Apparently @yoavweiss found himself dealing with a server that simply fails and sends PROTOCOL_ERROR if a request with both headers is sent (even if they contain the same value).

mhils:
Thanks for chiming in @Lukasa! :)

In reverse proxy mode (and only in reverse proxy mode), we change the value of the host header to whatever upstream server the user has specified. This is indeed not standard-conformant, but we had a lot of users complaining that e.g. the following would not work:
```
$ mitmdump -R http://example.com/
$ curl localhost:8080
```
Without rewriting the host header, that'd fail with a 404 because we forwarded the original `Host: localhost` the client has sent. We should add a `--keep-host-header` flag to enforce standard-conformant behaviour, but since we added this auto-rewrite this topic basically did not come up anymore.

The issue at hand is that the current rewriting is hardcoded to the "Host" header, not taking HTTP/2 into account. This is more or less a general issue we still have, see https://github.com/mitmproxy/mitmproxy/pull/1791.
I believe the nicest way to solve that is adding a `request.host_header` property that routes to whatever is the right header. I'll open a separate issue for that in a second to discuss it a bit further.

Lukasa:
> One could argue now that, "copying" implies that the :authority header remains. Or do you think it should be translated, i.e., replaced by a host header?

The pseudo-header fields must be removed in a HTTP/1.1 request. That means that copying means "move its data to the Host header, and then delete the header".

> The question here is now, is it valid to have :authority AND host headers, and should h2-servers accept that.

Yes, it is. Specifically, RFC 7540 says that emitting only `:authority` is SHOULD, not MUST, and so conformant h2 servers should accept it so long as both fields are the same.

I agree with @mhils however: the caveat is that both fields must be the same. In fact, if you had hyper-h2 header validation on you'd find that sending mismatched `Host` and `:authority` headers is forbidden. :wink:

Kriechi:
Thanks! I'm not sure about the precise circumstances of the original issue, but from what @yoavweiss reported, `:authority` and `host` were both present, and contained the same value (i.e. matched). And yet the server still sent an error.

Any way, the best way forward is to only change those headers that are already present and not add any more in mitmproxy, unless we are translating between H1 and H2 - which we are not doing anyway at this time.

Lukasa:
So, as with all things: just because the server SHOULD tolerate this doesn't mean it will. ;)

mhils:
FYI, #2040 should also fix this. :smiley:

yoavweiss:
> FYI, #2040 should also fix this. 😃

Cool, closing this one.

> Thanks! I'm not sure about the precise circumstances of the original issue, but from what @yoavweiss reported, :authority and host were both present, and contained the same value (i.e. matched). And yet the server still sent an error.

Just to emphasize, I hit this issue with 3 different server implementations (2 of them returned a PROTOCOL_ERROR after the request, and one simply returned a 400 response. I failed to find an H2 sites served by a server that accepted these requests (even though there may be ones).

mhils:
Yes, very good to get this fixed. Thanks again for the initial PR! :smiley:

mnot:
Which servers were doing this? They're broken.

yoavweiss:
I found 3 servers that were doing that: nginx, Akamai's H2 server and Fastly's H2 server (probably H2O).

But, I think I misdiagnosed and the reason for the PROTOCOL_ERROR/400 wasn't the presence of `Host` but the fact that it contained uppercased letters, which is spec compliant behavior AFAICT.

```
Just as in HTTP/1.x, header field names are strings of ASCII
   characters that are compared in a case-insensitive fashion.  However,
   header field names MUST be converted to lowercase prior to their
   encoding in HTTP/2.  A request or response containing uppercase
   header field names MUST be treated as malformed (Section 8.1.2.6).
```

Lukasa:
Heh, yes, mitmproxy should probably be fixing that up a bit too. ;)

mnot:
Aha! That makes a lot more sense.

mhils:
> But, I think I misdiagnosed and the reason for the PROTOCOL_ERROR/400 wasn't the presence of Host but the fact that it contained uppercased letters, which is spec compliant behavior AFAICT.

Ugh - this was caused by the same HTTP/2 unaware host header rewriting, sorry. This specific case has been fixed by #2040 already, @Kriechi enabled the uppercase checks for all headers in #2055 now.

PR #2043, opened 2017-02-18
http2: disable priority forwarding

Kriechi:


Kriechi:
Too complex? It's two if statements...
The logging is set to `debug`, not very verbose IMO...

mitmproxy/options.py
```diff
@@ -65,15 +66,18 @@ def __init__(
...
         ciphers_server: Optional[str]=None,
         clientcerts: Optional[str] = None,
-        http2: bool = True,
         ignore_hosts: Sequence[str] = [],
         listen_host: str = "",
         listen_port: int = LISTEN_PORT,
         upstream_bind_address: str = "",
         mode: str = "regular",
         no_upstream_cert: bool = False,
-        rawtcp: bool = False,
```
mhils: I think this was sorted alphabetically before.

mitmproxy/tools/cmdline.py
```diff
@@ -137,16 +137,19 @@ def get_common_options(args):
         ciphers_client = args.ciphers_client,
         ciphers_server = args.ciphers_server,
         clientcerts = args.clientcerts,
-        http2 = args.http2,
```
mhils: alphabetically as well

mitmproxy/options.py
Kriechi: Nope - should I make it sorted?

mitmproxy/tools/cmdline.py
Kriechi: Nope - again...

PR #2050, opened 2017-02-20
nuke tcp.Address and add proper IPv6 support

### LINKED ISSUE #1994: Cannot bind mitmproxy to IPv6 address

Kriechi:
This PR helps us to get IPv6-ready.

fixes #1994
closes #1995

### DIFF
--- a/mitmproxy/flowfilter.py
+++ b/mitmproxy/flowfilter.py
@@ -348,7 +348,10 @@ class FSrc(_Rex):
     is_binary = False
 
     def __call__(self, f):
-        return f.client_conn.address and self.re.search(repr(f.client_conn.address))
+        if not f.client_conn or not f.client_conn.address:
+            return False
+        r = "{}:{}".format(f.client_conn.address[0], f.client_conn.address[1])
+        return f.client_conn.address and self.re.search(r)
 
 
 class FDst(_Rex):
@@ -357,7 +360,10 @@ class FDst(_Rex):
     is_binary = False
 
     def __call__(self, f):
-        return f.server_conn.address and self.re.search(repr(f.server_conn.address))
+        if not f.server_conn or not f.server_conn.address:
+            return False
+        r = "{}:{}".format(f.server_conn.address[0], f.server_conn.address[1])
+        return f.server_conn.address and self.re.search(r)
 
 
 class _Int(_Action):
--- a/mitmproxy/io_compat.py
+++ b/mitmproxy/io_compat.py
@@ -88,6 +88,11 @@ def convert_019_100(data):
 
 def convert_100_200(data):
     data["version"] = (2, 0, 0)
+    data["client_conn"]["address"] = data["client_conn"]["address"]["address"]
+    data["server_conn"]["address"] = data["server_conn"]["address"]["address"]
+    data["server_conn"]["source_address"] = data["server_conn"]["source_address"]["address"]
+    if data["server_conn"]["ip_address"]:
+        data["server_conn"]["ip_address"] = data["server_conn"]["ip_address"]["address"]
     return data
 
 
--- a/mitmproxy/net/tcp.py
+++ b/mitmproxy/net/tcp.py
@@ -30,8 +30,7 @@ socket_fileobject = socket.SocketIO
 
 # workaround for https://bugs.python.org/issue29515
 # Python 3.5 and 3.6 for Windows is missing a constant
-if not hasattr(socket, 'IPV6_V6ONLY'):
-    socket.IPV6_V6ONLY = 41
+IPPROTO_IPV6 = getattr(socket, "IPPROTO_IPV6", 41)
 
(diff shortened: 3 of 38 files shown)


Kriechi:
@mhils: only 1 failing test, that is related to binding to an IPv6 socket. At the moment I hard-coded it to use the v4 family. We need to introduce a nifty way of fixing this. The second example could be useful: https://docs.python.org/3/library/socket.html#example

@ujjwal96 ping FYI

mitmproxy/flowfilter.py
```diff
@@ -357,7 +360,10 @@ class FDst(_Rex):
     is_binary = False
 
     def __call__(self, f):
-        return f.server_conn.address and self.re.search(repr(f.server_conn.address))
```
mhils: see https://github.com/mitmproxy/mitmproxy/pull/2050/files#r102108753

mitmproxy/net/tcp.py
```diff
@@ -915,19 +821,18 @@ class TCPServer:
     request_queue_size = 20
 
     def __init__(self, address):
-        self.address = Address.wrap(address)
+        self.address = address
         self.__is_shut_down = threading.Event()
         self.__shutdown_request = False
-        self.socket = socket.socket(self.address.family, socket.SOCK_STREAM)
+        self.socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
mhils: As you mentioned, this is why IPv6 binding fails. We probably should implement the strategy proposed in https://github.com/mitmproxy/mitmproxy/pull/1995 here: Try IPv6 with V6ONLY=false, and fall back to IPv4 if that fails. We can do that in a separate PR if you want.

mitmproxy/io_compat.py
```diff
@@ -88,6 +88,8 @@ def convert_019_100(data):
 
 def convert_100_200(data):
     data["version"] = (2, 0, 0)
+    data["client_conn"]["address"] = data["client_conn"]["address"]["address"]
+    data["server_conn"]["address"] = data["server_conn"]["address"]["address"]
```
Kriechi: Done.

mitmproxy/flowfilter.py
Kriechi: Done.

mitmproxy/flowfilter.py
```diff
@@ -348,7 +348,10 @@ class FSrc(_Rex):
     is_binary = False
 
     def __call__(self, f):
-        return f.client_conn.address and self.re.search(repr(f.client_conn.address))
+        r = repr(f.client_conn.address)
```
Kriechi: Done.

mitmproxy/net/tcp.py
```diff
@@ -743,12 +653,12 @@ def convert_to_ssl(self, sni=None, alpn_protos=None, **sslctx_kwargs):
 
     def makesocket(self):
```
Kriechi: Does the cuckoo sandbox team use this API? I would like to limit the API changes if possible.
Or do you know how they use it?

mhils:
Copying here from Slack: We can't use `create_connection` as we need to do the `spoof_source_address` logic between `s = socket.socket()` and `s.connect()`. We need to so something like this:

1. Move `if self.spoof_source_address:` and `connection.bind(self.source_address)` into `_makesocket()`. 
2. Copy the implementation of `create_connection` and adapt it to use our _makesocket: https://hg.python.org/cpython/file/3.6/Lib/socket.py#l688

mitmproxy/net/tcp.py
```diff
@@ -915,19 +852,36 @@ class TCPServer:
...
+        self.address = address
         self.__is_shut_down = threading.Event()
         self.__shutdown_request = False
-        self.socket = socket.socket(self.address.family, socket.SOCK_STREAM)
-        self.socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-        self.socket.bind(self.address())
-        self.address = Address.wrap(self.socket.getsockname())
+
+        if self.address == 'localhost':
+            raise ValueError("Binding to 'localhost' is prohibited. Please use '::1' or '127.0.0.1' directly.")
```
mhils: socket.error here so that this function only raises this?

Kriechi:
Yes, I grep-ed for `*.host` and `*.port` in `console/`.

Kriechi:
Done.

PR #2053, opened 2017-02-22
Adds --keep-host-header option (#2039)

### LINKED ISSUE #2039: Add --keep-host-header

1. `mitmdump -R https://httpbin.org/`
2. `curl localhost:8080/headers`

This returns "Host": "httpbin.org" instead of "Host": "localhost", which was what the original curl request sent. We want to add a `--keep-host-header` to disable our automatic host header rewrite [here](https://github.com/mitmproxy/mitmproxy/blob/6a1e03ac6f5fbbfda70200d465cfb2f7cc01d253/mitmproxy/proxy/protocol/http.py#L292-L294). For a discussion of the rationale, see https://github.com/mitmproxy/mitmproxy/pull/2034#issuecomment-280770573.

Once implemented, we should change the dns_spoofing example to use that instead of doing the `requestheaders` dance. First, however, let's keep the PR small and only add the switch.

krsoninikhil:
As per the discussion in [#2034](https://github.com/mitmproxy/mitmproxy/pull/2034#issuecomment-280770573) (closed) and #2039 issue, this adds a `--keep-host-header` option to avoid auto header rewrite. 

fixes #2039

### DIFF
--- a/mitmproxy/tools/cmdline.py
+++ b/mitmproxy/tools/cmdline.py
@@ -112,6 +112,7 @@ def get_common_options(args):
         replacements=args.replacements,
         replacement_files=args.replacement_files,
         setheaders=args.setheaders,
+        keep_host_header=args.keep_host_header,
         server_replay=args.server_replay,
         scripts=args.scripts,
         stickycookie=stickycookie,
@@ -386,6 +387,11 @@ def proxy_options(parser):
         "--upstream-bind-address",
         action="store", type=str, dest="upstream_bind_address",
         help="Address to bind upstream requests to (defaults to none)"
+    )
+    group.add_argument(
+        "--keep-host-header",
+        action="store_true", dest="keep_host_header",
+        help="Keep the host header as proxy addres"
     )
 
 
--- a/mitmproxy/tools/cmdline.py
+++ b/mitmproxy/tools/cmdline.py
@@ -391,7 +391,7 @@ def proxy_options(parser):
     group.add_argument(
         "--keep-host-header",
         action="store_true", dest="keep_host_header",
-        help="Keep the host header as proxy addres"
+        help="Reverse Proxy: Keep the original host header instead of rewriting it to the reverse proxy target."
     )
 
 
--- a/test/mitmproxy/proxy/test_server.py
+++ b/test/mitmproxy/proxy/test_server.py
@@ -482,6 +482,26 @@ class TestHTTPSNoCommonName(tservers.HTTPProxyTest):
 class TestReverse(tservers.ReverseProxyTest, CommonMixin, TcpMixin):
     reverse = True
 
+    def test_host_header(self):
+        self.config.options.keep_host_header = True
+        p = self.pathoc()
+        with p.connect():
+            resp = p.request("get:/p/200:h'Host'='example.com'")
+        assert resp.status_code == 200
+
+        req = self.master.state.flows[0].request
+        assert req.host_header == "example.com"
+
+    def test_overridden_host_header(self):
+        self.config.options.keep_host_header = False  # default value
+        p = self.pathoc()
+        with p.connect():
+            resp = p.request("get:/p/200:h'Host'='example.com'")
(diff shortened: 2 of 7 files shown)


mitmproxy/tools/cmdline.py
```diff
@@ -387,6 +388,11 @@ def proxy_options(parser):
         action="store", type=str, dest="upstream_bind_address",
         help="Address to bind upstream requests to (defaults to none)"
     )
+    group.add_argument(
+        "--keep-host-header",
+        action="store_true", dest="keep_host_header",
+        help="Keep the host header as proxy addres"
```
Kriechi: We should be a bit more descriptive here, something along this line:
`Keep the original host header in reverse-mode. The default is to modify the host header to reflect the upstream server address during reverse proxying request.`
@mhils ?

mhils: Yes. Let's do this:
```
Reverse Proxy: Keep the original host header instead of rewriting it to the reverse proxy target.
```

krsoninikhil: This sounds more clear. Thanks, I'll update it.

krsoninikhil:
Updated the docs and help message.

mitmproxy/tools/cmdline.py
```diff
@@ -387,6 +388,14 @@ def proxy_options(parser):
         action="store", type=str, dest="upstream_bind_address",
         help="Address to bind upstream requests to (defaults to none)"
     )
+    group.add_argument(
+        "--keep-host-header",
+        action="store_true", dest="keep_host_header",
+        help="""
+            Reverse Proxy: Keep the original host header instead of rewriting it
+            to the reverse proxy target.
+        """
```
mhils: Please remove the whitespace here, that doesn't work well on the console.

Kriechi:
@mhils: do we have a test for this in general which we can extend for the new option?

mhils:
I think the best we have is https://github.com/mitmproxy/mitmproxy/blob/bb2fa6dc7d871d703c6759926521d8c16aae80f1/test/mitmproxy/proxy/test_server.py#L482, but yes, we need a dedicated test for this.

krsoninikhil:
So should I just add `keep_host_header = True` below this test or more tests will be needed?

mhils:
This needs a dedicated test.

krsoninikhil:
Added a test for it, please have a look.

test/mitmproxy/proxy/test_server.py
```diff
@@ -481,6 +481,14 @@ def test_http(self):
...
     reverse = True
 
+    def test_host_header(self):
+        self.config.options.keep_host_header = True
+        f = self.pathod("304")
+        assert f.status_code == 304
+
+        r = self.master.state.flows[0].request
+        assert r.host_header == "localhost"
+
```
mhils: Nice. Can we add a reverse test for `keep_host_header = False`? 😃

krsoninikhil: Sure.

mhils:
Thanks. As I see it, the overriden header equals the original one in both cases so we cannot really verify that things are working? We can use `.pathoc` to specify a host header like so (untested):

```python
p = self.pathoc()
with p.connect():
    resp = p.request("get:/p/200:h'Host'='foo'")
```

krsoninikhil:
As I checked, the overridden header was in IP form (`127.0.0.1`) while original was `localhost`, that's why the `assert not` case is passing in case of `keep_host_header = False`.  But if it's not convincing, I'll try changing host with `.pathoc`, that will be more cleaner.

krsoninikhil:
`.pathoc` made it more clear, thanks. Should I squash last five commits into one?

mhils:
Yes, please. LGTM other than that!

krsoninikhil:
Squashed the redundant commits. Please let me know if this requires any further changes.

mhils:
Thanks for the great PR!

PR #2063, opened 2017-02-25
Remove pytz in hardump

ujjwal96:
Removes pytz from har_dump script.
Also, this uses utcfromtimestamp 
https://github.com/mitmproxy/mitmproxy/blob/master/examples/complex/har_dump.py#L92
whereas this uses fromtimestamp
https://github.com/mitmproxy/mitmproxy/blob/master/examples/complex/har_dump.py#L201

both of these were being passed to the same function which then set the tzinfo to UTC. I think both should have been utcfromtimestamp. 
In the fix both the times give UTC date time

### DIFF
--- a/examples/complex/har_dump.py
+++ b/examples/complex/har_dump.py
@@ -10,7 +10,7 @@ import zlib
 import os
 
 from datetime import datetime
-import pytz
+from datetime import timezone
 
 import mitmproxy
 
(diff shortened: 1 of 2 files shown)


Kriechi:
Can you remove pytz here please?
https://github.com/mitmproxy/mitmproxy/blob/master/setup.py#L116

ujjwal96:
Done

PR #2064, opened 2017-02-25
connections: coverage++

Kriechi:


Kriechi:
Unfortunatley no full individual coverage yet. 
Maybe @mhils could point me into the right direction on how to easily test the TLS functions without building up a huge infrastructure around it first...

PR #2066, opened 2017-02-26
certs: coverage++

Kriechi:


mitmproxy/certs.py
```diff
@@ -477,5 +477,9 @@ def altnames(self):
                 except PyAsn1Error:
                     continue
                 for i in dec[0]:
-                    altnames.append(i[0].asOctets())
+                    if i[0] is None and isinstance(i[1], univ.OctetString) and not isinstance(i[1], char.IA5String):
+                        e = b'.'.join([str(e).encode() for e in i[1].asNumbers()])
+                    else:
+                        e = i[0].asOctets()
```
mhils: One reason why people switched from the CN to SANs is that SANs are typed. Type confusion between DNS names and IP addresses may be a rather esoteric exploit vector, but still I feel very strongly that we should not discard the type information here if we support multiple types. The proper way would be to make altnames entirely typed, but then again nobody has ever complained so far that we just support DNS names. 
Instead of adding complexity in these (security-)sensitive parts of the codebase I'd rather vote to keep that out for now until someone complains. If we start supporting multiple types, that really needs a lot more testing and I don't think that's worth it for now.

Kriechi:
@mhils: There was untested code - which actually didn't work. So I tried to implement the missing bits and make it work as intended.
If you feel more comfortable without that, I can remove the whole thing entirely?

Kriechi:
Done - please take another look.

PR #2081, opened 2017-03-01
web: Fix #1888

### LINKED ISSUE #1888: mitmweb: --host Not Working

mitmweb is not respecting --host. All hosts are ips, never domain names. "showhost" is displayed at bottom of mitmweb web page.

I had reported this long ago. The issue was switches weren't implemented. Now they are but --host isn't working anyway.

##### System information

Mitmproxy version: 1.0.0
Python version: 3.5.2
Platform: Windows-10-10.0.14393
Windows version: 10 10.0.14393  Multiprocessor Free

s4chin:


s4chin:
Yes, it I copied it from the Python function. I think we should pass pretty_host in the JSON. Am I to add this here? https://github.com/mitmproxy/mitmproxy/blob/8f13f357adb6aa0e952eb99200ef3fbd5cc73c36/mitmproxy/tools/web/app.py

mhils:
Yes, that's the right place. Thanks!

s4chin:
I think this is done

mhils:
Thanks. This does not fix #1888 as we now default to showing the pretty host, but the change we do here is the right one nonetheless.

PR #2085, opened 2017-03-02
[web] Add mouse support for FilterInput popover.

MatthewShao:
This PR makes the FilterInput box more user-friendly. Originally, the popover will appear when you click the FilterInput box, but you could not select the filter item by mouse, you have to type into the box. 
I make the item in the popover click-able, and the selected filter will appear in the input box.
img

### DIFF
--- a/web/src/css/header.less
+++ b/web/src/css/header.less
@@ -113,5 +113,12 @@ header {
     .popover-content {
         max-height: 500px;
         overflow-y: auto;
+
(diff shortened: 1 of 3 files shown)


mhils:
Thanks. Is there an easy way to retain the focus on the input field after selecting a filter?

mhils:
Looks great now, thanks! 😄

PR #2087, opened 2017-03-02
link to stackoverflow for chrome certificate management

TomTasche:
the old link doesnt lead anywhere. i replaced it with a link to an answer on stackoverflow which helped me import the mitmproxy-certificate to chrome

ps: i'm not sure if the new link also applies to chrome on other operating systems? i guess it's the same for all...?


mhils:
Thanks! There's also https://chromium.googlesource.com/chromium/src/+/master/docs/linux_cert_management.md, but I guess the post you linked is actually nicer for our users.

Issue #2083, opened 2017-03-01
Unable to scroll mitmweb details in Firefox


wdormann:
##### Steps to reproduce the problem:

1. Install Firefox ESR 45.7.0 on Linux
2. Capture with mitmweb
3. Find a message that requires scrolling on the right pane, and attempt to view all content. 

##### Any other comments? What have you tried so far?

This seems very similar to issue #615
Chromium scrolls fine.

##### System information

Mitmproxy version: 2.0.0 (release version) 
Python version: 3.5.2
Platform: Linux-3.10.0-514.6.1.el7.x86_64-x86_64-with-centos-7.3.1611-Core
SSL version: OpenSSL 1.0.1e-fips 11 Feb 2013
Linux distro: CentOS Linux 7.3.1611 Core

MatthewShao:
I tried this out and found that in `.flow-detail section`, if we change the css code like this:
```css
section {
    /* display: flex;
     flex-direction: column;*/
    overflow-y: scroll;
}
```
That works for me on Ubuntu with Firefox 51.0.1 and Chrome 52.0.2743.116, don't know if this works on other browsers.

mhils:
@MatthewShao: We only care about Chrome/Firefox for now, so if that fixes things, please submit a PR! 😃

pihug12:
It seems like this fix wasn't commit in the 2.0.x branch as well.

* commit e064439 : 2017-03-04
* release v2.0.1 : 2017-03-14
* release v2.0.2 : 2017-04-28

The bug is still present in this last release.

Bjornskjald:
Still present...

kira0204:
I will take a look at it @mhils tomorrow?

PR #2095, opened 2017-03-04
[web] fix scroll bug in flow detail. (#2083)

MatthewShao:
This fix the scroll bus when viewing flow details in Firefox. Works for me in both Firefox and Chrome.


cle1000:
works also for me in firefox and chrome, thanks a lot!

PR #2097, opened 2017-03-05
Added automatic port increment feature if the port is already in use.

madhusai3113:
mitmproxy by default uses **8080** port.Added a script to change the port number by incrementing it if it is not free.


cortesi:
Hi there. Thanks for contributing. I'd like to think about the utility of this feature for a bit - I'm unsure whether it's helpful in a situation where, say, test browsers or devices are usually configured to use a pre-defined known port. Usually if the expected port is occupied, there is some issue the user should address, and we should fail visibly. 

Do you have a specific use-case for this? If so, it would be helpful to outline it here so we can discuss.

madhusai3113:
@cortesi  Yes, you are absolutely right, but in case the port is already in use by some important process, we can have this running on another port, if required I can add additional functionality where the user gets a prompt that the port is in use and if he wants to check preceding ports for running mitmproxy.

Kriechi:
We already display a nice error message: 
`Error starting proxy server: OSError(48, 'Address already in use')`. 

IMHO this is enough for the user to investigate *why* mitmproxy failed to start. Simply bumping up the port numbers is not really a good idea, other things might run there, or expect different ports to be available. And I certainly don't want to reconfigure my fleet of test devices again and again.

PR #2100, opened 2017-03-05
Options revamp

cortesi:
This patch does the following: 

- Centralise options and documentation for options in one place. 
- Regularise names across the command line, options and config files.
- Add a --set option[=value] flag to generically set any option.
- Add an --options flag that dumps an annotated, syntactically correct config file with all options and their default values.
- Re-assess the flags exposed on our commands (especially in the light of --set), and reduce them to a minimum. 

Todo in future work:

- Fix the mitmproxy console options editor
- Allow addons to add their own options. We're ready for this now.


mitmproxy/optmanager.py
```diff
@@ -45,33 +92,37 @@ class OptManager(metaclass=_DefaultsMeta):
...
-        instance = super().__new__(cls)
-        instance.__dict__["_opts"] = {}
-        return instance
-
     def __init__(self):
+        self.__dict__["_options"] = {}
         self.__dict__["changed"] = blinker.Signal()
         self.__dict__["errored"] = blinker.Signal()
-        self.__dict__["_initialized"] = True
+        self.__dict__["_processed"] = {}
```
mhils: I only see this used for `body_size_limit`, which we could just parse at the few places where we use it. Do you have bigger plans for this?

mitmproxy/optmanager.py
```diff
@@ -1,37 +1,84 @@
...
+        self.value = unset
+        self.help = help
+        self.choices = choices
+
+    def __repr__(self):
+        return "{value} [{type}]".format(value=self.current(), type=self.typespec)
+
+    @property
+    def default(self):
+        return copy.deepcopy(self._default)
```
mhils: Would we ever have mutable defaults? Right now we only have sequences, for which we could use tuples. We don't have dicts yet, but we could use `types.MappingProxyType` as a "frozendict". This would even be a bit nicer as that you cannot accidentally call `.append()` without noticing that it's a no-op.

mitmproxy/optmanager.py
```diff
@@ -95,61 +146,51 @@ def _call(options, updated):
...
         """
             Restore defaults for all options.
         """
-        self.update(**self._defaults)
-
-    @classmethod
-    def default(klass, opt):
-        return copy.deepcopy(klass._defaults[opt])
+        for o in self._options.values():
+            o.reset()
```
mhils: Does not trigger an update anymore. Intentional?

mitmproxy/tools/main.py
```diff
@@ -34,134 +35,80 @@ def assert_utf8_env():
...
 
-def process_options(parser, options, args):
+def process_options(parser, opts, args):
     if args.version:
         print(debug.dump_system_info())
         sys.exit(0)
+    if args.options:
+        print(optmanager.dump(opts))
+        sys.exit(0)
+    if args.quiet:
```
mhils: Any reason why we special-case quiet? Seems like we could just define the command line arg as store_const 0.

mitmproxy/optmanager.py
cortesi: Let's think about this. Immutable datastructures are great, but are often not nice to use in Python specifically.

mitmproxy/optmanager.py
cortesi: I'd love for this to just go away. In any other case than core, we can hang processed data off an addon instance. In the case of body_size_limit, my feeling is that _processed is marginally  nicer than having the core of mitmproxy depend on human-friendly value parsing in multiple places. 

In the longer term we should a) move as much as we can into addons, and b) get rid of this particular instance by passing it up and down our chain as needed. Let's punt on this right at the moment.

mitmproxy/tools/main.py
cortesi: The handling of quiet here is unchanged from before the patch. I _think_ the idea here is that -q should take precedence over any -v flags, regardless of the order they were passed in. 

I think we need to clarify this anyway. I'd like verbosity to be a string that matches our log level names, which is clearer.

mitmproxy/optmanager.py
cortesi: Yes, it probably should. Fixed.

PR #2107, opened 2017-03-06
Fix WBXML parsing to work with Python3 and the MITMProxy 2.0 codebase

davidpshaw:
Fixes for the WBXML parser to work correctly with MITM Proxy 2.0, which uses python 3


Kriechi:
This is the same as #2106?

davidpshaw:
Yes, sorry, I don't know why this showed up twice.  You can reject this one.

PR #2112, opened 2017-03-07
[WIP] Fixes #2103, eventlog disappearing problem

### LINKED ISSUE #2103: Make the eventlog more sticky

1. Start mitmproxy (console UI).
2. Press E to show the eventlog.
3. Do something to replace the entire view. For example, press ? to show help.
4. Press Q to go back. The eventlog is gone.
5. But apparently its state is still remembered somewhere, because, to show the eventlog again, I have to press E *twice*.

##### Any other comments?

I think it would be better if the eventlog stayed on screen after returning from another view. If not, at least the double-E bug should be fixed.

##### System information

(this is mitmproxy master)

Python version: 3.5.2
Platform: Linux-4.4.0-64-generic-x86_64-with-Ubuntu-16.04-xenial

krsoninikhil:
Function `sig_replace_view_state` seemed to be deleting the current view itself while it should be deleting the one below it.

Fixes #2103


krsoninikhil:
Added the tests for the same.

krsoninikhil:
Please let me know if this needs any changes?

cortesi:
I think the real answer here is to change how we handle the event log. It's not really a view state that should be pushed onto the stack. We should move the event log to the window.Window object, and toggle it upon change of the event log option. This means that we'll immediately be able to display the event log on any screen, rather than just the flow list, which is desirable. If we do this, we might also want an option to control the height of the event log so the user has a bit of control over how to lay it out.

krsoninikhil:
I will get back on this in few days. Apologies for the delay.

krsoninikhil:
Hi @cortesi, I moved the eventlog as you suggested, still working on the height part. But now it is a just another window and doesn't look like something to toggle? Please let me know if I am getting it wrong.

cortesi:
This PR is now obsolete - I changed the way that the eventlog relates to the rest of the console app entirely, and we now have a full 2-pane layout that manages this. Thanks for your work here - there's a LOT more to do on the console side, so if you're interested in helping there, please get in touch!

Issue #1458, opened 2016-08-01
Add Replace from file to Replacement Editor


dufferzafar:
We have a replacement editor (press `o` then `R` in mitmproxy) that allows us to add replacement patterns for flows. 

There's also a `--replace-from-file` cmdline option that gets the replacement text from a local file.

But this functionality can't be used from the replacement editor itself. 

---

A user tried to enter the path to file in the Replacement text field: https://discourse.mitmproxy.org/t/emulate-charles-proxys-map-local/116

arjun23496:
Is anyone currently working on this issue ? If no, may I work on this ?

dufferzafar:
@arjun23496 I don't think anyone is working so feel free to take it up, but before you start I think it'll be nice if you discuss your ideas with the dev team over at Slack.

cortesi:
Fixed by #2120

PR #2120, opened 2017-03-08
Replace from file through replacement editor #1458

ujjwal96:
Replacement file can be specified from replacement editor. 
I think this kind of defeats the purpose of `--replace-from-file` though, because now a user can also do `--replace :~q:foo:file://<filepath>` 
Should this only be allowed in replacement editor and not from `--replace` ?


cortesi:
This is  a step in the right direction. Could you please:

- Rebase this on the current options work I've just done. It's all different now, but better. 
- Use @ for the "read from file" marker. This is what I've done in proxyauth and is a convention familiar from some other tools (e.g. curl). So, the invocation would be --replace /filter/foo/@~/path/to/file. 
- Remove the replace_from_file option entirely. 
- Ignore the grid editor for now. I'm going to rewrite the options editing stuff entirely in light of the new options structures. Leave that aspect for me, and just focus on the options in core for now.

ujjwal96:
Should I change the docs as well?

mhils:
That'd be great, yes. Don't worry if you don't feel too comfortable in written english, one of us will read over it. :)

mhils:
Looks great! Do we want to merge Replace and _ReplaceBase at this point?

cortesi:
Yes, let's ditch ReplaceBase, and then this looks ready to merge.

mhils:
Not sure why this is failing on Linux and OSX. @Kriechi, any hints?

Kriechi:
The coverage check blocks this - only 88% for the replace addon.

mhils:
Yes, but that's a symptom of failing tests before that: https://travis-ci.org/mitmproxy/mitmproxy/jobs/210326235#L2528

mitmproxy/addons/replace.py
```diff
@@ -72,6 +72,10 @@ def configure(self, options, updated):
                     raise exceptions.OptionsError(
                         "Invalid regular expression: %s - %s" % (rex, str(e))
                     )
+                if s.startswith("@") and not os.path.isfile(s[1:]):
```
ujjwal96: I think we should use os.path.expanduser here since we are doing that in the editor and also in the replace function.

PR #2128, opened 2017-03-09
disable h2c prior knowledge connections

Kriechi:


test/mitmproxy/addons/test_disable_h2c_upgrade.py
```diff
@@ -1,17 +0,0 @@
-from mitmproxy.addons import disable_h2c_upgrade
-from mitmproxy.test import tflow
-
-
-class TestTermLog:
```
mhils: 😄

PR #2129, opened 2017-03-09
Custom options for addons

cortesi:
This PR extends and refactors our events and options mechanisms to enable custom options for addons. 

Todo:

- [x] Set custom options with --set
- [x] Set custom options from config files
- [x] Show custom options in --options output

Changes:

- Add an options parameter to the start() event. This is to be used by addons
on startup to add custom options.
- Add a running() event that is called once the proxy is up and running.
- With the new paradigm we can't log during master __init__, so add a tiny
termstatus addon to print proxy status to terminal once we're running.

### DIFF
--- a/mitmproxy/master.py
+++ b/mitmproxy/master.py
@@ -76,12 +76,16 @@ class Master:
 
     def run(self):
         self.start()
+        running = False
         try:
             while not self.should_exit.is_set():
                 # Don't choose a very small timeout in Python 2:
                 # https://github.com/mitmproxy/mitmproxy/issues/443
                 # TODO: Lower the timeout value if we move to Python 3.
                 self.tick(0.1)
+                if not running:
+                    running = True
+                    self.addons.invoke_all_with_context("running")
         finally:
             self.shutdown()
 
--- a/mitmproxy/master.py
+++ b/mitmproxy/master.py
@@ -42,6 +42,7 @@ class Master:
         self.event_queue = queue.Queue()
         self.should_exit = threading.Event()
         self.server = server
+        self.first_tick = True
         channel = controller.Channel(self.event_queue, self.should_exit)
         server.set_channel(channel)
 
@@ -76,20 +77,19 @@ class Master:
 
     def run(self):
         self.start()
-        running = False
         try:
             while not self.should_exit.is_set():
                 # Don't choose a very small timeout in Python 2:
                 # https://github.com/mitmproxy/mitmproxy/issues/443
                 # TODO: Lower the timeout value if we move to Python 3.
                 self.tick(0.1)
-                if not running:
-                    running = True
-                    self.addons.invoke_all_with_context("running")
         finally:
             self.shutdown()
 
     def tick(self, timeout):
+        if self.first_tick:
+            self.first_tick = False
+            self.addons.invoke_all_with_context("running")
         with self.handlecontext():
             self.addons("tick")
         changed = False
--- a/mitmproxy/tools/main.py
+++ b/mitmproxy/tools/main.py
@@ -45,9 +45,6 @@ def process_options(parser, opts, args):
     if args.quiet:
         args.flow_detail = 0
 
-    for i in args.setoptions:
-        opts.set(i)
-
     adict = {}
     for n in dir(args):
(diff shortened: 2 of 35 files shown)


mhils:
I know this is WIP, but what I see here so far is super confusing to me. 😕 

> Add an options parameter to the start() event. This is to be used by addons on startup to add custom options.

If I understand things correctly an addon is not supposed to access options at this point, only to add custom options, right? Assuming that's the case, this is a really counterintuitive API. If I see a "start" event that gets passed an argument called "options", I'd expect to be able to process options at this point.

> Add a running() event that is called once the proxy is up and running.

Why is this needed?

> With the new paradigm we can't log during master init

Not sure I understand why this is the case.

Two more general questions what would help me a lot understanding this:

1. Which challenges in this PR led you to which design decisions? The semantics of start/configure/running seem to have changed quite a bit.
2. What's our intended lifecycle of an addon? Can addons be enabled/disabled while mitmproxy is running? How do we deal with unknown options? How do we deal with options for addons that are not enabled?

cortesi:
Max, I'm excited that you're excited by this patch. Let's discuss even though, as you say, it's premature. ;)

- You're right - the fact that we have a ```start(options)``` event where the user shouldn't read options is confusing. We can talk about how to phrase the API to make this less confusing, we have lots of options. The underlying structure is clear, though - addons need to create options *before* configure or any of the other lifecycle methods are called. If we don't use the start event for this, we'd have to create a near-duplicate event called at the same point. We've also just discussed a declarative options attribute in Slack -  I could be amenable to that, though I feel that should come after we've created a base Addon class for everything to derive from. This is not a structural change, though - it would still be tantamount to initialising in __start__, and it would be a bit less flexible in some circumstances. 

- The __running()__ event fires after the proxy is up and running, and has been on the todo list for a while. It's used in to addons in this patch, with more to follow. For instance, now we have running, flow reading can finally become an addon. It's the last vestige of the old way of doing things that has code duplicated through all our tools. 

- The lifecycle is still the same, with the addition of running. ```start``` -> ```configure``` -> ```running```, with further ```configure``` calls whenever options change. Addons can of course be added while mitmproxy is running (the script loader does this), at and in that case these events are still called in that same sequence. Unknown options are always an error. Addon enabling/disabling is handled by the addons themselves, and is not part of this work (it's a common requirement, though, and one we may or may not cater for).

mitmproxy/master.py
```diff
@@ -76,12 +76,16 @@ def start(self):
...
+        running = False
         try:
             while not self.should_exit.is_set():
                 # Don't choose a very small timeout in Python 2:
                 # https://github.com/mitmproxy/mitmproxy/issues/443
                 # TODO: Lower the timeout value if we move to Python 3.
                 self.tick(0.1)
+                if not running:
+                    running = True
+                    self.addons.invoke_all_with_context("running")
```
mhils: Could we just do `self.addons.invoke_all_with_context("running")` after `self.start()` ? Or is the first `tick` doing something special?

mitmproxy/tools/main.py
```diff
@@ -39,15 +39,10 @@ def process_options(parser, opts, args):
     if args.version:
         print(debug.dump_system_info())
         sys.exit(0)
-    if args.options:
-        print(optmanager.dump(opts))
-        sys.exit(0)
-    if args.quiet:
+    if args.quiet or args.options:
```
mhils: Add comment why `args.options` is here?

mitmproxy/master.py
cortesi: No, first tick is special - it comes after all addons have been configured and the proxy is up and running. This is the semantic we need if we're, for instance, reading flows from file.

mitmproxy/tools/main.py
cortesi: This is just to suppress all other output when we are dumping options (for instance, info on ignored options). The comment would say "# Suppress all other output when dumping options", which is what the code says. ;)

mhils: I expected that dumping options would be done before we'd log anything (just like e.g. `--version`). I can't think of a smart comment either though, so let's go with this.

PR #2133, opened 2017-03-11
[web] Fixes all existed tests

MatthewShao:
This PR include two modifications:
1. I found the `updatFlow` action was out of dated, we should test `stopEdit` instead, however, `stopEdit` has a flaw, so I fix it as well as its test.
2. Solve all test errors.
The javascript tests on travis should be passed now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,6 @@ git:
 
 matrix:
   fast_finish: true
-  allow_failures:
-    - language: node_js
-      node_js: "node"
   include:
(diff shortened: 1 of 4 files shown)


mhils:
Awesome!
Can we remove the node tests from allowed failures right away?

@Kriechi, I think you have a new friend. :smile:

MatthewShao:
Sure! I will add up the coverage step by step later on.

mhils:
🎉

PR #2135, opened 2017-03-12
mark tutils.tmpdir as deprecated

mhils:
All further code should use pytest's fixture. Raising a DeprecationWarning here makes this clear to anyone reading the code and in particular to anyone using PyCharm:

image


cortesi:
Let's not do this - if we really want to switch, we should just do it. I recognise a cunning scheme to irritate poor @resam enough to make him do this instead, @mhils. ;)

mhils:
#2136 it is.

PR #2137, opened 2017-03-13
[web] Add tests for ducks/utils/store.js

MatthewShao:
Try to add some tests for store.js. Since this is a new test for a file that we haven't covered before, if there is anything not good enough, please let me know.

### DIFF
--- a/web/src/js/__tests__/ducks/utils/storeSpec.js
+++ b/web/src/js/__tests__/ducks/utils/storeSpec.js
@@ -14,64 +14,73 @@ describe('store reducer', () => {
     })
 
     it('should handle add action', () => {
-        let state = reduceStore(undefined, {})
-        expect(state = reduceStore(state, storeActions.add({id: 1}))).toEqual({
-                byId: { [1]: {id: 1} },
-                listIndex: { [1]: 0 },
-                list: [ {id: 1} ],
-                view: [ {id: 1} ],
+        let a = {id: 1},
+            b = {id: 9},
+            state = reduceStore(undefined, {})
+        expect(state = reduceStore(state, storeActions.add(a))).toEqual({
+                byId: { 1: a },
+                listIndex: { 1: 0 },
+                list: [ a ],
+                view: [ a ],
                 viewIndex: { 1: 0 },
         })
 
-        expect(reduceStore(state, storeActions.add({id: 9}))).toEqual({
-            byId: { [1]: {id:1}, [9]: {id:9} },
-            listIndex: { [1]: 0, [9]: 1 },
-            list: [ {id: 1}, {id: 9} ],
-            view: [ {id: 1}, {id: 9} ],
+        expect(reduceStore(state, storeActions.add(b))).toEqual({
+            byId: { 1: a, 9: b },
+            listIndex: { 1: 0, 9: 1 },
+            list: [ a, b ],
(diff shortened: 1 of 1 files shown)


web/src/js/__tests__/ducks/utils/storeSpec.js
```diff
@@ -0,0 +1,77 @@
...
+            listIndex: {},
+            view: [],
+            viewIndex: {},
+        })
+    })
+
+    it('should handle add action', () => {
+        let state = reduceStore(undefined, {})
+        expect(state = reduceStore(state, storeActions.add({id: 1}))).toEqual({
+                byId: { [1]: {id: 1} },
```
mhils: Minor nitpick: I think we can just write `1` instead of `[1]`?

web/src/js/__tests__/ducks/utils/storeSpec.js
```diff
@@ -0,0 +1,77 @@
...
+                viewIndex: { 1: 0 },
+        })
+
+        expect(reduceStore(state, storeActions.add({id: 9}))).toEqual({
+            byId: { [1]: {id:1}, [9]: {id:9} },
+            listIndex: { [1]: 0, [9]: 1 },
+            list: [ {id: 1}, {id: 9} ],
+            view: [ {id: 1}, {id: 9} ],
+            viewIndex: { 1: 0, 9: 1 },
+        })
```
mhils: I think these tests would be more expressive if we'd assign the entries to variables, e.g. like this:
```javascript
let a = {id: 1},
    b = {id: 9}

expect(reduceStore(state, storeActions.add(b))).toEqual({
    byId: { 1: a, 9: b },
    listIndex: { 1: 0, 9: 1 },
    list: [ a, b ],
    view: [ a, b ],
    viewIndex: { 1: 0, 9: 1 },
})
```

web/src/js/__tests__/ducks/utils/storeSpec.js
```diff
@@ -0,0 +1,77 @@
...
+    })
+
+    it('should not add the item with duplicated id', () => {
+        let state = reduceStore(undefined, storeActions.add({id: 1}))
+        expect(reduceStore(state, storeActions.add({id: 1}))).toEqual(state)
+    })
+
+    it('should handle update action', () => {
+        let state = reduceStore(undefined, storeActions.add({id: 1, foo: "foo"}))
+        expect(reduceStore(state, storeActions.update({id:1, foo:"foo1"}))).toEqual({
```
mhils: I'm getting nitpicky here again: Make this "bar", that's easier to spot for humans. You can assign to a variable again like this:
```javascript
let a = {id : 1, foo: "foo"},
    updated = {...a, foo: "bar"}
```

web/src/js/__tests__/ducks/utils/storeSpec.js
MatthewShao: Yes, it should be like that.

web/src/js/__tests__/ducks/utils/storeSpec.js
MatthewShao: Appreciate for the advices, these will definitely make our code more readable. Now I get a better concept about how a good test case should be.

mhils:
Awesome, thanks!

PR #2138, opened 2017-03-13
[web] Fixes the flaw in #2133

MatthewShao:
It seems that I made something wrong in last PR(#2133):

The flows in the table can't update the size entry and the icon on the left when finish loading. It turned out that we should not change `data` to `flow` on line 63, but change the parameter name of the `stopEdit()` function.
Sorry for the trouble.


mhils:
No worries. One more reason why we need tests! 😄

PR #2150, opened 2017-03-14
Match ~d and ~u filters against pretty_host too

### LINKED ISSUE #2146: Can't filter by hostname when --host is used

1. mitmproxy -T --host
2. Capture something
3. Press f and filter by something like `~d domain.com` or `~u domain com`

Going to detail view to get the resolved IP address and using it for the filter works, but it's uncomfortable when you want to match things like "google.com" (lots of IP ranges, lots of unrelated domains, but many mentioning google)

##### System information

Mitmproxy version: 2.0.0 (release version)
Python version: 3.6.0
Platform: Linux-4.9.11-1-ARCH-x86_64-with-arch
Linux distro: arch

dequis:
Changed the ~u filter in the console UI to match the behavior of
mitmweb, which only matches against pretty_url, never against url.

-----

Fixes #2146 (thanks mhils for the hint on where to fix it)

### DIFF
--- a/mitmproxy/flowfilter.py
+++ b/mitmproxy/flowfilter.py
@@ -319,10 +319,14 @@ class FDomain(_Rex):
     code = "d"
     help = "Domain"
     flags = re.IGNORECASE
+    is_binary = False
 
     @only(http.HTTPFlow)
     def __call__(self, f):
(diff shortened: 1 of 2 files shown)


mhils:
Thanks!

PR #2151, opened 2017-03-14
Update .travis.yml to fix the failure about yarn

MatthewShao:
Using `npm install -g yarn` is easy to get into trouble, so we change to install yarn with the script given by their developer. [reference](https://yarnpkg.com/lang/en/docs/install-ci/#travis-tab)


mhils:
Thanks!

PR #2152, opened 2017-03-14
Extract flow reading into addons

cortesi:
This patch moves the final pieces of master functionality into addons.

- Add a ReadFile addon to read from file
- Add a separate ReadStdin addon to read from stdin, only used by mitmdump
- Remove all methods that know about io and serialization from master.Master

There are a few more things to do before this is complete: we need an event mechanism so applications can trigger file reading on the hoof, and we need to clear up the fact that keepserving has a different default value depending on which tool is running.

### DIFF
--- a/mitmproxy/addonmanager.py
+++ b/mitmproxy/addonmanager.py
@@ -1,5 +1,6 @@
 from mitmproxy import exceptions
 from mitmproxy import eventsequence
+from . import ctx
 import pprint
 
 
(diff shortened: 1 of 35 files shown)


mhils:
Looks great. Side note: We currently also use `keepserving` in the context of server replay.

cortesi:
Yeah, also in client replay. I'm working on a consistent way to approach this.

PR #2153, opened 2017-03-15
[web] Reach 100% coverage for ducks/ui/flow.js

MatthewShao:
default
Now `ducks/ui/flow.js` has 100% coverage tests.


mhils:
Awesome, thanks! 😄

PR #2155, opened 2017-03-15
bump cryptography

### LINKED ISSUE #2154: cryptography dependency update?
The template here wasn't relevant so I cut it, apologies.

`mitmproxy` currently mandates `cryptography <1.8` which when built from source on macOS/OS X El Capitan (10.11) using Xcode 8 results in `Symbol not found: _getentropy` errors. Reported and explained in more detail [here](https://github.com/pyca/cryptography/issues/3332), hence my short summary. These issues were fixed in https://github.com/pyca/cryptography/pull/3354, which was released in the 1.8 branch.

This is also currently causing some "fun" for downstream package managers having to carry their own somewhat hacky patches like [this](https://github.com/Homebrew/homebrew-core/blob/37abcfc55099f635d0e187657a55a1eed36b5ccf/Formula/mitmproxy.rb#L193-L202). The gist of my request is whether `mitmproxy` has plans to/would consider update/updating [cryptography](https://github.com/mitmproxy/mitmproxy/blob/e723a58af5dc4fc7a46958aa9ce8c386a7387450/setup.py#L67) to use the latest release without these issues.

Kriechi:
fixes #2154

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,7 @@ setup(
         "click>=6.2, <7",
         "certifi>=2015.11.20.1",  # no semver here - this should always be on the last release!
         "construct>=2.8, <2.9",
-        "cryptography>=1.3, <1.8",
+        "cryptography>=1.3, <1.9",
         "cssutils>=1.0.1, <1.1",
         "h2>=2.5.1, <3",
         "html2text>=2016.1.8, <=2016.9.19",


Kriechi:
Looks like @requires needed some manual convincing 🔨 ... #2157

PR #2169, opened 2017-03-16
mypy without skipping imports 

cortesi:
This is based on PR #2168. 

We get little benefit at the moment from mypy, because we skip imports when we run it. This is an experiment to see what's needed to make mypy run with imports and without errors on mimtmproxy/master.py. See commits for more information. 

From here, validation of the complete project is not that far away.


mhils:
The GitHub diff is borked, trying to fix this now.

mitmproxy/net/http/encoding.py
```diff
@@ -10,18 +10,22 @@
...
+CachedDecode = collections.namedtuple(
+    "CachedDecode", "encoded encoding errors decoded"
+)
 _cache = CachedDecode(None, None, None, None)
 
 
-def decode(encoded: Union[str, bytes], encoding: str, errors: str='strict') -> Union[str, bytes]:
+def decode(
+    encoded: Optional[bytes], encoding: str, errors: str='strict'
+) -> Optional[str]:
```
mhils: This should be `-> Optional[AnyStr]`. For example `gzip` decodes from gzipped to ungzipped bytes. See https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/net/http/test_encoding.py

mitmproxy/net/http/encoding.py
```diff
@@ -62,7 +66,7 @@ def decode(encoded: Union[str, bytes], encoding: str, errors: str='strict') -> U
         ))
 
 
-def encode(decoded: Union[str, bytes], encoding: str, errors: str='strict') -> Union[str, bytes]:
+def encode(decoded: Optional[str], encoding: str, errors: str='strict') -> Optional[bytes]:
```
mhils: Same here. `decoded: Optional[AnyStr]` and `-> Optional[AnyStr]`

mitmproxy/net/http/request.py
```diff
@@ -82,8 +82,8 @@ def make(
             cls,
             method: str,
             url: str,
-            content: AnyStr = b"",
-            headers: Union[Dict[AnyStr, AnyStr], Iterable[Tuple[bytes, bytes]]] = ()
+            content: bytes = b"",
```
mhils: `content` can be `str` here, see https://github.com/mitmproxy/mitmproxy/pull/2169/files#diff-477e6d69234066ab1a95fcfa034f295cR119.

mitmproxy/net/http/response.py
```diff
@@ -69,8 +69,8 @@ def __repr__(self):
     def make(
             cls,
             status_code: int=200,
-            content: AnyStr=b"",
-            headers: Union[Dict[AnyStr, AnyStr], Iterable[Tuple[bytes, bytes]]]=()
+            content: bytes=b"",
```
mhils: Same here.

mitmproxy/utils/typecheck.py
```diff
@@ -1,7 +1,7 @@
 import typing
 
 
-def check_type(name: str, value: typing.Any, typeinfo: type) -> None:
+def check_type(name: str, value: typing.Any, typeinfo: typing.Any) -> None:
```
mhils: This should be `type` shouldn't it?

cortesi: This makes underscore attribute access on the type object in the method barf. We can tackle this a different way by peppering the function body with ignore annotations.

cortesi:
Ok, the annotations have been relaxed where you recommended. I wonder if the strategy here might not be to err on the side of strictness until the mypy checker tells us otherwise (as we expand the checks over the rest of the project).

PR #2170, opened 2017-03-16
Fix test_format_xml with dot in path

felixonmars:
When the path contains dot ".", replacing all dots will generate a non-exist result and raises a FileNotFoundError. Replacing only the last dot fixes this.


mhils:
Uh, thanks!

PR #2182, opened 2017-03-18
Fix #2134

### LINKED ISSUE #2134: mitmweb event log doesn’t show messages at level higher than info

1. Create the following `script1.py`:

```
import mitmproxy.ctx

def response(flow):
    mitmproxy.ctx.log.info('this is an info')
    mitmproxy.ctx.log.warn('this is a warning')
    mitmproxy.ctx.log.error('this is an error')
```

2. Start mitmweb with this script: `mitmweb -s script1.py`
3. Enable *Options* → *Display Event Log* to show the event log pane.
4. Run a flow through mitmweb, for example: `curl -vx http://localhost:8080 http://httpbin.org/get/`
5. Observe that three messages are printed to the system console (“this is an info”, “this is a warning”, “this is an error”), but only one of them is displayed in the event log pane in mitmweb (“this is an info”).

Checking the “debug” checkbox at the top right-hand corner of the event log pane has no effect on this.

##### System information

(this is mitmproxy master)

Python version: 3.5.2
Platform: Linux-4.4.0-66-generic-x86_64-with-Ubuntu-16.04-xenial

syahn:
I picked the shape of icon and added toggle buttons for error and warn.

### DIFF
--- a/web/src/js/components/EventLog.jsx
+++ b/web/src/js/components/EventLog.jsx
@@ -55,7 +55,7 @@ class EventLog extends Component {
                 <div onMouseDown={this.onDragStart}>
                     Eventlog
                     <div className="pull-right">
-                        {['debug', 'info', 'web'].map(type => (
+                        {['debug', 'info', 'web', 'warn', 'error'].map(type => (
                             <ToggleButton key={type} text={type} checked={filters[type]} onToggle={() => toggleFilter(type)}/>
                         ))}
                         <i onClick={close} className="fa fa-close"></i>
--- a/web/src/js/components/EventLog/EventList.jsx
+++ b/web/src/js/components/EventLog/EventList.jsx
@@ -83,7 +83,12 @@ class EventLogList extends Component {
 }
 
 function LogIcon({ event }) {
-    const icon = { web: 'html5', debug: 'bug' }[event.level] || 'info'
+    const icon = {
+      web: 'html5',
+      debug: 'bug',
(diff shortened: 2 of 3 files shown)


MatthewShao:
Looks great! I also got some consideration about this issue: 
1. It seems that 5 filter levels are a little bit clumsy for our users,  `error` and `warn` levels should be of top priority to display, is it a good idea to hide that toggle button for these two options and display the logs anyway?
2. It will be awesome if we could add some color to the logs, just like what we did in the console interface.

@mhils, thoughts?

syahn:
@MatthewShao Yes, I agree with you. It would be better!

MatthewShao:
@syahn Don't push it too quick, let's hear what @mhils say. :stuck_out_tongue:

mhils:
Thanks for the PR @syahn and thanks for the feedback @MatthewShao! I'll merge this right away because it's clearly better than what we do right now. For everything else, please feel free to just open a new PR.

> It seems that 5 filter levels are a little bit clumsy for our users, error and warn levels should be of top priority to display, is it a good idea to hide that toggle button for these two options and display the logs anyway?

I'm not sure what's the best UI here. A drop-down may be appropriate now, we could also probably get away with saying web = info.

> It will be awesome if we could add some color to the logs, just like what we did in the console interface.

I'm very much a fan of this - either just the icon or the whole text, whatever you think looks best. :)

Issue #2109, opened 2017-03-07
IPv6 IPs are not formatted properly in console output


mhils:
##### Steps to reproduce the problem:

1. `mitmproxy -R http://example.com`
2. `curl https://[::1]:8080`

The output currently looks like this:
```
('::1', 56795, 0, 0): clientconnect
```
But should probably look more like this:
```
[::1]:56795: clientconnect
```

##### System information

Mitmproxy version: 3.0.0.dev (2.0.0dev0064-0xbda3933)
Python version: 3.6.0
Platform: Windows-10-10.0.14393-SP0
SSL version: OpenSSL 1.0.2k  26 Jan 2017
Windows version: 10 10.0.14393 SP0 Multiprocessor Free

mhils:
Also:
```
Proxy server listening at http://:::8081
```

madhusai3113:
@mhils I would like to take a shot at this issue.

PiyushGoyal443:
Is this issue still open?

mhils:
@PiyushGoyal443: If you look closely, you'll see that there's an open PR linked above.

PR #2183, opened 2017-03-18
Fixes mitmproxy crashes when can't resolve address ( due to internet connectivity issues )

LuD1161:
##### Steps to reproduce the problem : ( There may be other ways )

1. Disconnect the internet.
2. Try to access a website, this comes up ( 172.31.1.5 is my college proxy server )

image

3. When you open the flow and move on to Details tab , this error comes up .

image

##### Reason :
This code in **/mitmproxy/mitmproxy/tools/console/flowdetailview.py** at *line 34*
```
    if sc is not None:
        text.append(urwid.Text([("head", "Server Connection:")]))
        parts = [
            ["Address", "{}:{}".format(sc.address[0], sc.address[1])],
            ["Resolved Address", "{}:{}".format(sc.ip_address[0], sc.ip_address[1])],
        ]
        if resp:
            parts.append(["HTTP Version", resp.http_version])
        if sc.alpn_proto_negotiated:
            parts.append(["ALPN", sc.alpn_proto_negotiated])

        text.extend(
            common.format_keyvals(parts, key="key", val="text", indent=4)
        )
```

Even though **sc** ( here a **ServerConnection** object defined in **/mitmproxy/mitmproxy/connections.py** ) is _not_ **None** but its attribute `sc.ip_address` is **None** , since without internet it couldn't be resolved.

#### After Fix 

image

### ToDo
I still couldn't think of any possible use cases apart from internet being disconnected in between the work.
Would love to know any other use cases where this could have prop up. 

#### System information

Mitmproxy version: 3.0.0 (2.0.0dev0203-0xc0882496) 
Python version: 3.5.3
Platform: Linux-4.6.0-kali1-amd64-x86_64-with-Kali-kali-rolling-kali-rolling
SSL version: OpenSSL 1.1.0e  16 Feb 2017
Linux distro: Kali kali-rolling kali-rolling

### DIFF
--- a/mitmproxy/tools/console/flowdetailview.py
+++ b/mitmproxy/tools/console/flowdetailview.py
@@ -29,23 +29,26 @@ def flowdetails(state, flow: http.HTTPFlow):
 
     if sc is not None:
         text.append(urwid.Text([("head", "Server Connection:")]))
-        print(sc)
-        try:
+        if sc.address:
             parts = [
                 ["Address", "{}:{}".format(sc.address[0], sc.address[1])],
-                ["Resolved Address", "{}:{}".format(sc.ip_address[0], sc.ip_address[1])],
             ]
-            if resp:
-                parts.append(["HTTP Version", resp.http_version])
-            if sc.alpn_proto_negotiated:
-                parts.append(["ALPN", sc.alpn_proto_negotiated])
+        else:
+            parts = [
+                ["Address", "Couldn't Resolve"],
+            ]
+        if sc.ip_address:
+            parts.append(["Resolved Address", "{}:{}".format(sc.ip_address[0], sc.ip_address[1])])
+        else:
+            parts.append(["Resolved Address", "Couldn't Resolve"])
+        if resp:
+            parts.append(["HTTP Version", resp.http_version])
+        if sc.alpn_proto_negotiated:
+            parts.append(["ALPN", sc.alpn_proto_negotiated])
 
-            text.extend(
-                common.format_keyvals(parts, key="key", val="text", indent=4)
-            )
-        except Exception as e:
-            print(e)
-
+        text.extend(
+            common.format_keyvals(parts, key="key", val="text", indent=4)
+        )
 
         c = sc.cert
         if c:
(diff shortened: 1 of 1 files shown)


mitmproxy/tools/console/flowdetailview.py
```diff
@@ -29,10 +29,18 @@ def flowdetails(state, flow: http.HTTPFlow):
 
     if sc is not None:
         text.append(urwid.Text([("head", "Server Connection:")]))
-        parts = [
-            ["Address", "{}:{}".format(sc.address[0], sc.address[1])],
-            ["Resolved Address", "{}:{}".format(sc.ip_address[0], sc.ip_address[1])],
-        ]
+        if sc.address:
```
mhils: Is there a case where `sc.address` is None?

mitmproxy/tools/console/flowdetailview.py
```diff
@@ -29,10 +29,18 @@ def flowdetails(state, flow: http.HTTPFlow):
...
+                ["Address", "{}:{}".format(sc.address[0], sc.address[1])],
+            ]
+        else:
+            parts = [
+                ["Address", "Couldn't Resolve"],
+            ]
+        if sc.ip_address:
+            parts.append(["Resolved Address", "{}:{}".format(sc.ip_address[0], sc.ip_address[1])])
+        else:
+            parts.append(["Resolved Address", "Couldn't Resolve"])
```
mhils: This is a bit tricky here because `.ip_address` is only set after successful connection. It could very well be that we resolved the hostname successfully, but then failed to connect. In this case, we should not display `"Couldn't Resolve"`, because that is not what is happening. I suggest to just not display "Resolved Address" when `ip_address` is None.

ujjwal96:
@LuD1161 Any update on this?

LuD1161:
@ujjwal96 
Left it for a while due to some other engagements and since was stalled earlier due to some other issue being worked upon by  @abhisheksachdeva.
Seems to me that I will have to solve the IP resolution also ( the formatting problem )

Am currently resuming work on this with effect from now.

LuD1161:
@mhils 
I contacted  @abhisheksachdeva , he said to me to go on with #2109 , but I see that @ujjwal96 is working on it so , if he finalises something I can go on with implementing it on my PR too.

mhils:
This is fixed on master now, thanks again.

Issue #2184, opened 2017-03-18
mitmweb -n: Proxy server listening at http://d:u/


mhils:
##### Steps to reproduce the problem:

1. `mitmweb -n`
2. Incorrect information is displayed on the console: "Proxy server listening at http://d:u/" 

##### Any other comments? What have you tried so far?

mitmweb should use the termstatus addon for printing this as well.

##### System information

Mitmproxy version: 3.0.0 (2.0.0dev0203-0xc088249)
Python version: 3.5.2
Platform: Linux-3.4.0+-x86_64-with-Ubuntu-14.04-trusty
SSL version: OpenSSL 1.0.2g-fips  1 Mar 2016
Linux distro: Ubuntu 14.04 trusty

PR #2187, opened 2017-03-19
Fix#2184, mitmweb -n dispalys incorrect message.

MatthewShao:
Check if the `option.server` is on before printing the information about the proxy server.
In mitmweb, add `server` to `state.setting`, so we could display the correct message in the termstatus.
default

### DIFF
--- a/web/src/js/components/Footer.jsx
+++ b/web/src/js/components/Footer.jsx
@@ -48,9 +48,11 @@ function Footer({ settings }) {
                 <span className="label label-success">stream: {formatSize(stream_large_bodies)}</span>
             )}
             <div className="pull-right">
-            <span className="label label-primary" title="HTTP Proxy Server Address">
-                { server ? ((listen_host||"*") + ":" + listen_port) : "" }
-            </span>
+                {server && (
+                    <span className="label label-primary" title="HTTP Proxy Server Address">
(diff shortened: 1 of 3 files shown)


mhils:
Thanks. I may not have been clear in #2184 - we should re-use the [termstatus addon](https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/addons/termstatus.py) for mitmweb. Also updating the web ui sounds good, but it's currently inconsistent with `mitmproxy`, which just hides the `[*:8080]` when `-n` is passed. I suggest we do the same for mitmweb. 😃

MatthewShao:
Sorry for the misunderstanding, no more `[*:8080]` when `-n` is passed.
default

web/src/js/components/Footer.jsx
```diff
@@ -49,7 +49,7 @@ function Footer({ settings }) {
             )}
             <div className="pull-right">
             <span className="label label-primary" title="HTTP Proxy Server Address">
```
mhils: You can make this 
```javascript
{ server && (
    <span className="label label-primary">...</span>
)}
```
as with the other options (see above)

MatthewShao:
Yes, that's better. I haven't noticed this notation before, thanks for the advice.

mhils:
👍

PR #2189, opened 2017-03-19
Fix IPv6 address rendering

### LINKED ISSUE #2109: IPv6 IPs are not formatted properly in console output

abhisheksachdeva:
This fixes the IPv6 formatting issue.
screenshot from 2017-03-19 17-26-07

fixes #2109

### DIFF
--- a/mitmproxy/utils/human.py
+++ b/mitmproxy/utils/human.py
@@ -62,3 +62,10 @@ def format_timestamp(s):
 def format_timestamp_with_milli(s):
     d = datetime.datetime.fromtimestamp(s)
     return d.strftime("%Y-%m-%d %H:%M:%S.%f")[:-3]
+
+
+def format_address(s):
+    # s is the address tuple
+    if "." in s[0]:
+        return s[0].split(":")[-1]+":"+str(s[1])
+    return "["+s[0]+"]:"+str(s[1])
--- a/test/mitmproxy/utils/test_human.py
+++ b/test/mitmproxy/utils/test_human.py
@@ -46,3 +46,8 @@ def test_pretty_duration():
     assert human.pretty_duration(10000) == "10000s"
     assert human.pretty_duration(1.123) == "1.12s"
     assert human.pretty_duration(0.123) == "123ms"
+
+
+def test_format_address():
+    assert human.format_address(('::1','55700','0','0')) == "[::1]:55700"
+    assert human.format_address(('::ffff:127.0.0.1','55700','0','0')) == "127.0.0.1:55700"
--- a/mitmproxy/proxy/server.py
+++ b/mitmproxy/proxy/server.py
@@ -12,6 +12,7 @@ from mitmproxy.proxy import modes
 from mitmproxy.proxy import root_context
 from mitmproxy.net import tcp
 from mitmproxy.net.http import http1
+from mitmproxy.utils import human
 
 
 class DummyServer:
@@ -152,5 +153,5 @@ class ConnectionHandler:
         self.client_conn.finish()
 
     def log(self, msg, level):
-        msg = "{}: {}".format(repr(self.client_conn.address), msg)
+        msg = "{}: {}".format(human.format_address(self.client_conn.address), msg)
         self.channel.tell("log", log.LogEntry(msg, level))


mhils:
Thanks. This fixes the use case for IPv6, but won't look right for hostnames or IPv4 addresses. How about we add a utility method to format `(addr, port)` tuples in https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/utils/human.py?

Kriechi:
👍 for a new utility function. We probably should use is in a few more places (see https://github.com/mitmproxy/mitmproxy/commit/3e9125a3c13084f4abb0e5b74554909f24df9454).

abhisheksachdeva:
@mhils Working on it. I am not able to produce the same format while using hostnames, such as 
1.  `mitmproxy -R http://example.com`

abhisheksachdeva:
@mhils Working on it. But I am not able to reproduce the issue while working with hostnames, such as
1.  `mitmproxy -R example.com`
2.  `curl https://example.com:8080`
Any help would be appreciated.

@Kriechi I'll add it in other places too.

abhisheksachdeva:
@mhils Hi, I have updated the PR. Please have a look at it.

mitmproxy/utils/human.py
```diff
@@ -62,3 +62,9 @@ def format_timestamp(s):
 def format_timestamp_with_milli(s):
     d = datetime.datetime.fromtimestamp(s)
     return d.strftime("%Y-%m-%d %H:%M:%S.%f")[:-3]
+
+
+def format_ip_address(addr, port):
```
mhils: 1. Let's call this `format_address` - after all, it should also work for hostnames.
2. Let's just accept a single `(host, port)` tuple here, not two different arguments. That's not particularly beautiful, but matches all the other Python socket APIs.

abhisheksachdeva: Thanks for the review, I'll append these changes.

abhisheksachdeva: @mhils done! 
Please review the changes.

test/mitmproxy/utils/test_human.py
```diff
@@ -46,3 +46,8 @@ def test_pretty_duration():
     assert human.pretty_duration(10000) == "10000s"
     assert human.pretty_duration(1.123) == "1.12s"
     assert human.pretty_duration(0.123) == "123ms"
+
+
+def test_format_address():
+    assert human.format_address(('::1','55700','0','0')) == "[::1]:55700"
+    assert human.format_address(('::ffff:127.0.0.1','55700','0','0')) == "127.0.0.1:55700"
```
mhils: This is missing tests for plain IPv4 addresses and hostnames.

mitmproxy/utils/human.py
```diff
@@ -62,3 +62,10 @@ def format_timestamp(s):
 def format_timestamp_with_milli(s):
     d = datetime.datetime.fromtimestamp(s)
     return d.strftime("%Y-%m-%d %H:%M:%S.%f")[:-3]
+
+
+def format_address(s):
+    # s is the address tuple
```
mhils: This function should have a proper docstring explaining that it accepts hostnames and IPv4/IPv6 address tuples. It may also be useful to add type annotations to make this clear.

test/mitmproxy/utils/test_human.py
abhisheksachdeva: Even if I use IPv4 address, the `client_conn.address` will convert it into IPv6 type.
eg. if I use `curl 127.0.0.1:8080`, client_conn.address[0] will be `::ffff:127.0.0.1`.
That's why I've used this assertion. 
Also, in the case of hostnames, client_conn.address will be an IPv6 address only, as you had pointed earlier in our slack discussion. This method takes a tuple having an IP address and identifies the version (v4 or v6).
So, aren't these tests enough for the `format_address` method?

mhils: As I mentioned a couple of times now, this function will also be used in other contexts (see e.g. https://github.com/mitmproxy/mitmproxy/pull/2183#pullrequestreview-27747348) where it can be a plain IPv4 address or a hostname.

mhils:
Any update on this?

abhisheksachdeva:
I am currently busy, will continue this in a few days.
Sorry for the delay.

mhils:
No worries, please take your time. Just wanted to make sure that this doesn't go stale.

PR #2190, opened 2017-03-20
mypy for all of ./mitmproxy

cortesi:
Does whatever is needed to make ./mitmproxy pass mypy, and enables this in tox.

Also factors out an io module and pulls tnetstring out of contrib.


mitmproxy/tools/console/flowlist.py
```diff
@@ -305,7 +306,9 @@ def get_prev(self, pos):
 
 class FlowListBox(urwid.ListBox):
 
-    def __init__(self, master: "mitmproxy.tools.console.master.ConsoleMaster"):
+    def __init__(
+        self, master: "mitmproxy.tools.console.master.ConsoleMaster"
```
Kriechi: Do we still need the quotes around this?

mitmproxy/tools/console/flowview.py
```diff
@@ -103,7 +104,11 @@ def _mkhelp():
 
 class FlowViewHeader(urwid.WidgetWrap):
 
-    def __init__(self, master: "mitmproxy.console.master.ConsoleMaster", f: http.HTTPFlow):
+    def __init__(
+        self,
+        master: "mitmproxy.tools.console.master.ConsoleMaster",
```
Kriechi: quotes?

mitmproxy/tools/console/flowlist.py
cortesi: True. Fixed.

PR #2195, opened 2017-03-21
safer url handling

Kriechi:
While playing around I got a traceback and mitmproxy crashed due to host or port not being a proper value.
This PR makes the url handling safe to use even if the values are `None`.

### DIFF
--- a/mitmproxy/net/http/request.py
+++ b/mitmproxy/net/http/request.py
@@ -68,13 +68,9 @@ class Request(message.Message):
         self.data = RequestData(*args, **kwargs)
 
     def __repr__(self):
-        if self.host and self.port:
-            hostport = "{}:{}".format(self.host, self.port)
-        else:
-            hostport = ""
         path = self.path or ""
         return "Request({} {}{})".format(
-            self.method, hostport, path
+            self.method, mitmproxy.net.http.url.hostport(self.scheme, self.host, self.port), path
         )
 
     @classmethod
@@ -283,7 +279,7 @@ class Request(message.Message):
         The URL string, constructed from the request's URL components
         """
         if self.first_line_format == "authority":
-            return "%s:%d" % (self.host, self.port)
+            return "{}:{}".format(self.host, self.port)
         return mitmproxy.net.http.url.unparse(self.scheme, self.host, self.port, self.path)
 
     @url.setter
--- a/mitmproxy/net/http/url.py
+++ b/mitmproxy/net/http/url.py
@@ -77,9 +77,11 @@ def unparse(scheme, host, port, path=""):
     Args:
         All args must be str.
     """
-    if path == "*":
+    if not path or path == "*":
         path = ""
-    return "%s://%s%s" % (scheme, hostport(scheme, host, port), path)
+    if not scheme:
+        scheme = "<unknown-scheme>"
+    return "{}://{}{}".format(scheme, hostport(scheme, host, port), path)
 
 
 def encode(s: Sequence[Tuple[str, str]], similar_to: str=None) -> str:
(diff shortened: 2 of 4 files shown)


Kriechi:
macOS latest with webserver running on 127.0.0.1:8384

```
$ mitmdump -p 8384
Proxy server listening at http://:::8384
('::1', 59314, 0, 0): clientconnect
  File "/Users/kriechi/.pyenv/versions/mitmproxy-3.5.2/bin/mitmdump", line 11, in <module>
    load_entry_point('mitmproxy', 'console_scripts', 'mitmdump')()
  File "...mitmproxy/tools/main.py", line 126, in mitmdump
    m = run(dump.DumpMaster, args, extra)
  File "...mitmproxy/tools/main.py", line 94, in run
    master.run()
  File "...mitmproxy/master.py", line 80, in run
    self.tick(0.1)
  File "...mitmproxy/master.py", line 96, in tick
    self.addons.handle_lifecycle(mtype, obj)
  File "...mitmproxy/addonmanager.py", line 79, in handle_lifecycle
    self.trigger(name, message)
  File "...mitmproxy/addonmanager.py", line 116, in trigger
    self.invoke_addon(i, name, *args, **kwargs)
  File "...mitmproxy/addonmanager.py", line 107, in invoke_addon
    func(*args, **kwargs)
  File "...mitmproxy/addons/dumper.py", line 228, in error
    self.echo_flow(f)
  File "...mitmproxy/addons/dumper.py", line 196, in echo_flow
    self._echo_request_line(f)
  File "...mitmproxy/addons/dumper.py", line 127, in _echo_request_line
    url = flow.request.url
  File "...mitmproxy/net/http/request.py", line 287, in url
    return mitmproxy.net.http.url.unparse(self.scheme, self.host, self.port, self.path)
  File "...mitmproxy/net/http/url.py", line 82, in unparse
    return "%s://%s%s" % (scheme, hostport(scheme, host, port), path)
  File "...mitmproxy/net/http/url.py", line 140, in hostport
    return "%s:%d" % (host, port)
TypeError: %d format: a number is required, not NoneType
```

and

`curl http://localhost:8384/`

mhils:
Just to be clear, you expect mitmproxy to fail, just in a different way?

Kriechi:
With this PR, it behaves like this:
```
Proxy server listening at http://:::8384
('::1', 59376, 0, 0): clientconnect
('::1', 59376, 0, 0): GET None://None:None/
 << Invalid HTTP request form (expected: authority or absolute, got: relative)
('::1', 59376, 0, 0): HTTP protocol error in client request: Invalid HTTP request form (expected: authority or absolute, got: relative)
('::1', 59376, 0, 0): clientdisconnect
```

mhils:
This looks like we should return `http://unknown/path` in `request.url` when `request.host` is `None` ?

cortesi:
Agree with @mhils here - this should be fixed in request.url. I'd prefer some notation to make clear that we're not just dealing with a host called "unknown". Maybe ```http://<unknown>/path```?

Kriechi:
This unfortunately affect a bit more than `request.url`. I updated the PR, please take a look.

cortesi:
The fact that port, scheme or host can be unspecified is not an attribute of URLs, it's an attribute of a corner case of our Request representation. Modifying the way we parse and unparse URLs is simply the wrong layer to fix this. 

Here are two more principled solutions. Option 1: change the Request.url method to do something a bit like this:

```python
    @property
    def url(self):
        """
        The URL string, constructed from the request's URL components
        """
        if self.first_line_format == "authority":
            return "%s:%d" % (self.host, self.port)
        elif not self.port:
            # We have a corner case where port, host and scheme can be None if
            # there's an error. It's not clear whether there are further
            # circumstances where a subset of these might be None, so let's die
            # loudly if that's the case.
            if self.host or self.scheme:  # pragma: no cover
                raise NotImplementedError("Unexpected partial request.")
            return "<unknown-scheme>://<unknown-host>" + self.path
        else:
            return mitmproxy.net.http.url.unparse(
                self.scheme,
                self.host,
                self.port,
                self.path
            )
```

The second option is to look a bit more deeply into why we get to this juncture in the first place. You can make a good argument that a malformed partial request like this should never make it through our event mechanism, and that trying to pass the request to the error handler at this juncture is wrong. That would mean that we construct an error without a request attribute, that instead includes text to describe what went wrong (including the partial URL), analogous to what we do if there's a protocol error. This means fixing things in *protocol/http.py*. 

Thinking about it, I actually prefer option 2, which clarifies all of this stuff and also clears up an ambiguity that might trip us up elsewhere.

cortesi:
I've put a minimal PR up at #2203 that prevents passing malformed requests over the event mechanism when there's an HTTP protocol error. There's more we can do there to give users a clearer idea of what happened, but this seems like a good start.

cortesi:
Discussion of how to fix this has moved to #2203.

PR #2200, opened 2017-03-22
Add coverage for eventLog.js and update Jest config.

MatthewShao:
This one is easy to test.
default

The `testPathDirs` in Jest configuration has been replaced by `roots`, so we update that to avoid the deprecation warning.


mhils:
Thanks!

PR #2203, opened 2017-03-22
Don't pass malformed request objects to our error handler

cortesi:


mhils:
+1 to this, but I think this still passes an invalid request object to `requestheaders`. Can we call `validate_request_form()` right after reading the request, i.e. before `f.request = request`?

cortesi:
Hmm... let's think about this. First, an observation - if the user can change scheme or first_line_format in the request_header handler, then we'll need to re-validate after the handler is called. Second, a question: are there circumstances where we want the user to be able to intercept an invalid inbound request, and turn it into a valid one? Previously this could be done in request_headers, but if we validate first that would be impossible.

cortesi:
I've amended this to validate the request form pre and post the `requestheaders` event. We have to zero the request field on the flow anyway to handle post validation, so I leave that intact.

PR #2207, opened 2017-03-23
scripts: redirect stdout to ctx.log.warn

### LINKED ISSUE #1530: Handle print() statements for inline scripts
print() statements in inline scripts should be suppressed, and produce into ctx.log.warn() calls instead.

nikofil:
Redirect messages written to stdout in scripts to ctx.log.warn.
(closes #1530)

### DIFF
--- a/mitmproxy/addons/script.py
+++ b/mitmproxy/addons/script.py
@@ -65,14 +65,28 @@ def cut_traceback(tb, func_name):
         return tb
 
 
+class StreamLog:
+    """
+        A class for redirecting output using contextlib.
+    """
+    def __init__(self, log):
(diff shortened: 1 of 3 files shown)


mhils:
This looks great, thanks. Could you please add a test?

nikofil:
Thanks! Will do.
There is a small issue, it seems like `print` is calling `sys.stdout.write` more than once to print a newline, resulting in empty lines in the log. Should I change it to not print empty lines?

mhils:
Yes, that sounds about right. Thanks!

cortesi:
Good stuff, thanks.

PR #2209, opened 2017-03-24
[WIP] options: use new Path class for path option

### LINKED ISSUE #2198: Teach options to know about a Path type
At the moment, the type we use for Path values is str. We use paths in many options, and we'd like to handle them specially in user interfaces (so we can, for instance, do path completion on tab). 

We should teach optmanager to know about a Path type, which is treated just like a string. Ergonomically, I'd like user to still be able to say ```options.path = "value"``` rather than have to say ```options.path = types.Path("value")```, so it would be good to automatically handle type promotion.

nikofil:
Add a Path class for use with user options. (closes #2198)

Signed-off-by: Nikos Filippakis <aesmade@gmail.com>

### DIFF
--- a/mitmproxy/options.py
+++ b/mitmproxy/options.py
@@ -7,6 +7,8 @@ from mitmproxy.net import tcp
 # We redefine these here for now to avoid importing Urwid-related guff on
 # platforms that don't support it, and circular imports. We can do better using
 # a lazy checker down the track.
+from mitmproxy.optmanager import Path
+
 console_palettes = [
     "lowlight",
     "lowdark",
@@ -24,7 +26,7 @@ view_orders = [
 
 APP_HOST = "mitm.it"
 APP_PORT = 80
-CA_DIR = "~/.mitmproxy"
+CA_DIR = Path("~/.mitmproxy")
 LISTEN_PORT = 8080
 
 # Some help text style guidelines:
@@ -218,7 +220,7 @@ class Options(optmanager.OptManager):
             """
         )
         self.add_option(
-            "cadir", str, CA_DIR,
+            "cadir", Path, CA_DIR,
             "Location of the default mitmproxy CA files."
         )
         self.add_option(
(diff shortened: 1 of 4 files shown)


mitmproxy/options.py
```diff
@@ -7,6 +7,8 @@
 # We redefine these here for now to avoid importing Urwid-related guff on
 # platforms that don't support it, and circular imports. We can do better using
 # a lazy checker down the track.
+from mitmproxy.optmanager import Path
```
mhils: We already import `optmanager`, so let's just use `optmanager.Path` below.

mitmproxy/optmanager.py
```diff
@@ -20,6 +20,27 @@
 unset = object()
 
 
+class Path(str):
```
mhils: Would it make sense to use [`pathlib.Path`](https://docs.python.org/3/library/pathlib.html) instead?

cortesi:
Hi @nikofil - thanks for working on this. Let's ponder the approach here a bit. I would prefer not to have a fat Path object of our own - if we can get away with something lighter weight, we should. 

I see @mhils suggests we use pathlib.Path. It's slightly more than a simple type, in the sense that it instantiates different concrete types depending on the platform we're on. We need to do some analysis to see if there are hidden traps, but if there aren't, this might be a good option. 

Another possibility is to simply declare Path as an alias for str with the [typing module](https://docs.python.org/3/library/typing.html#newtype). That means we can keep the semantics almost identical to what they are now, just with a narrower type spec that will let us treat things specially in the UIs.

nikofil:
Thank you for the feedback.

`pathlib.Path` can indeed be used nicely in the place of `str` for paths, I managed to get mitmproxy running with it rather quickly. However there is a lot of searching and replacing involved and I am a bit worried about changing that many lines, so hopefully the tests will catch any errors that I don't.

I will continue in this direction then and ask you again if I get stuck somewhere.

mhils:
I do like @cortesi's aliasing proposal as well. Please go for whatever you feel is more appropriate! 😃

PR #2212, opened 2017-03-25
[web] Update Jest config and coverage ++

MatthewShao:
This PRs is for:
+ Add a small js dependency `jest-fetch-mock`, because `node` does not support the `fetch` function, we have to mock it when running the test. This dependency would make our test jobs easier when dealing with some async actions.
+ Add `setupJest.js` to load `jest-fetch-mock` when starting Jest.
+ Add `--coverage` argument when doing `npm test`, so we could have the coverage report on travis.
+ Reach 100% coverage for `ducks/settings.js`

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -14,11 +14,11 @@
     "unmockedModulePathPatterns": [
       "react"
     ],
-      "coverageDirectory":"./coverage",
-      "collectCoverage": true,
-      "coveragePathIgnorePatterns": [
-          "<rootDir>/src/js/filt/filt.js"
-      ]
+    "coverageDirectory":"./coverage",
+    "collectCoverage": true,
+    "coveragePathIgnorePatterns": [
+        "<rootDir>/src/js/filt/filt.js"
+    ]
   },
   "dependencies": {
     "bootstrap": "^3.3.7",
--- a/web/package.json
+++ b/web/package.json
@@ -13,7 +13,9 @@
     ],
     "unmockedModulePathPatterns": [
       "react"
-    ]
+    ],
+      "coverageDirectory":"./coverage",
+      "collectCoverage": true
   },
   "dependencies": {
     "bootstrap": "^3.3.7",
--- a/web/package.json
+++ b/web/package.json
@@ -15,7 +15,10 @@
       "react"
     ],
       "coverageDirectory":"./coverage",
-      "collectCoverage": true
+      "collectCoverage": true,
+      "coveragePathIgnorePatterns": [
+          "<rootDir>/src/js/filt/filt.js"
+      ]
   },
   "dependencies": {
     "bootstrap": "^3.3.7",
--- a/web/src/js/__tests__/ducks/settingsSpec.js
+++ b/web/src/js/__tests__/ducks/settingsSpec.js
@@ -1,8 +1,7 @@
 jest.unmock('../../ducks/settings')
 jest.mock('../../utils')
 
-import reduceSettings from '../../ducks/settings'
-import * as SettingsActions from '../../ducks/settings'
+import reduceSettings, * as SettingsActions from '../../ducks/settings'
 
 describe('setting reducer', () => {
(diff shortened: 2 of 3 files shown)


mhils:
Now that's an interesting one. When we introduced jest for mitmweb, it still had automocking enabled by default, but [this has apparently changed](https://facebook.github.io/jest/blog/2016/09/01/jest-15.html). As such, all the nice `jest.unmock('...')` statements we have don't serve any purpose right now 🙈. We need to make a decision on this to either

- manually re-enable automocking
- follow along and remove all those unmock statements.

@cortesi, any preferences? `automock` is unfortunately boolean an would apply to the whole codebase.

Coming back to this PR: Do we actually need `jest-fetch-mock`? For this particular purpose it looks like using plain jest to mock `utils.fetchApi` would be good enough. 

Somewhat related, we should add JS coverage to codecov: https://github.com/codecov/example-node provides some examples.

MatthewShao:
@mhils 
> Do we actually need `jest-fetch-mock`?

For this question, I have reconsidered, it seems that for now we don't need to reduce the state base on the response of `fetch`. So, no, we don't have to add this now. 
And we have two solution to mock the fetch:
1. explicitly use `global.fetch = jest.fn()` in every test file which will invokes `fetch`
2. keep the new `setupJest.js` and write `global.fetch=jest.fn()` in it

but if we want to reduce the state base on the response from the server in the future, we'd better use this `jest-fetch-mock` dependency to simplify the mocking works.

For the decision about automock, I personally have no strong preference. 
About adding JS coverage to codecov, I will certainly keep that in mind but I feel it is still too early to do so... 

Please leave a message about how to fix this PR when you have a decision. Thanks!

mhils:
> About adding JS coverage to codecov, I will certainly keep that in mind but I feel it is still too early to do so...

There is no too early for this. We should add it even if it initially shows that we have very little coverage. No wrong excuses. :)

I agree on `jest-fetch-mock` - we may need it at some point, but not for this. Can we just mock `utils`/`utils.fetchApi` here? That's what automock would do.

mhils:
Looks good to me. We should exclude `web/src/js/filt/filt.js` from coverage as that's autogenerated and for the most part impossible to test.

@Kriechi @cortesi: You can both survive the drop in coverage?

web/package.json
```diff
@@ -13,7 +13,12 @@
     ],
     "unmockedModulePathPatterns": [
       "react"
-    ]
+    ],
+      "coverageDirectory":"./coverage",
```
mhils: nitpick: wrong indentation

web/src/js/__tests__/ducks/settingsSpec.js
```diff
@@ -0,0 +1,23 @@
+jest.unmock('../../ducks/settings')
+jest.mock('../../utils')
+
+import reduceSettings from '../../ducks/settings'
+import * as SettingsActions from '../../ducks/settings'
```
mhils: We can combine this to `import reduceSettings, * as SettingsActions from '../../ducks/settings'`

web/src/js/__tests__/ducks/settingsSpec.js
```diff
@@ -0,0 +1,23 @@
...
+    it('should handle receive action', () => {
+        let action = { type: SettingsActions.RECEIVE, data: 'foo' }
+        expect(reduceSettings(undefined, action)).toEqual('foo')
+    })
+
+    it('should handle update action', () => {
+        let action = {type: SettingsActions.UPDATE, data: {id: 1} }
+        expect(reduceSettings(undefined, action)).toEqual({id: 1})
+
+        expect(reduceSettings(undefined, SettingsActions.update())).toEqual({})
```
mhils: This line should probably be in a `describe('settings actions', ...` block?

MatthewShao:
@mhils Appreciate for the advices!

PR #2219, opened 2017-03-28
[web] Coverage++ for ducks/flows.js

MatthewShao:
+ Could not cover the line below, since we still don't have multiple selection implementation:
https://github.com/mitmproxy/mitmproxy/blob/master/web/src/js/ducks/flows.js#L43
+ Not sure whether I am doing this right, this could add up coverage, but obviously not a perfect solution:
https://github.com/MatthewShao/mitmproxy/blob/15c19d2bc63e667a95bdc1f6f5b5af056264280f/web/src/js/__tests__/ducks/flowsSpec.js#L90
I want to test if we are passing the right parameter(URL) to `fetchApi` here.  I googled a lot, but still got no idea about how to mock this function correctly. Any hints?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ MANIFEST
 .cache/
 .tox*/
 build/
+mitmproxy/contrib/kaitaistruct/*.ksy
 
 # UI
 
@@ -21,3 +22,4 @@ sslkeylogfile.log
 .tox/
 .python-version
 coverage.xml
+web/coverage/
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,8 +52,10 @@ matrix:
       before_install:
           - curl -o- -L https://yarnpkg.com/install.sh | bash
           - export PATH=$HOME/.yarn/bin:$PATH
-      install: cd web && yarn
-      script: npm test
+      install:
+          - cd web && yarn
+          - yarn global add codecov
+      script: npm test && codecov
       cache:
           yarn: true
           directories:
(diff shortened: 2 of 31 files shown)


mhils:
Ok, I finally found some time to look at this! 😃 
I needed to play around a bit to answer your question, so I already addressed some comments with my latest commit.

> Could not cover the line below, since we still don't have multiple selection implementation

I made one up: https://github.com/mitmproxy/mitmproxy/pull/2219/files#diff-03885a709bb05c5e6ceff6acbc875b2fR79

> I want to test if we are passing the right parameter(URL) to fetchApi here. I googled a lot, but still got no idea about how to mock this function correctly. Any hints?

Two parts to this: 
First, we should use a full store when testing redux-thunk action creators (the ones which return a `(dispatch, getState)` function). I've added an example [here](https://github.com/mitmproxy/mitmproxy/pull/2219/files#diff-03885a709bb05c5e6ceff6acbc875b2fR111), the other actions need to be adjusted. If it gets more complicated, we should use a mock store as described in http://redux.js.org/docs/recipes/WritingTests.html#async-action-creators, but let's keep it minimal for now.
Second, we can `jest.mock('../../utils'); import {fetchApi} from "../../utils"` and then use 
```javascript
expect(fetchApi).toBeCalledWith('/flows/1/resume', { method: 'POST' })
```

MatthewShao:
Awesome! Finally got a satisfying solution.

mhils:
🎉

PR #2220, opened 2017-03-28
eventlog: Override set_focus to check for index error

krsoninikhil:
Mitmproxy crashes on pressing `g` or `G` while `eventlog` is focused and is empty. This happens due to `IndexError` in `set_focus`, so this PR just put a check by overriding it.

#### Mitmproxy version

Mitmproxy version: 3.0.0 (1.0.1dev0502-0x60b2fdfe) 
Python version: 3.6.0
Platform: Linux-4.8.13-1-ARCH-x86_64-with-arch
Linux distro: arch

PS: Should I open a issue before sending PR?

### DIFF
--- a/mitmproxy/tools/console/flowlist.py
+++ b/mitmproxy/tools/console/flowlist.py
@@ -57,6 +57,10 @@ class LogBufferBox(urwid.ListBox):
         self.master = master
         urwid.ListBox.__init__(self, master.logbuffer)
 
+    def set_focus(self, index):
+        if 0 <= index < len(self.master.logbuffer):
+            super().set_focus(index)
+
     def keypress(self, size, key):
         key = common.shortcuts(key)
         if key == "z":


iharsh234:
@krsoninikhil  i think you doing it wrong way :thinking: 
it should be 
```
def set_focus(self, index):
        if self.master.view.inbounds(index):
            super().set_focus(index)
```
check [`addons/views.py`](https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/addons/view.py#L149) and  [`flowlist.py`](https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/tools/console/flowlist.py#L287)

mitmproxy/tools/console/flowlist.py
```diff
@@ -57,6 +57,10 @@ def __init__(self, master):
         self.master = master
         urwid.ListBox.__init__(self, master.logbuffer)
 
+    def set_focus(self, index):
+        if(0 <= index < len(self.master.logbuffer)):
```
Kriechi: Please remove the `(...)` around the if condition - we don't use this style of code in mitmproxy.

krsoninikhil: Oh sorry, should have noticed.

Kriechi:
Thanks! 👍 🎉

krsoninikhil:
@Kriechi Should I also write the tests to cover this?

Kriechi:
Yes please! I'm not sure how easy this can be tested, but give it a try - thanks!

krsoninikhil:
I added some tests, can you please check #2223?

Issue #2194, opened 2017-03-21
Extend mypy checking


cortesi:
A somewhat smaller amount of work than what was needed to cover the mitmproxy directory. I'd like to get us to the point where we have mypy checking for the entire codebase. 

  - [x] mitmproxy
  - [x] pathod
  - [x] examples/simple
  - [ ] examples/complex
  - [ ] examples/pathod

iharsh234:
@cortesi can i take this?

mhils:
@iharsh234: Sure, please go ahead.

kira0204:
Working on this

kira0204:
Hi,
When I was working on this I noticed that `examples/complex/xss_scanner.py` contains many errors. Some of those errors are calling `decode()` attribute on `str` tyoe. Can I edit the code as well?

kira0204:
I am not able to understand why there is a difference when we are talking about XSS and SQL

https://github.com/mitmproxy/mitmproxy/blob/6dd336fcec6dc32c4986b6c20189bf1b2132153c/examples/complex/xss_scanner.py#L168-L184

And in the test file as well, this has been done 

https://github.com/mitmproxy/mitmproxy/blob/6dd336fcec6dc32c4986b6c20189bf1b2132153c/test/examples/test_xss_scanner.py#L331-L350

kira0204:
@mhils @cortesi @Kriechi What do you guys think?

Kriechi:
The XSS scanner example was contributed by @ddworken in #1907 - I don't think any of us is deeply familiar with the example in this case, sorry.

kira0204:
Aah okay, I will wait for @ddworken to reply

ddworken:
@kira0204 If you look at what info is actually being logged, it is slightly different data so I kept it in different functions. The XSS logger logs a suggested exploit and the line while the SQLi logger logs the regex used and the suspected DBMS. 

You're definitely right though that there are a bunch of mypy errors, I had been using pycharm and apparently its checker isn't as comprehensive. Let me know if you want a hand with fixing it up. :) I'd also be happy to help out and take care of fixing up the XSS scanner if you want to focus elsewhere.

kira0204:
@ddworken Thank you so much for replying. I got why two separate functions are used but I don't understand why `sqli_info.url` needs to be decoded? They are already in the `str` format.

ddworken:
[...]

PR #2221, opened 2017-03-28
Extend mypy example/simple (#2194)

iharsh234:
Extended mypy checking to `example/simple`.
I am new to mypy , thats why worked on small part once get reviewed and get most of it , I will be updating whole` example/ `.
REF : #2194

### DIFF
--- a/examples/simple/add_header.py
+++ b/examples/simple/add_header.py
@@ -1,2 +1,5 @@
-def response(flow):
+from mitmproxy import http
+
+
+def response(flow: http.HTTPFlow) -> None:
     flow.response.headers["newheader"] = "foo"
--- a/examples/simple/filter_flows.py
+++ b/examples/simple/filter_flows.py
@@ -2,12 +2,12 @@
 This scripts demonstrates how to use mitmproxy's filter pattern in scripts.
 """
 from mitmproxy import flowfilter
-from mitmproxy import ctx
+from mitmproxy import ctx, http
 
 
 class Filter:
     def __init__(self):
-        self.filter = None
+        self.filter = None  # type: flowfilter.TFilter
 
     def configure(self, updated):
         self.filter = flowfilter.parse(ctx.options.flowfilter)
@@ -17,7 +17,7 @@ class Filter:
             "flowfilter", str, "", "Check that flow matches filter."
         )
 
-    def response(self, flow):
+    def response(self, flow: http.HTTPFlow) -> None:
         if flowfilter.match(self.filter, flow):
             print("Flow matches filter:")
             print(flow)
--- a/examples/simple/wsgi_flask_app.py
+++ b/examples/simple/wsgi_flask_app.py
@@ -10,7 +10,7 @@ app = Flask("proxapp")
 
 
 @app.route('/')
-def hello_world():
+def hello_world() -> str:
     return 'Hello World!'
 
 
--- a/examples/simple/add_header_class.py
+++ b/examples/simple/add_header_class.py
@@ -1,5 +1,8 @@
+from mitmproxy import http
+
+
 class AddHeader:
-    def response(self, flow):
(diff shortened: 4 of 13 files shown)


iharsh234:
@MatthewShao thanks , I know about the error and fixed it locally , just don't want to create unnecessary commits , waiting for reviews as if i am doing it right , if some changes is confirmed , I will be updating my commit.

mhils:
Please push what you have, we can squash later. I usually skip PRs during review where CI fails :wink:

iharsh234:
@mhils  done , passes CI :smile:

examples/simple/wsgi_flask_app.py
```diff
@@ -6,15 +6,16 @@
...
 
 
 @app.route('/')
-def hello_world():
+def hello_world() -> str:
     return 'Hello World!'
 
 
-def load(l):
+def load(l) -> wsgiapp.WSGIApp:
```
Kriechi: I think we should keep the generic types for our hooks as defined in mitmproxy-core. Otherwise the type information is different for each addon, which is basically useless information... @cortesi, @mhils?

iharsh234: Should we remove this or add `types.ModuleType` here.

examples/simple/add_header.py
```diff
@@ -1,2 +1,6 @@
-def response(flow):
+from mitmproxy import http
+import typing
+
+
+def response(flow: http.HTTPFlow) -> typing.Any:
```
mhils: Why is this returning `Any`?

examples/simple/filter_flows.py
```diff
@@ -4,14 +4,15 @@
...
 """
 import sys
-from mitmproxy import flowfilter
+from mitmproxy import flowfilter, http
+import typing
 
 
 class Filter:
-    def __init__(self, spec):
+    def __init__(self, spec: str) -> None:
```
mhils: Is the `-> None` required to make the tests pass?

iharsh234: `None` is required to pass mypy test. 
here is output without `None `
`filter_flows.py:12: error: The return type of "__init__" must be None`
I have cross checked with previous commits for extending mypy , and they all seems to using None with` __init__`

examples/simple/add_header.py
iharsh234: `Any` has no significance here. since return type in` None` , it good to put `Any` than to ignore type.

Kriechi: `response()` shouldn't return anything, should it?
At least the [docs](http://docs.mitmproxy.org/en/latest/scripting/events.html#http-events) don't mention anything.
So the correct return type should be `None`.

examples/simple/filter_flows.py
Kriechi: Yes, as @iharsh234 said. ` -> None` is the correct way to do it.

Kriechi:
@iharsh234 thanks for addressing our comments so far! 👍 
We are very sorry, but we due to recent refactoring we now got a few conflicts.
Would you mind resolving those, so we can get this PR merged?

iharsh234:
I have updated my branch with the master

Kriechi:
LGTM! Thanks! 👍 
@cortesi can we merge this - or do you have more conflicting changes lined up?

iharsh234:
Can i try my hands on to rest of `/example` , ` /test `& `/pathdoc`
@Kriechi @cortesi @mhils

Kriechi:
Sure! 👍 
I'd go for `pathod` first - otherwise we get conflicts with @cortesi 's recent efforts.
Not sure if we need/want mypy-checking for tests, @mhils?

Please make sure to update newly added mypy-checking folders here: https://github.com/mitmproxy/mitmproxy/blob/83f87607036aad642515efe4fcdb0de9a8e31ed6/tox.ini#L30

PR #2224, opened 2017-03-29
Replace phrase "an TLS" with "a TLS"

jonathanrfisher1:
Updating usage of 'a' or 'an' to be more consistent with general English grammatical rules. TLS is pronounced with a consonant sound for both the expanded and abbreviated versions.


Kriechi:
Thanks! 👍 🎉

PR #2227, opened 2017-04-02
Replace phrase 'the the' in 2 places

vinaydargar:
Replaced with the appropriate "in the" and "to the".
There are multiple other occurences of this but they are in upstream python modules.


mhils:
Thanks!

PR #2232, opened 2017-04-07
fixes ipv6 authority form parsing in CONNECT

r1b:
I noticed the following output in the event log:

```
127.0.0.1:58169: HTTP protocol error in client request: Invalid host specification: b'[2a03:2880:f003:c07:face:b00c::2]:443'
```

It appears that the authority form parser does not handle ipv6 hostnames properly. I changed the parser to:

* Split on a single colon from the right-hand-side
* Strip brackets when an ipv6 hostname is detected

I also added a test for this condition.


mhils:
Awesome, thanks! 😃 🍰 🎉

PR #2248, opened 2017-04-18
Copy change (lodge -> log)

fuho:
Found in two files.


mhils:
Thanks. I believe ["lodge"](https://www.merriam-webster.com/dictionary/lodge) is actually intended here though. 😃

PR #2251, opened 2017-04-20
Hack that fixes #2228

### LINKED ISSUE #2228: Sticky cookies are improperly formatted.

1. Go to http://www.html-kit.com/tools/cookietester/
2. Click 'Set Test Cookie'
3. Observe that one cookie is sent to the server.
4. Remove the cookie.
5. launch mitmproxy with `mitmproxy -t html-kit\.com` and tell your browser to use it as a proxy
6. Reload the page.
7. Click 'Set Test Cookie'
8. Observe that two 'cookies' are sent to the server.

There appears to be a comma in the output of mitmproxy, even though it is surrounded by quotes. It's possible, then that this is a parsing fail on the tool's end caused by a difference in what's sent back for the format of the date. Still, should it really be changing that?

##### System information
Arch Linux, freshly updated.

Mitmproxy version: 2.0.1 (release version) 
Python version: 3.6.0
Platform: Linux-4.10.6-1-ARCH-x86_64-with-glibc2.3.4

kitling:
The stickycookies implementation stores a list of lists of `(name, value)` tuples, including attributes, instead of a list of `(name, value, attributes=[(name, value), ...])` tuples that represent each cookie. This hack excludes everything except the first (name, value) pair, which is the actual cookie.

See #2228 for discussion on what would be a *proper* fix.

### DIFF
--- a/mitmproxy/addons/stickycookie.py
+++ b/mitmproxy/addons/stickycookie.py
@@ -78,7 +78,7 @@ class StickyCookie:
                     ]
                     if all(match):
                         c = self.jar[(domain, port, path)]
-                        l.extend([cookies.format_cookie_header(c[name].items(multi=True)) for name in c.keys()])
+                        l.extend([cookies.format_cookie_header([c[name].items(multi=True)[0]]) for name in c.keys()])
             if l:
                 # FIXME: we need to formalise this...
                 flow.request.stickycookie = True


mhils:
Thanks! I went for a more throughout fix in #2259, so closing this here. 😃

PR #2258, opened 2017-04-24
Integrate readstdin into readfile

mhils:
This PR integrates the readstdin addon into the readfile addon, which made mitmdump unusable on Windows under some consoles where `sys.stdin.isatty()` returns False. This is now back to the explicit approach where "-" means reading from stdin.


cortesi:
This doesn't address the issue that resulted in the split in the first place: reading from stdin breaks tools like mitmproxy console. I also like the clarity of having the functionality in two separate addons much better. If we're going to have an explicit flag for reading from stdin, I'd prefer something like a --stdin flag respected by the readstdin addon - there's no reason for it to be a magic filename.

mhils:
Ok. I think `-` is a fairly well-established practice for command line tools, so I think it'd be great to keep that. At the very least, I'd prefer that over having both `-r` and `--stdin`.

We can limit the readfile addon to files again, but add `-` as a special case that gets ignored by the addon. We then keep the separate readstdin addon that handles just this in mitmdump. Does that work for you?

cortesi:
It's a well-established convention, but only for tools where reading from stdin is appropriate (or possible). ;)  In interactive contexts, and for tools that can't read from stdin I want passing "-" to readfile to be an error (unless there's a file called "-" in the current directory). 

If you'd really like to special-case "-" for mitmdump I can live with it, but I think the best way is to inherit from readilfe.Readfile to make a special ReadFileStdin and over-ride `load_flows_file` to special-case "-". Leaving the current ReadFile pure is the best way to go - it's soon going to sprout events so users can use it to load flows interactively, and I'd like to keep this sort of special casing isolated. 

As an aside, it would also be great to understand why `isatty` is not working in some places - as I understood it, it should be cross-platform on Windows. By far the best option here is to just do the right thing if we can detect it.

PR #2259, opened 2017-04-24
fix #2228

### LINKED ISSUE #2228: Sticky cookies are improperly formatted.

mhils:


kitling:
Looks better than my hack, though I'd still like to see some of the other things I mentioned addressed, if it makes sense to do so.

Not personally testing due to OS upheaval.

mhils:
I'm happy for the feedback, but I'm not sure what you are referring to? :)

mhils:
(merging for now to avoid conflicts, still happy about feedback)

kitling:
Just tested, LGTM.

(When I made the comment about "OS Upheaval", I was dealing with the fallout of a bad stick of ram. Protip: if everything is crashing and segfaulting, don't try to run a system upgrade. Was attempting to use OpenBSD for a while, before I just gave up and recovered Arch.)

PR #2265, opened 2017-04-26
Addons and addon testing

cortesi:
- Fix some loading sequence bugs affecting command-line script invocation
- Allow addons to over-ride existing options (with a warning). We need this for
reloading.
- Convert har_dump to new-style arguments, fix and re-instate its test suite.
- Covnert miscelaneous other exmples to new-style args.


mhils:
Looks good to me. Options remain after addon unload? How does that work with script reloading?

cortesi:
Yes, options are left dangling, but are over-written when addons are re-added.

I think that's OK for now. The alternatives are a) a declarative options scheme so we can delete them on unload, or b) recommending that addons remove their options in .done(). I don't think either of these pay for themselves at the moment.

PR #2271, opened 2017-04-26
[web] add connection indicator [WIP]

mhils:
This PR adds a connection indicator to mitmweb in the top right corner:

The next logical step is to automatically reconnect etc, but this is a vast improvement over the current state already.

TODO:

- [x] Tests


web/src/js/components/Header/ConnectionIndicator.jsx
```diff
@@ -0,0 +1,29 @@
+import React, { PropTypes } from "react"
```
cle1000: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.

I will change its in a different pull request...

PR #2273, opened 2017-04-27
Commands

cortesi:
Add a commands mechanism for addons. This is a base implementation that outlines the fundamental structures, which I want to merge before we diverge too far from reality. This patch:

- Adds a command manager and sketches out the addon command registration system
- Demonstrates cross-addon communication with the internal console.resolve command, which resolves flow specifications.
- Adds a `--commands` option to list all commands (analogous to `--options`)
- Demonstrates an end-to-end user facing command by adding the `save.file flowspec path` command. 
- Adds the beginning of a vim-style command interface to mitmproxy.

### DIFF
--- a/mitmproxy/command.py
+++ b/mitmproxy/command.py
@@ -2,17 +2,17 @@ import inspect
 import typing
 import shlex
 from mitmproxy.utils import typecheck
-
-
-class CommandError(Exception):
-    pass
+from mitmproxy import exceptions
+from mitmproxy import flow
 
 
 def typename(t: type) -> str:
     if t in (str, int, bool):
         return t.__name__
+    if t == typing.Sequence[flow.Flow]:
+        return "[flow]"
     else:  # pragma: no cover
-        raise NotImplementedError
+        raise NotImplementedError(t)
 
 
 def parsearg(spec: str, argtype: type) -> typing.Any:
@@ -22,7 +22,7 @@ def parsearg(spec: str, argtype: type) -> typing.Any:
     if argtype == str:
         return spec
     else:
-        raise CommandError("Unsupported argument type: %s" % argtype)
+        raise exceptions.CommandError("Unsupported argument type: %s" % argtype)
 
 
 class Command:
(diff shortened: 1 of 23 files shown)


mitmproxy/utils/typecheck.py
```diff
@@ -1,20 +1,37 @@
 import typing
 
 
-def check_type(name: str, value: typing.Any, typeinfo: typing.Any) -> None:
+def check_command_return_type(value: typing.Any, typeinfo: typing.Any) -> bool:
```
mhils: I don't get this distinction. My understanding so far was that these functions are not really be specific to options or commands. If you want to check the return type, you can just `check_type("the return value", val, type)` ? 
I'd also prefer raising over returning here so that we have a more informative error message if required.

mitmproxy/command.py
```diff
@@ -0,0 +1,91 @@
...
+
+def typename(t: type, ret: bool) -> str:
+    """
+        Translates a type to an explanatory string. Ifl ret is True, we're
+        looking at a return type, else we're looking at a parameter type.
+    """
+    if t in (str, int, bool):
+        return t.__name__
+    if t == typing.Sequence[flow.Flow]:
+        return "[flow]" if ret else "flowspec"
```
mhils: Can we drop `ret` and just make this `[Flow]` for all cases? I think that's almost equally understandable for users and would provide more consistency across the board.
(we can then also move it to utils/typecheck)

mitmproxy/utils/typecheck.py
cortesi: We support a different restricted set of types for each of command arguments, command return values and options. They may converge later, but I'm carefully expanding the types one piece at a time at the moment.

mitmproxy/command.py
cortesi: These are completely different types - it may be that we later grow internal commands that do take a list of flows as an argument. Let me think about this.

cortesi:
Taking your general comments in order (we really should collect these somewhere that's not a PR I'm just about to merge):

- The UI will just report "returned N flows".
- I think so. There will be few of these, and I'd like people to be able to try things out manually. Down the track we may have piping or some other concept of joining commands together, in which case these will be useful directly.
- There is a plan for that. The idea is that we'll expose a "view.resolve" command, and that console, mitmdump and mitmweb will each implement the command in their own addon. For "historical reasons" the command prefix for the view addon is "console", which will change soon. 
- No, the flowspec resolution is an attribute of the view addon, which is only incidentally what console is based on. I want absolutely as little as possible of this sort of logic to be in the end tools. See above.

PR #2275, opened 2017-04-27
change import of prop-types

cle1000:
I updated the import in all files...

Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.


mhils:
👍

PR #2276, opened 2017-04-27
Prominently link to examples in script docs

mhils:
I think our simple example scripts are a very useful resource for new addon developers, so I'd like to link to them very prominently. We of course should not see them as a replacement for textual documentation, but I think they are a very useful complement.

@cortesi, ok with you?

### DIFF
--- a/docs/scripting/overview.rst
+++ b/docs/scripting/overview.rst
@@ -29,6 +29,12 @@ will be added to all responses passing through the proxy:
 >>> mitmdump -s add_header.py
 
 
+Examples
+--------
+
+A collection of addons that demonstrate popular features can be found at :src:`examples/simple`.
+
+
 Using classes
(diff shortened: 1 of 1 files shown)


docs/scripting/overview.rst
```diff
@@ -29,6 +29,12 @@ will be added to all responses passing through the proxy:
 >>> mitmdump -s add_header.py
 
 
+Examples
+--------
+
+A collection of addons that demonstrate popular features can be found at :src:`examples/simple`.
```
Kriechi: Where does this link to if I'm looking at `/stable` vs. `/latest` docs?

mhils: This should always link to the correct branch, @cortesi built something in docs/conf.py (on mobile, can't link).

PR #2285, opened 2017-04-28
[web] Add coverage for js/urlState.js

MatthewShao:
It's a little bit tricky to complete the tests for this, so I want to explain a bit here.
The reducers(`reduceFlows`...) are require here to create a store object which our target function will operate on, so we could not mock the whole js file as we usually do. However, we don't care about the implementation of certain actions (`setFilter`...) and want to mock them up.

In this case, `jest.mock()` does not work. After looking up the documentation, I found a solution which uses `jest.spyOn()`. We could import all the reducers and actions first, and mock the action functions before calling the target function. Then we could use the handy `toBeCalledWith()` function to test if we are passing the right arguments to the action functions.

***
Now that we have covered almost all of the `.js` files, it's time to move on to the `.jsx` file. :rocket: 
According to the official guide of Jest, we could either use [snapshot test](http://facebook.github.io/jest/docs/tutorial-react.html#snapshot-testing) or traditional [dom test](http://facebook.github.io/jest/docs/tutorial-react.html#dom-testing) on this task. 
@mhils @cle1000 Which way do you prefer?

### DIFF
--- a/web/src/js/__tests__/urlStateSpec.js
+++ b/web/src/js/__tests__/urlStateSpec.js
@@ -1,4 +1,5 @@
 import initialize from '../urlState'
+import { updateStoreFromUrl, updateUrlFromStore } from '../urlState'
 
 import reduceFlows from '../ducks/flows'
 import reduceUI from '../ducks/ui/index'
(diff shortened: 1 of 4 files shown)


mhils:
Hmm - this looks like a prime candidate for http://arnaudbenard.com/redux-mock-store/ instead - what do you think? I think we could just use a mocked store and then check if `.getActions()` equals what we expect?

For the DOM part, I think snapshot testing is the more efficient approach in most cases, so let's start with that. 😃

MatthewShao:
Now that we mock all the actions, `updateStoreFromUrl` could not affect our store object any more. So it's better to separate the test up and test `updateUrlFromStore` independently.
And I found a minor mistake in `initialize()` function of `urlState.js`, we should subscribe to `updateUrlFromStore` before `updateStoreFromUrl`

web/src/js/urlState.js
```diff
@@ -78,6 +78,6 @@ function updateUrlFromStore(store) {
 }
 
 export default function initialize(store) {
-    updateStoreFromUrl(store)
     store.subscribe(() => updateUrlFromStore(store))
+    updateStoreFromUrl(store)
```
mhils: I don't quite understand this change. What was wrong here before?

web/src/js/__tests__/urlStateSpec.js
```diff
@@ -0,0 +1,79 @@
...
+import reduceUI from '../ducks/ui/index'
+import reduceEventLog from '../ducks/eventLog'
+import * as flowsActions from '../ducks/flows'
+
+import configureStore from 'redux-mock-store'
+
+const mockStore = configureStore()
+
+describe('updateStoreFromUrl', () => {
+    history.replaceState = jest.fn()
```
mhils: Do we need this (and the initial State) for the updateStoreFromUrl tests?

web/src/js/urlState.js
MatthewShao: default
Here's the coverage report before the change, seems that it could not hit `updateUrlFromStore.`
I am confused about why this happen, but changing the order of the code can fix this.

web/src/js/__tests__/urlStateSpec.js
MatthewShao: ` history.replaceState = jest.fn()` is not necessary, it should be deleted.
But the initial state is necessary, because `urlStare.js` only export `initialize()` function, every time we invoke `updateStoreFromUrl`, `updateStoreFromUrl` will be invoked later, so if we omit these initial states, it will throw a exception like: ` TypeError: Cannot read property 'filter' of undefined` from `updateUrlFromStore`.

mhils: Ah, okay. Missed that part. :-)
How about we just export `updateStoreFromUrl` and `updateUrlFromStore`? 😁

web/src/js/urlState.js
mhils: This problem should solve itself if we export both functions and test them independently. :)

MatthewShao: OK, let me export them for the tests.

PR #2286, opened 2017-04-28
Added LDAP Auth

charlesdhdt:
Added basic LDAP Auth:
Based on ldap3, I just added the option ldap after --proxyauth :
--proxyauth ldap[s]:url_ldap:dn:group
example :
--proxyauth ldap:ldap.forumsys.com:uid=?,dc=example,dc=com:person
So the DN can be anything (should be based on the configuration of the ldap server), the username will be queried as ? in the dn. 
Basically it will connect and bind to the server and check that the user :
- Has proper creds for logging in the server
- Is part of a group (needs to be configured on the LDAP server) that will be allowed to use the proxy

### DIFF
--- a/mitmproxy/addons/proxyauth.py
+++ b/mitmproxy/addons/proxyauth.py
@@ -114,7 +114,6 @@ class ProxyAuth:
             if conn:
                 conn.search(parts[1][1:], '('+parts[0]+username+')', attributes=['objectclass'])
                 if ctx.options.proxyauth.split(":")[3] in conn.entries[0]['objectclass']:
-                    conn.unbind()
                     return username, password
 
         return None
@@ -135,6 +134,7 @@ class ProxyAuth:
             self.nonanonymous = False
             self.singleuser = None
             self.htpasswd = None
+            self.ldapserver = None
             if ctx.options.proxyauth:
                 if ctx.options.proxyauth == "any":
                     self.nonanonymous = True
--- a/mitmproxy/addons/proxyauth.py
+++ b/mitmproxy/addons/proxyauth.py
@@ -107,10 +107,10 @@ class ProxyAuth:
             dn = ctx.options.proxyauth.split(":")[2]
             parts = dn.split("?")
             conn = ldap3.Connection(
-                    self.ldapserver,
-                    parts[0] + username + parts[1],
-                    password,
-                    auto_bind=True)
+                self.ldapserver,
+                parts[0] + username + parts[1],
+                password,
+                auto_bind=True)
             if conn:
                 conn.search(parts[1][1:], '(' + parts[0] + username + ')', attributes=['objectclass'])
                 if ctx.options.proxyauth.split(":")[3] in conn.entries[0]['objectclass']:
--- a/mitmproxy/addons/proxyauth.py
+++ b/mitmproxy/addons/proxyauth.py
@@ -115,7 +115,6 @@ class ProxyAuth:
                 conn.search(parts[1][1:], '(' + parts[0] + username + ')', attributes=['objectclass'])
                 if ctx.options.proxyauth.split(":")[3] in conn.entries[0]['objectclass']:
                     return username, password
-
         return None
 
     def authenticate(self, f: http.HTTPFlow) -> bool:
--- a/mitmproxy/addons/proxyauth.py
+++ b/mitmproxy/addons/proxyauth.py
@@ -1,5 +1,6 @@
 import binascii
 import weakref
+import ldap3
 from typing import Optional
 from typing import MutableMapping  # noqa
 from typing import Tuple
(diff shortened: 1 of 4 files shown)


Kriechi:
Unfortunately a few of the tests are failing.
I'm guessing you need to add `ldap3` to the requirements in `setup.py`.

setup.py
```diff
@@ -71,6 +71,7 @@
         "hyperframe>=5.0, <6",
         "jsbeautifier>=1.6.3, <1.7",
         "kaitaistruct>=0.7, <0.8",
+        "ldap3>=2.2.0, <2.2.1",
```
mhils: Is there any issue with 2.2.3? If not this should be `<2.3`. 😃

test/mitmproxy/addons/test_proxyauth.py
```diff
@@ -109,6 +121,24 @@ def test_check():
         )
         assert not up.check(f)
 
+        ctx.configure(
```
mhils: Am I right in the assumption that this test connects to the `ldap.forumsys.com` server? I see why you do this here, but we'd be really keen to keep our tests running even if there is no network connection. How difficult would it be to run a minimal ldap server for the tests?

mitmproxy/addons/proxyauth.py
```diff
@@ -99,6 +101,20 @@ def check(self, f: http.HTTPFlow) -> Optional[Tuple[str, str]]:
         elif self.htpasswd:
             if self.htpasswd.check_password(username, password):
                 return username, password
+        elif self.ldapserver:
+            if not username or not password:
+                return None
+            dn = ctx.options.proxyauth.split(":")[2]
+            parts = dn.split("?")
+            conn = ldap3.Connection(
```
mhils: This is establishing a network connection, right?

setup.py
charlesdhdt: Yes indeed, I just run it against 2.2.3, It works fine ! :+1:

test/mitmproxy/addons/test_proxyauth.py
charlesdhdt: Yeah, indeed I connect to an already set up LDAP server, to be honest running a small ldap locally is possible, it just requires automatic installation and configuration (and it is a bit of pain even with simple solution like openldap) which is why I chose to use an already existing one

mitmproxy/addons/proxyauth.py
charlesdhdt: yes it is

test/mitmproxy/addons/test_proxyauth.py
mhils: Ok, that sounds overly tricky. Can we just mock the ldap calls then?

charlesdhdt: yeah actually, I looked a bit at solutions to install a small local ldap server and as I was saying it is a bit of an overkill for testing purposes but ldap3 integrates a way to mock an ldap server, I should look into that direction I guess

charlesdhdt: Ok so I looked at the mock option a bit and it is possible to have a mock connection, however it needs to be hard coded in the arguments of the Connection(), I can't just emulate a server and test  what I implemented against a fake server

charlesdhdt: Ok, to sum up : 
If we are not testing the connection itself, we can use a mock ldap server to test that we do search-write-update etc... correctly
If we are testing the LDAP connection then we are really doing an integration test. In which case it's probably best to connect to a real LDAP implementation.
In our case, we are just establishing a connection to an ldap server and performing one search request, so for the connection part I don't think there is a viable solution for off line testing (apart from running a full -even minimal- local ldap server). I could still take the search part out of the function (check) and do a unit test for that part by mocking an ldap server.
What do you think ?

mhils: I think we should just [monkeypatch](https://docs.pytest.org/en/latest/monkeypatch.html) `ldap3.Connection` module so that we don't have any real connections but 100% code coverage. Does that work for you?

mitmproxy/addons/proxyauth.py
mhils: Ok. We should add a comment that this is a potential performance bottleneck, because we are doing these connections in the main thread where they block everything else. If this becomes an actual issue, we can think about adding `@concurrent` to the respective addon hooks.

test/mitmproxy/addons/test_proxyauth.py
charlesdhdt: yes works perfectly for me :)

mhils:
Awesome. Thanks again for the great PR! 🎉 🍰

PR #2299, opened 2017-05-01
Mypy Checking to pathod

iharsh234:
This PR Extend Mypy checking to `/pathod` folder , fixed some bugs from previous` example/simple `and add these mypy checking to `tox`.
REF #2194 , #2221

### DIFF
--- a/examples/simple/custom_contentview.py
+++ b/examples/simple/custom_contentview.py
@@ -3,7 +3,10 @@ This example shows how one can add a custom contentview to mitmproxy.
 The content view API is explained in the mitmproxy.contentviews module.
 """
 from mitmproxy import contentviews
-from typing import Tuple, Iterable, AnyStr, List
+import typing
+
+
+CVIEWSWAPCASE = typing.Tuple[str, typing.Iterable[typing.List[typing.Tuple[str, typing.AnyStr]]]]
 
 
 class ViewSwapCase(contentviews.View):
@@ -14,7 +17,7 @@ class ViewSwapCase(contentviews.View):
     prompt = ("swap case text", "z")
     content_types = ["text/plain"]
 
-    def __call__(self, data: bytes, **metadata) -> Tuple[str, Iterable[List[Tuple[str, AnyStr]]]]:
+    def __call__(self, data: typing.AnyStr, **metadata) -> CVIEWSWAPCASE:
         return "case-swapped text", contentviews.format_text(data.swapcase())
 
 
--- a/examples/simple/io_read_dumpfile.py
+++ b/examples/simple/io_read_dumpfile.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python
+
+# type: ignore
 #
 # Simple script showing how to read a mitmproxy dump file
 #
-
 from mitmproxy import io
 from mitmproxy.exceptions import FlowReadException
 import pprint
--- a/examples/simple/io_write_dumpfile.py
+++ b/examples/simple/io_write_dumpfile.py
@@ -8,12 +8,13 @@ to multiple files in parallel.
 import random
 import sys
 from mitmproxy import io, http
+import typing  # noqa
 
 
 class Writer:
     def __init__(self, path: str) -> None:
         if path == "-":
-            f = sys.stdout  # type: io.TextIO
+            f = sys.stdout  # type: typing.IO[typing.Any]
         else:
             f = open(path, "wb")
         self.w = io.FlowWriter(f)
--- a/pathod/language/actions.py
+++ b/pathod/language/actions.py
@@ -2,9 +2,7 @@ import abc
 import copy
 import random
 from functools import total_ordering
-
 import pyparsing as pp
-
 from . import base
 
 
@@ -52,7 +50,7 @@ class _Action(base.Token):
 
 
 class PauseAt(_Action):
-    unique_name = None
+    unique_name = None  # type: ignore
 
     def __init__(self, offset, seconds):
         _Action.__init__(self, offset)
@@ -103,7 +101,7 @@ class DisconnectAt(_Action):
 
 
 class InjectAt(_Action):
-    unique_name = None
+    unique_name = None  # type: ignore
 
     def __init__(self, offset, value):
         _Action.__init__(self, offset)
--- a/pathod/language/base.py
+++ b/pathod/language/base.py
@@ -3,10 +3,9 @@ import os
 import abc
 import functools
 import pyparsing as pp
-
 from mitmproxy.utils import strutils
 from mitmproxy.utils import human
-
+import typing  # noqa
 from . import generators, exceptions
 
 
@@ -84,7 +83,7 @@ class Token:
         return None
 
     @property
-    def unique_name(self):
+    def unique_name(self) -> typing.Optional[str]:
         """
             Controls uniqueness constraints for tokens. No two tokens with the
             same name will be allowed. If no uniquness should be applied, this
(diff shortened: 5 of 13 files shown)


pathod/language/websockets.py
```diff
@@ -239,8 +239,14 @@ class NestedFrame(base.NestedMessage):
     preamble = "f"
     nest_type = WebsocketFrame
 
-COMP = Tuple[ Type[OpCode], Type[Length], Type[Fin], Type[RSV1], Type[RSV2], Type[RSV3], Type[Mask], Type[actions.PauseAt], Type[actions.DisconnectAt], Type[actions.InjectAt], Type[KeyNone], Type[Key], Type[Times], Type[Body], Type[RawBody]]
+
+COMP = Tuple[
```
iharsh234: @Kriechi Any better idea how to fix this , or ignoring type is the only choice?

cortesi:
Hi there. This looks good to me. One point about code style - as a general rule, we import modules, and not individual items from modules. So we prefer:

```python
import typing

x = []  # type: typing.Sequence
```

to

```python
from typing import Sequence

x = []  # type: Sequence
```

This is a bit more verbose, but is a lot clearer, doesn't require you to constantly adjust the imports as you use more or less of the module, and really pays dividends when refactoring or doing searches across the codebase.

The particular bit of code you're working on here doesn't stick to the import rules very consistently - feel free to fix the usage if you're working on the code anyway.

iharsh234:
@cortesi done :smile:

pathod/language/actions.py
```diff
@@ -52,7 +50,7 @@ def intermediate(self, settings):  # pragma: no cover
 
 
 class PauseAt(_Action):
-    unique_name = None
+    unique_name = None  # type: ignore
```
cortesi: This can be `typing.Optional[str]`. Same for all the similar occurrences.

iharsh234: I have done that before , but was getting some errors like
`error: Definition of "unique_name" in base class "_HeaderMixin" is incompatible with definition in base class "Token"
`
Then from [here](https://github.com/python/mypy/issues/1237) I get to know that either we should refactor code or go for `# type: ignore`

cortesi: Ah, OK then. Squash the commits and let's merge this puppy.

examples/simple/custom_contentview.py
```diff
@@ -14,7 +17,7 @@ class ViewSwapCase(contentviews.View):
     prompt = ("swap case text", "z")
     content_types = ["text/plain"]
 
-    def __call__(self, data: bytes, **metadata) -> Tuple[str, Iterable[List[Tuple[str, AnyStr]]]]:
+    def __call__(self, data: typing.AnyStr, **metadata) -> CVIEWSWAPCASE:
```
mhils: Why `AnyStr` here? This should be `bytes`.

examples/simple/io_read_dumpfile.py
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env python
+
+# type: ignore
```
mhils: What's the motivation for this?

examples/simple/io_write_dumpfile.py
```diff
@@ -8,12 +8,13 @@
...
 import sys
 from mitmproxy import io, http
+import typing  # noqa
 
 
 class Writer:
     def __init__(self, path: str) -> None:
         if path == "-":
-            f = sys.stdout  # type: io.TextIO
+            f = sys.stdout  # type: typing.IO[typing.Any]
```
mhils: The FlowWriter only writes bytes, so this should really be `typing.IO[bytes]`. If that raises errors, that's a strong indicator that the code here does not work as intended. I'd be fine with removing the `if path == "-":` special case altogether.

pathod/language/actions.py
mhils: That's not really what the linked issue says. These attributes should have the same type as the base class.

pathod/language/base.py
```diff
@@ -376,7 +375,7 @@ def freeze(self, settings):
 
 
 class Integer(_Component):
-    bounds = (None, None)
+    bounds = (None, None)  # type: typing.Tuple[typing.Union[int, None], typing.Union[int , None]]
```
mhils: This should be `Optional[int]`, not `typing.Union[int, None]`.

pathod/language/base.py
```diff
@@ -546,7 +545,7 @@ class NestedMessage(Token):
         A nested message, as an escaped string with a preamble.
     """
     preamble = ""
-    nest_type = None
+    nest_type = None  # type: ignore
```
mhils: We shouldn't ignore this here either.

pathod/language/websockets.py
```diff
@@ -239,7 +240,14 @@ class NestedFrame(base.NestedMessage):
     nest_type = WebsocketFrame
 
 
+COMP = typing.Tuple[
+    typing.Type[OpCode], typing.Type[Length], typing.Type[Fin], typing.Type[RSV1], typing.Type[RSV2], typing.Type[RSV3], typing.Type[Mask],
+    typing.Type[actions.PauseAt], typing.Type[actions.DisconnectAt], typing.Type[actions.InjectAt], typing.Type[KeyNone], typing.Type[Key],
+    typing.Type[Times], typing.Type[Body], typing.Type[RawBody]
+]
```
mhils: That looks wrong. First, this doesn't really match the semantics of [`typing.Tuple`](https://docs.python.org/3/library/typing.html#typing.Tuple), second, I think all those are `_Component`s?

pathod/pathod.py
```diff
@@ -71,7 +69,7 @@ def get_cert(self, name):
 
 class PathodHandler(tcp.BaseHandler):
     wbufsize = 0
-    sni = None
+    sni = None  # type: typing.Union[str, None, bool]
```
mhils: Why would this be a bool? Looks like `Optional[str]` to me?

examples/simple/custom_contentview.py
```diff
@@ -3,7 +3,10 @@
 The content view API is explained in the mitmproxy.contentviews module.
 """
 from mitmproxy import contentviews
-from typing import Tuple, Iterable, AnyStr, List
+import typing
+
+
+CVIEWSWAPCASE = typing.Tuple[str, typing.Iterable[typing.List[typing.Tuple[str, typing.AnyStr]]]]
```
mhils: This type annotations should be done here: https://github.com/mitmproxy/mitmproxy/blob/9d2350b6709532bf185c6e0f56342b7b780092c7/mitmproxy/contentviews/base.py#L17 and then re-used here. Please prefix types with `T`.

examples/simple/io_read_dumpfile.py
iharsh234: It was previous fix , but after adding `--follow-imports=skip` , it should not be needed now. thanks:)

pathod/language/base.py
iharsh234: Don't both of them same?

pathod/language/websockets.py
iharsh234: I wasn't sure about it , and asked but didn't received any reply.

pathod/language/actions.py
iharsh234: They should have , but mypy always says they are incompatible , don't know why

pathod/pathod.py
iharsh234: > sni: ``str`` for a custom SNI value,
                ``None`` for the client SNI value,
                ``False`` if no SNI value should be sent.

examples/simple/custom_contentview.py
iharsh234: indeed it should be, it working with mypy test now.

examples/simple/io_write_dumpfile.py
iharsh234:
```
def __init__(self, path: str) -> None:
        f = open(path, "wb")  # type: typing.IO[bytes]
        self.w = io.FlowWriter(f)
```
this worked as expected :smile:

PR #2312, opened 2017-05-05
Note to enable `Certificate Trust Settings` in iOS 10.3 or up

zhigang1992:
Even after installing certs from `mitm.it`, it's not trusted by default, user need to enable this on their own.

http://stackoverflow.com/a/43303169/1433273

simulator screen shot may 5 2017 5 13 34 pm


Kriechi:
Thanks! 🎉 👍

PR #2316, opened 2017-05-05

mhils:


Kriechi:
Looks like we got a mypy error:
`mitmproxy/addons/view.py:73: error: Incompatible return value type (got "Union[Any, int]", expected "datetime")`

mhils:
And it's a valid one! `timestamp_start` is an int, I fixed the type annotation. :-)

PR #2319, opened 2017-05-06
Fix examples/simple/custom_contentview.py to work

tkiapril:
I was looking at the examples to use mitmproxy inside a small project,
but this example wasn't quite working:

While trying to run this script on both Linux/Mac with python 3.6.1, mitmproxy
installed with venv and pip, this script had few problems and would not run.

- load(l) was not being called, trying to ctx.log.error() inside load(l) did
  not show anything on the Event Log.

- shortcut key in prompt had to be changed into something inside the title
  "swap case text" -- otherwise it would result in IndexError.

- "case-swapped text" inside __call__ is not defined in palette -- thus
  changed into generic "text".

### DIFF
--- a/examples/simple/custom_contentview.py
+++ b/examples/simple/custom_contentview.py
@@ -14,19 +14,14 @@ class ViewSwapCase(contentviews.View):
 
     # We don't have a good solution for the keyboard shortcut yet -
     # you manually need to find a free letter. Contributions welcome :)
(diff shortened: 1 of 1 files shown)


tkiapril:
Nevermind, I'm closing this because I noticed that this example works fine on master branch. Sorry for the bother ;)

PR #2322, opened 2017-05-07
[WIP][web] Coverage improve

MatthewShao:
Here's the brief road map about adding tests for our JavaScript code.

- [x] Clear up the code: remove `jest.unmock()`
- [x] Add full coverage to the files in the path we have touched, for example `js/ducks`, `js/ducks/ui`
    - [x] `js/ducks/index.js`
    - [x] `js/ducks/ui/index.js`
    - [x] `js/ducks/ui/keyboard.js`

Next step: Add tests for components in the `.jsx` files.

The reason why we need to remove `jest.unmock` has been discussed [here](https://github.com/mitmproxy/mitmproxy/pull/2212#issuecomment-289211106), I think we should keep the automocking disabled and manually mock the modules if necessary.

### DIFF
--- a/web/src/js/__tests__/ducks/ui/keyboardSpec.js
+++ b/web/src/js/__tests__/ducks/ui/keyboardSpec.js
@@ -7,151 +7,142 @@ import reduceUI from '../../../ducks/ui/index'
 import * as flowsActions from '../../../ducks/flows'
 import * as UIActions from '../../../ducks/ui/flow'
 import configureStore from 'redux-mock-store'
-import { createStore } from '../tutils'
+import thunk from 'redux-thunk'
 import { fetchApi } from '../../../utils'
 
-const mockStore = configureStore()
+const mockStore = configureStore([ thunk ])
 console.debug = jest.fn()
 
 describe('onKeyDown', () => {
     let flows = undefined
     for( let i=1; i <= 12; i++ ) {
-        flows = reduceFlows(flows, {type: flowsActions.ADD, data: {id: i}, cmd: 'add'})
+        flows = reduceFlows(flows, {type: flowsActions.ADD, data: {id: i, request: true, response: true}, cmd: 'add'})
     }
     let store = mockStore({ flows, ui: reduceUI(undefined, {}) })
     let createKeyEvent = (keyCode, shiftKey = undefined, ctrlKey = undefined) => {
-            return { keyCode, shiftKey, ctrlKey, preventDefault: jest.fn() }
+            return onKeyDown({ keyCode, shiftKey, ctrlKey, preventDefault: jest.fn() })
     }
 
     it('should handle cursor up', () => {
         store.getState().flows = reduceFlows(flows, flowsActions.select(2))
-        onKeyDown(createKeyEvent(Key.K))(store.dispatch, store.getState)
-        expect(store.getActions().pop()).toEqual({ flowIds: [1], type: flowsActions.SELECT })
+        store.dispatch(createKeyEvent(Key.K))
+        expect(store.getActions()).toEqual([{ flowIds: [1], type: flowsActions.SELECT }])
 
-        onKeyDown(createKeyEvent(Key.UP))(store.dispatch, store.getState)
+        store.dispatch(createKeyEvent(Key.UP))
         expect(store.getActions().pop()).toEqual({ flowIds: [1], type: flowsActions.SELECT })
     })
 
     it('should handle cursor down', () => {
-        onKeyDown(createKeyEvent(Key.J))(store.dispatch, store.getState)
+        store.dispatch(createKeyEvent(Key.J))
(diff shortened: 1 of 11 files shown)


mhils:
Sounds good! You can merge this right away if you want, but also feel free to follow through with the other points in this PR if you prefer that! :smiley:

MatthewShao:
Let me explain these tests a bit. We bound many actions to the keyboard, which have different mechanism. It took me more time than expected to figure out a solution to cover them up...
1. For the actions which change the state by invoking reducer, (for example cursor movement), we can create a mock store by using `redux-mock-store`, and assert the actions pass to the reducer.
2. For the actions which simply invoke `fetchApi`, we have to introduce the real store object because the mocked one does not actually dispatch things to the reducer. Then we assert the arguments pass to the `fetchApi`. 

If these new tests are good enough, I think it's time to move on to the components parts. I hope this PR can be merged before we touch those paths, because we are going to experience a dramatic drop on the coverage rate. Hold on for the drop! :smile:

web/src/js/__tests__/ducks/ui/keyboardSpec.js
```diff
@@ -0,0 +1,158 @@
...
+    }
+    let store = mockStore({ flows, ui: reduceUI(undefined, {}) })
+    let createKeyEvent = (keyCode, shiftKey = undefined, ctrlKey = undefined) => {
+            return { keyCode, shiftKey, ctrlKey, preventDefault: jest.fn() }
+    }
+
+    it('should handle cursor up', () => {
+        store.getState().flows = reduceFlows(flows, flowsActions.select(2))
+        onKeyDown(createKeyEvent(Key.K))(store.dispatch, store.getState)
+        expect(store.getActions().pop()).toEqual({ flowIds: [1], type: flowsActions.SELECT })
```
mhils: Nit-pick:
```javascript
expect(store.getActions()).toEqual([{expectedAction}])
```
would also check that only this one action has been dispatched.

web/src/js/__tests__/ducks/ui/keyboardSpec.js
```diff
@@ -0,0 +1,158 @@
...
+
+        onKeyDown(createKeyEvent(Key.UP))(store.dispatch, store.getState)
+        expect(store.getActions().pop()).toEqual({ flowIds: [1], type: flowsActions.SELECT })
+    })
+
+    it('should handle cursor down', () => {
+        onKeyDown(createKeyEvent(Key.J))(store.dispatch, store.getState)
```
mhils: If the redux-thunk part we just discussed on Slack works, this can just be
```javascript
store.dispatch(onKeyDown(createKeyEvent(Key.J)))
```

web/src/js/__tests__/ducks/ui/keyboardSpec.js
```diff
@@ -0,0 +1,158 @@
...
+        onKeyDown(createKeyEvent(Key.LEFT))(store.dispatch, store.getState)
+        expect(store.getActions().pop()).toEqual({ tab: 'details', type: UIActions.SET_TAB })
+    })
+
+    it('should handle switch to right tab', () => {
+        onKeyDown(createKeyEvent(Key.TAB))(store.dispatch, store.getState)
+        expect(store.getActions().pop()).toEqual({ tab: 'details', type: UIActions.SET_TAB })
+    })
+
+    it('should handle switch to left tab', () => {
```
mhils: Duplicate from line 74-82

web/src/js/__tests__/ducks/ui/keyboardSpec.js
```diff
@@ -0,0 +1,158 @@
...
+import { createStore } from '../tutils'
+import { fetchApi } from '../../../utils'
+
+const mockStore = configureStore()
+console.debug = jest.fn()
+
+describe('onKeyDown', () => {
+    let flows = undefined
+    for( let i=1; i <= 12; i++ ) {
+        flows = reduceFlows(flows, {type: flowsActions.ADD, data: {id: i}, cmd: 'add'})
```
mhils: We can use `{id: i, request: true, response: true}` for the flow to have better tests for left/right keys (https://github.com/mitmproxy/mitmproxy/blob/master/web/src/js/ducks/ui/keyboard.js#L64)

web/src/js/__tests__/ducks/ui/keyboardSpec.js
```diff
@@ -0,0 +1,158 @@
...
+    }
+    let store = mockStore({ flows, ui: reduceUI(undefined, {}) })
+    let createKeyEvent = (keyCode, shiftKey = undefined, ctrlKey = undefined) => {
+            return { keyCode, shiftKey, ctrlKey, preventDefault: jest.fn() }
```
mhils: We could also return `onKeyDown({ keyCode, shiftKey, ctrlKey, preventDefault: jest.fn() })` here, that'd make the tests less verbose.

PR #2327, opened 2017-05-11
[WIP][web] Add tests for UI components

MatthewShao:
- [x] comon  
- [x] helpers
- [ ] ValueEditor
- [ ] Header
- [ ] FlowView
- [ ] FlowTable
- [ ] EventLog
- [ ] ContentView
- [ ] ...

### DIFF
--- a/web/src/js/components/common/FileChooser.jsx
+++ b/web/src/js/components/common/FileChooser.jsx
@@ -21,7 +21,7 @@ export default function FileChooser({ icon, text, className, title, onOpenFile }
                 ref={ref => fileInput = ref}
                 className="hidden"
                 type="file"
-                onChange={e => { e.preventDefault(); if(e.target.files.length > 0) onOpenFile(e.target.files[0]); fileInput = "";}}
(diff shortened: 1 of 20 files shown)


mhils:
Looks good! I think the most interesting part will be in components that take flows as props. It probably makes sense to have a central test flow object that gets reused across a set of components for that.

web/src/js/components/common/FileChooser.jsx
```diff
@@ -21,7 +21,7 @@ export default function FileChooser({ icon, text, className, title, onOpenFile }
                 ref={ref => fileInput = ref}
                 className="hidden"
                 type="file"
-                onChange={e => { e.preventDefault(); if(e.target.files.length > 0) onOpenFile(e.target.files[0]); fileInput = "";}}
```
cle1000: `fileInput = ""` should be changed to `fileInput.value = ""`
this call resets the fileInput. without this extra call it is not possible to open the same file twice...

fileInput.value = ""

MatthewShao: Oh, that's right. This statement confuses me a lot... Thanks! @cle1000

PR #2333, opened 2017-05-12
Fixed LDAP Auth

charlesdhdt:
Fixed LDAP Auth. I realized what I did last week is clearly not good enough and is too specific so I did everything again with a better understanding of ldap


setup.py
```diff
@@ -96,6 +96,7 @@
         'dev': [
             "Flask>=0.10.1, <0.13",
             "flake8>=3.2.1, <3.4",
+            "mock>=2.0.0, <3.0",
```
Kriechi: Please use `from unittest import mock` instead of this package.

test/mitmproxy/addons/test_proxyauth.py
```diff
@@ -1,8 +1,8 @@
 import binascii
-import ldap3
 
 import pytest
 
+import mock
```
Kriechi: `from unittest import mock`

charlesdhdt:
I think it is ready to be pushed, what do you think ?

Kriechi:
Thanks! 🎉 👍

PR #2336, opened 2017-05-14
Increase net.tcp.py coverage

ujjwal96:
This PR increases net.tcp.py coverage to 100% making coverage of mitmproxy/net/ to 100%

### DIFF
--- a/mitmproxy/net/tcp.py
+++ b/mitmproxy/net/tcp.py
@@ -511,7 +511,7 @@ class _Connection:
         if log_ssl_key:
             context.set_info_callback(log_ssl_key)
 
-        if HAS_ALPN:
+        if HAS_ALPN:  # pragma: openssl-old no cover
             if alpn_protos is not None:
                 # advertise application layer protocols
                 context.set_alpn_protos(alpn_protos)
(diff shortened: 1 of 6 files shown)


mitmproxy/net/tcp.py
```diff
@@ -849,7 +849,7 @@ def __init__(self, address):
         self.__is_shut_down = threading.Event()
         self.__shutdown_request = False
 
-        if self.address == 'localhost':
+        if self.address[0] == 'localhost':
```
mhils: Good catch 👍

test/full_coverage_plugin.py
```diff
@@ -59,6 +60,12 @@ def pytest_runtestloop(session):
     if os.name == 'nt':
         cov.exclude('pragma: windows no cover')
 
+    if sys.platform == 'darwin':
+        cov.exclude('pragma: osx no cover')
+
+    if 'OPENSSL_OLD' in os.environ:
+        cov.exclude('pragma: alpn no cover')
```
Kriechi: These patterns use a name to indicate *when* the code is not being covered.
Following this idea, `alpn no cover` would mean, code is not covered when ALPN is present.
I think a better name would be `pragma: missing-alpn no cover` or `pragma: openssl-old no cover`.

mitmproxy/net/tcp.py
ujjwal96: Now I see the importance of test cases and code coverage :smile:

Kriechi:
Almost ready - please solve the conflict first.

PR #2337, opened 2017-05-16
[WIP][web] Add tests for UI components

MatthewShao:
- [x] ValueEditor 
- [ ] FlowTable

PR #2342, opened 2017-05-19
[web] Add  _flow.js for component tests.

MatthewShao:
Suggested by @mhils [here](https://github.com/mitmproxy/mitmproxy/pull/2337#discussion_r117033821), we add a python test to generate `js/__tests__/ducks/_flow.js` for the components tests. In this way, each time when we run the python test, it will produce a generator function for our js tests.

### DIFF
--- a/test/mitmproxy/tools/web/test_app.py
+++ b/test/mitmproxy/tools/web/test_app.py
@@ -1,5 +1,6 @@
 import json as _json
 from unittest import mock
+import os
 
 import tornado.testing
 from tornado import httpclient
@@ -11,7 +12,6 @@ from mitmproxy import options
 from mitmproxy.test import tflow
 from mitmproxy.tools.web import app
 from mitmproxy.tools.web import master as webmaster
-
 
 def json(resp: httpclient.HTTPResponse):
     return _json.loads(resp.body.decode())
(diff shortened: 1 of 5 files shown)


test/mitmproxy/tools/web/test_app.py
```diff
@@ -275,3 +276,13 @@ def test_websocket(self):
         # trigger on_close by opening a second connection.
         ws_client2 = yield websocket.websocket_connect(ws_url)
         ws_client2.close()
+
+    def test_generate_tflow_js(self):
+        tflow_json = _json.dumps(
+            app.flow_to_json(tflow.tflow(resp=True, err=True)), indent=4, sort_keys=True
+        )
+        web_root = os.path.join(os.getcwd(), 'web')
```
mhils: `cwd` is a bit dangerous as this may change depending on where you execute your tests. Use [this](https://github.com/mitmproxy/mitmproxy/blob/af0ad494348d0f2d773232b5913d5c86787e39f5/setup.py#L10) and construct the path relative from here. :)

web/src/js/__tests__/components/FlowTable/FlowColumnsSpec.js
```diff
@@ -1,101 +1,101 @@
 import React from 'react'
 import renderer from 'react-test-renderer'
 import * as Columns from '../../../components/FlowTable/FlowColumns'
-import { TFlow } from '../../ducks/tutils'
+import _tflow from '../../ducks/_tflow'
```
mhils: Do you want to have `tflow` as a "public" module or do we want to integrate it in `tutils`? I was thinking of the following in tutils:
```javascript
export {default as TFlow} from './_tflow'
```
Both really works, go with what you prefer. 😃

MatthewShao: This is a really good point! I would like to integrate all the test mock object into `tutils.js`

web/src/js/__tests__/ducks/_tflow.js
```diff
@@ -8,7 +8,7 @@ export default function(){
         "alpn_proto_negotiated": "http/1.1",
         "cipher_name": "cipher",
         "clientcert": null,
-        "id": "75bfd3cd-a084-4d84-a063-b0804dc91342",
+        "id": "4a18d1a0-50a1-48dd-9aa6-d45d74282939",
```
mhils: OK, this commit is helpful in unintended ways. We need to set the ids and timestamps to constant values before exporting, otherwise this file will always be changed.

MatthewShao: I'll check if there is any other value would be changed from time to time, and give a constant value in the Python test.

mhils:
:+1: :+1: :+1:

PR #2346, opened 2017-05-20
Tests for wbxml

ujjwal96:
Added tests for wbxml


Kriechi:
Please at mention the URL / git-commit where you got the xml files from.
Either as first-line comment in the file itself, or the commit message (after the first line).
It is MIT license, so it should be ok.

PR #2348, opened 2017-05-23
[WIP][web] Add tests for js/components/Header

MatthewShao:
- [x] FilterDocs.jsx
- [x] ConnectionIndicator.jsx
- [x] FIleMenu.jsx
- [x] FIlterInput.jsx
- [x] FlowMenu.jsx
- [x] MenuToggle.jsx
- [x] OptionMenu.jsx
- [x] MainMenu.jsx

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -25,7 +25,6 @@
   "dependencies": {
     "bootstrap": "^3.3.7",
     "classnames": "^2.2.5",
-    "jest-fetch-mock": "^1.1.1",
     "lodash": "^4.17.4",
     "prop-types": "^15.5.0",
     "react": "^15.4.2",
--- a/web/package.json
+++ b/web/package.json
@@ -25,6 +25,7 @@
   "dependencies": {
     "bootstrap": "^3.3.7",
     "classnames": "^2.2.5",
+    "jest-fetch-mock": "^1.1.1",
     "lodash": "^4.17.4",
     "prop-types": "^15.5.0",
     "react": "^15.4.2",
--- a/web/src/js/__tests__/components/Header/FilterDocsSpec.js
+++ b/web/src/js/__tests__/components/Header/FilterDocsSpec.js
@@ -1,0 +1,19 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+import FilterDocs from '../../../components/Header/FilterDocs'
+import mockFetch from 'jest-fetch-mock'
+
+global.fetch = mockFetch
+
+describe('FilterDocs Component', () => {
+
+    it('should render correctly', () => {
+        // fetch successes
+        fetch.mockResponseOnce(JSON.stringify({commands: [['cmd1', 'foo'], ['cmd2', 'bar']]}), {status: 200})
+        let filterDocs = renderer.create(<FilterDocs/>),
+            tree = filterDocs.toJSON()
+        // [TODO] doc in render() could not be set correctly.
+        console.log(tree)
+    })
+
+})
--- a/web/src/js/__tests__/components/Header/FilterDocsSpec.js
+++ b/web/src/js/__tests__/components/Header/FilterDocsSpec.js
@@ -4,10 +4,10 @@ import FilterDocs from '../../../components/Header/FilterDocs'
 
 describe('FilterDocs Component', () => {
     let mockResponse = { json:
-            jest.fn(() => { return { commands: [['cmd1', 'foo'], ['cmd2', 'bar']]}})
+            () => { return { commands: [['cmd1', 'foo'], ['cmd2', 'bar']]}}
         },
         promise = Promise.resolve(mockResponse)
-    global.fetch = jest.fn(r => { return promise })
+    global.fetch = r => { return promise }
 
     let filterDocs = renderer.create(<FilterDocs/>),
         tree = filterDocs.toJSON()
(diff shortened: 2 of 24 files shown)


MatthewShao:
@mhils  @cle1000 , I met some trouble here. To make the test go into the `FilterDocs.xhr` logic, I introduce a dependency `jest-fetch-mock`. But we still have two line missed.
1. https://github.com/mitmproxy/mitmproxy/blob/master/web/src/js/components/Header/FilterDocs.jsx#L20, even with `jest-fetch-mock`, I could not find a way to mock the connection error, some mock response with error status code could not trigger the `catch` method.
2. https://github.com/mitmproxy/mitmproxy/blob/master/web/src/js/components/Header/FilterDocs.jsx#L40 , it turns out that the render result is `<i className="fa fa-spinner fa-spin"></i>`, instead of the docs table.

It's kind of tricky to cover up the lines here, and I notice that there's a `TODO` saying: *move to redux* in this file, so I wonder should we do this right now to add up the testability, if so, please give me some guidance.

web/package.json
```diff
@@ -25,6 +25,7 @@
   "dependencies": {
     "bootstrap": "^3.3.7",
     "classnames": "^2.2.5",
+    "jest-fetch-mock": "^1.1.1",
```
mhils: This needs to go, otherwise LGTM! :)

web/src/js/components/Header/MainMenu.jsx
```diff
@@ -17,14 +17,18 @@ export default function MainMenu() {
...
+
 const InterceptInput = connect(
     state => ({
         value: state.settings.intercept || '',
         placeholder: 'Intercept',
         type: 'pause',
         color: 'hsl(208, 56%, 53%)'
     }),
-    { onChange: intercept => updateSettings({ intercept }) }
+    { onChange: setIntercept }
```
mhils: Why this?

web/src/js/__tests__/ducks/tutils.js
```diff
@@ -8,4 +14,58 @@ export function createStore(parts) {
     )
 }
 
-export { default as TFlow } from './_tflow'
+export { TFlow }
+
+export function TStore(){
+    let tflow = new TFlow()
+    return mockStore({
+        ui: {
```
mhils: Any particular reason why we don't use the respective reducers here? This currently is a good candidate where we could get out of sync.

web/src/js/__tests__/components/Header/FilterDocsSpec.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+import FilterDocs from '../../../components/Header/FilterDocs'
+
+describe('FilterDocs Component', () => {
+    let mockResponse = { json:
+            jest.fn(() => { return { commands: [['cmd1', 'foo'], ['cmd2', 'bar']]}})
+        },
+        promise = Promise.resolve(mockResponse)
+    global.fetch = jest.fn(r => { return promise })
```
mhils: Do we need that `jest.fn` here (and above)?

web/src/js/__tests__/components/Header/FilterInputSpec.js
```diff
@@ -0,0 +1,93 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+import FilterInput from '../../../components/Header/FilterInput'
+import FilterDocs from '../../../components/Header/FilterDocs'
+import TestUtil from 'react-dom/test-utils'
+import ReactDOM from 'react-dom'
+import { Key } from '../../../utils'
+
+describe('FilterDocs Component', () => {
```
mhils: FilterInput

web/src/js/components/Header/MainMenu.jsx
MatthewShao: Here we need to extract the anonymous function, otherwise this line could not be covered.

web/src/js/__tests__/components/Header/FilterDocsSpec.js
MatthewShao: Yeah, this should be eliminated.

web/src/js/__tests__/ducks/tutils.js
MatthewShao: If we use the reducers, it will look like this:
```javascript
export function TStore(){
    let tflow = new TFlow(),
        uiState = null,
        setttingState = null
    uiState = reduceUI(undefined, UIActions.setContentViewDescription('foo'))
    uiState = reduceUI(uiState, UIActions.setContent(['foo', 'bar']))
    uiState = reduceUI(uiState, UIActions.startEdit(tflow))
    setttingState = reduceSetting(setttingState, {type: SettingActions.UPDATE, data: { contentViews: ['Auto', 'Raw', 'Text']}})
    setttingState = reduceSetting(setttingState, {type: SettingActions.UPDATE, data: { anticache: true }})
    setttingState = reduceSetting(setttingState, {type: SettingActions.UPDATE, data: { anticomp: false }})
    return mockStore({
        ui: uiState,
        settings: settingState
        })
}
```
Basically, we need to pass a bunch of actions to the reducers to get the state object, it's a little bit redundant. Would it be more straightforward and clear to define the state object statically? In that way we could see the structure of the store object and modify some value easily.

mhils: Ok, your argument on seeing the structure is quite convincing. Let's continue as you proposed, if we see that this causes consistency issues in the long run, we could still change it. :)

PR #2356, opened 2017-05-26
bump cryptography

mhils:


release/setup.py
```diff
@@ -6,9 +6,9 @@
     py_modules=["rtool"],
     install_requires=[
         "click>=6.2, <7.0",
-        "twine>=1.6.5, <1.9",
+        "twine>=1.6.5, <1.10",
         "pysftp==0.2.8",
-        "cryptography>=1.6, <1.7",
+        "cryptography>=1.6, <1.9",
```
Kriechi: Hmm, didn't we just find out that 1.6 *does NOT* work?
We should bump the minimum as well.

PR #2359, opened 2017-05-29
[web] Add tests for components/EventLog & components/FlowView

MatthewShao:
- [x] EventList.jsx
- [x] Details.jsx
- [x] Header.jsx
- [x] Message.jsx
- [x] Nav.jsx
- [x] ToggleEdit.jsx

### DIFF
--- a/web/src/js/__tests__/components/FlowView/MessagesSpec.js
+++ b/web/src/js/__tests__/components/FlowView/MessagesSpec.js
@@ -1,0 +1,150 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+import TestUtils from 'react-dom/test-utils'
+import { Request, Response, ErrorView } from '../../../components/FlowView/Messages'
+import { Provider } from 'react-redux'
+import { TFlow, TStore } from '../../ducks/tutils'
+import { updateEdit } from '../../../ducks/ui/flow'
+import { parseUrl } from '../../../flow/utils'
+import ContentView from '../../../components/ContentView'
+import ContentViewOptions from '../../../components/ContentView/ContentViewOptions'
+import Headers from '../../../components/FlowView/Headers'
+import ValueEditor from '../../../components/ValueEditor/ValueEditor'
+import mockXMLHttpRequest from 'mock-xmlhttprequest'
+
+global.XMLHttpRequest = mockXMLHttpRequest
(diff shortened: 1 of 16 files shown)


web/src/js/components/FlowView/Nav.jsx
```diff
@@ -2,14 +2,15 @@ import React  from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 import classnames from 'classnames'
+import _ from 'lodash'
```
mhils: Any idea why this worked before?

web/src/js/__tests__/components/FlowView/MessagesSpec.js
```diff
@@ -0,0 +1,150 @@
...
+import mockXMLHttpRequest from 'mock-xmlhttprequest'
+
+global.XMLHttpRequest = mockXMLHttpRequest
+global.fetch = jest.fn()
+
+let tflow = new TFlow(),
+    store = TStore()
+store.getState().ui.flow.modifiedFlow = false
+
+mockXMLHttpRequest.onSend = xhr => {
```
mhils: What do we need this for? :)

web/src/js/components/FlowView/Nav.jsx
MatthewShao: Without this statement, it works fine in the the browser, but it will fail the tests. Throwing `ReferenceError: _ is not defined`. I guess it is because `lodash` has been import by other part of our js code, but here, when we doing unit tests, it just fails to find `_`.

web/src/js/__tests__/components/FlowView/MessagesSpec.js
MatthewShao: The `ContentView` component here will invoke `updateContent` in `ContentLoad`, that function use XMLHttpRequest. So I mocked XMLHttpRequest here to feed `updateContent`.
But when I reconsider this just now, I found that we could actually mock the `ContentView` to bypass the error. I'll update the new version.

MatthewShao: But shall we keep `mock-xmlhttprequest` in the dependencies? We need it in the `ContentView` after all.

mhils:
\o/

PR #2365, opened 2017-06-01
[WIP][web] Add tests for components/ContentView

MatthewShao:
- [x] CodeEditor.jsx
- [ ] ContentLoader.jsx
- [x] ContentViewOptions.jsx
- [ ] ContentView.jsx
- [ ] DownloadContentButton.jsx
- [ ] MetaView.jsx
- [ ] ShowFullContentButton.jsx
- [ ] UploadContentButton.jsx
- [ ] ViewSelector.jsx

### DIFF
--- a/web/src/js/__tests__/components/ContentView/__snapshots__/CodeEditorSpec.js.snap
+++ b/web/src/js/__tests__/components/ContentView/__snapshots__/CodeEditorSpec.js.snap
@@ -1,0 +1,8 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`CodeEditor Component should render correctly 1`] = `
+<div
+  className="codeeditor"
+  onKeyDown={[Function]}
+/>
+`;
--- a/web/src/js/components/ContentView/ContentViewOptions.jsx
+++ b/web/src/js/components/ContentView/ContentViewOptions.jsx
@@ -4,6 +4,7 @@ import { connect } from 'react-redux'
 import ViewSelector from './ViewSelector'
 import UploadContentButton from './UploadContentButton'
 import DownloadContentButton from './DownloadContentButton'
+import { uploadContent } from '../../ducks/flows'
 
 ContentViewOptions.propTypes = {
     flow: PropTypes.object.isRequired,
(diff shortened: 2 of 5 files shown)


web/src/js/components/ContentView/ContentViewOptions.jsx
```diff
@@ -28,5 +29,8 @@ export default connect(
     state => ({
         contentViewDescription: state.ui.flow.viewDescription,
         readonly: !state.ui.flow.modifiedFlow,
-    })
+    }),
+    {
+        uploadContent: uploadContent
```
mhils: Are you sure about this change? I think this gets passed in here: https://github.com/mitmproxy/mitmproxy/blob/5f91b7a45d0b983d3e0f3fff1be8af93736d35c6/web/src/js/components/FlowView/Messages.jsx#L114

web/src/js/__tests__/components/ContentView/__snapshots__/CodeEditorSpec.js.snap
```diff
@@ -0,0 +1,8 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`CodeEditor Component should render correctly 1`] = `
+<div
```
mhils: It looks like all CodeMirror is missing entirely here?

MatthewShao: `CodeMirror` itself is too complex, it use `React.findDOMNode` and `ref` tricks to implement its logic. Unfortunately, `react-test-render` has poor support for these two statements. It will throw some strange exceptions.
The only way that came up to me is mock the whole `CodeMirror`, as a dependency, it should have been fully tested, right? Or we could turn to `TestUtils.renderIntoDocument` to do DOM test, but we need to mock many other functions like `document.createRange` for `CodeMirror`, basically doing another render test for `CodeMirror` component.
Do you have any idea about how to deal with this `CodeMirror` ?
BTW, I don't understand why the codecov coverge dropped.

mhils: > BTW, I don't understand why the codecov coverge dropped.

Codecov bug. 

Other than that, looking at the issues you are describing the current test is good enough. I'll add a comment and we're good to go! :)

Issue #1147, opened 2016-05-15
Response streaming does not work in replay mode


ghost:
##### What's the problem?

While testing mitmdump replay I see that it consumes multiple GB of RAM and multiple GB of data went through the network. Because mitmdump eats up all physical memory as well as nearly all memory of swap the system was nearly dead. The client could not load a lot of the test pages.
Friendly killing mitmdump was not successful for more than 20 minutes and in the log files I could see crash reports.
After some search it seems that my client start download of several big files, many of them are bigger than one GB, but abort it after less than two MB.
##### Steps to reproduce the problem:
1. start mitmdump in transparent proxy mode, SSL and --no-pop
2. start on the client a big download like Linux DVD
3. abort download 
4. watch mitmdump downloading the whole file for the next minutes or hours and eating up memory
##### What is the expected behaviour?

The client disconnect and thus does not need the information any more. Therefore mitmdump should abort the download. Additionally for me it seems to be a good idea to not load big files completely into the memory.
##### What went wrong?

One of my test pages contains JavaScript/links to movies. These mp4 files are mostly bigger than one GB bur some of them are only about 500 MB. It seems that Internet Explorer peek into these files. I could not see Firefox doing this. As a result of the behaviour of the Internet Explorer mitmdump tried to download all these files completely and dumping them in the memory. But the sum of all of these files are more than the memory of machine mitmdump is running on. Thus the system slows down a lot and mitmdump is not able to replay other sites.
With this behaviour it is easy for a client to do a denial of service attack against mitmdump, even though it is not the intention of the client.
##### What is an example of the crash log?

For me it seems that mitmdump finished a 'smaller' download and then tries to send it to the client which disconnected minutes ago.
Here is an example:

```
  << 200 OK 692.45MB

 Error in processing of request from 192.168.*.*:50000
 Traceback (most recent call last):
   File "/usr/local/lib/python2.7/dist-packages/netlib/tcp.py", line 854, in connection_thread
     self.handle_client_connection(connection, client_address)
   File "/usr/local/lib/python2.7/dist-packages/mitmproxy/proxy/server.py", line 62, in handle_client_connection
     self.channel
   File "/usr/local/lib/python2.7/dist-packages/mitmproxy/proxy/server.py", line 75, in __init__
     None)
   File "/usr/local/lib/python2.7/dist-packages/mitmproxy/models/connections.py", line 17, in __init__
     super(ClientConnection, self).__init__(client_connection, address, server)
[...]

PR #2369, opened 2017-06-02
HTTP request & WebSocket message streaming

### LINKED ISSUE #1764: Overhaul Streaming Support
ref #856
ref #1147
ref #1506

Recent core refactoring efforts touched our streaming support for large responses a couple of times.
Here a couple of proposals which could improve things a lot:
  * support request streaming
     - useful for file uploads, inspecting S3 clients and other HTTP-based storage APIs
     - see #1506 for an unfortunately now obsolete PR
  * stream by default for files larger than X bytes
     - AFAIK the current default is to just fail the request if the returned `Content-Length` is larger than `--body-size-limit`. This is not very convenient, nor intuitive.
  * support WebSocket streaming for messages and/or frames
     - typical WebSocket messages are small JSON-ish things, but one could do large binary file transfers as well.

ujjwal96:
* [x] http1
* [x] http1 tests
* [x] http2
* [x] http2 tests
* [x] websockets  
* [x] websockets tests

closes #1764
ref #1147


mitmproxy/proxy/protocol/http.py
```diff
@@ -327,7 +330,15 @@ def _process_flow(self, f):
                 )
 
                 def get_response():
-                    self.send_request(f.request)
+                    if f.request.stream:
```
Kriechi: Does `f.request.stream` get set to `True` anywhere? Or is this inherited from the flow?

ujjwal96: Here:
https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/addons/streambodies.py#L30

mitmproxy/proxy/protocol/http2.py
```diff
@@ -487,9 +487,17 @@ def read_request_headers(self, flow):
 
     @detect_zombie_stream
     def read_request_body(self, request):
-        self.request_data_finished.wait()
```
Kriechi: We probably want to keep the wait for non-streaming requests (as you had before)

mitmproxy/proxy/protocol/http2.py
```diff
@@ -487,11 +487,89 @@ def read_request_headers(self, flow):
...
+    def send_request_body(self, request, chunks):
+        if not self.no_body:
+            self.connections[self.server_conn].safe_send_body(
+                self.raise_zombie,
+                self.server_stream_id,
+                chunks
+            )
 
     @detect_zombie_stream
     def send_request(self, message):
```
Kriechi: `send_request(...)` now can just be:
```
  send_request_header(...)
  send_request_body(...)
```
Or did you change anything else?

test/mitmproxy/proxy/protocol/test_http1.py
```diff
@@ -77,3 +79,35 @@ def test_head_content_length(self):
...
+            self.add_log(v)
+            return self.o.sendall(v)
+        else:
+            r = self.o.write(v)
+            self.add_log(v[:r])
+            return r
+
+
+class TestRequestStreaming(tservers.HTTPProxyTest):
+    @mock.patch("mitmproxy.net.tcp.Writer.write", side_effect=write, autospec=True)
```
Kriechi: If you move the mocking after creating the pathoc object, we should only affect new connections, but not pathoc...

test/mitmproxy/proxy/protocol/test_http1.py
```diff
@@ -83,7 +82,7 @@ def test_head_content_length(self):
         assert resp.headers["Content-Length"] == "42"
 
 
-class TestStreaming(tservers.HTTPProxyTest):
+class TestRequest(tservers.HTTPProxyTest):
```
Kriechi: This still tests response streaming as well, why did you change the class name?

test/mitmproxy/proxy/protocol/test_http2.py
```diff
@@ -862,3 +863,91 @@ def test_connection_terminated(self):
...
         assert connection_terminated_event.additional_data == b'foobar'
+
+
+class TestRequestStreaming(_Http2Test):
+
+    @classmethod
+    def handle_server_event(cls, event, h2_conn, rfile, wfile):
+        if isinstance(event, h2.events.ConnectionTerminated):
+            return False
+        elif isinstance(event, h2.events.RequestReceived):
```
Kriechi: I think there should be a `DataReceived` event which we should use instead.

test/mitmproxy/proxy/protocol/test_http2.py
```diff
@@ -862,3 +863,91 @@ def test_connection_terminated(self):
...
+            req = rfile.read(109)
```
Kriechi: Why 109 bytes?

test/mitmproxy/proxy/protocol/test_http2.py
```diff
@@ -862,3 +863,91 @@ def test_connection_terminated(self):
...
+            assert req
+            h2_conn.close_connection(error_code=5, last_stream_id=42, additional_data=req[9:])
+            wfile.write(h2_conn.data_to_send())
+            wfile.flush()
+
+        return True
+
+    def _send_request(self,
```
Kriechi: Please modify the existing `_send_request` function so we can change this with a new argument / config.

test/mitmproxy/proxy/protocol/test_http2.py
ujjwal96: 9 bytes headers + 100 bytes bodies
Anything less would throw an error [here](https://github.com/mitmproxy/mitmproxy/pull/2369/files#diff-ae3ce6b83d0e48d6158f2e5fe6583e6bR911) inside `safe_read` because it then reads the length incorrectly

mitmproxy/addons/streambodies.py
```diff
@@ -15,8 +15,15 @@ def configure(self, updated):
             except ValueError as e:
                 raise exceptions.OptionsError(e)
 
-    def run(self, f, is_request):
+    def run(self, f, is_request, is_wsflow):
         if self.max_size:
+            if is_wsflow:
```
Kriechi: IMO in this case it is cleaner to use `isinstance(f, WebSocketFlow)` instead of a new bool argument.

Kriechi:
@mitmproxy/devs this is ready for review!
IMO this is ready to merge, but I might have missed something important.

mitmproxy/proxy/protocol/http.py
```diff
@@ -327,7 +330,15 @@ def _process_flow(self, f):
                 )
 
                 def get_response():
```
mhils: `get_response` can be called twice (see [here](https://github.com/mitmproxy/mitmproxy/pull/2369/files#diff-1a48d22b11223fa300cac1e75cbcfafcR351) for the reasoning). If `stream` is set to True, the second send will have no body, because we have already depleted the body generator on the first attempt.

Unfortunately this is really hard to fix. I see two possible ways forward:

1. Store all chunks _somewhere_ so that we can play them again. That'd be nice as we then could always store the content while streaming, but that also adds substantial complexity.
2. Figure out a better way to see if the socket is closed. Maybe it's enough to just try reading from it with MSG_PEEK. Alternatively, maybe we can make `get_response` a `send_headers` function, and then do everything else (sending body + receiving response) outside of the try-except-repeat block. I am not sure if any of this works, we need to test that.

I think the second option is the generally nicer way (we can still implement 1 later), so we should probably try and improve that in a separate PR.

mitmproxy/proxy/protocol/websocket.py
```diff
@@ -70,43 +71,50 @@ def _handle_data_frame(self, frame, source_conn, other_conn, is_server):
...
+                def get_chunk(payload):
+                    if len(payload) == length:
+                        # message has the same length, we can reuse the same sizes
+                        pos = 0
+                        for s in original_chunk_sizes:
+                            yield payload[pos:pos + s]
+                            pos += s
+                    else:
+                        # just re-chunk everything into 10kB frames
+                        chunk_size = 10240
```
mhils: Minor suggestion: Make this a power of two, see e.g. the note in https://docs.python.org/3/library/socket.html#socket.socket.recv.

Kriechi: Good point! @ujjwal96 please change it to 4096.

mitmproxy/proxy/protocol/http.py
Kriechi: Sounds all reasonable and we should fix this, but isn't this a general response streaming issue unrelated to this PR? If so, please open a new issue, and we can address it after merging this PR.

mitmproxy/proxy/protocol/websocket.py
```diff
@@ -70,43 +71,50 @@ def _handle_data_frame(self, frame, source_conn, other_conn, is_server):
...
+                def get_chunk(payload):
+                    if len(payload) == length:
+                        # message has the same length, we can reuse the same sizes
+                        pos = 0
+                        for s in original_chunk_sizes:
+                            yield payload[pos:pos + s]
+                            pos += s
+                    else:
+                        # just re-chunk everything into 4kB frames
+                        chunk_size = 4096
```
Kriechi: Thinking about it again, we should consider the frame header as well.
So the size of each chunk should be 4096 - header.
We also have to account for the direction a frame is sent, since the masking key is only for client->server frames.
https://tools.ietf.org/html/rfc6455#section-5.2

mitmproxy/proxy/protocol/http.py
mhils: As discussed on Slack, this is specific to this PR - `get_response` is retried, and on retry we have potentially already streamed some parts of the body.

mitmproxy/addons/streambodies.py
```diff
@@ -28,12 +29,18 @@ def run(self, f, is_request):
...
 
-    # FIXME! Request streaming doesn't work at the moment.
     def requestheaders(self, f):
         self.run(f, True)
 
     def responseheaders(self, f):
         self.run(f, False)
+
+    def websocket_start(self, f):
+        if ctx.options.stream_websockets and isinstance(f, websocket.WebSocketFlow):
```
Kriechi: I guess the `isinstance` check is not needed, who else would call `websocket_start`?

mitmproxy/addons/streambodies.py
```diff
@@ -28,12 +29,18 @@ def run(self, f, is_request):
...
+            f.stream = True
+            ctx.log.info("Streaming WebSocket message {client} - {server}".format(
```
Kriechi: From here on we are streaming all messages in this flow. The info text could be misleading.
Maybe rephrase to something like: **Streaming all WebSocket messages between {client} and {server}**

mitmproxy/options.py
```diff
@@ -155,6 +155,13 @@ def __init__(self, **kwargs) -> None:
             """
         )
         self.add_option(
+            "stream_websockets", bool, False,
+            """
+            Stream WebSocket frames as and when they are received
+            by the proxy. If True message payload cannot be modified.
```
Kriechi: Stream WebSocket **messages between client and server. Messages are captured and cannot be modified.**

docs/features/streaming.rst
```diff
@@ -0,0 +1,100 @@
...
+
+As the name suggests, request streaming feature can be used to stream request body to the server incrementally
+before it has been fully received by the proxy. This may be useful for large file uploads.
+
+
+By default, mitmproxy will read the entire request/response, perform any indicated
+manipulations on it and then send the (possibly modified) request or response to
+the server or the client respectively. In some cases this is undesirable and you may wish to "stream"
+the request/response. When streaming is enabled, the request/response is
+not buffered on the proxy but directly sent to the server/client instead.
```
Kriechi: Move this paragraph to the top of the file, underneath a top-level headline: **HTTP Streaming**
Also mention that HTTP headers are still fully buffered before being sent.

Then the two sub-headlines for **Request Body Streaming** and **Response Body Streaming**.

test/mitmproxy/proxy/protocol/test_websocket.py
```diff
@@ -328,3 +334,30 @@ def test_invalid_frame(self):
...
+
+        if not streaming:
+            with pytest.raises(exceptions.TcpDisconnect):  # Reader.safe_read get nothing as result
+                frame = websockets.Frame.from_file(self.client.rfile)
+        else:
+            frame = websockets.Frame.from_file(self.client.rfile)
+
+        if streaming:
+            assert frame
+            assert self.master.state.flows[1].messages == []
```
Kriechi: We should asserting *something* if we are not streaming...
Or remove the `False` parameter option.

docs/features/streaming.rst
Kriechi: For future reference: this was fixed with a new commit.

Kriechi:
@mhils the HTTP/1 retry issue should be sufficiently resolved. Do we get a final seal of approval?

PR #2375, opened 2017-06-04
Always pretend to have a tty in tests

mhils:
This should fix appveyor. 
Pytest's monkeypatch [doesn't work out of the box](https://github.com/pytest-dev/pytest/issues/363), so I'm just using `unittest.mock` from stdlib.

### DIFF
--- a/test/mitmproxy/tools/console/test_master.py
+++ b/test/mitmproxy/tools/console/test_master.py
@@ -1,3 +1,5 @@
+import pytest
+
 from mitmproxy.test import tflow
 from mitmproxy.test import tutils
 from mitmproxy.tools import console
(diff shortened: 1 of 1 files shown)


Kriechi:
Why is this failing on Windows but not on Linux/macOS in the first place?

mhils:
I guess `isatty()` is true on travis?

Kriechi:
I'm not sure what is going on here exactly, but I think just excluding all ConsoleMaster tests from appveyor makes more sense? `test/mitmproxy/tools/console/test_master.py`

mhils:
I think it's actually quite good to have as much coverage as (easily) possible for Windows developers. Besides, the surprising behaviour is Linux/OSX for me here, at least it seems weird that tests are run in a tty :)

PR #2376, opened 2017-06-05
[WIP][web] Add tests for components/ContentView (con't)

MatthewShao:
- [x] ContentLoader.jsx
- [x] ContentView.jsx
- [x] DownloadContentButton.jsx
- [x] MetaView.jsx
- [x] ShowFullContentButton.jsx
- [x] UploadContentButton.jsx
- [x] ViewSelector.jsx

### DIFF
--- a/web/src/js/__tests__/components/ContentView/ContentLoaderSpec.js
+++ b/web/src/js/__tests__/components/ContentView/ContentLoaderSpec.js
@@ -1,0 +1,75 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+import ContentLoaderGenerator from '../../../components/ContentView/ContentLoader'
+import { TFlow } from '../../ducks/tutils'
+import TestUtils from 'react-dom/test-utils'
+import mockXMLHttpRequest from 'mock-xmlhttprequest'
+
+global.XMLHttpRequest = mockXMLHttpRequest
+
+class tComponent extends React.Component {
+    constructor(props, context){
+        super(props, context)
+    }
+    render() {
+        return (<p>foo</p>)
+    }
+}
+
+let tflow = new TFlow(),
+    ContentLoader = ContentLoaderGenerator(tComponent)
+
+describe('ContentLoader Component', () => {
+    it('should render correctly', () => {
+        let contentLoader = renderer.create(<ContentLoader flow={tflow} message={tflow.response}/>),
+            tree = contentLoader.toJSON()
+        expect(tree).toMatchSnapshot()
+    })
+
+    let contentLoader = TestUtils.renderIntoDocument(<ContentLoader flow={tflow} message={tflow.response}/>)
+
+    it('should handle updateContent', () => {
+        tflow.response.content = 'foo'
+        contentLoader.updateContent({flow: tflow, message: tflow.response})
+        expect(contentLoader.state.request).toEqual(undefined)
+        expect(contentLoader.state.content).toEqual('foo')
+        // when content length is 0 or null
+        tflow.response.contentLength = 0
+        tflow.response.content = undefined
+        contentLoader.updateContent({flow: tflow, message: tflow.response})
+        expect(contentLoader.state.request).toEqual(undefined)
+        expect(contentLoader.state.content).toEqual('')
+    })
+
+    it('should handle componentWillReceiveProps', () => {
+        contentLoader.updateContent = jest.fn()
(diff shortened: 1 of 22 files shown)


MatthewShao:
@mhils If it looks good, please do not hurry to merge this PR, I would like to cover the whole `ContentView` directory and then merge. Giving it an approval would be welcome though.:smile:

web/src/js/__tests__/components/ContentView/MetaViewsSpec.js
```diff
@@ -0,0 +1,37 @@
...
+            tree = contentMissing.toJSON()
+        expect(tree).toMatchSnapshot()
+    })
+})
+
+describe('ContentTooLarge Components', () => {
+    it('should render correctly', () => {
+        let clickFn = jest.fn(),
+            uploadContentFn = jest.fn(),
+            contentTooLarge = renderer.create(<ContentTooLarge
```
mhils: indentation looks quite confusing here.

web/src/js/__tests__/components/ContentView/ContentLoaderSpec.js
```diff
@@ -0,0 +1,75 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+import ContentLoaderGenerator from '../../../components/ContentView/ContentLoader'
```
mhils: Any particular reason why you added "Generator" here? :)

web/src/js/__tests__/components/ContentView/ContentLoaderSpec.js
```diff
@@ -0,0 +1,75 @@
...
+import { TFlow } from '../../ducks/tutils'
+import TestUtils from 'react-dom/test-utils'
+import mockXMLHttpRequest from 'mock-xmlhttprequest'
+
+global.XMLHttpRequest = mockXMLHttpRequest
```
mhils: I'm not sure if I understand this without a comment - which purpose does this serve here and in ContentViewSpec.js?

web/src/js/__tests__/components/ContentView/ContentLoaderSpec.js
MatthewShao: As far as I consider, the anonymous class in `ContentLoader` is just a wrapper, which takes a `View` component and attach some methods to it, it is not the actual content loader. So I named it `ContentLoaderGenerator` and its return component `ContentLoader`. I know this is wired, so I wonder if you have any suggestions?

web/src/js/__tests__/components/ContentView/ContentLoaderSpec.js
MatthewShao: Because we use `XMLHttpRequest` here:
https://github.com/mitmproxy/mitmproxy/blob/5f91b7a45d0b983d3e0f3fff1be8af93736d35c6/web/src/js/components/ContentView/ContentLoader.jsx#L59
Introducing `mock-xmlhttprequest` dependency can save us a lot of troubles.

web/src/js/__tests__/components/ContentView/ContentLoaderSpec.js
mhils: Speaking in React terms, it's a [Higher-Order Component](https://facebook.github.io/react/docs/higher-order-components.html), like react-redux's `connect`. So maybe just call it `contentLoader` or `withContentLoader`? I have no strong preference, but we should do the same as we do in the actual non-test code. Please feel free to adjust both :)

web/src/js/__tests__/components/ContentView/ContentLoaderSpec.js
mhils: Ok, but we never tell the mock function what to return, so what would happen if we didn't do this here?

MatthewShao: If we don't mock `XMLHttpRequest`, it fails like this:
default

web/src/js/components/ContentView/ContentLoader.jsx
```diff
@@ -2,98 +2,100 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { MessageUtils } from '../../flow/utils.js'
 
-export default View => class extends React.Component {
+export default function withContentLoader(View) {
+    
+    return class extends React.Component {
```
mhils: Good call on making this more clear! 👍

mhils:
Awesome! :)

PR #2390, opened 2017-06-12
Remove private key from PKCS12 file

### LINKED ISSUE #2349: Private key is present in mitmproxy-ca-cert.p12

Private key is inserted in .p12 file, although filename says "mitmproxy-ca-**cert**.p12"
Output can be verified with:
openssl pkcs12 -in mitmproxy-ca-cert.p12 -nocerts -nodes

I don't have a Windows machine to test if a .p12 file would work without the private key inside.
We can generate a .p12 file without key with:
openssl pkcs12 -export -nokeys -in mitmproxy-ca-cert.pem -out mitmproxy-ca-cert.p12

-> If private key is needed, file could be renamed "mitmproxy-ca.p12" for better understanding
-> If private key is not needed, .p12 generation should be modified

##### System information

Mitmproxy version: 2.0.1 (release version) Precompiled Binary
Python version: 3.5.2
Platform: Darwin-16.6.0-x86_64-i386-64bit
Mac version: 10.12.5 ('', '', '') x86_64

ograff:
Fixes #2349

### DIFF
--- a/mitmproxy/certs.py
+++ b/mitmproxy/certs.py
@@ -266,6 +266,12 @@ class CertStore:
         with open(os.path.join(path, basename + "-ca-cert.p12"), "wb") as f:
             p12 = OpenSSL.crypto.PKCS12()
             p12.set_certificate(ca)
+            f.write(p12.export())
+
+        # Dump the certificate and key in a PKCS12 format for Windows devices
+        with open(os.path.join(path, basename + "-ca.p12"), "wb") as f:
+            p12 = OpenSSL.crypto.PKCS12()
+            p12.set_certificate(ca)
             p12.set_privatekey(key)
             f.write(p12.export())
 


ograff:
@Kriechi D'oh, didn't think of that. Thanks!

ograff:
@mhils My bad! I really appreciate you guys walking through this with me as I'm sure it would be quicker to just do it yourselves.

mhils:
No worries! I personally would have fixed it as you did in your initial PR, so this is actually good here 😄.

Kriechi:
Do we need to update the docs?
http://docs.mitmproxy.org/en/stable/certinstall.html#windows-automated
http://docs.mitmproxy.org/en/stable/certinstall.html#ca-and-cert-files

ograff:
@Kriechi I don't believe so. Both of those reference installing the certificate, the private key need not be installed anywhere but the host MITMing the traffic.

Kriechi:
Thanks! 🎉

tomlabaude:
Awesome thank you guys, few hours too late, I've lost an opportunity for my 1st simple PR!

I haven't seen the 1st PR, where the private key was lost, but guess, it was simply deleting `p12.set_privatekey(key)`, without -ca.p12.

Is it really a problem as private key is still in -ca.pem?

-ca.p12 won't certainly be used without openssl command lines tricks (which could be done on -ca.pem), so does it worth keeping the file?

mhils:
> Is it really a problem as private key is still in -ca.pem?

Not for mitmproxy's core operation, but still some users may want the private key exported in PKCS format.

PR #2395, opened 2017-06-13
[WIP] Add RESTful API for mitmweb option

MatthewShao:
- [x] Implement GET /options
- [x] Implement POST /options
The return json looks like this:
```python
[
    {
        "default": false, 
        "help": "Add all certificates of the upstream server to the certificate chain that will be served to the proxy client, as extras.", 
        "name": "add_upstream_certs_to_client_chain", 
        "type": "bool", 
        "value": false
    }, 
    {
        "default": false, 
        "help": "Strip out request headers that might cause the server to return 304-not-modified.", 
        "name": "anticache", 
        "type": "bool", 
        "value": false
    }, 
...]
```
The tests passed on my local repo, but kept falling on Travis... Is that because of some compatibility problem?

### DIFF
--- a/mitmproxy/optmanager.py
+++ b/mitmproxy/optmanager.py
@@ -420,12 +420,27 @@ def dump_dicts(opts):
     """
         Dumps the options into a list of dict object.
 
-        Return: A list like: [ { name: "anticahce", type: "bool", default: false, value: true, help: "help text"}]
+        Return: A list like: [ { name: "anticache", type: "bool", default: false, value: true, help: "help text"} ]
     """
     options_list = []
     for k in sorted(opts.keys()):
         o = opts._options[k]
-        option = {'name': k, 'type': o.typespec.__name__, 'default': o.default, 'value': o.current(), 'help': o.help.strip()}
+        if o.typespec in (str, int, bool):
+            t = o.typespec.__name__
+        elif o.typespec == typing.Optional[str]:
+            t = 'Union'
+        elif o.typespec == typing.Sequence[str]:
+            t = 'Sequence'
+        else:
+            raise NotImplementedError
+        option = {
+            'name': k,
+            'type': t,
+            'default': o.default,
+            'value': o.current(),
+            'help': o.help,
+            'choices': o.choices
+        }
         options_list.append(option)
     return options_list
 
--- a/mitmproxy/optmanager.py
+++ b/mitmproxy/optmanager.py
@@ -425,14 +425,7 @@ def dump_dicts(opts):
     options_list = []
     for k in sorted(opts.keys()):
         o = opts._options[k]
-        if o.typespec in (str, int, bool):
-            t = o.typespec.__name__
-        elif o.typespec == typing.Optional[str]:
-            t = 'Union'
-        elif o.typespec == typing.Sequence[str]:
-            t = 'Sequence'
-        else:
-            raise NotImplementedError
+        t = typecheck.typespec_to_str(o.typespec)
         option = {
             'name': k,
             'type': t,
--- a/mitmproxy/optmanager.py
+++ b/mitmproxy/optmanager.py
@@ -416,6 +416,20 @@ def dump_defaults(opts):
     return ruamel.yaml.round_trip_dump(s)
 
 
+def dump_dicts(opts):
+    """
+        Dumps the options into a list of dict object.
+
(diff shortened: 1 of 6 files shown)


mitmproxy/optmanager.py
```diff
@@ -416,6 +416,20 @@ def dump_defaults(opts):
     return ruamel.yaml.round_trip_dump(s)
 
 
+def dump_dicts(opts):
+    """
+        Dumps the options into a list of dict object.
+
+        Return: A list like: [ { name: "anticahce", type: "bool", default: false, value: true, help: "help text"}]
```
mhils: typo: anticache

mitmproxy/optmanager.py
```diff
@@ -416,6 +416,20 @@ def dump_defaults(opts):
...
+    """
+    options_list = []
+    for k in sorted(opts.keys()):
+        o = opts._options[k]
```
mhils: You can simplify this a tiny bit:
```python
for k, o in sorted(opts.items()):
```

mitmproxy/optmanager.py
```diff
@@ -416,6 +416,20 @@ def dump_defaults(opts):
...
+        option = {'name': k, 'type': o.typespec.__name__, 'default': o.default, 'value': o.current(), 'help': o.help.strip()}
```
mhils: Maybe add some line breaks here to make this easier to read?

cortesi: Also, the help.strip() here should be un-necessary - we dedent and format help text on initialisation. Let's keep the help formatting in one place.

mhils: Good catch. This also needs to support `.choices`.

mitmproxy/optmanager.py
```diff
@@ -416,6 +416,35 @@ def dump_defaults(opts):
...
+def dump_dicts(opts):
+    """
+        Dumps the options into a list of dict object.
+
+        Return: A list like: [ { name: "anticache", type: "bool", default: false, value: true, help: "help text"} ]
+    """
+    options_list = []
+    for k in sorted(opts.keys()):
+        o = opts._options[k]
+        if o.typespec in (str, int, bool):
```
mhils: Move this typespec to string conversion into its own function in mitmproxy/utils/typecheck? :)

mitmproxy/optmanager.py
MatthewShao: I'm afraid we could not do this, `opts` here is not a `dict`, but a `OptManager`, or do you think we should add a `.items` method for it?

mhils: Ah, I see. Let's keep it as is then. :)

mitmproxy/optmanager.py
MatthewShao: Good advice!

test/mitmproxy/tools/web/test_app.py
```diff
@@ -253,6 +253,15 @@ def test_settings_update(self):
...
         assert self.put_json("/settings", {"wtf": True}).code == 400
 
+    def test_options(self):
+        j = json(self.fetch("/options"))
+        assert type(j) == list
+        assert type(j[0]) == dict
+
+    def test_option_update(self):
+        assert self.put_json("/options", {"anticache": True}).code == 200
+        assert self.put_json("/options", {"wtf": True}).code == 400
```
mhils: We should also test type confusion here (e.g. `{"anticache": "foo"}`)

mitmproxy/utils/typecheck.py
```diff
@@ -98,3 +98,15 @@ def check_option_type(name: str, value: typing.Any, typeinfo: typing.Any) -> Non
         return
     elif not isinstance(value, typeinfo):
         raise e
+
+
+def typespec_to_str(typespec: typing.Any) -> str:
```
mhils: We should also use this in `dump_defaults`

mitmproxy/utils/typecheck.py
```diff
@@ -98,3 +98,15 @@ def check_option_type(name: str, value: typing.Any, typeinfo: typing.Any) -> Non
...
+    if typespec in (str, int, bool):
+        t = typespec.__name__
+    elif typespec == typing.Optional[str]:
+        t = 'Union'
```
mhils: Make this `"optional str"` (and `"sequence of str"` below this) to match what is currently being done in `dump_defaults`

Issue #2409, opened 2017-06-23
mitm.it different certs


fulldecent:
Please clarify, in program documentation, if mitm.it installs a certificate a different certificate for each user.

If not then basically anyone that also has the certificate can spoof any website on my phone.

Either way is OK, but this should be clearly specified.

Kriechi:
You mean like this?
> The mitmproxy CA cert is located in ~/.mitmproxy after it has been generated at the first start of mitmproxy.

Take from:
http://docs.mitmproxy.org/en/stable/certinstall.html#installing-the-mitmproxy-ca-certificate-manually

mhils:
Thanks for the feedback, but I think we already do that. The following text is displayed on mitm.it:

> This page is served by your local mitmproxy instance. The certificate you are about to install has been uniquely generated on mitmproxy's first run and is not shared between mitmproxy installations.

fulldecent:
Thank you, now scrolling all the way down I can see the full text. I got scared and closed it when I saw "**the** mitmproxy certificate", assuming there is only one.

PR coming.

PR #2410, opened 2017-06-23
Slight wording change

fulldecent:
THE implies there is only one. THE Eiffel tower.

YOUR implies everyone has a different one. YOUR teeth.


fulldecent:
Fixes #2409

mhils:
😄

That's definitely better, thanks! 🍰

PR #2413, opened 2017-06-25
[web] Add tests for js/components/*

MatthewShao:


web/src/js/__tests__/components/EventLogSpec.js
```diff
@@ -0,0 +1,57 @@
...
+            <ConnectedComponent/>
+        </Provider>),
+        tree = provider.toJSON()
+
+    it('should connect to state and render correctly', () => {
+        expect(tree).toMatchSnapshot()
+    })
+
+    it('should handl toggleFilter', () => {
+        let debugToggleButton = tree.children[0].children[1].children[0]
```
mhils: This seems weird. Shouldn't we rather test EventLog.toggleFilter?
Or is the issue that we don't have coverage for `() => toggleFilter(type)` then?

web/src/js/__tests__/components/EventLogSpec.js
```diff
@@ -0,0 +1,57 @@
+jest.mock('../../components/EventLog/EventList')
+
+import React from 'react'
+import renderer from 'react-test-renderer'
+import TestUtils from 'react-dom/test-utils'
+import ConnectedComponent, { EventLog } from '../../components/EventLog'
```
mhils: Maybe call this EventLog / PureEventLog? I was reading the code below and it was not really clear to me what `ConnectedComponent` is.

web/src/js/__tests__/components/EventLogSpec.js
MatthewShao: Yes, that looks weird, but that's the only way I came up with to cover `() => toogleFilter(type)`.

PR #2416, opened 2017-06-27
[WIP] [web] Mitmweb options editor UI

MatthewShao:
- [x] Add reducer for the modal
- [x] Implement Modal Component
- [x] Add tests


MatthewShao:
I implemented the modal component with the help of the work done by @cle1000 last summer. However, my implementation might by immature, so please feel free to comment on it.
Here's how it looks now:
peek 2017-06-28 18-39

web/src/js/components/Modal/Modal.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+class PureModal extends Component {
+
+    constructor(props, context) {
+        super(props, context)
+    }
+
+    render() {
+        const { activeModal, hideModal } = this.props
+        const ActiveModal = _.find(ModalList, m => m.name === activeModal )
```
cle1000: _.find returns undefined if the element is not found 
https://lodash.com/docs/#find
you can rewrite this to 

> const ActiveModal = _.find(ModalList, m => m.name === activeModal ) || Fallback case (<div/>) <
then you can just return ActiveModal

>

MatthewShao: I'm afraid it looks weired if we add a fallback component in this way, because we have ` <div className="modal-backdrop fade in"></div>` below, which will make the background goes dark. 
How about keep this line and changing the retun like:
```javascript
return(
    activeModal && ActiveModal ? ():<div/>
)
```

cle1000: ok I see, maybe we can write it like this:
> return(
>     ActiveModal && <div/>
> )

MatthewShao: Should we remove `<div className="modal-backdrop fade in"></div>` to omit the dark background  or put it into every modal component in the ModalList?

cle1000: -> slack

web/src/js/components/Header/FileMenu.jsx
```diff
@@ -50,5 +57,6 @@ export default connect(
         clearFlows: flowsActions.clear,
         loadFlows: flowsActions.upload,
         saveFlows: flowsActions.download,
+        openModal: modalActions.setActiveModal,
```
mhils: We can make this `openOptions: () => modalActions.setActiveModal('OptionModal')`. I'm surprised we don't need to call `e.preventDefault()?` as with the other options?

web/src/js/components/Modal/Modal.jsx
mhils: Haven't seen your Slack discussion, but my five cents: If `activeModal` is not undefined, I think it's fine to assume that `.find()` is sucessful, so there shouldn't be a need for a fallback. It's ok (if not useful) if it breaks visibly if there's an invalid state.

Other than that, no need to use lodash here - ES6 has Array.prototype.find. 😃

MatthewShao:
There was a discussion between me and @cle1000 about the Modal component, he suggests that we should have a generic component for the modal, which should contains all the redundant information. Besides, we should make the `Modal.jsx` as simple as possible. Therefore I reconsider the structure of the modal component and then make some changes:
1. `Modal.jsx`: be responsible for switching between different modal components, and turning them on and off.
2. `ModalLayout.jsx`: returns a generic component, containing `<div className='modal-backdrop'>` for the dark background and  `div className='modal'`for the layout.
3. `OptionModal.jsx`: returns the header, body, footer which below to Modal, also takes care of the logic of fetching and updating options. We need to connect it to the state.
4. `ModalList.jsx`: exporting all the available modals. Because we select the modal component by function name in `Modal.jsx`, so we need to give a specific name to every modal component. However, if we want to connect the component to the state, the `connect` function can not return a function with specific name. My solution is wrapping up the component in the `ModalList.jsx`, so we could query the modal by name while still have it connect to the state.

PR #2418, opened 2017-06-28
Fixes the issue of script reloader not unloading previous script

### LINKED ISSUE #2315: Script Reloader does not unload old script

1. mitmdump -s script.py
2. Modify script on disk.

##### System information

Mitmproxy version: 3.0.0 (2.0.0dev0407-0x315daa0)
Python version: 3.5.2
Platform: Linux-4.4.0-43-Microsoft-x86_64-with-Ubuntu-16.04-xenial

ujjwal96:
Fixes #2315
Well, this was a tricky one.
**The issue -** 
We use [`load_module`](https://docs.python.org/3/library/importlib.html#importlib.abc.Loader.load_module) to load the scripts. According to the docs -

> If the requested module already exists in sys.modules, that module should be used and reloaded

So, the modified script was [`reloaded`](https://docs.python.org/3/library/importlib.html#importlib.reload) 

> When a module is reloaded, its dictionary (containing the module’s global variables) is retained. Redefinitions of names will override the old definitions, so this is generally not a problem. If the new version of a module does not define a name that was defined by the old version, the old definition remains.

**How I solved it-**
Before loading a script remove it from sys.modules. 
If you have a better solution I'll implement that :slightly_smiling_face: 

This PR also fixes the issue caused when a script is deleted during execution

### DIFF
--- a/mitmproxy/addons/script.py
+++ b/mitmproxy/addons/script.py
@@ -52,11 +52,19 @@ class Script:
 
     def tick(self):
         if time.time() - self.last_load > self.ReloadInterval:
-            mtime = os.stat(self.fullpath).st_mtime
+            try:
+                mtime = os.stat(self.fullpath).st_mtime
+            except FileNotFoundError:
+                scripts = ctx.options.scripts
+                scripts.remove(self.path)
+                ctx.options.update(scripts=scripts)
+                return
+
             if mtime > self.last_mtime:
                 ctx.log.info("Loading script: %s" % self.path)
                 if self.ns:
                     ctx.master.addons.remove(self.ns)
+                    del sys.modules[self.ns.__name__]
                 self.ns = load_script(ctx, self.fullpath)
                 if self.ns:
                     # We're already running, so we have to explicitly register and
--- a/test/mitmproxy/addons/test_script.py
+++ b/test/mitmproxy/addons/test_script.py
@@ -1,6 +1,7 @@
 import traceback
 import sys
 import time
+import os
 import pytest
 
 from unittest import mock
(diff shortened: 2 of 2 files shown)


mhils:
This only fixes the removal issue, right?

ujjwal96:
@mhils No, it fixes both the issues.

mitmproxy/addons/script.py
```diff
@@ -51,12 +51,25 @@ def addons(self):
         return [self.ns] if self.ns else []
 
     def tick(self):
+        if not os.path.exists(self.fullpath):  # File is deleted
```
mhils: Doing IO on every tick is a bad idea from a performance standpoint.

test/mitmproxy/addons/test_script.py
```diff
@@ -84,6 +85,25 @@ def test_reload(self, tmpdir):
                 time.sleep(0.1)
             tctx.master.has_log("Loading")
 
+    def test_deletion(self):
+        with open(tutils.test_data.path("mitmproxy/data/addonscripts/") + "/dummy.py", 'w') as f:
```
mhils: Simplify concatenation?

test/mitmproxy/addons/test_script.py
```diff
@@ -84,6 +85,25 @@ def test_reload(self, tmpdir):
...
+            f.write("\n")
+        with taddons.context() as tctx:
+            sc = script.Script(
+                tutils.test_data.path("mitmproxy/data/addonscripts/dummy.py")
+            )
+            tctx.master.addons.add(sc)
+            tctx.configure(sc)
+
+            sl = script.ScriptLoader()
```
mhils: I do not quite understand why we have both a `Script` and a `ScriptLoader` instance here. This also probably needs to be moved to `TestScriptLoader`

mhils:
Fun times - the original issue in #2315 is already fixed apparently, but it looks like you managed to find a much more nasty one - good job! 😄 I think the current approach to delete it from sys.modules is the best we can do. 👍

test/mitmproxy/addons/test_script.py
ujjwal96: `ScriptLoader` addon is the parent addon which manages the `Script` addons. 
Once the script file is deleted the script addon should be removed from both `master.addons.lookup` and `ScriptLoader.addons` 
I used `ScriptLoader` so as to assert that the script is removed from it's addons as well

mhils: Ok, but why do you need to instantiate `Script`, can't that be done by `ScriptLoader`?

ujjwal96: It can be. Pushed a new commit

test/mitmproxy/addons/test_script.py
```diff
@@ -183,6 +184,20 @@ def test_dupes(self):
                     scripts = ["one", "one"]
                 )
 
+    def test_script_deletion(self):
+        tdir = tutils.test_data.path("mitmproxy/data/addonscripts/")
+        with open(tdir + "/dummy.py", 'w') as f:
```
mhils: Why not `tutils.test_data.path("mitmproxy/data/addonscripts/dummy.py")`? Am I missing something entirely obvious? 😄 
(also: re-use var below)

ujjwal96: It will raise a ValueError as the file doesn't exist.

mhils: Fair point 😄

PR #2423, opened 2017-07-02
[web] [WIP] Mitmweb options editor content

MatthewShao:
Following the pattern of `settings` and for the convenience of later operation, I changed the response of `GET /options` into something like this:
``` javascript
{
    anticache: {type: "bool", default: false, value: true, help: "help text"},
    intercept: {type: "str", default: null, value: null, help: "help text"},
    ....
}
```
Then I ported the options to the state, and made a toggle component for the option.
However, when clicking the toggle component in the option modal, it will only change the state of `settings`, instead of `options`. So we fail to switch it on and off.

@mhils @cle1000 Am I missing anything here?


web/src/js/components/Modal/OptionTypes.jsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+OptionsToggle = connect(
+    state => ({
+        options: state.options,
+    }),
+    {
+        updateOptions,
+    }
+)(OptionsToggle)
```
mhils: This still looks a bit unintuitive to me. If we want to separate data ("OptionsToggle") from view ("MenuToggle"), we should make the data part a [higher-order component](https://facebook.github.io/react/docs/higher-order-components.html), so that we can just do
```javascript
MenuToggle = connectOptions(PureMenuToggle);
```
instead of defining our own data component for each datatype (i.e. instead of "OptionsToggle").

`connectOptions` could look like this for example:
```javascript
// this is untested
function connectOptions(WrappedComponent) {
    function Wrapper({name, options, updateOptions, ...props}){
        let option = options[name],
            onChange = (val) => updateOptions({ [option]: val });
        return <WrappedComponent name={name} value={option.value} onChange={onChange} {...props} />;
    }
    Wrapper.displayName = /*...*/;
    return connect(
        state = ({
            options: state.options,
        }),
        {
            updateOptions,
        }
    )(Wrapper)
}
```

I had a few beers earlier, I hope that makes sense. Ping me or Clemens on Slack if it doesn't. 😄

mhils: Minor remark `MenuToggle` sounds a bit weird to me. Maybe just `BooleanOption`?

MatthewShao:
peek 2017-07-05 08-56
Following @mhils 's advices, I made some adjustments, here. I changed `OptionTypes.jsx` -> `OptionMaster.jsx`, which can switch from different option components according to the option type. 
Now `bool`, `str`, `int` and `choices` option types were implemented, but I'm not sure about how to deal with `optional str` and `sequence of str`. Also some options relevant to file operations maybe should have they corresponding component.
The view is still so ugly now, but we could add some CSS to beautify it once we nail down the structure.

web/src/js/components/Modal/OptionModal.jsx
```diff
@@ -28,9 +27,11 @@ class PureOptionModal extends Component {
                 </div>
 
                 <div className="modal-body">
-                    <OptionsToggle option="http2">HTTP/2.0</OptionsToggle>
-                    <OptionsToggle option="anticache">Anticache</OptionsToggle>
-                    <OptionsToggle option="anticomp">Anticomp</OptionsToggle>
+                    {
+                        Object.keys(options).sort().map((key) => (
+                            <Option option={key}/>
```
mhils: Add a `key=` here or React will probably complain. :-)

MatthewShao:
These commits includes following changes:
1. I ditch the connect in `OptionTypes`, because that will make our test work easier. 
2. Move `<div className="menu-entry">` to the `OptionMaster`( we called it `Wrapper` originally)
3. Update the tests for ui as well as the backend websocket part.
4. Add `key=` for components in `maps()`
Please merge this PR for me if everything looks OK, then I'll open a new PR for the `optional str` and `sequence of str` components, and maybe give it a better look by adding some CSS.

mhils:
:cake:

Issue #1293, opened 2016-06-23
Consider binding to localhost by default


colemickens:
I think users might be surprised if their box starts running an open proxy by nature of running `mitmproxy`. I propose changing the default to act like `mitmproxy -b 127.0.0.1` instead.

mhils:
I generally welcome this proposal, but unfortunately this would break transparent mode by default, without providing any useful error message out of the box. 
I see three options:

- When `-T` is specified, we bind to `0.0.0.0` by default. This would probably cater for 95% of the cases, but it's a bit inconsistent.
- Actually bind to `0.0.0.0` and enforce the client IP in mitmproxy. This way we could display an error message in mitmproxy. I don't really like this approach though.
- Decide to keep everything as-is. A lot of services are listening on 0.0.0.0 by default. We may want to explore displaying client ips != localhost more prominently in the UI.

colemickens:
Sorry, I now realize this was the conversation going on in Slack, I didn't pick up on the context properly.

I was just about to say "maybe it could make it obvious it's listening on all devices" and then I realized that it doesn't just say the port in the bottom-right, it actually says "*:9000" or in my case now "127.0.0.1:9000", so I ought to have put it together sooner. (Normally this wouldn't have been an issue thanks to external firewalls, but I just happened to have `mitmproxy` listening on a port that I also use to do testing with various things that I actually need to be able to hit from outside from time to time.)

I don't really like option #2 just from a cleanliness/separation-of-concerns standpoint either and I appreciate the consistency concern in #1.

#3 might help... At least if I'd known those requests came from off box I wouldn't have had the minor panic at seeing unknown traffic (that I thought was originating on box). But it may not be worth the effort to implement or design a new indicator.

Given all this, I guess feel free to close it unless you think option #3 is worth pursuing, and even then it might make sense as a new issue. Thanks for the time and response.

javabrett:
I'm usually all-for conservative approaches when it comes to security, but in this case I think it is reasonable that, in deliberately starting a known proxy-server including the word proxy in its binary name, you should be willing to accept that by-default this will install an open-proxy server.  Hopefully it is sufficient that there is already a mechanism to restrict the address/adapter binding if desired.

My 2c worth.

colemickens:
Agreed. I'm going to close this. Thanks for the thoughts @javabrett and @mhils .

KOLANICH:
[...]

PR #2428, opened 2017-07-05
Bind it to localhost by default

KOLANICH:


Kriechi:
Could you please explain why you made this change?
Just sending a PR without any reason or comment is not very helpful.

KOLANICH:
see #1293 . Will this pr fix it? Does this allow the proxy receive the connections only from localhost?

Kriechi:
Please see https://github.com/mitmproxy/mitmproxy/issues/1293#issuecomment-313208810.

PR #2429, opened 2017-07-06
Fixes bug where loaded certs are stored in the certs dict as str

donald-m-ritter:
Certs provided through the `--cert` argument or config yaml get stored in the `certs` dictionary as a `str`.  This causes issues because the keys created by `get_certs` method are byte strings.  So any provided certs wind up being ignored and the generated certs still get stored as byte strings.

This was found when trying to modify requests while keeping certificate pinning enabled using a second cert I have pinned and have the private key to.  Once I converted the spec to bytes the domain and certificate were correctly identified.


donald-m-ritter:
Closing until I fix tests

PR #2430, opened 2017-07-07
[web] [WIP] Mitmweb options 

MatthewShao:
- [x] Add two columns layout for name-value.
- [x] Adjust the structure of `OptionMaster`.
- [x] Add a basic implementation for `PureStringSequenceOption`.
- [x] Complete validation and updateOption logic for all types.


web/src/js/ducks/options.js
```diff
@@ -21,12 +22,26 @@ export default function reducer(state = defaultState, action) {
...
+    let error = ''
+    fetchApi.put('/options', options).then(
+        (response) => {
+            response.text().then(errorMsg => {
+                error = errorMsg
+                console.log(error)
+            })
+        }
+    )
+    return {type: ERROR, error}
```
MatthewShao: we could send the update request to the backend in the way, but there are two problems:
1. the statement in the promise could not change the value of error.
2. even we connect this in the component, the reducer could not receive this action, therefore the error msg could not be stored in the state.
I'm quite confusing about this, how should we fix this?

web/src/js/components/Modal/OptionMaster.jsx
```diff
@@ -1,20 +1,24 @@
...
 
 PureBooleanOption.PropTypes = {
     value: PropTypes.bool.isRequired,
     onChange: PropTypes.func.isRequired,
 }
 
-function PureBooleanOption({ value, onChange, name, help}) {
+function PureBooleanOption({ value, onChange, ...props}) {
+        let onMouseEnter = () => { props.onMouseEnter() },
+            onMouseLeave = () => { props.onMouseLeave() }
```
mhils: Minor: Is there a particular reason why you don't just pass `props.onMouseEnter` down?
(We can probably also just use `...props` below?)

web/src/js/ducks/options.js
mhils: I _think_ I understand what your problem is, but I'm not sure and I may have misunderstood you. Let me try. :-)

First, we have an async request here, which, once completed should dispatch an action. This doesn't work with plain redux. However, we can use [`redux-thunk`](https://github.com/gaearon/redux-thunk) (we already use at other places) to fix that. I a nutshell, redux-thunk makes 
`dispatch` available in an action creator so that we can use it asynchronously. 😃 

Let me try to sketch a very minimal example:
```javascript
export function update(options) {
    // return a function instead of an action. redux-thunk will just evaluate this right away.
    return dispatch => {
        dispatch({action: OPTION_UPDATE_START, ...});
        fetchApi.put('/options', options).then(response => {
             // maybe do nothing here because get get a state update via websockets anyways?
            dispatch({action: OPTION_UPDATE_SUCCESS, ...});
        }).catch(error => {
            dispatch({action: OPTION_UPDATE_ERROR, err: error.message})
        });
    }
}
```

It probably makes sense to have a new optionEditor reducer that contains the client-side state (the existing option reducer then only handles server-side events):

```javascript
const defaultState = {
   /* optionName -> {isUpdating, value (client-side), error} */
}

export default function reducer(state = defaultState, action) {
    switch (action.type) {

        case OPTION_UPDATE_START:
            return {
                ...state,
                [action.option]: {
                    isUpdating: true,
                    value: action.value,
                    error: false,
                }
            }

        case OPTION_UPDATE_SUCCESS:
            s = {...state};
            del s[action.option];
            return s;

        case OPTION_UPDATE_ERROR:
            return {
                ...state,
                [action.option]: {
                    ...state[action.option],
                    isUpdating: false,
                    error: action.error,
                }
            }

        default:
            return state
    }
}
```

Every row would then connect to both the server-side and the client-side state and render appropriately.
This is all sketched out very quickly and may have some fundamental flaws - does it make any sense? 😃

web/src/js/components/Modal/OptionMaster.jsx
MatthewShao: Yes, we could pass `props.onMouseEnter` and `props.onMouseLeave` here, I was confused about that, but understand how it works now.

MatthewShao:
The new structure is awesome. :tada: :cake: The error msg below the option names looks good: that's more conspicuous than the tooltip. I did some minor changes:
1. sort the options by alphabet order, just like the one in our console version.
2. make the input border turns to red color if its value is wrong.

While I still have a little complain about the `StringSequence` textarea, now we could not start a new line by pressing `Enter`. That's awful when we want to input, for example, a list of ignore hosts... Can we fix this?

mhils:
`ducks/ui/option.js` seems to be a duplicate of  `ducks/ui/optionsEditor.js`?

Other than that, LGTM!

PR #2435, opened 2017-07-10

Kriechi:


mhils:
mypy is handled in 7b0485d6d6d1f3cc5d6704db39971b69dc030b80, leaving cryptography/pyOpenSSL, which is #2456.

PR #2436, opened 2017-07-11
Allow remote Addon

ujjwal96:
This addon decides whether clientconnect should be accepted or not.
If allow_remote option is set to false client will not be able to connect to the proxy unless it is on the same network or the proxyauth option is set. All connections from private IPs will be allowed and remote clients will be able to connect if the proxyauth option is set.
            
ref: https://github.com/mitmproxy/mitmproxy/issues/1293#issuecomment-313208810

### DIFF
--- a/mitmproxy/addons/allowremote.py
+++ b/mitmproxy/addons/allowremote.py
@@ -1,0 +1,29 @@
+import ipaddress
+from mitmproxy import ctx
+
+
+class AllowRemote:
+    def load(self, loader):
+        loader.add_option(
+            "allow_remote", bool, False,
+            """
+            Allow remote clients to connect to proxy. If set to false,
+            client will not be able to connect to proxy unless it is on the same network
+            or the proxyauth option is set
+            """
+        )
+
+    def clientconnect(self, layer):
+        address = layer.client_conn.address
+
+        accept_connection = (
+            ctx.options.allow_remote or
+            ipaddress.ip_address(address[0]).is_private or
+            ctx.options.proxyauth is not None
+        )
+
+        if not accept_connection:
+            layer.reply.kill()
+            ctx.log.warn("Client connection was killed because allow_remote option is set to false, "
+                         "client IP was not a private IP and proxyauth was not set.\n"
+                         "To allow remote connections set allow_remote option to true or set proxyauth option.")
--- a/test/mitmproxy/addons/test_allowremote.py
+++ b/test/mitmproxy/addons/test_allowremote.py
@@ -5,37 +5,34 @@ from mitmproxy.addons import allowremote
 from mitmproxy.test import taddons
 
 
-class TestAllowRemote:
-    @pytest.mark.parametrize("allow_remote, ip, should_be_killed",
-                             [
-                                 (True, "192.168.1.3", False),
-                                 (True, "122.176.243.101", False),
-                                 (False, "192.168.1.3", False),
-                                 (False, "122.176.243.101", True),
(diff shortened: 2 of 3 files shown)


mitmproxy/addons/allowremote.py
```diff
@@ -0,0 +1,16 @@
+import ipaddress
+from mitmproxy import ctx
+
+
+class AllowRemote:
+    def clientconnect(self, layer):
+        address = layer.client_conn.address
+
+        if not (ctx.options.allow_remote or
+                ipaddress.ip_address(address[0]).is_private or
```
Kriechi: `is_private` could be anything that is considered *private* by the IANA. I think this is a bit too much, and we should restrict it to only allow the local network (netmask in IPv4 or link-local address in IPv6).
Thoughts?

mhils: I think that's fine to be honest, the more interesting question is how this behaves with IPv6?

mitmproxy/addons/allowremote.py
```diff
@@ -0,0 +1,16 @@
...
+                ctx.options.proxyauth is not None):
```
mhils: The indentation here is confusing, let's make this a bit more explicit:
```python
accept_connection = (
    # conditions here
)
if not accept_connection:
```

mitmproxy/options.py
```diff
@@ -207,6 +207,14 @@ def __init__(self, **kwargs) -> None:
...
         # Proxy options
         self.add_option(
+            "allow_remote", bool, True,
+            """
+            Allow remote clients to connect to proxy. If set to false,
+            client will not be able to connect to proxy unless it is on the same network
+            or the proxyauth option is set
+            """
+        )
+        self.add_option(
```
mhils: As @Kriechi remarked on Slack, this should be moved into the addon itself. :)

dwfreed:
As mentioned in Slack, rather than making some kind of arbitrary conditions, just allow the user to set a list of allowed networks and addresses.  In other words, just make it a generic IP ACL.

mhils:
Let's add tests first. Generic IP ACL sounds overengineered to me.

test/mitmproxy/addons/test_allowremote.py
```diff
@@ -0,0 +1,78 @@
+from unittest import mock
+
+from mitmproxy.addons import allowremote
+from mitmproxy.test import taddons
+
+
+class TestAllowRemote:
```
Kriechi: Please use `pytest.parametrize` on the two test functions for `allow_remote` and the expected log entries (list of strings or similar).
Ideally we cover both values for `allow_remote` with all important IPs

mhils:
I'm a bit late to the party, but 👍 👍 👍. I like the test a lot.

Kriechi:
Unfortunately we are getting tracebacks in our test suite now.
@ujjwal96 please look into it and send a new PR to fix this. Thanks!

```
test/mitmproxy/tools/web/test_master.py Addon error: Traceback (most recent call last):
  File "...mitmproxy/addons/allowremote.py", line 17, in clientconnect
    address = layer.client_conn.address
AttributeError: 'ClientConnection' object has no attribute 'client_conn'

  File "...mitmproxy/addons/allowremote.py", line 17, in clientconnect
    address = layer.client_conn.address
AttributeError: 'ClientConnection' object has no attribute 'client_conn'

  File "...mitmproxy/addons/allowremote.py", line 17, in clientconnect
    address = layer.client_conn.address
AttributeError: 'ClientConnection' object has no attribute 'client_conn'
```

ref as part of #2468

Issue #2334, opened 2017-05-12
Unable to use upstream proxy and proxy authentification


charlesdhdt:
##### Steps to reproduce the problem:

1. git clone https://github.com/mitmproxy/mitmproxy.git
2. cd mitmproxy
3. ./dev.sh 
4. . venv/bin/activate
5. mitmdump --proxyauth test:test -m upstream:http://10.7.0.225:8080/ --ssl-insecure

On 10.7.0.225, I have a VM that runs mitmproxy :
6. mitmdump --insecure

##### Any other comments? What have you tried so far?
It seems that each time you have a clientdisconnect/clientconnect, you get a new prompt to login (proxyauth). Basically for each domain you have to login to the proxyauth, here test/test. 
I wild guess would be that the information in the header to pass by the upstreamproxy somehow is not compatible with the header auth information 

##### System information
HOST : 
```
Mitmproxy version: 3.0.0 (2.0.0dev0435-0x1c6b33f) 
Python version: 3.6.0
Platform: Linux-4.9.0-0.bpo.2-amd64-x86_64-with-debian-8.7
SSL version: OpenSSL 1.1.0e  16 Feb 2017
Linux distro: debian 8.7 
```

VM (10.7.0.225):
```
Mitmproxy version : 2.0.0 (release version)
Python version : 3.5.2
Platform : Linux-4.4.0-75-generic-x86_64-with-Ubuntu-16.04-xenial
SSL version: OpenSSL 1.0.2g 1 Mar 2016
Linux distro: Ubuntu 16.04 xenial
```

mhils:
Thanks for the report. [This](https://github.com/mitmproxy/mitmproxy/blob/69c5a0b6993f70340a0a4b99159086c638d77103/mitmproxy/addons/proxyauth.py#L64) is faulty, because the mode is now `upstream:...`, not just `upstream`.

charlesdhdt:
I fixed it with just a minor change, show I create a pull request ?

-        return ctx.options.mode in ("regular", "upstream")
+        if ctx.options.mode == "regular" or "upstream" in ctx.options.mode:
+            return True
+        else:
+            return False

mhils:
Go ahead please!

PR #2443, opened 2017-07-17
Fixed compability issue : upstream proxy and proxy auth 

charlesdhdt:
Solved issue #2334


test/mitmproxy/addons/test_proxyauth.py
```diff
@@ -1,8 +1,8 @@
 import binascii
-import ldap3
```
Kriechi: If the tests don't depend on the `ldap3` package any more, please also remove it from the dep-list in `setup.py`.

mhils:
What's the state here?

Issue #2448, opened 2017-07-19
Options: Improve flow_detail and verbosity options.


mhils:
`flow_detail` and `verbosity` are currently specified as being integers. The UX of those options is terrible at the moment: there's neither an indication of allowed values nor an explanation on what those values would do.

```
$ mitmdump --options
# Log verbosity. Type int.
verbosity: 2
```

I would propose to make them somewhat useful choices instead of ints. This can hardly be fixed by extending the help text alone, see the [help text style guidelines](https://github.com/mitmproxy/mitmproxy/blob/374c27feb64b0215a7e94cb9f30c27c84bcc437b/mitmproxy/options.py#L35-L39).

PR #2454, opened 2017-07-20
Minor improvements

mhils:


mitmproxy/options.py
```diff
@@ -450,11 +445,7 @@ def __init__(self, **kwargs) -> None:
         self.add_option(
             "flow_detail", int, 1,
             """
-            The display detail level for flows in mitmdump: 0 (almost quiet) to 3 (very verbose).
-              0: shortened request URL, response status code, WebSocket and TCP message notifications.
-              1: full request URL with response status code
-              2: 1 + HTTP headers
-              3: 2 + full response content, content of WebSocket and TCP messages.
+            Flow display detail level in mitmdump from 0 (almost quiet) to 3 (full message contents).
```
Kriechi: Didn't we just add that because nobody knew *what* is displayed at *which* level?

mhils: Yes, this needs to be properly fixed in #2448. Right now it looks absolutely unreadable in mitmweb as the text is floated. This commit is actually from a few days ago, let me revert this part so that we retain the info until a proper fix is found.

PR #2457, opened 2017-07-20
make --keep-host-header available from command line

kopchik:
Hi! I noticed --keep-host-header was not available from command line and didn't display in help. This trivial patch should fix this.

Cheers,
Alex

### DIFF
--- a/mitmproxy/tools/cmdline.py
+++ b/mitmproxy/tools/cmdline.py
@@ -79,6 +79,7 @@ def common_options(parser, opts):
     opts.make_parser(group, "upstream_auth", metavar="USER:PASS")
     opts.make_parser(group, "proxyauth", metavar="SPEC")
     opts.make_parser(group, "rawtcp")
+    opts.make_parser(group, "keep_host_header")
 
     # Proxy SSL options
     group = parser.add_argument_group("SSL")


mhils:
Thanks for the PR! I assume you are only referring to master? We're currently restructuring how options work, and the majority of the less-often used ones has been moved into the new option editor and can be `--set` on the console. 
In other words, what you are seeing is currently intended (yet not documented) behaviour. If it turns out that this is confusing to users in general, we may need to reconsider our current approach before the next release, so this PR alone is useful feedback. Thoughts?

kopchik:
Thank you for explanation. I see two potential solutions. Or three.

1) Do not hide options. The online doc is good, it explains typical scenarios. So, this leaves built-in help for more experienced users and online docs for beginners.

2). Hide extra options by default, but display them with --extended-help or something.

3) Introduce subcommands and display only quick help by default. But, if user requires, display more details about specific command or mode of operation. Like `openssl help` will display available command, while `openssl s_client --help` will show only small portion of help relevant to this specific command.

PS shall I close this PR?

mhils:
Thanks for the suggestions. I think the primary motivation was to not introduce a million command line switches, but discoverability certainly is a concern. Let's see what @cortesi thinks.
I linked this to our general Options UX issue, so we can close this here. Nonetheless, thanks again, this is very helpfull feedback!

Issue #2450, opened 2017-07-19
Options: Distinguish between user-set value and default.


mhils:
It would be very useful to distinguish between options that have been explicitly set by a user and options with default values. For example, Firefox's about:config prints all user-set options bold:

![image](https://user-images.githubusercontent.com/1019198/28370871-bef202c4-6c9b-11e7-87d9-304a8a34e2f3.png)

This distinction is useful so that we can offer a save button in mitmweb, that saves all user-set options into a config file on disk. 

Steps:

 - [x] See how we can distinguish default from user-set options. Is it just enough to compare `.default` to `.value` or something like this?
 - [x] Indicate state in mitmproxy
 - [x] Indicate state in mitmweb
 - [x] Offer save button?

ujjwal96:
This is already done in mitmproxy. User set options are shown in a different color

MatthewShao:
I'm working on the mitmweb part, basically just compare `.default` and `.value`, maybe adding some judgement about the special cases.

PR #2460, opened 2017-07-21
Add default value suggester in mitmweb options

MatthewShao:
From #2450 , `OptionDefault` component will compare current value and the default value, if they are not the same, it will show the default value.

### DIFF
--- a/mitmproxy/tools/web/app.py
+++ b/mitmproxy/tools/web/app.py
@@ -451,6 +451,22 @@ class Options(RequestHandler):
             raise APIError(400, "{}".format(err))
 
 
+class DumpOptions(RequestHandler):
+    def get(self):
+        self.set_header("Content-Disposition", "attachment; filename=options")
+        self.set_header("Content-Type", "application/octet-stream")
+
+        data = optmanager.save(self.master.options)
+        self.write(data)
+
+    def post(self):
+        try:
+            data = optmanager.parse(self.filecontents)
+            self.master.options.update(**data)
+        except Exception as err:
+            raise APIError(400, "{}".format(err))
+
+
 class Application(tornado.web.Application):
     def __init__(self, master, debug):
         self.master = master
@@ -475,7 +491,8 @@ class Application(tornado.web.Application):
                 FlowContentView),
             (r"/settings", Settings),
             (r"/clear", ClearAll),
-            (r"/options", Options)
+            (r"/options", Options),
+            (r"/options/dump", DumpOptions)
         ]
         settings = dict(
             template_path=os.path.join(os.path.dirname(__file__), "templates"),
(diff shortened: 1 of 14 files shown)


MatthewShao:
Here's a simple demo for this new feature:
peek 2017-07-22 21-18

MatthewShao:
I made the config file write to `~/.mitmproxy/config.yaml`, which is identical to the path in the console. So when we update the options in `mitmweb` and then switch to `mitmproxy`, the dumped options can be loaded automatically.

mitmproxy/tools/web/app.py
```diff
@@ -18,8 +18,11 @@
 from mitmproxy import log
 from mitmproxy import version
 from mitmproxy import optmanager
+from mitmproxy import options
 import mitmproxy.tools.web.master # noqa
 
+CONFIG_PATH = os.path.join(options.CA_DIR, 'config.yaml')
```
mhils: Don't redefine, just import the existing one in `cmdline.py`. :)

mitmproxy/tools/web/app.py
```diff
@@ -475,7 +486,8 @@ def __init__(self, master, debug):
                 FlowContentView),
             (r"/settings", Settings),
             (r"/clear", ClearAll),
-            (r"/options", Options)
+            (r"/options", Options),
+            (r"/options/dump", DumpOptions)
```
mhils: Really no strong opinion, just curious - why "dump" and not "save"?

mitmproxy/tools/web/master.py
```diff
@@ -125,6 +125,12 @@ def run(self):  # pragma: no cover
                     "No web browser found. Please open a browser and point it to {}".format(web_url),
                     "info"
                 )
+        unknown_opts = optmanager.load_paths(app.CONFIG_PATH)
```
mhils: This should already happen by default [here](https://github.com/mitmproxy/mitmproxy/blob/8526ca9e175af4a903348c39928535ae10f7543d/mitmproxy/tools/main.py#L69) if I understand things correctly, is that not the case?

web/src/js/__tests__/ducks/optionsSpec.js
```diff
@@ -49,3 +49,18 @@ describe('sendUpdate', () => {
         ])
     })
 })
+
+describe('dump', () => {
```
mhils: Pressing Escape does not save the file now, does it? I was also thinking about making the modal disappear by clicking on the background. Maybe we actually want to do this in `hideModal`, or OptionModal's `componentWillUnmount`? I totally see the motivation why you wanted to separate this, but that looks it leaves a lot of room for error.

mitmproxy/tools/web/app.py
MatthewShao: I followed the pattern of `DumpFlows` befoe, but seems `SaveOptions` sounds more reasonable here.

mitmproxy/tools/web/master.py
MatthewShao: Did not notice that... This should be removed.

PR #2463, opened 2017-07-23
[sans-io] WebSocket layer

ujjwal96:
Websocket layer using wsproto

### DIFF
--- a/mitmproxy/proxy/protocol2/http.py
+++ b/mitmproxy/proxy/protocol2/http.py
@@ -1,13 +1,16 @@
 import typing
 from warnings import warn
+from unittest import mock
 
 import sys
 
 import h11
+from mitmproxy.net import http
 from mitmproxy.proxy.protocol2 import events, commands, websocket
 from mitmproxy.proxy.protocol2.context import ClientServerContext
 from mitmproxy.proxy.protocol2.layer import Layer
 from mitmproxy.proxy.protocol2.utils import expect
+from mitmproxy.net import websockets
 
 
 class HTTPLayer(Layer):
@@ -22,7 +25,7 @@ class HTTPLayer(Layer):
     def __init__(self, context: ClientServerContext):
         super().__init__(context)
         self.state = self.read_request_headers
-
+        self.flow = mock.Mock("mitmproxy.http.HTTPFlow")
         self.client_conn = h11.Connection(h11.SERVER)
         self.server_conn = h11.Connection(h11.CLIENT)
 
@@ -63,6 +66,7 @@ class HTTPLayer(Layer):
             if self.client_conn.client_is_waiting_for_100_continue:
                 raise NotImplementedError()
 
+            self.flow.request.headers = http.Headers(event.headers)
             self.flow_events[0].append(event)
             self.state = self.read_request_body
             yield from self.read_request_body()  # there may already be further events.
@@ -108,9 +112,9 @@ class HTTPLayer(Layer):
             self.state = self.read_response_body
             yield from self.read_response_body()  # there may already be further events.
         elif isinstance(event, h11.InformationalResponse):
-            if event.status_code == 101:
-                # FIXME: check if this is actually WebSocket
-                child_layer = websocket.WebsocketLayer(self.context, None)
+            self.flow.response.headers = http.Headers(event.headers)
+            if event.status_code == 101 and websockets.check_handshake(self.flow.response.headers):
+                child_layer = websocket.WebsocketLayer(self.context, self.flow)
                 yield from child_layer.handle_event(events.Start())
                 self._handle_event = child_layer.handle_event
                 return
--- a/mitmproxy/proxy/protocol2/test/test_websocket.py
+++ b/mitmproxy/proxy/protocol2/test/test_websocket.py
@@ -1,0 +1,146 @@
+from unittest import mock
+
+import pytest
+
+from . import tutils
+from .. import commands
+from .. import events
+from .. import websocket
+from mitmproxy.test import tflow
+
+
+@pytest.fixture
+def ws_playbook(tctx):
+    tctx.server.connected = True
+    playbook = tutils.playbook(
+        websocket.WebsocketLayer(
+            tctx,
+            tflow.twebsocketflow().handshake_flow
+        )
+    )
+    with mock.patch("os.urandom") as m:
+        m.return_value = b"\x10\x11\x12\x13"
+        yield playbook
+
+
+def test_simple(tctx, ws_playbook):
+    f = tutils.Placeholder()
+
+    assert (
+        ws_playbook
+        << commands.Hook("websocket_start", f)
+        >> events.HookReply(-1, None)
+        >> events.DataReceived(tctx.client, b"\x82\x85\x10\x11\x12\x13Xt~\x7f\x7f")  # Frame with payload b"Hello"
+        << commands.Hook("websocket_message", f)
+        >> events.HookReply(-1, None)
+        << commands.SendData(tctx.server, b"\x82\x85\x10\x11\x12\x13Xt~\x7f\x7f")
+        >> events.DataReceived(tctx.server, b'\x81\x05Hello')  # Frame with payload "Hello"
+        << commands.Hook("websocket_message", f)
+        >> events.HookReply(-1, None)
+        << commands.SendData(tctx.client, b'\x81\x05Hello')
+        >> events.DataReceived(tctx.client, b'\x88\x82\x10\x11\x12\x13\x13\xf9')  # Closing frame
+        << commands.SendData(tctx.server, b'\x88\x82\x10\x11\x12\x13\x13\xf9')
+        << commands.SendData(tctx.client, b'\x88\x02\x03\xe8')
+        << commands.Hook("websocket_end", f)
+        >> events.HookReply(-1, None)
+        >> events.DataReceived(tctx.server, b'\x81\x05Hello')
+        << None
+    )
+
+    assert len(f().messages) == 2
+
+
+def test_server_close(tctx, ws_playbook):
+    f = tutils.Placeholder()
+
+    assert (
+        ws_playbook
+        << commands.Hook("websocket_start", f)
+        >> events.HookReply(-1, None)
+        >> events.DataReceived(tctx.server, b'\x88\x02\x03\xe8')
+        << commands.SendData(tctx.client, b'\x88\x02\x03\xe8')
+        << commands.SendData(tctx.server, b'\x88\x82\x10\x11\x12\x13\x13\xf9')
+        << commands.Hook("websocket_end", f)
+        >> events.HookReply(-1, None)
+        << commands.CloseConnection(tctx.client)
+    )
+
+
+def test_ping_pong(tctx, ws_playbook):
+    f = tutils.Placeholder()
+
+    assert (
+        ws_playbook
+        << commands.Hook("websocket_start", f)
+        >> events.HookReply(-1, None)
+        >> events.DataReceived(tctx.client, b'\x89\x80\x10\x11\x12\x13')  # Ping
+        << commands.Log("info", "Websocket PING received ")
+        >> events.HookReply(-1, None)
+        << commands.SendData(tctx.server, b'\x89\x80\x10\x11\x12\x13')
+        << commands.SendData(tctx.client, b'\x8a\x00')
+        >> events.DataReceived(tctx.server, b'\x8a\x00')  # Pong
+        << commands.Log("info", "Websocket PONG received ")
+        >> events.HookReply(-1, None)
+    )
+
+
+def test_connection_failed(tctx, ws_playbook):
+    f = tutils.Placeholder()
+    assert (
+        ws_playbook
+        << commands.Hook("websocket_start", f)
+        >> events.HookReply(-1, None)
+        >> events.DataReceived(tctx.client, b"Not a valid frame")
+        << commands.SendData(tctx.server, b'\x88\x94\x10\x11\x12\x13\x13\xfb[}fp~zt1}cs~vv0!jv')
+        << commands.SendData(tctx.client, b'\x88\x14\x03\xeaInvalid opcode 0xe')
+        << commands.Hook("websocket_error", f)
+        >> events.HookReply(-1, None)
+        << commands.Hook("websocket_end", f)
+        >> events.HookReply(-1, None)
+    )
+
+
+def test_extension(tctx):
+    f = tutils.Placeholder()
+    tctx.server.connected = True
+    handshake_flow = tflow.twebsocketflow().handshake_flow
+    handshake_flow.request.headers["sec-websocket-extensions"] = "permessage-deflate;"
+    handshake_flow.response.headers["sec-websocket-extensions"] = "permessage-deflate;"
+    playbook = tutils.playbook(websocket.WebsocketLayer(tctx, handshake_flow))
+    with mock.patch("os.urandom") as m:
+        m.return_value = b"\x10\x11\x12\x13"
+        assert (
+            playbook
+            << commands.Hook("websocket_start", f)
+            >> events.HookReply(-1, None)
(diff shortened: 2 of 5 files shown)


mitmproxy/proxy/protocol2/websocket.py
```diff
@@ -16,14 +18,64 @@ def __init__(self, context: ClientServerContext, handshake_flow: http.HTTPFlow):
...
+                source = self.server_conn
+                other = self.client_conn
+                fb = self.server_frame_buffer
+                send_to = self.context.client
 
-    _handle_event = start
+            source.receive_bytes(event.data)
+            ws_event = next(source.events())
+
+            if isinstance(ws_event, wsevents.TextReceived) or isinstance(ws_event, wsevents.BytesReceived):
```
Kriechi: Use `wsevents.DataReceived` here.

mitmproxy/proxy/protocol2/websocket.py
```diff
@@ -16,14 +18,64 @@ def __init__(self, context: ClientServerContext, handshake_flow: http.HTTPFlow):
...
+                send_to = self.context.server
+            else:
+                source = self.server_conn
+                other = self.client_conn
+                fb = self.server_frame_buffer
+                send_to = self.context.client
 
-    _handle_event = start
+            source.receive_bytes(event.data)
+            ws_event = next(source.events())
```
mhils: This could potentially yield more than one event, right?

ujjwal96: Potentially, yes but here's a [list of events](https://github.com/ujjwal96/wsproto/blob/master/wsproto/events.py) that wsproto generates. Since we're not handling the handshake here, only events that should be generated are ConnectionClosed and DataReceived. 
We can have an else block maybe and raise Unexpected event exception?

mhils: No, I mean currently you just take off one event. There might be more than that, e.g. two DataReceived events.

mitmproxy/proxy/protocol2/websocket.py
```diff
@@ -16,14 +19,84 @@ def __init__(self, context: ClientServerContext, handshake_flow: http.HTTPFlow):
...
+                        yield commands.Hook("websocket_message", self.flow)
+
+                    other.send_data(ws_event.data, ws_event.message_finished)
+                    yield commands.SendData(send_to, other.bytes_to_send())
+
+                elif isinstance(ws_event, wsevents.ConnectionClosed):
+                    yield commands.SendData(send_to, source.bytes_to_send())
+                    if isinstance(ws_event, wsevents.ConnectionFailed):
+                        self.flow.error = flow.Error(
+                            "WebSocket connection closed unexpectedly by {}: {}".format("client" if from_client else "server",
```
mhils: Move into new line from "client" on :)

mitmproxy/proxy/protocol2/websocket.py
```diff
@@ -16,14 +19,84 @@ def __init__(self, context: ClientServerContext, handshake_flow: http.HTTPFlow):
...
+                                                                                        ws_event.reason)
+                        )
+                        yield commands.Hook("websocket_error", self.flow)
+
+                    yield commands.Hook("websocket_end", self.flow)
+                    self._handle_event = self.done
+        elif isinstance(event, events.ConnectionClosed):
+            if event.connection == self.context.server:
```
mhils: Is this an expected event according to protocol? My understanding would be that receiving `ConnectionClosed` in `relay_messages` means ungraceful shutdown of one end, as a graceful shutdown would first cause a `wsevents.ConnectionClosed`? In this case, we should add an error here.

mitmproxy/proxy/protocol2/test/test_websocket.py
```diff
@@ -0,0 +1,104 @@
...
+from . import tutils
+from .. import commands
+from .. import events
+from .. import websocket
+from mitmproxy.test import tflow
+
+
+def test_simple(tctx):
+    f = tutils.Placeholder()
+    tctx.server.connected = True
```
mhils: You can make a ws_playbook pytest fixture to simplify your tests:

```python
@pytest.fixture
def ws_playbook(tctx):
    tctx.server.connected = True
    playbook = tutils.playbook(
        websocket.WebsocketLayer(
            tctx, 
            tflow.twebsocketflow().handshake_flow
        )
    )
    with mock.patch("os.urandom") as m:
        m.return_value = b"\x10\x11\x12\x13"
        yield playbook

def test_simple(ws_playbook):
    assert (
        ws_playbook
        << ...
    )
```

mitmproxy/proxy/protocol2/test/test_websocket.py
```diff
@@ -0,0 +1,104 @@
...
+            >> events.DataReceived(tctx.server, b'\x81\x05Hello')  # Frame with payload "Hello"
+            << commands.Hook("websocket_message", f)
+            >> events.HookReply(-1, None)
+            << commands.SendData(tctx.client, b'\x81\x05Hello')
+            >> events.DataReceived(tctx.client, b'\x88\x82\x10\x11\x12\x13\x13\xf9')
+            << commands.SendData(tctx.server, b'\x88\x82\x10\x11\x12\x13\x13\xf9')
+            << commands.SendData(tctx.client, b'\x88\x02\x03\xe8')
+            << commands.Hook("websocket_end", f)
+            >> events.HookReply(-1, None)
+            << commands.CloseConnection(tctx.server)
```
mhils: Ok, it looks like I haven't made the current closing semantics clear yet. It's a bit nonobvious: If a connection is completed, all we want to do is closing the client connection. Once the proxy server (the thing that does IO) notices that the client connection is shut down, it will automatically close all other connections as well. 

If we close a server connection, the client connection will remain open. This has the purpose as e.g. a HTTP/2 connection could have multiple server connections at the same time because a user changed the hostname in a script for one of the requests.

With the current code here, we rely on the client actually closing the connection, otherwise we end up with a stale connection, as both ends (client and proxy) wait for the other end to close. Does that make sense and do you see the problem?

mitmproxy/proxy/protocol2/websocket.py
```diff
@@ -16,14 +19,104 @@ def __init__(self, context: ClientServerContext, handshake_flow: http.HTTPFlow):
...
+        self.server_frame_buffer = []
+        self.extension = None
+        self.make_extension()
+        self.client_conn = wsconn.WSConnection(wsconn.SERVER, wsconn.ConnectionState.OPEN,
+                                               extensions=[self.extension] if self.extension else None)
+
+        self.server_conn = wsconn.WSConnection(wsconn.CLIENT, wsconn.ConnectionState.OPEN,
+                                               extensions=[self.extension] if self.extension else None)
+
+    def make_extension(self):
```
mhils: We can make our code simpler by making this function side-effect free, i.e. let's provide `server_extensions` as an argument and return the extension here instead of storing it as `self.extension`. In principle it's a `@staticmethod` then.

mitmproxy/proxy/protocol2/websocket.py
```diff
@@ -71,7 +80,8 @@ def relay_messages(self, event: events.Event) -> commands.TCommandGenerator:
                 if isinstance(ws_event, wsevents.DataReceived):
                     fb.append(ws_event.data)
                     if ws_event.message_finished:
-                        payload = b''.join(f for f in fb)
+                        payload = b'' if isinstance(ws_event, wsevents.BytesReceived) else ''
+                        payload = payload.join(f for f in fb)
```
mhils: Simpler:
```python
if isinstance(ws_event, wsevents.BytesReceived):
    payload = b"".join(fb)
else:
    payload = "".join(fb)
```

mitmproxy/proxy/protocol2/websocket.py
```diff
@@ -82,19 +92,29 @@ def relay_messages(self, event: events.Event) -> commands.TCommandGenerator:
                     yield commands.SendData(send_to, other.bytes_to_send())
 
                 elif isinstance(ws_event, wsevents.ConnectionClosed):
-                    yield commands.SendData(send_to, source.bytes_to_send())
-                    if isinstance(ws_event, wsevents.ConnectionFailed):
+                    other.close(ws_event.code, ws_event.reason)
+                    yield commands.SendData(send_to, other.bytes_to_send())
+                    # FIXME: Wait for other end to actually send the closing frame
```
mhils: I think it's okay if we document this here and terminate both ends immediately. Makes the code considerably simpler. But yes, good catch. 👍

mitmproxy/proxy/protocol2/test/test_websocket.py
ujjwal96: Should I add it in conftest or in test_websocket itself??

mitmproxy/proxy/protocol2/websocket.py
```diff
@@ -16,14 +19,104 @@ def __init__(self, context: ClientServerContext, handshake_flow: http.HTTPFlow):
...
+                    if param.startswith('client_no_context_takeover'):
+                        args["client_no_context_takeover"] = True
+                    elif param.startswith('server_no_context_takeover'):
+                        args["server_no_context_takeover"] = True
+                    elif param.startswith('client_max_window_bits'):
+                        args["client_max_window_bits"] = int(param.split('=', 1)[1].strip())
+                    elif param.startswith('server_max_window_bits'):
+                        args["server_max_window_bits"] = int(param.split('=', 1)[1].strip())
+
+                self.extension = PerMessageDeflate(**args)
```
Kriechi: I think we should move parts of this into wsproto itself, and provide a simple API (to be used by mitmproxy here) to access these parameters.
Ideally IMO we shouldn't deal with extensions at all, but simply pass the completed handshake back to wsproto to "update" it's internal state.

mitmproxy/proxy/protocol2/websocket.py
```diff
@@ -16,14 +19,104 @@ def __init__(self, context: ClientServerContext, handshake_flow: http.HTTPFlow):
...
+                            "WebSocket connection closed unexpectedly by {}: {}".format(
+                                "client" if from_client else "server",
+                                ws_event.reason
+                            )
+                        )
+                        yield commands.Hook("websocket_error", self.flow)
+
+                    yield commands.Hook("websocket_end", self.flow)
+                    yield commands.CloseConnection(send_to)
+                    self._handle_event = self.done
```
Kriechi: please add a log entry for websocket PING & PONGs (with payload in the log).

mitmproxy/proxy/protocol2/http.py
```diff
@@ -108,9 +112,9 @@ def read_response_headers(self):
             self.state = self.read_response_body
             yield from self.read_response_body()  # there may already be further events.
         elif isinstance(event, h11.InformationalResponse):
-            if event.status_code == 101:
-                # FIXME: check if this is actually WebSocket
-                child_layer = websocket.WebsocketLayer(self.context, None)
+            self.flow.response.headers = http.Headers(event.headers)
+            if event.status_code == 101 and websockets.check_handshake(self.flow.respone.headers):
```
Kriechi: `self.flow.respone`? Does this actually work?
Are the tests not catching this typo?

mitmproxy/proxy/protocol2/websocket.py
```diff
@@ -16,14 +18,98 @@ def __init__(self, context: ClientServerContext, handshake_flow: http.HTTPFlow):
...
+                            ws_event.payload.decode()
+                        )
+                    )
+                    if isinstance(ws_event, wsevents.PingReceived):
+                        other.ping(ws_event.payload)
+                        yield commands.SendData(send_to, other.bytes_to_send())
+                        source.bytes_to_send()  # We don't want send PONG unless we receive it
+                    else:
+                        other.pong(ws_event.payload)
+                        yield commands.SendData(send_to, other.bytes_to_send())
```
Kriechi: If I understand wsproto correctly, it automatically replies to a PING. We only need to `yield commands.SendData(send_to, other.bytes_to_send())`. Calling `other.pong` would create two pongs!

mitmproxy/proxy/protocol2/websocket.py
```diff
@@ -16,14 +18,98 @@ def __init__(self, context: ClientServerContext, handshake_flow: http.HTTPFlow):
...
+                        "info",
+                        "Websocket {} received {}".format(
+                            "PING" if isinstance(ws_event, wsevents.PingReceived) else "PONG",
+                            ws_event.payload.decode()
+                        )
+                    )
+                    if isinstance(ws_event, wsevents.PingReceived):
+                        other.ping(ws_event.payload)
+                        yield commands.SendData(send_to, other.bytes_to_send())
+                        source.bytes_to_send()  # We don't want send PONG unless we receive it
```
Kriechi: I'm not sure what is best here. For h2 the ping+pong is only between endpoints: client<->mitmproxy, and mitmproxy<->server. We do not forward & wait for pings. The WebSocket RFC doesn't mention proxies in the context of pings.

Let's trigger a ping to `other`, and reply immediately with a pong to `source`.
I think we should implement this for h2 as well. See https://github.com/python-hyper/hyper-h2/issues/834

mitmproxy/proxy/protocol2/http.py
ujjwal96: Oops..
There are no tests present for http layer yet. Also, since this is a mock object python doesn't complain about self.flow not having a respone attribute

mitmproxy/proxy/protocol2/websocket.py
ujjwal96: So, what should we do on a PongReceived event? 
If we choose to do nothing then what happens to the unsolicited pongs?

Kriechi:
Nothing, just log them.

mitmproxy/proxy/protocol2/websocket.py
mhils: > Let's trigger a ping to other

I would recommend to just reply with a pong to source, I don't think there's a need to forward that to the server. After all, that'd be a communication channel through mitmproxy that is not logged.

> So, what should we do on a PongReceived event?


The RFC says that "A Pong frame MAY be sent unsolicited.", so let's just do nothing.

mitmproxy/proxy/protocol2/websocket.py
mhils: Good catch. This should be visible in tests right away :-)

ujjwal96: I think it won't create two pongs as we are using up the PONG bytes in line 81

mitmproxy/proxy/protocol2/test/tutils.py
```diff
@@ -34,6 +35,8 @@ def _eq(
             if x.obj is None:
                 x.obj = y
             x = x.obj
+        if isinstance(x, log.LogEntry):
```
mhils: How about we implement `log.LogEntry.__eq__` instead?

mitmproxy/proxy/protocol2/test/test_websocket.py
```diff
@@ -0,0 +1,118 @@
...
+        << commands.SendData(tctx.server, b'\x88\x82\x10\x11\x12\x13\x13\xf9')
+        << commands.SendData(tctx.client, b'\x88\x02\x03\xe8')
+        << commands.Hook("websocket_end", f)
+        >> events.HookReply(-1, None)
+    )
+
+    assert len(f().messages) == 2
+
+
+def test_server_close(tctx):
```
mhils: Why not `ws_playbook` here?

mitmproxy/proxy/protocol2/websocket.py
```diff
@@ -16,14 +18,95 @@ def __init__(self, context: ClientServerContext, handshake_flow: http.HTTPFlow):
...
+                            )
+                        )
+                        yield commands.Hook("websocket_error", self.flow)
+
+                    yield commands.Hook("websocket_end", self.flow)
+                    if not from_client:
+                        yield commands.CloseConnection(self.context.client)
+                    self._handle_event = self.done
+        elif isinstance(event, events.ConnectionClosed):
+            yield commands.Log("error", "Connection closed abnormally")
```
mhils: How about setting `flow.error` here?

mhils:
🎉

PR #2471, opened 2017-07-25
Fixed addon error in test suite

ujjwal96:
Changed address value in `tclient_conn` to a valid IP address
ref: https://github.com/mitmproxy/mitmproxy/pull/2436#issuecomment-317660902


mhils:
We also need to update the mitmweb snapshots, you can ask @MatthewShao on how to do that. :)

PR #2473, opened 2017-07-26
Respect response to CONNECT created in http_connect function in upstream mode

### LINKED ISSUE #2464: Postpone establishing connection with server
Hi, I use mitmproxy in upstream mode as a flow replayer by loading a couple scripts and resource files. It's never intended to deliver a request to remote server. Everything is ok except that, although each of the canned responses (loaded from a file on the disk) is correctly returned to the client, it actually makes connection to remote server on each of the flow via the upstream proxy server I configured in the cmd line, which is totally unnecessary and time-wasting in my senario. 
Would it be possible to postpone the establishing of connection with server (after the request() hook has been revoked) so that my upstream mitmproxy server can run with a dummy -U flag, which don't have to be a real proxy server at all, as it never should really forward a request/establish a connection to remote server?

mengbiping:
Example script that creates response to CONNECT: 

```
from mitmproxy.http import make_connect_response
def http_connect(flow):
  if flow.request.method == 'CONNECT':
    # You may also selectively deny CONNECT request from certain IPs here.
    flow.response = make_connect_response(flow.request.http_version)
```

Fixes #2464

### DIFF
--- a/mitmproxy/proxy/protocol/http.py
+++ b/mitmproxy/proxy/protocol/http.py
@@ -217,16 +217,17 @@ class HttpLayer(base.Layer):
         return False
 
     def handle_upstream_connect(self, f):
-        self.establish_server_connection(
-            f.request.host,
-            f.request.port,
-            f.request.scheme
-        )
-        self.send_request(f.request)
-        f.response = self.read_response_headers()
-        f.response.data.content = b"".join(
-            self.read_response_body(f.request, f.response)
-        )
+        if not f.response:
+            self.establish_server_connection(
+                f.request.host,
+                f.request.port,
+                f.request.scheme
+            )
+            self.send_request(f.request)
+            f.response = self.read_response_headers()
+            f.response.data.content = b"".join(
+                self.read_response_body(f.request, f.response)
+            )
         self.send_response(f.response)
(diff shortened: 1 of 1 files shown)


mitmproxy/proxy/protocol/http.py
```diff
@@ -217,16 +217,17 @@ def handle_regular_connect(self, f):
...
-            f.request.host,
-            f.request.port,
-            f.request.scheme
-        )
-        self.send_request(f.request)
-        f.response = self.read_response_headers()
-        f.response.data.content = b"".join(
-            self.read_response_body(f.request, f.response)
-        )
+        if not f.response:
```
mhils: Add a comment here: 
```
# if the user specifies a response in the http_connect hook, we do not connect upstream here.
# https://github.com/mitmproxy/mitmproxy/pull/2473
```

mhils:
This looks good, thanks. I can't guarantee that this will survive a eager/lazy switch implementation as specified in #2464, but this should do it for now. :-)

mengbiping:
Thank you, @mhils for the review!

Would it possibly be cherrypicked into v2.0.x release? I'd prefer to refer to officially released versions instead of a monkey patch of mine.

PR #2475, opened 2017-07-26

Kriechi:


ujjwal96:
According to pyasn1 [changelog](http://pyasn1.sourceforge.net/changelog.html#revision-0-3-1-released-26-07-2017) 
> The .getComponent*() methods of constructed ASN.1 types changed to lazily instantiate underlying type rather than return None. This should simplify its API as initialization like X[0][1] = 2 becomes possible. WARNING: this change introduces a deviation from the original API.

So, `i[0]` will never be None and `i[1]` will be an instance of `univ.OctetString`

PR #2480, opened 2017-07-28
Fix #2448: Improve flow_detail and verbosity options

### LINKED ISSUE #2448: Options: Improve flow_detail and verbosity options.

MatthewShao:
- [x] Improve verbosity options, change it to `str` type.
- [ ] Improve flow_detail.

### DIFF
--- a/mitmproxy/addons/termlog.py
+++ b/mitmproxy/addons/termlog.py
@@ -20,7 +20,7 @@ class TermLog:
         else:
             outfile = self.outfile or realstdout
 
-        if ctx.options.verbosity >= log.log_tier(e.level):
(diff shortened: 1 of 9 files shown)


mhils:
Nice! This is a great UX improvement. :-)

PR #2484, opened 2017-07-29
Enable tool-specific options

mhils:
This started as a naive attempt to move mitmweb's options into a mitmweb-specific addon, and ended in something slightly more complex:

- **Disentagle `ProxyServer/DummyServer` from the `Master` constructor:** The proxy server should ultimately be an addon itself, so this is the first step towards this by moving it out of the function signature. 
- **Replace usage of `ProxyConfig` with `Options`:** Now that `Master` may not have a server, there also is no `ProxyConfig`, so a whole bunch of tests failed. To fix this, I replaced all usages of `ProxyConfig` with the proper equivalent in `Options`, which should have been the case from in the first place.
- **Actually move mitmweb options into mitmweb-specific addon:** We can now follow up with all other options and move them into their respective addons. I plan to update `Loader.add_option` to not log repeated `add_option` calls with _exactly_ the same arguments (same option name, type, help, ...).
- **Fix mitmproxy's and mitmweb's cmdline** to actually display tool-specific options and not just the mitmdump ones.

### DIFF
--- a/mitmproxy/master.py
+++ b/mitmproxy/master.py
@@ -10,6 +10,7 @@ from mitmproxy import exceptions
 from mitmproxy import command
 from mitmproxy import http
 from mitmproxy import log
+from mitmproxy.net import server_spec
 from mitmproxy.proxy.protocol import http_replay
 from mitmproxy.types import basethread
 
@@ -32,16 +33,25 @@ class Master:
     """
         The master handles mitmproxy's main event loop.
     """
-    def __init__(self, opts, server):
-        self.options = opts or options.Options()
+    def __init__(self, opts):
+        self.options = opts or options.Options()  # type: options.Options
         self.commands = command.CommandManager(self)
         self.addons = addonmanager.AddonManager(self)
         self.event_queue = queue.Queue()
         self.should_exit = threading.Event()
-        self.server = server
+        self._server = None
         self.first_tick = True
-        channel = controller.Channel(self.event_queue, self.should_exit)
-        server.set_channel(channel)
+
+    @property
+    def server(self):
+        return self._server
+
+    @server.setter
+    def server(self, server):
+        server.set_channel(
+            controller.Channel(self.event_queue, self.should_exit)
+        )
+        self._server = server
 
     @contextlib.contextmanager
     def handlecontext(self):
@@ -71,7 +81,8 @@ class Master:
 
     def start(self):
         self.should_exit.clear()
-        ServerThread(self.server).start()
+        if self.server:
+            ServerThread(self.server).start()
 
     def run(self):
         self.start()
@@ -101,7 +112,8 @@ class Master:
         return changed
 
     def shutdown(self):
-        self.server.shutdown()
+        if self.server:
+            self.server.shutdown()
         self.should_exit.set()
         self.addons.trigger("done")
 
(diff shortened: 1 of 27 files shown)


mitmproxy/master.py
```diff
@@ -32,16 +33,25 @@ class Master:
...
+    @property
+    def server(self):
+        return self._server
+
+    @server.setter
+    def server(self, server):
+        server.set_channel(
+            controller.Channel(self.event_queue, self.should_exit)
+        )
+        self._server = server
```
mhils: This is a bit ugly, but the plan is to ultimately get rid of `master.server` anyways, so...

mitmproxy/tools/web/webaddons.py
```diff
@@ -0,0 +1,18 @@
+class WebOptions:
```
Kriechi: File name does not represent class name... if we ever cared for such things?

mitmproxy/tools/web/webaddons.py
```diff
@@ -0,0 +1,18 @@
...
+            "web_debug", bool, False,
+            "Mitmweb debugging."
+        )
+        loader.add_option(
+            "web_port", int, 8081,
+            "Mitmweb port."
+        )
+        loader.add_option(
+            "web_iface", str, "127.0.0.1",
+            "Mitmweb interface."
```
Kriechi: IMHO `mitm*` should always be spelled with lower case letters only!
Feel free to disagree.

mitmproxy/tools/web/webaddons.py
mhils: This is to stay consistent with https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/tools/console/consoleaddons.py.

mitmproxy/tools/web/webaddons.py
mhils: Just copied here, but I'm with you. 😄

test/filename_matching.py
```diff
@@ -22,7 +22,7 @@ def check_src_files_have_test():
 def check_test_files_have_src():
     unknown_test_files = []
 
-    excluded = ['test/mitmproxy/data/', 'test/mitmproxy/net/data/', '/tservers.py']
+    excluded = ['test/mitmproxy/data/', 'test/mitmproxy/net/data/', '/tservers.py', '/conftest.py']
```
Kriechi: did `conftest.py` actually show up for you?
`test_files` should never contain it...

test/mitmproxy/proxy/protocol/test_http2.py
```diff
@@ -508,8 +505,10 @@ def handle_server_event(cls, event, h2_conn, rfile, wfile):
         return True
 
     def test_body_size_limit(self):
-        self.config.options.body_size_limit = "20"
-        self.config.options._processed["body_size_limit"] = 20
+        self.options.body_size_limit = "20"
+
+        # FIXME: This should not be required?
+        self.options._processed["body_size_limit"] = 20
```
Kriechi: No idea - if this code is from me, I probably just copied it from somewhere without knowing what it is doing...

test/filename_matching.py
mhils: Yes, conftest.py in the console subdirectory did. I later on removed it because it's not required anymore with the new location of the isatty check, but figured that this would be useful to leave.

mitmproxy/tools/web/webaddons.py
Kriechi: how is this consistent?
`consoleaddons.py` contains a `ConsoleAddon` class.
`webaddons.py` contains a `WebOptions` class.

test/filename_matching.py
Kriechi: ok, let's keep it.

mitmproxy/tools/web/webaddons.py
mhils: Uh... I guess you have a point here. I'll fix it. 😄

test/mitmproxy/proxy/protocol/test_http2.py
ujjwal96: This is from https://github.com/mitmproxy/mitmproxy/commit/201c65960eb19b97e6c499f22082aa176bdaa6b1#diff-ae3ce6b83d0e48d6158f2e5fe6583e6b
we aren't using the core addon which probably sets it

mhils: Thanks for pointing to that! I think the proper solution here is to cache parsing results, e.g. just by annotating the parse function with `functools.lru_cache`. Anyways, that's for another PR.

PR #2489, opened 2017-08-01
Move console-related options into their own addon

MatthewShao:


mitmproxy/tools/console/consoleaddons.py
```diff
@@ -11,6 +11,29 @@
 from mitmproxy.tools.console import signals
 from mitmproxy.tools.console import keymap
 
+# We redefine these here for now to avoid importing Urwid-related guff on
+# platforms that don't support it, and circular imports. We can do better using
+# a lazy checker down the track.
```
Kriechi: Is this comment still true & valid?

MatthewShao: I searched through the project and did not find any duplicated definition... So no more **redefinition** here and I guess this comment should be removed.

mitmproxy/test/taddons.py
```diff
@@ -105,6 +105,8 @@ def configure(self, addon, **kwargs):
             Options object with the given keyword arguments, then calls the
             configure method on the addon with the updated value.
         """
+        loader = addonmanager.Loader(self.master)
+        self.master.addons.invoke_addon(addon, "load", loader)
```
mhils: 👍👍👍, but that repeatedly calls "load", which maybe isn't optimal. How about
```python
if addon not in self.master.addons:
    self.master.addons.register(addon)
```
(this requires us to implement AddonManager.__contains__, which should be easy though)

test/mitmproxy/tools/console/test_statusbar.py
```diff
@@ -13,14 +13,12 @@ def test_statusbar(monkeypatch):
         stickycookie="~dst example.com",
         stickyauth="~dst example.com",
         default_contentview="javascript",
-        console_order="url",
```
mhils: Let's add the addon here, because these actually should be set to have better coverage of the statusbar :)

test/mitmproxy/addons/test_view.py
```diff
@@ -546,11 +549,13 @@ def test_configure():
...
         with pytest.raises(Exception, match="Unknown flow order"):
-            tctx.configure(v, console_order="no")
+            tctx.configure(v, view_order="no")
 
-        tctx.configure(v, console_order_reversed=True)
+        tctx.configure(v, view_order_reversed=True)
 
+        console_addon = consoleaddons.ConsoleAddon(tctx.master)
+        tctx.configure(console_addon)
         tctx.configure(v, console_focus_follow=True)
```
mhils: How do you feel about `console_focus_follow` -> `view_focus_follow` and then redeclaring the option in the eventlog addon? Ideally addons don't depend on other addons. :)

MatthewShao: Sorry, but I can't catch what you mean here, do you mean redeclaring `view_focus_follow` in `consoleaddons`?
This is the case where same option is used by two different addons... How should we deal with this option?

mhils: I think `view_focus_follow` is used by the eventlog addon and the view addon. Maybe just keep it as `console_focus_follow`, but let's have 
```python
        loader.add_option(
            "view_order_reversed", bool, False,
            "Reverse the sorting order."
        )
```
in both addons? alternatively, we could add a `consoleaddons.common_options(loader)` function which we invoke from both `load` calls to express the dependency. In any case, that should solve the issue that we need to add the ConsoleAddon here and it's clear what the addon depends on. What do you think?

MatthewShao: >         loader.add_option(
            "view_order_reversed", bool, False,
            "Reverse the sorting order."
        )

Did you really mean `view_order_reversed` here or `view_focus_follow` ? Because I can't see any usage of `view_order_reversed` in eventlog addon :sweat_smile:

mhils: copy paste error 😇. I meant `view_focus_follow`/`console_focus_follow`, whatever we want to call it. If we have it in consoleaddons.common_options, then probably the latter. :)

MatthewShao:
It seems that importing `consoleaddons` in `addons/view.py` will fail the build process.... so `consoleaddons.common_options` does not work, maybe we have to add the option in two different addons.

mhils:
Looks good! You need to add a test for `__contains__` to the addonmanager tests, as we enforce individual coverage there. :)

PR #2494, opened 2017-08-02
Improve intercept toggle

ujjwal96:
This PR does 2 things - 
* Renamed intercept_toggle command to console.intercept.toggle to be consistent with existing commands  
* Update options when <kbd>I</kbd> is pressed. Currently we need to press <kbd>I</kbd> and <kbd>Enter</kbd> to toggle


Kriechi:
What happens to all the other instances of `intercept_active`?

ujjwal96:
I didn't change the option name. Just the command

Issue #2498, opened 2017-08-03
Options: Provide good upgrade experience from 2.0.x


mhils:
##### Steps to reproduce the problem:

1. Existing mitmproxy users updates to 3.0 once it is released.
2. User invokes `mitmproxy -R http://example.com`
3. Mitmproxy returns cryptic "Invalid replacement specifier:" error.

##### Any other comments? What have you tried so far?

The changes to the CLI in 3.0 are quite big, so we should try to provide a somewhat reasonable migration experience for existing users. As I see it, it would make sense to have a very dirty function that checks `argv` for common migration errors. We should at least provide useful hints for the following issues:

- [x] `-T`
- [x] `-R` (now repurposed, remove that?)
- [x] `-U` 
- [x] `--http2` / `--no-http2` (#2500)

Again, implementation of this can be dirty, it's just useful to have for a while to ease migration.

ujjwal96:
Options which mean a different thing in 3.0
----
- R (reverse mode) -> R (replacements)
- k (replay-kill-extra) -> k (ssl-insecure)

Options which are removed - 
----
- cadir
- Z, body-size-limit
- stream
- palette
- palette-transparent
- e, eventlog
- follow
- order
- no-mouse
- reverse
- socks
- http2-priority
- np-http2-priority
- no-websocket
- websocket
- spoof-source-address
- upstream-bind-address
- ciphers-client 
- ciphers-server
- client-certs 
- no-upstream-cert
- add-upstream-certs-to-client-chain
- upstream-trusted-cadir
- upstream-trusted-ca
- ssl-version-client
- ssl-version-server
- no-onboarding
- onboarding-host
- onboarding-port
- server-replay-use-header
- norefresh
- no-pop
- replay-ignore-content
- replay-ignore-payload-param
- replay-ignore-param
- replay-ignore-host
- replace-from-file
- nonanonymous
- singleuser
- htpasswd

Options that are invoked differently -
----
- t -> stickycookie
- u -> stickyauth
- wfile -> save-stream-file
- a, afile -> Use -w and prefix path with + to append
- z -> anticomp
- T -> m transparent
- U -> m upstream:SPEC
- b, bind-address -> listen-host
- port -> listen-port
- I, ignore -> ignore-hosts
- tcp -> tcp-hosts
- cert -> certs
- insecure -> ssl-insecure, k
- c -> C (client-replay)
- replace -> replacements
- i -> intercept
- f, filter -> view-filter

[...]

PR #2503, opened 2017-08-05
Improve options UX

ujjwal96:
More checks can be added to this.
Almost all the options which will be removed can be set using `--set option_name=value` Maybe add a check for that?
ref: #2498

### DIFF
--- a/mitmproxy/utils/arg_check.py
+++ b/mitmproxy/utils/arg_check.py
@@ -1,0 +1,148 @@
+import sys
+
+DEPRECATED = """
+--cadir
+-Z
+--body-size-limit
+--stream
+--palette
+--palette-transparent
+--follow
+--order
+--no-mouse
+--reverse
+--socks
+--http2-priority
+--no-http2-priority
+--no-websocket
+--websocket
+--spoof-source-address
+--upstream-bind-address
+--ciphers-client
+--ciphers-server
+--client-certs
+--no-upstream-cert
+--add-upstream-certs-to-client-chain
+--upstream-trusted-cadir
+--upstream-trusted-ca
+--ssl-version-client
+--ssl-version-server
+--no-onboarding
+--onboarding-host
+--onboarding-port
+--server-replay-use-header
+--no-pop
+--replay-ignore-content
+--replay-ignore-payload-param
+--replay-ignore-param
+--replay-ignore-host
+--replace-from-file
+"""
+
+REPLACED = """
+-t
+-u
+--wfile
+-a
+--afile
+-z
+-b
+--bind-address
+--port
+-I
+--ignore
+--tcp
+--cert
+--insecure
+-c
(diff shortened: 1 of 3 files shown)


mitmproxy/addons/replace.py
```diff
@@ -38,7 +38,9 @@ def parse_hook(s):
         patt, a, b = parts
     else:
         raise exceptions.OptionsError(
-            "Invalid replacement specifier: %s" % s
+            "Invalid replacement specifier: {}\n"
+            "If you want to use mitmproxy in reverse proxy mode. "
+            "Use --mode reverse:SPEC instead".format(s)
```
mhils: This seems to be the wrong place to me. I think we should rather check for "-R" followed by something that starts with "http"  in main.py so that we don't spread out those legacy checks over the whole codebase.

ujjwal96: Right now, I'm calling the check function only if parse_args() throws an exception.  "-R" won't throw an exception in parse_args. So should we just call check function every time before calling parse_args?

mhils: Let's just check that one particular case before calling parse_args. :)

mitmproxy/utils/arg_check.py
```diff
@@ -0,0 +1,61 @@
...
+--palette-transparent
+-e
+--eventlog
+--follow
+--order
+--no-mouse
+--reverse
+--socks
+--http2-priority
+--np-http2-priority
```
mhils: typo

mitmproxy/utils/arg_check.py
```diff
@@ -0,0 +1,61 @@
...
+--norefresh
+--no-pop
+--replay-ignore-content
+--replay-ignore-payload-param
+--replay-ignore-param
+--replay-ignore-host
+--replace-from-file
+--nonanonymous
+--singleuser
+--htpasswd
```
mhils: Those three are now `--proxyauth`

mitmproxy/utils/arg_check.py
```diff
@@ -0,0 +1,61 @@
...
+    if "-U" in args:
+        print("-U is deprecated, please use --mode upstream:SPEC instead")
+
+    if "-T" in args:
+        print("-T is deprecated, please use --mode transparent instead")
+
+    if any([i in args for i in DEPRECATED.splitlines()]):
+        print("Looks like you used a flag which is deprecated.\n"
+              "To set an option please use `--set option_name=value` "
+              "To show all options and their default values use --options")
```
mhils: Imagine someone calling this with four, five options. This would be more useful if we call out the actual flag that has been removed. Just make it a for-loop and print for each affected option? :)

ujjwal96: Updated

PR #2504, opened 2017-08-07
[WIP] [web] Static viewer for mitmweb

MatthewShao:
ref: https://github.com/mitmproxy/mitmweb/issues/23 -> mhils' comment
- [x] Add `static.js` to `index.html`
- [x] Add `StaticBackend` beside `WebSocketBackend`
- [x] Create `<HideInStatic>` component
- [x] Hide some components in  static mode

For now, we hide following responsive components:
- `<ConnectionIndicator/>`
- `Protocol Support` menu in Options Header
- `Http Options menu` in Options Header
- `Use Host Header` toggler in Options Header
- `Flow Modification menu` in Flow Header
- `Interception menu` in Flow Header
Also:
- set proxy server information label to `Static` in static mode.

And there are some components which I'm not sure whether we should hide in static mode:
- Option Modal
- `Install Certificates Option` in File Menu
- Eventlog: This probably should be kept in static mode, however we need to have some designs to store the log information in a file, maybe just in the `flow` file.


web/src/js/backends/static.js
```diff
@@ -0,0 +1,54 @@
...
+       this.activeFetches[resource] = queue
+       fetchApi(`/${resource}`)
+           .then(res => res.json())
+           .then(json => {
+               if (this.activeFetches[resource] === queue)
+                   this.receive(resource, json)
+           })
+   }
+
+   onMessage(msg) {
```
mhils: I think this (and `activeFetches` in general) can be removed if we don't have WebSockets anymore?

web/src/js/components/Footer.jsx
```diff
@@ -49,11 +49,13 @@ function Footer({ settings }) {
                 <span className="label label-success">stream: {formatSize(stream_large_bodies)}</span>
             )}
             <div className="pull-right">
-                {server && (
+                { MITMWEB_STATIC ?
```
mhils: Use `<HideInStatic>` here?

web/src/js/components/common/HideInStatic.jsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react'
+
+export default function HideInStatic({className, children }) {
+    return MITMWEB_STATIC ? null : ( <div className={className}>{children}</div> )
```
mhils: Shall we try to update to React 16 beta right away so that we do not have to use the wrapper div here? This would allow us to do
```jsx
<HideInStatic>
    <p>foo</p>
    <p>bar</p>
</HideInStatic>
```

See https://github.com/facebook/react/issues/3415, https://github.com/facebook/react/issues/2127, https://github.com/facebook/react/issues/10294

web/src/js/backends/static.js
MatthewShao: For now, the only way to load flows into the static view is to use the `open` option in the `FileMenu`, and then we could refresh the page to see the flows. So, we need the whole `activeFetches` thing to update the flows view, is that unnecessary if we have the `flow` -> static converter?

web/src/js/backends/static.js
```diff
@@ -0,0 +1,36 @@
...
+ * This backend uses the REST API only to host static instances,
+ * without any Websocket connection.
+ */
+import { fetchApi } from "../utils"
+
+const CMD_RESET = 'reset'
+
+export default class StaticBackend {
+   constructor(store) {
+       this.activeFetches = {}
```
mhils: minor leftover

web/src/js/components/Header/OptionMenu.jsx
```diff
@@ -17,6 +19,9 @@ export default function OptionMenu() {
                 </div>
                 <div className="menu-legend">Protocol Support</div>
             </div>
+            </HideInStatic>
+
+            <HideInStatic>
```
mhils: We can remove these three lines :)

web/src/js/components/common/HideInStatic.jsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react'
+
+export default function HideInStatic({ children }) {
+    return global.MITMWEB_STATIC ? null : [children]
```
mhils: `children` instead of `[children]` doesn't work?

web/src/js/__tests__/components/Header/__snapshots__/FileMenuSpec.js.snap
```diff
@@ -72,17 +72,9 @@ exports[`FileMenu Component should render correctly 1`] = `
         />
          Options
       </a>
-       
-    </li>
-    <li>
```
mhils: I'll bite... any idea why there is a diff here? 😄

web/src/js/components/common/HideInStatic.jsx
MatthewShao: yeah, the new feature has to return a array of components.

web/src/js/__tests__/components/Header/__snapshots__/FileMenuSpec.js.snap
MatthewShao: I think this is because we just upgraded the `react-test-renderer`, some rendering algorithm might have been updated, so it makes a diff... This file is generated automatically by `jest`, so I believe that the diff is innocuous.

mhils: FWIW I don't think it's the rendering algorithm - looks more like a bug to me. There clearly should be a `<li>` here generated by this: https://github.com/mitmproxy/mitmproxy/blob/6d9b5b0c062126d2ac47fa50795c7d560321e0f7/web/src/js/components/common/Dropdown.jsx#L49

Does the file menu still look the same? If so, I think we can tolerate this for now.

MatthewShao: yeah, don't worry, it looks totally the same.

PR #2505, opened 2017-08-07
fix #2470

### LINKED ISSUE #2470: Transitive import of mitmproxy.version causes warning
Since #1837, we import `.script`, will imports `.flow`, which imports `.version`.
This causes the following warning in pytest:

```
test/mitmproxy/test_version.py::test_version
  /Users/kriechi/.pyenv/versions/3.5.3/lib/python3.5/runpy.py:125: 
RuntimeWarning: 'mitmproxy.version' found in sys.modules after import of package 
'mitmproxy', but prior to execution of 'mitmproxy.version'; this may result in 
unpredictable behaviour
    warn(RuntimeWarning(msg))

-- Docs: http://doc.pytest.org/en/latest/warnings.html
```

[Note](http://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html#the-double-import-trap)
> This next trap exists in all current versions of Python, including 3.3, and can be summed up in the following general guideline: “Never add a package directory, or any directory inside a package, directly to the Python path”.

> The reason this is problematic is that every module in that directory is now potentially accessible under two different names: as a top level module (since the directory is on sys.path) and as a submodule of the package (if the higher level directory containing the package itself is also on sys.path).

Maybe using the approach described [here](https://stackoverflow.com/questions/27947639/how-to-properly-create-a-pyinstaller-hook-or-maybe-hidden-import) works better?

Kriechi:


Kriechi:
@mhils: warning still present: https://travis-ci.org/mitmproxy/mitmproxy/jobs/261727626#L1873
or did you have a different patch in mind?

mhils:
I would have done it without `exec`, but in any case that should not register mitmproxy.version. Could it maybe be https://github.com/mitmproxy/mitmproxy/blob/380ff50e57a56d2ff25f04350341ae013dcd2443/test/mitmproxy/test_version.py#L7 ?

PR #2510, opened 2017-08-10
[WIP][web]Static viewer converter for mitmweb

MatthewShao:
This PR is trying to add a static viewer converter, which can convert the dumped `flows` file into a group of static file, so that we could view the flows without backend server support.

This is about the first step, copying the resource files to a destination path. Also modifying `index.html` a bit, so we can have the interface by opening this html file.

If I understand the thing correctly, the next step is about converting the `flows` file into a bunch of `json` files in certain structure, so that our frontend could fetch the flow information from those local files.

However, there are two troubles here as far as I can see:

1. The `fetch` api we used in `StaticBackend` does not support the `file:///` protocol, ref: https://github.com/github/fetch/pull/92, I tried `XMLHttpRequest` too, but it only works in Firefox, Chrome prevents us to fetch from local file. We need to change the design of `StaticBackend`, maybe importing the json files directly from our js code.
2. Since we will mimic the API calls in a folder structure of local files, `updatesStoreFromUrl` and `updateUrlFromStore` can not work as excepted, we need to find some workarounds for them.


mhils:
> The fetch api we used in StaticBackend does not support the file:/// protocol, ref: 

Doesn't matter for the final use case. For testing we can just start a webserver (`python3 -m http.server` or `devd .`), ultimately those files would be uploaded to our testing server anyways.

> Since we will mimic the API calls in a folder structure of local files, updatesStoreFromUrl and updateUrlFromStore can not work as excepted, we need to find some workarounds for them.

Good catch. AFAIK they currently work on the location hash though, so this should just work. 😃 

Other than that, this is going in the right direction. Let's continue with the flows! :)

web/src/js/flow/utils.js
```diff
@@ -49,7 +49,14 @@ export var MessageUtils = {
         } else if (message === flow.response) {
             message = "response";
         }
-        return `/flows/${flow.id}/${message}/content` +  (view ? `/${view}` : '');
+        if (global.MITMWEB_STATIC) {
```
mhils: Let's make `.json` the default and not have two different cases. :)

web/src/js/utils.js
```diff
@@ -81,7 +81,21 @@ function getCookie(name) {
 }
 const xsrf = `_xsrf=${getCookie("_xsrf")}`;
 
+
 export function fetchApi(url, options={}) {
+    if (global.MITMWEB_STATIC) {
+        let path = url.split('/'),
```
mhils: Same here. Let's add ".json" to wherever `fetchApi` is called. :) No need to do it within here.

web/src/js/flow/utils.js
MatthewShao: Our backend could not redirect the url like `/flows.json` to `/flows`, it will return a 404. Is there any setting that can help us with this?

mhils: Maybe we can just add `(?:\.json)?` to our routes, e.g. [here](https://github.com/mitmproxy/mitmproxy/blob/96c8c75ce6b8c656e5f03312479c32082d5a89b7/mitmproxy/tools/web/app.py#L472):

```python
(r"/flows(?:\.json)?", Flows),
```
Thoughts? 😃

MatthewShao: Yeah, we could do that and have `.json` in all requested url. 
AFAIK, `GET` operation on a `.json` url is natural, but `POST` or `PUT` operation on a `.json` url sounds so wired. And I'm not sure whether it meets the guideline of RESTful API design...

mhils: I strongly agree - hence the regex to make `.json` optional, see https://www.debuggex.com/r/TVc9q4VcGlGO6znM. We can do `GET /flows.json`, but still do `POST /flows`. The static viewer only needs `GET`, so this covers our use case.

mitmproxy/tools/web/static_viewer.py
```diff
@@ -54,7 +54,7 @@ def save_flows_content(self) -> None:
                 path = os.path.join(self.path, 'flows', f.id, m)
                 if not os.path.exists(path):
                     os.makedirs(path)
-                with open(os.path.join(path, 'content.json'), 'wb') as content_file:
+                with open(os.path.join(path, '_content'), 'wb') as content_file:
```
MatthewShao: Change `content.json` to `_content` here, or it will be weired to access `<flow-id>/response/content.json` to get a picture file. We could not specify the download file name for the user in static mode now, so use this temporary solution.

MatthewShao:
If there is anything that I'm not satisfied with this implementation yet, that should be the speed. We are doing a lot of IO operations here, basically creating a mass of static files, so it is kind of slow... It takes over 30s to process a `flows` file which is only 2MB on my computer. How could we improve this?
I will write the tests for this addon once we nail down the design of it.

mitmproxy/addons/static_viewer.py
```diff
@@ -0,0 +1,93 @@
...
+        shutil.rmtree(str(path / "static"))
+    shutil.copytree(str(web_dir / "static"), str(path / "static"))
+    shutil.copyfile(str(web_dir / 'templates' / 'index.html'), str(path / "index.html"))
+
+    with open(web_dir / "static" / "static.js", "w") as f:
+        f.write("MITMWEB_STATIC = true;")
+
+
+def save_filter_help(path: pathlib.Path) -> None:
+    with open(path / 'filter-help.json', 'w') as f:
```
MatthewShao: > pathlib integrates seemlessly with open only in Python 3.6 and later

so we need to make it `str` type to work for both 3.5 and 3.6. ref:https://stackoverflow.com/questions/42694112/when-using-pathlib-getting-error-typeerror-invalid-file-posixpathexample-t

mitmproxy/addons/static_viewer.py
```diff
@@ -0,0 +1,93 @@
...
+def save_flows(path: pathlib.Path, flows: typing.Iterable[flow.Flow]) -> None:
+    with open(path / 'flows.json', 'w') as f:
+        json.dump(
+            [flow_to_json(f) for f in flows],
+            f
+        )
+
+
+def save_flows_content(path: pathlib.Path, flows: typing.Iterable[flow.Flow]) -> None:
+    for f in flows:
```
MatthewShao: The variable `f` has a collision with the filetype var on line 51, let's make it `flow` here.

mitmproxy/addons/__init__.py
```diff
@@ -39,6 +40,7 @@ def default_addons():
         script.ScriptLoader(),
         serverplayback.ServerPlayback(),
         setheaders.SetHeaders(),
+        static_viewer.StaticViewer(),
```
MatthewShao: Adding `StaticViewer` here as default addons makes the building process fail, because we are importing `mitmproxy.tools.web.app.flow_to_json` as dependency. Maybe we should load it only in `WebMaster`?

PR #2509, opened 2017-08-09
Fixes #2495

ujjwal96:
Fixes #2495

PR #2513, opened 2017-08-10
Parse Client hello from buffer

ujjwal96:


mhils:
Looks good. 👍 

Can we please copy `get_client_hello` into `proxy2/layers/tls.py` so that `mitmproxy/proxy/protocol/tls.py` remains completely untouched? sans-io is still a bit away from getting into master, so keeping merges/rebases simple is crucial. The code can also be simplified if it only has to work with bytes.

Kriechi:
Please rebase / cherry-pick to get #2509

mhils:
Merged, thanks!

PR #2520, opened 2017-08-13
[WIP] Use wsproto for websockets

ujjwal96:
- [x] Fix FIXMEs
- [x] Add tests
- [x] Fix printing websocket logs

### DIFF
--- a/mitmproxy/net/websockets/utils.py
+++ b/mitmproxy/net/websockets/utils.py
@@ -3,11 +3,10 @@ Collection of WebSocket protocol utility functions (RFC6455)
 Spec: https://tools.ietf.org/html/rfc6455
 """
 
+
 import base64
 import hashlib
 import os
-
-from wsproto.extensions import PerMessageDeflate
 
 from mitmproxy.net import http
 from mitmproxy.utils import strutils
@@ -89,26 +88,3 @@ def get_client_key(headers):
 
 def get_server_accept(headers):
     return headers.get("sec-websocket-accept", None)
-
-
-def make_extension(extension):
-    name, *params = extension.split(';')
-    if name == 'permessage-deflate':
-        args = {
-            "client_no_context_takeover": False,
-            "server_no_context_takeover": False,
-            "client_max_window_bits": None,
-            "server_max_window_bits": None
-        }
-
-        for param in params:
-            if param.startswith('client_no_context_takeover'):
-                args["client_no_context_takeover"] = True
-            elif param.startswith('server_no_context_takeover'):
-                args["server_no_context_takeover"] = True
-            elif param.startswith('client_max_window_bits'):
-                args["client_max_window_bits"] = int(param.split('=', 1)[1].strip())
-            elif param.startswith('server_max_window_bits'):
-                args["server_max_window_bits"] = int(param.split('=', 1)[1].strip())
-
-        return PerMessageDeflate(**args)
--- a/mitmproxy/net/websockets/utils.py
+++ b/mitmproxy/net/websockets/utils.py
@@ -3,10 +3,11 @@ Collection of WebSocket protocol utility functions (RFC6455)
 Spec: https://tools.ietf.org/html/rfc6455
 """
 
-
 import base64
 import hashlib
 import os
+
+from wsproto.extensions import PerMessageDeflate
 
 from mitmproxy.net import http
 from mitmproxy.utils import strutils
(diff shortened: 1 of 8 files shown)


mitmproxy/proxy/protocol/websocket.py
```diff
@@ -111,8 +165,9 @@ def _handle_connection_closed(self, event, source_conn, other_conn, is_server):
 
         print(self.connections[other_conn])
         self.connections[other_conn].close(event.code, event.reason)
+        other_conn.send(self.connections[other_conn].bytes_to_send())
+        source_conn.send(self.connections[source_conn].bytes_to_send())
```
ujjwal96: Immediately sending close frame to other end because after we call `self.connections[other_conn].close(event.code, event.reason)` it won't generate any more events.

test/mitmproxy/proxy/protocol/test_websocket.py
```diff
@@ -272,19 +268,22 @@ def test_pong(self):
         frame = websockets.Frame.from_file(self.client.rfile)
         assert frame.header.opcode == websockets.OPCODE.PONG
         assert frame.payload == b'foobar'
+        from time import sleep
+        sleep(0.5)
+        assert self.master.has_log("Pong Received from server", "info")
```
ujjwal96: Better way for this?
Without this it sometimes returns False

mitmproxy/proxy/protocol/websocket.py
```diff
@@ -54,7 +52,9 @@ def __init__(self, ctx, handshake_flow):
         extensions = []
         if 'Sec-WebSocket-Extensions' in handshake_flow.response.headers:
             if PerMessageDeflate.name in handshake_flow.response.headers['Sec-WebSocket-Extensions']:
```
ujjwal96: We'll need to make a few more changes if we remove this `if` 
because there may be case when extensions=[None] which will throw an error in wsproto

mitmproxy/proxy/protocol/websocket.py
```diff
@@ -81,27 +81,80 @@ def _handle_event(self, event, source_conn, other_conn, is_server):
...
+
         return True
 
     def _handle_ping_received(self, event, source_conn, other_conn, is_server):
         # PING is automatically answered with a PONG by wsproto
-        # TODO: log this PING and its payload
-        self.connections[other_conn].ping(event.payload)
+        self.connections[other_conn].ping()
         other_conn.send(self.connections[other_conn].bytes_to_send())
+        source_conn.send(self.connections[source_conn].bytes_to_send())
```
ujjwal96: Once the we get a PingReceived event a PONG frame will eventually be sent to the sender as it will be in the buffer but the pong is not being sent in L259.
Receive_bytes only adds bytes to the buffer so the pong will not be present before the PingReceived event is generated in the for loop

mitmproxy/proxy/protocol/websocket.py
```diff
@@ -81,27 +81,80 @@ def _handle_event(self, event, source_conn, other_conn, is_server):
...
         # fail-safe for unhandled events
         return True
 
     def _handle_data_received(self, event, source_conn, other_conn, is_server):
+        fb = self.server_frame_buffer if is_server else self.client_frame_buffer
+        fb.append(event.data)
+
+        if event.message_finished:
+            original_chunk_sizes = [len(f) for f in fb]
+            message_type = 0x01 if isinstance(event, events.TextReceived) else 0x02
```
Kriechi: Please use https://github.com/python-hyper/wsproto/blob/24cb9bc5967e106684e19dd0afd1210aedbfc510/wsproto/frame_protocol.py#L58
At some point we might move our websocket stack into pathod, when mitmproxy no longer uses it.

mitmproxy/proxy/protocol/websocket.py
Kriechi: True. 👍

mitmproxy/net/websockets/utils.py
```diff
@@ -88,3 +89,26 @@ def get_client_key(headers):
 
 def get_server_accept(headers):
     return headers.get("sec-websocket-accept", None)
+
+
+def make_extension(extension):
```
Kriechi: I think we can remove all of this, and use the `finalize` and `accept` functions of `wsproto.extensions.PerMessageDeflate` instead.
It takes the raw header value, so we don't have to parse it ourself.

test/mitmproxy/proxy/protocol/test_websocket.py
Kriechi: We could close the connection (close handshake with two close frames), after receiving a `CloseReceived` event, we should also see the log message.

mitmproxy/net/websockets/utils.py
ujjwal96: I don't think it will work because we need to pass the extension(initialized with correct params) as arg to the WSConn object

PR #2533, opened 2017-08-18
Fixing incorrect custom_contentview.py example.

mason-bially:
* Changing `load` to `start`.
    * Removing the old copy of the contentView (since `done` is only called on shutdown).

### DIFF
--- a/examples/simple/custom_contentview.py
+++ b/examples/simple/custom_contentview.py
@@ -21,8 +21,10 @@ view = ViewSwapCase()
 
 
 def start():
+    old_view = contentviews.get(view.name)
+    if old_view is not None:
+        contentviews.remove(old_view)
     contentviews.add(view)
-
 
 def done():
     contentviews.remove(view)


mhils:
Hi, thanks for the PR!
Which version does this affect and how? The example works for me on master, including livereload.

mason-bially:
Specifically:

```
Mitmproxy version: 2.0.2 (release version)
Python version: 3.5.1
Platform: Windows-10-10.0.14393-SP0
SSL version: OpenSSL 1.1.0f  25 May 2017
Windows version: 10 10.0.14393 SP0 Multiprocessor Free
```

Which is why I pull-requested it against the 2.0.x branch.

For that version when using mitmweb, done is never called, so live reload constantly causes errors.

mhils:
Ok. I think this is fixed on master (please feel free to verify it yourself with the snapshots from https://snapshots.mitmproxy.org/), so I guess we can close this?

Thanks again!

mason-bially:
The problem is that this example - as documentation for 2.0.x - is wrong.

mhils:
Ok, thanks.

PR #2542, opened 2017-08-24
Replace cssutils with custom, faster, css prettifier

mhils:
This is a quite pragmatic implementation based on regular expressions. The "trick" is to first replace all special characters in strings, comments, etc., and then perform replacements to only modify non-significant whitespace. The disadvantage of this approach is that we cannot do any lazy evaluation, the upside is that it really is quite fast as regular expressions are super optimized. Here's an example of mitmweb's vendor.css (138KB):

```
beautify() took 0.039s
jsbeautifier took 6.1s
```

@ujjwal96: `strutils.escape_special_areas` should also work for your JS contentview, we have to see if the performance is still good enough for this! 😄


mitmproxy/utils/strutils.py
```diff
@@ -141,3 +142,87 @@ def hexdump(s):
...
+    """
+    patterns = "|".join(
+        r"{lchar}.*?{rchar}".format(
+            lchar=a,
+            rchar=b,
+        ) for (a, b) in area_delimiter)
+    return re.split(
+        "({})".format(patterns),
+        data,
+        re.MULTILINE
```
ujjwal96: We need to specify `flags=`. Third arg is maxsplit
This doesn't work for multi line comments.
```
d = """
test
/* don't
* modify
* me
*/
foo
"""
strutils.split_special_areas(d,  [(r"\s*/\*", r"\*/\s*")])
```
results in `["test\n/* don't\n* modify\n* me\n*/\nfoo\n"]`
Maybe use `re.DOTALL`  here?

mhils: Good catch. Fixed both.

PR #2551, opened 2017-08-30
[web] Add settings.json to Static viewer

MatthewShao:
The generation of `setting.json` improve the extensibility of static viewer, for now, it could display the version of mtimproxy. We could include other settings in `settings.json` in the future.


mhils:
This looks great. Somewhat related, we should get rid of /options at some point 😄

PR #2556, opened 2017-09-01
fix #2477

### LINKED ISSUE #2477: xss_scanner.py issue
hi
in the function test_query_injection the body is converted to lower case (line 153) but in the get_SQLi_data function it is compared with upper case words (line 187) that results in no match

surajt97:
I've made regex to lower case so that lower case body can be compared with lower case regex.


Kriechi:
Would the regex option `IGNORECASE` be a better fit here?

mhils:
Thanks! 😃 🍰

PR #2558, opened 2017-09-02
update installation docs

Kriechi:


docs/install.rst
```diff
@@ -14,153 +14,148 @@ Once installation is complete, you can run :ref:`mitmproxy`, :ref:`mitmdump` or
...
 Installation on Windows
 -----------------------
 
-The recommended way to install mitmproxy on Windows is to use the installer
-provided at `mitmproxy.org`_. After installation, you'll find shortcuts for
-:ref:`mitmweb` (the web-based interface) and :ref:`mitmdump` in the start menu.
-Both executables are  added to your PATH and can be invoked from the command
-line.
+The recommended way to install mitmproxy on Windows is to download our
+self-contained pre-built binary packages from our `releases`_ page.
```
mhils: We should continue to recommend the installer, not the portable binaries.

docs/install.rst
```diff
@@ -14,153 +14,148 @@ Once installation is complete, you can run :ref:`mitmproxy`, :ref:`mitmdump` or
...
+
+After installation, you'll find shortcuts for :ref:`mitmweb` (the web-based
+interface) and :ref:`mitmdump` in the start menu. Both executables are added to
+your PATH and can be invoked from the command line.
 
 .. note::
-    Mitmproxy's console interface is not supported on Windows, but you can use
-    mitmweb (the web-based interface) and mitmdump.
+    The console interface `mitmproxy` is not supported on Windows, but you can
+    use mitmweb (the web-based interface) and mitmdump.
```
mhils: Maybe `Mitmproxy's console interface (``mitmproxy``) is not supported...`?

docs/install.rst
```diff
@@ -14,153 +14,148 @@ Once installation is complete, you can run :ref:`mitmproxy`, :ref:`mitmdump` or
...
-provided at `releases`_.
-
-Our pre-built binaries provide you with the latest version of mitmproxy, a
-self-contained Python 3.5 environment and a recent version of OpenSSL that
-supports HTTP/2. Of course, you can also install mitmproxy from source if you
-prefer that (see :ref:`install-advanced`).
-
-.. _install-advanced:
+The recommended way to install mitmproxy on a recent Linux distribution (with
+Python 3.5 or higher) is to use pip3:
```
mhils: Why pip > prebuilt binaries? I think I'm -1 on this. The portable binaries avoid all the pip/Python version/Python package management issues listed here. For the presumably small set of mitmproxy users that require custom packages in their inline scripts, re-installing using pip seems ok to me.

(Besides, this would also make it much more difficult to upgrade to Python 3.6 for example)

docs/install.rst
```diff
@@ -14,153 +14,148 @@ Once installation is complete, you can run :ref:`mitmproxy`, :ref:`mitmdump` or
...
+OpenSSL that support ALPN and HTTP/2, and other dependencies that would
+otherwise we cumbersome to compile and install.
 
-Installation from Source on Fedora
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Please be advised that we do not updates these binaries after releasing a new
+mitmproxy version. This means we do not include security-related updates of our
+dependencies in already released mitmproxy versions. If there is a severe issue,
+we might consider releasing a bugfix release of mitmproxy and corresponding
+binary packages.
```
mhils: +1 on mentioning this, but this also applies to the non-prebuilt installations as we pin e.g. the cryptography version. Maybe we should generalize it to a simple "Only the latest release is guaranteed to receive security updates".

docs/install.rst
Kriechi: Would you mind adding some information about the Windows installer to this PR?
Maybe a word about the Windows Store as well, or is this not ready yet?

docs/install.rst
Kriechi: My main intention here was, the fact that I don't like `Mitmproxy` spelled with a capital `M`.

docs/install.rst
Kriechi: I personally am in favour of promoting pip as default installation method.

However, to ease future Python upgrades and reduce dependency / version conflicts, I will revert this change to again recommend our binary packages over pip.

docs/install.rst
Kriechi: We don't pin cryptography to a specific version, only to a `v2`: https://github.com/mitmproxy/mitmproxy/blob/781369a3269c05d40fa8d37cfc0eae0401558d8d/setup.py#L67
Or what do you mean with this?

mhils: Once cryptography releases their next minor release (2.1), they won't provide security updates for 2.0 anymore, so we end up with pretty much the same situation - old releases will have old OpenSSL versions. Of course there are small advantanges (e.g. certifi is unbounded in setup.py), but the issue generally remains.

PR #2560, opened 2017-09-03
Split TLS parts from net.tcp into net.tls

mhils:
We can re-use the context bits for the sans-io branch, so it makes sense to separate this now.


Kriechi:
A general 👍  - although there are a few linting & coverage issues.

Issue #2563, opened 2017-09-07
Custom certificate is ignored in reverse proxy mode


chorner:
https://github.com/mitmproxy/mitmproxy/blob/master/docs/certinstall.rst#using-a-custom-certificate
Following the documentation, mitmproxy fails to use the provided custom certificate.

##### Steps to reproduce the problem:

1. `openssl genrsa -out cert.key 2048`
`openssl req -new -x509 -key cert.key -out cert.crt`
`... <specify CSR>`
`cat cert.key cert.crt > cert.pem`
2. `mitmproxy -e -v --cert ./cert.pem -p 8000 -R https://www.google.com:443`
3. `echo | openssl s_client -connect 127.0.0.1:8000`
> ...
> Certificate chain
>  0 s:/CN=www.google.com
>    i:/CN=mitmproxy/O=mitmproxy
>  1 s:/CN=mitmproxy/O=mitmproxy
>    i:/CN=mitmproxy/O=mitmproxy

(the self-signed CSR just created should appear here, rather than the default mitmproxy cert)

mhils:
Thanks for the detailed steps. Which mitmproxy version is this?

chorner:
mitmproxy --version

> Mitmproxy version: 2.0.2 (release version)
> Python version: 3.6.2
> Platform: Darwin-16.7.0-x86_64-i386-64bit
> SSL version: OpenSSL 1.1.0f  25 May 2017
> Mac version: 10.12.6 ('', '', '') x86_64

mhils:
This is fixed on master now, thanks!

PR #2570, opened 2017-09-13
Fixes bug where loaded certs are stored in the certs dict as str

donald-m-ritter:
### Overview
Certs provided through the --cert argument or config yaml get stored in the certs dictionary as a str. This causes issues because the keys created by get_certs method are byte strings. So any provided certs wind up being ignored and the generated certs still get stored as byte strings.

### Steps to Reproduce
1. Start the proxy and provide a cert through the command line or config.yaml options:

  ```bash
  mitmproxy --cert [domain=path to pem]
  ```

2. Attach to the process using the debugger and set a breakpoint in `certs.py` at the `get_cert` method.  
3. Step through the method until you get to `if name:`

This was found when trying to modify requests while keeping certificate pinning enabled using a second cert I have pinned and have the private key to. Once I converted the spec to bytes the domain and certificate were correctly identified.

Additional Info:
Mid 2014 MacBook Pro with 15" Retina Display
macOS Sierra 10.12.5
Python 3.6.1
mitmproxy installed through Homebrew 1.2.4.

masked-screen-1
masked-screen-2

### DIFF
--- a/mitmproxy/proxy/config.py
+++ b/mitmproxy/proxy/config.py
@@ -113,7 +113,7 @@ class ProxyConfig:
                     "Certificate file does not exist: %s" % cert
                 )
             try:
-                self.certstore.add_cert_file(spec, cert)
+                self.certstore.add_cert_file(bytes(spec, "utf-8"), cert)
             except crypto.Error:
                 raise exceptions.OptionsError(
                     "Invalid certificate format: %s" % cert


mhils:
Thanks for the PR - I am right to assume that this only affects 2.0.x and is already fixed on master? 😃

donald-m-ritter:
I haven't checked master yet and won't be able to get to it today.

mhils:
This turned out to be the same issue as https://github.com/mitmproxy/mitmproxy/issues/2563, for which we just committed a fix on master. I'll merge this here, even though the next release will probably be 3.x right away.

Thanks again, and sorry that it took us that long! 😃 🍰

PR #2583, opened 2017-10-10
Fix typo in the docstring of flowfilter

rgerganov:
The expression for body of response is "bs", not "bq".


mhils:
Good catch, thanks! 😃

PR #2589, opened 2017-10-16
docs: clarify filter syntax for mitmproxy

rgerganov:
Filters in mitmproxy don't work with a leading "~" symbol and the user
gets a cryptic error message "Invalid filter pattern" when following the
docs. This patch tries to make the docs better in that respect.

### DIFF
--- a/docs/features/filters.rst
+++ b/docs/features/filters.rst
@@ -15,6 +15,7 @@ Filter expressions consist of the following operators:
 - Header matching (~h, ~hq, ~hs) is against a string of the form "name: value".
 - Strings with no operators are matched against the request URL.
 - The default binary operator is &.
+- When specifying filter in `mitmproxy`, the leading `~` is omitted.
 
 Examples
 --------


mhils:
Thanks for the PR and sorry for the slow response. I'm not sure if I really understand the issue. What did you try and what didn't work? Where is the leading "~" omitted?

rgerganov:
When I start `mitmproxy` like this:

    mitmproxy --replace :~bs:foo:bar

I get this error:

    mitmproxy: Invalid filter pattern: ~bs

but if I run `mitmproxy --replace :bs:foo:bar` it works

mhils:
Thanks for the quick reply! I think there's still some misunderstanding. The syntax for replacements is `:filter:replacement_pattern:replacement`. For example, if you want to replace "kittens" with "dogs" in all responses where the response body contains "parrots", your filter would be `:~bs parrots:kittens:dogs`. If you unconditionally want to replace all kittens with dogs in all responses, you could just do 
`:~s:kittens:dogs`. Does that clear up things?

In any case, that feedback is really helpful as we should improve the examples for replacements. 😃

rgerganov:
What I wanted is to unconditionally replace `foo` with `bar` in all responses. The problem is that `mitmproxy --replace :~bs:foo:bar` doesn't work. It does work when `~` is omitted. I have been using this feature in the past (don't remember which release) and I am pretty sure filters were working the way they are described in the docs. Right now, if you follow the docs, you get "Invalid filter pattern". Do I miss something here?

mhils:
> The problem is that mitmproxy `--replace :~bs:foo:bar` doesn't work. 

If you look at http://docs.mitmproxy.org/en/stable/features/filters.html, it's `~bs regex` - you're omitting the regex which should be matched on the response body (hence the complaint about an invalid filter). If you unconditionally want to mach any response, you want to use `~s`, not `~bs`.

> It does work when `~` is omitted

From the document linked above, "Strings with no operators are matched against the request URL". Your replacement will only be applied to pages with "bs" in the URL, which I don't think is what you want. 😛

Does that make more sense?

rgerganov:
Got it. Thanks for the explanation :) A few more replacement examples in the docs would be quite useful.

rgerganov:
Apparently this PR is not needed. I may open another one for adding more replacement examples in the docs.

mhils:
Thanks! 😃

PR #2591, opened 2017-10-18
Added scanning for CSS injection and iframe injection to XSS scanner

ddworken:
As @EdOverflow suggested, added scanning for CSS injection and iframe injection via unclaimed URLs.


mhils:
Thanks, @ddworken! 😃

PR #2593, opened 2017-10-19
docs-certinstall: Certificate pinning update link

rst0git:
Certificate pinning (HTTP Public Key Pinning) is better described on Wikipedia


mhils:
Thanks! :smiley:

amankumarjain:
@mhils  Hi, can we intercept traffic of pin certificate?

Issue #2110, opened 2017-03-07
`mitmdump --version` is confusing.


mhils:
##### Steps to reproduce the problem:

1. `mitmdump --version`

The output is 
```
Mitmproxy version: 3.0.0.dev (2.0.0dev0064-0xbda3933)
```
for me, stating two conflicting versions. Not sure what we actually want to do here?

Kriechi:
I saw multiple new issues in recent weeks that seem to have mitmproxy-master, but `--version` shows it as `release version`. Thats why I added the `.dev` to our `VERSION` variable.
Gathering the information from git seems to not work reliable, or people are using mitmproxy-master without git (e.g. straight from a tar.gz source or something)

mhils:
Yes. We should also get rid of that "release version" string in its current form, that's pretty misleading.

PR #2602, opened 2017-10-23
Simplify version output

mhils:
- fix display of conflicting version information (#2110)
- Remove `platform.linux/mac/win32` calls. We have all sufficient bits in `platform.platform()` and the `platform.linux` calls are deprecated and will be removed in Python 3.7.
- less verbose output, e.g. skip "version" for every entry.


Kriechi:
does this affect travis/snapshots/binary compilation? filename changes?
Otherwise, :shipit:

mhils:
No, all that is separate in rtool. 😃

PR #2603, opened 2017-10-23
Fix urwid display on WSL

mhils:
| before | after |
|-----|------|
| image | image |

### DIFF
--- a/mitmproxy/tools/console/common.py
+++ b/mitmproxy/tools/console/common.py
@@ -1,9 +1,13 @@
+import platform
+
 import urwid
 import urwid.util
 
 from functools import lru_cache
 from mitmproxy.utils import human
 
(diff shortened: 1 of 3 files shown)


Kriechi:
Strong 👎 on this one.

Isn't the whole point of WSL that applications just run as-is, without any adaption or emulation for Windows crap?

mhils:
That will hopefully be the case at some point, yes.

The Fall Creators Update already improved things remarkably (e.g. mouse support), but the Windows console host still does not implement the shift in / shift out control codes yet, which leads to the rendering artifacts on the left side of the console. If we want to recommend WSL to use mitmproxy on Windows, I think it's fair to add this as a temporary workaround to render properly. Nothing that's meant to stay forever.

PR #2606, opened 2017-10-24
Fix #2563

### LINKED ISSUE #2563: Custom certificate is ignored in reverse proxy mode

mhils:
We previously did not use custom certificates because `"*" != b"*"`.

### DIFF
--- a/mitmproxy/certs.py
+++ b/mitmproxy/certs.py
@@ -114,46 +114,6 @@ def dummy_cert(privkey, cacert, commonname, sans):
     return SSLCert(cert)
 
 
-# DNTree did not pass TestCertStore.test_sans_change and is temporarily replaced by a simple dict.
-#
-# class _Node(UserDict.UserDict):
-#     def __init__(self):
-#         UserDict.UserDict.__init__(self)
-#         self.value = None
-#
-#
-# class DNTree:
-#     """
-#         Domain store that knows about wildcards. DNS wildcards are very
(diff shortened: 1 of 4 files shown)


mitmproxy/certs.py
```diff
@@ -295,10 +261,10 @@ def add_cert_file(self, spec, path):
             privatekey = self.default_privatekey
         self.add_cert(
             CertStoreEntry(cert, privatekey, path),
-            spec
+            spec.encode("idna")
```
mhils: This is the actual fix, everything else is clarification/simplification.

PR #2608, opened 2017-10-25
[web] minor improvements, move options editor

mhils:
@MatthewShao, would you mind looking if the new place for the options editor works out for now? 😃


web/src/js/__tests__/ducks/flowsSpec.js
```diff
@@ -157,7 +157,9 @@ describe('flows actions', () => {
         file     = new window.Blob(['foo'], { type: 'plain/text' })
```
MatthewShao: If we are expecting `anything` on line 162, can we just eliminate these two lines?

PR #2636, opened 2017-11-21
Fix#2635, mitmdump does not apply filter to saved data

MatthewShao:
This bug is caused by a missing condition. Now we only write the stream to file when `save_stream_file` is updated, therefore the `save_stream_filter` can not be applied to the output file.


mhils:
Thanks, @MatthewShao! 😃

PR #2642, opened 2017-11-27
Bug in expected_http_body_size fix

surajt97:


surajt97:
Can you help me understand the error in builds?

Kriechi:
Thanks for the PR! 👍 

Looks good so far - the only thing missing is a test to make sure the fix actually works.
Otherwise the builds will not pass... Sorry about that... ☹️ 

Would you mind writing a simple test here?https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/net/http/http1/test_read.py

surajt97:
@Kriechi done.

Kriechi:
Thanks - Perfect! 👍 🎉

PR #2643, opened 2017-11-29
Fix #2594

### LINKED ISSUE #2594: Transparent mode fail with looking up failure.

1. Launch Wifi Access Point(OS X)
2. Setup pfctl configuration so that http packet will be forwarded.
3. Launch mitmproxy ( `sudo mitmproxy -p 8080 -m transparent --showhost` )
4. Access web page after connecting to AP which launched before.
5. See event log.

When I tried to use transparent mode with OS X(10.11.6).
RuntimeError("Could not resolve original destination.") raised.

I investigated this bug.
And I found that this is caused by difference between AF_INET's and AF_INET6's peername.
https://github.com/mitmproxy/mitmproxy/blob/de006ea8adc08b9a8a6aa94eda2b30468727c307/mitmproxy/net/tcp.py#L567

If we use AF_INET, getpeername() return string like `"192.168.2.5:45670"`.
But if we use AF_INET6, getpeername() return string like `"::ffff:192.168.2.5:45670"`.

`pfctl -s state` 's result is like below.
`ALL tcp 192.168.2.5:45670 -> xx.xx.xx.xx:33291 -> xx.xx.xx.xx:443       ESTABLISHED:ESTABLISHED`

As you see, `::ffff:` doesn't exist.

So [lookup](https://github.com/mitmproxy/mitmproxy/blob/f17c0fdac636f7269f4885294e2a8d2c52c23590/mitmproxy/platform/pf.py#L4) function raises RuntimeError() because `spec in i` condition won't become true.

##### System information

Python version: 3.6.2
Platform: Darwin-15.6.0-x86_64-i386-64bit
SSL version: OpenSSL 1.0.2l  25 May 2017
Mac version: 10.11.6 ('', '', '') x86_64

Ga-ryo:
Just remove "::ffff:".

### DIFF
--- a/mitmproxy/platform/osx.py
+++ b/mitmproxy/platform/osx.py
@@ -1,7 +1,6 @@
 import subprocess
 
 from . import pf
-import re
 
 """
     Doing this the "right" way by using DIOCNATLOOK on the pf device turns out
@@ -16,7 +15,6 @@ import re
 """
 
 STATECMD = ("sudo", "-n", "/sbin/pfctl", "-s", "state")
-ipv4_mapped = re.compile("^::ffff:\d+.\d+.\d+.\d+$")
 
 
 def original_addr(csock):
@@ -35,6 +33,4 @@ def original_addr(csock):
         raise RuntimeError(
             "Insufficient privileges to access pfctl. "
             "See http://docs.mitmproxy.org/en/latest/transparent/osx.html for details.")
-    if ipv4_mapped.match(peer[0]):
-        return pf.lookup(peer[0].replace("::ffff:", ""), peer[1], stxt)
     return pf.lookup(peer[0], peer[1], stxt)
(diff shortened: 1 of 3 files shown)


mhils:
Thanks for tracking this down! 😃 🍰 
Can we move the the replacement to pf.py and add a small test [here](https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/platform/test_pf.py)? 😊

Ga-ryo:
yes please.

cortesi:
@Ga-ryo Hi there. I think @mhils was politely but not very clearly asking you to make a modification to this patch. :grinning: If you could shift the replacement over to pf.py and add a test, we can merge. Thanks for looking at this.

mhils:
No worries 😀 I adjusted the code now, this is ready for review.

Ga-ryo:
Ah, sorry for late reply X(
Thank you for replacing codes.

PR #2646, opened 2017-11-30
Include a warning about leaving the CA cert installed

webxl:
Malicious MITM hosts might be able to detect if a client trusts a mitmproxy certificate and re-encrypt all traffic.

### DIFF
--- a/docs/certinstall.rst
+++ b/docs/certinstall.rst
@@ -13,6 +13,9 @@ mitmproxy CA certificates have to be installed on the client device.
 Quick Setup
 -----------
 
+WARNING!!! Use at your own risk. The CA certificate should be removed before 
+connecting to an untrusted network.
+
 By far the easiest way to install the mitmproxy certificates is to use the
 built-in certificate installation app. To do this, just start mitmproxy and
 configure your target device with the correct proxy settings. Now start a


mhils:
Hi,

Your mitmproxy CA certificate is uniquely generated on your machine at mitmproxy's first start. Unless an attacker can access arbitrary files on your machine, this cannot be used to attack you. As such, it's relatively safe to keep the mitmproxy CA around. :smiley:

webxl:
> Your mitmproxy CA certificate

Can't any MITM attacker generate their own unique certificate using that CA, and then impersonate the servers I'm talking to?

mhils:
> Can't any MITM attacker generate their own unique certificate using that CA

Yes, iff the attacker has access to the CA's private key. The private key however is randomly/uniquely generated on your local machine and unknown to the attacker.

webxl:
Okay. It's not clear that mitm.it is being hosted on the machine that's running mitmproxy. I thought that was coming from the internet.

What about adding something to make it more apparent that this isn't a public CA certificate?

mhils:
There literally is a message on mitm.it about that (see [here](https://github.com/mitmproxy/mitmproxy/blob/5067438ec178643ac6ffe9ce518f502bc680dc99/mitmproxy/addons/onboardingapp/templates/index.html#L86-L92)), but I'm totally open to suggestions on how we can communicate this better without adding more text :smiley:

cortesi:
I think we're covered here, chaps. @webxl - if you think there's more clarity we can bring to the mitm.it page, please feel free to open a PR or an issue to discuss. Thanks!

PR #2656, opened 2017-12-10
browser addon: start an isolated browser attached to the proxy

cortesi:
A simple addon that starts an instance of Chrome attached to the current
proxy. The instance is isolated in its own user data directory, and addons are
turned off.

Future work:

- I wasn't able to test the Windows executable path - a Windows dev should
confirm this for us.
- In future it would be nice to support other browsers like Firefox.


mhils:
Looks great! I'll check Windows tomorrow morning. 👍

PR #2669, opened 2017-12-13
commands: add a Path argument type

### LINKED ISSUE #2198: Teach options to know about a Path type

cortesi:
This is just an alias for str, and in this patch is used mostly to give an
appropriate type in help strings. More to come.

Fixes #2198

### DIFF
--- a/mitmproxy/command.py
+++ b/mitmproxy/command.py
@@ -37,6 +37,9 @@ class Choice(str):
     options_command = ""
 
 
+Path = typing.NewType("Path", str)
+
+
 def typename(t: type, ret: bool) -> str:
     """
         Translates a type to an explanatory string. If ret is True, we're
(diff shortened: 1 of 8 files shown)


mitmproxy/command.py
```diff
@@ -37,15 +37,16 @@ class Choice(str):
     options_command = ""
 
 
+Path = typing.NewType("Path", str)
```
mhils: Have you considered using the following?
```python
class Path(str):
    pass
```
That would enable the `issubclass` check in [line 189](https://github.com/mitmproxy/mitmproxy/pull/2669/files#diff-c0ab5de8fa401179a3c28d8a4496437cL189) to work, but I presume there's a different issue with that approach?

cortesi: I'm neutral. NewType is the "blessed way" to declare an unadorned equivalent type, on the theory that it's zero cost. Path is not exactly performance critical so either way works. I'll merge and we can adjust as we work on this.

PR #2671, opened 2017-12-13
Introduce @command.argument

mhils:
This makes it possible to specify more specific type annotations at runtime, so that both mypy and our command system are happy. The `.argument(name, type=)` syntax is similar to click's, so it should be fairly extensible if we need it.


cortesi:
I really like this approach - it's much nicer. Having a flexible type here might let us further formalise the Choices mechanism down the track. We could, for instance, auto-generate the matching .choices commands rather than manually specify them, and then the Choices type could just take a list of options or a callable with a signature similar to commands. Lets get this merged!

PR #2674, opened 2017-12-14
Fix #2620

### LINKED ISSUE #2620: Mitmproxy odd UI behaviour

1. Run `mitmproxy` and start intercepting flows.

2. Let the UI fill with flows and then goto the last flow.

m1

3. Press up key to go one flow above.

4. The UI shifts one flow downwards instead of marker going one flow upwards.

m2

This is not a deal breaker or anything but a bit annoying. This seems to happen only with the last flow in UI.

##### System information

```
Mitmproxy: 3.0.0 
Python:    3.6.2
OpenSSL:   OpenSSL 1.0.2g  1 Mar 2016
Platform:  Linux-4.8.0-53-generic-x86_64-with-debian-stretch-sid
```

mhils:
We previously had the problem that overriding keypress() skipped the proper calculation of the top widget's size, leading to broken scrolling behavior in the flowlist. We now always use urwid.Frame's keypress method, but we make sure that urwid.Pile and urwid.Columns delegate to the currently focused component.

@cortesi: this seems suspiciously simple - anything I missed?

### DIFF
--- a/mitmproxy/tools/console/window.py
+++ b/mitmproxy/tools/console/window.py
@@ -156,12 +156,14 @@ class Window(urwid.Frame):
             w = urwid.Pile(
                 [
                     wrapped(i) for i, s in enumerate(self.stacks)
-                ]
+                ],
+                focus_item=self.pane
             )
         else:
             w = urwid.Columns(
                 [wrapped(i) for i, s in enumerate(self.stacks)],
-                dividechars=1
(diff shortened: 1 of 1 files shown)


cortesi:
Welp, it gets a "seems to work for me" endorsement. :) I'll merge, on the assumption that I'm working in this area right now, and will notice if something weird breaks.

PR #2677, opened 2017-12-14
Commander

cortesi:
Commands are awesome but impose a complexity burden on users. One way to alleviate this is to have a smarter command prompt that can check syntax and perform tab completions. This PR introduces a basic commander widget, puts in place tab completions for command names only, and sets the stage for completions for choices, paths, cut specifications and so forth.


mhils:
commander gif

mitmproxy/command.py
```diff
@@ -172,6 +208,8 @@ def parsearg(manager: CommandManager, spec: str, argtype: type) -> typing.Any:
                 "Invalid choice: see %s for options" % cmd
             )
         return spec
+    if argtype in (Path, Cmd):
+        return spec
```
mhils: not required, should be caught by the clause below?

PR #2680, opened 2017-12-15
commands: Reassess the cuts API

cortesi:
Make the cuts API more transparent. Cut specifications are no longer a
centrally resolved core type, and flows are now passed explicitly.


cortesi:
Superseded

PR #2683, opened 2017-12-16
Allow shorthands for command.Choice strings

mhils:
Not feeling very strong about this, but this seems to be a useful addition to me.


cortesi:
Hmm. I think i'm a net negative on this one, and in fact I've been removing shortcuts from other places (like cut specifications). We now have tab completion on the command prompt. So, the user types the prefix, presses tab to complete and confirm that the option is valid, and then completes the command. I think that's clear and works well. 

There are places where the user is prompted to choose an option in a popup. There I think work should be done in the popup selector to allow shortcut or keystroke selection. 

Either way, I don't think this should be in the parser.

mhils:
Fair enough, makes sense. Let's close this here!

Issue #2196, opened 2017-03-21
Refactor handling of persistent connection flows


cortesi:
We currently have a structure where we have TCPFlow and WebsocketFlow objects that represent persistent connections. These objects then accumulate messages while the connection lasts. The entire thing is serialised in one chunk when the persistent connection ends if we're writing to file.

This structure has two issues. First, all messages are accumulated in memory if we're writing to disk. We'd like to be able to shed data to disk as quickly as we can to keep memory usage low (especially in the light of the session proposal in #2175). Second, this is out of step with the HTTP/HTTP2 structure, where the flow is the primary target of user interaction. So, a user "replays a flow" or "intercepts a flow" - this sits awkwardly with persistent connections, where the target is really a contained message. This in turn has flow-on effects on how we might design our user interfaces.

I propose that we promote TCPMessage and WebsockeMessage to top-level Flow objects that can be serialized to disk immediately. Each message keeps a reference to the parent connection object's unique ID. 

With the current mechanism, we then have a situation where a WebsocketMessage appears in the flow file before the WebsocketConnection (which is written only when the connection terminates). If we didn't gracefully terminate, the connection  might not be written at all, so the file could potentially be invalid. To address that, I propose that we allow partial records to be written to the flow file - a WebsocketConnection entry is made before the first WebsocketMessage (containing partial data), and another entry with the same unique ID is made after the last WebsocketMessage (containing complete data). This keeps the file consistent chronologically, and ensures that we never have partial data on disk.

If we can represent persistent connections in this way, I also propose that we consider doing this for our ClientConnection and ServerConnection primitives. That is, promote them to full flows, with partial records streamed to disk, and with child objects keep a reference to parent objects through their unique IDs.

mhils:
> To address that, I propose that we allow partial records to be written to the flow file - a WebsocketConnection entry is made before the first WebsocketMessage (containing partial data), and another entry with the same unique ID is made after the last WebsocketMessage (containing complete data). This keeps the file consistent chronologically, and ensures that we never have partial data on disk.

This sounds good to me. If we take this a step further, we'd have an event store instead of a flow store, e.g. we'd stream (connection open, message, message, message, connection close) sequentially to disk. I think it's quite interesting to make this a bit redux-like.

Josue-T:
Hello,

I'm implementing a flow support for HTTP/2. My questions is now, in your point of view, what would be the best solution:

1. One flow per HTTP/2 frame:
[...]

PR #2702, opened 2017-12-17
improve websocket dumps

### LINKED ISSUE #2640: Dump with websocket messages
Hello,

I would like to record full http + websocket traffic including websocket messages. 

I have tried `mitmdump -w dump`

However, it seems that in the recorded file, only the http requests+ responses are recorded, and the websocket initialisation, but not the separate websocket messages (called frames in chrome Dev tool).

Is it possible to record the websocket frames too with the command line directly ?

I am running master (46901d1d55cb5bad86e17e093b85ade8111315aa)

Thanks in advance

Kriechi:
websocket supports domain and url filters now
fixes #2640

### DIFF
--- a/mitmproxy/flowfilter.py
+++ b/mitmproxy/flowfilter.py
@@ -322,13 +322,14 @@ class FDomain(_Rex):
     flags = re.IGNORECASE
     is_binary = False
 
-    @only(http.HTTPFlow)
+    @only(http.HTTPFlow, websocket.WebSocketFlow)
     def __call__(self, f):
+        if isinstance(f, websocket.WebSocketFlow):
+            f = f.handshake_flow
         return bool(
             self.re.search(f.request.host) or
             self.re.search(f.request.pretty_host)
         )
-
 
 class FUrl(_Rex):
     code = "u"
@@ -342,9 +343,11 @@ class FUrl(_Rex):
             toks = toks[1:]
         return klass(*toks)
 
-    @only(http.HTTPFlow)
+    @only(http.HTTPFlow, websocket.WebSocketFlow)
     def __call__(self, f):
-        if not f.request:
+        if isinstance(f, websocket.WebSocketFlow):
+            f = f.handshake_flow
+        if not f or not f.request:
             return False
         return self.re.search(f.request.pretty_url)
 
--- a/mitmproxy/flowfilter.py
+++ b/mitmproxy/flowfilter.py
@@ -331,6 +331,7 @@ class FDomain(_Rex):
             self.re.search(f.request.pretty_host)
         )
 
+
 class FUrl(_Rex):
     code = "u"
     help = "URL"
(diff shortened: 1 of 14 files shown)


mitmproxy/flowfilter.py
```diff
@@ -322,8 +322,10 @@ class FDomain(_Rex):
     flags = re.IGNORECASE
     is_binary = False
 
-    @only(http.HTTPFlow)
+    @only(http.HTTPFlow, websocket.WebSocketFlow)
     def __call__(self, f):
+        if isinstance(f, websocket.WebSocketFlow):
+            f = f.handshake_flow
```
mhils: This is - excuse the language - a bit half-assed. Let's add a `@websocket_as_http` decorator that does this "conversion" and add it to all applicable filters.
```python
def websocket_as_http(fn):
    @functools.wraps(fn)
    def wrapped(self, flow):
        if isinstance(flow, websocket.WebSocketFlow):
            flow = flow.handshake_flow
        return fn(self, flow)
    return wrapped
```

mitmproxy/master.py
```diff
@@ -117,15 +119,28 @@ def shutdown(self):
         self.should_exit.set()
         self.addons.trigger("done")
 
+    def _change_reverse_host(self, f):
+        if self.options.mode.startswith("reverse:"):
```
mhils: Let's add a doc string since you asked on Slack what this does:
"When we load flows in reverse proxy mode, we adjust the target host to the reverse proxy destination for all flows we load. This makes it very easy to replay saved flows against a different host."

mitmproxy/master.py
```diff
@@ -117,15 +119,28 @@ def shutdown(self):
...
-            if self.options.mode.startswith("reverse:"):
-                _, upstream_spec = server_spec.parse_with_mode(self.options.mode)
-                f.request.host, f.request.port = upstream_spec.address
-                f.request.scheme = upstream_spec.scheme
+            self._change_reverse_host(f)
+            if 'websocket' in f.metadata:
+                self.waiting_flows.append(f)
+
+        if isinstance(f, websocket.WebSocketFlow):
+            hf = [hf for hf in self.waiting_flows if hf.id == f.metadata['websocket_handshake']][0]
```
mhils: If someone asks, I did not see this 🙈. 
Ok for now. Let's tackle this properly with https://github.com/mitmproxy/mitmproxy/issues/2196 by maybe combining HTTP and WebSocket Flows.

mitmproxy/proxy/protocol/http.py
```diff
@@ -321,6 +321,7 @@ def _process_flow(self, f):
 
         try:
             if websockets.check_handshake(request.headers) and websockets.check_client_version(request.headers):
+                f.metadata['websocket'] = True
```
mhils: Can't we just use `metadata['websocket_flow']` to identify handshake flows and not add another attribute?

mitmproxy/master.py
Kriechi: 👍

Issue #2693, opened 2017-12-17
Flowview Keybindings are broken


mhils:
##### Steps to reproduce the problem:

1. Open flow
2. Press `e`
3. `Invalid value type: ['console.edit.focus', '{choice}'] - expected <class 'mitmproxy.command.Arg'>`

`v` also doesn't work. `o` in the flowlist also does not work.
##### System information

Mitmproxy: 3.0.0dev1042 (ef9ad11)
Python:    3.5.2
OpenSSL:   OpenSSL 1.0.2g  1 Mar 2016
Platform:  Linux-4.4.0-43-Microsoft-x86_64-with-Ubuntu-16.04-xenial

cortesi:
I've got a fix for this one inbound as part of the type refactoring.

cortesi:
Fixed in #2712

PR #2703, opened 2017-12-17
Test that default key bindings are valid commands

mhils:
This seems to generally work, but it's currently complains about https://github.com/mitmproxy/mitmproxy/issues/2693 and fails. Not sure if there are further issues beyond that.


rebeccagrcia470:
Thanks

PR #2705, opened 2017-12-17
Re-add body editing

mhils:
@cortesi: Not sure if I am doing this as intended - see https://github.com/mitmproxy/mitmproxy/compare/master...mhils:issue-2694?expand=1#diff-14fc04d5ac58857f7ce7f02e8ff05b50R375.

Is there a particular reason why this command is `console.edit.focus` and not `console.edit @focus`?

### DIFF
--- a/mitmproxy/tools/console/consoleaddons.py
+++ b/mitmproxy/tools/console/consoleaddons.py
@@ -348,6 +348,8 @@ class ConsoleAddon:
             "reason",
             "request-headers",
             "response-headers",
+            "request-body",
+            "response-body",
             "status_code",
             "set-cookies",
             "url",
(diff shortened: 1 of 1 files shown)


mitmproxy/tools/console/consoleaddons.py
```diff
@@ -369,6 +371,22 @@ def edit_focus(self, part: str) -> None:
...
         elif part == "response-headers":
             self.master.switch_view("edit_focus_response_headers")
+        elif part in ("request-body", "response-body"):
+            # FIXME: Is this the right approach?
+            # FIXME: This needs to be guarded against having no flows.
+            flow = self.master.view.focus.flow
+            if part == "request-body":
+                message = flow.request
+            else:
+                message = flow.response
```
cortesi: flow.response can also be None here

PR #2707, opened 2017-12-18
commands: refactor types

cortesi:
The type system was scattered over a number of places, making it hard to
follow. This collects all command types in types.py, and completion, validation
and parsing for each type is centralised. We should use the same mechanism for
options.

### DIFF
--- a/mitmproxy/exceptions.py
+++ b/mitmproxy/exceptions.py
@@ -112,6 +112,10 @@ class AddonHalt(MitmproxyException):
     pass
 
 
+class TypeError(MitmproxyException):
+    pass
+
+
 """
     Net-layer exceptions
 """
--- a/mitmproxy/types.py
+++ b/mitmproxy/types.py
@@ -1,0 +1,330 @@
+import os
+import glob
+import typing
+
+from mitmproxy import exceptions
+from mitmproxy import flow
+
+
+class Path(str):
+    pass
+
+
+class Cmd(str):
+    pass
+
+
+class Arg(str):
+    pass
+
+
+class CutSpec(typing.Sequence[str]):
+    pass
+
+
+class Data(typing.Sequence[typing.Sequence[typing.Union[str, bytes]]]):
+    pass
+
+
+class Choice:
+    def __init__(self, options_command):
+        self.options_command = options_command
+
+    def __instancecheck__(self, instance):  # pragma: no cover
+        # return false here so that arguments are piped through parsearg,
+        # which does extended validation.
+        return False
+
+
+# One of the many charming things about mypy is that introducing type
+# annotations can cause circular dependencies where there were none before.
+# Rather than putting types and the CommandManger in the same file, we introduce
+# a stub type with the signature we use.
+class _CommandStub:
+    commands = {}  # type: typing.Mapping[str, typing.Any]
+
+    def call_args(self, path: str, args: typing.Sequence[str]) -> typing.Any:  # pragma: no cover
+        pass
+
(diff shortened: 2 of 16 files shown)


mitmproxy/exceptions.py
```diff
@@ -112,6 +112,10 @@ class AddonHalt(MitmproxyException):
     pass
 
 
+class TypeError(MitmproxyException):
```
mhils: This looks wrong - we should use a regular Python TypeError?

mitmproxy/types.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+class BaseType:
+    typ = object  # type: typing.Type
+    display = ""  # type: str
+
+    def completion(
+        self, manager: _CommandStub, t: type, s: str
+    ) -> typing.Sequence[str]:  # pragma: no cover
+        pass
```
mhils: If you make this a `raise NotImplementedError()`, we don't need the `pragma: no cover` annotation.

mitmproxy/types.py
```diff
@@ -0,0 +1,330 @@
...
+        # return false here so that arguments are piped through parsearg,
+        # which does extended validation.
+        return False
+
+
+# One of the many charming things about mypy is that introducing type
+# annotations can cause circular dependencies where there were none before.
+# Rather than putting types and the CommandManger in the same file, we introduce
+# a stub type with the signature we use.
+class _CommandStub:
```
mhils: Ugh. I feel your pain, but this looks even more painful. Can we make this `_Command` or `ICommand` and make the actual `Command` inherit from it? This way we can at least assure that both stay in sync.

mitmproxy/types.py
```diff
@@ -0,0 +1,330 @@
...
+
+class FlowsType(BaseFlowType):
+    typ = typing.Sequence[flow.Flow]
+    display = "[flow]"
+
+    def parse(self, manager: _CommandStub, t: type, s: str) -> typing.Sequence[flow.Flow]:
+        return manager.call_args("view.resolve", [s])
+
+
+class DataType:
```
mhils: Intentional that you don't inherit from BaseType here?

mitmproxy/types.py
```diff
@@ -0,0 +1,330 @@
...
+    ) -> typing.Sequence[str]:  # pragma: no cover
+        raise exceptions.TypeError("data cannot be passed as argument")
+
+    def parse(
+        self, manager: _CommandStub, t: type, s: str
+    ) -> typing.Any:  # pragma: no cover
+        raise exceptions.TypeError("data cannot be passed as argument")
+
+
+class ChoiceType:
```
mhils: (same)

mitmproxy/types.py
cortesi: I'll think about this. It's a shame that the type checker is futzing with our code structure to the extent that we might have inheritance patterns that are just bogus.

mitmproxy/types.py
cortesi: I'll think about this. It's a shame that the type checker is futzing with our code structure to the extent that we might have inheritance patterns that are just bogus.

cortesi:
My tree has diverged, so I'll merge this PR and address comments in the next push. Basically +1 on all of this. Let's consider where to put the types as we write the documentation - with types shared between options an commands (not there yet, but we will be), .types seems more natural.

mitmproxy/exceptions.py
cortesi: This is an unfortunate clash of names, but I'd like to keep this type separate from the built-in TypeError.

PR #2708, opened 2017-12-18
@concurrent decorator works for class methods.

fate0:
``` python
import time

class ConcurrentClass:

    @concurrent  # Remove this and see what happens
    def request(self, flow):
        # You don't want to use mitmproxy.ctx from a different thread
        print("handle request: %s%s" % (flow.request.host, flow.request.path))
        time.sleep(5)
        print("start  request: %s%s" % (flow.request.host, flow.request.path))


addons = [ConcurrentClass()]
```

`@concurrent` only works for class staticmethod for now, I make it works for normal class method.

### DIFF
--- a/mitmproxy/script/concurrent.py
+++ b/mitmproxy/script/concurrent.py
@@ -18,7 +18,9 @@ def concurrent(fn):
         )
 
     def _concurrent(*args):
-        # Support @concurrent for class-based addons
+        # When annotating classmethods, "self" is passed as the first argument.
+        # To support both class and static methods, we accept a variable number of arguments
+        # and take the last one as our actual hook object.
         obj = args[-1]
 
         def run():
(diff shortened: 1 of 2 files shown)


mitmproxy/script/concurrent.py
```diff
@@ -17,9 +17,11 @@ def concurrent(fn):
             "Concurrent decorator not supported for '%s' method." % fn.__name__
         )
 
-    def _concurrent(obj):
+    def _concurrent(*args):
+        obj = args[0] if len(args) == 1 else args[1]
```
mhils: This won't be obvious if we look at it in a year, let's add a comment explaining what we are doing here. Also, we can simplify to `args[-1]` 😃

fate0: nice, thank you for the tips 😄

mhils:
Thanks again! 😃

PR #2709, opened 2017-12-18
Add notes on making changes persistent

jimhigson:
It took me a while to work out how to do this exactly on my raspberry pi (running rasbian) so adding some notes for the next person who tries.

### DIFF
--- a/docs/transparent/linux.rst
+++ b/docs/transparent/linux.rst
@@ -50,8 +50,9 @@ Debain
 To make the changes pperminent on on debian (inc ubuntu and raspbian)
 systems:
 
-  1. Write the sysctl changes to a new config file at (for example) /etc/sysctl.d/mitm.conf    
-    .. code-block: : none
+ 1. Write the sysctl changes to a new config file at (for example) /etc/sysctl.d/mitm.conf    
+ 
+    .. code-block:: none
     
         net.ipv4.ip_forward=1
         net.ipv4.conf.all.accept_redirects = 0
--- a/docs/transparent/linux.rst
+++ b/docs/transparent/linux.rst
@@ -58,5 +58,6 @@ systems:
         net.ipv4.conf.all.accept_redirects = 0
         net.ipv4.conf.all.send_redirects = 0
     
-  2. To make iptables changes persistent, install the package iptables-persistent:
+ 2. To make iptables changes persistent, install the package iptables-persistent:
+ 
     >>> sudo apt-get install iptables-persistent
--- a/docs/transparent/linux.rst
+++ b/docs/transparent/linux.rst
@@ -12,15 +12,15 @@ achieve transparent mode.
 
     >>> sysctl -w net.ipv4.ip_forward=1
 
-    You may also want to consider enabling this permanently in ``/etc/sysctl.conf``.
+    You may also want to consider enabling this permanently in ``/etc/sysctl.conf`` or newly created ``/etc/sysctl.d/mitmproxy.conf ``.
 
  3. If your target machine is on the same physical network and you configured it to use a custom
     gateway, disable ICMP redirects:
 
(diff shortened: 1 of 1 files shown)


mhils:
This looks useful, thanks. There are some typos that need to be fixed, but we can take care of that if you don't feel comfortable doing it. 

@cortesi @Kriechi: Do you see any particular risks with persisting sysctl/iptables changes that we need to mention?

jimhigson:
Thanks. I wasn't completely familiar with rst format, which probably explains the typos. I fixed a couple just now.

Kriechi:
Maybe a: *Make sure to undo these settings when you are done with using mitmproxy in transparent mode!* warning wouldn't hurt?

mhils:
If "undo these settings when you are done" is our recommended behaviour, we shouldn't explain how to persist them.

jimhigson:
Intersting. I didn't realise that was the recommendation. In my case I was setting up a dedicated headless raspberry pi zero purely for use as a mitm proxy so I wanted these settings to persist. Fair enough if that's not a generally recommended setup.

mhils:
Well, not sure what we [should] recommend. The iptables rules should be generally harmless if mitmproxy is not running, the only issue might be `net.ipv4.ip_forward` in some very uncommon network setups (uncommon for mitmproxy at least). 

Long term, I'd love to move something that tries to set up transparent mode automagically (especially on Windows for self-proxying) and then drops privileges, but that's a thing of the future. @cortesi, opinions?

jimhigson:
In my instructions, adding an undo for the sysctl variables would be a one liner - `rm /etc/sysctl/mitm.conf`

Kriechi:
`rm` + triggering sysctl to re-read and apply the settings, right?

docs/transparent/linux.rst
```diff
@@ -43,3 +43,23 @@ achieve transparent mode.
 
 
 For a detailed walkthrough, have a look at the :ref:`transparent-dhcp` tutorial.
+
+Debain
```
Kriechi: Deb**ia**n

docs/transparent/linux.rst
```diff
@@ -43,3 +43,23 @@ achieve transparent mode.
...
+To make the changes permanent on on Debian (inc ubuntu and raspbian)
+systems:
+
+ 1. Write the sysctl changes to a new config file at (for example) /etc/sysctl.d/mitm.conf    
+ 
+    .. code-block:: none
+    
+        net.ipv4.ip_forward = 1
+        net.ipv4.conf.all.accept_redirects = 0
+        net.ipv4.conf.all.send_redirects = 0
```
Kriechi: Why are ICMP redirects important for mitmproxy?

Kriechi:
Merged via https://github.com/mitmproxy/mitmproxy/commit/907ef4a02c3f37edc25f512a0714ad95cac628dd.
Thanks! 🎉 👍

docs/transparent/linux.rst
```diff
@@ -12,15 +12,15 @@ achieve transparent mode.
...
     >>> sysctl -w net.ipv4.ip_forward=1
 
-    You may also want to consider enabling this permanently in ``/etc/sysctl.conf``.
+    You may also want to consider enabling this permanently in ``/etc/sysctl.conf`` or newly created ``/etc/sysctl.d/mitmproxy.conf ``.
 
  3. If your target machine is on the same physical network and you configured it to use a custom
     gateway, disable ICMP redirects:
 
-    >>> echo 0 | sudo tee /proc/sys/net/ipv4/conf/*/send_redirects
+    >>> sysctl -w net.ipv4.conf.all.accept_redirects=0
```
mhils: @jimhigson: Any particular reason why we are disabling `accept_redirects` here? My understanding would be that we just need to not send them.

PR #2710, opened 2017-12-18
Move trust instructions for ios

jimhigson:
Actually, these instructions were mostly already there, but not under the iOS section which made them hard to find (I completely missed them). Put them in the correct section and expand them a little.


mhils:
Good catch. Thanks!

PR #2728, opened 2017-12-27
websocket: docs++ and kill messages

### LINKED ISSUE #2724: Proxy crashes on flow.kill()

0. (Sorry if this is my own mistake, I just started poking around with mitmproxy)
1. Load the proxy with the scenario which is supposed to kill every websocket request with 50% chance
```
from mitmproxy import ctx, websocket
import random


def websocket_message(flow: websocket.WebSocketFlow) -> None:
    ctx.log.info(flow.messages[-1].content.strip()[:20])
    if random.random() < 50:
        flow.kill()
```
2. Start the test which opens a websocket connection through the proxy and sends several binary chunks of data.
=> Proxy crashes

```
  File "/Users/eugene/gitrepo/mitmproxy/mitmproxy/tools/console/master.py", line 216, in run
    self.loop.run()
  File "/Users/eugene/gitrepo/mitmproxy/venv/lib/python3.6/site-packages/urwid/main_loop.py", line 278, in run
    self._run()
  File "/Users/eugene/gitrepo/mitmproxy/venv/lib/python3.6/site-packages/urwid/main_loop.py", line 376, in _run
    self.event_loop.run()
  File "/Users/eugene/gitrepo/mitmproxy/venv/lib/python3.6/site-packages/urwid/main_loop.py", line 682, in run
    self._loop()
  File "/Users/eugene/gitrepo/mitmproxy/venv/lib/python3.6/site-packages/urwid/main_loop.py", line 715, in _loop
    alarm_callback()
  File "/Users/eugene/gitrepo/mitmproxy/venv/lib/python3.6/site-packages/urwid/main_loop.py", line 164, in cb
    callback(self, user_data)
  File "/Users/eugene/gitrepo/mitmproxy/mitmproxy/tools/console/master.py", line 175, in ticker
    changed = self.tick(timeout=0)
  File "/Users/eugene/gitrepo/mitmproxy/mitmproxy/master.py", line 109, in tick
    self.addons.handle_lifecycle(mtype, obj)
  File "/Users/eugene/gitrepo/mitmproxy/mitmproxy/addonmanager.py", line 234, in handle_lifecycle
    message.reply.take()
  File "/Users/eugene/gitrepo/mitmproxy/mitmproxy/controller.py", line 88, in take
    "Reply is {}, but expected it to be start.".format(self.state)
mitmproxy.exceptions.ControlException: Reply is committed, but expected it to be start.

mitmproxy has crashed!
```

```
Mitmproxy: 3.0.0dev1075 (a6a4b1c3) 
Python:    3.6.3
Platform:  Darwin-17.2.0-x86_64-i386-64bit
```

Kriechi:
fixes #2724

### DIFF
--- a/docs/scripting/api.rst
+++ b/docs/scripting/api.rst
@@ -10,6 +10,9 @@ API
     - `mitmproxy.http.HTTPRequest <#mitmproxy.http.HTTPRequest>`_
     - `mitmproxy.http.HTTPResponse <#mitmproxy.http.HTTPResponse>`_
     - `mitmproxy.http.HTTPFlow <#mitmproxy.http.HTTPFlow>`_
+- WebSocket
+    - `mitmproxy.websocket.WebSocketFlow <#mitmproxy.websocket.WebSocketFlow>`_
+    - `mitmproxy.websocket.WebSocketMessage <#mitmproxy.websocket.WebSocketMessage>`_
 - Logging
     - `mitmproxy.log.Log <#mitmproxy.controller.Log>`_
     - `mitmproxy.log.LogEntry <#mitmproxy.controller.LogEntry>`_
@@ -31,6 +34,15 @@ HTTP
     :inherited-members:
 
 .. autoclass:: mitmproxy.http.HTTPFlow
+    :inherited-members:
+
+WebSocket
+---------
+
+.. autoclass:: mitmproxy.websocket.WebSocketFlow
+    :inherited-members:
+
+.. autoclass:: mitmproxy.websocket.WebSocketMessage
     :inherited-members:
 
 Logging
--- a/docs/scripting/events.rst
+++ b/docs/scripting/events.rst
@@ -187,8 +187,8 @@ are issued, only new WebSocket messages are called.
 
         - Called when a WebSocket message is received from the client or server. The
           sender and receiver are identifiable. The most recent message will be
-          ``flow.messages[-1]``. The message is user-modifiable. Currently there are
-          two types of messages, corresponding to the BINARY and TEXT frame types.
+          ``flow.messages[-1]``. The message is user-modifiable and is killable.
+          A message is either of TEXT or BINARY type.
 
           *flow*
             A ``models.WebSocketFlow`` object.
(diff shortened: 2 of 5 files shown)


Kriechi:
@EugeneAbramchuk does this help you with #2724?

EugeneAbramchuk:
@Kriechi how odd. I pulled from the branch and at first the script worked, however it killed way more messages than it should have.

```
def websocket_message(flow: websocket.WebSocketFlow) -> None:
    global i
    i += 1
    if i % 3 == 0:
        flow.kill()
```

I sent 34 messages and only 3 got through (but no crash). When I started to debug the script and restarted the proxy, it started crashing with the same stacktrace again.

Does restart leave any lasting effects on the state? Could there be any sticky remaining connections to the endpoint, which might have caused this behaviour?

Kriechi:
@EugeneAbramchuk I think you are still killing the whole connection.
The new API for killing messages is: `flow.messages[-1].kill()`.
Sorry, I didn't mention it before, I updated the docs to reflect this.

EugeneAbramchuk:
Just to clarify, there are two distinct issues here: 

- proxy crashing on `flow.kill()` - which I filed the ticket originally, and it still crashes with the update
- single message killing. I updated the test to `flow.messages[-1].kill()` and proxy no longer crashes with the change, but looks like it does not kill messages this way either:
```
error: Addon error: Traceback (most recent call last):
  File "log-websocket-requests.py", line 21, in websocket_message
    flow.messages[-1].kill()
AttributeError: 'WebSocketMessage' object has no attribute 'kill'
```

Do I need to update to use the new API?

Thanks!

Kriechi:
Yes, point 1 is still unresolved, but you should not trigger this anyway with the new API.
You point 2 should work if you are running this branch. I added this function https://github.com/mitmproxy/mitmproxy/pull/2728/files#diff-0de40cbc6458548283e26e099bf4edccR48
So the error you get should not happen... Please make sure to checkout the correct branch & code path.

EugeneAbramchuk:
Oh! You are right, I pulled from the repository but forgot to switch branches.

The problem does not reproduce anymore with `flow.messages[-1].kill()` and the messages are indeed killed.

PR #2736, opened 2017-12-31
Parse Commands in POSIX mode

mhils:
If posix=False, quotes are not stripped from parsed tokens:

```python
>>> list(shlex.shlex("foo 'bar baz'"))
['foo', "'bar baz'"]
>>> list(shlex.shlex("foo 'bar baz'", posix=True))
['foo', 'bar baz']
```

@cortesi, was this intentionally left in non-posix mode? I haven't fully looked into this, but I believe this breaks flowspec filter resolution at the moment: `replay.client "~m GET"` does not work, `replay.client ~s` does.


mhils:
I'm merging this as it's surely better than the super annoying current state. We can still improve on this later.

PR #2742, opened 2018-01-01
Dataviewer: disable (broken) edit

mhils:
The edit functionality was previously broken due to encoding issues, so we just replace the column type with a generic read-only variant.


mitmproxy/tools/console/grideditor/editors.py
```diff
@@ -175,11 +175,23 @@ def is_error(self, col, val):
...
+            master,
+            vals: typing.Union[
+                typing.List[typing.List[typing.AnyStr]],
+                typing.List[typing.AnyStr],
+                str,
+            ]):
         if vals:
+            # Whatever vals is, make it a list of rows containing lists of column values.
+            if isinstance(vals, str):
+                vals = [vals]
```
mhils: This is an additional bugfix, where we previously rendered a single string as one character per row.

PR #2744, opened 2018-01-02
Fix crash when using body filters

Fusion86:
They still don't work, but at least the web interface won't crash.


mhils:
Oops, thanks! Could you please fix this in filt.peg, which is used to generate filt.js? 😃

mhils:
Very cool, thanks for the fast turnaround! 😃 🍰 🎉

Issue #2738, opened 2017-12-31
Mitmproxy lets us interact with other tab, when focusing on the other one.


kajojify:
##### Steps to reproduce the problem:

1. Open mitmproxy.
2. Press `-` to open two pane layout.
3. Start intercepting something.
4. Choose any flow to see **Flow details**.
5. Change focus on other tab by pressing `Shift+Tab`.
6. Click on "Request" or "Response" by pressing left button of mouse.

![2017-12-31 19-52-30](https://user-images.githubusercontent.com/20267977/34463805-5bb3ebd0-ee70-11e7-883b-c6dc720e4127.png)

I am getting:
```
Traceback (most recent call last):
  File "mitmproxy/tools/console/master.py", line 216, in run
  File "site-packages/urwid/main_loop.py", line 278, in run
  File "site-packages/urwid/main_loop.py", line 376, in _run
  File "site-packages/urwid/main_loop.py", line 682, in run
  File "site-packages/urwid/main_loop.py", line 719, in _loop
  File "site-packages/urwid/raw_display.py", line 393, in <lambda>
  File "site-packages/urwid/raw_display.py", line 493, in parse_input
  File "site-packages/urwid/main_loop.py", line 403, in _update
  File "site-packages/urwid/main_loop.py", line 500, in process_input
  File "mitmproxy/tools/console/window.py", line 277, in mouse_event
  File "site-packages/urwid/container.py", line 1169, in mouse_event
  File "site-packages/urwid/container.py", line 1687, in mouse_event
  File "site-packages/urwid/container.py", line 1169, in mouse_event
  File "site-packages/urwid/container.py", line 1169, in mouse_event
  File "site-packages/urwid/container.py", line 1148, in mouse_event
  File "site-packages/urwid/container.py", line 2201, in mouse_event
  File "mitmproxy/tools/console/tabs.py", line 19, in mouse_event
  File "mitmproxy/tools/console/tabs.py", line 34, in change_tab
  File "mitmproxy/tools/console/tabs.py", line 73, in show
  File "mitmproxy/tools/console/flowview.py", line 86, in view_response
  File "mitmproxy/tools/console/flowview.py", line 160, in conn_text
  File "mitmproxy/command.py", line 221, in call
  File "mitmproxy/command.py", line 212, in call_args
  File "mitmproxy/command.py", line 101, in call
  File "mitmproxy/command.py", line 251, in wrapper
[...]

Issue #2676, opened 2017-12-14
Clicking on tab head does not focus tab


mhils:
##### Steps to reproduce the problem:

1. Open mitmproxy
2. Press `-` to open two pane layout.
3. Click on the header of the "Events" tabs.
4. The "Events" tab does not get focus.

##### Any other comments? What have you tried so far?

This is particularly annoying on Windows, where "shift tab" to switch tabs does not register under WSL. The only way to switch tabs under Windows right now is to open a command prompt and enter `console.panes.next`.

##### System information

Mitmproxy: 3.0.0dev0994 (842c9f7)
Python:    3.5.2
OpenSSL:   OpenSSL 1.0.2g  1 Mar 2016
Platform:  Linux-4.4.0-43-Microsoft-x86_64-with-Ubuntu-16.04-xenial

kajojify:
As I understood, we need mitmproxy to change focus, when we click **exactly** on the headers(Flows, Events), not on the other places of a tab? Am I correct?

mhils:
I had the circled area in mind, but I can be convinced for something else :)

![image](https://user-images.githubusercontent.com/1019198/34447184-d3c1dadc-ece1-11e7-8c73-ffada6a016bb.png)

kajojify:
What about this? 
![2017-12-31 18-16-56](https://user-images.githubusercontent.com/20267977/34462938-cfa56448-ee56-11e7-9149-19dde5eaa6ed.png)
Can we switch between "Key Bindings" - "Filter Expressions" tabs. while focusing on "Events"? 
We can still do it.

mhils:
If the events pane is focused, we shouldn't be able to modify stuff in the Help pane. So yes, this looks like a (related) bug.

kajojify:
@mhils So, do we need to open new issue or leave this bug for current issue?

mhils:
@kajojify: This may have been a misunderstanding. If I have the events pane focused as in your example, pressing "tab" (or any other key) does not switch the tab in the left pane. How do you switch between tabs while focusing the right pane?

kajojify:
@mhils by pressing left button of mouse. It is possible because of mitmproxy/tools/console/tabs.py

bksahu:
@kajojify I agree with you but it doesn't works as it is supposed to. Also by pressing left or right key doesn't works as mitmproxy/tools/console/tabs.py
[...]

PR #2746, opened 2018-01-02
Bugs with console fixed.

kajojify:
- Clicking on tab head focuses a tab now. #2676 fixed.
- We can't click on different stuff in not focused tab by pressing left button of mouse now. #2738 fixed.

### DIFF
--- a/mitmproxy/tools/console/window.py
+++ b/mitmproxy/tools/console/window.py
@@ -15,17 +15,41 @@ from mitmproxy.tools.console import grideditor
 from mitmproxy.tools.console import eventlog
 
 
+class Header(urwid.AttrWrap):
+    def __init__(self, master, title, focus):
+        self.master = master
+        super().__init__(urwid.Text(title),
+                         "heading" if focus else "heading_inactive"
+                         )
+
+    def mouse_event(self, size, event, button, col, row, focus):
+        if event == "mouse press" and button == 1:
+            if self.attr == "heading_inactive":
+                self.master.window.switch()
+                return True
+
+
+class Body(urwid.WidgetWrap):
+    def __init__(self, w, focus):
+        super().__init__(w)
+        self.f = focus
+
+    def mouse_event(self, size, event, button, col, row, focus):
+        if event == "mouse press" and button == 1:
+            if not self.f:
+                return False
+            else:
+                super().mouse_event(size, event, button, col, row, focus)
+
+
 class StackWidget(urwid.Frame):
-    def __init__(self, widget, title, focus):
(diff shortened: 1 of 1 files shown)


mhils:
This looks great, thanks. How do you feel about 299df48 as an alternative fix for #2738? This would retain the ability to switch tabs, and would also allow us to have e.g. the command list in the right view and a grideditor on the left.

kajojify:
@mhils Your fix looks more naturally and fixes #2738 :)
But it doesn't deal with a problem we discussed here https://github.com/mitmproxy/mitmproxy/issues/2676#issuecomment-354612811
We can still modify stuff in not focused tab.
For example:
screenshot from 2018-01-04 15-14-27
When being focused on **Command Reference** I can click on **Flow Details** by pressing left button of mouse and start interacting with it (without changing the colour of the header). Nevertheless all global keybindings still work only for tab with highlighted header. 
It looks a little bit strange, but if it isn't cosidered as an issue - ok.

mhils:
Right, ideally we should change the focused pane to wherever the user clicked. How about we get rid of StackWidgetHeader/StackWidgetBody and just implement `StackWidget.mouse_event` to always change focus to the pane that the user clicked on?

mitmproxy/tools/console/window.py
```diff
@@ -16,7 +16,10 @@
 
 
 class StackWidget(urwid.Frame):
-    def __init__(self, widget, title, focus):
+    def __init__(self, window, widget, title, focus):
+        self.f = focus
```
mhils: This needs a more self-explaining name. Maybe `.is_focused`?

mitmproxy/tools/console/window.py
```diff
@@ -29,6 +32,14 @@ def __init__(self, widget, title, focus):
             header=header
         )
 
+    def mouse_event(self, size, event, button, col, row, focus):
+        if event == "mouse press" and button == 1:
+            if not self.f:
+                self.window.switch()
+                return True
+            else:
+                super().mouse_event(size, event, button, col, row, focus)
```
mhils: `return super()...`

kajojify:
Talking about a thing, you described. It doesn't activate underlying action - just switching.

mhils:
Right. My question was if it would be more intuitive to also activate the underlying action, like e.g. Chrome does it.

kajojify:
@mhils Yes, it would be more intuitive, I think. So, do I need to include this feature into this PR or into next?

kajojify:
Sorry for making things more complex :) I will try not to do that again.

Issue #2720, opened 2017-12-20
console: shift+tab is broken in WSL for some Windows terminals


cortesi:
See https://github.com/Microsoft/WSL/issues/1770. This seems to affect some but not all Windows terminals. 

We use shift+tab by default to switch to the next pane in the console app. We can:

- Say that this is not our problem, and wait for upstream to fix it.
- Find a different binding for next pane - which would be a shame, because shit+tab is very natural.

@mhils what say you?

mhils:
This unfortunately affects the default terminal, which I guess the vast majority uses. 

I think the most pragmatic approach would be to bind to something else on Windows only as a temporary workaround, and unify to shift-tab once Microsoft has fixed it. How do you feel about this?

If you dislike this approach, #2676 would provide at least a somewhat convenient way to switch between panes.

PR #2757, opened 2018-01-04
Add alternative keybindings for changing the layout pane on WSL (#2720)

mhils:
This is a bit verbose, but intuitive in all layout configurations. Let's ditch this once the Windows guys have fixed it.


cortesi:
Having poked around, I now believe that the shift-tab handling is not WSL's fault, that it's a feature of Cmd used for autocompletion, and that it's unlikely ever to change. Happy for a Windows user to confirm or deny this.

My inclination would be to pick one alternative to shift-tab and then to create a global alias for it across all of our platforms, rather than handling WSL specially.

mhils:
Picking this up again - suggestions for an alternative? I found `ctrl + arrow` to work quite nicely, but I'm open to suggestions.

Issue #1833, opened 2016-12-09
mitmproxy display "\t" as "?" in request's UA


hexucheng:
My request's UA has "\t" and mitmproxy display it as "?", usually it display as blank space like fiddler or charles.

<img width="1368" alt="2016-12-10 3 24 13" src="https://cloud.githubusercontent.com/assets/9985186/21061654/366dd14c-be88-11e6-9153-22675532e3f9.png">

mhils:
Thanks! Please use the issue template next time so that we know your mitmproxy version right away.

This still affects master:
```
pathoc -c example.com:80 localhost:8080 'get:/:u"foo\tbar"'
```

mhils:
This seems to be an urwid issue:
```python
import urwid

txt = urwid.Text(u"Hello\tWorld")
fill = urwid.Filler(txt, 'top')
loop = urwid.MainLoop(fill)
loop.run()
```

hexucheng:
mitmproxy version is 0.18.2.

kajojify:
Are there any problems with fixing this issue?

kajojify:
We can just change https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/tools/console/flowview.py#L156 to
`[(h + ":", v.encode("unicode_escape")) for (h, v) in conn.headers.items(multi=True)]` or 
`[(h + ":", repr(v)[1:-1]) for (h, v) in conn.headers.items(multi=True)]`

If we use pathoc as @mhils suggested `pathoc -c example.com:80 localhost:8080 'get:/:u"foo\tbar"'`:
![2018-01-03 17-06-44](https://user-images.githubusercontent.com/20267977/34525693-813c9c00-f0a8-11e7-9f06-83d62eeed84f.png)

What do you think?

PR #2758, opened 2018-01-05
Fixing #1833

kajojify:
Fix for #1833.
There is also more readable 
`[(h + ":", v.encode("unicode_escape")) for (h, v) in conn.headers.items(multi=True)]`,
 but I thought, that there was no a good reason to change `str` type of second element of tuple to `bytes`.

Applying `pathoc -c example.com:80 localhost:8080 'get:/:u"f\too\tba\tr"'`, we will get:
screenshot from 2018-01-05 14-59-58

### DIFF
--- a/mitmproxy/tools/console/flowview.py
+++ b/mitmproxy/tools/console/flowview.py
@@ -153,7 +153,8 @@ class FlowDetails(tabs.Tabs):
     def conn_text(self, conn):
         if conn:
             txt = common.format_keyvals(
-                [(h + ":", v) for (h, v) in conn.headers.items(multi=True)],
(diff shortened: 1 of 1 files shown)


mhils:
Sorry, I missed https://github.com/mitmproxy/mitmproxy/issues/1833#issuecomment-354957287 earlier. These kinds of issues are ridicolously hard to get right and I should have warned you 😄. Your fix is pretty close, but it will also now _sometimes_ escape " or ', which really isn't what we want. I commited a fix in #2759, which starts at the Header's underlying bytes and works up from there.

kajojify:
@mhils ok, thanks :)

PR #2764, opened 2018-01-05
rename attributes, function and class names: s/ssl/tls/

Kriechi:
SSL is an outdated protocol superseeded by TLS: https://tools.ietf.org/html/rfc7568

Although the commonly used library is called OpenSSL, it is no reason to still use outdated language for attributes, function and class names. Some parts in the code are kept with their original name because they refer to OpenSSL (or other) library interfaces.


mitmproxy/io/compat.py
```diff
@@ -112,8 +112,22 @@ def convert_200_300(data):
     data["version"] = (3, 0, 0)
     data["client_conn"]["mitmcert"] = None
     data["server_conn"]["tls_version"] = None
+    data["client_conn"]["tls_established"] = data["server_conn"]["ssl_established"] or None
```
mhils: This should be a simple `.pop`, the attribute is always present.

mitmproxy/io/compat.py
```diff
@@ -112,8 +112,22 @@ def convert_200_300(data):
...
+    data["client_conn"]["timestamp_tls_setup"] = data["server_conn"]["timestamp_ssl_setup"] or None
+    data["server_conn"]["tls_established"] = data["server_conn"]["ssl_established"] or None
+    data["server_conn"]["timestamp_tls_setup"] = data["server_conn"]["timestamp_ssl_setup"] or None
+    del data["client_conn"]["ssl_established"]
```
mhils: ...and this is not required then :)

mitmproxy/io/compat.py
```diff
@@ -112,8 +112,22 @@ def convert_200_300(data):
...
+    del data["server_conn"]["ssl_established"]
+    del data["server_conn"]["timestamp_ssl_setup"]
+
     if data["server_conn"]["via"]:
         data["server_conn"]["via"]["tls_version"] = None
+        data["server_conn"]["via"]["tls_established"] = data["server_conn"]["via"]["ssl_established"] or None
+        data["server_conn"]["via"]["timestamp_tls_setup"] = data["server_conn"]["via"]["timestamp_ssl_setup"] or None
+        del data["server_conn"]["via"]["ssl_established"]
+        del data["server_conn"]["via"]["timestamp_ssl_setup"]
+
```
mhils: This should be a new flow file version in version.py so that dumps created with developer snapshots remain working. That's why we split version and flow file version. :)

mitmproxy/io/compat.py
```diff
@@ -149,6 +154,25 @@ def convert_4_5(data):
...
 
+def convert_5_6(data):
+    data["version"] = 6
+    data["client_conn"]["tls_established"] = data["client_conn"].pop("ssl_established")
+    data["client_conn"]["timestamp_tls_setup"] = data["client_conn"].pop("timestamp_ssl_setup")
+    data["server_conn"]["tls_established"] = data["server_conn"].pop("ssl_established")
+    data["server_conn"]["timestamp_tls_setup"] = data["server_conn"].pop("timestamp_ssl_setup")
+    if data["server_conn"]["via"]:
+        data["server_conn"]["via"]["tls_established"] = data["server_conn"]["via"].pop("ssl_established", None)
+        data["server_conn"]["via"]["timestamp_tls_setup"] = data["server_conn"]["via"].pop("timestamp_ssl_setup", None)
```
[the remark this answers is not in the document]
Kriechi: Sorry, it's debug leftover - and gone now.

PR #2769, opened 2018-01-07
Fix #2688

### LINKED ISSUE #2688: console: ctx.log.alert output is not seen for comands
Issue a command that emits an alert from the console commander. Note that the alert output doesn't appear.

kajojify:
Alert output appears in Status Bar now.
I decided to set **magenta colour** for alert output. I am not good at colours, so we always can change the colour, if you think it isn't suitable. 
Light background
light

Dark background
lowdark

Mitmdump
image


mhils:
Looks great! 🎉

PR #2778, opened 2018-01-10
Fix #2768

### LINKED ISSUE #2768: Traceback appears, when trying to set mitmproxy's address as upstream server for reverse/upstream mode

1. Run mitmproxy in **reverse** or **upstream** mode, using its own address as  upstream server address:
`mitmproxy --mode reverse:http://127.0.0.1:8080` or
`mitmproxy --mode upstream:http://127.0.0.1:8080`
2. Make a request using pathoc `pathoc 127.0.0.1:8080 "get:/"` or a browser.

I am seeing:
screenshot from 2018-01-06 23-29-35

https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/proxy/protocol/base.py#L115
should be handled.

##### System information

Mitmproxy: 3.0.0.dev1101 (commit d9d4d15) binary
Python: 3.5.2
OpenSSL: OpenSSL 1.1.0g 2 Nov 2017
Platform: Linux-4.4.0-104-generic-x86_64-with-debian-stretch-sid

kajojify:


test/mitmproxy/test_proxy.py
```diff
@@ -80,6 +82,29 @@ def test_simple(self):
 
 class TestConnectionHandler:
 
+    def check_proxy_for_mode(self, mode):
```
mhils: Let's add a small comment here that indicates what this test is doing:
```python
def check_proxy_for_mode(self, mode):
    "Check that ..."
```

test/mitmproxy/test_proxy.py
```diff
@@ -80,6 +82,29 @@ def test_simple(self):
...
+        try:
+            c.handle()
+        except exceptions.ProtocolException as e:
+            pytest.fail(repr(e))
+
+        c = pathoc.Pathoc(("127.0.0.1", 8080))
+        try:
+            c.connect()
+        except exceptions.TcpException:
+            pass
```
mhils: ...speaking of which: What does actually happen now? We should have an assertion here that a log statement is generated or something like that :)

kajojify: The traceback appears, when we run mitmproxy in reverse/upstream mode, setting mitmproxy's address as upstream server's address and then making a request using **pathoc**.

First of all, I create `ConnectionHandler` object, according to this: https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/proxy/server.py#L62-L68
Second of all, I make a request using pathoc to reproduce the issue, we are talking about. 

The thing is two exceptions are raised: `mitmproxy.exceptions.ProtocolException` from mitmproxy side and `mitmproxy.exceptions.TcpException` from pathoc side. I decided to check our code only for `ProtocolException`, so I wrote: _if ProtocolException is raised (it hasn't to be raised), test must be failed_. Since pathoc exception not a business of this issue, I just skip it. That's it.

kajojify:
@mhils 1) I rewrote test. Now it checks, whether  "The proxy shall not connect to itself."  is in logentry or not.
2) When I try to make a request to mitmproxy in `reverse/upstream` mode using pathoc, mitmproxy tries to handle the connection https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/proxy/server.py#L114-L118
and gives me such events:
image

I think it is expected behaviour. 
I am not a lazy, I am just curious :) 
Of course it would be better to write it in `test_server.TestReverse`, but my test ins't nonsense, I think.
Looking at `test_server.TestReverse` now...

test/mitmproxy/test_proxy.py
```diff
@@ -80,6 +82,37 @@ def test_simple(self):
...
+        Checks that we get correct log message in specific mode
+        """
+        log_msg = "The proxy shall not connect to itself."
+        opts = options.Options(mode=mode)
+        pconf = config.ProxyConfig(opts)
+
+        channel = mock.Mock()
+
+        def tell(mtype, m):
+            if m.level == "warn":
```
Kriechi: What happens in this test if there is no "warn" message?
Maybe set a bool flag `warning_triggered=True` and then at the end of the test assert that.

kajojify: @Kriechi yes, you're right!

mhils:
> 2) When I try to make a request to mitmproxy in reverse/upstream mode using pathoc

My point is exactly that you are not doing that. 😉 In [line 113](https://github.com/mitmproxy/mitmproxy/pull/2778/files#diff-7b99050809dd83af3af1bb034280352bR113) you try to connect to `("127.0.0.1", 8080)`, but you never started a server on port 8080 here. Of course, pathoc returns a TcpException because it cannot connect. Had there been a server, you  could actually connect and get a HTTP error message in return.

PR #2784, opened 2018-01-11
Update GIF parser

### LINKED ISSUE #2658: Traceback when viewing certain gifs
Open the attached flow dump in mitmproxy console. When viewing the content, observe the traceback that flashes to screen. A number of issues here:

- [ ] Kaitai should be able to parse this gif. It's not technically correct, but I've seen this often, especially with single-pixel tracking gifs. 
- *Even if Kaitai fails*, we shouldn't crash - the user should see an "invalid image" message (or something more informative). (#2698)
- *Even if we crash*, the traceback shouldn't be flashed to screen and disappear. This is an issue with the way we handle super-long error events in console. (#2699)

[bar.log](https://github.com/mitmproxy/mitmproxy/files/1545955/bar.log)

ujjwal96:
Fixes #2658

### DIFF
--- a/mitmproxy/contrib/kaitaistruct/gif.py
+++ b/mitmproxy/contrib/kaitaistruct/gif.py
@@ -35,9 +35,11 @@ class Gif(KaitaiStruct):
             self.global_color_table = self._root.ColorTable(io, self, self._root)
 
         self.blocks = []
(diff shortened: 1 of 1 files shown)


mhils:
Awesome, thanks @ujjwal96! 😃

Issue #2788, opened 2018-01-12
Store ClientHello Extensions


mhils:
@0x4d31 was interested in adding JA3 hash support to mitmproxy, so we need to give him a way to access all TLS extensions sent during the initial handshake.

abbbe:
@0x4D31 @mhils any clue if that story from 2018 has resulted in any publicly available mitmproxy addon for JA3 fingerprinting?

raxod502:
At the very least, is it possible to export the raw packets so that they can be processed by another tool? https://github.com/muzuiget/mitmpcap drops all the TLS data.

PR #2789, opened 2018-01-12
Move ClientHello and related functions to mitmproxy.net.tls

mhils:
This PR moves the ClientHello parsing code in .proxy.protocol to .net.tls, which is a first step to make #2788 possible.


Kriechi:
General 👍
Although I'm not very fond of weird imports like `... import tls as net_tls` and such... ¯\_(ツ)_/¯

PR #2802, opened 2018-01-18
Add share.flows command, fix #2779

### LINKED ISSUE #2779: Implement `share` command
@MatthewShao implemented a really cool static viewer based on mitmweb last summer: http://honeynet.org/node/1359.

Long story short, one can upload a mitmproxy flow file at share.mitmproxy.org and will get a shareable link in return. 😃 To make this more accessible, I think it would be awesome to have a mitmproxy command that takes a bunch of flows as argument, uploads them directly from within mitmproxy, and returns a URL to share.

kira0204:
Fixes #2779 . Added the command 'upload.file'


mitmproxy/addons/__init__.py
```diff
@@ -20,6 +20,7 @@
 from mitmproxy.addons import streambodies
 from mitmproxy.addons import save
 from mitmproxy.addons import upstream_auth
+from mitmproxy.addons import upload
```
mhils: Let's call this `share` and not `upload` - the user wants to share their flows, uploading is just the implementation of that. :)

mitmproxy/addons/upload.py
```diff
@@ -0,0 +1,124 @@
...
+import mitmproxy.types
+
+from mitmproxy.addons import multipart
+
+
+class Upload:
+    def __init__(self):
+        self.stream = None
+        self.filt = None
+        self.active_flows = set()  # type: Set[flow.Flow]
```
mhils: I don't understand why all this is here. The `share` command accepts a list of flows which should be uploaded, everything else here is not necessary.

mitmproxy/addons/upload.py
```diff
@@ -0,0 +1,124 @@
...
+
+        return encoded
+
+    @command.command("upload.file")
+    def upload(self, flows: typing.Sequence[flow.Flow]) -> None:
+
+        d = datetime.datetime.utcnow()
+        id = self.base36encode(int(time.mktime(d.timetuple()) * 1000 * random.random()))[0:7]
+        try:
+            f = self.open_file(id)  # Making temporary file to store the flows
```
mhils: As mentioned in https://github.com/mitmproxy/mitmproxy/issues/2779#issuecomment-358385276, you can pass a BytesIO object to FlowWriter. No need to write anything to disk.

mitmproxy/addons/multipart.py
```diff
@@ -0,0 +1,43 @@
+from urllib.request import Request, urlopen
+from urllib.error import HTTPError, URLError
+import mimetypes
+
+
+def get_content_type(filename):
+    return mimetypes.guess_type(filename)[0] or 'application/octet-stream'
```
mhils: This isn't really necessary, the content type is constant.

mitmproxy/addons/multipart.py
```diff
@@ -0,0 +1,43 @@
...
+    L.append(content)
+    L.append(b'--' + LIMIT + b'--')
+    L.append(b'')
+    body = CRLF.join(L)
+    content_type = b'multipart/form-data; boundary=%b' % LIMIT
+    return content_type, body
+
+def post_multipart(host, filename, content):
+    content_type, body = encode_multipart_formdata(filename, content)
+    req = Request("http://upload.share.mitmproxy.org.s3.amazonaws.com", data=body, method='POST')
```
mhils: Let's use Python 3's http.client, urllib really is a bit of a pain in the ass. :-)

mitmproxy/addons/multipart.py
```diff
@@ -0,0 +1,43 @@
...
+    L.append(b'Content-Type: %b' % get_content_type(filename).encode("utf-8"))
+    L.append(b'')
+    L.append(content)
+    L.append(b'--' + LIMIT + b'--')
+    L.append(b'')
+    body = CRLF.join(L)
+    content_type = b'multipart/form-data; boundary=%b' % LIMIT
+    return content_type, body
+
+def post_multipart(host, filename, content):
```
mhils: This function can live in the upload addon once you have removed the unnecessary cruft there :)

kira0204:
I will make all the changes, this was a learning experience, thanks

mitmproxy/addons/__init__.py
kira0204: Okay

mitmproxy/addons/upload.py
kira0204: I am not able to understand mitmproxy/addons/save.py works. Since I did not want to break anything, I did not change any of the methods. for example I couldn't figure out why tcp_start and tcp_end methods are there? Can you please give little explaination on this

mhils: The save addon can be configured to continually stream all requests to a file. In this case we use the tcp_start/tcp_end hooks to register when a flow has been finished.

mitmproxy/addons/multipart.py
kira0204: Yeah okay

mitmproxy/addons/upload.py
kira0204: Can you tell me how it can be configured.

mitmproxy/addons/multipart.py
kira0204: Done, but can you explain why do we prefer http.client over urllib

mitmproxy/addons/upload.py
kira0204: Okay thanks

mitmproxy/addons/share.py
```diff
@@ -27,12 +28,6 @@ def open_file(self, path):
         path = os.path.expanduser(path)
         return open(path, mode)
 
-    def read_file(self, path):
-        path = os.path.expanduser(path)
-        with open(path, "rb") as f:
-            content = f.read()
-        return content
-
     def start_stream_to_path(self, path, flt):
```
kira0204: I need to look into these methods more before removing them. The next commit will be without all the unnecessary methods.

kira0204:
@mhils Regarding the test case, we have two ways to do it:

1. Take the response that `post_multipart` will provide and use `assert` to check whether it matches with `"URL: share.mitmproxy.org\" + u_id`, but there is a better way to do this

2. Send a `GET` request at `share.mitmproxy.org\unique_id` and then check the repsonse code. I was going forward with this approach but wanted your advice on this before I write the test file

mhils:
@kira0204: Keep in mind that the testsuite should definitely run offline - you need to mock external communication. 😉

kira0204:
It will be a little challenging and exciting for me :) Lets do it

mitmproxy/addons/share.py
```diff
@@ -0,0 +1,79 @@
...
+        u_id = self.base36encode(int(time.mktime(d.timetuple()) * 1000 * random.random()))[0:7]
+        f = _io.BytesIO()
+        stream = io.FlowWriter(f)
+        for i in flows:
+            stream.add(i)
+        f.seek(0)
+        content = f.read()
+        res = self.post_multipart('upload.share.mitmproxy.org.s3.amazonaws.com', u_id, content)
+        self.res = res
+        self.u_id = u_id
```
Kriechi: Are these variables really necessary? Right now you only use them to access the result in the tests, would is not really a nice design. I think just logging both the result string + uid with `ctx.log.alert` is enough.
You can assert that a specific log message was received with `assert tctx.master.has_log("some message")`

mitmproxy/addons/share.py
```diff
@@ -0,0 +1,79 @@
...
+        l.append(b'--' + LIMIT)
+        l.append(b'Content-Disposition: form-data; name="file"; filename="%b"' % filename.encode("utf-8"))
+        l.append(b'Content-Type: application/octet-stream')
+        l.append(b'')
+        l.append(content)
+        l.append(b'--' + LIMIT + b'--')
+        l.append(b'')
+        body = CRLF.join(l)
+        content_type = b'multipart/form-data; boundary=%b' % LIMIT
+        return content_type, body
```
Kriechi: We already depend on the `requests` package, why not use its support for multipart uploads? http://docs.python-requests.org/en/master/user/quickstart/#post-a-multipart-encoded-file

mitmproxy/addons/share.py
kira0204: No they aren't necessary as such and even I was wondering that but it made the test more refined.

mitmproxy/addons/share.py
kira0204: Apart from `requests`, there was `urllib` also implemented in some of the files but @mhils recommended to use http.client.

Kriechi: Yes, we use `urllib` only for low-level parsing of HTTP-related things.
But in this case (upload files to a server), I think using `requests` is the better and easier choice.

mhils:
@resam: AFAIK core doesn't depend on requests, and I'd be very much
inclined to keep it that way.

Kriechi:
Ah yes - you are right. I only looked at setup.py and our install-dependencies.
The only places we actually import it are examples and tests.

Ok, then I'm fine with building our own multipart stuff.

mitmproxy/addons/share.py
kira0204: So we are having our own `multipart`?  :)

mitmproxy/addons/share.py
kira0204: I will change it asap :) Thankyou for the review

mitmproxy/addons/share.py
Kriechi: Yes - lets keep it.

mitmproxy/addons/share.py
kira0204: @Kriechi I was wondering how would I be able to check the log message using `tctx.master.has_log("some message")` because in the test case I wont be knowing what `u_id` will be if i don't introduce it as a class varible. Any suggestions?

Kriechi: The ID itself is not important. Just check something was written to the log. If you want something nice, try using a regex, but you need to adapt the `has_log` function to accept one first!

kira0204: Okay Done :)

mitmproxy/addons/share.py
```diff
@@ -0,0 +1,73 @@
...
+        while integer > 0:
+            integer, remainder = divmod(integer, 36)
+            encoded = chars[remainder] + encoded
+
+        return encoded
+
+    @command.command("share.flows")
+    def share(self, flows: typing.Sequence[flow.Flow]) -> None:
+        d = datetime.datetime.utcnow()
+        u_id = self.base36encode(int(time.mktime(d.timetuple()) * 1000 * random.random()))[0:7]
```
mhils: We can do this a bit more Pythonic (and skip the whole base36encode function):
```python
"".join(
    random.choice(string.ascii_lowercase + string.digits) 
    for _ in range(7)
)
```

mitmproxy/addons/share.py
```diff
@@ -0,0 +1,73 @@
+import typing
+import random
+import datetime
+import time
+import _io
```
mhils: This is a really bad idea, we definitely want to import `io` itself here. You can replace `from mitmproxy import io` with `import mitmproxy.io`.

mitmproxy/addons/share.py
```diff
@@ -0,0 +1,73 @@
...
+
+        return encoded
+
+    @command.command("share.flows")
+    def share(self, flows: typing.Sequence[flow.Flow]) -> None:
+        d = datetime.datetime.utcnow()
+        u_id = self.base36encode(int(time.mktime(d.timetuple()) * 1000 * random.random()))[0:7]
+        f = _io.BytesIO()
+        stream = io.FlowWriter(f)
+        for i in flows:
```
mhils: Nit-pick: `for f in flows:`, `for x in flows:`, or `for flow in flows:` - `i` is usually expected to be a numeric index variable.

mitmproxy/addons/share.py
```diff
@@ -0,0 +1,73 @@
...
+import http.client
+
+from mitmproxy import command
+from mitmproxy import io
+from mitmproxy import ctx
+from mitmproxy import flow
+
+
+class Share:
+    def encode_multipart_formdata(self, filename, content):
```
mhils: Let's add type annotations!
```python
def encode_multipart_formdata(self, filename: str, content: bytes) -> typing.Tuple[bytes, bytes]:
```

mitmproxy/addons/share.py
```diff
@@ -0,0 +1,73 @@
...
+        l.append(content)
+        l.append(b'--' + LIMIT + b'--')
+        l.append(b'')
+        body = CRLF.join(l)
+        content_type = b'multipart/form-data; boundary=%b' % LIMIT
+        return content_type, body
+
+    def post_multipart(self, host, filename, content):
+        content_type, body = self.encode_multipart_formdata(filename, content)
+        conn = http.client.HTTPConnection(host, 80)
```
mhils: This definitely needs to be a HTTPS connection here!

mitmproxy/addons/share.py
```diff
@@ -0,0 +1,73 @@
...
+        headers = {'content-type': content_type, 'content-length': str(len(body))}
```
mhils: Content-Length is automatically set by http.client, no need to add it here.

mitmproxy/addons/share.py
```diff
@@ -0,0 +1,73 @@
...
+        try:
+            conn.request("POST", "", body, headers)
+        except http.client.CannotSendRequest:
+            return 'We failed to reach a server.'
```
mhils: Returning error messages as strings is a really bad practice. We get a much nicer API if this function either returns a correct URL or raises an IOError.

mitmproxy/addons/share.py
```diff
@@ -0,0 +1,73 @@
...
+        content_type = b'multipart/form-data; boundary=%b' % LIMIT
+        return content_type, body
+
+    def post_multipart(self, host, filename, content):
+        content_type, body = self.encode_multipart_formdata(filename, content)
+        conn = http.client.HTTPConnection(host, 80)
+        headers = {'content-type': content_type, 'content-length': str(len(body))}
+        try:
+            conn.request("POST", "", body, headers)
+        except http.client.CannotSendRequest:
```
mhils: This is catching the wrong error.
```python
>>> x = http.client.HTTPConnection("unknown", 80)
>>> x.request("POST", "")
```
raises a `socket.gaierror`, which would not be caught here.

mitmproxy/addons/share.py
```diff
@@ -0,0 +1,73 @@
...
+            return 'We failed to reach a server.'
+        try:
+            conn.getresponse()
```
mhils: We need to check the returned HTTP response here - this could very well be a server error.

mitmproxy/addons/share.py
```diff
@@ -0,0 +1,73 @@
...
+        except http.client.RemoteDisconnected:
+            return 'The server couldn\'t fulfill the request.'
+        else:
+            conn.close()
```
mhils: The logic here is flawed - we should always call close, no matter if the connection succeeded. 

The number of comments on this function is getting a bit overboard. In total it should look roughly like this:
```python
    def post_multipart(self, host: str, filename: str, content: bytes) -> str:
        """
        Upload flows to the specified S3 server.

        Returns:
            - The share URL, if upload is successful.
        Raises:
            - IOError, otherwise.
        """
        content_type, body = self.encode_multipart_formdata(filename, content)
        headers = {'content-type': content_type}

        conn = http.client.HTTPSConnection(host)

        try:
            conn.request("POST", "/", body, headers)
            resp = conn.getresponse()
        finally:
            conn.close()

        if resp.status != 200:
            if resp.reason:
                reason = resp.reason
            else:
                 reason = status_codes.RESPONSES.get(resp.status, str(resp.status))
            raise IOError(reason)
        return "https://share.mitmproxy.org/%s" % filename
```

`share` needs to properly catch IOErrors now of course.

mitmproxy/addons/share.py
kira0204: I didn't know that `_io` is the internal code that `io` uses. I will change it

mitmproxy/addons/share.py
kira0204: This will also decrease the need to import `datetime` and `time` modules. Thanks

mitmproxy/addons/share.py
mhils: Sure, please feel free to change that as well.

mitmproxy/addons/share.py
kira0204: Thankyou for the advice :) I will implement it right away

mitmproxy/addons/share.py
kira0204: Hi, HTTPS Connection is throwing an error

'''
  File "/home/arushit/Desktop/mitmproxy/mitmproxy/tools/console/master.py", line 216, in run
    self.loop.run()
  File "/home/arushit/Desktop/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 278, in run
    self._run()
  File "/home/arushit/Desktop/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 376, in _run
    self.event_loop.run()
  File "/home/arushit/Desktop/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 682, in run
    self._loop()
  File "/home/arushit/Desktop/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 719, in _loop
    self._watch_files[fd]()
  File "/home/arushit/Desktop/mitmproxy/venv/lib/python3.5/site-packages/urwid/raw_display.py", line 393, in <lambda>
    event_loop, callback, self.get_available_raw_input())
  File "/home/arushit/Desktop/mitmproxy/venv/lib/python3.5/site-packages/urwid/raw_display.py", line 493, in parse_input
    callback(processed, processed_codes)
  File "/home/arushit/Desktop/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 403, in _update
    self.process_input(keys)
  File "/home/arushit/Desktop/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 503, in process_input
    k = self._topmost_widget.keypress(self.screen_size, k)
  File "/home/arushit/Desktop/mitmproxy/mitmproxy/tools/console/window.py", line 308, in keypress
    k = super().keypress(size, k)
  File "/home/arushit/Desktop/mitmproxy/venv/lib/python3.5/site-packages/urwid/container.py", line 1116, in keypress
    return self.footer.keypress((maxcol,),key)
  File "/home/arushit/Desktop/mitmproxy/mitmproxy/tools/console/statusbar.py", line 149, in keypress
    return self.ab.keypress(*args, **kwargs)
  File "/home/arushit/Desktop/mitmproxy/mitmproxy/tools/console/statusbar.py", line 104, in keypress
    self.prompt_execute(self._w.get_edit_text())
  File "/home/arushit/Desktop/mitmproxy/mitmproxy/tools/console/statusbar.py", line 124, in prompt_execute
    msg = p(txt)
  File "/home/arushit/Desktop/mitmproxy/mitmproxy/tools/console/commandexecutor.py", line 17, in __call__
    ret = self.master.commands.call(cmd)
  File "/home/arushit/Desktop/mitmproxy/mitmproxy/command.py", line 221, in call
    return self.call_args(parts[0], parts[1:])
  File "/home/arushit/Desktop/mitmproxy/mitmproxy/command.py", line 212, in call_args
    return self.commands[path].call(args)
  File "/home/arushit/Desktop/mitmproxy/mitmproxy/command.py", line 101, in call
    ret = self.func(*pargs)
  File "/home/arushit/Desktop/mitmproxy/mitmproxy/command.py", line 251, in wrapper
  File "/home/arushit/Desktop/mitmproxy/mitmproxy/addons/share.py", line 63, in share
    res = self.post_multipart('upload.share.mitmproxy.org.s3.amazonaws.com', u_id, content)
  File "/home/arushit/Desktop/mitmproxy/mitmproxy/addons/share.py", line 40, in post_multipart
    conn.request("POST", "", body, headers)
  File "/usr/lib/python3.5/http/client.py", line 1106, in request
    self._send_request(method, url, body, headers)
  File "/usr/lib/python3.5/http/client.py", line 1151, in _send_request
    self.endheaders(body)
  File "/usr/lib/python3.5/http/client.py", line 1102, in endheaders
    self._send_output(message_body)
  File "/usr/lib/python3.5/http/client.py", line 934, in _send_output
    self.send(msg)
  File "/usr/lib/python3.5/http/client.py", line 877, in send
    self.connect()
  File "/usr/lib/python3.5/http/client.py", line 1260, in connect
    server_hostname=server_hostname)
  File "/usr/lib/python3.5/ssl.py", line 377, in wrap_socket
    _context=self)
  File "/usr/lib/python3.5/ssl.py", line 752, in __init__
    self.do_handshake()
  File "/usr/lib/python3.5/ssl.py", line 988, in do_handshake
    self._sslobj.do_handshake()
  File "/usr/lib/python3.5/ssl.py", line 638, in do_handshake
    match_hostname(self.getpeercert(), self.server_hostname)
  File "/usr/lib/python3.5/ssl.py", line 297, in match_hostname
    % (hostname, ', '.join(map(repr, dnsnames))))
ssl.CertificateError: hostname 'upload.share.mitmproxy.org.s3.amazonaws.com' doesn't match either of '*.s3.amazonaws.com', 's3.amazonaws.com'
'''

Also: https://github.com/boto/boto/issues/2836
Should I monkey patch ssl?

mhils: Ugh. Please leave it HTTP for now but add a big prominent FIXME in the code. We'll merge this to a separate branch then until we fix it on our end.

kira0204: Okay

mitmproxy/addons/share.py
kira0204: Hello sir, I couldn't understand why are we checking if `resp.reason` is empty or not, don't we always receive a reason?

mhils: Not always, e.g. HTTP/2.0 does not include a reason in responses. It would be awkward to have an error message that reads like this: "Error uploading flow: "

kira0204:
`test_xss_scanner.py` depends upon external communication due to which earlier build failed. I will fix it in #2819

kira0204:
@mhils Should I add an keybinding to this API?

kira0204:
Hi, is there anything left in this to be done? :)

mhils:
No, this looks good - thanks! 🍰 As mentioned on Slack, we still need to work on some server-side optimisations, so I'm moving this into the mitmproxy repo on a separate branch. 😃

kira0204:
Yayyyy, so happy to see this. 💯

PR #2803, opened 2018-01-19
Persist flowfilter when editing it, fix #2777

### LINKED ISSUE #2777: Flow filters don't persist between invocations of `f`
1 - Hit `f` type in filter, make a typo
2 - Hit 'f' again to fix the typo

##### Expected behaviour
On second hit of f, when filter in use already the command line shows the current filter.

##### Current behaviour
On second hit of f, whn filter in use already, the filter is cleared.

##### System information
Mitmproxy 3 RC2

kajojify:
The issue says only about _view filter_. But I think we should extend it. The issue is relevant to intercept option(`i`) and stream to file(`W`) as well.

Now after pressing `f`, `i` or `W` and setting any value, you can press the button again and correct a typo or set new similar value without reprinting the whole value from scratch.

### DIFF
--- a/mitmproxy/tools/console/commander/commander.py
+++ b/mitmproxy/tools/console/commander/commander.py
@@ -47,9 +47,12 @@ CompletionState = typing.NamedTuple(
 )
 
 
-class CommandBuffer():
+class CommandBuffer:
     def __init__(self, master: mitmproxy.master.Master, start: str = "") -> None:
         self.master = master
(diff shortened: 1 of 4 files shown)


mhils:
Let's go ahead with a `console.command.set` command for now, removing this should always be easy.  `console.command.set` just needs to accept the option name and then get the option current value and call `console.command set {name}={value}`. Makes sense?

mhils:
👍

Issue #2783, opened 2018-01-11
Mitmproxy crashes, when trying to export something in the file with incorrect path


kajojify:
Related to #2750.
##### Steps to reproduce the problem:

1. Run mitmproxy.
2. Start intercept something.
3.  Press `e` to export specific flow to the file. 
_input   -    the corresponding exception_

`/file.txt`   -   `PermissionError: [Errno 13] Permission denied: '/file.txt'`
`/home`   -    `IsADirectoryError: [Errno 21] Is a directory: '/home'`
`/ome/file.txt`   -   `FileNotFoundError: [Errno 2] No such file or directory: '/ome/file.txt'`

4) The same exceptions will be relevant, if you press `b` and try to save  reponse body to the file with these incorect paths.

##### Any other comments? What have you tried so far?

PR for related #2750  - https://github.com/mitmproxy/mitmproxy/pull/2761

##### System information
Mitmproxy: 3.0.0.dev0004 (com
mit 68c32d8) 
Python:    3.5.2
OpenSSL:   OpenSSL 1.1.0g  2 Nov 2017
Platform:  Linux-4.4.0-104-generic-x86_64-with-Ubuntu-16.04-xenial

kajojify:
Working on it.

kajojify:
Closed in https://github.com/mitmproxy/mitmproxy/pull/2809

PR #2761, opened 2018-01-05
Catch file exceptions

ritiek:
Fixes #2750.

This PR catches and logs any exceptions when exporting flow either by selecting a specific flow or through the parent user interface (and it also generated an `.egg` file in `dist/` for me, so `dist/` goes to `.gitignore`).

Issue #2750, opened 2018-01-03
Mitmproxy crashes, when trying to export the flow in the folder with incorrect path


kajojify:
##### Steps to reproduce the problem:

1. Run mitmproxy without root privileges.
2. Start intercepting something.
3. Choose any flow.
4. Press `e`. Choose a part of the flow - `cookies` for example. 
5. Press `w` to save `cookies`.
6. Try to save `cookies` to the directory which requires only root permissions.

I am getting:
```
Traceback (most recent call last):
  File "/home/kajoj/Mitmproxy/mitmproxy/mitmproxy/tools/console/master.py", line 216, in run
    self.loop.run()
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 278, in run
    self._run()
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 376, in _run
    self.event_loop.run()
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 682, in run
    self._loop()
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 719, in _loop
    self._watch_files[fd]()
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/raw_display.py", line 393, in <lambda>
    event_loop, callback, self.get_available_raw_input())
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/raw_display.py", line 493, in parse_input
    callback(processed, processed_codes)
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 403, in _update
    self.process_input(keys)
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 503, in process_input
    k = self._topmost_widget.keypress(self.screen_size, k)
  File "/home/kajoj/Mitmproxy/mitmproxy/mitmproxy/tools/console/window.py", line 318, in keypress
    k = super().keypress(size, k)
[...]

PR #2809, opened 2018-01-22
Fix for #2750 and #2783

kajojify:
Continuation of https://github.com/mitmproxy/mitmproxy/pull/2761 
I added tests with mocked `open` function. I hope it is the right way to test such things. 

Now, when you press `w`, `e` or `b` and input invalid file path, you will get something like:

image

### DIFF
--- a/mitmproxy/addons/cut.py
+++ b/mitmproxy/addons/cut.py
@@ -109,7 +109,7 @@ class Cut:
                             [strutils.always_str(x) or "" for x in vals]  # type: ignore
                         )
                 ctx.log.alert("Saved %s cuts over %d flows as CSV." % (len(cuts), len(flows)))
-        except (PermissionError, IsADirectoryError, FileNotFoundError) as e:
+        except IOError as e:
             ctx.log.error(str(e))
 
     @command.command("cut.clip")
--- a/mitmproxy/addons/cut.py
+++ b/mitmproxy/addons/cut.py
@@ -88,26 +88,29 @@ class Cut:
         if path.startswith("+"):
             append = True
             path = mitmproxy.types.Path(path[1:])
-        if len(cuts) == 1 and len(flows) == 1:
-            with open(path, "ab" if append else "wb") as fp:
-                if fp.tell() > 0:
-                    # We're appending to a file that already exists and has content
-                    fp.write(b"\n")
-                v = extract(cuts[0], flows[0])
-                if isinstance(v, bytes):
-                    fp.write(v)
-                else:
-                    fp.write(v.encode("utf8"))
-            ctx.log.alert("Saved single cut.")
-        else:
-            with open(path, "a" if append else "w", newline='', encoding="utf8") as fp:
-                writer = csv.writer(fp)
-                for f in flows:
-                    vals = [extract(c, f) for c in cuts]
-                    writer.writerow(
-                        [strutils.always_str(x) or "" for x in vals]  # type: ignore
-                    )
-            ctx.log.alert("Saved %s cuts over %d flows as CSV." % (len(cuts), len(flows)))
+        try:
+            if len(cuts) == 1 and len(flows) == 1:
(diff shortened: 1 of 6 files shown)


test/mitmproxy/addons/test_export.py
```diff
@@ -93,6 +93,17 @@ def test_export(tmpdir):
         assert qr(f)
         os.unlink(f)
 
+        with mock.patch("mitmproxy.addons.export.open") as m:
+            m.side_effect = [PermissionError("Permission denied"),
+                             IsADirectoryError("Is a directory"),
+                             FileNotFoundError("No such file or directory")]
+            for i in range(3):
```
Kriechi: I would prefer looping through each error individually. You could make use of a new test function and `pytest.parametrize` to set the side effect (exception), and also the expected log message.

mitmproxy/addons/cut.py
```diff
@@ -88,26 +88,29 @@ def save(
...
+            else:
+                with open(path, "a" if append else "w", newline='', encoding="utf8") as fp:
+                    writer = csv.writer(fp)
+                    for f in flows:
+                        vals = [extract(c, f) for c in cuts]
+                        writer.writerow(
+                            [strutils.always_str(x) or "" for x in vals]  # type: ignore
+                        )
+                ctx.log.alert("Saved %s cuts over %d flows as CSV." % (len(cuts), len(flows)))
+        except (PermissionError, IsADirectoryError, FileNotFoundError) as e:
```
mhils: FileNotFound does not make sense here, but how about we just say `except IOError`?

test/mitmproxy/addons/test_export.py
kajojify: Thanks for the advice :) 
I just followed this comment https://github.com/mitmproxy/mitmproxy/pull/2761#issuecomment-355808629
I'll update this PR.

mitmproxy/addons/cut.py
kajojify: @mhils I think `FileNotFound` makes sense :)
We can get this exception, when inputting `/ome/file.txt` as a filepath. 
I was thinking about `OSError`, but `IOError` looks more reasonable, thanks.

test/mitmproxy/addons/test_cut.py
```diff
@@ -111,6 +111,36 @@ def test_cut_save(tmpdir):
...
+            m.side_effect = [PermissionError("Permission denied"),
+                             IsADirectoryError("Is a directory"),
+                             FileNotFoundError("No such file or directory")]
+            for i in range(3):
+                tctx.command(c.save, "@all", "request.method", f)
+            assert tctx.master.has_log("Permission denied", level="error")
+            assert tctx.master.has_log("Is a directory", level="error")
+            assert tctx.master.has_log("No such file or directory",
+                                       level="error")
+
```
mhils: Doesn't the test below replace the stuff above?

test/mitmproxy/addons/test_export.py
```diff
@@ -94,6 +94,21 @@ def test_export(tmpdir):
         os.unlink(f)
 
 
+@pytest.mark.parametrize("exception, log_message", [
+    (PermissionError, "Permission denied"),
+    (IsADirectoryError, "Is a directory"),
+    (FileNotFoundError, "No such file or directory")
```
mhils: Man, those FileNotFoundErrors when writing new files... 😄

mhils: Ok, nevermind... https://mitmproxy.slack.com/archives/C060SGTJ6/p1516711868000193 🙈 😄

test/mitmproxy/addons/test_cut.py
kajojify: \0/
Forget to get rid of it. Sorry.

mhils:
👍

PR #2811, opened 2018-01-22
don't crash when being scanned with nmap locally, fix #2771

### LINKED ISSUE #2771: Traceback appears, when trying to scan myself with nmap.

1. Run mitmproxy.
2. Scan yourself with **nmap**: `nmap -sT localhost`

Traceback:
image

##### System information

Mitmproxy: 3.0.0.dev0004 (commit 68c32d8) 
Python:    3.5.2
Platform:  Linux-4.4.0-104-generic-x86_64-with-Ubuntu-16.04-xenial

ankitsaini2609:
I have solved issue #2771 kindly take a look :-)

### DIFF
--- a/test/mitmproxy/net/test_tcp.py
+++ b/test/mitmproxy/net/test_tcp.py
@@ -803,3 +803,9 @@ class TestPeekSSL(TestPeek):
         with c.connect() as conn:
             c.convert_to_tls()
             return conn.pop()
+class Test_nmap_localhost_scan(tcp.BaseHandler):
+    connection = socket.socket(socket.AF_INET6, socket.SOCK_STREAM, 0)
+    address = ('::ffff:127.0.0.1', 37656, 0, 0)
+    server = None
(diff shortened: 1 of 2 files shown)


mhils:
Thanks! Can you please

1. Add a test
2. Check your code - `self._makefile()` probably doesn't need to be there twice, and `except` should specify an error message. I'm in fact surprised that the linter didn't catch the latter. @Kriechi?

test/mitmproxy/net/test_tcp.py
```diff
@@ -611,7 +600,6 @@ def test_dhparams(self):
 
 
 class TestTCPClient(tservers.ServerTestBase):
-
```
Kriechi: Please don't delete all those lines.

ankitsaini2609: Actually, I am trying to create a new class so I will not make any changes to existing one.

PR #2819, opened 2018-01-29
Extend mypy checking, fix #2194

### LINKED ISSUE #2194: Extend mypy checking

kira0204:
Fixes #2194

### DIFF
--- a/examples/complex/xss_scanner.py
+++ b/examples/complex/xss_scanner.py
@@ -395,26 +395,23 @@ def get_XSS_data(body: Union[str, bytes], request_URL: str, injection_point: str
 
 # response is mitmproxy's entry point
 def response(flow: http.HTTPFlow) -> None:
-    cookiesDict = get_cookies(flow)
-    try:
-        resp = flow.response.content.decode('utf-8')
-    except AttributeError:
-        resp = flow.response.content
+    cookies_dict = get_cookies(flow)
+    resp = flow.response.get_text(strict=False)
     # Example: http://xss.guru/unclaimedScriptTag.html
     find_unclaimed_URLs(resp, flow.request.url)
-    results = test_end_of_URL_injection(resp, flow.request.url, cookiesDict)
+    results = test_end_of_URL_injection(resp, flow.request.url, cookies_dict)
     log_XSS_data(results[0])
     log_SQLi_data(results[1])
     # Example: https://daviddworken.com/vulnerableReferer.php
-    results = test_referer_injection(resp, flow.request.url, cookiesDict)
+    results = test_referer_injection(resp, flow.request.url, cookies_dict)
     log_XSS_data(results[0])
     log_SQLi_data(results[1])
     # Example: https://daviddworken.com/vulnerableUA.php
-    results = test_user_agent_injection(resp, flow.request.url, cookiesDict)
+    results = test_user_agent_injection(resp, flow.request.url, cookies_dict)
     log_XSS_data(results[0])
     log_SQLi_data(results[1])
     if "?" in flow.request.url:
         # Example: https://daviddworken.com/vulnerable.php?name=
-        results = test_query_injection(resp, flow.request.url, cookiesDict)
+        results = test_query_injection(resp, flow.request.url, cookies_dict)
         log_XSS_data(results[0])
         log_SQLi_data(results[1])
--- a/examples/complex/xss_scanner.py
+++ b/examples/complex/xss_scanner.py
@@ -83,12 +83,13 @@ def get_cookies(flow: http.HTTPFlow) -> Cookies:
     return {name: value for name, value in flow.request.cookies.fields}
 
 
-def find_unclaimed_URLs(body: Union[str, bytes], requestUrl: bytes) -> None:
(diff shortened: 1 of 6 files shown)


examples/complex/xss_scanner.py
```diff
@@ -202,20 +201,21 @@ def get_SQLi_data(new_body: str, original_body: str, request_URL: str, injection
         "Sybase": (r"(?i)Warning.*sybase.*", r"Sybase message", r"Sybase.*Server message.*"),
     }
     for dbms, regexes in DBMS_ERRORS.items():
-        for regex in regexes:
+        for regex in regexes:  # type: ignore
```
mhils: Why is that required?

examples/complex/xss_scanner.py
```diff
@@ -202,20 +201,21 @@ def get_SQLi_data(new_body: str, original_body: str, request_URL: str, injection
...
                 return SQLiData(request_URL,
                                 injection_point,
                                 regex,
                                 dbms)
+    return None
 
 
 # A qc is either ' or "
-def inside_quote(qc: str, substring: bytes, text_index: int, body: bytes) -> bool:
+def inside_quote(qc: str, substring_bytes: bytes, text_index: int, body_bytes: bytes) -> bool:
```
mhils: That's not really required IMO, we already have a type annotation here.

examples/complex/xss_scanner.py
```diff
@@ -380,20 +380,21 @@ def handle_starttag(self, tag, attrs):
...
             return None
+    return None
 
 
 # response is mitmproxy's entry point
 def response(flow: http.HTTPFlow) -> None:
     cookiesDict = get_cookies(flow)
     # Example: http://xss.guru/unclaimedScriptTag.html
-    find_unclaimed_URLs(flow.response.content, flow.request.url)
+    find_unclaimed_URLs(flow.response.content.decode('utf-8'), flow.request.url)
```
mhils: This will fail for any non-utf8 contents, take a look at `response.get_text` :)

examples/complex/xss_scanner.py
kira0204: It's not to explicitly tell about the type but earlier `substring` variable was being used to store `bytes` type and after decoding it was being used to store `str`. Thus for a better approach, I renamed this variable

examples/complex/xss_scanner.py
kira0204: Because mypy was throwing an error. More about this can be read in https://github.com/python/mypy/issues/2598 . I thought this would be the best solution as of now

examples/complex/xss_scanner.py
kira0204: Hi, we need to rewrite most of the code since there are many places where we are taking that the content will be utf8 only. `response` function has been written on this assumption so is most of the code. I thought this assumption was wrong but since this was an already existing code didn't ask.
@mhils I will go forward and check all of the code, it would be great if you could review the whole existing code once. Thank you

examples/complex/xss_scanner.py
```diff
@@ -202,20 +201,21 @@ def get_SQLi_data(new_body: str, original_body: str, request_URL: str, injection
...
 
 # A qc is either ' or "
-def inside_quote(qc: str, substring: bytes, text_index: int, body: bytes) -> bool:
+def inside_quote(qc: str, substring_bytes: bytes, text_index: int, body_bytes: bytes) -> bool:
     """ Whether the Numberth occurence of the first string in the second
         string is inside quotes as defined by the supplied QuoteChar """
-    substring = substring.decode('utf-8')
-    body = body.decode('utf-8')
+    substring = substring_bytes.decode('utf-8')
+    body = body_bytes.decode('utf-8')
```
kira0204: As you can see here, therefore used `substring_bytes`

kira0204:
@mhils Is there anything else that needs to be done? :)

examples/complex/xss_scanner.py
mhils: Makes sense, thanks.

examples/complex/xss_scanner.py
```diff
@@ -380,33 +383,38 @@ def handle_starttag(self, tag, attrs):
...
         else:
             return None
+    return None
 
 
 # response is mitmproxy's entry point
 def response(flow: http.HTTPFlow) -> None:
     cookiesDict = get_cookies(flow)
+    try:
+        resp = flow.response.content.decode('utf-8')
```
mhils: Again, you assume that the response content is utf8, which is not guaranteed to be true. In fact It's latin-1 by default. Use `.get_text(strict=False)` here.

kira0204:
Hi, @mhils Simplifying mypy invocations will need `__init__.py` , without that the build is failing because `mypy` can't find .py files

mhils:
Thanks! 🍰

kira0204:
Thankyou so much for guiding me :)

PR #2825, opened 2018-01-30
Real-time flow counter

kajojify:
Now flow counter in the bottom left corner is updated only when we press `down`-`up` buttons or use mouse wheel. I think it will be more useful, if we make this counter work in real-time.

When new flow is added, it is sent to all receivers through `sig_view_add` - https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/addons/view.py#L305
So, we just redraw our statusbar every time, when new flow is sent through `sig_view_add`
https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/tools/console/statusbar.py#L145


mhils:
👍

Issue #2785, opened 2018-01-12
Editing message attributes is cumbersome on 3.0 / console.choose.cmd needs shortcuts.


mhils:
##### Steps to reproduce the problem:

Editing a flow's URL currently requires the following keystrokes:
`e` -> `down` -> `down` -> `down` -> `down` -> `down` -> `down` -> `down` -> `down` -> `down` -> `down` -> `down` -> `down` -> `enter`.

This is a clear downgrade from 2.0 and before where each option in that dialog could be selected by pressing a specific key (e.g. `u` to edit the url). It'd be very useful to add these kinds of shortcuts to `console.choose.cmd` again. A bit tricky because we can't pass an explicit list as before, but still something we should do.

##### System information

Mitmproxy: 3.0.0.dev1147 (commit b7db304)
Python:    3.5.2
OpenSSL:   OpenSSL 1.1.0g  2 Nov 2017
Platform:  Linux-4.4.0-43-Microsoft-x86_64-with-Ubuntu-16.04-xenial

kajojify:
@mhils How should it look like? 
I see two and a half variants :)
1) Like in mitmproxy 2.0.x. When we press `e` in flowview, Part overlay appears along with message in status bar: 
![image](https://user-images.githubusercontent.com/20267977/35242172-801cffd6-ffc1-11e7-9dd4-c71536864476.png)
Then when I press the certain button, the corresponding editor opens.

2) Each line in Part overlay has its shortcut written on this line.
When I press the certain button:
a) The corresponding line is focused  - I need to press `Enter` to confirm my decision and enter the editor.
b) I directly get into the corresponding editor.
Do you have any thoughts about it?

mhils:
@kajojify: I think both 1) and 2b) are good alternatives. My initial idea was that we have `console.choose.cmd` auto-select keyboard shortcuts so that it maybe looks like this:
![image](https://user-images.githubusercontent.com/1019198/35243179-3cf94160-ffbc-11e7-9594-cbb678216487.png)
(this could get aesthetically nicer if we reorder the entries)

or like this:
![image](https://user-images.githubusercontent.com/1019198/35243292-a70fc3ee-ffbc-11e7-8ad5-52a4da7ab711.png)

Looking at it now, I think the last option is the nicest one, both in terms of implementation and UX. No definite opinion though. 🙂

kajojify:
Ok, I'll try to work on it :)

kajojify:
[...]

PR #2834, opened 2018-02-01
Issue #2785. New shortcuts.

kajojify:
Shortcuts for all overlays. 
Now it is much easier to choose menu item at the bottom of any overlay.

If there is a situation, when your overlay has so many items, that mitmproxy doesn't have so many shortcuts (what if!), you will see the warning message: `Too few available shortcuts.` and anyway you will be able to use remaining shortcuts.
Examples:

If there aren't available shortcuts:
image

P.S. Now you can leave overlay by pressing `q` as well.

### DIFF
--- a/mitmproxy/tools/console/overlay.py
+++ b/mitmproxy/tools/console/overlay.py
@@ -2,7 +2,6 @@ import math
 
 import urwid
 
-from mitmproxy import ctx
 from mitmproxy.tools.console import signals
 from mitmproxy.tools.console import grideditor
 from mitmproxy.tools.console import layoutwidget
@@ -74,7 +73,7 @@ class ChooserListWalker(urwid.ListWalker):
 
     def _get(self, idx, focus):
         c = self.choices[idx]
-        return Choice(c, focus, c == self.current, self.shortcuts[idx:idx+1])
+        return Choice(c, focus, c == self.current, self.shortcuts[idx:idx + 1])
 
     def set_focus(self, index):
         self.index = index
--- a/mitmproxy/tools/console/overlay.py
+++ b/mitmproxy/tools/console/overlay.py
@@ -133,8 +133,8 @@ class Chooser(urwid.WidgetWrap, layoutwidget.LayoutWidget):
             self.callback(self.choices[self.walker.index])
             signals.pop_view_state.send(self)
             return
-        elif key in self.possible_shortcuts:
-            shortcut_index = self.possible_shortcuts.find(key)
+        elif key in self.shortcuts:
+            shortcut_index = self.shortcuts.index(key)
             self.callback(self.choices[shortcut_index])
             signals.pop_view_state.send(self)
             return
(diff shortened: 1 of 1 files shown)


mitmproxy/tools/console/overlay.py
```diff
@@ -120,7 +133,12 @@ def keypress(self, size, key):
             self.callback(self.choices[self.walker.index])
             signals.pop_view_state.send(self)
             return
-        elif key == "esc":
+        elif key in self.shortcuts:
+            shortcut_index = self.shortcuts.index(key)
+            self.callback(self.choices[shortcut_index])
+            signals.pop_view_state.send(self)
+            return
+        elif key in ["q", "esc"]:
```
mhils: If `q` binds to exit, let's not use it as a shortcut. :-)

mitmproxy/tools/console/overlay.py
```diff
@@ -131,6 +149,11 @@ def keypress(self, size, key):
         elif key in keymap.navkeys:
             return super().keypress(size, key)
 
+    def get_shortcuts(self, choices):
+        if len(self.possible_shortcuts) < len(choices):
+            ctx.log.warn("Too few available shortcuts.")
```
mhils: It's good that we think of this edge case, but I feel that it's not worth to emit a warning here. A choice without shortcut is still fully functional.

kajojify:
image

image

image

mitmproxy/tools/console/overlay.py
```diff
@@ -60,14 +65,15 @@ def keypress(self, size, key):
 
 
 class ChooserListWalker(urwid.ListWalker):
-    def __init__(self, choices, current):
+    def __init__(self, choices, current, shortcuts):
         self.index = 0
         self.choices = choices
         self.current = current
+        self.shortcuts = shortcuts
```
mhils: Let's use a class variable as in https://github.com/mitmproxy/mitmproxy/pull/2834#pullrequestreview-93365506 - no need to pass and slice the same string all the time.

mitmproxy/tools/console/overlay.py
```diff
@@ -120,7 +128,12 @@ def keypress(self, size, key):
             self.callback(self.choices[self.walker.index])
             signals.pop_view_state.send(self)
             return
-        elif key == "esc":
+        elif key in self.shortcuts:
+            shortcut_index = self.shortcuts.index(key)
```
mhils: This raises a ValueError for keys which are not in `self.shortcuts` I think?

I'd prefer a `choice_by_shortcut` function as in https://github.com/mitmproxy/mitmproxy/pull/2834#pullrequestreview-93365506, mainly because that can be tested.

kajojify: @mhils it won't raise ValueError, because we check, whether key is in `self.shortcuts` or not -> `elif key in self.shortcuts`. So `key` is always in `self.shortcuts`, when we call `self.shortcuts.index(key)`
Ok, I'll change it on `choice_by_shortcut` :)

kajojify:
@mhils thank you for your review! 👍

PR #2835, opened 2018-02-02
[web] Fix unstable sorting

MatthewShao:
This is more complex then it looks like:
1. we want to have the descending/ascending argument, so we should use [_.orderBy](https://lodash.com/docs/4.17.4#orderBy) here.
2. The `makeSort` function will be invoked in other places like [here](https://github.com/mitmproxy/mitmproxy/blob/master/web/src/js/ducks/flows.js#L108) and [here](https://github.com/mitmproxy/mitmproxy/blob/master/web/src/js/ducks/utils/store.js#L195), so we cannot just eliminate it.

We should have a stable sort now, I tried it with a bunch of flows which have same status code, they remained the same order after the sorting.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -37,7 +37,8 @@
     "redux-logger": "^3.0.6",
     "redux-mock-store": "^1.3.0",
     "redux-thunk": "^2.2.0",
-    "shallowequal": "^1.0.2"
+    "shallowequal": "^1.0.2",
+    "stable": "^0.1.6"
   },
   "devDependencies": {
     "babel-core": "^6.26.0",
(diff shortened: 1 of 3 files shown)


mhils:
Argh, I thought that would be easier. Good job! 

This certainly is a bit uglier than I had hoped. Do you prefer this solution or shall we go for the `stable` package considering all things? Both is fine with me.

MatthewShao:
Then I just added the stable dependency, saving lots of troubles 😌 .
for reference: https://github.com/Two-Screen/stable

mhils:
Much better :sweat_smile:
Let's fix package.json (yarn and correct folder) and this is ready to merge! :)

MatthewShao:
sorry for the mistakes :sweat_smile

PR #2841, opened 2018-02-04
Modify Linux transparent proxy docs to include ipv6 configuration.

aniketpanjwani:


aniketpanjwani:
I found that mitmproxy would not intercept some websites using ipv6 unless I also made the above changes to configuration.

docs/transparent/linux.rst
```diff
@@ -11,13 +11,15 @@ achieve transparent mode.
...
+    >>> sysctl -w net.ipv6.conf.all.forwarding=1
 
     You may also want to consider enabling this permanently in ``/etc/sysctl.conf`` or newly created ``/etc/sysctl.d/mitmproxy.conf``, see `here <https://superuser.com/a/625852>`__.
 
  3. If your target machine is on the same physical network and you configured it to use a custom
     gateway, disable ICMP redirects:
 
     >>> sysctl -w net.ipv4.conf.all.accept_redirects=0
+    >>> sysctl -w net.ipv6.conf.all.accept_redirects=0
     >>> sysctl -w net.ipv4.conf.all.send_redirects=0
```
mhils: Any particular reason why you didn't add a call for `ipv6.conf.all.send_redirects` as well?

I'm not even sure why we have the `accept_redirects` one in here, I asked the original contributor https://github.com/mitmproxy/mitmproxy/pull/2709/files#r165928069 ...) to also maybe shed some light on this.

aniketpanjwani: Honestly, I don't really understand what step 3 is referring to, or exactly why the ipv4 accept/send_redirects is necessary. I don't have a call for ipv6.conf.all.send_redirects, because I don't think there is any such setting: http://lists.cluenet.de/pipermail/ipv6-ops/2010-January/002785.html . I was just trying to be consistent with the previous instructions.

mhils: So the reason why we need to set `net.ipv4.conf.all.send_redirects=0` is that some OSes helpfully inform clients that there is a faster route: going directly to the router (this assumes everyone is on the same LAN). This then cirumvents mitmproxy. It is quite annoying that there is no such flag for IPv6, but I guess this is nothing we can help with for now.
I don't know why we set `accept_redirects`, that shouldn't matter for our case as far as I understand it.

Kriechi: Would IPv6 work in such a scenario? Don't we have to send a **router advertisment** or some other NDP magic?

mhils: @Kriechi: This assumes you set a custom gateway on your client.

mhils: Ok, let's move forward here and just remove the accept_redirects settings. I'll take care of this. :-)

aniketpanjwani: Cool, sounds good!

Issue #2849, opened 2018-02-07
import ctype error when used with --scripts SCRIPT option from mitmdump


YuriyPryyma:
##### Steps to reproduce the problem:

1. install binary release
2. run mitmdump -s test.py (with import ctypes)

#### Output:

```
  Addon error: Traceback (most recent call last):
  File "site-packages/PyInstaller/loader/pyiboot01_bootstrap.py", line 149, in __init__
  File "ctypes/__init__.py", line 347, in __init__
OSError: command_path: cannot open shared object file: No such file or directory

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "mitmproxy/addonmanager.py", line 67, in safecall
  File "mitmproxy/addons/script.py", line 76, in tick
  File "mitmproxy/addonmanager.py", line 177, in register
  File "mitmproxy/command.py", line 134, in collect_commands
  File "ctypes/__init__.py", line 417, in __getattr__
  File "site-packages/PyInstaller/loader/pyiboot01_bootstrap.py", line 151, in __init__
PyInstallerImportError: Failed to load dynlib/dll 'command_path'. Most probably this dynlib/dll was not found when the application was frozen.
```

##### System information
UBUNTU 16.04

Mitmproxy: 3.0.0.dev1136 (commit 15f525e) binary
Python:    3.5.2
OpenSSL:   OpenSSL 1.1.0g  2 Nov 2017
Platform:  Linux-4.13.0-32-generic-x86_64-with-debian-stretch-sid

mhils:
Hi,

If you want to use third-party modules in your mitmproxy scripts, you need to install mitmproxy from source (http://docs.mitmproxy.org/en/stable/install.html#advanced-installation). Our precompiled binaries contain a frozen Python environment which does not ship with `ctypes`:

Houd1ny:
Man I will have the same error even if I run it from sources

Houd1ny:
see https://github.com/mitmproxy/mitmproxy/pull/2850

mhils:
You definitely shouldn't get a `PyInstallerImportError` when running from source, but yes, we should probably catch this better.

Houd1ny:
This is possible that user redefined hasattr method as ctypes libraries does
And it could throw when started with argument "command_path"

mhils:
Yes - we definitely should catch this better and I very much welcome #2850, the root cause of @YuriyPryyma's problem is the use of the binaries though.

[...]

PR #2850, opened 2018-02-07
Add exption handling for hasattr because it could throw

Houd1ny:
Fixed https://github.com/mitmproxy/mitmproxy/issues/2849


Houd1ny:
I will update it and add test

mhils:
Hi, any update on this? :)

mhils:
Fixed in #2956, thanks again for the initial PR!

PR #2857, opened 2018-02-09
Add octet-stream json viewer

LokiSharp:
A content viewer for json with Content-Type: application/octet-stream compress by zlib.


Kriechi:
Thanks for the PR! 👍 

I'm not familiar with json streaming, but as far as I understand it there are multiple different ways that are accepted and used (https://en.wikipedia.org/wiki/JSON_streaming#Applications_and_tools). Which one are you implementing? I can see you are looking for an `x` character in the beginning of the data before decompressing - would you mind explaining & documenting this in a doc-string?

LokiSharp:
Hi, It not for JSON streaming. It for decompress some json compress by zlib and content-type in header is application/octet-stream.
For eliminate ambiguity I rename stream_json.py to octet_stream_json.py

That x character in the beginning of the data is zlib's Compression Method and flag.(https://tools.ietf.org/html/rfc1950)
I modify it to check hex with '0x78' and ['0x01', '0x5e', '0x9c', '0xda']

mhils:
Thanks. Is this some standard or just a custom format you encountered?

LokiSharp:
I'm not sure if this is a standard.
In order to prevent Content being parsed, someone set the Content-Type to application/octet-stream.
The situation I encountered was that it compressed JSON with zilb and set the Content-Type to application/octet-stream.

mhils:
Makes sense - thanks. So I'm very happy to see this PR as this is exactly how we want this API to be used, but I think there is very little value in merging this into mitmproxy as it's a rather unique thing you are experiencing. Of course, you should definitely keep this as an addon for your use! What do you think?

LokiSharp:
OK, I will try to learn how to create an addon and close this PR

PR #2858, opened 2018-02-10
Fix reverting of a flow

kajojify:
When pressing `V`, flow wasn't reverted to the previous state after editing some part of it and after encoding/decoding of a flow body.
It is fixed.

### DIFF
--- a/mitmproxy/addons/core.py
+++ b/mitmproxy/addons/core.py
@@ -178,6 +178,7 @@ class Core:
         for f in flows:
             p = getattr(f, part, None)
             if p:
+                f.backup()
                 current_enc = p.headers.get("content-encoding", "identity")
                 if current_enc == "identity":
                     p.encode("deflate")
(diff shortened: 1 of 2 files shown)


mitmproxy/addons/core.py
```diff
@@ -178,6 +178,7 @@ def encode_toggle(self, flows: typing.Sequence[flow.Flow], part: str) -> None:
         for f in flows:
             p = getattr(f, part, None)
             if p:
+                f.backup()
```
mhils: Isn't this something we should do in `flow.encode` and `flow.decode`?

mitmproxy/tools/console/consoleaddons.py
```diff
@@ -383,6 +383,7 @@ def edit_focus(self, part: str) -> None:
             part in ("response-headers", "response-body", "set-cookies") and
             flow.response is None
         )
+        flow.backup()
```
mhils: Nit: Don't move it between `require_dummy_response` definition and usage, this can live above or below :)

PR #2847, opened 2018-02-06
 Handling empty key with value for cookies, fix #2827 

kira0204:
Fixes #2827 ,
The earlier solution I implemented was bad. I think this will do a decent job.  

Thankyou so much @kajojify for helping out on the flow of the code :)

Issue #2827, opened 2018-01-30
Mitmproxy crashes, when setting only value in cookies/set-cookies grideditor


kajojify:
##### Steps to reproduce the problem:

1. Run mitmproxy.
2. Press `n` -> `enter` -> `enter` -> `e` -> `cookies` (or `set-cookies`)
3. You are in grideditor. Press `a`. Choose `Value` column and add any value.
You will see something like:
![image](https://user-images.githubusercontent.com/20267977/35590192-be00b434-060e-11e8-8ebd-e5bb20952aef.png)

4. Press `q` -> `e` -> `cookies` (or `set-cookies` if you're doing that for it)

Traceback: 
```
Traceback (most recent call last):
  File "/home/kajoj/Mitmproxy/mitmproxy/mitmproxy/tools/console/master.py", line 216, in run
    self.loop.run()
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 278, in run
    self._run()
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 376, in _run
    self.event_loop.run()
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 682, in run
    self._loop()
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 710, in _loop
    self._entering_idle()
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 671, in _entering_idle
    callback()
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 564, in entering_idle
    self.draw_screen()
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/main_loop.py", line 578, in draw_screen
    canvas = self._topmost_widget.render(self.screen_size, focus=True)
  File "/home/kajoj/Mitmproxy/mitmproxy/venv/lib/python3.5/site-packages/urwid/widget.py", line 141, in cached_render
    canv = fn(self, size, focus=focus)
[...]

PR #2859, opened 2018-02-11
Allow more permissive cookie parsing, Fix #2829

### LINKED ISSUE #2829: Incorrect displaying of set-cookie attributes, if no name-value pairs were set in set-cookie header editor

1. Run mitmproxy.
2. Press `n` -> `Enter` -> `Enter` -> `e` -> `set-cookies`
3. Press `a` to add a new row.
4. Choose `Attributes` column and enter subeditor. 
5. Add any name-value pair, for example `a    1`.
6. Go to flowview tab by pressing `q` twice. You will see this in Response tab:
image
I am not sure it is correct to display just `=` without anything 

7. Press `e` -> `set-cookies`. You will see, that name-value from `Attributes` column moved to high-level name-value:
image

`set-cookies` was changed in Response tab as well. 
Strange behaviour.
 
This small bug is relevant for 2.0.x version as well.

##### System information

Mitmproxy: 3.0.0.dev75 (commit 02d2ef7) 
Python:    3.5.2

tran-tien-dat:
Currently, having empty cookie key and value will lead to weird behaviour, as mentioned in #2829. Additionally, having cookie value without a key will crash the program when trying to edit the cookie header.

Initially, I thought we should reject malformed `Cookie` or `Set-Cookie` header and allow the user to edit the raw header instead. However, `cookies.py` says we are trying to be as permissive as possible, accepting even malformed header. Hence, I think `=;a=b`, `=bc` should all be accepted, as they can be interpreted as having empty cookie key (and value). The only thing that should return an empty array is something like `;` or the empty string, which will be parsed as `(lhs, rhs) == ('', None)`. That is the change I have made in this PR.

This is my first PR, so please let me know if I miss out on anything. Thank you!

### DIFF
--- a/mitmproxy/net/http/cookies.py
+++ b/mitmproxy/net/http/cookies.py
@@ -114,11 +114,11 @@ def _read_cookie_pairs(s, off=0):
         lhs, off = _read_key(s, off)
         lhs = lhs.lstrip()
 
-        if lhs:
-            rhs = None
-            if off < len(s) and s[off] == "=":
-                rhs, off = _read_value(s, off + 1, ";")
+        rhs = None
+        if off < len(s) and s[off] == "=":
+            rhs, off = _read_value(s, off + 1, ";")
 
+        if not (lhs, rhs) == ('', None):
             pairs.append([lhs, rhs])
 
         off += 1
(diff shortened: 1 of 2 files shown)


mitmproxy/net/http/cookies.py
```diff
@@ -114,11 +114,11 @@ def _read_cookie_pairs(s, off=0):
         lhs, off = _read_key(s, off)
         lhs = lhs.lstrip()
 
-        if lhs:
-            rhs = None
-            if off < len(s) and s[off] == "=":
-                rhs, off = _read_value(s, off + 1, ";")
+        rhs = None
+        if off < len(s) and s[off] == "=":
```
kira0204: As @kajojify mentioned in #2847 , initializing `rhs` as `None` will throw an error when the cookie is `"one"` which as you can see from the test cases will parsed as `["one", None]`

tran-tien-dat: Thanks for pointing that out. I think we should initialize it as an empty string then.

kira0204: Hi, I was already working on it, I am so sorry about this.

tran-tien-dat:
When trying to use the empty string instead of `None` to represent the value in a key-value pair is empty, I realize the intended difference between them. `None` is used to represent something like `foo`, while the empty string represents something like `foo=`. Nevertheless, this distinction is lost when the grid editor reconstructs the cookie string. Hence I suggest that we still change to use the empty string, but also when reconstructing, we drop the `=` if the right hand side is empty.

tran-tien-dat:
Anyway, I realize that this PR and #2847 are basically fixing the same thing. So I will abandon this PR.

kira0204:
@tran-tien-dat Hi, In my PR I am only fixing #2827 ans haven't thought about the issue you are dealing with. I know it may sound weird but do you want to wait for my PR to get merged and then work on this issue so that you can make the necessary changes to solve the issue you are dealing with ? :) I don't know how @mhils will react to this 🙈

mitmproxy/net/http/cookies.py
kira0204: If you initialise `rhs` as an empty string, then you need to find a way to not append in the case of an empty cookie , but if there is a cookie like this `=` you need to parse it in a way that it looks like `[["", ""]]` . I am not working on the issue you are trying to solve, sorry for the confusion

tran-tien-dat:
@kira0204 I think you are right. Even though your PR does not directly address the issue I am trying to fix, it will affect how the issue can be fixed, since we are changing the same part of the code. So I think to avoid conflict, I will wait and see how your PR turns out. In the mean time, I will find other issues to contribute to.

mhils:
I'm closing this for now, please feel free to open a new PR once merge conflicts have been resolved!

PR #2861, opened 2018-02-11
Add a JSON dump script: write data into a file or send to an endpoint as JSON

emedvedev:
For my use case, I needed to analyze the traffic patterns and store them in Elasticsearch, so here's a script that will serialize all captured traffic as JSON while mitmproxy is running (or repeating) and write it to a file or send to a remote endpoint. 

It also performs some processing to make sure the data is fully compatible with ES — and readable by humans, too (if reading raw traffic dumps is what you're into).

It does slow the traffic flow down, but not very noticeably because of `@concurrent` (`threading.Lock()` is implemented if the dump destination is a file).

Let me know if there's something you want updated or changed!


emedvedev:
Flake8 wasn't happy, now it is. :)

examples/complex/jsondump.py
```diff
@@ -0,0 +1,245 @@
...
+                ctx.log.info('HTTP Basic auth enabled.')
+        else:
+            self.outfile = open(ctx.options.dump_destination, 'a')
+            self.url = None
+            self.lock = Lock()
+            ctx.log.info('Writing all data frames to %s' % ctx.options.dump_destination)
+
+        self._init_transformations()
+
+    @concurrent
```
mhils: Have you considered to just put messages into a single `queue.Queue` and have a dedicated thread for writing to disk? This way you don't need to spawn multiple threads per HTTP request.

examples/complex/jsondump.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    def __init__(self):
+        self.outfile = None
+        self.transformations = None
+        self.encode = None
+        self.url = None
+        self.lock = None
+        self.auth = None
+
+    def __del__(self):
```
mhils: You can use `done()`, which is triggered on mitmproxy shutdown. :)

emedvedev:
> Have you considered to use mitmweb's `flow_to_json` instead of a custom implementation?
I've looked at it, but there's too many changes I'd have to make to make it ES-compatible. I do like the idea of a flag to drop content though, a lot of people probably wouldn't need to store the contents in their dumps for analysis.

> Last but not least, it would be useful to have a small test in https://github.com/mitmproxy/mitmproxy/tree/master/test/examples so that we keep this up-to-date. 😉
Will do!

mhils:
Any update on this? :-)

Thanks!
Max

emedvedev:
Hey,

Really sorry, got caught up in another semi-urgent project! Will get back to this PR in the next couple days hopefully. :)

Gigiarum:
In this [Discourse post](https://discourse.mitmproxy.org/t/har-mongo-dump-script/901) there is my script to save requests and responses to MongoDB.

emedvedev:
@mhils sorry for the delay! Implemented proper queueing and wrote a few tests for both file and HTTP dumps.

I've had to add [requests-mock](http://requests-mock.readthedocs.io/) to dev dependencies to test HTTP dumps properly, hope you don't mind. :)

emedvedev:
Apparently `requests` were moved to dep dependencies in https://github.com/mitmproxy/mitmproxy/issues/2899 while I was busy being slow. :) It's required for this script, but if there's a general consensus to exclude script dependencies, then I don't mind.

examples/complex/jsondump.py
```diff
@@ -0,0 +1,257 @@
+"""
+This script serializes the entire traffic dump, including websocket traffic,
+as JSON, and either sends it to a URL or writes to a file. The serialization
+format is optimized for Elasticsearch; the script can be used to send all
+captured traffic to Elasticsearch directly.
+
+Usage:
+
+    mitmproxy
+        -R http://example.com/
```
Kriechi: To use mitmproxy in reverse mode please use `--mode reverse:SPEC` instead:
https://docs.mitmproxy.org/stable/concepts-modes/#reverse-proxy

Kriechi:
Maybe we should keep the examples a bit simpler with regards to features.
Right now the new example provides two unrelated features: dump to a local json file on disk, and send urls to a server.

IMHO we should split these two to improve readability and complexity. Thoughts?

examples/complex/jsondump.py
emedvedev: Fixed, thanks!

emedvedev:
@Kriechi fair point, but I just couldn't find a good way to split. The main feature this script really provides is data processing (mitmproxy -> JSON); output method (dump/send) is secondary. There'd be too much code reuse between the two, and I didn't want to introduce a `lib/` directory inside `examples/` or modify the code structure in any other way.

fenilgandhi:
I completely liked the idea of having a (mitmproxy--> JSON) feature in mitmproxy. I couldn't help but notice that this PR a lot of concurrent code in the master. and since we are progressing towards `asyncio` so maybe we can update this to `asyncio` and then merge it.

I am completely new to asyncio but i would like to add some commits for it. Shall i add them to this PR or    open a new PR once this one is merged ?

Kriechi:
@emedvedev I would argue that sending the URL (and payload) to a server is a trivial task, compared to the json dumper.
Examples don't need to be fully working end-to-end use cases, most users will adapt and modify them to their own needs anyway.
The main contribution of this PR is the flow-to-json extraction of data - as such I would remove the server posting of data or extract it into a rather minimalistic generic example.

emedvedev:
@Kriechi I was actually aiming for a fully working end-to-end case, so now I see your point. :)

Problem is, the server posting of data depends on the flow-to-json extraction, so if I extract it into a minimalistic example, it would still need to import the extractor code, and the queueing code would also have to be reused.

A specific use case that this PR serves is sending the data to a search engine (like Elasticsearch) or a JSON-based DB (like Mongo). I'd say it's a use case that arises quite often if you need to analyze traffic patterns rather than intercept/modify. That's why I would really prefer to keep the script in one piece: splitting would make anyone with the same use case repeat the work I've already done.

That said, you're the maintainer, and I do see your point, so I'll remove/extract the server posting if that's the final decision.

suhanimirza:
I love PayPal Inc.

drzax:
At the risk of being that guy who seeks support in a pull request thread ...

I've used the code in this PR to dump a bunch of JSON to file without the `dump_encodecontent` flag. I now want to re-construct the content of those requests so I can inspect it, but I'm hitting dead ends.

Any hints?

kajojify:
Hi there! :) 
More than 6 months without any activity. What is the status of PR?

mhils:
Eh, tests failing. @kajojify, this seems to be easy to fix, can you give it a stab?

mhils:
This was outstanding for way too long, thanks again for the patience @emedvedev! 🍰

emedvedev:
Yayyy! Glad I could help, and thanks a lot for merging :)

PR #2870, opened 2018-02-16
Linux: support IPv6 destinations in transparent mode

### LINKED ISSUE #2869: Inability to connect to ipv6 websites through mitmproxy from ipv6 network.

1. Follow [instructions](https://github.com/mitmproxy/mitmproxy/blob/master/docs/transparent/linux.rst) to setup transparent proxy on Linux on my own machine, substituting PREROUTING in the iptables ruleset with OUTPUT.
2. Run `mitmdump --mode transparent`
3. Navigate to ipv4 site (e.g. espn.com) and ipv6 site (e.g. facebook.com).

I've done this from two networks. On one network, I have an ipv6 address (according to http://test-ipv6.com/). On the other network, I do not have an ipv6 address.

+ ipv4 network: Able to connect to both espn.com and facebook.com
+ ipv6 network: Only able to connect to espn.com. Additionally, when I use http://test-ipv6.com after doing the first two steps above, the website can no longer find my ipv6 address.

On the ipv6 network, I receive a message saying the following:

```
Client connection was killed because allow_remote option is set to false, client IP was not a private IP and proxyauth was not set.
To allow remote connections set allow_remote option to true or set proxyauth option.
[2601:246:100:cefc:7da3:3d6c:6195:5f1f]:42868: Connection killed
[2601:246:100:cefc:7da3:3d6c:6195:5f1f]:42868: clientdisconnect
```

When I set the allow_remote option to be true (`mitmproxy --mode transparent --set allow_remote=true`), I now receive the following message:

```
[2601:246:100:cefc:7da3:3d6c:6195:5f1f]:57032: Transparent mode failure: FileNotFoundError(2, ‘No such file or directory’)
```

I should try putting mitmproxy in a VM or Docker container. Then, I think I would instead have to use the PREROUTING chain. If the transparent proxy works from the VM/Docker container, this would indicate that the problem has to do with configuration when using the OUTPUT chain.

I think it's reasonable that mitmproxy should support ipv6 when run on your own computer. If we can figure out what the problem is, I'd be happy to submit a pull request to the Linux transparent proxy documentation with more explicit instructions on how to set up mitmproxy. Apologies if this is a better fit in discourse, but I think this issue at the very least reflects something which could be improved in the transparent proxy documentation.

One hacky option would be to shut down ipv6 altogether through configuration, but I haven't figured out how to make it work. Another possibility is that this has to do with [router advertisements](http://strugglers.net/~andy/blog/2011/09/04/linux-ipv6-router-advertisements-and-forwarding/), but simply setting the accept_ra flag to be equal to two did not work for me. 

##### System information

Mitmproxy: 3.0.0.dev1179 (commit 6d8731e) 
Python:    3.6.4
Platform:  Linux-4.4.0-21-generic-x86_64-with-debian-stretch-sid

mhils:
This fixes #2869 with the help of @aniketpanjwani.

@aniketpanjwani: If you find some time could you check if this code works for you? I improved the IPv4 detection heuristic a bit. Pretty sure it should work, but I'm still mostly coding in the dark here as we can't write sensible tests for this.

### DIFF
--- a/mitmproxy/platform/linux.py
+++ b/mitmproxy/platform/linux.py
@@ -1,12 +1,34 @@
 import socket
 import struct
+import typing
 
-# Python socket module does not have this constant
+# Python's socket module does not have these constants
 SO_ORIGINAL_DST = 80
+SOL_IPV6 = 41
 
 
-def original_addr(csock: socket.socket):
-    odestdata = csock.getsockopt(socket.SOL_IP, SO_ORIGINAL_DST, 16)
-    _, port, a1, a2, a3, a4 = struct.unpack("!HHBBBBxxxxxxxx", odestdata)
-    address = "%d.%d.%d.%d" % (a1, a2, a3, a4)
-    return address, port
+def original_addr(csock: socket.socket) -> typing.Tuple[str, int]:
+    # Get the original destination on Linux.
+    # In theory, this can be done using the following syscalls:
+    #     sock.getsockopt(socket.SOL_IP, SO_ORIGINAL_DST, 16)
+    #     sock.getsockopt(SOL_IPV6, SO_ORIGINAL_DST, 28)
+    #
+    # In practice, it is a bit more complex:
+    #  1. We cannot rely on sock.family to decide which syscall to use because of IPv4-mapped
+    #     IPv6 addresses. If sock.family is AF_INET6 while sock.getsockname() is ::ffff:127.0.0.1,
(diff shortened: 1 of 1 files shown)


aniketpanjwani:
Just checked it - I can confirm that it works. Wouldn't break despite my strongest tab-opening efforts.

mhils:
Awesome, thanks!

PR #2873, opened 2018-02-18
Fix a bug with two-words contentview modes

kajojify:
There was a bug:
1. Create a flow and  enter its details.
2. Press `m` and choose `multipart form` or `protocol buffer`.
You will see `command argument mismatch: too many positional arguments`

We call `console.flowview.mode.set` here https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/tools/console/consoleaddons.py#L264-L267
When `repl` is `multipart form` or `protocol buffer`, it is splitted into two parts inside. Thus `console.flowview.mode.set` gets more args that it can accept.

### DIFF
--- a/mitmproxy/tools/console/consoleaddons.py
+++ b/mitmproxy/tools/console/consoleaddons.py
@@ -498,10 +498,11 @@ class ConsoleAddon:
 
     @command.command("console.flowview.mode.set")
     @command.argument("mode", type=mitmproxy.types.Choice("console.flowview.mode.options"))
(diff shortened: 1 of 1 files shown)


mhils:
Isn't that something that should be fixed in `console.choose.cmd`? 😃

kajojify:
I hope such quotes are fine :)

mhils:
This definitely is the right place, `shlex.quote` is your friend! 😉

PR #2880, opened 2018-02-19
Updating Readme for better testing

kira0204:
I believe this is something that might be useful for contributors if they miss lines in their testing and want to know which lines they missed :)


mhils:
Thanks, this generally makes sense. Any particular reason why you are not using the pytest command mentioned below this?

kira0204:
Not a particular reason as such but testing gives me an excuse to take a break for like 3-4 minutes. Earlier I used to use `tox` and sometimes I used to miss statements in the test case I wrote and had to change the directory and use the `pytest` command. I wanted all this to be a single process, so I had to explore a little to find how to do that with tox. So that's the reason I thought this might be beneficial for someone out there. :see_no_evil:

kira0204:
Any thoughts?

cortesi:
Looks good to me. I use both tox and pytest, depending on the circumstance. For full tests I prefer tox. I'd tweak the grammar ever so slightly. Something like this:

> To run complete tests with a full coverage report, you can use the following command

kira0204:
@cortesi Thanks, I have changed the grammar :)

PR #2888, opened 2018-02-21
Fix verbosity mode. Fix #2885

### LINKED ISSUE #2885: Verbose mode not working in v3.0.0 RC2

1. Start mitmdump with -v or --verbose flag
2. No DEBUG level logs prints on standard output

In old stable version (2.0.2) the same steps produce desired output.

##### System information

Mitmproxy: 3.0.0.dev1136 (commit 15f525e)
Python:    3.6.3
Platform:  Linux-3.16.0-5-amd64-x86_64-with-debian-8.9

kajojify:
I delete `args.options or args.commands`, because it seemes to haven't any effect.
I suppose, the idea was to set other verbosity level and show it in `--options` output. It won't work, because only default options values are collected for output https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/optmanager.py#L404

### DIFF
--- a/mitmproxy/tools/main.py
+++ b/mitmproxy/tools/main.py
@@ -50,7 +50,7 @@ def process_options(parser, opts, args):
         args.flow_detail = 0
     if args.verbose:
         args.verbosity = 'debug'
-        args.flow_detail = 3
+        args.flow_detail = 2
 
     adict = {}
     for n in dir(args):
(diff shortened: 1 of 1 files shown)


mhils:
Thanks! Correct call on the removal, this lacked documentation and wasn't clear.

PR #2898, opened 2018-02-23
Explicitly declare pyperclip requirement

yan12125:
pyperclip version 1.6.0 introduces an API change: the class `pyperclip.exceptions.PyperclipException` is moved to `pyperclip.PyperclipException`. As mitmproxy uses the latter, it's better to explicitly declare the requirement.

https://github.com/mitmproxy/mitmproxy/blob/7733252/mitmproxy/addons/cut.py#L144


mhils:
Oh, great catch. Thanks! 🍰

PR #2912, opened 2018-02-24
Spelling corrected for think*

smonami:
Replaced "You should thin* of the mitmproxy" to "You should think* of the mitmproxy"


mhils:
Good catch. Thanks for the PR! 🍰 😃

PR #2914, opened 2018-02-24
Updating docs

### LINKED ISSUE #2913: New proxy mode documentation is outdated.
https://mitmproxy.org/docs/latest/concepts-modes/ still lists at least `-R`, which should be `--mode reverse:...`. This should be trivial to fix, one just needs to go through the whole document and check all proxy modes.

kira0204:
This fixes #2913 :)
The pictures also needs to be changed, is there any source of these images through which it can directly change? Otherwise I will go ahead and photoshop them.


cortesi:
@kira0204 The sources for those images are in the Vision file in the /docs/raw directiory (I presume  - I don't have Vision handy). If you have Visio, an update would be appreciated. If not, we'll have to bug @mhils.

mhils:
It's  Microsoft Visio, but other than that I agree entirely. :-)

cortesi:
That was actually autocorrect, Microsoft Max. ;)

PR #2920, opened 2018-02-25
Docs updates

kira0204:
HI, in addition to minor changes for #2913.
1. I found some more docs with deprecated option for proxy modes, changed them
2. Modified the `images` (reduced the size a little for better UI of docs) as well as the `Visio` raw file.

Please take a look :)


mhils:
This looks fantastic except for proxy-modes-transparent-1.png. Thanks! 🍰

kira0204:
I missed it :see_no_evil: , replaced the original file with the original file itself

mhils:
No problem, that's why we have peer review. Thanks again! 😃

PR #2924, opened 2018-02-25
Split dumper_filter off from view_filter

cortesi:
These don't even do the same thing.


cortesi:
Yes. They don't do the same thing, they live in different addons, and there's no reason for them to have the same name. I'm busy paving the way for a world where options are prefixed by their addon names, exactly like commands. `view_filter` becomes `view.filter` because it's an option in the `view` addon. Common options will live in the `core` namespace, and should be used as sparingly as possible. 

Options are going to be in a state of transition for a bit while I tease all these things out.

Issue #2839, opened 2018-02-03
Different traceback messages appear, when try to run inappropriate file as a script


kajojify:
##### Steps to reproduce the problem:

1. Run mitmproxy.
2. Press `|`(`shift+\`). Input the path to not script file: 
`: script.run @focus /home/kajoj/myphoto.JPG`

Traceback: 
![image](https://user-images.githubusercontent.com/20267977/35767610-4eb01d1c-08f8-11e8-97dc-aca5225fd9a5.png)

2a. `: script.run @focus /home/kajoj/passlist.txt`

Traceback: 
![image](https://user-images.githubusercontent.com/20267977/35767631-c6d8dafe-08f8-11e8-9413-177fe69f4f51.png)

##### Any other comments? What have you tried so far?

I think these aren't only traceback messages, which may appear. 
We should take a look at this place: https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/addons/script.py#L30

##### System information

Mitmproxy: 3.0.0.dev79 (commit 957a630)
Python: 3.5.2
OpenSSL: OpenSSL 1.1.0g 2 Nov 2017
Platform: Linux-4.4.0-112-generic-x86_64-with-Ubuntu-16.04-xenial

kira0204:
Hi, If noone else is working on it, I will take this :)

kira0204:
I have a question
Are we stringent that the script the user is trying to run should be a python file such as `example.py`, 
or we are giving them the flexibility to run any file until and unless it's contents are python code(which we are doing right now)? 

I would like to go with the second one

mhils:
@kira0204: Are you referring to the filename extension (e.g. `.py`)? We shouldn't enforce anything in that regard.

kira0204:
Yes I was, thanks for the clarification :)

kira0204:
Okay since we don't want the user to see all the traceback messages we can use `cut_traceback` with the parameter `_call_with_frames_removed` , I know this is not a neat solution but I am not able to find any solution apart from this :(

kajojify:
@kira0204 we can try to handle these exceptions here -> https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/addons/script.py#L26

kira0204:
I know, but I want the user to know what is the issue with the code that he is trying to load and where. If we ignore this and go forward with the handling of the exceptions and display a generic message then the solution becomes neat but we would be disturbing the UX

[...]

PR #2927, opened 2018-02-25
Exception when running inappropriate file as a script, fix 2839

kira0204:
@kajojify and I had a discussion on this and this was a very difficult choice for us to make.
With the solution that this PR brings forward, although it's a neat solution but it might disturb the user experience when the error from his/her script is not a `SyntaxError` because it won't be informative enough

Another solution is not a neat solution which is proposed in #2839 comments.

### DIFF
--- a/mitmproxy/addons/script.py
+++ b/mitmproxy/addons/script.py
@@ -30,11 +30,11 @@ def load_script(path: str) -> types.ModuleType:
         loader.exec_module(m)
         if not getattr(m, "name", None):
             m.name = path  # type: ignore
-        return m
     except Exception as e:
         ctx.log.error(str(e))
     finally:
         sys.path[:] = oldpath
+        return m
 
 
 class Script:
--- a/mitmproxy/addons/script.py
+++ b/mitmproxy/addons/script.py
@@ -14,7 +14,7 @@ from mitmproxy import eventsequence
 from mitmproxy import ctx
 
 
-def load_script(path: str) -> types.ModuleType:
+def load_script(path: str) -> typing.Optional[types.ModuleType]:
     fullname = "__mitmproxy_script__.{}".format(
         os.path.splitext(os.path.basename(path))[0]
     )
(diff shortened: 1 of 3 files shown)


mitmproxy/addons/script.py
```diff
@@ -30,9 +30,11 @@ def load_script(path: str) -> types.ModuleType:
         loader.exec_module(m)
         if not getattr(m, "name", None):
             m.name = path  # type: ignore
-        return m
+    except Exception as e:
+        ctx.log.error(str(e))
```
cortesi: A few comments here. First, the structure of this except bothers me - we will see an unbound local error for the m variable if there's an error somewhere before m is assigned. I think the right approach is to change the signature of the function to possibly return None, and then to make sure that m has a default value of None before entering the except handler. Second, we know the path here, so we should add that to the error message being logged, so the user knows which file had the error.



I think a better approach is to add another try/except just around the exec_module

test/mitmproxy/data/addonscripts/recorder/error.py
```diff
@@ -0,0 +1,3 @@
+impotr recorder
```
cortesi: Let's make it more obvious that this is intended to be an invalid file with a syntax error, just so I don't accidentally fix it someday. :)

kira0204: Sure thankyou :) I will write this in the comments that this is an intended errors :)

mitmproxy/addons/script.py
kira0204: Thanks for the extensive review :) When I am printing the error as `e` , it also prints the filename and the lineno with it like `invalid syntax(example.py, line 4)`, then this is only for SyntaxError and I couldn't find a way to print the filename for other errors like `ValueError` mentioned in the original error. Even the traceback module didn't help :(

kira0204:
Closed because #2962

PR #2928, opened 2018-02-26
Better error when no type-annotations, fix #2787

### LINKED ISSUE #2787: Can't run mitmproxy after adding a command without type annotations

1. Open `mitmproxy/tools/console/consoleaddons.py`
2. Add arbitrary command without type annotations. For example:
```
    @command.command("console.mycommand")
    def mycommand(self, flow):
        pass
```
3. Try to run mitmproxy. 

Nothing happens. Absolutely. No exceptions, no UI. 

If you add a command with type annotations
```
    @command.command("console.mycommand")
    def mycommand(self, flow: flow.Flow) -> None:
        pass
```
everything will work ok.

##### System information

Mitmproxy: 3.0.0.dev0004 (commit 68c32d8) 
Python:    3.5.2
Platform:  Linux-4.4.0-104-generic-x86_64-with-Ubuntu-16.04-xenial

kira0204:
Hi,
I have replaced the `NotImplementedError` with the `SyntaxError` and this fixes #2787 . Added a test for the same.
Also there is traceback because I have made an assumption that the developers will mostly have this issue and a `traceback` would be useful. Would love to hear your opinions? :)

### DIFF
--- a/mitmproxy/command.py
+++ b/mitmproxy/command.py
@@ -1,5 +1,5 @@
 """
-    This module manges and invokes typed commands.
+    This module manages and invokes typed commands.
 """
 import inspect
 import types
@@ -37,7 +37,7 @@ def typename(t: type) -> str:
     """
     to = mitmproxy.types.CommandTypes.get(t, None)
     if not to:
-        raise NotImplementedError(t)
+        raise SyntaxError("missing type annotations in consoleaddons.py")
     return to.display
 
 
--- a/test/mitmproxy/test_command.py
+++ b/test/mitmproxy/test_command.py
@@ -35,6 +35,10 @@ class TAddon:
     def empty(self) -> None:
         pass
 
+    @command.command("annotation_error")
+    def annotation_error(self):
+        return "ok"
+
     @command.command("varargs")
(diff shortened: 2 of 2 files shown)


mitmproxy/command.py
```diff
@@ -37,7 +37,7 @@ def typename(t: type) -> str:
     """
     to = mitmproxy.types.CommandTypes.get(t, None)
     if not to:
-        raise NotImplementedError(t)
+        raise SyntaxError("missing type annotations in consoleaddons.py")
```
cortesi: Two things here. That consoleaddons.py in this error string is obviously a leftover from debugging or something. :) This should be the command path, and you may have to modify the intermediary functions to make sure you have access to it. Secondly, `SyntaxError` is not the right exception here - it's a builtin that has a very specific meaning. Have a look at `mitmproxy.exceptions` for an appropriate exception here - perhaps `mitmproxy.exceptions.TypeError`?

I haven't validated the approach here deeply yet. Once these things are fixed, I'll take a closer look.

mhils:
What's the status here?

kira0204:
Hi, sorry for the slow response, let's close it since we will be approaching this problem differently keeping user scripts error also in mind using #2971

PR #2941, opened 2018-03-02
Fix for wrapping up extra-long urls in Flowdetail view

### LINKED ISSUE #2818: It is difficult to get information from request/response/detail tab, if url is very-very long

1. Run mitmproxy.
2. Make a request with pathoc: `pathoc -c example.com:80 localhost:8080 'get:/SOMETHING VERY LONG'`
3. Enter the corresponding flow.
You won't be able to see request/response/detail of the flow without changing size of your window. 
image

But if we decide to make a request with url over for example 10k length, even changing the size of window won't help you to see request/response/detail tab.

About maximum length of a URL
https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers

##### System information

Mitmproxy: 3.0.0.dev64 (commit 6dd336f) 
Python:    3.5.2

fenilgandhi:
While displaying extra-extra long urls in Flow Detail View, there was a no `max_url_len` value set causing a n overflow of url on the screen. The fixes get the max_cols from the ui and sets it to max_url_len
- Fixes #2818

### DIFF
--- a/mitmproxy/addons/core.py
+++ b/mitmproxy/addons/core.py
@@ -1,5 +1,7 @@
 import typing
 
+import os
+
 from mitmproxy.utils import human
 from mitmproxy import ctx
 from mitmproxy import exceptions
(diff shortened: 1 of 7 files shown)


mitmproxy/tools/console/flowview.py
```diff
@@ -30,17 +30,20 @@ def __init__(
...
                 extended=True,
-                hostheader=self.master.options.showhost
+                hostheader=self.master.options.showhost,
+                max_url_len=cols,
             )
         else:
             self._w = urwid.Pile([])
 
 
+
```
kira0204: Hi, `flake` is not happy :) That's why your build is failing

fenilgandhi:
Thanks @kira0204 . I removed that extra line and all is good to go.

PR #2944, opened 2018-03-02
readd Wireshark SSL master secrets docs

Kriechi:


cortesi:
I'd say this is a HOWTO, not a tute. The distinction is a that a tute is a freestanding step-by-step example that incidentally demonstrates features, and a HOWTO answers a "how do I..." question. In this case "how do I use mitmproxy and wireshark".

PR #2945, opened 2018-03-02
Polite shortening of statusbar messages. Fix #1433

### LINKED ISSUE #1433: console: notifications cope very poorly with long data

When we send a long notification to the statusbar - e.g. a traceback from some internal error - the statusbar gets messed up. We should snip this to the first line, and invite the user to view the rest in the event log. 

This behaviour can be seen in, e.g. #1431

kajojify:
PR snips any message that came to statusbar to the first line. It also invites users to see more in eventlog according to #1433. 
**P.S.**
https://github.com/mitmproxy/mitmproxy/compare/master...kajojify:shortening?expand=1#diff-5e13e3689fcd12fb5c83c66addc0e13bR99
Sometimes statusbar gets messages with "\n" or "\t" symbols at the beginning. Thus sometimes statusbar shows just `Error: `.  We want more.

### Result:

Variant | Result
------------ | -------------
1 | image
2 | image
3 | image
4 | image

**Very narrow window:**
image

### DIFF
--- a/mitmproxy/tools/console/statusbar.py
+++ b/mitmproxy/tools/console/statusbar.py
@@ -49,7 +49,7 @@ class ActionBar(urwid.WidgetWrap):
     def sig_message(self, sender, message, expire=1):
         if self.prompting:
             return
-        w = urwid.Text(message)
+        w = urwid.Text(self.prep_message(message))
         self._w = w
         if expire:
             def cb(*args):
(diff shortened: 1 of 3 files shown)


mitmproxy/tools/console/statusbar.py
```diff
@@ -60,6 +60,33 @@ def cb(*args):
     def prep_prompt(self, p):
         return p.strip() + ": "
 
+    def prep_message(self, msg):
```
mhils: `prep_message` isn't very telling, maybe `shorten_message`? We can add a docstring to the function that explains why we do this, e.g. `"""Shorten message so that it fits into a single line in the statusbar."""`

test/mitmproxy/tools/console/test_statusbar.py
```diff
@@ -31,3 +34,29 @@ def test_statusbar(monkeypatch):
...
+                                            ("warn", "")]),
+    ("Line doesn't fit into statusbar", [(None, "Line does..."),
+                                         ("warn", "(more in eventlog)")]),
+    (("alert", "Two lines.\nFirst fits"), [("alert", "Two lines."),
+                                           ("warn", "(more in eventlog)")]),
+    ("Two long lines\nFirst doesn't fit", [(None, "Two long ..."),
+                                           ("warn", "(more in eventlog)")])
+])
+def test_prep_message(message, ready_message):
+    m = mock.Mock()
```
mhils: We can avoid mocking here if you move the `self.master.ui.get_cols_rows()` call into sig_message and pass it as a `max_width` argument.

PR #2956, opened 2018-03-04
Test for #2850

kira0204:
Hi,
I still couldn't figure out why @mhils method was not working but I will not give up on that and come to a conclusion some day.
This provides the test for #2850.
closes #2850

### DIFF
--- a/test/mitmproxy/test_command.py
+++ b/test/mitmproxy/test_command.py
@@ -321,9 +321,9 @@ def test_collect_commands():
             c = command.CommandManager(tctx.master)
             a = TDec()
             c.collect_commands(a)
-            assert not "cmd1" in c.commands
-            assert not "cmd2" in c.commands
-            assert not "empty" in c.commands
+            assert "cmd1" not in c.commands
+            assert "cmd2" not in c.commands
+            assert "empty" not in c.commands
 
 
 def test_decorator():
--- a/test/mitmproxy/test_command.py
+++ b/test/mitmproxy/test_command.py
@@ -321,6 +321,9 @@ def test_collect_commands():
             c = command.CommandManager(tctx.master)
             a = TDec()
             c.collect_commands(a)
+            assert not "cmd1" in c.commands
+            assert not "cmd2" in c.commands
+            assert not "empty" in c.commands
 
 
 def test_decorator():
--- a/test/mitmproxy/test_command.py
+++ b/test/mitmproxy/test_command.py
@@ -1,5 +1,4 @@
 import typing
-from unittest import mock
 from mitmproxy import command
 from mitmproxy import flow
 from mitmproxy import exceptions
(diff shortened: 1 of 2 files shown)


kira0204:
Is there a reason why we don't follow PEP8 guidelines in the given code below `spaces around keyword\parameter equal`?
https://github.com/mitmproxy/mitmproxy/blob/5297e66a875e4d0260b0a683ad2c530cce92155f/test/mitmproxy/addons/test_script.py#L197

kira0204:
The build is failing when `TOXENV=individual_coverage`, I really don't know how individual_coverage is working but the namhe suggests `unit testing` and the only think that is crossing my mind is that the patch is in `command.py` file and the test is in `test_script.py` which might be casing the unit tests to not meet the full coverage cause the integrated test suite worked.

Kriechi:
@kira0204 individual_coverage makes sure that each test-file results in 100% coverage of the corresponding mitmproxy python file.

The current tests fail with this message: 
`FAIL:               mitmproxy/command.py     178      2    99%   136-137`

For example: `mitmproxy/command.py` has a matching test file in `test/mitmproxy/test_command.py`.
Running all tests in this file MUST result in 100% coverage of command.py. Your PR does not pass requirement.

The general idea here that maintenance and refactoring is a lot easier if all tests are combined in a single file that is responsible for the code in question. If other test files would cover the lines in command.py, it makes our lives much harder because we need to look for the tests while refactoring etc.

All in all a minor thing to fix, just add a new test in the right file  - and everybody is happy 👍

kira0204:
Thankyou @Kriechi. 
Since the patch is in `command.py` and handles the error that `hasattr` throws, now since we need to write the test in `test_command.py` , I was thinking of mocking `hasattr` and throw an error as a `side_effect` which will make the test cleaner. Any thoughts?

test/mitmproxy/test_command.py
```diff
@@ -309,6 +310,19 @@ def empty(self) -> None:
...
+    """
+        This tests for the error thrown by hasattr()
+    """
+    with mock.patch("mitmproxy.command.hasattr") as mock_hasattr:
+        mock_hasattr.return_value = False
+        with taddons.context() as tctx:
+            mock_hasattr.side_effect = OSError
+            c = command.CommandManager(tctx.master)
+            a = TDec()
+            c.collect_commands(a)
```
Kriechi: Don't you want to assert something? How would we tell if the test actually works?

kira0204: The idea here was to see if the command throws any error or not. If it doesn't then the test is successful, I thought that `assert` will not be useful in this case hence decided Thoughts?

Kriechi:
The assert look good now.
Only the linter is not happy yet: https://travis-ci.org/mitmproxy/mitmproxy/jobs/350104905#L496

kira0204:
Thank you :)

kira0204:
Hi, is there anything else that needs to be done in this?

test/mitmproxy/test_command.py
```diff
@@ -309,6 +310,22 @@ def empty(self) -> None:
         pass
 
 
+def test_collect_commands():
```
mhils: This test is much more implementation-specific than it should be. Mocking hasattr is wrong, it's not too hard to come up with an addon that raises when hasattr() is invoked on it.

kira0204: Thanks for the feedback, I will make the necessary changes ASAP.

test/mitmproxy/test_command.py
```diff
@@ -309,6 +309,29 @@ def empty(self) -> None:
...
+    """
+        This tests for the error thrown by hasattr()
+    """
+    with taddons.context() as tctx:
+        c = command.CommandManager(tctx.master)
+        a = TCmds()
+        c.collect_commands(a)
+        assert "cmd1" not in c.commands
+        assert "cmd2" not in c.commands
+        assert "empty" not in c.commands
```
mhils: Much better! The asserts don't make sense though, maybe we can add an actual command that still should be loaded?

test/mitmproxy/test_command.py
```diff
@@ -309,6 +309,31 @@ def empty(self) -> None:
...
+
+def test_collect_commands():
+    """
+        This tests for the error thrown by hasattr()
+    """
+    with taddons.context() as tctx:
+        c = command.CommandManager(tctx.master)
+        a = TCmds()
+        c.collect_commands(a)
+        assert "empty" in c.commands
```
kira0204: This should do the work now :)

mhils:
Perfect, thanks!

PR #2957, opened 2018-03-05
updating README according to recent doc updates

kira0204:
Hi,
The installation instructions linked back to old docs website.
I have updated the same


fenilgandhi:
There a few other places where the old url still exists. They can be updated as well.

PR #2961, opened 2018-03-05
Update remaining old docs url with new ones.

fenilgandhi:
Extending work of PR #2957  to replace remaining old docs url with new ones.


cortesi:
Hi there! Thanks for this. Two quick things:

- First, we're just about to transition back to the docs.mitmproxy.org domain. Your efforts are not wasted, though, and I would still like to merge this patch because this would give us consistency and allow us to do a grep/replace to migrate everything in the docs at the same time. 
- Second, there's a UI unit test that broke with this patch. You can see the output here:

https://travis-ci.org/mitmproxy/mitmproxy/jobs/349404260

If you make the adjustment there, we can merge!

fenilgandhi:
I have fixed the broken test. Thanks for helping me out.

PR #2971, opened 2018-03-07
A unified handler for user's scripts exception

### LINKED ISSUE #2962: User's script error handler function
As discussed with @cortesi on slack, right now whenever a user's script throws an error due to various reasons, it is being handled at different places differently.
Therefore we can have a consistent error handler function which can be invoked whenever there is an error
This will also handle #2837 #2838 #2839 
### Function

Signature
`script_error(path, message, lineno, exception)`

What function will do
>"Error in script XXX:NNN MMM” where XXX is the path as specified by the user (the .path attribute of Script), NNN is a line number if we have one, and MMM is a short message

The idea here is to display the above mentioned message in the console app and display the traceback related to the error in the event log.

kira0204:
Hi,
This PR introduces a unified handler which will handle all the user's scripts exception as mentioned in #2962 .
I was finally able to include a short message in the signature which is an optional argument and if it's not provided then the name of the exception type will be printed instead.

The test is a very basic one because of the fact that once merged other tests will do an extensive testing of the function as we move forward to using this function to display error messages. 
Closes #2962

### DIFF
--- a/mitmproxy/addons/script.py
+++ b/mitmproxy/addons/script.py
@@ -5,6 +5,7 @@ import time
 import sys
 import types
 import typing
+import traceback
 
 from mitmproxy import addonmanager
 from mitmproxy import exceptions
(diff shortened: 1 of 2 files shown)


cortesi:
Hey there. This looks like a pretty good start. Could you please take the next step in this PR and use this function throughout the script addon. This will let us see whether the API here needs any modifications.

kira0204:
Hi, the next step would be to use this function to solve the existing issues. I would like to open a new PR and work on top of this code(once this is merged) to fix them.

PR #2972, opened 2018-03-07
Revamp CI generation of snapshots and documentation

cortesi:
This PR prepares the way for generation of snapshots and docs from CI.

- Docs live at docs.mitmproxy.org. The current master is rendered by CI to /master. The current stable release lives at /stable, and is indexed by crawlers (everything else is prohibited in `robots.txt`). When we have archive releases, they will live under their respective major versions, i.e. `/v3`, `/v4`, etc.
- Snapshots live at snapshots.mitmproxy.org. Snapshots are created for all commits on the primary repo - i.e. NOT for PRs. The snapshot name is the TAG if the CI run is due to a tag, otherwise it is the BRANCH. So, the location for the v3.0.0 tag would be "v3.0.0". **Before our next release, we need to work out how this naming scheme interacts with the new model we're following, where each release gets a maintenance branch.** These snapshots are now behind a CDN, so we never have to upload assets to Github - instead, we will link to assets on our own servers in the release notes (except for the source tarfiles, which we don't bother to generate, and can stay on Github). 
- This patch also moves us to Python 3.6 wherever possible. We will probably ditch 3.5 before the next release (fingers crossed). 

Loose ends:

- There's a bit of awkwardness in the naming of some snapshot files. The Windows installer build is not integrated into the `ci.py` file yet, and still uses the version returned by the mitmproxy tool. The wheel file format is mandated, and will always include the version as set in `setup.py`. 
- We need directory listings or something similar for snapshots.mitmproxy.org. A it turns out, haha, this is an ENORMOUS PAIN in S3. 
- We probably need to revert to building Linux binaries under `precise`, but we need a better solution (or an insistence that we don't have one) soon.


cortesi:
I just noticed that there is a corner-case bug here: at the moment, we rely on the CI mechanism to tell us when we're a "privileged commit", by passing us decrypted secrets set in the environment configuration. This doesn't happen for PRs... unless the PR comes from a branch in the repo, like this one. This means that this PR has caused assets to upload to /master. We need to explicitly check the Travis/Appveyor env variables to see if we're a PR and bail before upload if we are. Chalk this up as another loose end.

Issue #1261, opened 2016-06-15
Add transparent mode for own outgoing traffic


tlbdk:
I did a bit of research on how to get PF to also handle own outgoing traffic, it works a bit like this:

Browser -> (outgoing interface) ->  (PF filter rule) -> (lo interface) -> (PF translation rule) -> mitmproxy

There are some limits, the "Browser" needs to run as a different user than mitmproxy as the uid is used as a filter rule to make sure the mitmproxy traffic is also not sent back to it self.

I have a more in-depth explanation of how to do this:
http://tlbdk.github.io/mac/proxy/mitmproxy/fiddler/2016/04/14/redirect-outgoing-traffic-for-user-on-mac.html

All this could be turned into a switch for mitmproxy so it was a bit less work to setup.

mhils:
Thanks for the fantastic write-up! Easily intercepting traffic for our own machine is certainly on the agenda for mitmweb.

refs #630

tomlabaude:
Thanks tlbdk for your detailed post, I was stuck testing transparent mode and it helped me understand my errors.

(btw link is not good anymore, found your post here: https://github.com/tlbdk/tlbdk.github.io/blob/master/_posts/2016-04-19-redirect-outgoing-traffic-for-user-on-mac.md)

Now I question myself about seeing the whole traffic of macOS, for all users & system traffic, not only the user running the browser.
I tried a negation of the proxy user like: `keep state {not or !} user { <user id running the proxy> }` without success.

A solution could be to write a line per other user, but with the possibility to miss one, or to automate by first listing candidate system users and then creating the list:
```
keep state user root
keep state user macports
keep state user _hidd
keep state user _ctkd
keep state user _softwareupdate
keep state user _mdnsresponder
...
```

Also not sure if it's technically feasible, but this hard work with pf is done in 1 line with iptables, would compiling iptables work on macOS?

tlbdk:
If PF supports user negation and you want to transparently proxy all traffic on the machine, you would only need to run mitmproxy under it's own user/uid.

iptables is Linux only and uses the Linux kernel firewall to do it's magic, PF is the macOS equivalent and does not seem to support filtering based on pid's, so I think the easiest options is to change random uid and add rules for that.

srenatus:
[...]

PR #2986, opened 2018-03-11
 Workaround for MacOS transparent proxy

miheerdew:
This workaround (adopted from #1261) allows MacOS users to redirect their machine's outgoing traffic to mitmproxy transparently.


Kriechi:
Thanks! This generally looks good, although I haven't tested it yet.
Would you mind updating this PR to resolve the conflicts?

miheerdew:
Done. Although I checked these instructions last time, perhaps you should also test that it works.

PR #2988, opened 2018-03-14
Fix #2935, Add save button for option panel.

### LINKED ISSUE #2935: Mitmweb changes default value of options

1. run `mitmweb`
2. enable `spoof_source_address` from options
3. close `mitmweb` and open `mitmproxy`
4. Go to options by pressing `O` , `spoof_source_address` will be enabled by default

Even if you try to change the setting using `options` from `mitmproxy` UI, the default is changed until and unless you open `mitmweb` and untick it back.
Idk if it's a solid issue or not but this can trouble the user a lot. 

##### System information

```
Mitmproxy: 4.0.0.dev1295 (commit 10b6f62) 
Python:    3.5.2
Platform:  Linux-4.13.0-36-generic-x86_64-with-Ubuntu-16.04-xenial
```

MatthewShao:
1. Add save button for users to save the option manually.
2. Only save those options differ from the defaults. (as @cortesi  requested).

### DIFF
--- a/web/src/js/components/Modal/OptionModal.jsx
+++ b/web/src/js/components/Modal/OptionModal.jsx
@@ -102,9 +102,6 @@ export default connect(
     }),
     {
         hideModal: modalAction.hideModal,
-        save: () => {
-            optionAction.save()
-            alert('Options saved, your options will be loaded automatically on next start up.')
-        },
+        save: optionAction.save,
     }
 )(PureOptionModal)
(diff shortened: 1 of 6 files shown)


mhils:
This looks good to me! We only save if a user actually clicks on "Save", right?

MatthewShao:
Yes, users are supposed to save options mannully now, when they click the save btn, an alert will appear, saying `Options saved, your options will be loaded automatically on next start up.`.
I just updated the jest snapshot, so the tests should be fixed.

web/src/js/components/Modal/OptionModal.jsx
```diff
@@ -105,6 +102,6 @@ export default connect(
     }),
     {
         hideModal: modalAction.hideModal,
-        save: optionAction.save,
+        save: optionAction.save(),
```
mhils: Doesn't this call .save() right away? This seems wrong.

PR #2997, opened 2018-03-18
Update cut.py

jplochocki:
Line 132: cut.clip command always copy "utf-8" when single cut specified (ie :cut.clip @focus request.method gets "utf-8")


mhils:
That's a bit embarassing. 😄 Thanks! 🍰

PR #3001, opened 2018-03-19
Adding Signal Receiver to Flow Update

### LINKED ISSUE #3000: Number of flows not updated after setting `view filters`
After setting or  removing the `view_filter` in flows list, the number of flows in the status bar is not updated. 
1. open `mitmproxy`
2. open some url (populate flow list)
3. Without navigating through the list,  set a view filter e.g.  `:set view_filter=google` 
4. The flow list  is filtered but the status bar is not updated.
5. On navigating through the focus, the status is updated.
6. Try removing the `view_filter`. Same thing happens.

Before pressing any key | After pressing key
:-------------------------:|:-------------------------:
a |  b

##### System information
Mitmproxy: 4.0.0.dev103(commit 3c7725a)
Python: 3.6.4
OpenSSL: OpenSSL 1.1.0g 2 Nov 2017
Platform: Linux-4.14.21-1-MANJARO-x86_64-with-arch-Manjaro-Linux

fenilgandhi:
- Fixes #3000


Kriechi:
Thanks for this PR! Is there an easy way to test this bugfix / new behaviour?
Unfortunately the current test suite is not very exhaustive for urwid-related things - but we kinda want to not introduce more untested code.

fenilgandhi:
@Kriechi  I searched on it but could not find a test for this. I guess we will have to wait untill some better support is provided by urwid.

PR #3021, opened 2018-03-30
Fixed sequence options setting only last element

### LINKED ISSUE #3015: No way to set a sequence option on the command line

1. Create a file `/tmp/example.py` with the following text:
```python
import typing
from mitmproxy import ctx

class Example:

    def load(self, loader):
        loader.add_option(name='myoption', typespec=typing.Sequence[str],
                          default=[], help='')

    def response(self, flow):
        ctx.log.info(repr(ctx.options.myoption))

addons = [Example()]
```
2. Run `mitmdump -s /tmp/example.py --set myoption=foo --set myoption=bar`
3. Run a request through it: `curl -sx http://localhost:8080 http://httpbin.org/get`
4. Observe that `['bar']` is printed, meaning that `myoption` is set to a sequence of one element — the last one passed on the command line, with the first one ignored.

Looking at [the code](https://github.com/mitmproxy/mitmproxy/blob/79fc4e8ad6cab5181ba39d7f0b0d40f7c11a8389/mitmproxy/optmanager.py#L312-L316), I’m not immediately sure why the multiple `--set` don’t work as I expected them to.

##### System information

Mitmproxy: 4.0.0 (79fc4e8)
Python:    3.5.2
Platform:  Linux-4.4.0-116-generic-x86_64-with-Ubuntu-16.04-xenial

madt1m:
Fixes #3015.

https://github.com/mitmproxy/mitmproxy/blob/79fc4e8ad6cab5181ba39d7f0b0d40f7c11a8389/mitmproxy/optmanager.py#L312-L316

https://github.com/mitmproxy/mitmproxy/blob/79fc4e8ad6cab5181ba39d7f0b0d40f7c11a8389/mitmproxy/optmanager.py#L276-L279

The values are saved only when the whole line has been parsed. Thus, `getattr` will never retrieve the updated sequence from the options. 

The solution I propose to fix the issue, here, is to update the options at each round, to ensure consistent retrieval of previous set effects. 

I understand that this makes a call of `update_known()` just for a couple each time, but the idea was to avoid changing the logic in `parse_setval()`, which in turn would have required some control code in `set` to manually update vals dictionary when the typespec was Sequence[str].
 
Let me know about anything that could/should be changed/improved :)

### DIFF
--- a/mitmproxy/optmanager.py
+++ b/mitmproxy/optmanager.py
@@ -273,10 +273,9 @@ class OptManager:
         )
 
     def set(self, *spec):
-        vals = {}
         for i in spec:
-            vals.update(self._setspec(i))
-        self.update(**vals)
+            self.update(**(self._setspec(i)))
+        
 
     def parse_setval(self, optname: str, optstr: typing.Optional[str]) -> typing.Any:
         """
--- a/mitmproxy/optmanager.py
+++ b/mitmproxy/optmanager.py
@@ -275,7 +275,6 @@ class OptManager:
     def set(self, *spec):
         for i in spec:
             self.update(**(self._setspec(i)))
-        
 
     def parse_setval(self, optname: str, optstr: typing.Optional[str]) -> typing.Any:
         """


cortesi:
Unfortunately, I think this doesn't work. Imagine an addon with 2 options that are inter-dependent (very common). With this patch, when both options are updated simultaneously, the addon gets an update event once for each setting. This means it could be in an inconsistent state in between. 

The correct approach for #3015 is indeed for the parser to know about sequences of items. This means defining a format for a comma-separated sequence of strings (the only sequence option we have at the moment?), and properly handling it on the command-line.

PR #3029, opened 2018-04-02
shift core event loop to asyncio

cortesi:
This patch shifts mitmproxy's core event loop to asyncio. It also changes both mitmproxy console and mitmweb to run off the same asyncio event loop as core, unifying event handing in the main thread. This is just a first step - there are many obvious next steps that will follow in the weeks to come. 

# Performance improvements

I used a modified version of the benchmark to pump 1000 requests through each proxy tool with a concurrency of 50. 

## mitmdump

Before:

mitmdump-pre

After:

mitmdump-asyncio

## mitmproxy

Before (!!!!!):
mitmproxy-pre

After:

mitmproxy-asyncio

## mitmweb

Before:

mitmweb-pre

After:

mitmweb-asyncio

### DIFF
--- a/test/mitmproxy/proxy/protocol/test_http2.py
+++ b/test/mitmproxy/proxy/protocol/test_http2.py
@@ -6,11 +6,11 @@ import tempfile
 import traceback
 import pytest
 import h2
+import time
 
 from mitmproxy import options
 
 import mitmproxy.net
-from mitmproxy.addons import core
 from ...net import tservers as net_tservers
 from mitmproxy import exceptions
 from mitmproxy.net.http import http1, http2
@@ -92,6 +92,10 @@ class _Http2TestBase:
         cls.options = cls.get_options()
         cls.proxy = tservers.ProxyThread(tservers.TestMaster, cls.options)
         cls.proxy.start()
+        while True:
+            if cls.proxy.tmaster:
+                break
+            time.sleep(0.01)
 
     @classmethod
     def teardown_class(cls):
@@ -118,6 +122,7 @@ class _Http2TestBase:
     def teardown(self):
         if self.client:
             self.client.close()
+        self.server.server.wait_for_silence()
 
     def setup_connection(self):
         self.client = mitmproxy.net.tcp.TCPClient(("127.0.0.1", self.proxy.port))
--- a/test/mitmproxy/proxy/protocol/test_http2.py
+++ b/test/mitmproxy/proxy/protocol/test_http2.py
@@ -6,7 +6,6 @@ import tempfile
 import traceback
 import pytest
 import h2
-import time
 
 from mitmproxy import options
 
--- a/test/mitmproxy/proxy/protocol/test_http2.py
+++ b/test/mitmproxy/proxy/protocol/test_http2.py
@@ -92,10 +92,6 @@ class _Http2TestBase:
         cls.options = cls.get_options()
         cls.proxy = tservers.ProxyThread(tservers.TestMaster, cls.options)
         cls.proxy.start()
-        while True:
-            if cls.proxy.tmaster:
-                break
-            time.sleep(0.01)
 
     @classmethod
     def teardown_class(cls):
--- a/test/mitmproxy/proxy/protocol/test_http2.py
+++ b/test/mitmproxy/proxy/protocol/test_http2.py
@@ -90,9 +90,7 @@ class _Http2TestBase:
     @classmethod
     def setup_class(cls):
         cls.options = cls.get_options()
-        tmaster = tservers.TestMaster(cls.options)
-        tmaster.addons.add(core.Core())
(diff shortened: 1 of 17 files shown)


test/mitmproxy/proxy/protocol/test_http2.py
```diff
@@ -90,10 +90,12 @@ class _Http2TestBase:
...
         cls.options = cls.get_options()
-        tmaster = tservers.TestMaster(cls.options)
-        tmaster.addons.add(core.Core())
-        cls.proxy = tservers.ProxyThread(tmaster)
+        cls.proxy = tservers.ProxyThread(tservers.TestMaster, cls.options)
         cls.proxy.start()
+        while True:
+            if cls.proxy.tmaster:
+                break
+            time.sleep(0.01)
```
Kriechi: I'm seeing this block of code multiple times in this PR - maybe we should extract it into a helper function: `cls.proxy.wait_for_master()` or similar?

test/mitmproxy/proxy/protocol/test_websocket.py
```diff
@@ -329,7 +331,12 @@ def test_pong(self):
 
         assert frame.header.opcode == websockets.OPCODE.PONG
         assert frame.payload == b'foobar'
-        assert self.master.has_log("Pong Received from server", "info")
+        for i in range(20):
+            if self.master.has_log("Pong Received from server", "info"):
+                break
+            time.sleep(0.01)
+        else:
+            raise AssertionError("No pong seen")
```
Kriechi: Just to clarify:
This basically tries 20 times to look for a pong log message, and then fails if nothing is found?
So kinda like an implicit timeout / early exit if the test would fail?

test/mitmproxy/tools/test_main.py
```diff
@@ -1,19 +1,25 @@
...
 shutdown_script = tutils.test_data.path("mitmproxy/data/addonscripts/shutdown.py")
 
 
-def test_mitmweb():
+@pytest.mark.asyncio
+async def test_mitmweb():
     main.mitmweb([
         "--no-web-open-browser",
         "-q",
+        "-p", "0",
```
Kriechi: Why port 0?

test/mitmproxy/tservers.py
```diff
@@ -107,11 +103,23 @@ def port(self):
...
+
     def run(self):
+        self.event_loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(self.event_loop)
+        self.tmaster = self.masterclass(self.options)
+        self.tmaster.addons.add(core.Core())
+        self.name = "ProxyThread (%s:%s)" % (
+            self.tmaster.server.address[0],
+            self.tmaster.server.address[1],
+        )
```
Kriechi: I think we have a `human_address` function for this somewhere...?

test/mitmproxy/tservers.py
```diff
@@ -344,11 +356,15 @@ def setup_class(cls):
         cls.chain = []
         for _ in range(cls.n):
             opts = cls.get_options()
-            tmaster = cls.masterclass(opts)
-            tmaster.addons.add(core.Core())
-            proxy = ProxyThread(tmaster)
+            proxy = ProxyThread(cls.masterclass, opts)
             proxy.start()
             cls.chain.insert(0, proxy)
+            while 1:
```
Kriechi: `while True` is a bit more pythonic.

test/mitmproxy/tools/test_main.py
cortesi: Port 0 makes the service bind to an arbitrary open port. We don't care where these daemons are listening, we just want the code to run. Asyncio makes synchronising service termination harder in the presence of threads, so this is a quick fix to prevent these two tests from clashing.

test/mitmproxy/proxy/protocol/test_websocket.py
cortesi: Yes. The test suite will hopefully become nicer as we progress here. This should really be a `wait_for_log` method in our built-in taddons mechanism. Much of the test suite should be migrated over to use that, and server variants should really be fixtures, etc. Our test suite is becoming a burden, to be honest - it took much, much longer to adapt the tests for this patch than it did to do the core migration.

test/mitmproxy/proxy/protocol/test_http2.py
cortesi: Fixed.

test/mitmproxy/tservers.py
cortesi: Fixed. For extra points, guess what language I've been auditing recently?

test/mitmproxy/tservers.py
cortesi: Done.

PR #3033, opened 2018-04-03
Fixed --no-mouse parameter message

emilstahl:
`--no-mouse` is deprecated. Changed parameter to `--set console_mouse=false`


Kriechi:
Thanks! 🎉 👍

Issue #3053, opened 2018-04-13
Use Python 3.6 variable type annotations 


mhils:
A super easy task for someone who wants to get started working on mitmproxy: There are quite a few occasions in the codebase where we still use old-style variable type annotations, e.g. here:
https://github.com/mitmproxy/mitmproxy/blob/660aa87a2471587f3d8a7597a7397be9006db83f/mitmproxy/websocket.py#L66

In Python 3.6 this can be expressed much more nicely without the use of a comment:
```python
self.messages: List[WebSocketMessage] = []
```
(see https://docs.python.org/3.6/whatsnew/3.6.html#whatsnew36-pep526)

A great beginner task to get familar with the whole development process would be to go through the codebase and replace all occurences of `# type:` with the proper new annotation form. That should be pretty straightforward to do, so this issue is strictly meant for first time contributors only. 😃 🚀

kira0204:
Let's close this? Solved via #3059.

PR #3059, opened 2018-04-14
fix Python 3.6 variable type annotations #3053

obscure76:
The purpose is to convert type annotations from old style to pep526. I do not know how to handle type: ignore at the moment. Please let me know if you have suggestions. The conversion in mitproxy/ctx.py is resulting in a circular import orignating from test/test_examples.py so it was not done.


cortesi:
This looks good! Regarding the "type ignore" annotations - I feel we rather over-use these, and it might be an idea to audit them one by one to see if they can be expressed in proper types instead. Scanning the code, there are a number of places where they seem legitimate and quite a few where my gut says we can do better.

PR #3076, opened 2018-04-25
add SO_KEEPALIVE

clyfish:
On our proxy server, after adding SO_KEEPALIVE, memory consumption is reduced from 23G to 16G.


mhils:
Thanks! I think I'm too afraid to ask why your instance even consumes 16G, but this looks like a sensible patch. A recent update to pytest broke our builds, I'll retrigger CI here once #3077 has landed.

Kriechi:
Generally reducing our memory footprint is always a good improvement.
However, in this case I don't really see WHY this might benefit us? Do you have a solid explanation or evidence on what is happening here?

As far as I understand SO_KEEPALIVE, it should not have any user-facing changes. It just puts the kernel into a different checking mode...? So why does this reduce memory consumption?

Kriechi:
ref http://tldp.org/HOWTO/TCP-Keepalive-HOWTO/usingkeepalive.html

PR #3086, opened 2018-04-29
improve server replay from flows support

jbremer:
See also the commit messages! :-)

### DIFF
--- a/mitmproxy/addons/serverplayback.py
+++ b/mitmproxy/addons/serverplayback.py
@@ -166,7 +166,7 @@ class ServerPlayback:
                 return ret
 
     def configure(self, updated):
-        if not self.configured and ctx.options.server_replay:
+        if not self.configured and not updated:
             self.configured = True
             try:
                 flows = io.read_flows_from_paths(ctx.options.server_replay)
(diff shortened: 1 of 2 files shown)


kajojify:
@mhils what do you think about this? Almost 6 months without comments/suggestions :disappointed:

mitmproxy/tools/main.py
```diff
@@ -115,6 +115,7 @@ def run(
         opts.set(*args.setoptions)
         if extra:
             opts.update(**extra(args))
+        master.addons.trigger("configure", ())
```
mhils: If I understand things correctly this shouldn't be necessary.

mitmproxy/addons/serverplayback.py
```diff
@@ -166,7 +175,7 @@ def next_flow(self, request):
                 return ret
 
     def configure(self, updated):
-        if not self.configured and ctx.options.server_replay:
+        if not self.configured and not updated:
```
mhils: This would read all flows form disk whenever any options is changed, so this has unintended side-effects. 

The intended way from within mitmproxy is to call `replay.server`, but that's confusing currently because editing the option is a no-op. For now, the easiest approach is probably to check here if `"server_replay" in updated` and if so, call either `replay.server.file` or `replay.server.stop`.

mhils:
Closing this in favor of #3703! :)

PR #3091, opened 2018-05-02
Remove the tick event

cortesi:
Mitmproxy: the tickless wonder.

### DIFF
--- a/examples/addons/events.py
+++ b/examples/addons/events.py
@@ -167,11 +167,6 @@ class Events:
             loaded.
         """
 
-    def tick(self):
-        """
-            A regular ticker - called approximately once every 100ms.
-        """
-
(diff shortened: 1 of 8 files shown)


juhakivekas:
Hey, I use mitmproxy as an authenticaiton proxy and a request signer for server-to-server communications while doing some more complex testing. The scripting interfaces are excellent for re-signing POST payloads in whatever ways the developers came up with. In some cases authentication tokens need to be kept alive and the tick() event was perfect for that.

Now that tick is removed, is there another way of doing things at regular time intervals? Is the recommended option to use `shced` or some other library for scheluling tasks?

vfaronov:
@cortesi I second Juha’s question. My use case is to programmatically inject delays into flows.

(By the way, it’d be great if breaking changes like this were mentioned in release notes. My script just silently stopped working after upgrading from 3.0 to 4.0.)

PR #3107, opened 2018-05-08
add har_dump example cmdline invocation

0xHJK:
When I was reading this example, I didn't know how to use it. The methods used on the Internet have expired, such as `mitmdump -s "./har_dump.py ./out.har"`. Later, after reading the document carefully, I learned how to use it. So I want to write usage in the file so that others can use it.


Kriechi:
Thanks! 👍 🎉

PR #3110, opened 2018-05-10
Changelog, change meaning of --conf flag

cortesi:


mitmproxy/tools/cmdline.py
```diff
@@ -25,10 +21,10 @@ def common_options(parser, opts):
         help="Show all commands and their signatures",
     )
     parser.add_argument(
-        "--conf",
-        type=str, dest="conf", default=CONFIG_PATH,
+        "--confdir",
+        type=str, dest="confdir", default=core.CONF_DIR,
         metavar="PATH",
-        help="Read options from a configuration file"
+        help="Path to the mitmproxy config directory"
```
Kriechi: Does this also affect where mitmproxy looks for its CA files?
I think it should.

cortesi: Yep, that's everything in the config directory. So firing up `mitmproxy --conf ~/foo.com` recreates the CA files in the new location, etc.

Kriechi: Ok. Just to be safe: do we display any kind of event/log message when we (re-)create the CA?
I feel like users might not be aware that the CA might have changed if they use `--conf`...

Kriechi: Not that strongly - but I'm pretty sure we will get some "my CA has changed and I lost 2 weeks of debugging because mitmproxy didn't tell me!!1!" issues...
Feel free to put it on the v5.0 roadmap - and add a "I told you so" tag with my name on it 😄

cortesi: Don't we already have a "resam told us so" tag? ;)

PR #3041, opened 2018-04-05
Improving the UX of mitmweb

kira0204:
HI,
This PR aims to improve the UX of mitmweb by fixing the position of edit button, a user had to scroll all the way up to start editing and once edited the content had to again scroll back to the top to lock the content.

Before

before-mitmweb

After

after-mitmweb

I hope this gives users a better experience

PR #3136, opened 2018-05-17
[WEB] Make "Edit Flow" button fixed

Mrigank11:
It's really annoying when you've to scroll all the way up just to click the "Edit Flow" button. 
This PR aims to improve user-experience by fixing that button.

Final result:
s

### DIFF
--- a/web/src/css/flowview.less
+++ b/web/src/css/flowview.less
@@ -9,7 +9,8 @@
 }
 
 .edit-flow-container {
-    position: relative;
+    position: fixed;
+    right: 20px;
 }
 
 .edit-flow  {


kira0204:
Hi there,
I had the same problem and worked on it a while back (https://github.com/mitmproxy/mitmproxy/pull/3041) , please feel free to see whether there is an overlap or not :) kinda busy here atm

Mrigank11:
@kira0204 right, they solve the same problem. 
Closing as it is duplicate of #3041.

mhils:
Thank you both! #3041 includes a bunch of unrelated changes in compiled files, we can merge this here.

kira0204:
@mhils I have a doubt here. When we make changes to `mitmweb`, don't we have to push the changes that it does to other files as well. For example the change proposed in this PR would only be applicable if it's compiled right? Am I missing something?

mhils:
We usually update the compiled assets in a separate commit before releases. The diff size can be quite big depending on node package versions, so we want to be a bit conservative here.

kira0204:
Okay thanks, this is great for future purposes.
No idea why that happened, I just tested the change using `gulp`.

PR #3138, opened 2018-05-18
add docker to the main repository

Kriechi:
The plan is to ditch the "automated build" feature of Docker Hub, and instead build & push our own images. We use Travis to perform this, the same way we already build our wheels and pyinstaller files and upload them. This means we do not need the separate docker-release repository any more, and our release check-list shrinks as well, because we don't need to update and tag and push in a second repo any more.

This PR contains general improvements and cleanup things regarding our release workflow.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -11,6 +11,7 @@ environment:
   matrix:
     - PYTHON: "C:\\Python36"
       TOXENV: "py36"
+      PYINSTALLER: "1"
 
 install:
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
@@ -21,7 +22,6 @@ test_script:
   - ps: "tox -- --verbose --cov-report=term"
   - ps: |
       $Env:VERSION = $(python -m mitmproxy.version)
-      $Env:SKIP_MITMPROXY = "python -c `"print('skip mitmproxy')`""
       tox -e cibuild -- build
   - ps: |
       if($Env:RTOOL_KEY) {
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,9 +21,12 @@ matrix:
     - os: osx
       osx_image: xcode7.3
       language: generic
-      env: TOXENV=py36 BDIST=1
+      env: TOXENV=py36 CIBUILD=1 PYINSTALLER=1
     - python: 3.6
-      env: TOXENV=py36 BDIST=1 WHEEL=1
+      env: TOXENV=py36 CIBUILD=1 PYINSTALLER=1 WHEEL=1 DOCKER=1
+      sudo: required
+      services:
+        - docker
     - python: 3.6
       env: TOXENV=individual_coverage
     - python: "3.7-dev"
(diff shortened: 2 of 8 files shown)


mhils:
Would be nice, but I'm not sure if this is going to work:

- The current workflow is: 1) tag release, 2) wait until release is on pypi 3) push docker tag. It's critical that 2 comes before 3, which wouldn't be the case here anymore. I'm not sure if there is a good workaround?
- Dockerhub mirrors branch and tags from GitHub. Our tags here don't all match with Docker releases. Could we make this tolerable somehow?

Kriechi:
Do we actually need the automated build for branches/tags?
Can't we have cibuild/rtool build a dedicated container image on Travis and then push it to DockerHub?

cortesi:
@Kriechi That would work, and would simplify a lot of things.

Kriechi:
This should be ready for review now.
Again lots of changes - but they will make our release workflow sooo much simpler!
I did a few dry runs - but obviously I couldn't test the actual uploads to PyPi and Docker yet.

cortesi:
I love this. It's hugely simpler, and is a big step towards more painless releases. 

We need to find a way to test this completely end-to-end, so we can modify it without me entering a state of existential terror. At the same time, we don't want to confuse users on pypi/Docker or accidentally push something live. I propose that we tackle this by setting up a "dev channel" with separate Docker and pypi accounts ("mitmproxy-dev"). Non-release tags use these accounts and go through the full end-to-end release process. If we like, we could also use the dev channel to publish nightlies down the track.

Kriechi:
On our snapshot server we can cleanup manually ourself.
PyPi has a testing repository that we could use.
Docker Hub is a bit more difficult - I haven't found an easy way yet - maybe setup our own registry?

cortesi:
Let's consider the pros and cons of using an actual separate pypi account so users can say "pip install mitmproxy-dev". For docker, this might be a separate image under the same account. Anything we tag with a "dev-" suffix goes to these channels rather than the main distribution channels (also clarifying the annotated/non-annotated tag thing we may or may not still have going on in our release process). 

This gives us a way to push out dev releases to interested folks for testing. This is quite different from just installing from source or the nightly, because we can explicitly control when dev releases are cut. So, say, we might cut one just after a big new feature has stabilised, or a series of RC releases as we approach a release. 

So the argument I'm making is for a proper dev channel, which is a real distribution channel, which can also act as a sandbox for us to work on the release process.

Kriechi:
I think PyPi already accepts versions for the same package with a `.dev` version. You can only install them with e.g., `pip install --pre mitmproxy`

pip help:
```
--pre
Include pre-release and development versions. By default, pip only finds stable versions.
```
https://pip.pypa.io/en/stable/reference/pip_install/#pre-release-versions


While I think this can be a valuable thing for our users, and a discussion we should have, it is not directly related to this PR, is it?

cortesi:
Pypi is probably our most sensitive spot for releases. The few times we've made errors there have been painful. I'm not fanatical about it, but I would certainly feel less uncomfortable if dev and test releases went to a different package name at least (different pypi account even better). We can discuss what the best way forward is.

The relevance to this PR is that I feel the PR looks good, but we've been unable to test it end-to-end. I'd like to work out a way to do that before the next release, and also to have a way to test it reasonably easily every time we modify the release process. I'm happy to merge this as stands, though.

PR #3150, opened 2018-05-24
check charset in HTML meta tags

0xHJK:
In some cases, there is no charset in response headers, such as www.baidu.com.
so I edited the `_guess_encoding` method to check charset in HTML meta tags.

And in some cases, sites using 'gb2312' encoding can cause decoding errors, such as www.qq.com.
So I changed 'gb2312' and 'gbk' to 'gb18030'. The latter is the superset of the first two.

### DIFF
--- a/mitmproxy/net/http/message.py
+++ b/mitmproxy/net/http/message.py
@@ -170,17 +170,18 @@ class Message(serializable.Serializable):
 
         if not enc:
             if "json" in self.headers.get("content-type", ""):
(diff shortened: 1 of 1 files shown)


0xHJK:
Sorry. I don't have time to write test cases now. I will add it when I have time.

rjt-gupta:
We can close this..

mhils:
Thanks for the heads-up!

PR #3152, opened 2018-05-24
Tame cibuild (somewhat)

cortesi:
Pull all the cibuild state management and computation into a testable class. Isolate environment access to a single function. Add unit tests simulating common CI scenarios. General cleanups.

There's more to be done, but this is a good start.

### DIFF
--- a/release/cibuild.py
+++ b/release/cibuild.py
@@ -48,9 +48,9 @@ class BuildEnviron:
         appveyor_repo_branch = "",
         appveyor_pull_request_number = "",
 
-        build_wheel = False,
-        build_docker = False,
-        build_pyinstaller = False,
+        should_build_wheel = False,
+        should_build_docker = False,
+        should_build_pyinstaller = False,
 
         has_aws_creds = False,
         has_twine_creds = False,
@@ -65,9 +65,9 @@ class BuildEnviron:
         self.travis_branch = travis_branch
         self.travis_pull_request = travis_pull_request
 
-        self.build_wheel = build_wheel
-        self.build_docker = build_docker
-        self.build_pyinstaller = build_pyinstaller
+        self.should_build_wheel = should_build_wheel
+        self.should_build_docker = should_build_docker
+        self.should_build_pyinstaller = should_build_pyinstaller
 
         self.appveyor_repo_tag_name = appveyor_repo_tag_name
         self.appveyor_repo_branch = appveyor_repo_branch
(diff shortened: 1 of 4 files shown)


release/cibuild.py
```diff
@@ -10,163 +10,254 @@
...
+                "TWINE_PASSWORD" in os.environ
+            ),
+
+            docker_username = os.environ.get("DOCKER_USERNAME"),
+            docker_password = os.environ.get("DOCKER_PASSWORD"),
+        )
+
+    @property
+    def has_docker_creds(self) -> bool:
+        return self.docker_username and self.docker_password
```
Kriechi: docker_creds and twine_creds are different.
`has_twine_creds` is an attribute, but `has_docker_creds` is a function-property... why?

cortesi: The docker creds have to be passed to docker login as arguments. The twine creds never leave the environment.

release/cibuild.py
```diff
@@ -10,163 +10,254 @@
...
+    @property
+    def branch(self):
+        return self.travis_branch or self.appveyor_repo_branch
+
+    @property
+    def version(self):
+        name = self.tag or self.branch
+        if not name:
+            raise BuildError("We're on neither a tag nor a branch - could not establish version")
+        return re.sub('^v', "", name)
```
Kriechi: I thought about an edge case when I implemented this regex a few days ago, maybe you find a nifty solution:
* branches that start with a `v` but are not a version number.

Is this an issue we need to catch?

cortesi: Yes. In the next revision, I will verify that tags and branches match the specification in PEP 440 (https://www.python.org/dev/peps/pep-0440/). This will let us distinguish between real releases, dev releases and feature branches based on name. Then we can constrain and document these names.

release/cibuild.py
```diff
@@ -10,163 +10,254 @@
...
         "python",
         "setup.py",
         "-q",
         "bdist_wheel",
-        "--dist-dir", DIST_DIR,
+        "--dist-dir", be.dist_dir,
     ])
-
-    whl = glob.glob(join(DIST_DIR, 'mitmproxy-*-py3-none-any.whl'))[0]
+    whl = glob.glob(os.path.join(be.dist_dir, 'mitmproxy-*-py3-none-any.whl'))[0]
```
Kriechi: Should we pullout the wheel filename into a property.

Kriechi:
Overall, this is sooooo much better! 🎉 👍

release/cibuild.py
cortesi: I wouldn't think it's needed - it's a static glob pattern that never changes and is not computed from the environment.

release/cibuild.py
Kriechi: 👍

PR #3156, opened 2018-05-25
More cibuild tweaks

cortesi:


setup.py
```diff
@@ -89,6 +89,7 @@
             "flake8>=3.5, <3.6",
             "Flask>=1.0,<1.1",
             "mypy>=0.590,<0.591",
+            "parver>=0.1<2.0",
```
Kriechi: Should there be a comma between the two version numbers?

PR #3174, opened 2018-05-28
Add support for IPv6 transparent mode on Windows

### LINKED ISSUE #3080: Support IPv6 transparent proxying on Windows
Transparent mode used to work OK on Windows in version 2.0.2. There was an issue with showing correct Host headers however, so I tried to upgrade to 3.0.4 and found that transparent mode now doesn't work at all. On every client request mitmweb shows the following error message:

> Transparent mode failure: RuntimeError('Cannot resolve original destination.',)

I've tried previous mitmproxy releases and found that this functionality was broken starting from v3.0.0 
RC2. I do understand that this mode is not "officially" supported on Windows, but since it used to work, I hope it can be fixed :)

1. Enable IP forwarding (https://gist.github.com/mhils/e9769fab7b1dd0775335)
2. Set port redirection, e.g.:
> netsh interface portproxy add v4tov4 listenport=80 connectaddress=192.168.0.90 connectport=8080
> netsh interface portproxy add v4tov4 listenport=443 connectaddress=192.168.0.90 connectport=8080
3. Start mitmproxy in transparent mode:
> mitmweb.exe --mode transparent --showhost
4. Install mitmproxy certificate on the phone
5. Set default gateway in phone's Wi-Fi connection properties to the mitmproxy host (e.g. 192.168.0.90)
6. Open any website in phone's browser

##### System information

Mitmproxy: 3.0.4 binary
Python:    3.6.4
Platform:  Windows-10-10.0.16299

mhils:
This fixes #3080. 

@spozdnyakov @Jazzelhawk: You both seem to be on Windows as well - could you please verify that this fixes all your issues? 😃

### DIFF
--- a/mitmproxy/platform/windows.py
+++ b/mitmproxy/platform/windows.py
@@ -338,7 +338,9 @@ class RedirectLocal(Redirect):
         if pid not in self.trusted_pids:
             self.redirect_request(packet)
         else:
-            self.windivert.send(packet, recalculate_checksum=False)
+            # It's not really clear why we need to recalculate the checksum here,
+            # but this was identified as necessary in https://github.com/mitmproxy/mitmproxy/pull/3174.
+            self.windivert.send(packet, recalculate_checksum=True)
 
 
 TConnection = typing.Tuple[str, int]
--- a/mitmproxy/platform/windows.py
+++ b/mitmproxy/platform/windows.py
@@ -1,25 +1,44 @@
-import collections
+import contextlib
 import ctypes
 import ctypes.wintypes
+import io
+import json
 import os
+import re
 import socket
-import struct
+import socketserver
 import threading
 import time
+import typing
 
-import argparse
+import click
+import collections
 import pydivert
 import pydivert.consts
-import pickle
-import socketserver
 
-PROXY_API_PORT = 8085
+REDIRECT_API_HOST = "127.0.0.1"
+REDIRECT_API_PORT = 8085
+
+
+##########################
+# Resolver
+
+def read(rfile: io.BufferedReader) -> typing.Any:
+    x = rfile.readline().strip()
+    return json.loads(x)
+
+
+def write(data, wfile: io.BufferedWriter) -> None:
+    wfile.write(json.dumps(data).encode() + b"\n")
+    wfile.flush()
 
 
 class Resolver:
+    sock: socket.socket
+    lock: threading.RLock
 
     def __init__(self):
-        self.socket = None
+        self.sock = None
         self.lock = threading.RLock()
 
     def setup(self):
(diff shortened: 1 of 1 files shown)


mhils:
Thanks!

How do you determine that you're actually using IPv6? That's definitely an IPv4 request going through the proxy. Those previously also failed due to the IPv6 refactorings.

> After leaving the proxy on for a short while I start getting errors like this:

Some other app on your computer talking something else than HTTP to some other machine. We need some more consistency with the Linux/macOS transparent modes here, I'll restrict it to ports 80/443 by default.

mhils:
Huh, interesting. What's the IPv6 address? Are you sure you're seeing `GET http://10.180.131.92:8081/` for IPv6 requests to that machine?

Jazzelhawk:
The web server machine's address is `fe80::99ce:2cc9:fde0:f2aa`. My machine's address is `fe80::3449:43e0:e7ff:b434%12`. I am running [this](https://gist.github.com/akorobov/7903307) as a IPv6 compatible web server and see logs such this on the web server machine:
```
fe80::3449:43e0:e7ff:b434%enp1s0 - - [28/May/2018 19:09:03] "GET / HTTP/1.1" 200 -
```

mhils:
This is very helpful feedback, thanks! Does 8ebcf8e solve these problems?

Jazzelhawk:
It fixed the address resolution but I'm now getting this error:
```
fe80::3449:43e0:e7ff:b434%17:53142: clientconnect
fe80::3449:43e0:e7ff:b434%17:53142: GET http://fe80::99ce:2cc9:fde0:f2aa:8081/
 << Server connection to ('fe80::99ce:2cc9:fde0:f2aa', 8081) failed: Error connecting to "fe80::99ce:2cc9:fde0:f2aa": [WinError 10060] A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond
fe80::3449:43e0:e7ff:b434%17:53142: clientdisconnect
```
The web server does work correctly as it properly responds when not using the proxy.

mhils:
Huh, that's interesting.

1. Can you access ipv6.cybernode.com with transparent mode turned on?
2. Do you see any indication why we are timing out in Wireshark?
3. Can you access the server if you insert an `addr[0] = addr[0] + "%17"` between these two lines?(assuming %17 is the correct scope) https://github.com/mhils/mitmproxy/blob/8ebcf8ec9647842bfd48e9c890109d2399621a4d/mitmproxy/platform/windows.py#L68-L69

Thanks for the help! My local IPv6 test setup works flawlessly, it's super helpful to catch errors in other configurations upfront.

Jazzelhawk:
1. Unfortunately I don't have access to IPv6 internet
2. Here is a [zip file](https://github.com/mitmproxy/mitmproxy/files/2048225/ipv6.zip) containing capture files for with and without the proxy and also what the server sees when the proxy is in use. I feel you are probably more capable to interpret these than I am.
3. Reinserting the scope ID didn't make a change.

mhils:
Thanks - those PCAPs are really _almost_ identical. Are you using real machines here, or are we talking about something virtualized (VirtualBox/...)? Any changes if you set recalculate_checksum to True in line 341?

Jazzelhawk:
I'm using separate real machines, but changing the recalculate_checksum to True fixed it!

mhils:
This makes no sense, but I take it. Thanks a bunch! 🍰

Jazzelhawk:
I would also suggest a way of changing the default filter when using `mitmdump` if different ports are required to not be restricted. I might be wrong but I don't see a way of doing this at the moment?

mhils:
You can run `python3 -m mitmproxy.platform.windows --help` standalone next to mitmproxy and mitmproxy should use that redirector instance. You can then specify a custom filter there. I totally agree that we need to do better here, but that needs to go in hand with changes for Linux and OSX. Ideally we'd just enable --raw-tcp by default at some point (that requires UI support) and then listen on any ports.

Issue #3159, opened 2018-05-25
Command language implementation


kajojify:
﻿
So, after long and super hot #3074 and #3087 discussions about command language syntax, I am ready to show ideas and some developments regarding the implementation.

The main goal of this issue is to represent the steps of implementation plan and to start implementation discussion in general.

### Parser choosing

In order to construct our language, we need to find suitable parser.
Primary requirements for parser:

- Save current usability features like tab expansion, syntax highlighting (also for partial commands), type checking and underlying type conversion;

- ease of integration with current implementation;

- speed of lexing/parsing and ease of use;

- high quality code, maintenance and no additional dependencies;

- compatible license.

Actually, we have two ways: *write our own lexer/parser* or *use an existing solution*.

**Own parser**

From my point of view, it is worth writing our own parser, only if we have some very special requirements. Yeah, we have one - integration with command line features. But after looking through available parsers, I can say there are pretty good solutions, which will let us integrate without PITA.

So, please, don’t think I am too lazy or I am just afraid of creating parser from scratch. I just think we have a ready-made qualitative solution and can go with it easily.

**Use an existing parser**

There are many of them: [first](https://github.com/webmaven/python-parsing-tools) list, [second](https://wiki.python.org/moin/LanguageParsing) list. These lists are pretty old but still can show, that there are many available variants. I found only one overview of modern parsers - [https://tomassetti.me/parsing-in-python/](https://tomassetti.me/parsing-in-python/). So, yes, I can agree with @cortesi, that not very much has been written about “Python parsing libraries” and “Creating parsers using Python” areas.

After some searchings and dropping completely unsuitable and unmaintained parsers, I chose a few more or less suitable ones and then my choice fell on PLY. Why so?

1.  Compatible license.
    

PLY-3.2 and newer releases are distributed under a BSD-license. So we can use it in mitmproxy codebase.

2.  PLY hasn’t any additional dependencies. It is well-maintained and qualitative. It has well-known creator - [David Beazley](https://github.com/dabeaz).

3.  Yes, we are implementing small language, but it wouldn’t be too bad to have fast-working implementation. For example, this is parsing speed comparison between PLY, funcparselib and pyparsing, when processing long summing expression (мс=ms) :

[...]

Issue #3087, opened 2018-04-29
Command language


cortesi:
# Command language proposal

In mimtproxy, users interact with addons (and by extension with mitmproxy itself) ONLY through [commands](https://docs.mitmproxy.org/stable/addons-commands/) and [options](https://docs.mitmproxy.org/stable/addons-options/). Commands have globally unique names, a set of typed arguments, and a single typed return value. The **command language** is the textual language users use to invoke and combine these typed addon commands.

- *Command names* are always in the form of period separated paths (`addon.command`). Top level commands are indicated with a leading period (`.cut`), and are a shorthand for a fully qualified command where the addon and command name are the same (`.cut` == `cut.cut`).
- *Command invocation* is in the style of Python function invocation, but without commas to separate arguments. In the top scope, brackets can optionally be omitted. In inner scopes brackets are mandatory. Invoked commands are executed once immediately when a command string is run.
- *Deferred commands* are command invocations preceded by a `&`. Deferred commands are executed every time they are used as a value.
- *Command pointers* are command invocations preceded by a `*`. Command pointers are used as a value, with the command and all arguments passed to a receiving command. They are not invoked by the command language itself - whether and how to evaluate a command pointer is up to the receiving command. In our type notation we write a command pointer type as `*cmd`.
- A `source<T>` yields a possibly infinite stream of items of type `T`. Sources are the return value from a **source command**. The Python representation is as yet undefined, but these might be represented as asynchronous iterators or queues.
- A `sink<T>` accepts a stream of items of type `T`. Concretely, a `sink<T>` accepts values of type `T`, `[T]` and `source<T>`, and can be considered to be a union of these types. Sinks are the first argument of a **sink command**. The Python representation is as yet undefined, but might be iterators over asynchronous iterator objects or consumers of asynchronous queues.
- *Two-sided pipe commands*. A command can be both a sink and a source, by accepting a `sink` as its first argument, and returning a `source`.
- *Pipes* are denoted with the `|` operator, and link sinks and sources by passing the LHS as the first argument to the RHS. A command string can compose multiple pipe components into a pipe chain by using two-sided pipe commands. If a pipe chain ends with a source, the final source's emitted values are consumed and discarded. Using the `|` symbol with a non-sink command on the RHS is a type error. Sources can only be used with the pipe operator.
[...]

PR #3183, opened 2018-06-01
Porting the current command language on PLY

kajojify:
Hello! This is my first GSoC PR :tada: 
Sorry for the delay - I spent much time on [thinking about the language](https://github.com/mitmproxy/mitmproxy/issues/3087#issuecomment-386811558), asking @cortesi questions and [playing with PLY](https://github.com/mitmproxy/mitmproxy/issues/3159). This PR transfers the current version of command language to PLY, adapting command features and preparing to have new command language syntax.

Henceforth, I will try to keep my contributions pace high.
### Step 1. Lexing.

So the sequence of actions was:
_Define the places, where we lex anything_ -> _see the interface_ -> _write the lexer according to PLY docs and available examples_ -> _adapt it to the current implementation_ -> _replace the current lexer with the new lexer_ -> _check broken things_.
- **Places.** The current [lexer](https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/command.py#L25) occurs 2 times: in `execute` [method](https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/command.py#L235) and in `parse_partial` [method](https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/command.py#L153).
- **Interface.** Changed in comparison to the current implementation: 
_Creating lexer:_
```
from mitmproxy.language import lexer
lexer = lexer.create_lexer("flow.mark.toggle @focus")
```
Created lexer supports the iteration protocol and also has `token()` method to give you next token from lexed string.

_Getting_ tokens:
```
from mitmproxy.language import lexer
tokens = lexer.get_tokens("flow.mark.toggle @focus")
```
or
```
from mitmproxy.language import lexer
lexer = lexer.create_lexer("flow.mark.toggle @focus")
tokens = [token.value for token in lexer]
```
- **New lexer.** The whole lexer class is built according to [PLY documentation](http://www.dabeaz.com/ply/ply.html#ply_nn3). Separate `build` method is a recommendation from docs. 

- **Broken things.** I know, that some command don't work now and `view.load"mypath"` will be lexed like `view.load` and `"mypath"`, which is incorrect. I still fix all these things. I know this, but my main goal is adapting.

My next steps:
- get rid of lexing bugs;
- implement parser. It is almost ready.


mhils:
This looks like a very reasonable change! 👍 
We also need to convince CI, but other than that I'm good with this. 😃

cortesi:
Thumbs up on this approach. Once we have tests passing, we should merge so we can keep moving on.

setup.py
```diff
@@ -70,6 +70,7 @@
         "kaitaistruct>=0.7,<0.9",
         "ldap3>=2.5,<2.6",
         "passlib>=1.6.5, <1.8",
+        "ply>=3.4, <3.12",
```
Kriechi: Did you actually test this with ply 3.4?
That release is pretty old (2011), so I think we can bump this to at least 3.6 (2015) or even 3.10 (2017)...

kajojify:
### Step 2. Parsing.
According to the wiki page:

> **Parsing** (syntax analysis) is the process of analysing a string of symbols, either in natural language, computer languages or data structures, conforming to the rules of a formal grammar.

**What do we need?** 
_get user input_ -> _lex it_ -> _juxtapose the tokens with formal language grammar_ -> _execute_
The first two steps are done. We get user input, [create a lexer](https://github.com/kajojify/mitmproxy/blob/language/mitmproxy/language/lexer.py#L55) basing on the user input and pass the lexer to the parser. So, now it's time for parser to show itself. 
Before writing the code, we need to describe the formal grammar of our language using BNF notation.

**Minimized BNF notation**, which takes actual tokens into account :

```
command_line : empty 
             | command_call
command_call : COMMAND argument_list"
argument_list : empty
              | argument_list argument
argument : PLAINT_STR
         | QUOTED_STR
         | COMMAND
empty : ""
```
Now let's join BNF notation with the code.

**Parser class.** It was written according to the [PLY docs](http://www.dabeaz.com/ply/ply.html#ply_nn23). In short, every BNF rule has the corresponding method, which describes what to do:
> The semantic behavior of a language is often specified using a technique known as syntax directed translation. In syntax directed translation, attributes are attached to each symbol in a given grammar rule along with an action. Whenever a particular grammar rule is recognized, the action describes what to do.

The main idea of the parser is to pass command manager into parser class as an `__init__` argument and then use `call_strings` method from within, when defining COMMAND token and its argument list. 
`command_manager.call_strings(COMMAND, argument_list)`

**Parser interface.**
_Parser creating:_
```
from mitmproxy.language import parser

command_parser = parser.create_parser(command_manager)
```
_Parsing_
```
from mitmproxy.language import lexer

lexer = lexer.create_lexer("flow.mark.toggle @focus")
command_parser.parse(lexer)
```

kajojify:
### Step 3. Additional features.
Lexer and parser are implemented, but we still have to do something with our interactive features: **syntax highlighting**, **autocompletion**, **underlying type conversion**.
In fact, all these features work like they worked before, except one thing.
Now you can use arbitrary number of white spaces between your tokens and this won't break anything:
image

The lexer has two states: _INITIAL_ and _interactive_. _INITIAL_ state doesn't take into account white spaces and is needed for parsing and execution. _interactive_ - takes into account white spaces and is needed for command prompt features.
When we use `execute` method, we use _INITIAL_ state. When we use `parse_partial`, we use _interactive_ state. So, for example `lexer.get_tokens("cut.clip @focus request.host ")` will give us such list inside `parse_partial`:`['cut.clip', ' ', '@focus', 'request.host', ' ']`, but if to use it with `state=INITIAL` param, it will return `['cut.clip', '@focus', 'request.host']`. It was done in order to take control over tokens separation on command prompt. 

The only thing I still need is to deal with test coverage of lexer and parser. The other tests seem to be fixed.

cortesi:
Excellent progress. Let's fix the coverage so we can merge! :)

Kriechi:
Sorry, I didn't mean to close the PR or the language branch.
Would you mind pushing it again?

PR #3184, opened 2018-06-02
Remove over-zealous assert on metadata types

### LINKED ISSUE #3180: proxyauth option breaks flow state access in addons

1. Create `test.py` addon
```py
from mitmproxy import ctx
class AuthTest:
    def response(self, flow):
        ctx.log.info('Getting flow state')
        flow.get_state()
        ctx.log.info('Got flow state')

addons = [
    AuthTest()
]
```
2. Start `mitmdump`, navigate to example.com and observe the log
```bash
mitmdump -s server/test.py
Loading script server/test.py
...
[::1]:56410: GET http://example.com/
          << 200 OK 606b
Getting flow state
Got flow state
```
3. Re-start `mitmdump` with proxyauth applied and navigate to example.com again
```bash
mitmdump -s server/test.py --set proxyauth=foo:bar
Loading script server/test.py
...
[::1]:56290: GET http://example.com/
          << 200 OK 606b
Getting flow state
  File "server/test.py", line 5, in response
    flow.get_state()
  File "/usr/local/Cellar/mitmproxy/4.0.1/libexec/lib/python3.6/site-packages/mitmproxy/flow.py", line 94, in get_state
    d = super().get_state()
  File "/usr/local/Cellar/mitmproxy/4.0.1/libexec/lib/python3.6/site-packages/mitmproxy/stateobject.py", line 31, in get_state
    state[attr] = get_state(cls, val)
  File "/usr/local/Cellar/mitmproxy/4.0.1/libexec/lib/python3.6/site-packages/mitmproxy/stateobject.py", line 94, in get_state
    return _process(typeinfo, val, False)
  File "/usr/local/Cellar/mitmproxy/4.0.1/libexec/lib/python3.6/site-packages/mitmproxy/stateobject.py", line 77, in _process
    for k, v in val.items()
  File "/usr/local/Cellar/mitmproxy/4.0.1/libexec/lib/python3.6/site-packages/mitmproxy/stateobject.py", line 77, in <dictcomp>
    for k, v in val.items()
  File "/usr/local/Cellar/mitmproxy/4.0.1/libexec/lib/python3.6/site-packages/mitmproxy/stateobject.py", line 81, in _process
    assert isinstance(val, (int, str, bool, bytes))
AssertionError
```

There's a FIXME [right near the code that's breaking](https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/stateobject.py#L80), but I think it's a red herring. My best guess is that proxy auth adds something to the flow state that isn't in the list of allowed types `int, str, bool, bytes`—possibly a `dict`.

##### System information

```bash
mitmdump --version
Mitmproxy: 4.0.1
Python:    3.6.5
OpenSSL:   OpenSSL 1.0.2o  27 Mar 2018
Platform:  Darwin-17.5.0-x86_64-i386-64bit
```

cortesi:
The only real rule here is that we want the metadata to be serializable, and
the current constraint fails in the face of compound types. We don't want to
re-implement a type checker here, so just remove the assert for now.

Once the session mechanism is in core, we'll have better ways to deal with
this.

Fixes #3180

### DIFF
--- a/mitmproxy/stateobject.py
+++ b/mitmproxy/stateobject.py
@@ -77,8 +77,6 @@ def _process(typeinfo: typecheck.Type, val: typing.Any, make: bool) -> typing.An
             for k, v in val.items()
         }
     elif typename.startswith("typing.Any"):
-        # FIXME: Remove this when we remove flow.metadata
-        assert isinstance(val, (int, str, bool, bytes))
         return val
     else:
         return typeinfo(val)


Kriechi:
This change will make `test/mitmproxy/test_stateobject.py:129` fail...

mhils:
I think this assert may actually be quite useful to spot some bugs that may be harder to diagnose if we only have a traceback from serialization. Maybe we just lessen it to what is [supported by tnetstrings](https://github.com/mitmproxy/mitmproxy/blob/09ff5df2fb354a017afa02138b329e615a80c1d0/mitmproxy/io/tnetstring.py#L69-L149)?
```
assert val is None or isinstance(val, (bool, int, float, bytes, str, list, tuple, dict))
```

No strong feelings though, it can go if that is what you all prefer. :)

cortesi:
I'll close this for now and come up with a different solution. On reflection, we probably do just want to check for serializability here. We can't just check the outer types or we'll miss unserializable data nested in containers.

PR #3186, opened 2018-06-04
Add scripting example to intercept & change requests

andersonvom:
This example captures requests to a specific URL and returns an
arbitrary response instead. It's common enough that is useful to have in
the docs.

It also adds links to the Request and Response classes to help
developers writing scripts.


mhils:
Thanks for the PR - feedback on developer experience is always appreciated! Is there any reason why you did not use https://github.com/mitmproxy/mitmproxy/blob/master/examples/simple/send_reply_from_proxy.py? This should achieve pretty much the same, with the difference that we don't need to send the request to the server (in many cases the server can even be offline). 

The link to the source code is also good - we're in fact planning to make this much better with pdoc (https://github.com/mitmproxy/pdoc/issues/148), but linking there in the iterim is definitely a good idea! :smiley:

andersonvom:
Hi @mhils, I actually didn't realize there already was a script elsewhere that did the same thing. I've changed the PR to use that one instead.

andersonvom:
bump

PR #3188, opened 2018-06-05
fix syntax-error on older python interpreters

Kriechi:
This should be included and released in a 4.0.2-bugfix release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ git:
 matrix:
   fast_finish: true
   include:
+    - python: 3.5
+      env: TOXENV=py35 # this just makes sure that our version detection shows an appropriate error message
     - python: 3.6
       env: TOXENV=lint
     - os: osx
--- a/mitmproxy/tools/_main.py
+++ b/mitmproxy/tools/_main.py
@@ -1,0 +1,175 @@
+"""
+This file contains python3.6+ syntax!
+Feel free to import and use whatever new package you deem necessary.
+"""
+
+import os
+import sys
+import asyncio
+import argparse  # noqa
+import signal  # noqa
+import typing  # noqa
+
+from mitmproxy.tools import cmdline  # noqa
+from mitmproxy import exceptions, master  # noqa
+from mitmproxy import options  # noqa
+from mitmproxy import optmanager  # noqa
+from mitmproxy import proxy  # noqa
+from mitmproxy import log  # noqa
+from mitmproxy.utils import debug, arg_check  # noqa
+
+OPTIONS_FILE_NAME = "config.yaml"
+
+
+def assert_utf8_env():
+    spec = ""
+    for i in ["LANG", "LC_CTYPE", "LC_ALL"]:
+        spec += os.environ.get(i, "").lower()
+    if "utf" not in spec:
+        print(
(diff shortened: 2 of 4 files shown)


mhils:
I'm slightly 👎 on this. Can we describe in setup.py that we're 3.6+ only?

Kriechi:
We got multiple confused users opening GitHub issues because of the weird failure mode.
The current check we have is useless and simply does not work.
We already only list Python3.6 in setup.py - unless there is another way to prevent installations on Py3.5 and earlier?

Kriechi:
I renamed the file as requested.
I think this is a good way to solve it, as the error message / behaviour is fully under our control.
If we rely on a weird setup.py hack (?) I'm not sure what will happen in the future...

cortesi:
I think this is too elaborate. As of 9.0, pip supports  [python_requires](https://setuptools.readthedocs.io/en/latest/setuptools.html?highlight=python_requires#new-and-changed-setup-keywords). This gives nice error on the command-line when attempting to install with the wrong Python version:

screen shot 2018-06-11 at 09 46 29

I don't know what percentage of users will have pip < 9.0, but my feeling is that its not large.

Kriechi:
In general I would agree, but this doesn't solve our problem.
Ubuntu xenial ships with pip 8.1.1.
Debian jessie ships with pip 1.5.6.
Raspbian jessie ships with pip 1.5.6.

cortesi:
Ugh. Another possibility - we only really need this check on install. I don't think we need it on invocation. Can we get away with a check in setup.py? I know I'm grasping for "creative solutions" here.

mhils:
1.5.6... 😭

Kriechi:
Isn't setup.py evaluated BEFORE building the wheels?
If you find a working solution I'm happy to switch, but right now we have full control over the error message and version detection.

cortesi:
Well, [wheels include a Python compatibility tag in the filename](https://wheel.readthedocs.io/en/stable/#defining-the-python-version) which can include a minor Python version. The issue currently is that we build them with `-py3-`, which says we're compatible with all Python3 versions. Instead, we could build wheels explicitly for the versions we're compatible with (`-py36-`, `-py367-`). 

So, an alternative to this patch would look like this:

- Specify the python version using `python_requires`
- Add an explicit compatibility tag to our wheel builds 
- Add a Python version check at the top of setup.py that bails with an error if Python version doesn't match. This catches non-wheel installations.

This is all pretty non-invasive, and there's a good argument that we should do all of this anyway, since this lets us catch issues at installation rather than runtime.

mhils:
This would also mean that users on 3.5 will continue to automatically get 3.0 (or whatever the last -py3- release is) when they pip install. Do we want this?

Kriechi:
This would just shift our support requests from "I'm getting a weird syntax error" over to "I'm trying to install the latest mitmproxy, but I only get 3.0"...

cortesi:
OK, so where we're at right now is that we need to cut at least one release that contains the runtime check for py3x. From that point on, the install checks described will be adequate for future releases, where we can roll wheels for specific Python versions we support. I'll merge this and we can add the rest of the paraphernalia later in the current release cycle.

PR #3190, opened 2018-06-05
Adding Docker build for Raspberry Pi 3 (ARMv7 Processor Family)

ironsalsa:
Adding in a build step to create and deliver an image specifically for the Raspberry Pi 3.

### DIFF
--- a/release/cibuild.py
+++ b/release/cibuild.py
@@ -287,7 +287,7 @@ def build_wheel(be: BuildEnviron):  # pragma: no cover
 
 
 def build_docker_image(be: BuildEnviron, whl: str):  # pragma: no cover
-    click.echo("Building Docker image...")
+    click.echo("Building Docker images...")
     subprocess.check_call([
         "docker",
         "build",
@@ -297,10 +297,6 @@ def build_docker_image(be: BuildEnviron, whl: str):  # pragma: no cover
         "--file", "docker/Dockerfile",
         "."
     ])
-
-
-def build_docker_arm_image(be: BuildEnviron, whl: str):  # pragma: no cover
-    click.echo("Building ARMv7 Docker image...")
     subprocess.check_call([
         "docker",
         "build",
--- a/release/cibuild.py
+++ b/release/cibuild.py
@@ -299,6 +299,18 @@ def build_docker_image(be: BuildEnviron, whl: str):  # pragma: no cover
     ])
 
 
+def build_docker_image(be: BuildEnviron, whl: str):  # pragma: no cover
+    click.echo("Building ARMv7 Docker image...")
+    subprocess.check_call([
+        "docker",
+        "build",
+        "--tag", be.docker_tag + "ARMv7",
+        "--build-arg", "WHEEL_MITMPROXY={}".format(whl),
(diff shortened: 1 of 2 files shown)


ironsalsa:
We'll have to see if this passes - I'm uncertain whether copying in the artifacts from earlier in the process is going to work for building for ARMv7. I'm also entirely inexperienced with TravisCI, so I may have butchered the CIBuild.py.

ironsalsa:
It passed everything but the linter step - guess I'll try figuring out Tox to find what I did wrong like I should have beforehand anyway.

Kriechi:
linting fails because you copy-pasted the `build_docker_image` function - you need to rename it, otherwise we have two functions with the same name. But feel free to add the second docker-build subprocess command into the existing function.

And before & after a function need to be two empty lines.

ironsalsa:
Reread the above comment - I'll see if I can combine that function into the existing subprocess.

release/cibuild.py
```diff
@@ -499,6 +508,16 @@ def upload():  # pragma: no cover
         ])
         subprocess.check_call(["docker", "push", be.docker_tag])
 
+    if be.should_upload_docker:
+        click.echo("Uploading ARMv7 Docker image to tag={}...".format(be.docker_tag + "ARMv7"))
+        subprocess.check_call([
```
Kriechi: I think this is not needed - we are already logged in at this point.
You can combine the two `if`s as well.

Kriechi:
Looking good!

Is the Docker tag the only way to distinguish images for different architectures?
I want to avoid having every release as image twice:
... `4.0.1` and `4.0.1ARMv7` ...

release/cibuild.py
ironsalsa: Updated to be in the same block.

ironsalsa:
@Kriechi Using a multitude of tags to distinguish the underlying OS/architecture/etc is the most common way to organize these - check out https://hub.docker.com/r/library/httpd/tags/ for an example of similar things that looks standard to my eyes.

The 2 options here are to either use tags, or separate out the various architecture into multiple separate Docker Hub repos. I prefer tags myself, but we could go the other way as well if that makes more sense here.

ironsalsa:
I'm not sure what I did wrong in the cibuild.py combining the steps - let me know if someone has a chance to look, as it's probably something obvious that I'm not seeing.

Kriechi:
CI is happy now - just some flaky tests that @cortesi promised to fix...

ironsalsa:
All right - that's my fixed PR. I'll keep an eye out for any further requested changes.

Kriechi:
This looks good to me now. @mitmproxy/devs any comments on providing Raspberry Pi ARM-based Docker images of mitmproxy?

cortesi:
@ironsalsa Many thanks!

lhzw:
Hi, there, there's still no arm architecture docker image, I'd like to use it, thank you.

lhzw:
I found the arm images, but it's here:
TAG
latest-ARMv7
docker pull mitmproxy/mitmproxy:latest-ARMv7

I first thecked the "OS/ARCH" column, and didn't find it.
It'll be much better if we can see the architecture under "OS/ARCH".

PR #3196, opened 2018-06-09
Adding Addon to check SSL-Pinning on client

su-vikas:
I have created an Add-On which can help in checking SSL Pinning implementation on client side. For SSL-Pinning broadly there are 5 test cases:
1. Certificate issuer
2. Certificate CommonName
3. Certificate Being On time
4. Certificate Expire time
5. Verifying complete certificate chain

Case 1 can be easily tested by just sending traffic via a proxy, whose root cert is not imported in the trust store of the device/browser/client. 
For Case 2, 3, 4 in my experience I didn't find many easy to use tools, and ended up creating this Add-on. Currently I am not able to hack my way to create Case 5 (any help is appreciated).  

I have tested this for few browsers: Android Chrome, iOS Safari, and Firefox on MacOS, and is working fine. 

I am triggering the Add-On into action on 'clientconnect' event, I hope its the right way to do it. 

Please let me know if something is not as per project's coding practice

Thanks


Kriechi:
Thanks for the PR! 👍 
It looks like you are adding a new addon to our default addon list (e.g., core) that monkey-patches a function. I feel like we probably should address this more directly by modifying & implementing your changes in the `mitmproxy.certs.Cert` class itself. 

We would also appreciate a few tests to make sure your new options actually work as expected.

su-vikas:
Hi, 
Initially I wanted to implement this without monkey-patching, but I couldn't come-up with a neat solution. If you have some ideas on how this can be achieved, please let me know. 

For tests, I will update the PR once I implement it.

mhils:
I moved this into our new examples/contrib directory. Thanks gain! 🍰

PR #3204, opened 2018-06-15
cibuild: cater for tags with "v" prefix

cortesi:


test/release/test_cibuild.py
```diff
@@ -101,8 +101,32 @@ def test_buildenviron_releasetag():
     assert be.is_prod_release
 
 
+def test_buildenviron_namedtag():
+    # Simulates a non-release tag on a branch.
```
Kriechi: Is this even possible?
I think travis builds either the tag or a branch when it is pushed...
https://github.com/travis-ci/travis-ci/issues/4745

cortesi: Wow, that's confusing - let's hope we never have some madman that creates a tag named "master". So, it looks like TRAVIS_TAG == TRAVIS_BRANCH whenever TRAVIS_TAG is not empty? We'll have to tweak the tests, because there's a number of places where we don't adhere to that. 

The idea of the test is less to do with Travis, and more to do with having a test in the suite for a tag that doesn't conform to a version naming convention. We don't have this in our release process anywhere, but I can totally see circumstances in which it might happen (even by if by accident).

PR #3218, opened 2018-06-22
Provide message object for content view addons

### LINKED ISSUE #3213: How to access current message from content view
I use mitmweb (because of Windows). My view requires access to *request* headers that seems impossible in case of *response* tab (`metadata["headers"]` points to response headers).

Is it possible to access the current (selected) flow from a view object? 

I've tried `ctx.master.view.focus`, but it always contains the first flow (it would be what I need otherwise).
I saw a `view.flows.resolve` command but it isn't clear how to call it from an addon. `ctx.master.commands.execute('view.flows.resolve @focus')` just hangs.

It looks like the best option would be to pass the current flow (request/response) as an additional object in `metadata`

pravic:
Sometimes you need to access data from the original request to implement a Content View addon. Right now, such an addon has access only to the request/response headers:

https://github.com/mitmproxy/mitmproxy/blob/9047c4f2a58d6e6b08d8066fb890b2a13de21fb3/mitmproxy/contentviews/__init__.py#L101-L106

So, there is no way to access request headers from response content view.

Also it might be useful to add an additional metadata field to indicate whether it is a request or a response. Something like `metadata["is_request"]`. Or we can do it manually via `isinstance(metadata["message"], http.Request)` or via `if 'query' in metadata`.

closes #3213

### DIFF
--- a/mitmproxy/contentviews/__init__.py
+++ b/mitmproxy/contentviews/__init__.py
@@ -75,7 +75,7 @@ def safe_to_print(lines, encoding="utf8"):
         yield clean_line
 
 
-def get_message_content_view(viewname, message):
+def get_message_content_view(viewname, message, flow=None):
     """
     Like get_content_view, but also handles message encoding.
     """
@@ -99,6 +99,7 @@ def get_message_content_view(viewname, message):
         return "", iter([[("error", "content missing")]]), None
 
     metadata = {}
+    metadata["flow"] = flow
     metadata["message"] = message
     if isinstance(message, http.Request):
         metadata["query"] = message.query
--- a/mitmproxy/contentviews/__init__.py
+++ b/mitmproxy/contentviews/__init__.py
@@ -99,6 +99,7 @@ def get_message_content_view(viewname, message):
         return "", iter([[("error", "content missing")]]), None
 
     metadata = {}
+    metadata["message"] = message
     if isinstance(message, http.Request):
         metadata["query"] = message.query
     if isinstance(message, http.Message):
--- a/mitmproxy/tools/web/app.py
+++ b/mitmproxy/tools/web/app.py
@@ -390,7 +390,7 @@ class FlowContentView(RequestHandler):
         message = getattr(self.flow, message)
 
         description, lines, error = contentviews.get_message_content_view(
-            content_view.replace('_', ' '), message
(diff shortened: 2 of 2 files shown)


mhils:
Thanks for the PR. Why does your response view require access to the request headers? That seems wrong. The PR looks harmless enough to be merged if there's a good use case, but I'm struggling to understand where this would be required. :-)

pravic:
In one case I needed a _request host_ field to properly decode response.
In another one I needed some _request_ headers. 

It's not _that_ wrong. Imagine a stateful protocol wrapped by HTTP, where to decode the next frame you need to maintain a state that depends on previous frames.

mhils:
> It's not that wrong. Imagine a stateful protocol wrapped by HTTP, where to decode the next frame you need to maintain a state that depends on previous frames.

Well I guess those two sentences contradict each other. Are there any standard protocols that do that? Can you give me a specific use case?

Again, I'm not entirely against this, but we at least need to fix every invocation of get_message_content_view and add tests.

pravic:
1) of course, it's not about standard protocols. One specific case: http payload was encrypted, encryption key is derived from _request_ headers. I can't rely on `addon.requestheaders` because traffic in mitmproxy is mixed (i.e. in ContentView there's no single connection where I could maintain some state)

2) I used default parameter, so it's not mandatory to adjust all calls.

mhils:
> of course, it's not about standard protocols. One specific case: http payload was encrypted, encryption key is derived from request headers. I can't rely on addon.requestheaders because traffic in mitmproxy is mixed (i.e. in ContentView there's no single connection where I could maintain some state)

Fair enough. Let's get the other relevant calls adjusted (flow should also be passed for mitmproxy console and mitmdump, that'd be needless inconsistency), and we can merge this. Sorry for being generally sceptic, but we're ultimately the ones bearing the maintenance costs.

And no, we're definitely not feature frozen. :)

pravic:
Huh?

mhils:
Sorry @pravic - I've stumbled upon your response only just now. I've closed this PR as there wasn't any progress after https://github.com/mitmproxy/mitmproxy/pull/3218#issuecomment-400342431.

Again, I'd be happy to merge this, but we please need to adjust all calls to `get_message_content_view` and pass flow objects where possible.

pravic:
I see.

PR #3225, opened 2018-06-26
Update CLI usage instructions for dns_spoofing example

amm0nite:
Hello!
I found that the CLI usage comment is not up to date while updating to mitmproxy 4.


Kriechi:
Thanks! 🎉 👍

PR #3232, opened 2018-06-30
Shifting to Protobuf Serialization

madt1m:
Back with some - hopefully - good news.

After having discussed with @cortesi about the current serialization process, which involves lots of nested traversals of state objects and state attributes, I have tried to define a bit more explicitly our interface of serialization for HTTP Flow. 

Basically, I have observed how HTTPFlows are serialized from states, and tried to "standardize" it, and offer the same result, via Google Protocol Buffers

Some insights on the code:

### Proto
Proto folder contains _proto_ file, alongside with its compiled python code.
The HTTPFlow message definition should act as the _herald_ of the standardization I was talking about before: it has to define, in a single place, what mitmproxy expects from the `mitmproxy.http.HTTPFlow` object.

### DBHandler
This actually is a hint from the future: once protobuf serialization is somewhat mature, the plan is to use it to implement live sessions on disk, storing, retrieving and ordering flows from the SQlite DB - certainly most performant and efficient than storing everything on file. It doesn't do much, right now: it stores and collects list of flows...mostly an interface to how mitmproxy should wrap protobuf, just like what `FlowReader` and `FlowWriter` do right now for tnetstrings.

### Protobuf
Here's the sauce! I have tried to find a balance between cryptic and explicit, that is, I have tried to make automagic and Pythonic everything that I was capable of, without using the states methods.
In some parts you could see some "scripted" behaviour. Some of these were actually necessary, because of how protobuf handles `None` objects -> disclaimer: it cannot, at least not right out of the box.
I know that the overall code can be prettified and can be hardened, so I hope to gather useful feedbacks on that!
Both the dump and load processes should extract and process fields in the same way the `from_state` and `get_state` do.

### The results
I have slightly modified the View addon to quickly test the behaviour of the protobuf code.
This
```
      dh = io.DbHandler(path)
      for f in dh.load():
          self.add([f.copy()]) 
```
actually loads previous stored flows into the view, and to a manual testing the flows are parsed correctly, in all of their fields. 
Furthermore, I have used this code to test in PyCharm that the loaded objects were almost equal (if not for the _stateobject_attribute):
```
def response(f):

    pd = protobuf.dumps(f)
    pl = protobuf.loads(pd)

    with open("flowtest", "wb") as handle:
        fw = FlowWriter(handle)
        fw.add(f)

    with open("flowtest", "rb") as handle:
        fr = FlowReader(handle)
        for i in fr.stream():
            pass
```

Mind that these tests were extremely superficial. So...

### What next
- The implemented Message Definition represents the **current** mitmproxy HTTPFlow object. I will need something like `compat.py` to handle old messages and convert them into the current format.
- Once all of this is mature enough, we probably wouldn't need anymore state objects and methods. This could represent a huge cleanup in the codebase.
- Adjust benchmark code to reflect this new interface, and re-test performances.

### UPDATE
CI is now happy. After the first reviews - and possible changes - there's a fan of possible roads ahead. Some ideas out of the hat:
- Extend this interface to WebSocket and TCP flows, implementing more tests, hardening and expanding this interface to be ready to be released.
- Using this functioning prototype to start working on live sessions on disk, optimizing the code on the road while testing the new features.
- Write new scripts, and change pieces in code where flows are serialized, to start integrating the interface into the codebase.
- Develop some documentation, or script, that prove the benefit of having a standardized module to dump flows, and set the stage for GRPCs messaging exchanging flows.


mitmproxy/io/db.py
```diff
@@ -0,0 +1,40 @@
+import sqlite3
+import os
+
+from mitmproxy.io import protobuf
+
+
+class DbHandler:
```
cortesi: A tiny point, but can we name this DBHandler, please. It's a borderline case, but I think consistent caps in DB is less surprising.

mitmproxy/io/proto/http.proto
```diff
@@ -0,0 +1,93 @@
...
+
+message HTTPFlow {
+    optional HTTPRequest request = 1;
+    optional HTTPResponse response = 2;
+    optional HTTPError error = 3;
+    optional ClientConnection client_conn = 4;
+    optional ServerConnection server_conn = 5;
+    optional bool intercepted = 6;
+    optional bool marked = 7;
+    optional string mode = 8;
```
cortesi: Next step: intercepted, marked and mode should all be items in a general annotation mechanism in the format. That is, this should be in a separate annotations table with a nice general interface.

mitmproxy/io/db.py
```diff
@@ -0,0 +1,40 @@
...
+        self.db_path = db_path
+        self._con = sqlite3.connect(self.db_path)
+        self._c = self._con.cursor()
+        self._create_db()
+
+    def _create_db(self):
+        with self._con:
+            self._con.execute('CREATE TABLE IF NOT EXISTS FLOWS('
+                              'id INTEGER PRIMARY KEY,'
+                              'pbuf_blob BLOB)')
```
cortesi: Next step: the schema here is going to become more complex as we add more tables, indexes and the like. We should pull this out, probably into its own separate .sql file.

mitmproxy/io/proto/http.proto
```diff
@@ -0,0 +1,93 @@
...
+message HTTPRequest {
+    optional string first_line_format = 1;
+    optional string method = 2;
+    optional string scheme = 3;
+    optional string host = 4;
+    optional int32 port = 5;
+    optional string path = 6;
+    optional string http_version = 7;
+    repeated HTTPHeader headers = 8;
+    optional bytes content = 9;
```
cortesi: The content here and in response will often dwarf the rest of the items in this message. When we iterate over flows to show indexes or do searches, we will often not need to read the content, just the surrounding data. I think the content objects should go in a separate large object table and retrieved as needed.

mitmproxy/io/db.py
madt1m: Oooops. That was unintended-persistent typo, DB is the way to go.

mitmproxy/io/proto/http.proto
madt1m: DB wrapper just served as stub code to test protobuf to/from file, so table schemas did not receive much attention (yet). I will add for sure those fields in annotation table, but do you think that should be reflected in the proto format, too?

mitmproxy/io/proto/http.proto
madt1m: For some reason I decoupled the message definition  from the discussions about view and sessions, only focusing on reproducing state object's behavior. Obviously, it wouldn't make sense to parse and save every time HTTP body. I'll move that out in its own message.

PR #3239, opened 2018-07-05
Update howto-transparent.md

bunnyhero:
i got a syntax error on macOS high sierra with the existing example pf.conf line. based on sample pf rules i found online, i believe it was missing the `pass` keyword, which i added.

also, the surrounding text refers to "two lines," when there is only one, so i've updated that too.


Kriechi:
Thanks! 🎉 👍

Issue #3234, opened 2018-07-01
mitmweb isn't protected against DNS rebinding


atx:
The mitmweb interface does not seem to include protection against DNS rebinding. This could be exploited by a malicious website to either access the sniffed data or run arbitrary Python scripts on the filesystem by setting the `scripts` config option. 

I have hacked together a PoC [here](http://arya.atx.name/mmp.html) (nothing really special to be seen though).

mhils:
Thanks for raising this. Any recommendations on how we can fix this best?

atx:
I like the Jupyter implementation - effectively password protect the web interface, but pass an access token to your `webbrowser.open` call not to annoy the user too much. As a simpler alternative, you can implement Host header based whitelist (allowing localhost or IP address access by default).

mhils:
Thanks again - we've just released mitmproxy 4.0.4, which includes the fix from #3243. :)

ILoveStreaming3:
Can I allow binding? Trying to use ngrok

chrisooo3:
Is it possible to turn it off and allow DNS rebinding?

stellarpower:
Would also like to know if this can be disabled - my instance is behind a reverse proxy so I am presuming this is safe enough for what I need it for, and accessing it externally would be helpful.

ananthb:
Yes would love for this to be configurable.

DenisBalan:
As @Kriechi merged https://github.com/mitmproxy/mitmproxy/pull/3243 into the master.
Also mentioning @mhils @atx 

Whats the problem having by default DnsBinding protection enabled,
but have this option configurable?

In our case, for internal usage purposes, we need to fork this repo and revert mentioned PR in order to fulfill our needs.

IMHO, thats not the best way to do that.
Just make secure by default, and configurable for those who `knows what they are doing`

chrisbecke:
What the actual ...... ..... A day spent trying to figure out why a docker swarm Traefik router for `mitmweb.example.com` always returned 404.

hoogi91:
@chrisbecke regarding traefik you need to disable passHostHeader see below example using compose - hope it helps you too :)
```
mitmproxy:
    image: mitmproxy/mitmproxy
    tty: true
    command: "mitmweb --web-host *"
    labels:
      - "traefik.http.services.mitmproxy.loadbalancer.server.port=8081"
      - "traefik.http.services.mitmproxy.loadbalancer.passHostHeader=false"
```

[...]

PR #3243, opened 2018-07-11
mitmweb: protect against dns rebinding

mhils:
This PR attempts to guard mitmweb against DNS rebinding attacks. 

See also:
 - https://github.com/mitmproxy/mitmproxy/issues/3234
 - http://www.tornadoweb.org/en/stable/guide/security.html#dns-rebinding

@atx: My understanding is that enforcing the hostname to match `'(localhost|\d+\.\d+\.\d+\.\d+)'` should effectively mitigate this while still enabling users to access mitmweb from other hosts if they so desire. Do you see errors in my thinking?

In the long term we should probably adopt a jupyter-style solution, but it's more important to get this fixed quickly now.

### DIFF
--- a/mitmproxy/tools/web/app.py
+++ b/mitmproxy/tools/web/app.py
@@ -463,10 +463,20 @@ class SaveOptions(RequestHandler):
         pass
 
 
+class DnsRebind(RequestHandler):
+    def get(self):
+        raise tornado.web.HTTPError(
+            403,
+            reason="To protect against DNS rebinding, mitmweb can only be accessed by IP at the moment. "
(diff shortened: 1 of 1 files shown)


Kriechi:
@mhils: do you want to allow IPv6 addresses too?

atx:
> My understanding is that enforcing the hostname to match '(localhost|\d+\.\d+\.\d+\.\d+)' should effectively mitigate this ...

Yes, that should be sufficient.

mhils:
Thanks @atx! I added support for IPv6 and included an error message for other hostnames, this is ready for final review.

PR #3251, opened 2018-07-19
Command history

kajojify:
This PR implements command history. It tries to work similar to the history in the ubuntu terminal.

### How to use
1. Start mitmproxy.
2. Press `:` and start typing in the status bar.
3. Input a few commands.
4. Press `up` button to see previous command if available. Press `down` button to see next command. 

**A few moments to know.** 
The history contains of `CommandBuffer` objects, not text objects. It was done in order to save cursor position and autocompletion state.
 
When we press `up` button, the command buffer currently typed in the status bar will be saved to the command history. But if the last history buffer has the same text as the current one buffer, the last buffer will be replaced with the current one. We can also add command with `execution` param. It will mean, whenever we decide to execute command. It will be added at the end of the history or will replace either command buffer has the same text or with no text.

ezgif com-video-to-gif

### DIFF
--- a/mitmproxy/tools/console/commander/commander.py
+++ b/mitmproxy/tools/console/commander/commander.py
@@ -1,5 +1,7 @@
 import abc
+import copy
 import typing
+import collections
 
 import urwid
 from urwid.text_layout import calc_coords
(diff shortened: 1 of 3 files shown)


mitmproxy/tools/console/commander/commander.py
```diff
@@ -172,6 +214,11 @@ def keypress(self, size, key):
             self.cbuf.left()
         elif key == "right":
             self.cbuf.right()
+        elif key == "up":
+            self.history.add_command(self.cbuf)
+            self.cbuf = self.history.get_prev() or self.cbuf
+        elif key == "down":
+            self.cbuf = self.history.get_next() or self.cbuf
         elif key == "tab":
```
cortesi: No need to do anything about this now, but we're growing a special little garden of hard-coded keys here. We need to add a key binding context for the command editor, make all of these actions into bind-able commands, and then during interaction with the command edit buffer, we first check if a key is bound in this context, and if not, treat it as an ordinary edit action. No need to do that in this PR,  though.

cortesi:
The proper place for a persistent command-history is probably within the session store, which is not there yet. I feel we should wait on that until the session store can handle it.

PR #3252, opened 2018-07-20
Session - Hybrid DB

madt1m:
First section of Session addon.

Right now, the code implements the logic to create/load a new session DB from file. Store/Retrieve methods will be implemented later, together with View code, depending on the needs.
Schema which will be used by View has been defined:

- **FLOW** table: Flows are dumped here. Will be used for full storage/retrievals.
- **META** table: contains informations that can be quickly retrieved for View modifications --  intercepted, marked, mode, all related to Flow ID.
- **BODY** table: Every time a flow is stored to DB, its request/response body will be extracted and saved here. This way, we will parse bodies only when needed, and FLOW table can be quickly parsed.
- **ANNOTATION** table: This can be used to store View Settings, so to save them in Session, and also to store custom user content related to Flows. A "type" column is there to quicken selections.


madt1m:
PR has been updated with changes of schema and temp DB logic!

mitmproxy/addons/session.py
```diff
@@ -0,0 +1,75 @@
...
+        self.temp = None
+        self.con = None
+        if db_path is not None and os.path.isfile(db_path):
+            self._load_session(db_path)
+        else:
+            if db_path:
+                path = db_path
+            else:
+                # We use tempfile only to generate a path, since we demand file creation to sqlite, and removal to os.
+                self.temp = tempfile.NamedTemporaryFile()
```
cortesi: This creates a file on the file system and returns you an opened file. It's probably not what we want when we're creating a database _de nuovo_. I suggest that you create a temporary directory instead, and then join the tmpdir path with some filename to get the database file path. 

This might be the root cause for the Windows unit test failure.

cortesi:
I'm off to bed, but once the new file creation issue is handled, feel free to merge.

PR #3256, opened 2018-07-22
Protobuf - Benchmark Script

madt1m:
A cleaner and updated addon to test performance of protobuf module.

Right now, the addon: 
- Constantly generates flows with optional size content;
- Periodically gathers generated flows in bunch of N=150;
- Measures time to dump, and store in DB, the bunch;
- Write stats to file (in the form of num_dumps / time_spent) and exits.

Any idea you'd want to see realized here to improve this profiling? Or more, something you feel is not correct in the benchmark process, which makes it not reliable?  
I felt it was necessary to clean and use this, since protobuf interface has evolved more complex, and I had to make sure that performances stayed on par. 

serbenchmark1

serbenchmark2

### DIFF
--- a/test/bench/serialization-bm.py
+++ b/test/bench/serialization-bm.py
@@ -1,0 +1,102 @@
+import asyncio
+import time
+import os
+from statistics import mean
+
+from mitmproxy import ctx
+from mitmproxy.io import db
+from mitmproxy.test import tflow
+
+
+class StreamTester:
+
+    """
+    Generates a constant stream of flows and
+    measure protobuf dumping throughput.
+    """
+
+    def __init__(self):
+        self.loop = asyncio.get_event_loop()
+        self.queue = asyncio.Queue(loop=self.loop)
+        self.dbh = None
+        self.streaming = False
+        self.tf = None
+        self.out = None
+        self.hot_flows = []
+        self.results = []
+        self._fflushes = 0
+        self._stream_period = 0.001
+        self._flush_period = 3.0
+        self._flush_rate = 150
+
+    def load(self, loader):
+        loader.add_option(
+            "testflow_size",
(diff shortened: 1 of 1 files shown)


cortesi:
A few comments/questions:

- Does this obsolete #3126? If so, let's close that PR.
- This looks legit, and the performance is great. There's one aspect of this that we need to fix/think about. At the moment, the flow stream source generates many, many more flows that we actually consume. This is because we wake up, read 150 flows, write them, and then sleep for another 3 second interval. So the queue backs up infinitely. This raises a few questions for me. First, we should probably adjust the benchmark to check the queue length and never let the queue grow bigger than, say, 3x the flush rate. Second, in production we will need to think of a similar back-pressure mechanism - that is, cap the number of "hot" flows we tolerate, and pause flow processing in some way when we exceed that. That's for the future, though.

madt1m:
>Does this obsolete #3126? If so, let's close that PR.

Definitely. I'll close it.

The growing queue issue definitely opens some questions for the future. I'll fix it here as you suggested, but there's something more I am concerned with, if on-the-fly capture method for Sessions will be similar to this: that hardcoded flush period.
It just doesn't feel right to establish a priori a period to "wake" the flushing method, but without a sleep on the `writer`, there would be a ping pong between producer that puts flow in queue, and consumer that immediately flush it, thus losing the optimization of writing flows in stocks.
A way I see to "solve" this is via async writes to DB. This way, while a group of flows is being stored to DB, there would be a convenient time window for `streamer` (or any flow capture method) to fill as much as possible the queue. No idle time, and probably a good throughput.

I could start with testing the [aiosqlite]( https://github.com/jreese/aiosqlite) module right here in the benchmark script, actually.

PR #3268, opened 2018-07-28
Pipes initial support

kajojify:
Hello everyone!
This PR provides basic pipes support.

As we discussed earlier pipes syntax is just a syntactic sugar.
`@focus | replay.client` is the same as `replay.client @focus` or `replay.client(@focus)`
Pipes are just moving the resulting value from the left-hand side  to the right-hand side command as the first argument. One more example:
`console.choose Part [request response] | cut.clip @focus` is the same as 
`cut.clip console.choose(Part [request response]) @focus`

**Implementation details**
Pipe elements are executed one by one from left-hand side to the right-hand side, like in this picture:
image

As soon as the result of the particular pipe element is given, we save it into `self._pipe_value`. In other words, `self._pipe_value` stores the current pipe element value.
As soon as the next pipe element is executed, we take the value from `self._pipe_value` and put it as the first argument into the element.

**Results:**
ezgif com-video-to-gif 2

Now you can also write a plain string into the status bar and no errors will occur:
ezgif com-video-to-gif 3

### DIFF
--- a/mitmproxy/language/lexer.py
+++ b/mitmproxy/language/lexer.py
@@ -7,6 +7,7 @@ import ply.lex as lex
 class CommandLanguageLexer:
     tokens = (
         "WHITESPACE",
+        "PIPE",
         "LPAREN", "RPAREN",
         "LBRACE", "RBRACE",
         "PLAIN_STR", "QUOTED_STR",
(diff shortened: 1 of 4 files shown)


mhils:
Nice! This looks good to me overall - @cortesi may have some more thoughts. :)

What's the reasoning behind allowing plain strings without errors? That doesn't seem too intuitive, is there a particular use case? :)

kajojify:
@mhils I am not insisting on this :) 
My primary thought was to let users see, how any legal token may look like. 
So, imagine I want to use some structure like this `[one two three myaddon.somecommand()]` as an argument of some other command, but I am not sure about the value it will have after `myaddon.somecommand` execution. I can experiment, changing the content of my array to reach the state, where I am sure, that the accepting command will be glad.

There is not any serious case about allowing plain strings without errors. But if we can use array without errors, why not to use plain strings? :)

cortesi:
I like the idea of letting user experiment with the parsing like this, but I think it might be more intuitive in a "show" or "print" command that just outputs its input. I'm afraid users will get confused when they mist-type command names and things pop up on the screen. I'd rather have that be an error in the status bar that says "no such command". 

Not a blocker for this PR, but something I feel we should change.

PR #3274, opened 2018-07-31
Add DEFAULT_EXP_DUMMY_CERT and set to two years

### LINKED ISSUE #3273: Reduce certificate validity for Chrome >= 68

1. Install Chrome 68
2. Start mitmproxy, add mitmproxy-ca to your cert-store
3. Open any site via https
4. Chrome shows NET::ERR_CERT_VALIDITY_TOO_LONG error

In #815 the validity of the certificates was reduced to three years which was enough for Chrome's "not valid longer than 39 month". It seems like Chrome 68 changed this to a shorter value though I haven't found any evidence in the code yet.
After tinkering around I found that the value "71107200" (around 27 month) for DEFAULT_EXP is the greatest value that is accepted by Chrome.

I know that I'm not using the most recent version of mitmproxy but the value for DEFAULT_EXP is the same in the current version so I assume the problem is still the same.

Suggesting to just reduce the validity to something really short, like Let's Encrypt's three month. Should not be a problem and saves us from further reduction in Chrome.

##### System information
-> % mitmproxy --version
Mitmproxy: 3.0.3 
Python:    3.6.6
Platform:  Linux-4.17.8-1-ARCH-x86_64-with-arch

muffl0n:
Helps with Chrome's "certificates can not be valid longer than 27,5 month"
Fixes #3273

### DIFF
--- a/mitmproxy/certs.py
+++ b/mitmproxy/certs.py
@@ -15,7 +15,7 @@ from mitmproxy.coretypes import serializable
 
 # Default expiry must not be too long: https://github.com/mitmproxy/mitmproxy/issues/815
 DEFAULT_EXP = 94608000  # = 24 * 60 * 60 * 365 * 3
-DEFAULT_EXP_DUMMY_CERT = 7776000  # = 90 days
+DEFAULT_EXP_DUMMY_CERT = 63072000  # = 2 years
 
 # Generated with "openssl dhparam". It's too slow to generate this on startup.
 DEFAULT_DHPARAM = b"""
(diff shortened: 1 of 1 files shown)


muffl0n:
It doesn't matter to me as long as we are under 27,5 month. Would feel more relaxed if we'd go as short as possible. But I can surely understand your use case.
Maybe it would be nice if the certificate would be renewed periodically for those long running setups.

Shall I change the value for DEFAULT_EXP_DUMMY_CERT in this PR?

mhils:
> Maybe it would be nice if the certificate would be renewed periodically for those long running setups.

I totally agree, but that needs code which is not there yet and we should ship a bugfix release for this ASAP. :-)

> Shall I change the value for DEFAULT_EXP_DUMMY_CERT in this PR?

Yes, please! ☺️

mhils:
Thanks! 🍰

muffl0n:
Thank you for the quick response and help with this fix! :+1:

mhils:
Thanks for tracking this down and point to the relevant resources - makes our work much easier. Release artifacts for 4.0.4 are currently uploaded, which contains a fix for this issue! 😃 🎉

PR #3277, opened 2018-08-02
Session - Storage Layer

madt1m:
## Introduction

Here we are, this is a crucial step in the development of the Session addon. Considering this, I would appreciate so much any critic, review and comment which can ensure this piece of work to meet the standard quality of the codebase.
This PR contains code to implement, in Session, the following features:
- View (as ordered list of ids);
- View (re)ordering / (re)filtering;
- Storage layer (both in memory, for flows not flushed, and on disk);
- Logic to capture flows -> save them to hot flows -> periodically flush flows to disk -> retrieve them;
- Logic to detect for updated flows already in state, and to replace them opportunely in storage;
- An initial strategy to adapt flush parameters to the incoming load (much prone to improvements);

### Flow Lifecycle

I've sketched down how the logic behind the storage layer behaves. 
I apologize for the DIY feel, and I'll try to enrich the "diagram" *wink-wink* with some comments below.

image

#### Main Data Structures:
- **_view**: a list of tuples (flow.order, flow.id) sorted by first element. It basically keeps which flows are in the current view, ordering them.
- **hot_store**: an ordered dictionary keyed on flow ids. It stores flows when they first arrive or when they are updated, and serve as the first layer of storage.
- **db_store**: the class wrapping sqlite database. Flows are periodically flushed here as protobuf blobs, retrieved and loaded when some other tool is trying to access to storage.

#### Main Behaviour:
- When first arriving in storage - for example via `request()` call, the flow is added to hot store; then, if filter matches, (flow.order, flow.id) - where order is generated based on the current set order - is inserted ordered into the view.
- If afterwards a captured flow is detected to be already in storage, the flow is replaced both in storage (via hot store) and in view.
- Flows in hot storage are periodically flushed to db_store, inserting new tuples in database (for new flows) or replacing existing ones (for updated flows).
- When some tool needs access to storage (for example via `load.view()`), the hot store is checked first. This allows to retrieve flows even if they have not been flushed yet, and always guarantees us to retrieve the **most updated version of flow**. If not present there, the db store is queried.

#### Other features:
- Some components of flows cannot be serialized. Due to this fact, the Session DB keeps a dictionary of "live" components of flow when they are stored, reassembling them upon retrieval.
- As discussed previously, big flow contents are stored in a separate table. A body ledger ensures one-to-one relationship between _flow_ and _body_  table to be sure that only **one** between request and response is stored.
- Session implements a simple, initial strategy to handle bursts of flows: basically, flush period and flush rate are periodically tweaked -- if hot_store starts to grow, flush period decreases, and flush rate increases (and the opposite way around).
- I have developed filtering and ordering routines, since those were strongly related to flow capture and insertion. Their behaviour mimic the one seen in View addon.

### What's Next

Once storage layer is ok, it will be time to implement an API which mirrors View one, to finally test for interoperability and complete the layer. This will require some work out of the addon - as debated some weeks ago - to extend command typing, handle blinker Signals subscription via API and implement some way to "plug" storage modules, to replace View with Session during testing phase (and hopefully, afterwards).

During this piece of work, I wrote a fair amount of new code, and made lots of design choices in the process. Some of them could be good, some bad, so:
- Let me know for everything which just doesn't click in your mind;
- Let me know if you believe that something in logic or implementation doesn't work as it is supposed to do;
- Having full statement coverage doesn't mean that the testing process has good coverage. So, let me know if you believe that I should test for more use cases, or modify the current suite.

word to the CI!

### DIFF
--- a/mitmproxy/addons/session.py
+++ b/mitmproxy/addons/session.py
@@ -6,6 +6,7 @@ import shutil
 import sqlite3
 import os
 
+from mitmproxy import flowfilter
 from mitmproxy import types
 from mitmproxy import http
 from mitmproxy import ctx
@@ -36,8 +37,8 @@ class SessionDB:
     content_threshold = 1000
     type_mappings = {
         "body": {
-            "request" : 1,
-            "response" : 2
+            1: "request",
+            2: "response"
         }
     }
 
@@ -49,6 +50,9 @@ class SessionDB:
         """
         self.tempdir = None
         self.con = None
+        # This is used for fast look-ups over bodies already dumped to database.
+        # This permits to enforce one-to-one relationship between flow and body table.
+        self.body_ledger = set()
         if db_path is not None and os.path.isfile(db_path):
             self._load_session(db_path)
         else:
@@ -91,6 +95,7 @@ class SessionDB:
         is a valid Session SQLite DB.
         :return: True if valid, False if invalid.
         """
+        c = None
         try:
             c = sqlite3.connect(f'file:{path}?mode=rw', uri=True)
             cursor = c.cursor()
@@ -100,7 +105,7 @@ class SessionDB:
             if all(elem in rows for elem in tables):
                 c.close()
                 return True
-        except:
+        except sqlite3.Error:
             if c:
                 c.close()
         return False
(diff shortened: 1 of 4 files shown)


mitmproxy/addons/session.py
```diff
@@ -67,7 +111,278 @@ def is_session_db(path):
...
+            elif len(self._hot_store) < self._flush_rate:
+                self._flush_period *= 1.1
+                self._flush_rate *= 0.9
+
+    def load_view(self):
+        ids = [fid for _, fid in self._view]
+        flows = self.load_storage(ids)
+        return sorted(flows, key=lambda f: self._generate_order(f))
+
+    def load_storage(self, ids=None):
```
cortesi: Return type annotation for this function, please. Does this return a list of flows? Or a list of  `Optional[flow]`?

mitmproxy/addons/session.py
```diff
@@ -67,7 +111,278 @@ def is_session_db(path):
...
+        ids_from_store = []
+        if ids is not None:
+            for fid in ids:
+                # A same flow could be at the same time in hot and db storage. We want the most updated version.
+                if fid in self._hot_store:
+                    flows.append(self._hot_store[fid])
+                elif fid in self.db_store:
+                    ids_from_store.append(fid)
+                else:
+                    flows.append(None)
```
cortesi: What does it mean to append `None` to the list of flows here?

mitmproxy/addons/session.py
```diff
@@ -40,6 +73,16 @@ def __del__(self):
         if self.tempdir:
             shutil.rmtree(self.tempdir)
 
+    def __contains__(self, fid):
```
cortesi: I don't like the use of `__contains__` here. This is an expensive operation, which goes to disk and retrieves IDs of all rows. Being able to use `if x in db` encourages casual use of this operation (see comments below) that we should actually use carefully. 

We need to either make this operation cheap (keep an in-memory set of all IDs on disk or at least make it only retrieve one record), or make it a function like "has_id" so the user is a bit more prepared. I think having an in-memory set of IDs should be fine if this is an important operation.

mitmproxy/addons/session.py
```diff
@@ -67,7 +111,278 @@ def is_session_db(path):
...
+
+    def load_storage(self, ids=None):
+        flows = []
+        ids_from_store = []
+        if ids is not None:
+            for fid in ids:
+                # A same flow could be at the same time in hot and db storage. We want the most updated version.
+                if fid in self._hot_store:
+                    flows.append(self._hot_store[fid])
+                elif fid in self.db_store:
```
cortesi: It seems to me this check is un-necessary - the flow has to be either in the hot store or on disk? As noted above, membership checking for `db_store` is very expensive, so we shouldn't do it unless we have to.

mitmproxy/addons/session.py
```diff
@@ -67,7 +111,278 @@ def is_session_db(path):
...
+
+    def clear_storage(self):
+        self.db_store.clear()
+        self._hot_store.clear()
+        self._view = []
+
+    def store_count(self):
+        ln = 0
+        for fid in self._hot_store.keys():
+            if fid not in self.db_store:
```
cortesi: This as above, the membership check here is really expensive.

mitmproxy/addons/session.py
```diff
@@ -67,7 +111,278 @@ def is_session_db(path):
...
+
+    def set_order(self, order: str) -> None:
+        if order not in orders:
+            raise CommandError(
+                "Unknown flow order: %s" % order
+            )
+        if order != self.order:
+            self.order = order
+            newview = [
+                (self._generate_order(f), f.id) for f in self.load_view()
```
cortesi: This seems to reload all flows from disk. Changing the order of flows on the fly is going to be a very important operation, and needs to be very snappy. I'd rather have _generate_order generate a dict with all the order keys, so we can do this once. Then we can make changing the order an operation that just changes the sort key and re-sorts.

PR #3278, opened 2018-08-02
Charles Proxy HAR converter script

ironsalsa:
The current HAR converter script outputs a format that Charles displays in a way the user doesn't expect, since it doesn't have a display option to show the DNS name instead of the IP address in the Host column. To work around it, I've modified the script to overwrite the response.url key with the value from the host header if present, and fall back to default behavior if not.

It's almost a line-for-line copy of the current har_dump.py with an additional function to make the switch.


ironsalsa:
Fixing the lint failures now.

mhils:
Are we following the HAR spec, is Charles following the HAR spec, or is the HAR spec just too vague?
I think there'd be value in having interop by default. I'd be more in favor of changing the current har dump script than adding a second one which is mostly a copy.

ironsalsa:
I can look into this with http://www.softwareishard.com/blog/har-12-spec/. IMO, it's not really a problem with what data is put in the log.entries.request.url key, but rather how the program interprets it for the user. Putting in the IP address of the server as the URL isn't incorrect when that's how it gets requested, but Charles doesn't automatically adjust the display for users to swap in the DNS name like MITMWeb does.

ironsalsa:
Per [HAR spec](http://www.softwareishard.com/blog/har-12-spec/#request):
"url [string] - Absolute URL of the request (fragments are not included)."
So if the IP is the Absolute URL, then that seems correct to me.

ironsalsa:
Let this one simmer a bit - I've unearthed bugs in the output when loading into Charles, since it's very picky about the format.

ironsalsa:
Ok, got the formatting changed so Charles will take the HAR (it was expecting the URL field to container the http or https). Also let me know if my assumptions are correct - I'm assuming that a positive SSL timing value means https was used, and a 0 or negative means http.

mhils:
Thanks for your work here - can you please integrate your changes in the existing har dumper (i.e. overwrite that one)? I don't think we need to scripts doing almost the same stuff. :)

mhils:
Hi - we'd love to get this merged. Any update on this?

mhils:
I think the current HAR script was just developed from the spec without compatibility with any particular application in mind. I'd be perfectly fine with adjusting our current HAR dumper entirely to what Charles does/supports. :)

PR #3281, opened 2018-08-03
Recreate certs that have expired

muffl0n:
In #3274 we talked about the possibility to replace certs that have expired. This PR sets the lifetime of created certs to 90 days (just like Let's Encrypt) and checks if the certs have expired when calling get_cert().

Things I don't like in this PR and would like to improve before merging:
* the existing test "test_expire", the method "expire" and the "expire_queue" should be renamed. This is more like a "overflow" than an "expire". Maybe even switch to lru_cache?
* the ```raise ValueError('Certificate has expired', name)``` feels kina ugly

Any feedback is greatly appreciated!

### DIFF
--- a/mitmproxy/certs.py
+++ b/mitmproxy/certs.py
@@ -15,7 +15,7 @@ from mitmproxy.coretypes import serializable
 
 # Default expiry must not be too long: https://github.com/mitmproxy/mitmproxy/issues/815
 DEFAULT_EXP = 94608000  # = 24 * 60 * 60 * 365 * 3
-DEFAULT_EXP_DUMMY_CERT = 63072000  # = 2 years
+DEFAULT_EXP_DUMMY_CERT = 7776000  # = 90 days
 
 # Generated with "openssl dhparam". It's too slow to generate this on startup.
 DEFAULT_DHPARAM = b"""
@@ -311,7 +311,7 @@ class CertStore:
             filter(lambda key: key in self.certs, potential_keys),
             None
         )
-        if name:
+        if name and not self.certs[name].cert.has_expired:
             entry = self.certs[name]
         else:
             entry = CertStoreEntry(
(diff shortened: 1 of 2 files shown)


Kriechi:
This would assume you are running mitmproxy for more than 90 days?! 😄 

Instead of raising an exception, you could refactor the try block and prevent the KeyError with a `if name in self.certs` check instead...

muffl0n:
To quote @mhils 
> I personally have a few long-term monitoring projects running for >90 days.

:smiley:

Gave it a shot and replaced the exceptions with a more simpler if/else. Thanks for the hint!

muffl0n:
In the end it was just a simple " and not self.certs[name].cert.has_expired:" added to the already existing "if name"-clause. Could have been so simple in the first place... :roll_eyes:

mhils:
This looks good, thanks! 🍰

mhils:
On a closer look, CI uncovered an actual issue: If someone specifies `--cert foo.pem` on the console, that should always be used, even if the certificate is expired. With the current change, we'd generate a new certificate in these cases, which shouldn't happen.

muffl0n:
That's a tricky one! My first idea is to mark the CertStoreEntry of manually specified certs so that the expiration check only affects the ones automatically created. But I'm not quite sure if that's the best way to go. Maybe storing them in another list (certs vs. custom_certs) would be better?

mhils:
How about adding
```python
def expire_outdated(self) -> None:
    while self.expire_queue[0].cert.has_expired:
        d = self.expire_queue.pop(0)
        self.certs = {
            k: v
            for k, v in self.certs.items()
            if v != d
        }
```
and calling that in `get_cert`?

muffl0n:
I'm still thinking about ditching the explicit "expire_queue" and using something like collections.deque that discards old entries automatically. Or even [lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache) which would discard entrys that haven't been used for some time and not just start with the ones that were created foremost.

mhils:
Any progress on this? :)

muffl0n:
Unfortunately not. I'm not quite sure how much time I can spend on this in the upcoming weeks. Sorry, guys.

mhils:
Closing this for now, happy to reopen if there is any progress! 😃

PR #3282, opened 2018-08-04
Async commands support

kajojify:
Hi there! This PR makes initial steps porting the current command system to asyncio. Some parts don't work correctly yet (like `console.key.bind`). And yes, I understand, that copying the same code in two different places is the last thing we want to do :) But after reading some documentation and stackoverflow answers, I understood, that dealing with sync and async code in one function isn't a good idea. So sometimes you can see two functions/methods, which do the same thing: one does synchronously, the other - asynchronously. 

As you remember, my last PR provided pipes support. Now the main features of the language are more or less in the workable state, but the thing is our commands and types checking implementations aren't adapted to the ideas of these features. I will give you an example. Now we have such command
```
  console.choose.cmd Format export.formats
  console.command export.file {choice} @focus
```

Since we have new language features, we need to rewrite it like
```
console.command [
        export.file 
        console.choose(Format export.formats()) @focus
]
```

And here 2 issues come:
1. `console.choose` must wait until a user makes a choice or closes an overlay. The previous version of this function used a callback for this, but the current version needs new approach. How to do that?
2. We want to call the command. We are calling `call` method from Command object and pass `str` arguments to it. String arguments will be checked for compilance, parsed and converted to some Python objects. The function corresponding to the command will be called with converted `str` arguments.
The thing is commands can be chained now `console.choose(Format export.formats())` and any enclosing command returns some value as is (not necessary `str`) -> https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/command.py#L105-L115 and outer command should do something with it. So, we have the situation, when `console.choose` is being called like `console.choose(Format ['curl', 'httpie', 'raw'])`. The return of `export.formats` is a Python list, Format is a string. Type checking system will try to parse these arguments and convert them to some Python objects, but fail, because arguments must be strings to be parsed.

After long discussions there are two solutions for these two issues:
1. Let mitmproxy execute async commands. 
Now you can define async command like:
```
@command.command("myaddon.acommand")
async acommand(self, arg1: SomeType) -> SomeType:
    ...
```
There are 2 ways to execute commands now: `command_manager.execute` and `command_manager.async_execute`. 
`command_manager.execute` is needed for sync commands.
`command_manager.async_execute` is needed for async commands.
If we try to run async command using sync executor `command_manager.execute`, we will get `ExecutionError` exception. We also have `AsyncExecutionManager`, which tracks running commands.
`AsyncExecutionManager` will probably become a separate addon and get new commands for convenient running commands management.

When you call async command or the command, which has enclosing async command, recursive `ParsedEntity` object is created and traversed then.

2. Simplified check for the return value of enclosing command.

Briefly speaking, what this PR brings:
- All default key bindings are rewritten according to the new language features;
- basic async commands support;
- simple async execution manager is provided;
- minor changes like f-strings support.

### DIFF
--- a/mitmproxy/command.py
+++ b/mitmproxy/command.py
@@ -221,7 +221,7 @@ class CommandManager(mitmproxy.types._CommandBase):
         """
             Parse a possibly partial command. Return a sequence of ParseResults and a sequence of remainder type help items.
         """
-        parts: typing.List[str] = [t.value for t in lexer.get_tokens(cmdstr)]
+        parts: typing.List[str] = lexer.get_tokens(cmdstr)
         if not parts:
             parts = [""]
         elif parts[-1].isspace():
@@ -305,7 +305,6 @@ class CommandManager(mitmproxy.types._CommandBase):
             Execute a command string. May raise CommandError.
         """
         lex = lexer.create_lexer(cmdstr, self.oneword_commands)
-        self.command_parser.async_exec = False
         parsed_cmd = self.command_parser.parse(lexer=lex)
         return parsed_cmd
 
--- a/mitmproxy/command.py
+++ b/mitmproxy/command.py
@@ -38,12 +38,12 @@ RunningCommand = typing.NamedTuple(
     "RunningCommand",
     [
         ("cmdstr", str),
-        ("task", asyncio.Task)
+        ("task", asyncio.Future)
     ],
 )
 
 
-class AsyncExectuionManager:
+class AsyncExecutionManager:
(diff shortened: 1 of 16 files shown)


cortesi:
We've already discussed the approaches here extensively on Slack. I think the summary is a) this is excellent, b) let's merge this to the language branch and rebase to master. :)

cortesi:
I think this looks ready to merge to me. We know that there are adjustments yet to come for the interactive features, and I'm happy for us to work towards that in master.

PR #3294, opened 2018-08-15
Fix XSS scanner failure in test_xss_scanner.py

elijahbal:
Two commits:
1. Fix the test failure (small)
2. Cleanup and refacto a little bit the test (bigger)


elijahbal:
Has anybody some comments on this ?
I actually found a way to fix this with a much shorter patch but with all the tests passing and all, is it really worth it ?

mhils:
Thanks for the PR and sorry for the slow turnaround - we've all been on vacation or otherwise busy. :)

This look good - the XSS scanner thing is not the most beautiful Python code we have, very happy to see improvements here. Thanks! 🍰 😃

PR #3297, opened 2018-08-17
Lib ruamel.yaml version 0.15.55 will cause typing.Sequence[str] unsupported in option.

guangchao0831:
Based on release note of recent version `0.15.55  ruamel.yam`l, it **unmade** CommentedSeq a subclass of list.
This will cause the sequence parsed from yaml file **cannot** be set to option whose type is `typing.Sequence[str]`.


Kriechi:
Thanks for the PR - I think this should already be fixed with ruamel.yaml 0.15.57.
Can you verify that?

guangchao0831:
Thank you for your info and sorry for giving feedback late.

I upgrade `ruamel.yaml` to version `0.15.64` which is the latest version when I add this comment. I found that the list option about ignore_host still cannot be loaded from yaml file which looks like:
`ignore_hosts:`
`    - sample.com`
`    - sample1.com`

BUT if version 0.15.54 is used, above option can be loaded as expected.

mhils:
@guangchao0831 is right. This is tracked in #3321.

PR #3328, opened 2018-09-14
addons/browser: fallback to google-chrome-unstable as last try

bennofs:
On archlinux, if you install `google-chrome-dev` from AUR, the chrome binary is called `google-chrome-unstable`. It shouldn't hurt to try if everything else fails.


mhils:
Sure, thanks! 🍰 😃

PR #3329, opened 2018-09-15
update readme slogan

Kriechi:


kajojify:
Can we merge it? Or this PR is waiting for its time? :)

mhils:
I believe I mentioned this in a previous PR before, but my feeling is that we should rather aim for a shorter slogan that stays constant.

1. Every part of the slogan needs to have a high information/character density. For me personally "HTTP/1, HTTP/2, and WebSockets" wouldn't cut the bar here. No strong feelings if you all think otherwise, but I think we can do without.
2. "HTTP/1, HTTP/2, and WebSockets" would need to be changed once we have better support for plain TCP proxying. AFAIK we also don't display WebSockets in the UI properly, so we should probably either have both (TCP and WebSocket) here or none.

Again, I'm not too invested in this. If you feel very strongly @Kriechi and find someone to support you let's go with this. :)

PR #3343, opened 2018-10-03
Added correct timestamp management in replay

### LINKED ISSUE #3307: Client replay doesn't update request timestamps

1. Save flows to a file, `saved.mitm`
2. Wait a few minutes
3. Use  a command like `mitmdump -n -C saved.mitm -s myscript.py -w replayed.mtim` to replay the saved flows. 
4. Then observe the flows with `mitmproxy -n -r replayed.mitm` and see that the request-response time of the flows are inaccurate because it seems to be calculated using the start time of the requests in `saved.mitm`.

If such behavior is intentional, what's the recommended way to set the request timestamps so that they reflect the time when the requests are actually replayed rather than the previously saved timestamps?

As a workaround, I've added these two lines after the `server.wfile.flush()` line in `addons/clientplayback.py` for my use case:

```
                r.timestamp_end = time.time()
                r.timestamp_start = server.wfile.first_byte_timestamp
```
It seems reasonable but I'm not sure if it's correct.

##### System information
Mitmproxy: 4.0.4
Python:    3.7.0
OpenSSL:   OpenSSL 1.0.2p  14 Aug 2018
Platform:  Darwin-17.5.0-x86_64-i386-64bit

kara71:
Fixes #3307.

### DIFF
--- a/mitmproxy/addons/clientplayback.py
+++ b/mitmproxy/addons/clientplayback.py
@@ -1,5 +1,6 @@
 import queue
 import threading
+import time
 import typing
 
 from mitmproxy import log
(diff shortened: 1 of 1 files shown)


Kriechi:
`request` doesn't exist in this context - I think you are looking for `f.request`?

Issue #2994, opened 2018-03-16
Limiting interception to only those domains that I have the SSL cert for in transparent mode


hariszukanovic:
I believe the correct way to accomplish this would be using --ignore-hosts as indicated with an example in the docs at https://mitmproxy.org/docs/latest/howto-ignoredomains/
using the concrete example
# Ignore everything but example.com and mitmproxy.org:
--ignore-hosts '^(?!example\.com)(?!mitmproxy\.org)'

I am certain that there is SNI support towards those servers that I DO want to intercept (these are my servers which run under HTTPS) because I can provide to mitmproxy the certs.
For those, I provide cert using --cert example.com=/opt/certs/example.com.pem which works perfectly when it actually intercepts for those domains.

##### Steps to reproduce the problem:
Running in transparent mode with --ignore-hosts using regex negative lookahead (to intercept SSL for only some hosts)

Full command (please use your own domain for which you would have the cert available):
/usr/bin/mitmweb --web-iface=192.168.0.252 --listen-host=192.168.0.252 --mode transparent --showhost --cert example.com=/opt/certs/example.com.pem --cert mitmproxy.org=/opt/certs/mitmproxy.org.pem --ignore-hosts '^(?!example\.com)(?!mitmproxy\.org)'

##### Any other comments? What have you tried so far?
I have tried an endless number of regex combinations without success. Even for the simplest possible case. It seems as if either all requests are ignored or all requests are intercepted.

##### System information

mitmproxy --version
Mitmproxy: 3.0.3 
Python:    3.5.5
OpenSSL:   OpenSSL 1.1.0g  2 Nov 2017
Platform:  Linux-3.10.0-693.21.1.el7.x86_64-x86_64-with-centos-7.4.1708-Core

mhils:
Thanks for the detailed write-up. You mentioned transparent mode, so the problem may be that we are matching on the IP as well and ignore if either IP or domain matches. I haven't verified this, but assuming IPv4, does `--ignore-hosts '^(?!example.com)(?![0-9\.]+:)'` work as expected?

hariszukanovic:
It seems to behave equally with the addition of (?![0-9\.]+:)
Also, I tried it using the IP addresses of my server, also without success.
I was unable to find where in the code this comparison takes place, I could try to look into it myself if you indicated the spot please?

mhils:
[...]

PR #3344, opened 2018-10-05
Fixed doc about domain whitelisting

Tey:
This PR fixes the documentation about domains whitelisting. Previous pattern did not work, and solution was given by @mhils in #2994.

I don't think this should close this ticket as there should be a clear way to specify if the pattern should match the IP address and/or the domain, but at least, this PR prevents confusion and gives a workaround.


mhils:
Yes, agreed on everything - thanks for taking the time to properly document the current workarounds. :cake:

PR #3348, opened 2018-10-09
Added link_expander.py example to simple examples

jbrule:
A script we use with mitmdump to expand webpage links before being ingested by a elasticsearch crawler. The crawler in this case is unable to resolve relative links so this step is done by mitmdump with link_expander.py.


examples/simple/link_expander.py
```diff
@@ -0,0 +1,82 @@
...
+			pageText = pageText.replace(*map)
+			
+			#Uncomment the following to print the expansion mapping
+			#print("{0} -> {1}".format(*map))
+		
+		flow.response.text = pageText
+
+
+
+# Python 2.7
```
Kriechi: mitmproxy doesn't support Python 2.7 since a few releases ago - you can safely remove this comment block.

examples/simple/link_expander.py
```diff
@@ -0,0 +1,82 @@
...
+			map_dict["{0}{1}{0}".format(delimiter,rel_link)] = "{0}{1}{0}".format(delimiter,abs_link)
+			
+		for map in map_dict.items():
+			
+			pageText = pageText.replace(*map)
+			
+			#Uncomment the following to print the expansion mapping
+			#print("{0} -> {1}".format(*map))
+		
+		flow.response.text = pageText
```
Kriechi: Please cleanup the empty lines a bit - I find it makes the code actually harder to read in this case.
Our code linting system isn't very happy either: https://travis-ci.org/mitmproxy/mitmproxy/jobs/439246856#L505

Kriechi:
Thanks - this looks much better already!
Please use spaces instead of tabs - and make sure our linting system is happy as well.
https://travis-ci.org/mitmproxy/mitmproxy/jobs/440361409#L522

Kriechi:
Thanks! 🎉 👍 

Yes - you can run `tox -e lint`.

PR #3353, opened 2018-10-19
Support content type suffixes (rfc6839)

kam800:
Currently mitmproxy only recognizes json when request/response content type is `application/json`. Mitmproxy uses default raw view mode, when content mode has vendor-specific media-type with json suffix: `application/vnd.my.company+json`. User has to change view mode to json each time he/she opens the flow.

Mitmproxy only recognises media-types and parameters in content type.
`type/subtype; parameter=value`

This PR introduces suffix support (suffixes are described in [rfc6839](https://tools.ietf.org/html/rfc6839)).
`type/subtype+suffix; parameter=value`

This PR introduces support for the most common suffixes:
- `json`,
- `xml`,
- `wbxml`.

### DIFF
--- a/examples/simple/custom_contentview.py
+++ b/examples/simple/custom_contentview.py
@@ -7,7 +7,7 @@ from mitmproxy import contentviews
 
 class ViewSwapCase(contentviews.View):
     name = "swapcase"
-    content_types = ["text/plain"]
+    media_types = ["text/plain"]
 
     def __call__(self, data, **metadata) -> contentviews.TViewResult:
         return "case-swapped text", contentviews.format_text(data.swapcase())
--- a/mitmproxy/contentviews/__init__.py
+++ b/mitmproxy/contentviews/__init__.py
@@ -27,7 +27,8 @@ from . import (
 from .base import View, VIEW_CUTOFF, KEY_MAX, format_text, format_dict, TViewResult
 
 views: List[View] = []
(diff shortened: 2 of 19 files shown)


mhils:
Thanks for the PR and sorry for the slow turnaround. Is there any widespread adoption of this? I only get 700 Google results for "RFC 6839" 😶. There are [dedicated libraries for parsing mimetypes](https://github.com/dbtsai/python-mimeparse) that don't support this. My current inclination would be to just ignore it and refactor the content views to each implement a should_render() method as opposed to specifying content types. This provides a bit more flexibility to folks who implement their own views, and we can then just check for "+json" for the JSON view and be done.

kam800:
Hi. Sorry for being slow from my side either :) The library mentioned by you looks not popular. The suffix notation has already been there for a while (https://en.wikipedia.org/wiki/Media_type#Suffix). Anyway it is good idea to implement should_render(), I will try to refactor. Thanks!

Kriechi:
In general I am 👍 on supporting stuff that has an RFC (without an later published deprecation RFC of course). What @mhils sounds more suitable for us, @kam800 if you are still working on this, we are happy to accept a new PR with the proposed `should_render` feature.

kam800:
Yeap, I totally aggree with you guys. I just was a little short of time. PR is coming shortly 🤞

PR #3355, opened 2018-10-19
Make content view lines limit configurable

kam800:
Currently mitmproxy by default limits a flow content to 512 lines. This PR makes that cutoff configurable.


mhils:
Thanks! 🍰 😃

PR #3356, opened 2018-10-20
Make private keys readable only by the owner

fghzxm:


mhils:
Thanks! Can we move this into a decorator to simplify things and add a test? 

I'm thinking of something like this, which could be tested in isolation:
```python
@contextmanager
def umask_owner_readable():
    original_umask = os.umask(0)
    os.umask(original_umask | 0o77)
    try:
        yield
    finally:
        os.umask(original_umask)

def test_umask_owner_readable():
    with umask_owner_readable():
        # write file
    # test permissions
```

fghzxm:
@mhils Thanks for your advice.  I have updated my branch.

mhils:
Awesome! Thank you so much for the quick turnaround, really appreciated. I'm not sure if Windows tests will be passing (we may just want to exclude them), but first it looks like GitHub needs to get things in order :)

PR #3364, opened 2018-10-25
Added howto information for client configuration

alexlauerman:
This method is significantly easier in my opinion and should also be included. Greatly reduces the barrier for entry.


q2dg:
Is there any technical explanation why this PR has been closed? 
Thanks!

PR #3366, opened 2018-10-25
Corrected keep_host_header usage option

alexlauerman:
I tested this locally to verify this change is correct.


mhils:
Thanks! 🍰

PR #3374, opened 2018-11-04
docs: Add mitm.it link in the certificates docs.

abhigyank:
In the page https://docs.mitmproxy.org/stable/concepts-certificates/, there is usage of mitm.it and it is not a link which seems odd on reading through the documentation. This PR adds the link to the usage.


abhigyank:
@mhils  Does the failing tests have anything to with the change I have made in this PR?

mhils:
That looks like a false positive to me. Thanks for the PR! 🍰

PR #3376, opened 2018-11-05
Add Organization field for the generated certificate

Abcdefghijklmnopqrstuvwxyzxyz:
Some clients (for example, the client below) will check organization fields`(O)` to avoid fake certificate generated by others. So I chang the tls layer and cert store to support the organization field`(O)` in the generated certificate.

So far, the organization field`(O)` will be duplicated from the original certificate from server.

example

> Though unbelievable, some clients check their certificate by its Organization field.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,5 +28,3 @@ sslkeylogfile.log
 coverage.xml
 web/coverage/
 .mypy_cache/
-
-# To triggle CI, will be delete later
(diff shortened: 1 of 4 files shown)


mhils:
@abhigyank: fair enough - maybe we just go with `organization`? Feel free to create a PR, this is too minor to track it in a separate issue.

abhigyank:
Sure, I am on it. :)

Issue #3072, opened 2018-04-23
query string arrays are not fully displayed


StevenVanAcker:
##### Steps to reproduce the problem:

1. visit through mitmproxy/mitmdump e.g. http://example.com/?first=value&arr[]=foo+bar&arr[]=baz
2. Check the query parameters in the request
3. Notice that they contain more data than mitmproxy/mitmdump shows

##### Any other comments? What have you tried so far?

The following script shows all the data:

```
#!/usr/bin/env python3

from urllib.parse import urlparse, parse_qs

url = "http://example.com/?first=value&arr[]=foo+bar&arr[]=baz"
parts = urlparse(url)
print(parse_qs(parts.query))
```

##### System information

Mitmproxy: 3.0.4
Python:    3.5.2
OpenSSL:   OpenSSL 1.0.2g  1 Mar 2016
Platform:  Linux-4.10.0-42-generic-x86_64-with-Ubuntu-16.04-xenial

mhils:
Repro:
```

ggarg2906sumo:
I will to work on this.

PR #3404, opened 2018-12-02
Query array fix

rjt-gupta:
Fix for #3072 and #3254 

Added `format_pair` function and changed the implementation of `format_dict`.
Did relevant changes in urlencoded.py

### DIFF
--- a/mitmproxy/contentviews/base.py
+++ b/mitmproxy/contentviews/base.py
@@ -35,6 +35,35 @@ class View:
         raise NotImplementedError()  # pragma: no cover
 
 
+def format_pair(item: TViewLine) -> typing.Iterator[TViewLine]:
+
+    """
+    Helper function that accepts a list of (k,v) pairs into a list of
+    [
+        ("key", key    )
+        ("value", value)
+    ]
+    where key is padded to a uniform width
+    """
+
+    max_key_len = max((len(k[0]) for k in item), default=0)
+    max_key_len = min((max_key_len, KEY_MAX), default=0)
+
+    for k in item:
+        if isinstance(k[0], bytes):
+
+            temp = b"k[0]" + b":"
+        else:
+            temp = k[0] + ":"
+
+        key = temp.ljust(max_key_len + 2)
+
+        yield [
+            ("header", key),
+            ("text", k[1])
+        ]
+
+
 def format_dict(
         d: typing.Mapping[TTextType, TTextType]
 ) -> typing.Iterator[TViewLine]:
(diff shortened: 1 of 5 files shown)


rjt-gupta:
This is ready for review.

mitmproxy/contentviews/base.py
```diff
@@ -35,6 +35,35 @@ def __call__(self, data: bytes, **metadata) -> TViewResult:
         raise NotImplementedError()  # pragma: no cover
 
 
+def format_pair(item: TViewLine) -> typing.Iterator[TViewLine]:
```
mhils: This method should be `format_pairs` and not `format_pair`, as it formats multiple (key, value) pairs.

mhils: Second, `item` is a bad name for a list. `items` would work, or `pairs`, or something like that.

mhils: Also, the TViewLine type is used to indicate `(style, text)` tuples for rendering. That's not really what we have as input here, which should just be an `Iterable[Tuple[TTextType,TTextType]]`.

mitmproxy/contentviews/base.py
```diff
@@ -35,6 +35,35 @@ def __call__(self, data: bytes, **metadata) -> TViewResult:
...
+        ("key", key    )
+        ("value", value)
+    ]
+    where key is padded to a uniform width
+    """
+
+    max_key_len = max((len(k[0]) for k in item), default=0)
+    max_key_len = min((max_key_len, KEY_MAX), default=0)
+
+    for k in item:
```
mhils: We can make this much more readable by using `for key, value in pairs`. :-)

mitmproxy/contentviews/base.py
```diff
@@ -35,6 +35,35 @@ def __call__(self, data: bytes, **metadata) -> TViewResult:
...
+        if isinstance(k[0], bytes):
+
+            temp = b"k[0]" + b":"
```
mhils: That's definitely wrong.

mitmproxy/contentviews/base.py
```diff
@@ -47,18 +76,7 @@ def format_dict(
...
-        if isinstance(key, bytes):
-            key += b":"
-        else:
-            key += ":"
-        key = key.ljust(max_key_len + 2)
-        yield [
-            ("header", key),
-            ("text", value)
-        ]
+    return format_pair(list(d.items()))
```
mhils: There's no need to wrap `d.items()` in `list()`.

mitmproxy/contentviews/base.py
rjt-gupta: okay!

rjt-gupta: Sorry i misinterpreted TViewLine type.

mitmproxy/contentviews/base.py
rjt-gupta: Sorry about this.

rjt-gupta:
It was a learning experience.Thanks!
I will make all the changes.

PR #3405, opened 2018-12-03

mhils:


rathann:
Would it be possible to support brotlipy 1.0.7 and wsproto 0.12.0? Fedora development branch has these versions.

PR #3410, opened 2018-12-07
Charset in meta tags

rjt-gupta:
Added tests for #3150 (for all cases).


rjt-gupta:
My other commits also came

Issue #3354, opened 2018-10-19
Two error responses on bad HTTP request


slepogin:
##### Steps to reproduce the problem:

1. Start mitmdump or mitmproxy
2. $ echo "GET bad request HTTP/1.1" | nc 0.0.0.0 8080

This would produce two errors in the response:

```
HTTP/1.1 400 Bad Request
Server: mitmproxy 5.0.0.dev
Connection: close
Content-Length: 290
Content-Type: text/html

<html>
            <head>
                <title>400 Bad Request</title>
            </head>
            <body>
            <h1>400 Bad Request</h1>
            <p>HttpSyntaxException(&quot;Bad HTTP request line: b&#x27;GET test me HTTP&#x27;&quot;,)</p>
            </body>
        </html>HTTP/1.1 502 Bad Gateway
Server: mitmproxy 5.0.0.dev
Connection: close
Content-Length: 327
Content-Type: text/html

<html>
            <head>
                <title>502 Bad Gateway</title>
            </head>
            <body>
            <h1>502 Bad Gateway</h1>
            <p>ProtocolException(&quot;HTTP protocol error in client request: Bad HTTP request line: b&#x27;GET test me HTTP&#x27;&quot;,)</p>
            </body>
        </html>
```
Expected behaivor - only one error.

##### Any other comments? What have you tried so far?

Happens due to propagating of ProtocolException from **protocol/http.py:292** to **server.py:143**

mhils:
Oops. Well I guess two is better than none. Thanks for the report and the easy repro! 
refs #1775

Kriechi:
We just merged a brand new protocol implementation - if you can repeat the problem with the latest version from git-master, please re-open this issue.

PR #3415, opened 2018-12-10
Fix: Two error responses on bad HTTP request

sudo45:
Fixed issue #3354

When the request is not sintactically correct, there is already an exception raised(HttpSyntaxException), so there is no need to raise a further ProtocolException. 
So this fix added a check to verify that, before raise a ProtocolException, it has not already been raised an HttpSyntaxException.

### DIFF
--- a/mitmproxy/proxy/protocol/http.py
+++ b/mitmproxy/proxy/protocol/http.py
@@ -289,9 +289,12 @@ class HttpLayer(base.Layer):
             f.request = None
             f.error = flow.Error(str(e))
             self.channel.ask("error", f)
-            raise exceptions.ProtocolException(
-                "HTTP protocol error in client request: {}".format(e)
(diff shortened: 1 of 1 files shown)


mitmproxy/proxy/protocol/http.py
```diff
@@ -289,9 +289,12 @@ def _process_flow(self, f):
             f.request = None
             f.error = flow.Error(str(e))
             self.channel.ask("error", f)
-            raise exceptions.ProtocolException(
-                "HTTP protocol error in client request: {}".format(e)
-            ) from e
+            if not isinstance(e, exceptions.HttpSyntaxException):
```
mhils: It's not clear from the PR when this is still triggered. What's the story here? Can we maybe add an explaining comment?

PR #3416, opened 2018-12-11
Modifications to examples/complex/har_dump.py

wishdev:
Updates the har_dump.py example to allow for the creation of a HAR file per flow.

Also allows for different compression options from within the standard Python library.

### DIFF
--- a/examples/complex/har_dump.py
+++ b/examples/complex/har_dump.py
@@ -1,24 +1,40 @@
 """
 This inline script can be used to dump flows as HAR files.
 
-example cmdline invocation:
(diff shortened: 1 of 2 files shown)


nealmcb:
See also intended changes to the same script in a separate PR: Charles Proxy HAR converter script #3278

mhils:
Super sorry for the nonexistent reply here - not sure how that slipped through. We now have https://github.com/mitmproxy/mitmproxy/tree/master/examples/contrib, if you are interested in getting your HAR changes in please submit a PR for that and I promise we'll be much faster! ❤️

PR #3420, opened 2018-12-14
multipart-fix

### LINKED ISSUE #3175: multipart/form-data current values are deleted right before editing

1. Edit with 'e' a POST flow with multipart/form-data type.
image

2. An empty editor will appear:
image

3. Add a row, and go back to check that previous fields have sadly been overwritten:
image

image

It seems to me that the editor expects a URL-encoded form type, and thus the unexpected format passes unnoticed, while the new form is saved with the URL-encoded format. I'll wait for some insights from anyone which knows something about it. Otherwise, I'll take a look at it in the next days.

##### System information

Mitmproxy: 4.0.0 binary
Python:    3.6.3
Platform:  Linux-4.15.0-kali3-amd64-x86_64-with-debian-kali-rolling

rjt-gupta:
Fixes #3175


armudgal:
Hi, can you write tests for the same. Thanks

mitmproxy/net/http/request.py
```diff
@@ -468,7 +468,8 @@ def _get_multipart_form(self):
         return ()
 
     def _set_multipart_form(self, value):
-        raise NotImplementedError()
+        self.headers["content-type"] = "multipart/form-data"
+        self.content = mitmproxy.net.http.url.encode(value, self.get_text(strict=False)).encode()
```
mhils: This sets a multipart header but an urlencode value. They are not the same, so this won't fly! This is akin to taking a Word document and changing the filename to ".pdf". It won't make it a PDF document either!

mitmproxy/tools/console/grideditor/editors.py
```diff
@@ -54,16 +54,24 @@ def set_data(self, vals, flow):
 
 
 class RequestFormEditor(base.FocusEditor):
-    title = "Edit URL-encoded Form"
```
mhils: Minor: I'd like to have two seperate editors here, which both accept either format as input but only output the respective format. This makes it possible to quickly change the form type from within mitmproxy.

mhils:
Awesome, thanks! 🍰 🚀

rjt-gupta:
Super duper happy 😁 

> 
> 
> Awesome, thanks! 🍰 🚀

PR #3424, opened 2018-12-16
[RFC] Tabular flow list

BkPHcgQL3V:
screenshot

- Present the flow list as a table with one line per flow, instead of two or more rows
- Use a combination of fixed-width and variable-width columns to present flow properties
- Indicate field truncation appropriately using an ellipsis
- Colour each column individually according to its content
  - Request type: spot POST or other (OPTIONS, HEAD) requests easily
  - Protocol: spot insecure connections easily
  - Hostname: right-align to give priority to top-level domains instead of subdomains
  - Content-type: colour according to general category
  - Size and duration: use a gradient (with full colour range on high-colour terminals) to spot slow / large requests

Current version only includes palette colour entries for the Dark and SolarizedDark (default) palette.

### DIFF
--- a/mitmproxy/tools/console/common.py
+++ b/mitmproxy/tools/console/common.py
@@ -1,5 +1,8 @@
 import platform
 import typing
+import datetime
+import time
+import math
 from functools import lru_cache
 
 import urwid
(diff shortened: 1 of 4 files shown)


BkPHcgQL3V:
I added highlighting of TLDs and URL parts:

url-highlighting

Any thoughts before I clean up / push the code here?

mhils:
Hi! This is very very cool, thanks for the PR. This is a fantastic improvement for large mitmproxy windows. We do need to figure out what to do with smaller terminals though, here's what it looks like for 80 characters:

image

I feel that is a case we can hardly ignore and I'm afraid we do need a two line display in these cases. However, I also do like your table style a lot. Can we maybe add a `console_flowlist_layout = {auto, table, list}` option, where auto (the default) switches to your style for window widths > 120?

/cc @mitmproxy/devs: Please do chime in, this is nothing I want to make calls on alone. :)

BkPHcgQL3V:
Before I spend too much time on this (fixing narrow windows and other colour palettes), any feedback if this is going in a direction the project is interested in so far?

mhils:
I'm definitely in favor of the overall change, with a slight preference for keeping the old layout for narrow windows and having an option for it. :-) 

@cortesi @Kriechi @kajojify @madt1m any opinions? Speak now or forever hold your peace. :)

madt1m:
This looks a nice improvement on the layout! I agree with the changes, while adding my two cents:
- Having a default for narrow windows, and a switch option, seems the right choice to me;
- I'd include an option/keybinding, for each relevant field, to change alignment e.g. I can imagine some specific use cases where I'd prefer to have the domain left-aligned, focusing on subdomains.

BkPHcgQL3V:
> Keep the old (or switch to a new) two-row layout for narrow windows. Probably place the path part of the URL on its own row.

How's this?



The URL is on its own row in narrow views, to get as much space as possible.

> I'd include an option/keybinding, for each relevant field, to change alignment e.g. I can imagine some specific use cases where I'd prefer to have the domain left-aligned, focusing on subdomains.

The left/right arrow keys in the default view don't seem to be bound to anything...

Issue #3165, opened 2018-05-27
How to configure the external viewer?


gajus:
mitmproxy v4 has key binding `v` ("View flow body in an external viewer").

It appears to open the content of the response in nano. How do I change this to, e.g. Atom or browser?

kajojify:
@gajus Hello! 
Mitmproxy lets you:
- edit request/response body of the flow;
- edit the content of grideditor cell;
- view request/response of the flow.
 in the **external editor**.

Mitmproxy uses environment variables in order to choose appropriate external editor/viewer:
- for editing:
https://github.com/mitmproxy/mitmproxy/blob/69aa5a0d2f799cc91ab5a7d5d75c55aa40ae58b8/mitmproxy/tools/console/master.py#L123

- for viewing:
https://github.com/mitmproxy/mitmproxy/blob/69aa5a0d2f799cc91ab5a7d5d75c55aa40ae58b8/mitmproxy/tools/console/master.py#L162

So, as you can see, in order to change standard editor you should change environment variable PAGER or EDITOR from your terminal.

All the best!

gajus:
> So, as you can see, in order to change standard editor you should change environment variable PAGER or EDITOR from your terminal.

Is this documented anywhere?

I could not find this anywhere in the docs.

mhils:
That sounds reasonable, we can make it `os.environ.get("MITMPROXY_EDITOR") or os.environ.get("EDITOR") or "vi"`.

This should also be in the docs, not sure where a good place would be.

ffe4:
@mhils can this issue be closed after #3425 ? Is it only about adding docs now?

q2dg:
The existence of MITMPROXY_EDITOR variable isn't added in docs yet :-(

mhils:
Please feel free to add it!

PR #3425, opened 2018-12-16
external-viewer-fix

rjt-gupta:
fix for #3165 

Should this be added in docs?

### DIFF
--- a/mitmproxy/tools/console/consoleaddons.py
+++ b/mitmproxy/tools/console/consoleaddons.py
@@ -113,6 +113,11 @@ class ConsoleAddon:
             "console_mouse", bool, True,
             "Console mouse interaction."
         )
+        loader.add_option(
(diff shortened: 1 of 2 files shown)


mhils:
Looking at the issue again I feel that it may be more useful if we just add a regular mitmproxy console option (https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/tools/console/consoleaddons.py) for this (where we mention the use of $EDITOR in the option description). What do you think?

rjt-gupta:
Yeah I guess this is a better solution.

rjt-gupta:
In my opinion we should only mention MITMPROXY_EDITOR in options as it will set only mitmproxy specific editor. what do you think?

mhils:
Thanks again! 🍰

PR #3428, opened 2018-12-17
Memoise command parsing

BkPHcgQL3V:
https://github.com/mitmproxy/mitmproxy/pull/3427#issuecomment-447928549

Refactor `verify_arg_signature` and the code to lex a command string so that they can be memoised using `functools.lru_cache`, then do so.

### DIFF
--- a/mitmproxy/command.py
+++ b/mitmproxy/command.py
@@ -14,10 +14,11 @@ from mitmproxy import exceptions
 import mitmproxy.types
 
 
-def verify_arg_signature(f: typing.Callable, args: list, kwargs: dict) -> None:
+@functools.lru_cache(maxsize=200)
+def verify_arg_signature(f: typing.Callable, args: tuple, kwargs: tuple) -> None:
     sig = inspect.signature(f)
     try:
-        sig.bind(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


mitmproxy/command.py
```diff
@@ -271,7 +278,7 @@ def command(path):
     def decorator(function):
         @functools.wraps(function)
         def wrapper(*args, **kwargs):
-            verify_arg_signature(function, args, kwargs)
+            verify_arg_signature(function, args, tuple(kwargs.items()))
```
BkPHcgQL3V: Doesn't look like this is working, as some of the commands' arguments themselves are lists. Looks like they need to be flattened for `@lru_cache` to accept them. However, the values themselves aren't necessary, are they? Any way we can pass just the types?

mhils: This is really surprising as *args is consistently a tuple on my system. 🤷‍♂️ 
Anyways, fix attempt below.

BkPHcgQL3V: I admit I didn't look into it too closely, but I think some of the arguments may contain objects that in turn contain non-hashable types, like lists?

PR #3444, opened 2019-01-06
Fix ResourceWarning: unclosed file, prevent resource leaks

BoboTiG:
Signed-off-by: Mickaël Schoentgen <contact@tiger-222.fr>


mhils:
Thanks again! 😊

BoboTiG:
My pleasure :)

PR #3448, opened 2019-01-14
Fix command injection vulnerability when exporting to curl or httpie

cript0nauta:
The function producing a curl/httpie command from a request doesn't escape single quotes properly. An attacker could craft a malicious request that when exported to a httpie/curl command will also contain any command he/she desires. This changes fixes the vulnerability.

**UPDATE**: Here is a copy of the email I sent to the maintainers indicating how to exploit the vulnerability:

```
The problem is in the export to curl/httpie command. An attacker can craft a
request that when it is exported will produce not only a curl/httpie command,
but any other one he/she desires.

To replicate the vuln, execute the following curl command (this is what the
attacker would do)

    curl -x localhost:8080 httpbin.org/get -H "X-Inject: '; cat
/etc/passwd #"

Then, in the mitmproxy flow list, select this request and run the command :

    export.clip curl @focus

A curl command will be copied to the clipboard. If you paste it in a bash
shell, the contents of the /etc/passwd file will be printed, indicating the
presence of the vulnerability.


By looking at the mitmproxy/addons/export.py file, I saw you try to escape
certain request fields like the POST data with the
strutils.bytes_to_escaped_str. This function isn't safe to escape single-quoted
shell arguments since single quotes are escaped with \' but the backslash
doesn't work in single-quoted bash strings. So I can exploit this even in
fields that are escaped with the bytes_to_escaped_str function. For example,
you can make this request:

    curl -x localhost:8080 httpbin.org/post --data-binary "'& cat /etc/passwd #"

If you export it to curl inside mitmproxy and paste the command in a bash shell
as in the previous step, the contents of the passwd file will also be printed.
I used a & instead of a ; to run another command because otherwise the curl
command wouldn't end because the content-length is greater the the actual POST
payload (because part of it is injected as a shell command).


To fix this vulnerability you should escape all request fields with the
shlex.quote python function instead of strutils.bytes_to_escaped_str. I can
make a PR you want to.


Greetings!

```


mitmproxy/addons/export.py
```diff
@@ -18,36 +19,42 @@ def raise_if_missing_request(f: flow.Flow) -> None:
...
-            request.content,
-            escape_single_quotes=True
-        )
-    return data
+        try:
+            content = strutils.always_str(request.content)
+        except UnicodeDecodeError:
+            # shlex.quote doesn't support a bytes object
+            # see https://github.com/python/cpython/pull/10871
+            raise exceptions.CommandError("Request content must be valid unicode")
```
mhils: I'm not really happy with this construct here - the idea of `strutils.bytes_to_escaped_str` is that it gives you something you can paste, even if it's binary. For example, if the request body is a null byte, you get a string with `\x00` back.

 How would you feel about `shlex.quote(strutils.bytes_to_escaped_str(request.content))`?

mhils: Great find on the Bash feature! 
If I understand things correctly, we should `shlex.quote` all other parameters and then append binary content:
```python
command = ' '.join(shlex.quote(arg) for arg in args)
if request.content:
    command += (
        "--data-binary $'" + 
        strutils.bytes_to_escaped_str(request.content, escape_single_quotes=True) + 
        "'"
    )
return command
```

The problem with shlex.quote is that it would break the $ env:
```python
>>> print(shlex.quote(r"foo\55'\55"))
'foo\55'"'"'\55'
```

Makes sense?

(and we definitely need tests here)

cript0nauta: I'm not sure of how to proceed to support this. Calling `shlex.quote` on the output of `bytes_to_escaped_str` and prepending a `$` doesn't work properly based on the example you posted. Removing the `shlex.quote` call makes it work and it properly escapes single quotes, but that doesn't sound safe to me. We could also copy the code done in [this PR](https://github.com/python/cpython/pull/10871/files) that makes `shlex.quote` support bytes objects, but I don't know if that has been properly tested.

Another thing I noted is that none of this techniques will work properly if the body has a null byte, since the body is being passed as an argument, and [command arguments can't contain null bytes](https://eklitzke.org/unexpected-places-you-can-and-cant-use-null-bytes).

I think we should have a solution that is secure and that works in most cases. For the rest of the cases (like bodies that contain null bytes or that cannot be decoded to unicode) you can always use mitmproxy to replay requests. What do you think?

mhils: > I think we should have a solution that is secure and that works in most cases. For the rest of the cases (like bodies that contain null bytes or that cannot be decoded to unicode) you can always use mitmproxy to replay requests. What do you think?

Fair enough - it's probably not a bad idea to not rely on  `bytes_to_escaped_str` for security.

mitmproxy/addons/export.py
```diff
@@ -18,36 +19,42 @@ def raise_if_missing_request(f: flow.Flow) -> None:
...
+        args += ["-X", request.method]
+    args.append(request.url)
     if request.content:
-        data += " --data-binary '%s'" % strutils.bytes_to_escaped_str(
-            request.content,
-            escape_single_quotes=True
-        )
-    return data
+        try:
+            content = strutils.always_str(request.content)
```
mhils: We should use `request.get_text()` here (which may raise a ValueError)

cript0nauta: Isn't that done by the `request.decode(strict=False)` line from above? It decodes the body and removes the `content-encoding` header

cript0nauta:
Hi! Can this PR be merged, or is there anything else to be done? This fixes a security vulnerability so it would be good to merge this soon.

cript0nauta:
Hi! Since I haven't received any response about the vuln since January, I'm publishing the details of how this vulnerability can be exploited by an attacker.

As a proud user of mitmproxy, I hope this can be fixed soon.

mhils:
Thank you again for disclosing this responsible @cript0nauta! I'm super sorry that it took us that long, but we'll finally have the fix merged and a new release hopefully soon 😃 

Thanks! 🍰 ❤️

PR #3451, opened 2019-01-18
openbsd: divert-to requires exact listen address

yjh0502:
divert-to does not work with '0.0.0.0' or similar listen address, so we
need to specify listen address that we provided to `pf`.


mhils:
Thanks! 🍰

PR #3453, opened 2019-01-23
fixes copying on macos

### LINKED ISSUE #2910: System copying shortcuts don't work with mitmweb

1. `mitmweb` - web browser opens
2. highlight any text on the page
3. use Cmd+C system shortcut (standard for copying text on macOS)

Result: nothing is copied - clipboard remains as-is.

Expected: highlighted text is copied to clipboard.

Right clicking and hitting copy works, but this is still a massive usability issue.

This has been an issue since at least v2.0.2 - but maybe earlier, and is still a problem on the latest version.

I have had this problem for about a year, using multiple different Macs, so I don't think this is something unique to my current setup.

Have searched, but there don't seem to be any issues mentioning this, surprisingly.

##### System information

```
Mitmproxy: 3.0.2 
Python:    3.6.4
OpenSSL:   OpenSSL 1.0.2n  7 Dec 2017
Platform:  Darwin-17.4.0-x86_64-i386-64bit
```

XZzYassin:
I tested `e.metaKey` and it returns `true` when the command key is pressed.

resolves #2910

### DIFF
--- a/web/src/js/ducks/ui/keyboard.js
+++ b/web/src/js/ducks/ui/keyboard.js
@@ -6,7 +6,7 @@ import * as modalActions from "./modal"
 
 export function onKeyDown(e) {
     //console.debug("onKeyDown", e)
-    if (e.ctrlKey) {
+    if (e.ctrlKey || e.metaKey) {
         return () => {
         }
     }


mhils:
Thanks! 😃 🍰

PR #3456, opened 2019-01-25
Fix: support wsproto 0.13

RPigott:
wsproto changed the location of WSConnection in 0.13


RPigott:
Ah that's too bad, I made this change on my machine and things seemed to work. Woulda been nice if it was that simple. Sorry for the hassle.

PR #3462, opened 2019-02-02
add MASQUERADE target for iptables

jgiglio:
A MASQUERADE line is necessary for this to work.  I spent quite a while trying to figure out what was missing from the howto to make it work.  It looks to be safe to not specify any target interfaces or ips, etc, the host running mitmproxy can still be used for local traffic fine.


mhils:
Thanks for the PR. Depending on your setup you actually don't want to MASQUERADE, so we prefer to keep these rules minimal.

PR #3465, opened 2019-02-06
Fix: wrong import name from wsproto.connection

ihsanturk:


mhils:
That's a problem with your wsproto version.

Issue #3469, opened 2019-02-07
Errors when trying to use server replay on the flows created with help of mitmproxy


kajojify:
##### Steps to reproduce the problem:

1. Run mitmproxy.
2. Press `n` -> `enter` -> `r`
3. Press `:` -> input "replay.server @focus" -> `enter` -> `r`
![image](https://user-images.githubusercontent.com/20267977/52443176-b7484400-2b2d-11e9-8c8f-dfcae3b6f898.png)

4. Create second flow using "n -> enter -> r" sequence.
5. Press `:` -> input `replay.server @all` -> `enter` -> replay any flow.
![image](https://user-images.githubusercontent.com/20267977/52443359-26be3380-2b2e-11e9-8521-87bb7934d783.png)

##### System information

Mitmproxy: 5.0.0.dev (+223, commit 139a385)
Python:    3.6.7
OpenSSL:   OpenSSL 1.0.2g  1 Mar 2016
Platform:  Linux-4.4.0-141-generic-x86_64-with-Ubuntu-16.04-xenial

PR #3482, opened 2019-02-18
fixes server-replay

### LINKED ISSUE #3167: Server replay does not cache POST requests
Using mitmproxy v2 and server replay I was able to server replay all requests, including POST.

It appears that as of version v4 POST requests are not being replayed.

It is unclear having read the entire documentation what is the reason POST requests are not being cached.

rjt-gupta:
Fixes #3167 and #3469 

Is there a better way to do this?


mhils:
This is fixed a bit differently in #3705 now. Thanks again Rajat! 😃

rjt-gupta:
> 
> 
> This is fixed a bit differently in #3705 now. Thanks again Rajat! 😃

Awesome! 😀

PR #3486, opened 2019-02-23
filter unicode fix

### LINKED ISSUE #2951: Cannot filter by UTF-8 string

1. run `mitmproxy`
2. type (in UTF-8) `: set view_filter=~bs Троцкий`
3. “Invalid interception filter: ~bs Троцкий”
4. type `: set view_filter=~bs trotskij`
5. the filter is set and correctly limits the view to flows where response body contains “trotskij”

mitmproxy should make it possible to filter by arbitrary UTF-8 strings.

I have `LANG=en_US.UTF-8`; setting `LC_ALL=en_US.UTF-8` doesn’t help.

These don’t work, either:

- `: set view_filter=~bs "Троцкий"` (the quotes are automatically removed after typing the second one)
- `: set "view_filter=~bs Троцкий"`
- `: set view_filter=~bs /Троцкий/`

##### System information

Mitmproxy: 3.0.3 binary
Python:    3.5.2
Platform:  Linux-4.4.0-116-generic-x86_64-with-debian-stretch-sid

rjt-gupta:
Fixes #2951 

Hi! Initially I thought this should be easy to fix but struggled due to wrong approach. Later I found that `pyparsing`  included several unicode character ranges that can help to set non-ascii filters.

 Would be Glad to add if I missed something.

### DIFF
--- a/mitmproxy/flowfilter.py
+++ b/mitmproxy/flowfilter.py
@@ -475,7 +475,30 @@ def _make():
         parts.append(f)
 
     simplerex = "".join(c for c in pp.printables if c not in "()~'\"")
+    alphdevanagari = pp.pyparsing_unicode.Devanagari.alphas
+    alphcyrillic = pp.pyparsing_unicode.Cyrillic.alphas
+    alphgreek = pp.pyparsing_unicode.Greek.alphas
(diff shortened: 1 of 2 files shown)


Kriechi:
Thanks! 🎉 👍

PR #3516, opened 2019-04-09
Remove the confdir argument

pierlon:
This removes the deprecated `--confdir` argument from the `--help` output, and no longer accepts the value even if specified.


mhils:
Awesome, thanks!

PR #3525, opened 2019-04-17
Fix #3502 by using email.utils.formatdate's usegmt argument.

### LINKED ISSUE #3502: Server replay corrupts Date header

1. Start `mitmproxy`.
2. Run a request through it and observe that the `Date` header of the response is correct:

       $ curl -six http://localhost:8080 http://httpbin.org/status/200 | grep Date:
       Date: Thu, 21 Mar 2019 12:45:00 GMT

3. In mitmproxy, start server replay:

       : replay.server @all
       
4. Run a request again and observe that the `Date` header now has `-0000` instead of `GMT`:

       $ curl -six http://localhost:8080 http://httpbin.org/status/200 | grep Date:
       Date: Thu, 21 Mar 2019 12:46:58 -0000

Per [RFC 7231 § 7.1.1](https://tools.ietf.org/html/rfc7231#section-7.1.1), a correct HTTP date always has `GMT`.

##### System information

    Mitmproxy: 4.0.4 binary
    Python:    3.6.3
    Platform:  Linux-4.15.0-46-generic-x86_64-with-debian-buster-sid

tonyb486:
This fixes issue #3502 by using the usegmt argument for email.utils.formatdate.  

https://docs.python.org/3/library/email.utils.html#email.utils.formatdate
https://docs.python.org/2/library/email.utils.html#email.utils.formatdate

> Optional usegmt is a flag that when True, outputs a date string with the timezone as an ascii string GMT, rather than a numeric -0000. This is needed for some protocols (such as HTTP). This only applies when localtime is False. The default is False.

### DIFF
--- a/mitmproxy/net/http/cookies.py
+++ b/mitmproxy/net/http/cookies.py
@@ -304,7 +304,7 @@ def refresh_set_cookie_header(c: str, delta: int) -> str:
             e = email.utils.parsedate_tz(attrs["expires"])
             if e:
                 f = email.utils.mktime_tz(e) + delta
-                attrs.set_all("expires", [email.utils.formatdate(f)])
+                attrs.set_all("expires", [email.utils.formatdate(f, usegmt=True)])
             else:
                 # This can happen when the expires tag is invalid.
                 # reddit.com sends a an expires tag like this: "Thu, 31 Dec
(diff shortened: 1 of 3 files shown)


mhils:
Awesome, thanks! 🍰 🎉

PR #3526, opened 2019-04-19
Add --allow_hosts option

pierlon:
This PR adds the `--allow_hosts` option, opposite to the `--ignore-hosts` option.


stevennyman:
A command flag like this one would be really helpful for my use case. It would be great to see this merged and released!

Kriechi:
Thanks! 🎉 👍

PR #3547, opened 2019-05-27
Use Brotli instead of brotlipy

felixonmars:
brotlipy is stuck at brotli 0.6 and upstream is inactive. Let's switch
to the official binding which is up-to-date and has same interfaces.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,7 +62,7 @@ setup(
     # It is not considered best practice to use install_requires to pin dependencies to specific versions.
     install_requires=[
         "blinker>=1.4, <1.5",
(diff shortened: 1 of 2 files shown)


mhils:
Thanks Felix!

@Lukasa: Any thoughts from your end?

Lukasa:
The only note is that you’ll suffer a bit with PyPy because the official binding uses the C API instead of cffi. It’s up to you how big a deal that is.

mhils:
We probably care about recent Python wheels more than about PyPy. Thanks again @Lukasa! 😃

Issue #3578, opened 2019-07-06
exported curl command is not properly quoted when payload has single quote


contrun:
##### Steps to reproduce the problem:

1. `http_proxy=http://127.0.0.1:8090 curl --data-binary "'test'=string" example.com`
2. Export flow as curl command
3. Run the exported curl command `curl -H 'Host:example.com' -H 'User-Agent:curl/7.64.0' -H 'Accept:*/*' -H 'Proxy-Connection:Keep-Alive' -H 'Content-Length:13' -H 'Content-Type:application/x-www-form-urlencoded' -X POST 'http://example.com/' --data-binary '\'test\'=string'`

##### Any other comments? What have you tried so far?
[shellcheck wiki SC1003](https://github.com/koalaman/shellcheck/wiki/SC1003) has more information above how to escape single quote inside single quote.

I copied its description to here.

Problematic code:
```
echo 'This is not how it\'s done'.
```
Correct code:
```
echo 'This is how it'\''s done'.
```

##### System information

```
Mitmproxy: 4.0.4
Python:    3.7.3
OpenSSL:   OpenSSL 1.0.2r  26 Feb 2019
Platform:  Linux-4.19.49-x86_64-with-glibc2.2.5
```

mhils:
Thanks for the report. This seems to be a duplicate of #3448.

PR #3579, opened 2019-07-06
use shlex to escape single quote

contrun:
This should fix https://github.com/mitmproxy/mitmproxy/issues/3578


user185953:
#3448 is the real issue link

mhils:
Thanks! This has been fixed by #3448.

Issue #3567, opened 2019-06-15
websockets: replace simple Masker with speed-optimized version


Kriechi:
see the recent discussion in https://github.com/python-hyper/wsproto/issues/127

Summary: Our ["simple" Masker](https://github.com/mitmproxy/mitmproxy/blob/8353f4a55afeec9d30727d91d642e8b8af4040f8/mitmproxy/net/websockets/masker.py) is not really performant. With a little code magic we can gain speedup of up to 60x!

MathieuBordere:
Hi, I'll pick this one up as my first contribution to mitmproxy

typoon:
I assume this can be closed?

PR #3580, opened 2019-07-06
websockets: replace masker with more optimized one

MathieuBordere:
fix for #3567


MathieuBordere:
Obtained a 40x speed up compared to the previous implementation of `mask`

mhils:
Awesome, thank you! 🎉 🍰

PR #3585, opened 2019-07-17
fix #3584

### LINKED ISSUE #3584: Addons: `running` is invoked twice.

1. Write an addon that implements `running()`
2. Running is invoked twice on startup.

mhils:


mhils:
This hell is much trickier and fragile than expected, I'll leave this to @cortesi.

PR #3590, opened 2019-07-18
fix #3584

### LINKED ISSUE #3584: Addons: `running` is invoked twice.

mhils:
This is a hell of invariants, but 100 monkeys on typewriters found a solution.


mhils:
Also doesn't work. 🙃

PR #3593, opened 2019-07-21
[RFC] Tabular flow list [Updated] 

Jessonsotoventura:
It looked like PR #3424 is no longer being updated. This PR adds the following on top of the other PR:
 
- Color schemes for lowlight, lowdark, solarized light, solarized dark, dark and light
- Colors for each request type
- The console_flowlist_type settings, which allows switching between default, list, and table view
   + Default auto switches between the two views if the cols are < 80
   + If the cols are <80 and a table view has been selected a two lined view is presented
   + list view is the current view
dark
light
lowdark
lowlight
options
solarized_dark
solarized_light


mhils:
Thanks again for picking up this fantastic PR and sorry for letting you wait. This looks like a really good improvement of the flow table. Some notes:

1. It looks like we have some issues regarding caching. [Here](https://uploads.hi.ls/2019-09/2019-09-07_16-45-35.mp4)'s a video that demonstrates the issue in the time column.
2. The FQDN highlighting is a bad security UX. Being used by "expert users" isn't a good excuse. :) We should (1) highlight the extension as well and (2) ideally use the public suffix list, e.g. by including https://pypi.org/project/publicsuffix2/. I would consider dropping `url_punctuation` and just having a highlight and a non-highlighted domain part.

Other that that, I'm quite strongly +1 to get this into the next release and on by default. @cortesi, any notes from you on this?

BkPHcgQL3V:
> It looked like PR #3424 is no longer being updated.

Thanks for picking it up! I didn't update it as nobody answered my questions regarding the design.

> It looks like we have some issues regarding caching. [Here](https://uploads.hi.ls/2019-09/2019-09-07_16-45-35.mp4)'s a video that demonstrates the issue in the time column.

Looks like these synthetic requests don't have a `timestamp_start`, so `human.format_timestamp` passes the `None` to `time.localtime`, which substitutes the current time instead of `None`. Here is a fix:

```diff
diff --git a/mitmproxy/tools/console/common.py b/mitmproxy/tools/console/common.py
index 49f5c247..a7e583fe 100644
--- a/mitmproxy/tools/console/common.py
+++ b/mitmproxy/tools/console/common.py
@@ -304,6 +304,8 @@ def raw_format_flow(f):
         s = human.format_timestamp(f["req_timestamp"])
     else:
         s = datetime.datetime.fromtimestamp(time.mktime(time.localtime(f["req_timestamp"]))).strftime("%H:%M:%S")
+    if f["req_timestamp"] is None:
+        s = ' ' * len(s)  # reuse formatted length
     req.append(fcol(s, uc))
 
     methods = {
```

>  The FQDN highlighting is a bad security UX

What makes you say that, especially in this context? I see that e.g. Microsoft Edge highlights the non-subdomain part of the domain in its URL bar.

> ideally use the public suffix list, e.g. by including https://pypi.org/project/publicsuffix2/.

There is actually a TODO in the code to do exactly this, but I didn't know if the additional dependency would be appreciated.

mitmproxy/tools/console/common.py
```diff
@@ -415,14 +501,15 @@ def raw_format_flow(f):
...
-def format_flow(f, focus, extended=False, hostheader=False, cols=False):
+def format_flow(f, focus, extended=False, hostheader=False, cols=False, layout='default'):
     acked = False
     if f.reply and f.reply.state == "committed":
         acked = True
     d = dict(
         focus=focus,
         extended=extended,
-        two_line=extended or cols < 100,
+        two_line=extended or cols < 80,
```
BkPHcgQL3V: I'm not sure 80 characters is enough to fit the content in the tabular view nicely.

image

mitmproxy/tools/console/common.py
```diff
@@ -266,9 +266,95 @@ def colorize_url(url):
         ('url_punctuation', 3),  # ://
     ] + colorize_host(parts[2]) + colorize_req('/' + parts[3])
 
+@lru_cache(maxsize=800)
+def raw_format_list(f):
```
BkPHcgQL3V: What is the rationale for adding this? Why not make the option use the two-row layout from the tabular view always?

mitmproxy/tools/console/common.py
Jessonsotoventura: I agree 80 is too low upped the value back to 100.

Jessonsotoventura:
@mhils I've pushed the changes you requested. Let me know if something seems off or a change is needed.

mhils:
Awesome, thank you! I incidentally talked about this PR with @cortesi
yesterday, who is also strongly +1. On mobile on vacation, will merge once
I get access to my laptop! 😃

mhils:
🎉🎉🎉

@BkPHcgQL3V and @Jessonsotoventura, thank you so much for pushing this. The next mitmproxy release will have a much nicer interface by default. There are also fantastic avenues for further improvements, such as a setting on which columns to display (shared for both mitmweb and mitmproxy). Looking forward to having this out! 😃

Issue #3288, opened 2018-08-12
UX Regression: Updating the URL is painful.


mhils:
##### Steps to reproduce the problem:

1. Select a flow, press `e` to edit a URL.
2. `flow.set @focus url str` appears

This is really impractical for editing URLs. I see two approaches here:

1. A popup appears where the existing URL can be updated, similar to Query editing.
2. `flow.set @focus url https://existing-url/` appears.

The same problem may also apply when editing other attributes, we should take a look.

##### System information

Mitmproxy: 5.0.0.dev (+137, commit e74e97b)
Python:    3.6.3
OpenSSL:   OpenSSL 1.1.0g  2 Nov 2017
Platform:  Linux-4.4.0-17134-Microsoft-x86_64-with-Ubuntu-16.04-xenial

elijahbal:
#3289 seems to be referencing the same issue.

Jessonsotoventura:
I don't think the other attributes have this issue. It appears that other options, which could be long strings are first shown inside of an editing viewing. These views allow the contents of a cell to be opened in an external editor. 

The only long attribute which can not be opened in the external editor was URL, but #3594 fixes this issues.

I don't think there is a reason to have status, method, or reason spawn an external editor or an edit view, but it might be a nice cosmetic feature just to keep constancy across all the edit options.

Thoughts?

PR #3594, opened 2019-07-23
Edit URLs in external editor

Jessonsotoventura:
Fixes: #3288 

Since URLs can be long and unwieldy they should be edited in an external editor instead of in the action bar.


Kriechi:
Thanks! 🎉 👍

PR #3597, opened 2019-07-24
Update howto-transparent-vms.md for newer versions

Nirusu:
Newer versions of Ubuntu have more dependencies on systemd. The names of the interfaces changed and systemd-resolved provides a stub DNS provider that interferes with dnsmasq. This commit/PR makes some adjustments in order to provide instructions for newer Ubuntu versions.

### DIFF
--- a/docs/src/content/howto-transparent-vms.md
+++ b/docs/src/content/howto-transparent-vms.md
@@ -14,9 +14,13 @@ Internal Network* setup can be applied to other setups.
 
 ## 1. Configure Proxy VM
 
-On the proxy machine, **eth0** or **enp0s3** (Ubuntu 15.10 and newer) is connected to the internet. **eth1**  or **enp0s8** (Ubuntu 15.10 and newer) is
-connected to the internal network that will be proxified and configured
-to use a static ip (192.168.3.1).
+First, we have to find out under which name Ubuntu has mapped our network interfaces. You can find this information with:
+
+{{< highlight bash  >}}
+ip link
+{{< / highlight >}}
+
+Usually with Ubuntu and Virtualbox, **eth0** or **enp0s3** (Ubuntu 15.10 and newer) is connected to the internet and **eth1**  or **enp0s8** (Ubuntu 15.10 and newer) is connected to the internal network that will be proxified and configured to use a static ip (192.168.3.1). If the names differ, use the ones you got from the *ip link* command.
 
 ### VirtualBox configuration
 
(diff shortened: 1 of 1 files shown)


docs/src/content/howto-transparent-vms.md
```diff
@@ -14,7 +14,7 @@ Internal Network* setup can be applied to other setups.
 
 ## 1. Configure Proxy VM
 
-On the proxy machine, **eth0** is connected to the internet. **eth1** is
+On the proxy machine, **eth0** or **enp0s3** (Ubuntu 15.10 and newer) is connected to the internet. **eth1**  or **enp0s8** (Ubuntu 15.10 and newer) is
```
Kriechi: The interface names are pretty much random at this point - and highly depend on your hardware.
`eth0` should be considered a placeholder here, so the user has to verify which interface name it really has.
Would you mind re-writing the sentence to make the user "find" the correct interface?

docs/src/content/howto-transparent-vms.md
```diff
@@ -93,10 +94,11 @@ IP address via DHCP:
 
 ## 3. Redirect traffic to mitmproxy
 
-To redirect traffic to mitmproxy, we need to add two iptables
+To redirect traffic to mitmproxy, we need to enable IP forwarding and add two iptables
 rules:
 
 {{< highlight bash  >}}
+sysctl -w net.ipv4.ip_forward=1
```
Kriechi: Since we use `sudo ...` below, we should use it here as well.

docs/src/content/howto-transparent-vms.md
Nirusu: Saying that the interface names are "pretty much random at this point" is not quite true. Actually, systemd calls them [Predictable Network Interface Names](https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/) and in this case, they're assigned by their "physical location".

Now, considering this tutorial covers VirtualBox, I'm fairly certain that enp0s3 maps to your first network interface and enp0s8 to the second one. Maybe one can create a different assignment by fiddling around with the virtual machine config? Might be the case, but I haven't seen this before. [And some other docus also cover it this way.](https://bertvv.github.io/notes-to-self/2015/09/29/virtualbox-networking-an-overview/).

Kriechi: Yes, some systems might have predictable names, but IMHO there are too many systems / managers / OS / Linux distributions to have a "one size fits all" documentation.
I would not rely on any interface name that is written in the docs - the user should always double-check what their own system is using - especially in multi-homed servers / appliances.

Nirusu: But the tutorial is specified to Ubuntu and VirtualBox, but yeah... I've updated the text to look for the correct names via `ip link`, might wanna take a look on the new version?

Kriechi:
Thanks! 🎉 👍

PR #3603, opened 2019-08-01
Transparent proxy on Linux for local originating traffic

tomsaleeba:
I struggled with getting this working and found a few posts around the internet from people with the same issue. The macOS section talks about this so I've just "ported" that to the linux world.

Thanks for the fantastic tool BTW :tada:


Kriechi:
Thanks! 👍 🎉

PR #3626, opened 2019-09-03
Http2 full support

Josue-T:
# Add full support for HTTP/2.

**Main work on this PR is:**

- Create some new class to represent the HTTP/2 frames
- Send all HTTP/2 frame in the channel and add the possibility to modify the frame by a plugin.
- Make the  `View` more generic to be able to manage also the HTTP/2 frame
- Add a interface in the console to show the list of all HTTP/2 frames

Here is the document of this project. You should more understand how this work happens and how the new architecture was made.

[Document_v1.0.pdf](https://github.com/mitmproxy/mitmproxy/files/3579167/Document_v1.0.pdf)


Josue-T:
Hello,

mhils:
Hi @Josue-T,

Thanks for the awesome massive PR and congratulations on your Bachelor thesis! 🍰 🚀 Time to dust off my French I guess. 😅 I'm very impressed with what you have built here, you should definitely consider applying for GSoC with us next summer. :) 

I'm really looking forward to play around with this. This of course is a _massive_ amount of new code (and while quality looks good, possible liabilities) and we need to figure out what's in the best interest of mitmproxy and our users. @cortesi, @Kriechi, @kajojify and I are super busy at the moment, so this may take a few weeks before some of us meet in person. In the meantime, please do reach out on Slack, we'd be super interested to hear from you how you perceived the codebase, what we could improve, etc!

Best,
Max

mitmproxy/http2.py
```diff
@@ -0,0 +1,946 @@
...
+            elif not end_stream:
+                for event in self._events:
+                    if hasattr(event, "stream_ended"):
+                        event.stream_ended = None
+                    elif isinstance(event, h2.events.StreamEnded):
+                        self._events.remove(event)
+        self._end_stream = end_stream
+
+
+class _PriorityFrame():
```
Kriechi: Instead of re-implementing all these frame classes, would it be possible to reuse the class hierarchy from our hyperframe dependency? https://github.com/python-hyper/hyperframe/blob/82abce835659f46d3944c65dce62d584152d4aab/hyperframe/frame.py

mitmproxy/tools/console/defaultkeys.py
```diff
@@ -7,12 +7,13 @@ def map(km):
     km.add("K", "console.view.keybindings", ["global"], "View key bindings")
     km.add("O", "console.view.options", ["global"], "View options")
     km.add("E", "console.view.eventlog", ["global"], "View event log")
+    km.add("2", "console.view.http2", ["global"], "View the HTTP/2 flow")
```
Kriechi: I'm not sure if "flow" is an intuitive description here? From what I understand, `consolve.view.http2` displays exactly one HTTP/2 connection (from a single TCP socket) and shows all frames.
If this is correct, we should just tell the user that it is a "connection", not an abstract-sounding "flow".

mhils:
This is really impressive work, but I'm sorry to say that I'm going to close this PR. I've tried to review this a couple of times now, but the changeset is too massive to be tamed. There is a large number of conceptual decisions in here we don't necessarily agree with, plus a lot of additional maintenance burden. I realize that the short timeframe of a bachelor thesis didn't give you the opportunity to wait for feedback from us earlier in the process, so sorry for that. If you are interested in getting some of your changes into mitmproxy we'd be more than happy to get this going, it just needs to happen in smaller steps.

Best,
Max

PR #3627, opened 2019-09-03
Updated har_dump,py timings dictionary function

RamiBerm:
the HAR file spec (http://www.softwareishard.com/blog/har-12-spec/#timings) states that timings that do not apply for a certain requests should be set to -1, this example may set -1000 as a timings value for certain requests.
This ends up producing invalid HAR files in many cases.

My proposed fix is to assign -1 into the dic and only multiply by 1000 for other values


mhils:
Thanks! 🍰

PR #3630, opened 2019-09-05
Add support for zstd Content-Encoding

tsaaristo:
```
Handles zstandard-compressed bodies labeled as zstd.
```

I doubt this is all that common, but I did run into zstd in the Twitter Android app and adding support for it was trivial enough.

### DIFF
--- a/mitmproxy/addons/core.py
+++ b/mitmproxy/addons/core.py
@@ -289,7 +289,7 @@ class Core:
         """
             The possible values for an encoding specification.
         """
-        return ["gzip", "deflate", "br"]
+        return ["gzip", "deflate", "br", "zstd"]
 
     @command.command("options.load")
     def options_load(self, path: mitmproxy.types.Path) -> None:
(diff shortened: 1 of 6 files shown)


mhils:
Thanks, this looks excellent! 🍰

Mikanoshi:
If only this could support zstd dictionaries, currently it fails do decompress such data.
For example ICQ New desktop client compresses requests and responses using 2 different dictionaries:
```
user-agent:		ICQ Desktop 000000000 ic1nmMjqg7Yu-0hL 10.0.0(40952) Windows_10 PC
accept:			*/*
accept-encoding:	zstd, gzip
im-zstd-request-dict:	im-zstd-dict-desktop-request-250919.zdict
im-zstd-response-dict:	im-zstd-dict-desktop-response-250919.zdict
content-encoding:	zstd
content-type:		application/json;charset=utf-8
content-length:		120
```

PR #3646, opened 2019-09-24
Update concepts-certificates.md

ahopkins:
Update to direct people to using Safari instead of another browser on iOS.


Kriechi:
Thanks! 🎉 👍

Issue #3625, opened 2019-08-30
Unknown setting ssl_insecure


FranklinYu:
##### Steps to reproduce the problem:

1. `docker run --rm -it -p 8080:8080 -p 127.0.0.1:8081:8081 mitmproxy/mitmproxy mitmweb --web-iface 0.0.0.0`
2. Navigate to http://localhost:8081
3. Switch to tab “Options”
4. Check “Verify server certificates”

```text
WARNING:tornado.general:400 PUT /settings?_xsrf=... (x.x.x.x): Unknown setting ssl_insecure
```

##### Any other comments? What have you tried so far?

I tried both Windows (with Docker-for-Windows) and Arch Linux.

##### System information

```text
Mitmproxy: 4.0.4
Python:    3.6.3
OpenSSL:   OpenSSL 1.0.2o  27 Mar 2018
Platform:  Linux-4.9.184-linuxkit-x86_64-with
```

Eluch:
Hi! I ran into the same problem. I fixed it and sent a PR.

If you want to check that box anyway (until PR accept / release), you should press the "Edit Options" button to the left, search for the ssl_insecure row, and enable it from there. (This will also check the "uncheckable" checkbox because those settings are the same)
Also I should note that it makes a very similar request, but to the "/options" endpoint.

If I would guess one of the endpoints will be obsolate in the future. :thinking: 

P.S.: High-Five for using Arch :smile:

Kriechi:
Fixed by #3648.

PR #3648, opened 2019-09-27
ssl_insecure added to option_whitelist because it was missing

Eluch:
I am using mitmweb and on clicking to "Verify server certificates"
image gave me an error like this:
image (Unknown setting ssl_insecure)
I investigated the problem and the "ssl_insecure" option was missing from the "option_whitelist".

The same setting could be toggled from the option popup:
image

Now it works from both places.

This should fix the following issue: #3625


Kriechi:
Thanks! 🎉 👍

PR #3657, opened 2019-09-29
Add key_size option to define rsa key size

vin01:
While using `mitmproxy` on a remote server I noticed that it uses the pre-defined `2048` bit key size for CA and also interception certificates.

A higher bit size would be better especially when testing with remote instances. /paranoia

I modified this manually for my use case but thought it would make sense to have this as an option as well.  Please review.


Kriechi:
Thanks! 🎉 👍

PR #3667, opened 2019-10-13
hugo server -D fails if html file in docs/src/generated are not present

jannst:
Wanted to build docs according to quick start and noticed that you have to run ./build-current script in order to start hugo


mhils:
Thanks for the PR! 🍰

PR #3676, opened 2019-10-25
curl command export has several problems:

ylmrx:
- authority can usually rely on actual URL. as `:authority` headers will
break curl command. (advise if it's better to change them to Host, or if
it should be reported on curl side)
- `content-length: 0` is added for each request. if it's found in the
curl argument list, it'll try to fetch an empty body (and crash).
- also trying to guess on `accept-encoding` header to add the
`--compress` option when fetching potentially compressed content.


mhils:
Thanks for the PR! 😃 🍰 
Very happy to merge this in, but could you please add tests so that we don't get any regressions? Our CI system is complaining about missing coverage: https://travis-ci.org/mitmproxy/mitmproxy/jobs/603873731#L276

PR #3679, opened 2019-10-28
Added support for IPv6 in pf.py for macOS

tomlabaude:
Added support to parse IPv6 entries in `sudo pfctl -s state`

Note that the part considering "freebsd" has not been copied from IPv4 as not being able to test it.


mhils:
Thanks! Could you please add a test in https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/platform/test_pf.py? 😃

tomlabaude:
> Thanks! Could you please add a test in https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/platform/test_pf.py? 😃

Oh tests !
I'm new to all those github/python/dev environment/tests/lint, but motivated to learn more to contribute.

Can you tell me how I can test this test? (haha)

Like your example you gave in docs `pytest --cov mitmproxy.addons.anticache --cov-report term-missing --looponfail test_anticache.py` but specifically for test_pf.py ?

mhils:
If you want to run the full test suite, just invoking `tox` should be good. :)
If you want to run test_pf.py only (which would be much faster), it works similar to the command above:
```
$ cd mitmproxy
$ pip install -e .[dev]  # install pytest
$ cd test/mitmproxy/platform
$ pytest --cov mitmproxy.platform.pf --cov-report term-missing test_pf.py
```

For the linter, you can run `tox -e lint` to see all errors before CI.

tomlabaude:
Well I thought I did everything well with two new commits, but a 3rd one "Merge branch" appeared due to my-first-steps-in-git-world... (And I don't understand as the merge commit seems to include the two previous ones ?)

Tell me if something's wrong.

mhils:
Awesome, thanks @tomlabaude! 😃

PR #3692, opened 2019-11-12
Update cryptography, enable TLS 1.3

mhils:
This works locally, but we're waiting for a cryptography release with `SSL._lib.SSL_clear_mode` exposed.


mhils:
The upstream cryptography PR is at https://github.com/pyca/cryptography/pull/5062.

test/mitmproxy/net/test_tcp.py
```diff
@@ -421,16 +421,18 @@ class TestServerCipherList(tservers.ServerTestBase):
...
+    @pytest.mark.xfail
     def test_echo(self):
+        # Not working for OpenSSL 1.1.1, see
+        # https://github.com/pyca/pyopenssl/blob/fc802df5c10f0d1cd9749c94887d652fa26db6fb/src/OpenSSL/SSL.py#L1192-L1196
         c = tcp.TCPClient(("127.0.0.1", self.port))
         with c.connect():
             c.convert_to_tls(sni="foo.com")
-            expected = b"['AES256-GCM-SHA384']"
-            assert c.rfile.read(len(expected) + 2) == expected
+            expected = b"['TLS_AES_256_GCM_SHA384']"
```
Kriechi: If the cipher name changes, shouldn't we also increase our minimum version dependency?
Or how does this all fit together?

Kriechi:
Upstream is merged - do we need to wait for an actual release, or should we use a dev-commit in our master branch until they release, so we can also cut a release?

test/mitmproxy/net/test_tcp.py
mhils: Well.. the confusing part here is OpenSSL:
```python
ctx.set_cipher_list('AES256-GCM-SHA384')
ctx.get_cipher_list()  # returns ['AES256-GCM-SHA384', 'TLS_AES_256_GCM_SHA384', ...]
```
This is because they also introduced `SSL_set_ciphersuites` for TLS 1.3, and `SSL_get_cipher_list` returns whatever you set with `SSL_set_cipher_list` _plus_ what you set with `SSL_set_ciphersuites`. 🤷‍♂ 

I guess the correct approach for us would be to also introduce a cipher_suites option, and always set both. A lot of the code we're touching here will probably be thrown out with sans-io, so I decided to not fix all that up.

Minimum version dependency: cryptography is not guaranteed to ship with OpenSSL 1.1.1, you can also build it with older versions. So in either case we'd need to support both.

mhils:
If we want to cut a release, I would recommend we keep using cryptography 2.4 and we wait with the PR here until a new cryptography release appears.

That being said, we probably should ship one to fix the iOS 13 stuff. I'm a bit egoistic at the moment and pushing sans-io instead of doing all the other stuff that should be done.

RadionBik:
Hello all! 
will it be possible to activate the eSNI feature with this PR?

mhils:
@RadionBik: What in particular are you thinking of?

RadionBik:
> @RadionBik: What in particular are you thinking of?

I am thinking about proxying requests with and without the eSNI option (e.g. TLSv1.2 vs eSNI TLSv1.3), so as to compare the same traffic/ behavior in both cases. Plain activation of eSNI in browsers works, but I haven't found a possibility  to reproduce experiments with the same requests

mhils:
We currently don't support eSNI, and probably won't as long as OpenSSL doesn't support it: https://github.com/openssl/openssl/issues/7482

mhils:
One reason our builds are failing is that our SSLKEYLOGFILE implementation breaks somewhere between OpenSSL 1.1.1d and 1.1.1f:

```
  File "d:\a\mitmproxy\mitmproxy\.tox\py37\lib\site-packages\OpenSSL\SSL.py", line 1321, in wrapper
    callback(Connection._reverse_mapping[ssl], where, return_code)
  File "D:\a\mitmproxy\mitmproxy\mitmproxy\net\tls.py", line 99, in __call__
    masterkey = binascii.hexlify(connection.master_key())
  File "d:\a\mitmproxy\mitmproxy\.tox\py37\lib\site-packages\OpenSSL\SSL.py", line 2168, in master_key
    assert length > 0
AssertionError
```

It looks like `SSL_SESSION_get_master_key` doesn't work in the `SSL_CB_HANDSHAKE_DONE` info callback anymore. 🤷‍♂ 

https://github.com/pyca/cryptography/pull/5179 will fix this long-term, for now I will come up with some workarounds.

PR #3693, opened 2019-11-13
Improve Command Bar UX

### LINKED ISSUE #3259: Unable to enter backslash or parenthesis in a display filter regex interactively
Hi, thanks for your great work with mitmproxy!

I beleive I found a bug.

# Example:
Let's say I want to show only requests containing a digit in url.
Filter `\d` does the trick. However, I cannot enter it interactively via `f` hotkey. Setting it via `--view-filter "\d"` works fine.

# System info
I'm running an image from [dockerhub](https://hub.docker.com/r/mitmproxy/mitmproxy/)
```
# mitmproxy --version
Mitmproxy: 4.0.3
Python:    3.6.3
OpenSSL:   OpenSSL 1.0.2o  27 Mar 2018
Platform:  Linux-4.17.3-200.fc28.x86_64-x86_64-with
```

### LINKED ISSUE #3689: Filter regex does not work like Python
#### Problem Description

Hello. I am trying to filter the flows in a case insensitive manner. The documentation (https://docs.mitmproxy.org/stable/concepts-filters/) says that the accepted regex is just like Python:

> Regexes are Python-style

It doesn't seem to be the case.

If I try to enter the filter `~bq (?i)user`, I get the error `Invalid interception filter: ~bq (?i)user`.

The expected result was that I would see any flows that contain the word `user` in the request in a case insensitive way.

1. Make a couple of requests to add a few flows to mitmproxy
2. Press `f` and type the filter ~bq (?i)user
3. Notice the error message

#### System Information

$ mitmproxy --version
Mitmproxy: 4.0.4
Python:    3.7.3
OpenSSL:   OpenSSL 1.1.1b  26 Feb 2019
Platform:  Linux-5.0.0-31-generic-x86_64-with-Ubuntu-19.04-disco

typoon:
These changes fix the issues with the command bar where using backslash and quotes was not properly working. 

With that fix, regexes can also now be properly written and work just fine.

closes #3259 
closes #3689


mitmproxy/command.py
```diff
@@ -86,6 +76,20 @@ def signature_help(self) -> str:
         return "%s %s%s" % (self.path, params, ret)
 
     def prepare_args(self, args: typing.Sequence[str]) -> typing.List[typing.Any]:
+
+        # Arguments that are just blank spaces aren't really arguments
+        # We need to get rid of those. If the user intended to pass a sequence
+        # of spaces, it would come between quotes
+        clean_args = []
+        for a in args:
+            if isinstance(a, str):
```
mhils: `args` is defined as `typing.Sequence[str]` in the method signature, so this check seems superfluous?

mitmproxy/command.py
```diff
@@ -159,23 +163,29 @@ def parse_partial(
...
-                break
-            parts.append(t)
+
+        rex = pyparsing.QuotedString("\"", escChar='\\', unquoteResults=False) |\
+            pyparsing.QuotedString("'", escChar='\\', unquoteResults=False) |\
+            pyparsing.Combine(pyparsing.Literal('"') + pyparsing.Word(pyparsing.printables + " ") + pyparsing.StringEnd()) |\
+            pyparsing.Word(pyparsing.printables) |\
+            pyparsing.Word(' ')
+
+        rex = rex.copy().leaveWhitespace()
```
mhils: Perf 3: superfluous `.copy()`

mitmproxy/command.py
```diff
@@ -159,23 +163,29 @@ def parse_partial(
...
-            try:
-                t = lex.get_token()
-            except ValueError:
-                parts.append(remainder)
-                break
-            if not t:
-                break
-            parts.append(t)
+
+        rex = pyparsing.QuotedString("\"", escChar='\\', unquoteResults=False) |\
```
mhils: @BkPHcgQL3V did some previous work that showed we're on the hot path here and need to be a bit mindful of performance.

Perf 1: Move the whole parsing step into an isolated function that can be cached with `functools.lru_cache`.
Perf 2: Create `rex` once only and then re-use it.

typoon:
I am not sure what is this `codecov/patch` failure that is showing up there.

Other than that, code is ready for review.

mitmproxy/command.py
```diff
@@ -159,23 +163,29 @@ def parse_partial(
...
+            pyparsing.QuotedString("'", escChar='\\', unquoteResults=False) |\
+            pyparsing.Combine(pyparsing.Literal('"') + pyparsing.Word(pyparsing.printables + " ") + pyparsing.StringEnd()) |\
+            pyparsing.Word(pyparsing.printables) |\
+            pyparsing.Word(' ')
+
+        rex = rex.copy().leaveWhitespace()
+
+        for t, start, end in rex.scanString(cmdstr):
+            parts.append(t[0])
```
mhils: Perf 4: If we only use the element itself, we can just go with `parseString`.

mitmproxy/command.py
```diff
@@ -159,23 +163,29 @@ def parse_partial(
...
-            parts.append("")
+            parts = []
+
+        # First item in parts has always to be the command
+        # so we remove any blank tokens from the start of it
+        while True:
+            if parts and parts[0].strip() == '':
+                del parts[0]
+            else:
+                break
```
mhils: What's the reasoning for special-casing whitespace at the start? It doesn't make much sense, but I don't think we need a special case for it. Am I missing something?

(if I am indeed missing something, can there be more than one blank token at the start? If not, this should become a simple if)

mitmproxy/command.py
```diff
@@ -186,10 +196,15 @@ def parse_partial(
                 if parts[i] in self.commands:
                     params.extend(self.commands[parts[i]].paramtypes)
             elif params:
-                typ = params.pop(0)
-                if typ == mitmproxy.types.Cmd and params and params[0] == mitmproxy.types.Arg:
-                    if parts[i] in self.commands:
-                        params[:] = self.commands[parts[i]].paramtypes
+                if parts[i].strip() != '':
```
mhils: `elif params and not parts[i].isspace():` :)

mitmproxy/command.py
```diff
@@ -203,6 +218,8 @@ def parse_partial(
                 else:
                     valid = True
 
+            # if ctx.log:
+            #     ctx.log.info('[gilga] before parse.append. value = %s' % parts[i])
```
mhils: those can go away :)

test/mitmproxy/tools/console/test_defaultkeys.py
```diff
@@ -3,24 +3,33 @@
...
+
         assert cmd in m.commands.commands
 
         cmd_obj = m.commands.commands[cmd]
         try:
             cmd_obj.prepare_args(args)
         except Exception as e:
+
+            import pdb
+            pdb.set_trace()
```
mhils: probably no :)

mitmproxy/command.py
typoon: Even though it is defined like that, for some reason python is not checking it? One of the tests was passing the number 3 there, and the function accepted it (causing an error since a number can't be `split()`ed. I am not much of a python programmer, but seems to me that typing in python is just for annotating, not to enforce? Anyways, that is the reason why I added that check.

mitmproxy/command.py
typoon: Hmmm... not sure why this is part of the hot path. It is supposed to be called when you are executing a command. Anyways, I separated it in a different function. How do I make it cacheable by functools.lru_cache?

mitmproxy/command.py
typoon: So, three options there. If I changes this back to parseString I either have to handle an exception for the case that the string is empty or I have to check that the string is not empty prior to calling parseString. The third option is leaving it as a loop. Any preferences?

mitmproxy/command.py
typoon: In my opinion there shouldn't be any situations where this happens (I changed the code to prevent a whitespace from being added there from the commander), but there was a test case for it with various line brakes or spaces before the actual command. 

If you feed the parser with `"\n \n \n p1"` the response will be `[' ', ' ', ' ', 'p1'], so there is a chance to have more than one blank token prior to the command making the loop necessary.

mitmproxy/command.py
typoon: Fixed

mitmproxy/command.py
typoon: Oops, my bad

test/mitmproxy/tools/console/test_defaultkeys.py
typoon: Forgot that one there :$

mitmproxy/command.py
typoon: Fixed

mitmproxy/command.py
typoon: Alright, just read the documentation for lru_cache. I have my doubts about the effectiveness of it though considering that every time a letter is typed this function will be called, so if the user decides to type the expression:

`: set view_filter=~bq "(?i)[a-z]{2,3}"`

This will take 36 entries in the cache already.

There are 2 commands that I see calling this function like once a second each: `view.properties.marked` and `view.properties.length`. For these two, the cache might help a bit, but I think that is the only occasion.

Just to clarify, this is the function I created:

```python
def _parse_cmd(cmdstr: str):

    import mitmproxy.ctx as ctx

    if ctx.log:
        ctx.log.info(cmdstr)
    parts: typing.List[str] = []

    regex = pyparsing.QuotedString("\"", escChar='\\', unquoteResults=False) |\
        pyparsing.QuotedString("'", escChar='\\', unquoteResults=False) |\
        pyparsing.Combine(pyparsing.Literal('"') + pyparsing.Word(pyparsing.printables + " ") + pyparsing.StringEnd()) |\
        pyparsing.Word(pyparsing.printables) |\
        pyparsing.Word(' ')

    regex = regex.leaveWhitespace()

    for t, start, end in regex.scanString(cmdstr):
        parts.append(t[0])

    if not parts:
        parts = []

    return parts
```

mhils: > I have my doubts about the effectiveness of it though considering that every time a letter is typed this function will be called, so if the user decides to type the expression:

You are right - it doesn't matter for users typing commands and we wouldn't need to/shouldn't cache here. It does matter for rendering the flow list: Some commands are called on each render (such as `view.properties.marked`), and this adds up when scrolling quickly or when lots of new flows are arriving.

mitmproxy/command.py
mhils: How about we just do a simple `if not cmdstr or cmdstr.isspace(): return []` at the start? :)

mitmproxy/command.py
mhils: Ok, two thoughts here:

1) Let's fix the parser to return `\n \n \n ` as a single whitespace thing. That seems more correct.
2) What happens if we allow starting spaces in the commander? Can't we treat all spaces equally at a later point and not special case anything before? Of course you are right that spaces at the start don't make much sense, but it shouldn't be harmful and maybe we can reduce complexity if we tolerate it? :)

mitmproxy/command.py
mhils: Yes, Python's type system is an afterthought and not enforced. If we specify in the function signature that we expect a sequence of strings, we should just assume that this is true. Let's remove the test and the special case here.

mitmproxy/command.py
mhils: Second amendment: let's move the creation of `regex` out of the function definition, that object can be re-used. :)

mitmproxy/command.py
typoon: Alright. Done!

mitmproxy/command.py
typoon: Done!

mitmproxy/command.py
typoon: That function ended up becoming useless at the end after moving the regex object out of it. It ended up being:

```python
        parts: typing.List[str] = []
        for t, start, end in self.regex.scanString(cmdstr):
            parts.append(t[0])
        return parts
```

So I put that loop back into `parse_partial` and cached it instead (which should give a better gain, considering it does more stuff)

mitmproxy/command.py
typoon: 1 - Done
2 - This is where I am now. If we allow whitespaces before the command, then we need to change `CommandManager.execute` to find the first value in the `params` array that has a type of `mitmproxy.types.Cmd` to pass it as the first param to `CommandManager.call_strings`. It is not too pretty, but has to be done.

In that note, we should turn `CommandManager.call_strings` and `CommandManager.call` into private methods. Currently there are 3 entry points available to execute a command, and I don't think that is a good idea. Anything that wants to execute a command, should just use `CommandManager.execute`. If you agree, I can make this change as well.

typoon: Alright, I made changes to make `CommandManager.execute` the only entry point for command execution and changed `call_strings` to a private method by prepending it with an underscore (it is no _call_strings).

The `execute` method does the appropriate cleaning prior to executing commands.

Hopefully with these changes things are in good shape. If you disagree with anything, just let me know.

mhils:
This looks really good! The only thing we don't do properly yet is remove quotation marks, i.e. 
```
view.flows.create get "https://example.com/"
```
should do the same as 
```
view.flows.create get https://example.com/
```
😃

typoon:
Alright, this should resolve the issue.

test/mitmproxy/test_command.py
```diff
@@ -356,20 +391,19 @@ def test_parse_partial(self):
...
         c.add("one.two", a.cmd1)
-        assert(c.commands["one.two"].help == "cmd1 help")
-        assert(c.execute("one.two foo") == "ret foo")
-        assert(c.execute("one.two \"foo\"") == "ret foo")
-        assert(c.execute("one.two \"foo bar\"") == "ret \"foo bar\"")
-        assert(c.call("one.two", "foo") == "ret foo")
+        assert (c.commands["one.two"].help == "cmd1 help")
+        assert (c.execute("one.two foo") == "ret foo")
+        assert (c.execute("one.two \"foo\"") == "ret foo")
+        assert (c.execute("one.two \"foo bar\"") == "ret foo bar")
```
typoon: So... this might not be correct. If we passed the string between quotes, we want it to come back between quotes when there is a space in the string. This signals that the params were treated as a single thing (in this case as a single string). 
This change re-introduces the bug I was trying to fix in the first place.

Try the following command:

`: set view_fiter=~bq "(?i)[^\"]ser"`

The application will say the filter is invalid, when in fact it is supposed to be valid.

mitmproxy/command.py
```diff
@@ -33,228 +34,236 @@ def typename(t: type) -> str:
...
-            to = mitmproxy.types.CommandTypes.get(typ, None)
+            arg_is_known_command = (
+                    expected_type == mitmproxy.types.Cmd and part in self.commands
+            )
+            arg_is_unknown_command = (
+                    expected_type == mitmproxy.types.Cmd and part not in self.commands
+            )
+            command_args_following = (
+                    next_params and next_params[0].type == mitmproxy.types.CmdArgs
+            )
```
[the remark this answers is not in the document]
mhils: I just auto-format with the default settings of PyCharm. Less spacing would be just as fine for me, but I also do like keeping default configs. :) 🤷‍♂

mitmproxy/addons/view.py
```diff
@@ -298,16 +298,16 @@ def get_order(self) -> str:
 
     # Filter
     @command.command("view.filter.set")
-    def set_filter_cmd(self, f: str) -> None:
+    def set_filter_cmd(self, filtstr: str) -> None:
```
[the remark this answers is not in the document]
mhils: Good call. Adjusting this and the other ones.

mitmproxy/addons/core.py
```diff
@@ -168,19 +168,20 @@ def flow_set_options(self) -> typing.Sequence[str]:
             "reason",
         ]
 
-    @command.command("flow.set")
-    @command.argument("spec", type=mitmproxy.types.Choice("flow.set.options"))
+    @command.command(
+        "flow.set")
```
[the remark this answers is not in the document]
mhils: Accidental. :)

mitmproxy/command.py
```diff
@@ -33,228 +34,236 @@ def typename(t: type) -> str:
...
+            ret = ""
+        return f"{self.name} {params}{ret}"
 
-    def retname(self) -> str:
-        return typename(self.returntype) if self.returntype else ""
+    def prepare_args(self, args: typing.Sequence[str]) -> inspect.BoundArguments:
+        try:
+            bound_arguments = self.signature.bind(*args)
+        except TypeError as v:
+            raise exceptions.CommandError(f"Command argument mismatch: {v.args[0]}")
```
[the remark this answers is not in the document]
mhils: So yes - if we annotate this as `typing.Sequence[str]`, we clearly expect a list of strings or a tuple of strings. 
Now the odd thing is that a simple `str` is also a Sequence of chars, or better, one-character strings, which makes this a bit weird (see: https://github.com/python/mypy/issues/5090). 

Long story short, the annotation here conveys intent, and the intent is that you should call it with a list/tuple. Good enough for me. :)

mitmproxy/command.py
```diff
@@ -266,27 +275,37 @@ def dump(self, out=sys.stdout) -> None:
...
+        raise exceptions.CommandError(f"Unsupported argument type: {argtype}")
     try:
-        return t.parse(manager, argtype, spec)  # type: ignore
+        return t.parse(manager, argtype, spec)
     except exceptions.TypeError as e:
         raise exceptions.CommandError from e
 
 
-def command(path):
+def command(name: typing.Optional[str]):
```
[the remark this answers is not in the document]
mhils: The idea is that you can do 
```python
@command.command()
def foo_bar():
    pass
```
and we'd automatically name that command `foo.bar`. You can override this by explicitly setting a different name. This pretty much mirrors what click is doing: https://click.palletsprojects.com/en/7.x/api/#click.command

mitmproxy/tools/console/commander/commander.py
```diff
@@ -1,52 +1,49 @@
...
-        self.options: typing.Sequence[str] = []
+        self.options: typing.List[str] = []
         for o in options:
             if o.startswith(start):
                 self.options.append(o)
         self.options.sort()
         self.offset = 0
 
-    def cycle(self) -> str:
+    def cycle(self, forward: bool) -> str:
```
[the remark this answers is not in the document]
mhils: That means we'd have another enum. Let's keep it super simple for now. I'll make it default to True, so you can just call `.cycle()`. :)

PR #3700, opened 2019-11-15
Ctrl-C should prompt the user before exiting

Jessonsotoventura:
Currently hitting q inside of mitmproxy prompts the user before quitting. Ctrl-C should match that behavior and prompt the user before quitting.

### DIFF
--- a/mitmproxy/tools/_main.py
+++ b/mitmproxy/tools/_main.py
@@ -11,7 +11,6 @@ import signal
 import typing
 
 from mitmproxy.tools import cmdline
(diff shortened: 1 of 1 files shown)


mhils:
👍  on the overall change, but our mitmdump executables don't ship with `mitmproxy.tools.console`: https://travis-ci.org/mitmproxy/mitmproxy/jobs/612168142#L2799

I'd be fine with changing the check to `getattr(master, "prompt_for_exit", master.shutdown)`

mhils:
Awesome, thanks! 👍

Issue #3701, opened 2019-11-15
Export Plaintext HTTP Response 


mckeimic:
A common irritation for my team when using mitmproxy is the inability to export the full HTTP response. While the request can be clipped or saved to a file with the `:export.clip raw @focus` or `:export.file raw @focus /tmp/request.txt`, there is no analogous method to save the full text of the response.

I suggest a modification to the export command which supports exporting both the raw request, and the raw response for the flow. Specifically, I suggest splitting the `raw` export type into `raw_request` and `raw_response`.

I've created a custom response export script as a stop-gap measure for myself; however I believe the ability to export the full, plaintext request/response pairs for documentation is a common requirement for penetration testers. As such, there may be a significant enough need to perform this action that a modification to mitmproxy itself is worthwhile. 

Unlike the other export options curl and httpie, which are often used to modify and resubmit the original request, my team and I use the raw option to produce documentation of the flow for our reports. For this purpose, the response is as important as the request. A clean way to export the plaintext HTTP of the response is an important usability feature for us.

mckeimic:
Feature added by https://github.com/mitmproxy/mitmproxy/pull/3702

PR #3702, opened 2019-11-15
Adding export raw http response

mckeimic:
Adding a new export type for raw http response, and changing export raw to export
raw_request to distinguish between the two. This is a proposed change for https://github.com/mitmproxy/mitmproxy/issues/3701


ylmrx:
I wouldn't rename `raw->raw_request`, having `raw_response` is indeed useful.

mhils:
Thanks! Not exporting responses clearly is an oversight. Adding to what @ylmrx said, maybe we just want to go for `raw_request`, `raw_response` and `raw` (which outputs both combined)?

Kriechi:
@mckeimic looks like we have a merge conflict now - would you mind looking into it?

mhils:
Ah, I wanted to merge those the other way around. I'll take care of the merge conflict, this should be straightforward to fix :)

mckeimic:
You folks are lightning fast! Thank you!

PR #3713, opened 2019-11-17
Use Github Actions for CI

mhils:
This PR replaces our existing Travis/AppVeyor infrastructure with GitHub Actions. This has a bunch of benefits:

1. Strong GitHub integration. For example, our linter now automatically annotates PRs. We'll hopefully get similar functionality for pytest (or codecov) at some point :)

image

2. Speed. We can now define separate jobs with dependencies between them and form pipelines. This allows us to run all tests and build all binaries on Windows/Linux/macOS in parallel, and then - if everything succeeds - deploy everything. This cuts build time by a factor of 4, and makes linting/mypy results available after ~30s.

3. Resources. GitHub actions comes with 20 parallel builds for free.

**TODO:**

- [ ] @cortesi: add DOCKER_PASSWORD to https://github.com/mitmproxy/mitmproxy/settings/secrets


release/cibuild.py
```diff
@@ -265,18 +297,24 @@ def should_upload_pypi(self) -> bool:
         ])
 
     @property
-    def tag(self):
-        return self.travis_tag or self.appveyor_repo_tag_name
+    def tag(self) -> str:
+        if self.travis_tag:
```
Kriechi: Do we still need this - and the one below for appveyor?

release/cibuild.py
```diff
@@ -232,20 +255,22 @@ def is_prod_release(self) -> bool:
 
     @property
     def is_pull_request(self) -> bool:
+        if self.github_event_name == "pull_request":
+            return True
         if self.appveyor_pull_request_number:
             return True
         if self.travis_pull_request and self.travis_pull_request != "false":
```
Kriechi: Are these two still needed?

release/cibuild.py
```diff
@@ -143,26 +156,34 @@ def bdists(self):
         return ret
 
     @property
-    def branch(self):
-        return self.travis_branch or self.appveyor_repo_branch
+    def branch(self) -> str:
+        if self.travis_branch:
+            return self.travis_branch
+        if self.appveyor_repo_branch:
+            return self.appveyor_repo_branch
```
Kriechi: same as below

release/cibuild.py
mhils: Likely no, but I would keep this until we are certain that GitHub Actions does everything we want.

PR #3714, opened 2019-11-18
Add support for IPv6-only environments

### LINKED ISSUE #3695: Add IPv6-only mode
**Is your feature request related to a problem? Please describe.**
When starting the proxy on a machine that has IPv4 disabled, the proxy fails to start. Taking a look at the code, it seems that issue occurs when the socket option IPV6_V6ONLY is set to 0 and the system does not allow IPv4-enabled sockets. 

Adding an ipv6_only option was pretty quick & easy, and I have a potential patch, but I wanted to see if this is a reasonable thing to do?

It could also make sense to--instead of adding an ipv6_only option--add an option that gives the user full control to select ipv6_only, ipv4_only, or both ipv4 + ipv6, where the proxy does not attempt to fall back on another protocol.

sethb157:
Previously, the proxy would attempt to bind to an IPv6 + IPv4 enabled socket.
On failure, it would try to bind to an IPv4 only socket. If that failed, the
proxy would fail to start. This update makes it so that the proxy also tries
to bind to an IPv6-only socket, which is necessary in environments where IPv4
is disabled.

In short, the proxy will try binding in the following order, only moving
to the next step when binding fails:
IPv6 + IPv4 -> IPv4 only -> IPv6 only -> proxy fails to start.

Closes #3695

### DIFF
--- a/mitmproxy/net/tcp.py
+++ b/mitmproxy/net/tcp.py
@@ -558,7 +558,7 @@ class TCPServer:
         self.socket = None
 
         try:
-            # First try to bind an IPv6 socket, with possible IPv4 if the OS supports it.
+            # First try to bind an IPv6 socket, attempting to enable IPv4 support if the OS supports it.
             # This allows us to accept connections for ::1 and 127.0.0.1 on the same socket.
             # Only works if self.address == ""
             self.socket = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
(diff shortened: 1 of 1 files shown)


sethb157:
Hey @mhils! I think this is roughly what we discussed. Haven't added any tests yet, as I'm not sure what the best way to do so in this case is. Any suggestions?

mhils:
Awesome, thanks! 🍰 
Testing this on CI is tricky - we can't really do much here without mocking everything, which then again becomes kind of a self-fulfilling prophecy. Let's get this in as-is! :shipit:

PR #3715, opened 2019-11-19
Several fixes

ylmrx:
- correct linux install procedure (sudo misuse)
- formatting in windows setup
- loop reference in source install procedure

### DIFF
--- a/docs/src/content/overview-installation.md
+++ b/docs/src/content/overview-installation.md
@@ -29,16 +29,19 @@ binary-packages from our
 [releases](https://github.com/mitmproxy/mitmproxy/releases/latest) page.
 
 Some Linux distributions provide community-supported mitmproxy packages through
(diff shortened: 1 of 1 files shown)


docs/src/content/overview-installation.md
```diff
@@ -114,7 +135,12 @@ pip3 install mitmproxy
...
-on GitHub.
+Download the [release
+package](https://github.com/mitmproxy/mitmproxy/releases).
+
+{{< highlight bash  >}}
+tar xvzf mitmproxy-[VERSION].tar.gz
+cd mitmproxy-[VERSION]
+pip install .
+{{< / highlight >}}
+
```
mhils: I'm 👎 on this. I think we do want to point to the README here, so that everyone looking at the source code (no matter by which entrypoint) uses the description in README.rst.

mhils:
Closed in favor of #3718. Thanks again! :smiley:

PR #3718, opened 2019-11-20
Simplify install instructions

mhils:
Prompted by @ylmrx's #3715 regarding the same issue, here's a counter-proposal. This simplifies install instructions in the following ways:

- Replace pip3 installation instructions with [pipx](https://pipxproject.github.io/pipx/). This is much less overhead for users unfamiliar with the Python ecosystem. I've used it a couple of times in the last months and was overall _very_ happy with how pipx performed.
- Fix location of binary packages. We now just link to mitmproxy.org, which has the big OS-specific download button.
- Streamline development setup a bit. Invoking `tox` to test _all_ Python versions more confusing than anything else, `tox -e py` and `tox -e lint` are more adequate.

@ylmrx, would that also address your concerns?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,7 +57,16 @@ please consider contributing in the following areas:
 Development Setup
 -----------------
 
-To get started hacking on mitmproxy, please follow the `advanced installation`_ steps to install mitmproxy from source, but stop right before running ``pip3 install mitmproxy``. Instead, do the following:
+To get started hacking on mitmproxy, please install a recent version of Python (we require at least 3.6).
+The following commands should work on your system:
+
+.. code-block:: bash
+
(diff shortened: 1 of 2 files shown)


ylmrx:
@mhils It's very good and  much clearer (I didn't know about pipx, and I need to play a bit with it).

My only worry is when the user need to `sudo` in order to use the transparent mode. Just running "`sudo mitmproxy`" _won't_ work, and anything implying 'pip' and 'sudo' scares me.

I'll sync and push another PR later on that topic if you feel it's needed. Good job :100: .

Kriechi:
Looks good in general.
I'm undecided on the pipx thing - I've not heard about it yet...

PR #3724, opened 2019-11-22
Save user executed commands to a history file

typoon:
This is the implementation of a feature that saves all user executed commands (`:` prompt) to a file so that the next time they open mitmproxy those commands are already in their command history.

This has been pretty useful for me when testing issues that would cause mitmproxy to crash (so I didn't have to re-type a bunch of commands again). It has also been pretty useful for the cases when I have to close mitmproxy for any reason during my workflow and reopen it and continue where I left of.

### DIFF
--- a/mitmproxy/addons/command_history.py
+++ b/mitmproxy/addons/command_history.py
@@ -15,7 +15,7 @@ class CommandHistory:
         self.filter_str: str = ''
 
         _command_history_path = os.path.join(os.path.expanduser(ctx.options.confdir), 'command_history')
-        _history_lines = []
+        _history_lines: typing.List[str] = []
         if os.path.exists(_command_history_path):
             _history_lines = open(_command_history_path, 'r').readlines()
 
--- a/mitmproxy/addons/command_history.py
+++ b/mitmproxy/addons/command_history.py
@@ -7,6 +7,7 @@ import mitmproxy.options
 import mitmproxy.types
 
 from mitmproxy import command
+from mitmproxy import ctx
 from mitmproxy.tools.console.commander.commander import CommandBuffer
 
 
@@ -22,8 +23,10 @@ class CommandHistory:
         self.filtered_commands: typing.Deque[str] = collections.deque()
         self.filter_active: bool = True
 
-        _command_history_path = os.path.join(os.path.expanduser(mitmproxy.options.CONF_DIR), 'command_history')
-        _history_lines = open(_command_history_path, 'r').readlines()
+        _command_history_path = os.path.join(os.path.expanduser(ctx.options.confdir), 'command_history')
+        _history_lines = []
+        if os.path.exists(_command_history_path):
+            _history_lines = open(_command_history_path, 'r').readlines()
 
         self.command_history_file = open(_command_history_path, 'w')
 
--- a/mitmproxy/addons/command_history.py
+++ b/mitmproxy/addons/command_history.py
@@ -1,6 +1,7 @@
 import collections
 import os
 import typing
+import atexit
 
 from mitmproxy import command
 from mitmproxy import ctx
@@ -27,6 +28,12 @@ class CommandHistory:
 
         for l in _history_lines:
             self.add_command(l.strip())
+
+        atexit.register(self.cleanup)
+
+    def cleanup(self):
+        if self.command_history_file:
+            self.command_history_file.close()
 
     @property
     def last_filtered_index(self):
--- a/mitmproxy/addons/command_history.py
+++ b/mitmproxy/addons/command_history.py
@@ -1,7 +1,7 @@
+import atexit
 import collections
 import os
 import typing
-import atexit
 
 from mitmproxy import command
 from mitmproxy import ctx
@@ -19,12 +19,12 @@ class CommandHistory:
         if not os.path.exists(_command_history_dir):
             os.makedirs(_command_history_dir)
 
-        _command_history_path = os.path.join(_command_history_dir, 'command_history')
+        self.command_history_path = os.path.join(_command_history_dir, 'command_history')
         _history_lines: typing.List[str] = []
-        if os.path.exists(_command_history_path):
-            _history_lines = open(_command_history_path, 'r').readlines()
+        if os.path.exists(self.command_history_path):
+            _history_lines = open(self.command_history_path, 'r').readlines()
 
-        self.command_history_file = open(_command_history_path, 'w')
+        self.command_history_file = open(self.command_history_path, 'w')
 
         for l in _history_lines:
             self.add_command(l.strip())
@@ -32,7 +32,8 @@ class CommandHistory:
         atexit.register(self.cleanup)
 
     def cleanup(self):
-        if self.command_history_file:
+        self._reload_saved_commands()
+        if self.command_history_file and not self.command_history_file.closed:
             self.command_history_file.close()
 
     @property
@@ -99,9 +100,10 @@ class CommandHistory:
         if command.strip() == '':
             return
 
+        self._reload_saved_commands()
+
         if command in self.saved_commands:
             self.saved_commands.remove(command)
-
         self.saved_commands.append(command)
 
         _history_str = "\n".join(self.saved_commands)
@@ -111,3 +113,17 @@ class CommandHistory:
         self.command_history_file.flush()
 
         self.restart()
+
+    def _reload_saved_commands(self):
+        # First read all commands from the file to merge anything that may
+        # have come from a different instance of the mitmproxy or sister tools
+        if not os.path.exists(self.command_history_path):
+            return
+
+        _history_lines = open(self.command_history_path, 'r').readlines()
+        self.saved_commands.clear()
+        for l in _history_lines:
+            l = l.strip()
+            if l in self.saved_commands:
+                self.saved_commands.remove(l)
+            self.saved_commands.append(l.strip())
--- a/mitmproxy/addons/command_history.py
+++ b/mitmproxy/addons/command_history.py
@@ -1,0 +1,152 @@
+import collections
+import copy
+import os
+import typing
+
+import mitmproxy.options
+import mitmproxy.types
+
+from mitmproxy import command
+from mitmproxy.tools.console.commander.commander import CommandBuffer
+
+
+class CommandHistory:
+    def __init__(self, size: int = 300) -> None:
+        self.saved_commands: typing.Deque[str] = collections.deque(
+            maxlen=size
+        )
+        self.index: int = 0
+
+        self.filter: str = ''
+        self.filtered_index: int = 0
+        self.filtered_commands: typing.Deque[str] = collections.deque()
+        self.filter_active: bool = True
+
+        _command_history_path = os.path.join(os.path.expanduser(mitmproxy.options.CONF_DIR), 'command_history')
+        _history_lines = open(_command_history_path, 'r').readlines()
+
+        self.command_history_file = open(_command_history_path, 'w')
+
+        for l in _history_lines:
+            self.add_command(l.strip(), True)
+
+    @property
+    def last_index(self):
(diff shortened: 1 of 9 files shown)


mhils:
Can we encapsulate the command history into a seperate addon that provides get/add/clear commands? This would automatically make this a singleton and we can re-use it for mitmweb, which should also get command support eventually. :-)

typoon:
I can for sure try to do it like that. Will give it a shot and commit the attempt as soon as it's ready.

Kriechi:
How will this behave when running multiple mitmproxy/mitmdump/mitmweb commands in parallel or different terminal windows?

typoon:
Yeah, I thought about this. Probably not a problem unless you are able to type a command on both windows at the same time. As soon as you type a command it is saved to the command history, now if you move to the other window, you will have that command in the history there as well. Kinda the same way that happens with bash.

If that is not the desired behavior, then we need to have the command history per PID and when the user leaves the application we copy everything in that process' command history to the main command history file.

Any preferences?

typoon:
Just to be clear, this is only saving the commands that are typed by the user in mitmproxy's TUI (so mitmdump and mitmweb won't be interfering here).

mitmproxy/addons/command_history.py
```diff
@@ -0,0 +1,113 @@
...
+    def __init__(self, size: int = 300) -> None:
+        self.saved_commands: typing.Deque[str] = collections.deque(maxlen=size)
+
+        self.filtered_commands: typing.Deque[str] = collections.deque()
+        self.current_index: int = -1
+        self.filter_str: str = ''
+
+        _command_history_dir = os.path.expanduser(ctx.options.confdir)
+        if not os.path.exists(_command_history_dir):
+            os.makedirs(_command_history_dir)
```
mhils: What happens if you do all this in `configure`? Is the directory available then?

mitmproxy/command.py
```diff
@@ -100,8 +100,10 @@ def signature_help(self) -> str:
     def prepare_args(self, args: typing.Sequence[str]) -> inspect.BoundArguments:
         try:
             bound_arguments = self.signature.bind(*args)
-        except TypeError as v:
-            raise exceptions.CommandError(f"Command argument mismatch: {v.args[0]}")
+        except TypeError:
+            expected = f'Expected: {str(self.signature.parameters)}'
+            received = f'Received: {str(args)}'
+            raise exceptions.CommandError(f"Command argument mismatch: \n\t{expected}\n\t{received}")
```
mhils: What is this solving here? IIRC `v.args[0]` was relatively sensible.

mitmproxy/tools/_main.py
```diff
@@ -70,19 +70,19 @@ def run(
...
         args = parser.parse_args(arguments)
     except SystemExit:
         arg_check.check()
         sys.exit(1)
+
+    # To make migration from 2.x to 3.0 bearable.
+    if "-R" in sys.argv and sys.argv[sys.argv.index("-R") + 1].startswith("http"):
+        print("-R is used for specifying replacements.\n"
+              "To use mitmproxy in reverse mode please use --mode reverse:SPEC instead")
+
```
mhils: Specific reason why this is moved?

mitmproxy/utils/debug.py
```diff
@@ -11,6 +11,11 @@
 from mitmproxy import version
 
 
+def remote_debug(host='localhost', port=4444):
+    import remote_pdb
+    remote_pdb.RemotePdb(host, port).set_trace()
+
+
```
mhils: Leftover :)

mitmproxy/addons/command_history.py
```diff
@@ -0,0 +1,113 @@
...
+
+        if self.current_index == -1 or self.current_index == self.last_filtered_index:
+            self.current_index = -1
+            return ''
+        elif self.current_index < self.last_filtered_index:
+            self.current_index += 1
+
+        ret = self.filtered_commands[self.current_index]
+
+        return ret
```
mhils: Hm... Do we want this as part of the addon, or do we want the addon to just provide a way to give us the entire list? The filtering etc. provides a lot of statefulness, which would be nice to avoid. Thinking ahead, if we integrate commands into mitmweb, we probably just want a single command that gives us the list (and no filtering on the Python side).

test/mitmproxy/tools/console/test_commander.py
```diff
@@ -1,9 +1,35 @@
...
+from mitmproxy.addons import command_history
 from mitmproxy.test import taddons
 from mitmproxy.tools.console.commander import commander
 
 
+@pytest.fixture(autouse=True)
+def tctx():
+    # This runs before each test
+    dir_id = str(uuid.uuid4())
+    confdir = os.path.expanduser(f"~/.mitmproxy-test-suite-{dir_id}")
```
mhils: `~/.mitmproxy-test-suite` could be a bit annoying. http://doc.pytest.org/en/latest/tmpdir.html should be the way to go.

mitmproxy/addons/command_history.py
```diff
@@ -0,0 +1,113 @@
...
+        self.filter_str: str = ''
+
+        _command_history_dir = os.path.expanduser(ctx.options.confdir)
+        if not os.path.exists(_command_history_dir):
+            os.makedirs(_command_history_dir)
+
+        _command_history_path = os.path.join(_command_history_dir, 'command_history')
+        _history_lines: typing.List[str] = []
+        if os.path.exists(_command_history_path):
+            _history_lines = open(_command_history_path, 'r').readlines()
```
mhils: Let's make sure that we always close files after reading:

1) Use `with open(...):` .... so that the file pointer is automatically closed.
2) Use pathlib and https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text

mitmproxy/addons/command_history.py
```diff
@@ -0,0 +1,113 @@
...
+        if command in self.saved_commands:
+            self.saved_commands.remove(command)
+
+        self.saved_commands.append(command)
+
+        _history_str = "\n".join(self.saved_commands)
+        self.command_history_file.truncate(0)
+        self.command_history_file.seek(0)
+        self.command_history_file.write(_history_str)
+        self.command_history_file.flush()
```
mhils: This really looks a bit fishy to me. Why can't we just 1) not keep a file handle open and 2) open in append mode and then add a line?

mitmproxy/command.py
typoon: v.args[0] doesn't give you any information about the args that were expected and what args were received. This facilitates debuging things if an error like that occurs.

mitmproxy/tools/_main.py
typoon: No reason whatsoever. Was trying to edit the file back to its original format and ended up moving that down by mistake. Will put it back to what is was.

mitmproxy/utils/debug.py
typoon: Not really. I added this as a facilitator for me to debug things. I thought that the `debug.py` file would be the place to have that. Is it a problem? If so, I can remove it.

typoon: Only thing I forgot is that for this I need to add remote_pdb as a dependency for mitmproxy. Not sure if you are cool with that or not.

mitmproxy/addons/command_history.py
typoon: This just filters the contents of the commands available. If you just want to use the whole list, just don't set a filter. `self.filtered` commands will have all available commands if the filter is empty. This allows for you viewing commands that were type before that start with something you know, for example, if you want to see the last `set` command you entered, just type `:set<up>` and you will get the last `set` command. Pretty useful, and mitmweb can make use of it if it wants to.

test/mitmproxy/tools/console/test_commander.py
typoon: Does it matter? The directory is deleted at the end of the test, so I thought that name would be fine.

mitmproxy/addons/command_history.py
typoon: I thought that the way I wrote it would cause the file to be closed after the readlines() call. My bad I guess. Will change it to use `with`. Any reason to use pathlib instead?

mitmproxy/addons/command_history.py
typoon: To handle multiple instances running at the same time. I can't just append to this file because I am checking if the command already exists in the history I just move it to the last command executed.

I could open and close the file every time a new command is entered or when the history is cleared. I would be okay with doing that, just thought that it would be "costly" for some reason, but thinking about it, this is probably not a problem.

mitmproxy/addons/command_history.py
typoon: Not sure... I will test and see if it works

mitmproxy/command.py
mhils: I'll get a proper `missing a required argument: 'y'` (as you would in the Python console).

mitmproxy/utils/debug.py
mhils: let's put it in https://github.com/mitmproxy/mitmproxy/blob/master/examples/complex/remote_debug.py.

test/mitmproxy/tools/console/test_commander.py
mhils: Probably not too much, but there are two arguments for `tmpdir`:

- Consistency: We already use `tmpdir` at other places. `tmpdir` gracefully handles test failures etc.
- Temporary files belong into the system temporary directory. 🙃

mitmproxy/addons/command_history.py
mhils: This is what pathlib is doing internally: https://github.com/python/cpython/blob/b9e5547f5814962964c4a5bd5cd36a2af8fbf974/Lib/pathlib.py#L1220-L1225
So I guess the answer is no. Personal preference. 😃

mitmproxy/addons/command_history.py
mhils: I agree on the not a problem part (given that this is only invoked for commands typed by humans), and we are probably less likely to run into issues with multiple mitmproxy instances running at the same time.

test/mitmproxy/tools/console/test_commander.py
typoon: Fixed!

mitmproxy/addons/command_history.py
typoon: Fixed!

mitmproxy/tools/_main.py
typoon: Fixed!

mitmproxy/addons/command_history.py
typoon: I changed it to use `with open` then. Should be good now

mitmproxy/command.py
typoon: Here is what I get if I give something too many parameters:

```
mitmproxy.exceptions.CommandError: Command argument mismatch: too many positional arguments
```

Doesn't tell me anything about how many extras I have given or what is expected.

typoon:
This seems ready for another review round. I think I covered everything discussed.

mitmproxy/addons/command_history.py
typoon: Seems to work fine with `configure`. Let me know what you think of the changes.

mitmproxy/addons/command_history.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @command.command("commands.history.add")
+    def add_command(self, command: str) -> None:
+        if not command.strip():
+            return
+
+        self.history.append(command)
+        if ctx.options.command_history:
+            with self.history_file.open("a") as f:
+                f.write(f"{command}\n")
```
Kriechi: should we add a timestamp to each line?

PR #3734, opened 2019-11-29
Fix converting to HAR in case of void response

worldmind:


mhils:
Awesome, thanks! :smiley:

Issue #3722, opened 2019-11-22
Search inside a flow causes mitmproxy to crash


typoon:
#### Problem Description

If you try to use the search feature inside a flow (by pressing '/') and type something to make a search, mitmproxy crashes.

#### Steps to reproduce the behavior:
1. Open mitmproxy and make a few requests
2. Selct one of the flows in the main screen and open it (by pressing Enter)
3. In the flow, press `/` and type something and then press Enter
4. Notice that mitmproxy crashes

#### System Information

$ mitmproxy --version
Mitmproxy: 5.0.0.dev (+413, commit 2f30981)
Python:    3.7.3
OpenSSL:   OpenSSL 1.1.0h  27 Mar 2018
Platform:  Linux-5.0.0-31-generic-x86_64-with-Ubuntu-19.04-disco

KORraNpl:
Can confirm, traceback:
```Traceback (most recent call last):
  File "/Users/username/Projects/mitmproxy/mitmproxy/master.py", line 86, in run_loop
    loop()
  File "/Users/username/Envs/mitmproxy-0GFVzhJ3/lib/python3.7/site-packages/urwid/main_loop.py", line 286, in run
    self._run()
  File "/Users/username/Envs/mitmproxy-0GFVzhJ3/lib/python3.7/site-packages/urwid/main_loop.py", line 384, in _run
    self.event_loop.run()
  File "/Users/username/Envs/mitmproxy-0GFVzhJ3/lib/python3.7/site-packages/urwid/main_loop.py", line 1484, in run
    reraise(*exc_info)
  File "/Users/username/Envs/mitmproxy-0GFVzhJ3/lib/python3.7/site-packages/urwid/compat.py", line 58, in reraise
    raise value
  File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
  File "/Users/username/Envs/mitmproxy-0GFVzhJ3/lib/python3.7/site-packages/urwid/raw_display.py", line 404, in <lambda>
    event_loop, callback, self.get_available_raw_input())
  File "/Users/username/Envs/mitmproxy-0GFVzhJ3/lib/python3.7/site-packages/urwid/raw_display.py", line 502, in parse_input
    callback(processed, processed_codes)
[...]

PR #3736, opened 2019-12-01
Bug fixed: Search inside a flow causes mitmproxy to crash #3722

peter-way:
https://github.com/mitmproxy/mitmproxy/issues/3722


mhils:
Thanks. Looking at the code there seems to be the same issue at line 137. Would you mind fixing it there as well? 😃

mhils:
Fixed, thanks!

PR #3738, opened 2019-12-05
Add message to ContentView metadata

gergesh:
This should greatly increase the flexibility of custom ContentViews


mhils:
@pravic already convinced me in #3218 that this may be useful, even though content views should ideally not depend on any context. I'm happy to accept a PR that also exposes the flow object, but that needs to adjust all relevant calls to `get_message_content_view`.

PR #3739, opened 2019-12-05
Add flow to ContentView metadata

gergesh:
The long-awaited sequel to #3738, adds the flow to the ContentView's metadata.

### DIFF
--- a/mitmproxy/addons/dumper.py
+++ b/mitmproxy/addons/dumper.py
@@ -86,10 +86,11 @@ class Dumper:
             )
             self.echo(out, ident=4)
 
-    def _echo_message(self, message):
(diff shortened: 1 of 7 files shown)


gergesh:
This brings up the question of whether we want to keep the `message` field in `metadata` or replace it with a flag for whether the selected field is the request or the response. Maybe both, to save boilerplate in all cases?

mhils:
Thanks! 🍰 

> This brings up the question of whether we want to keep the message field in metadata or replace it with a flag for whether the selected field is the request or the response.

We have some flows with N>2 messages, e.g. TCPFlows with TCPMessages. Moving to a flag would make that harder, so I think passing it separately is a good compromise. :)

PR #3754, opened 2019-12-21
docker: auto-update latest tag

mhils:


mhils:
@Kriechi, could you sanity-check this? We only support the latest release anyway, so updating `latest` on _any_ production release should be fine.

Kriechi:
👍 
(I'm surprised we didn't think of this the last time we refactored this bit 🙃)

PR #3774, opened 2020-01-04
Don't raise an exception when script is not found

sekkr1:
Is really helpful when you have scripts in your `config.yaml` and don't want to edit the file each time one gets deleted or moved.

### DIFF
--- a/mitmproxy/addons/script.py
+++ b/mitmproxy/addons/script.py
@@ -76,7 +76,8 @@ class Script:
         self.ns = None
 
         if not os.path.isfile(self.fullpath):
-            raise exceptions.OptionsError('No such script')
+            ctx.log.error('No such script: %s' % path)
+            return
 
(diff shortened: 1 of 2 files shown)


Kriechi:
Please also update the now failing test.

mhils:
I feel it's good that we visibly fail in this scenario. What's your use case of moving/deleting scripts?

sekkr1:
I have scripts in a specific repository branch as it takes time to get merged so in the meanwhile every time I work on a different branch I have to edit the config file.

mhils:
Hm - Apart from your particular use case, hard-erroring on a missing script feels like the correct thing to do. Could you maybe keep your config.yaml in the repository as well?

mhils:
I'm closing this in favor of the current behavior. I think it's more sensible to hard-error on missing scripts instead of requiring users to check stderr.

PR #3782, opened 2020-01-11
Exit on SIGTERM instead of prompting

ThinkChaos:
Catching SIGTERM and prompting for exit instead of just shutting down is pretty annoying.
I have scripts to manage mitmproxy, and this behavior was preventing them from working.


mhils:
That sounds like a sensible change. Thanks! 😃

Issue #2936, opened 2018-02-28
header names are needlessly line-wrapped


majewsky:
Even though there is plenty of horizontal space (as can be seen from the width of the "Request-Response-Details" tab bar).

![screen](https://user-images.githubusercontent.com/24696/36799077-8faf6490-1cac-11e8-9966-6f9181cf7925.png)

fenilgandhi:
Hi, i had a look into the source but it turns out that they are not needlessly wrapped up but it is intentional to be on the new line. It may not be clear for smaller urls but for longer urls it make the header responses visible which would otherwise be wrapped up. 
If you still have any other suggestions, feel free to share them.

majewsky:
I'm not sure I understand what you're saying. Just because header values *could* be larger, doesn't mean we have to line-wrap the header names when they aren't.

koknaroglu:
I want to think about this question. Which files can i work on?

mhils:
The correct fix here is to

1. Get available space.
2. Calculate maximum header name length.
3. Calculate maximum header value length.
4. Make sure that the header name takes up at least 30 chars, but more if the space is not fully used.

urvashi2004:
Hello
Wouldn't it be better to store the number of characters in the header length and then assign if they need wrapping or not according to their length?

PR #3785, opened 2020-01-17
Increase max header length

mprencipe:
This is a kludge to help with #2936. My knowledge of Urwid is slim but it really doesn't seem to offer any tools for this problem.

Using weight = 1 for the header key/value columns would use the available space more efficiently and get you wrapping for both key and value. The downside would be that the key and value would no longer be adjacent.

### DIFF
--- a/mitmproxy/tools/console/common.py
+++ b/mitmproxy/tools/console/common.py
@@ -34,7 +34,7 @@ def highlight_key(str, key, textattr="text", keyattr="key"):
     return l
 
 
-KEY_MAX = 30
+KEY_MAX = 50
 
 
 def format_keyvals(


mhils:
Thanks @mprencipe! If I understand things correctly, this PR would mean that we always allocate 50 characters for header keys if we have a key that long, and make values wrap in the rest of the space. Is that correct?

mprencipe:
Yes, nothing special, as you described. Maybe the previous 30 character limit was a bit too little for modern monitors.

mhils:
Hm. Assuming that header keys are generally smaller than header values, this would degrade the experience for everyone using a <100 col terminal window and we'll get a PR that reverts this in a couple of weeks. 😛 I'm not too much into urwid either, do we have any better approaches?

mprencipe:
Closing this as it's clearly not a good solution.

PR #3787, opened 2020-01-18
Fix console command quotes issues

itaisod:
This PR fixes several issues I've encountered while using mitmproxy in console mode:
1. Trying to manually change to "multipart form" contentview failed because ViewMultipart.name has a space in it. This also affected my custom contentview, so I had to choose a name without spaces.
2. Said error was not visible in the event log, just in the status bar (where it wasn't even informative because it just said "(more in eventlog)"...)
3. When I did manage to show the full exception in the event log, tabs ('\t') were displayed as '?'.
4. A re-raised error (when I run a command with an "invalid choice") would display an empty error string because it was raised like this: `raise exceptions.CommandError from e`

I solved (1) by exposing CommandManager.call_strings and calling it directly, because it seems far more reliable than formatting a full command string with `%s`, quote/unquote operations and so on. I also did this for any command call which involves a `%s` or `{choice}`.

I solved (2) by replacing `signals.status_message.send` with `ctx.log.error`.

### DIFF
--- a/mitmproxy/command.py
+++ b/mitmproxy/command.py
@@ -103,7 +103,7 @@ class Command:
         except TypeError:
             expected = f'Expected: {str(self.signature.parameters)}'
             received = f'Received: {str(args)}'
-            raise exceptions.CommandError(f"Command argument mismatch: \n\t{expected}\n\t{received}")
+            raise exceptions.CommandError(f"Command argument mismatch: \n    {expected}\n    {received}")
 
         for name, value in bound_arguments.arguments.items():
             convert_to = self.signature.parameters[name].annotation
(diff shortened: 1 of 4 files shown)


itaisod:
Didn't see the part about coverage, I'll fix that.

mhils:
Thanks @itaisod! That looks all very reasonable to me. 👍 
@typoon, any thoughts from your end?

typoon:
Looks good to me. Maybe we should establish that when we are programmatically trying to execute a command, we should use `call_strings` and when we are dealing with user input we should use `execute`. Seems like the best way to approach this to me.

itaisod:
Thanks!

Issue #3775, opened 2020-01-06
Replace "Toggle Proxy" addon recommendation in client replay tutorial


aesyondu:
#### Problem Description
In https://github.com/mitmproxy/mitmproxy/blame/master/docs/src/content/tute-clientreplay.md#L35, Toggle Proxy is linked, but it no longer exists.

#### Steps to reproduce the behavior:
1. GOTO https://docs.mitmproxy.org/stable/tute-clientreplay/
2. GOTO https://addons.mozilla.org/en-us/firefox/addon/toggle-proxy-51740/
3. SEE Oops! We can’t find that page

#### System Information
DOCS

mhils:
Thanks for the report. I'd be happy for a PR that replaces it with something else, e.g. FoxyProxy.

globefire:
@mhils is this issue taken? If not can you please assign this issue to me?

diegotony:
@mhils  the issue was resolved?

ghunaaugest:
Thanks community

PR #3789, opened 2020-01-19
Fix broken link with link to similar addon. This fixes issue #3775.

mprencipe:
Foxyproxy is a commonly used Firefox addon with more than 180k users. Would this be a good replacement for Toggle Proxy?


mprencipe:
Why did the coverage test fail? Does it take into account also non-code files?

mhils:
Coverage seems to be a bit flaky at the moment, not sure why. I any case, thanks for taking care of this, I think FoxyProxy is a good alternative! 😃 🍰 🚀

PR #3791, opened 2020-01-20
Fix wording for ssl_insecure option on mitmweb

Eluch:
I made previous PRs to fix the broken checkbox on mitmweb ("Verify server certificates"):
image

It represents the popup option:
image

But I realized that this wording is inaccurate so I adjusted the wording for it.
It's now "Don't verify server certificates"


mhils:
Oops. Thanks for catching this! 🍰

Issue #3797, opened 2020-01-24
Unable to omit true value of custom bool option in CLI


marwinxxii:
#### Problem Description
When setting `true` value of custom addon option from CLI you cannot omit its value.

#### Steps to reproduce the behavior:
1. Create custom addon with bool option
2. run `mitmproxy -s myaddon.py --set bool_option`
3. See exception in log

```
error: Addon error: Traceback (most recent call last):
  File "…/lib/python3.7/site-packages/mitmproxy/addonmanager.py", line 42, in safecall
    yield
  File "…/lib/python3.7/site-packages/mitmproxy/addons/script.py", line 102, in loadscript
    ctx.master.addons.register(ns)
  File "…/lib/python3.7/site-packages/mitmproxy/addonmanager.py", line 162, in register
    self.master.options.process_deferred()
  File "…/lib/python3.7/site-packages/mitmproxy/optmanager.py", line 326, in process_deferred
    self.update(**update)
  File "…/lib/python3.7/site-packages/mitmproxy/optmanager.py", line 225, in update
    u = self.update_known(**kwargs)
  File "…/lib/python3.7/site-packages/mitmproxy/optmanager.py", line 216, in update_known
    self._options[k].set(v)
  File "…/lib/python3.7/site-packages/mitmproxy/optmanager.py", line 57, in set
    typecheck.check_option_type(self.name, value, self.typespec)
  File "…/lib/python3.7/site-packages/mitmproxy/utils/typecheck.py", line 74, in check_option_type
    raise e
TypeError: Expected <class 'bool'> for bool_option, but got <class 'NoneType'>.
```

While the same works for options in addons shipped with mitmproxy, e.g. `mitmproxy --set anticache`

#### System Information
```
Mitmproxy: 5.0.0 (+3, commit 010248a)
Python:    3.7.1
OpenSSL:   OpenSSL 1.1.0j  20 Nov 2018
Platform:  Darwin-19.0.0-x86_64-i386-64bit
```

#### Example
```python
import mitmproxy


class MyAddon:
    def load(self, loader):
        loader.add_option(
            name='bool_option',
            typespec=bool,
            default=False,
            help='Bool option'
        )

    def configure(self, options):
        if 'bool_option' in options:
            print(mitmproxy.ctx.options.bool_option)

addons = [MyAddon()]
```

PR #3798, opened 2020-01-25
Fix for issue 3797

typoon:
I believe this would fix the issue #3797


mhils:
Yes, this looks about right. Thanks! 🍰

PR #3811, opened 2020-02-06
http2: send all stream reset types to other connection

rohfle:
I am simulating interrupts during http2 responses for Wget2's test suite, while using mitmproxy to debug some of the tests that are not working. 

RST_STREAM frames were received by mitmproxy from the server, but were not sent to the client from mitmproxy. As this is inconsistent with what happens when the proxy is disabled, I have modified mitmproxy to send all stream resets from server to client (and probably client to server) to correct this.

### DIFF
--- a/mitmproxy/proxy/protocol/http2.py
+++ b/mitmproxy/proxy/protocol/http2.py
@@ -209,13 +209,12 @@ class Http2Layer(base.Layer):
     def _handle_stream_reset(self, eid, event, is_server, other_conn):
         if eid in self.streams:
             self.streams[eid].kill()
-            if event.error_code == h2.errors.ErrorCodes.CANCEL:
-                if is_server:
-                    other_stream_id = self.streams[eid].client_stream_id
-                else:
-                    other_stream_id = self.streams[eid].server_stream_id
-                if other_stream_id is not None:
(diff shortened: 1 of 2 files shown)


Kriechi:
This condition is there since basically day-0 of our h2 implementation:
https://github.com/mitmproxy/mitmproxy/commit/e9eed5e4c206e507665f5a4ff92654e969f4dd89#diff-f9c6498a2f07babd399fab7b9cc1fa16R256

I'm a bit hesitant to just change it without an explanation and test case.
(yes - I don't have a test to begin with, but I'm sure back then I had a good reason to add this check 😸)

rohfle:
I have added a test that checks all RST_STREAM error codes.... the test passes on my PR but fails on master with a TcpTimeout.

http2 error code documentation here: https://http2.github.io/http2-spec/#ErrorCodes
See also the stream state transition table (albeit an old one) here: http://html5.ohtsu.org/HTTP2_Stream_State.pdf

http2 has many different RST_STREAM codes - each code may affect the connection state, so in order to maintain valid state on the client end,  you must pass these codes through to the client. What happens on master now is the server will reset the stream, the proxy does not forward the reset, and the client will timeout. 

I'm not an expert at http2, but this is my understanding.

PR #3814, opened 2020-02-07
Update concepts-filters.md

2NuFyeat4PZt:
Provide a working example for the ~m filter

### DIFF
--- a/docs/src/content/concepts-filters.md
+++ b/docs/src/content/concepts-filters.md
@@ -49,3 +49,8 @@ Requests whose body contains the string "test":
 Anything but requests with a text/html content type:
 
     !(~q & ~t "text/html")
+
+Replace entire GET string in a request (quotes required to make it work):
+
+   ":~q ~m GET:.*:/replacement.html"
+   


mhils:
@Kriechi @2NuFyeat4PZt: Did you test this? Pretty sure the regex should be `^.*$` and not `.*`. On my system, this rewrites everything to `/replacement.html/replacement.html`.

```python
>>> re.subn(b".*$", b"/replacement.html", b"/", flags=re.DOTALL,count=0)
(b'/replacement.html/replacement.html', 2)
>>> re.subn(b"^.*$", b"/replacement.html", b"/", flags=re.DOTALL,count=0)
(b'/replacement.html', 1)
```

This really is a bit of an abuse of the replacement patterns, not sure how long it will survive. For example, this only works because 1) we don't run replacements on the body if it's empty and 2) the `Header.replace` implementation keeps the original header line if the returned replacement doesn't contain a colon. I'm fine with keeping in the documentation for now, but we may change the behavior here at some point to accomodate other use cases.

Kriechi:
@mhils I have to admit, I haven't actually tested this myself.
Should we open a bug ticket to keep track of this - or revert now?

2NuFyeat4PZt:
.*
^.*
^.*$
All work exactly the same on my system.

Given the aim is to replace the entire path being used by GET, specifying the start and end in the regex seem redundant to me; not sure why that would be considered an abuse?

mhils:
> Given the aim is to replace the entire path being used by GET, specifying the start and end in the regex seem redundant to me; not sure why that would be considered an abuse?

Replacements also apply to headers and body. For headers, it only works because we do a "if resultung replacement doesn't contain `:`, keep existing header as-is" check. For the body, it only works because we don't run body replacements if the body is empty. In other words, there's a lot of luck involved that the implemention is just like that. 😛 


@2NuFyeat4PZt: You get a different output for `import re; re.subn(b".*$", b"/replacement.html", b"/", flags=re.DOTALL,count=0)` ? The problem with .* is that it apparently matches twice on my system, once for "/" and once for "<end of string>". A bit weird to me as well, but apparently that's what it is.

PR #3826, opened 2020-02-23
Fix wrong behavior of --allow-hosts options

BlownSnail:
The "not" with "any" statement mistaken if the expect behavior of --allow-hosts is that only allow host matched any given regex.


mhils:
Oof - thanks!

mhils:
For the record, this really should have proper tests. I'll let this skip here as I'll do some changes to these on my current proxy core refactor. Thanks again! 🍰

Issue #3833, opened 2020-02-27
Headers replace bug when count is set


eremeevdev:
#### Problem Description
Headers matching for replace are fully removed, if count=1

#### Steps to reproduce the behavior:
1. create header instance: `headers = Headers(host='foo.com', referrer='foo.com')`
2. replace `foo` with `bar` and `count=1`: `headers.replace('foo', 'bar', count=1)`
3. result is empty headers: `Headers[]`

```
>>> headers = Headers(host='foo.com', referrer='foo.com')
>>> headers.replace('foo', 'bar', count=1)
1
>>> headers
Headers[]
```

#### System Information
Mitmproxy: 5.0.1
Python:    3.7.6
OpenSSL:   OpenSSL 1.1.0j  20 Nov 2018
Platform:  Darwin-19.3.0-x86_64-i386-64bit

mhils:
Thanks for providing super easy steps to reproduce this! :smiley:

naivekun:
I'll try to fix it.

PR #3839, opened 2020-02-28
Fix for issue 3833

naivekun:
#3833

### DIFF
--- a/test/mitmproxy/net/http/test_headers.py
+++ b/test/mitmproxy/net/http/test_headers.py
@@ -90,10 +90,6 @@ class TestHeaders:
         assert replacements == 1
         assert headers["Host"] == "barbarfoo.com"
         assert headers["Accept"] == "foo/bar"
-        replacements = headers.replace("foo", "bar", count=2)
-        assert replacements == 2
-        assert headers["Host"] == "barbarbar.com"
-        assert headers["Accept"] == "bar/bar"
 
 
 def test_parse_content_type():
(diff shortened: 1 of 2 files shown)


test/mitmproxy/net/http/test_headers.py
```diff
@@ -88,6 +88,12 @@ def test_replace_with_count(self):
         headers = Headers(Host="foobarfoo.com", Accept="foo/bar")
         replacements = headers.replace("foo", "bar", count=1)
         assert replacements == 1
+        assert headers["Host"] == "barbarfoo.com"
+        assert headers["Accept"] == "foo/bar"
+        replacements = headers.replace("foo", "bar", count=2)
+        assert replacements == 2
+        assert headers["Host"] == "barbarbar.com"
+        assert headers["Accept"] == "bar/bar"
```
mhils: This test does not really verify that we're doing the correct thing now, is it? It seems to pass with and without the patch.

naivekun: * Maybe because the test is committed ahead of my patch? should I open a new PR for test? or commit the test first to verify the issue #3833 and then commit the patch for the issue?
* line 91 and 92 actually make sense to the issue #3833. is the line 93 to 96 unnecessary?

mhils: Sorry, I got a bit confused here. In short:

1. Yes, this looks like we already had a test for this, but this test checked the wrong things (it just checked the replacement count). You fixed that correctly with lines 91/92.
2. 93-96 are just duplicating what is done before. That somehow confused me, but that's my fault. :) I'll take them out and we're good here.

PR #3846, opened 2020-03-05
Example for detecting/blocking DNS over HTTTPS queries

kimbo:
This is an example mitmproxy script to detect and block DNS queries over HTTPS. It generates a blocklist of DoH hostnames and IP addresses on startup, and also inspects request headers, paths, and query params that are often associated with DoH queries.

I wrote this for some research/testing I was doing and figured it might be useful to contribute here.


examples/complex/block_dns_over_https.py
```diff
@@ -0,0 +1,248 @@
...
+    https_url_re = re.compile(r'https://'
+                              r'(?P<hostname>[0-9a-zA-Z._~-]+)'
+                              r'(?P<port>:[0-9]+)?'
+                              r'(?P<path>[0-9a-zA-Z._~/-]+)?')
+
+    provider_re = re.compile(r'(\[([^\]]+)\]\(([^)]+))\)|(.*)')
+    # URLs that are not DoH URLs
+    do_not_include = ['my.nextdns.io', 'blog.cloudflare.com']
+    found_table = False
+    with urllib.request.urlopen('https://raw.githubusercontent.com/wiki/curl/curl/DNS-over-HTTPS.md') as fp:
```
Kriechi, line 50: Depending on an external website that is not under our control is very brittle.
Please add some error handling and maybe a fallback static list of a handful common providers so this example still works in a few months from now after that wiki page has changed significantly.

kimbo: > Depending on an external website that is not under our control is very brittle.
Very good point. I'll just replace all the wiki scraping and DNS lookups and what with a static list.

kimbo: Alright made the changes.

Kriechi:
Thanks! 🎉 👍

Issue #3558, opened 2019-06-04
Addon OptionsError is neither logged, nor does it stop mitmproxy


cventers:
##### Steps to reproduce the problem:

1. Following the master branch documentation here:

https://docs.mitmproxy.org/master/addons-options/

```
(venv) chase@turbotaz .../Stage/mitmproxy $ which mitmdump
/home/chase/Stage/mitmproxy/venv/bin/mitmdump
(venv) chase@turbotaz .../Stage/mitmproxy $ mitmdump -s ./examples/addons/options-configure.py --set addheader=1000
Loading script ./examples/addons/options-configure.py
Proxy server listening at http://*:8080
```

##### System information

Mitmproxy: 5.0.0.dev (+237, commit 8353f4a)
Python:    3.6.6
OpenSSL:   OpenSSL 1.1.0j  20 Nov 2018
Platform:  Linux-4.13.0-sabayon-x86_64-Intel-R-_Core-TM-_i5-6500_CPU_@_3.20GHz-with-gentoo-2.6

mhils:
This should be fixed in mitmproxy 8! 😃

PR #3849, opened 2020-03-07
Fix:Addon OptionsError is neither logged, nor does it stop mitmproxy

sarthak212:
#3558

### DIFF
--- a/test/mitmproxy/data/addonscripts/configure.py
+++ b/test/mitmproxy/data/addonscripts/configure.py
@@ -1,27 +1,21 @@
 import typing
 
-from mitmproxy import ctx
 from mitmproxy import exceptions
 
 
-class TestHeader:
+class OptionAddon:
     def load(self, loader):
         loader.add_option(
-            name = "testheader",
+            name = "optionaddon",
             typespec = typing.Optional[int],
             default = None,
-            help = "test header",
+            help = "Option Addon",
         )
 
     def configure(self, updates):
         raise exceptions.OptionsError("Options Error")
 
-    def response(self, flow):
-        if ctx.options.testheader is not None:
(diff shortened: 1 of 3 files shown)


sarthak212:
@mhils I think Configure in scripts is not raising any type of error.

mhils:
Thanks! Can we start by adding a test that demonstrates the wrong behavior?

sarthak212:
Yeah but should i write custom dummy script to pass as argument or i use prebuild script?

test/mitmproxy/data/addonscripts/configure.py
```diff
@@ -0,0 +1,27 @@
...
+        )
+
+    def configure(self, updates):
+        raise exceptions.OptionsError("Options Error")
+
+    def response(self, flow):
+        if ctx.options.testheader is not None:
+            flow.response.headers["testheader"] = str(ctx.options.testheader)
+
+
```
mhils: This really isn't needed here, let's remove it.

test/mitmproxy/data/addonscripts/configure.py
```diff
@@ -0,0 +1,27 @@
+import typing
+
+from mitmproxy import ctx
+from mitmproxy import exceptions
+
+
+class TestHeader:
```
mhils: Give this thing a better name. :)

sarthak212:
@mhils I have done changes.

test/mitmproxy/data/addonscripts/configure.py
sarthak212: I changed the name to "OptionAddon" if that doesn't fit can you give me a little hint, I am not too good with names.

mhils:
Great, thanks! 🍰

Issue #2894, opened 2018-02-22
Properly check for UTF8 console


luckydonald:
##### Steps to reproduce the problem:

1. `mitmproxy` and `mitmproxy --save-stream-file +/tmp/mitmdump`
2. select some event, enter
3. scroll down

```python
Traceback (most recent call last):
  File "/usr/local/Cellar/mitmproxy/3.0.0/libexec/lib/python3.6/site-packages/mitmproxy/tools/console/master.py", line 216, in run
    self.loop.run()
  File "/usr/local/Cellar/mitmproxy/3.0.0/libexec/lib/python3.6/site-packages/urwid/main_loop.py", line 286, in run
    self._run()
  File "/usr/local/Cellar/mitmproxy/3.0.0/libexec/lib/python3.6/site-packages/urwid/main_loop.py", line 384, in _run
    self.event_loop.run()
  File "/usr/local/Cellar/mitmproxy/3.0.0/libexec/lib/python3.6/site-packages/urwid/main_loop.py", line 788, in run
    self._loop()
  File "/usr/local/Cellar/mitmproxy/3.0.0/libexec/lib/python3.6/site-packages/urwid/main_loop.py", line 816, in _loop
    self._entering_idle()
  File "/usr/local/Cellar/mitmproxy/3.0.0/libexec/lib/python3.6/site-packages/urwid/main_loop.py", line 777, in _entering_idle
    callback()
  File "/usr/local/Cellar/mitmproxy/3.0.0/libexec/lib/python3.6/site-packages/urwid/main_loop.py", line 572, in entering_idle
    self.draw_screen()
  File "/usr/local/Cellar/mitmproxy/3.0.0/libexec/lib/python3.6/site-packages/urwid/main_loop.py", line 586, in draw_screen
    canvas = self._topmost_widget.render(self.screen_size, focus=True)
  File "/usr/local/Cellar/mitmproxy/3.0.0/libexec/lib/python3.6/site-packages/urwid/widget.py", line 144, in cached_render
    canv = fn(self, size, focus=focus)
  File "/usr/local/Cellar/mitmproxy/3.0.0/libexec/lib/python3.6/site-packages/urwid/container.py", line 1086, in render
    focus and self.focus_part == 'body')
  File "/usr/local/Cellar/mitmproxy/3.0.0/libexec/lib/python3.6/site-packages/urwid/widget.py", line 144, in cached_render
    canv = fn(self, size, focus=focus)
[...]

PR #3874, opened 2020-03-30
UTF 8 dependency

techathena:
I would be really helpful for newbies if docs had a clear "we need UTF encoding " in the installation section


mhils:
Thanks - I appreciate the effort, but I think we need to do proper UTF8 checks on startup instead and fail with a useful error message there.

refs #2894

PR #3875, opened 2020-03-30
c is deprecated

sarthak212:
In mitmproxy -c file command -c is replace with -C but docs still shows -c


mhils:
Awesome, thanks! 🍰

PR #3883, opened 2020-04-02
Proper check for UTF8 dependency 

techathena:
Added a test for UTF do proper UTF8 checks on startup and if check fails,it fails with a useful error message there.
as suggested here : https://github.com/mitmproxy/mitmproxy/pull/3874#issuecomment-607691135


Kriechi:
Thanks for the PR - unfortunately this won't do very much in the setup.py file.
I suppose this logic should go somewhere in the files in `mitmproxy/tools/`.

techathena:
Hey, 
Thanks for reviewing the PR! 
Can you tell me which linter is being used in the script, since there is no issue on my system.
Also should  add the logic in the _main.py ?

Issue #3457, opened 2019-01-25
allow mitmweb to export request as cURL command


baptx:
It looks like it is only possible to export a request as cURL command in mitmproxy command-line version and not in mitmweb.

hillz2:
+1

We really need this feature, please add this on mitmweb

Congee:
need it badly because I run mitmweb is a remote host. The export as curl feature in mitmproxy requires a system clipboard command that obviously does not exist in a headless linux box.

apuppy:
Really need it. Thanks.

Nowaker:
TOP2 most upvoted ticket... Anyone?

oxr463:
I started working on this...but I'm not good at the front end bits.

mhils:
I'm 99% confident we'll get there this summer. @gorogoroumaru is currently interning as our Google Summer of Code student on improving mitmweb, and we've been talking about adding this just the other day. The main complication here is that we first want to add a new API which allows the execution of arbitrary commands (https://docs.mitmproxy.org/dev/concepts-commands/) from mitmweb, and then have copy as curl/... use that. The mitmproxy 7 release coming in a few days will focus heavily on back-end stuff, over the summer we will focus heavily on mitmweb.


@oxr463: Please drop by on our Slack anytime to coordinate! We're always happy for helping hands, and there definitely are some low-hanging fruits here. :)

Nowaker:
Awesome. Thanks for sharing some inside knowledge, @mhils!

mhils:
This is fixed on main now, you can either install from there or use our development snapshots from https://mitmproxy.org/downloads/#branches/main/ 😃

lionel126:
I run into an issue what I copied is "[object Object]".
I installed with 2 ways below and same results:
1. pip install https://snapshots.mitmproxy.org/branches/main/mitmproxy-8.0.0.dev0-py3-none-any.whl
2. docker image: mitmproxy/mitmproxy:dev

> This is fixed on main now, you can either install from there or use our development snapshots from https://mitmproxy.org/downloads/#branches/main/ 😃

mhils:
Thanks @lionel126 - looks like I screwed this up last minute. 🙊 
#4779 has the fix, should be out later today.

mhils:
And fixed! :)

PR #3887, opened 2020-04-03
WIP: Export a request as cURL command

oxr463:
Bug: https://github.com/mitmproxy/mitmproxy/issues/3457


mhils:
This button only does the same as the Download button, which I think isn't too useful. Closing this here for now. :)

oxr463:
> 
> 
> This button only does the same as the Download button, which I think isn't too useful. Closing this here for now. :)

Yes, I was still trying to figure out how it all worked.

PR #3899, opened 2020-04-04
Don't force host header on outgoing requests

mhils:
It is not really clear why we did this, so we stop doing this and see who complains.

### DIFF
--- a/mitmproxy/net/http/http1/assemble.py
+++ b/mitmproxy/net/http/http1/assemble.py
@@ -1,4 +1,3 @@
-import mitmproxy.net.http.url
 from mitmproxy import exceptions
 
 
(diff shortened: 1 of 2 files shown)


Kriechi:
If I get asked in 5 years from now, why I approved this PR, I will deny everything!

wchasekelley:
Should i still be seeing the `Host:` header in flow views if it doesn't get output by `_assemble_request_headers`? I noticed that even if the `Host:` header is displayed in mitmproxy ui, it still doesn't show up in the raw export for a request and presumably the actual request as well.

Also it might be a good idea to always add the host header? From https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host
```text
    A Host header field must be sent in all HTTP/1.1 request messages. A 400 (Bad
    Request) status code may be sent to any HTTP/1.1 request message that lacks a
    Host header field or that contains more than one.
```

mhils:
> Also it might be a good idea to always add the host header? From https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host

Generally yes, but our use case for mitmproxy is a bit different: We want to pass-through requests _as-is_. It would be not very helpful if your requests work with your debugger (mitmproxy) on the line, but not without.

mhils:
Regarding your first question, we have a bunch of related changes on master. If you have inconsistencies there, please open a new issue :)

Issue #3886, opened 2020-04-03
Dark mode for mitmweb


oxr463:
Love the web ui, but the all white / really bright interface is not the best on the eyes in the evenings.

See also: https://postwoman.io

mhils:
Thanks for the suggestion! I'm happy to merge a PR that adds dark-mode CSS depending on system settings, but it's nothing I see one of the core team spend time on in at the moment.

Prinzhorn:
Just my 2 cents for anyone that wants to tackle this: I've made good experiences using CSS variables and `@media (prefers-color-scheme: dark)`. Optionally you can force an overwrite of the media query using a class added via JavaScript (and maybe persisted in `localStorage` for extra swag).

```css
:root {
  --text: #000;
  --background-default: #fff;
}

@media (prefers-color-scheme: dark) {
  :root {
    --text: #fff;
    --background-default: #222;
  }
}

body {
  color: var(--text);
  background: var(--background-default);
}
```

oxr463:
Trying to figure out how to toggle the theme. I'm missing something on the server-side, (See: https://github.com/mitmproxy/mitmproxy/issues/3925)

Prinzhorn:
> Trying to figure out how to toggle the theme. I'm missing something on the server-side, (See: #3925)

I personally don't think the server needs to be involved at all. Toggling is something that happens in the browser. If persistence is needed it can be done via `localStorage`. Or is there a use-case that requires server interaction that I'm missing?

oxr463:
How can I execute a function based on the `SETTINGS_UPDATE` action?

![image](https://user-images.githubusercontent.com/43783393/82447156-df835a00-9a75-11ea-9a28-f684aaea9838.png)

The `localStorage` part is fine, I can easily read/write values there. But I don't know where in [`js`](https://github.com/mitmproxy/mitmproxy/tree/master/web/src/js) to do that.

AClon314:
Share the css style, you can use **stylus** browser extension to apply that on http://127.0.0.1
```css
@-moz-document url-prefix("http://127.0.0.1") {
    @media (prefers-color-scheme: dark) {
        *:not(ul) {
            background: none !important;
        }
        :root {
            --text: #fff;
            --background-default: #000;
        }
        body,button,pre,a {
            color: var(--text) !important;
        }
        pre,
        body,
        header,
        .modal-content {
            background: var(--background-default) !important;
        }
        .selected {
            background-color-color: #333 !important;
[...]

PR #3903, opened 2020-04-05
WIP: Implement dark mode for mitmweb

oxr463:
Bug: https://github.com/mitmproxy/mitmproxy/issues/3886

Reference:

- https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

- https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage

### DIFF
--- a/mitmproxy/tools/web/webaddons.py
+++ b/mitmproxy/tools/web/webaddons.py
@@ -21,6 +21,10 @@ class WebAddon:
             "web_host", str, "127.0.0.1",
             "Web UI host."
         )
+        loader.add_option(
+            "web_dark_mode", bool, False,
+            "Toggle dark mode."
+        )
 
     def running(self):
         if hasattr(ctx.options, "web_open_browser") and ctx.options.web_open_browser:


oxr463:
```python
self.app = app.Application(
            self, self.options.web_dark_mode
        )
```

Adapted from: https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/tools/web/master.py#L46

mhils:
Thanks for the preliminary PR! I think I would be in favor of just respecting the browser's `prefers-color-scheme` setting, everything else seems overkill to me. Most crucially, I don't want to maintain additional dependencies and options.

oxr463:
@mhils will do.

oxr463:
**Bootstrap Variables**

```
@gray-darker:  lighten(#000, 13.5%); // #222
@gray-dark:    lighten(#000, 20%);   // #333
```

Source: `node_modules/bootstrap/less/variables.less`

oxr463:
See: https://stackoverflow.com/a/56757527/8507637

image

mhils:
Closing this here, see https://github.com/mitmproxy/mitmproxy/pull/3903#issuecomment-609464530. I'm fine with having CSS that auto-adjusts to the browser settings, but we don't need a separate setting within mitmproxy for this.

Issue #3911, opened 2020-04-07
Not able to bind one key for two different cmd under two different contexts


KevCui:
#### Problem Description

I'd like to bind `l` key as `console.nav.select` on `flowlist` and `console.nav.right` on `flowview`.

The `keys.yaml` looks like:

```
- 
  key: l
  ctx: ['flowlist']
  cmd: console.nav.select
- 
  key: l
  ctx: ['flowview']
  cmd: console.nav.right
```

However, mitmproxy seems to only get the latest keybinding in yaml file. In this case, `l` key won't perform `select` nav action on `flowlist`, only perform `right` nav action on `flowview`.

#### System Information

Mitmproxy: 5.0.1
Python:    3.8.2
OpenSSL:   OpenSSL 1.1.1d  10 Sep 2019
Platform:  Linux-5.5.9-arch1-2-x86_64-with-glibc2.2.5

mhils:
Fixed by #3912, thanks again!

PR #3912, opened 2020-04-07
#3911 Support keybinding for different contexts with different commands

KevCui:
When reading `keys.yaml` file for a specific key and context, instead of removing all contexts bound with this key, it will only overwrite the specific context. This change allows user to bind 1 key with different commands for different contexts.


mhils:
Awesome, thanks! 😃 🍰

Issue #3748, opened 2019-12-18
iOS 13.3: Certificate installation instructions need improvement


andig:
#### Problem Description
After installing profile and activating it, Safari still shows websites as "not trusted" and apps show "network error"

#### Steps to reproduce the behavior:
1. Configure wifi to use mitmproxy
2. Install and activate profile from mitm.it
3. Navigate anywhere in Safari

refs https://github.com/mitmproxy/mitmproxy/issues/3649 (added EKU) and #3647 

#### System Information
Paste the output of "mitmproxy --version" here.

```
Mitmproxy: 5.0.0
Python:    3.7.5
OpenSSL:   OpenSSL 1.1.1d  10 Sep 2019
Platform:  Darwin-19.2.0-x86_64-i386-64bit
```

mhils:
Yes, this should work in 5.0. Did you delete your ~/.mitmproxy folder and then reinstall the certificate on your device (https://docs.mitmproxy.org/stable/concepts-certificates/#installing-the-mitmproxy-ca-certificate-manually)?

andig:
> Did you delete your ~/.mitmproxy folder and then reinstall the certificate on your device (https://docs.mitmproxy.org/stable/concepts-certificates/#installing-the-mitmproxy-ca-certificate-manually)?

I have only started working with mitmproxy just now, so no old settings. I will try explicit certificate trust on top of the profile installation and see if that works.

ohld:
I finally found out how to run all of this.
### My settings

``` bash
mitmproxy --version
  Mitmproxy: 5.0.0
  Python:    3.7.5
  OpenSSL:   OpenSSL 1.1.1d  10 Sep 2019
  Platform:  Darwin-18.7.0-x86_64-i386-64bit
```

iPhone XS (current latest OS).

### Step-by-step
#### Macbook
1) `brew install mitmproxy`
2) run `mitmweb` for web interface
3) in new Terminal window run `ifconfig` to find out your laptop local IP address (probably starts with `192.168. ... `)

#### iPhone
1) Open settings -> Wi-Fi -> "your Wi-Fi network" -> Configure Proxy
2) Insert into "Server" your local laptop's IP and default port "8080"
3) In Safari open website 'mitm.it' and install sertificate
4) Open Settings -> General -> Profiles -> mitmproxy -> Verify the sertificate

And now everything will be `Trusted` and `Awesome`!

andig:
[...]

PR #3914, opened 2020-04-08
Reduce leaf certificate lifetime to one year

mhils:
Safari will, later this year, no longer accept new HTTPS certificates that expire more than 13 months from their creation date.

#3748 was a helpful reminder that we need to do this.

### DIFF
--- a/mitmproxy/certs.py
+++ b/mitmproxy/certs.py
@@ -15,8 +15,8 @@ import OpenSSL
 from mitmproxy.coretypes import serializable
 
 # Default expiry must not be too long: https://github.com/mitmproxy/mitmproxy/issues/815
-DEFAULT_EXP = 94608000  # = 24 * 60 * 60 * 365 * 3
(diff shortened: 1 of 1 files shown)


Kriechi:
How about doing 90 days - to match Let's Encrypt?

Prinzhorn:
I have a related question that crossed my mind the other day: I think mitmproxy caches certificates for a certain hostname to not generate a new one for every request to the same host. When will the cache invalidate? I didn't look through the code. I'm just wondering what would happen in theory if I run mitmproxy for a year straight or whatever (or 91 days).

mhils:
We currently don't detect if certs expire, so mitmproxy would serve outdated certs after running for 90 days. All browsers accept one year, so I think that's good for now.

mhils:
Race conditions 😄

mhils:
FWIW, the more practical issue is with root certs after three years, but anyways. Let's merge this and deal with that later. 😃

Prinzhorn:
> Race conditions smile

😂 branch prediction

PR #3915, opened 2020-04-08
Limit WebSocketFlow messages array to a fixed size

fascinated:
If mitmproxy/mitmdump is used with long-running (or very busy) websocket connections, memory usage grows continuously while a socket connection remains active. One approach could be to limit this array to a fixed size, so that only the last N messages are held in memory. I selected 1000 arbitrarily, but this could be a constant defined elsewhere.

It seems most of the user scripts interact with messages as they come in (vs. accessing the complete websocket message history), so this would likely not break any existing functionality.

### DIFF
--- a/mitmproxy/proxy/protocol/websocket.py
+++ b/mitmproxy/proxy/protocol/websocket.py
@@ -110,6 +110,7 @@ class WebSocketLayer(base.Layer):
             websocket_message = WebSocketMessage(message_type, not is_server, payload)
             length = len(websocket_message.content)
             self.flow.messages.append(websocket_message)
+            while(len(self.flow.messages)>1000): self.flow.messages.pop(0)
             self.channel.ask("websocket_message", self.flow)
 
             if not self.flow.stream and not websocket_message.killed:


mhils:
Hey @fascinated,

I appreciate the effort (and the idea isn't too bad either), but silently discarding data is out of question for us. If you have this rare use-case of long-running websocket connections that exceed your memory, you can always do this yourself in an addon until we come up with a better solution. 

refs #2196

fascinated:
That makes sense, and yeah, I am using this fix locally. Overall I was surprised to learn that even while using mitmdump, the entire persistent connection was stored in memory (vs. being partially discarded after processing or saving to disk as requested).

mhils:
Yes - we would like to change that eventually (see #2196), but keeping connections in memory until they are closed is just good enough for almost all HTTP use cases where response streaming is not activated.

Kriechi:
I would be open to having something like this as addon or example script - with good documentation.

fascinated:
@Kriechi that's a good idea, I'll try truncating flow.messages inside an addon script. Then I can add it to an existing websocket handler demo, or elsewhere.

Kriechi:
Maybe also add a CLI option for this addon: `--truncate-long-websockets=1000` messages.
I think we have a similar option for flows in general.

PR #3921, opened 2020-04-10
Example docs

Kriechi:
This PR makes the docs render all examples into a markdown page so they are browsable and accessible directly through our main docs page.


docs/build-current
```diff
@@ -8,5 +8,7 @@ for script in scripts/* ; do
   "$script" > "src/generated/${output%.*}.html"
 done
 
+python3 render_examples > src/content/addons-examples.md
```
mhils, line 11: This should probably live in scripts/ so that it is covered by lines 5-9.

Kriechi: I played around with this a bit - I couldn't get it to work. Maybe your Hugo knowledge can solve it: I'm generating a markdown file from my python script. Hugo then does the usual of transforming markdown into html, while running the shortcut functions / templating inline.

If I would make use of the scripts loop, then I would only get a html, which breaks the whole pipeline. Ideas?

Kriechi:
My main motivation is: why not have all examples also neatly integrated and accessible on our documentation. We already include a few selected examples - so why not have all of them.
This would save the user from going back and forth between Github and our docs.

Also - we never really mentioned the whole examples directory anywhere in the docs - this PR now  shows them in a prominent place.

Issue #1020, opened 2016-03-10
Add support for non-HTTP flows to the UI.


mhils:
We want to support other protocols than basic HTTP in the UI. In particular, we want to display HTTP2-specific content, WebSockets and raw tcp streams at some point.

The basic foundations for this are laid out: We have a generic Flow object (mitmproxy.models.flow) and a HTTPFlow that subclasses it (libmproxy.models.http). For TCP, we would introduce a TCPFlow object that stores a list of messages objects from client and server (in comparison, the HTTPFlow stores a distinct request and response). libmproxy.protocol.tcp would then create one TCPFlow per connection.

Each connection thread passes Flow objects to the FlowMaster (`handle_request`, `handle_response`, ...), which stores them in a FlowStore (see libmproxy.flow). The UI renders the store contents.

The high level implementation steps are as follows:

1. Set up TCPFlow class
2. Send TCPFlows to the FlowMaster (e.g. `handle_new_tcp_flow`, `handle_update_tcp_flow`)
3. Add printing to mitmdump.
4. Make `mitmdump -w` and `mitmdump -r` work.
5. Make the mitmproxy console work.

tomlandia:
Is anybody working on this? I'd be interested to do it this week as I've had a use case where it would have been very useful.

tomlandia:
Obviously it would only work in SOCKS mode and you'd need a command line flag to filter by port, but I still wonder if it should log flows through ALL or NO ports by default.

mhils:
@tomlandia: Feel free to join us on Slack if you have any questions. Proxy mode and port are solved problems (see http://docs.mitmproxy.org/en/latest/dev/protocols.html), the hard part is the UI side. In any case, if you tackle this, please do baby-step pull requests.

mhils:
1-4 are tackled in #1127. :tada:

cortesi:
This refactoring should happen first - #2196

jnsnow:
Does this issue cover concepts like the ability to MITM TLS/TCP protocols, and have those packets appear as flows? Or is that a separate kind of issue that would merit a new issue?

Presently it appears with --rawtcp or --tcp-hosts that you can convince mitmproxy to facilitate traffic it doesn't understand, but the data goes to the logging page where you cannot seemingly interact with it further. I have a use case right now where a custom protocol starts dumping zlib data into the stream, and I can't convince mitmproxy to write it out anywhere into a usable format where I could actually try to decompress it.

[...]

PR #3922, opened 2020-04-10
Display TCP flows at flow list!

nikitastupin:
Hi @mhils,

In this pull request I've implemented part of the 1st step of https://github.com/mitmproxy/mitmproxy/issues/1020#issuecomment-609906840. Right now it's not so fancy nor sexy but does the job:

 Screenshot 2020-04-10 at 20 57 09

There is no detail view yet. I'm planning to add it with another commit.

### DIFF
--- a/mitmproxy/addons/view.py
+++ b/mitmproxy/addons/view.py
@@ -22,6 +22,7 @@ from mitmproxy import connections
 from mitmproxy import ctx
 from mitmproxy import io
 from mitmproxy import http  # noqa
+from mitmproxy import tcp  # noqa
 
 # The underlying sorted list implementation expects the sort key to be stable
 # for the lifetime of the object. However, if we sort by size, for instance,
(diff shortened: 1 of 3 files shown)


mitmproxy/addons/view.py
```diff
@@ -69,30 +70,42 @@ def __call__(self, f):
 
 class OrderRequestStart(_OrderKey):
     def generate(self, f: http.HTTPFlow) -> int:
-        return f.request.timestamp_start or 0
+        if isinstance(f, http.HTTPFlow):
+            return f.request.timestamp_start or 0
+        else:
+            return f.timestamp_start
```
mhils, line 76: We probably have quite a few changes like this, so let's make sure that we do it a bit more systematic:

```python
    def generate(self, f: flow.Flow) -> int:  # note the updated type definition
        if isinstance(f, http.HTTPFlow):
            return f.request.timestamp_start or 0
        elif isinstance(f, tcp.TCPFlow):
            return f.timestamp_start
        else:
            raise NotImplementError(f)
```

This makes it hopefully more straightforward to add additional protocols. For example, your `OrderRequestMethod` would silently fail with "TCP" instead of complaining with an error.

mitmproxy/addons/view.py
```diff
@@ -69,30 +70,42 @@ def __call__(self, f):
...
+                s += len(f.request.raw_content)
+            if f.response and f.response.raw_content:
+                s += len(f.response.raw_content)
+        else:
+            s = 1337 # Stub
         return s
 
 
-matchall = flowfilter.parse(".")
+matchall = flowfilter.parse(". | ~tcp")
```
mhils, line 108: `~http | ~tcp`, but I realize that nothing is really finished here yet :)

nikitastupin:
Closed in flavor of #3927

PR #3927, opened 2020-04-11
TCP Flow UI

mhils:
This PR adds a very simple TCP Flow UI based on the fabulous work by @Jessonsotoventura and @nikitastupin. Thank you two!

- I've taken Nikita's PR and added the most important bits from Jesson's PR, as that seemed easier than removing the TCPViewEntry stuff in there. The end result is somewhere in the middle.
- After delving into the code I got a bit sidetracked with fixing the terrible mess of `format_flow` and its descendants¹. Two main changes here: First, I've removed the two_line/extended/layout convolution and replaced with it with a single RenderMode enum that signals in which context we're currently rendering. Second, I've removed the awful dict-passing code and replaced it all with explicit variables. This may seem verbose, but allows us to do proper typechecking, which makes the whole thing much less error-prone. Our flow list has been slightly improved in the process.
- There is still tons of work to be done, but it can now be tackled in smaller quantities.

@Jessonsotoventura, @nikitastupin, @mitmproxy/devs: Feedback welcome!
@Kriechi: This should land after 5.1 so that it receives some light testing on master before release.

closes #3922, closes #3680, refs #1020

¹) No blame here, it has just grown organically.

### DIFF
--- a/mitmproxy/addons/view.py
+++ b/mitmproxy/addons/view.py
@@ -21,8 +21,10 @@ from mitmproxy import command
 from mitmproxy import connections
 from mitmproxy import ctx
 from mitmproxy import io
-from mitmproxy import http  # noqa
-from mitmproxy import tcp  # noqa
+from mitmproxy import http
+from mitmproxy import tcp
+from mitmproxy.utils import human
+
 
 # The underlying sorted list implementation expects the sort key to be stable
 # for the lifetime of the object. However, if we sort by size, for instance,
@@ -39,7 +41,7 @@ class _OrderKey:
     def __init__(self, view):
         self.view = view
 
-    def generate(self, f: http.HTTPFlow) -> typing.Any:  # pragma: no cover
+    def generate(self, f: mitmproxy.flow.Flow) -> typing.Any:  # pragma: no cover
         pass
 
     def refresh(self, f):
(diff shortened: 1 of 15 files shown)


mitmproxy/addons/view.py
```diff
@@ -68,32 +71,49 @@ def __call__(self, f):
...
-        if f.request.raw_content:
-            s += len(f.request.raw_content)
-        if f.response and f.response.raw_content:
-            s += len(f.response.raw_content)
-        return s
-
+    def generate(self, f: mitmproxy.flow.Flow) -> int:
+        if isinstance(f, http.HTTPFlow):
+            size = 0
+            if f.request.raw_content:
```
Prinzhorn, line 102: In `clientplayback.py` you're using `if f.request.raw_content is None:`. Should this be consistent or is there a difference between the two? I'm just looking through this PR as I'm learning Python.

mhils: Good find, but it doesn't really matter here. :-)

.raw_content is of type Optional[bytes], which only evaluates to False if it is `None` or `b''`. We need to check for `None` because `len(None)` would rightly error, but it doesn't matter for `b''` as that would just add zero to size. I usually go for the shortest form in these cases, even though `is not None` would work just as well here.

Prinzhorn: Gotcha, so it's like `foo === undefined` vs relying on falsy values (`!foo`) in JavaScript and this catches both None and the empty value. I guess it's a matter of taste, I think I'm spoiled by using Go where everything is pretty explicit and self-explanatory (I mean Python probably beats JavaScript in this regard :smile: )

PR #3930, opened 2020-04-12
Towards v5.1

### LINKED ISSUE #3748: iOS 13.3: Certificate installation instructions need improvement

Kriechi:
* update docs with a getting-started page
* closes #3748 by updating and simplifying the iOS instructions
* drop Docker image for ARMv7 - the base image is deprecated, and ARMv7 is only the "old" RaspberryPis, which are becoming less and less popular.


mhils:
We have so far kept everyone from @mitmproxy/devs out of the changelog. Do we want to continue doing that, or shall we just name authors for everything? I'm really indifferent on that, just pointing out the change. :-)

Huge 👍 on the ARMv7 images, they greatly add onto CI time.

Kriechi:
Yes - I thought a bit about mentioning everyone or just non-core-team... I have no strong feelings either. I Looked at the changelog and it felt odd that a few items had names, and other didn't.
Ideally for the next release, I would like to try and keep the changelog "always up2date", instead of having to review 100s of PRs and commits before the next releases.

PR #3940, opened 2020-04-15
Map editor for GSoC idea 'Map Editor', includes code, test, document

naivekun:
Map Editor is now available to map local file. This PR includes code, test and document

### DIFF
--- a/mitmproxy/addons/mapeditor.py
+++ b/mitmproxy/addons/mapeditor.py
@@ -21,10 +21,18 @@ def parse_mapeditor(mexpr):
         )
     return parts[0], parts[1]
 
+# performance optimize
+# 0b01 for request
+# 0b10 for response
+mapchoice_option_mapper = {
+    "request": 0b01,
+    "response": 0b10,
+}
 
 class MapEditor:
     def __init__(self):
         self.map_list = []
+        self.map_choice = 0b10
 
     def load(self, loader):
         loader.add_option(
@@ -33,6 +41,18 @@ class MapEditor:
             map files like "pattern:MAP_TO:/path/to/local/file"
             all response body matched by 'pattern' will be replaced to the content of file
             """
+        )
+        loader.add_option(
+            "mapeditor_choice", str, "response",
+            """
+            select whether request or response or both to replace.
+            by default, mapeditor only replace response content
+            """,
+            choices=sorted([
+                "response",
+                "request",
+                "response + request",
+            ])
         )
 
     def configure(self, updated):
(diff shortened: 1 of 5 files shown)


naivekun:
Oops, the test file `/tmp/mapeditor_test` seems not work on windows.

Prinzhorn:
Hey, this looks great! I wanted to give my initial thoughts:

1. Instead of having a hard-coded `:MAP_TO:` keyword, could this be solved similar to `replacements` (like `/patt/file`)? So that you can define your own separator? I love my edge cases and I bet some day someone finds a URL in the wild that actually contains `:MAP_TO:` and they can't replace it :scream:  :smile: 
2. Could this be adjusted to also work with requests in addition to responses? It looks like `replacements` can do both and there are use-cases (e.g. when testing upload functionality).
3. I'm not sure about the implications of using `content` vs `raw_content` here. I assume the current implementation could have trouble with binary files? Maybe you can add a test case for binary and also move the test files to a "fixtures" folder instead of generating temporary files for each test run. If you continue to use temporary files it would be best practice to generate non-predictable file names. I think Python can do this for you, so you also don't have to worry about operating system differences https://docs.python.org/3/library/tempfile.html

Prinzhorn:
Disclaimer: I'm not affiliated with mitmproxy, so especially regarding number 3 you might want to wait for feedback from the maintainers. I don't know how fixtures in the test suite are handled.

naivekun:
Wow, learned a lot. Thanks for your suggestion. 
1. That's a problem. I initially thought `/` in path will break `/patt/path`. Now the best solution might be use the first char as the separator an document that user should check if path contains the separator.
2. That seems easy to implement, just do the same thing to the request is OK.
3. Thanks to your suggestion, binary file is actually a big problem. I firstly use this add-on to modify JS file which is normal ASCII text and did not noticed the binary file. I'll fix it later.

mitmproxy/addons/mapeditor.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class MapEditor:
+    def __init__(self):
+        self.map_list = []
+        self.map_choice = 0b10
+
+    def load(self, loader):
+        loader.add_option(
+            "mapeditor", typing.Sequence[str], [],
```
mhils, line 41: This should describe what it is doing, not how we name that component internally. How about `map_to_file` (or `map_local`)?

test/mitmproxy/addons/test_mapeditor.py
```diff
@@ -0,0 +1,193 @@
...
+        with taddons.context(me) as tctx:
+            with pytest.raises(Exception, match="Invalid map editor filter pattern"):
+                tctx.configure(me, mapeditor = ["~b:MAP_TO:/etc/hostname"])
+            tctx.configure(me, mapeditor = ["foo:MAP_TO:/etc/hostname"])
+
+    def test_mapeditor_matching(self):
+        me = mapeditor.MapEditor()
+        with taddons.context(me) as tctx:
+            # test matching rules
+            text_test_fp = os.path.join(tempfile.gettempdir(), os.urandom(24).hex())
```
mhils, line 29: pytest has a wonderful `tmp_path` fixture which also takes care of cleaning up: https://docs.pytest.org/en/latest/tmpdir.html

mitmproxy/addons/mapeditor.py
```diff
@@ -0,0 +1,95 @@
...
+                    "Invalid map editor filter pattern: %s" % fpatt
+                )
+            self.map_list.append((fpatt, path_to_file, flt))
+
+        m_options = ctx.options.mapeditor_choice.split(" + ")
+        self.map_choice = 0
+        for op in m_options:
+            self.map_choice |= mapchoice_option_mapper[op]
+
+    def run(self, f, flow):
```
mhils, line 77: Add type annotations and use better variable names. It's very unexpected that a variable that is either a HTTP request or a response is named `flow`.

test/mitmproxy/addons/test_mapeditor.py
mhils: On a second note: I don't think we need to create temp files at all, you can use have the resources in the test directory. See https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/test_certs.py#L151 for example.

Prinzhorn:
@mhils ooooh, I thought this was going in a completely different direction. Now I get it.

naivekun:
deprecated
see #3948

PR #3944, opened 2020-04-16
Fix docker smoke test

mhils:
@Kriechi: Docker CI doesn't run for external PRs because it is too slow and only tests the deploy, so if we work on that we need to use branches on mitmproxy/mitmproxy.


Kriechi:
oh - sorry!

Did you also mean to remove "mitmproxy/mitmproxy:"? does the `be.docker_tag` contain it already?

PR #3945, opened 2020-04-17
Improve replacements docs

Prinzhorn:
I thought #3940 was about using files to replace bodies, but it turns out `replacements` already does that. I completely missed this multiple times because the docs are not meant for scanning. But nobody actually reads walls of text and some sentences are incredibly long.

1. That `patt` is optional was completely missing
2. That `replacement` can be a `@file_path` is now completely obvious and right in your face without reading any of the text
3. I broke up the wall of text so that you don't have to actually read the thing if you're just looking for a particular thing

Now your brain can go pew pew "patt -> filter expression", nice gotcha. "replace -> string literal" neat. You get the idea.

### DIFF
--- a/docs/src/content/overview-features.md
+++ b/docs/src/content/overview-features.md
@@ -53,18 +53,25 @@ proxy mode.
 ## Replacements
 
 The `replacements` option lets you specify an arbitrary number of patterns that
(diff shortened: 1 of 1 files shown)


mhils:
refs #3948

Prinzhorn:
Nice, thanks!

PR #3950, opened 2020-04-21
Fix #3863 Use XDG base directory

### LINKED ISSUE #3863: Use XDG_CONFIG_HOME for the configuration
Currently mitmproxy stores the configuration in the `~/.mitmproxy` which is hard to manage and increases clutter in the home directory.

Use [FreeDesktop XDG directory recommendations](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html) which mean that the used directory is `$XDG_CONFIG_HOME/mitmproxy` where `XDG_CONFIG_HOME` defaults to the `$HOME/.config`. That makes everything much easier to manage. Of course for backwards compatibility it should fall back to current behaviour if the above directory is not found.

palb91:
Hi!
This morning, I found #3863 .
I already made some dirty changes on my side, I spend some time to reimplement it in a better way.
I also change some parts of the documentation to reflect the changes.
I'm not a Python developer in any way (although I like writing my tools with this language), so in my mind it's good, it works, but in the real world I don't know.

About the changes:
- `mitmproxy` check for an existing `~/.mitmproxy` first, to keep a compatibility with previous versions;
- If this directory does not exist, `mitmproxy` will use `$XDG_CONFIG_HOME/mitmproxy` that default to `~/.config/mitmproxy`.

I also thought about splitting configuration (`*.yaml`) and data (certificates), following the freedesktop recommendations, but if you consider this useless, I didn't want to spend time on it and let it in the configuration directory.


Prinzhorn:
Thanks for the PR.
Here are my thoughts. I'm not a maintainer and I'm also not a Python developer, but whatever:

1. You've duplicated the logic in multiple places. It should exist in a central place where the option is first normalized.
2. I think we should rather migrate an existing folder to the new location than supporting both in parallel. Or not? This makes the code easier because the migration is a separate code path and all other code can simply use a consistent location.
3. I don't have Windows at hand by I assume your changes will result in `~/.config/mitmproxy` on Windows (because of the fallback value for the env var) but it should probably stay as is? Or do we care? Does XDG_CONFIG_HOME apply to all platforms? I don't know.
4. The docs are not super clear, it says "either A or B" without explaining the difference. E.g. in `concepts-certificates.md `. However, I think the main issue is that with these changes we now have to differentiate between platforms?

> I also thought about splitting configuration (`*.yaml`) and data (certificates), following the freedesktop recommendations, but if you consider this useless, I didn't want to spend time on it and let it in the configuration directory.

I can't comment on the usefulness, but it's easier if a PR only solves one problem at a time.

palb91:
Hello, thank you for your interesting feedback.

> 1. You've duplicated the logic in multiple places. It should exist in a central place where the option is first normalized.

I followed what already existed, but it is repetitive for sure! Where can I factorize that? Is it relevant to do it in a `mitmproxy/utils/config_home.py` file for example?

> 2. I think we should rather migrate an existing folder to the new location than supporting both in parallel. Or not? This makes the code easier because the migration is a separate code path and all other code can simply use a consistent location.

I don't have a specific opinion on it, except that it would help to keep a clear documentation. I will wait for other input from maintainers, your « Or not? » doesn't help me :wink: 

> 3. I don't have Windows at hand by I assume your changes will result in `~/.config/mitmproxy` on Windows (because of the fallback value for the env var) but it should probably stay as is? Or do we care? Does XDG_CONFIG_HOME apply to all platforms? I don't know.

My bad, did not took Windows in account… I don't know how to handle that. There are equivalents, so I can use it as fallback (found this [link](https://stackoverflow.com/questions/43853548/xdg-basedir-directories-for-windows) where the suggestion is `%LOCALAPPDATA%`). I will need some Windows users feedback.

> 4. The docs are not super clear, it says "either A or B" without explaining the difference. E.g. in `concepts-certificates.md `. However, I think the main issue is that with these changes we now have to differentiate between platforms?

Yep, I knew I would have to clarify this part… however, an answer to your question 2) will give me an idea on how to proceed.

> > I also thought about splitting configuration (`*.yaml`) and data (certificates), following the freedesktop recommendations, but if you consider this useless, I didn't want to spend time on it and let it in the configuration directory.
> 
> I can't comment on the usefulness, but it's easier if a PR only solves one problem at a time.

Fine! Let's keep things simple.

mhils:
### Platform Consistency

Thanks. I think the current approach (~/.mitmproxy) isn't beautiful, but at least consistent across platforms. I'm open to switching to XDG on Linux, but then we also need to stay consistent for macOS and Windows and also use what's best practice there.

- Linux: `$XDG_CONFIG_HOME/mitmproxy`, fallback to `~/.config/mitmproxy` if not set.
- Windows: `%APPDATA%/mitmproxy`
- macOS: ??? @Kriechi?

### DRY

> I followed what already existed, but it is repetitive for sure! Where can I factorize that? Is it relevant to do it in a `mitmproxy/utils/config_home.py` file for example?

Yes, please do just that. Please expose a pathlib.Path while we're at it. :-)

### Migration

> * I think we should rather migrate an existing folder to the new location than supporting both in parallel. Or not? This makes the code easier because the migration is a separate code path and all other code can simply use a consistent location.

Yes, please. Let's kill it with fire and move ~/.mitmproxy to its new location on startup (while emitting a log message). 

### Docs

 > The docs are not super clear, it says "either A or B" without explaining the difference. E.g. in concepts-certificates.md . However, I think the main issue is that with these changes we now have to differentiate between platforms?


Probably the most important part. How about we add a section at the beginning of https://docs.mitmproxy.org/stable/concepts-options/ that looks as follows?

------

##### Configuration Directory
The mitmproxy configuration is stored in the following directory:
- **Windows:** ...
- **Linux**: ...
- **macOS**:

----

We can then link the term "configuration directory" at all other places in the docs to that section.

palb91:
Fine! I'll do my best to implement that. Let's improve my Python skills!

Prinzhorn:
> * Linux: `$XDG_CONFIG_HOME/mitmproxy`, fallback to `~/.config/mitmproxy` if not set.
> 
> * Windows: `%APPDATA%/mitmproxy`
> 
> * macOS: ??? @Kriechi?

Electron might not be authoritative, but I guess they have smart people that already thought about that
https://www.electronjs.org/docs/api/app#appgetpathname

> %APPDATA% on Windows
> $XDG_CONFIG_HOME or ~/.config on Linux
> ~/Library/Application Support on macOS

Kriechi:
IMHO on macOS we should use `~/.config`, same as on Linux.
If we consider mitmproxy a "terminal application", it feels more natural to behave like a Linux environment. `~/Library/Application Support` is mostly commonly used by proper macOS GUI apps.

palb91:
Thanks @Kriechi, so I will only differenciate Windows and the other.

In order to set the configuration path, I already have to use the `os` lib to check existing environment variables. Why should I use `pathlib` ? I don't get the advantages over `os`. I tried it, it works well but I don't understand the need here.

mhils:
pathlib is nice for two reasons. Most importantly, it's typed: We can clearly annotate paths in function signatures and don't have to go for generic strings. Second, pathlib code is often easier to read. Consider this example from `cibuild.py`:

```python3
os.path.normpath(os.path.join(os.path.dirname(__file__), ".."))
```
vs
```python3
Path(__file__).parent.resolve()
```

mhils:
Hi, any progress on this? 😃

palb91:
Hi! I'm sorry but no. I lost myself in python documentation, how to do the things correctly, sometimes just how to do the things… I also don't manage git well so I'm lost between new stuff and the few I already did.

I used to work on small python scripts that are standalone, it's hard to me to handle a big code base like that. I was not expecting to be blocked by such a small change (not that small after your comments ;) ), but the fact is… I'm blocked.

mhils:
No problem! Feel free to reopen this at any time, I'll close the PR for now.

palb91:
Sounds good. It is not for now as I'm quite busy, but do you know if it is possible to start a fresh PR while keeping all these comments ?
Does it work if I just delete my own branch and recreate with the same name?

mhils:
You can probably just force-push new commits to your branch. :)

palb91:
Sounds good, thanks :+1:

PR #3952, opened 2020-04-23
Update howto-wireshark-tls.md

PJF16:
Wireshark renamed the protocol from SSL to TLS in settings


mhils:
Not sure why CI is botched, but this is uncontroversial enough. 😄

Issue #3949, opened 2020-04-19
mitmproxy crashes when entering regex escape character in half-open string


liona24:
#### Problem Description
When adding a regex to the `replacements` option in interactive mode mitmproxy crashes when entering a leading `"` which is not yet closed followed by an escape character. 

#### Steps to reproduce the behavior:
1. Start `mitmproxy`
2. Type `: set replacements "/abc`
3. Now upon entering an escape character `\` mitmproxy crashes

It is working if you preemptively close the quotation marks, i.e. `: set replacements ""` and then continue entering your pattern within.

#### System Information

Crash Dump:
```
Traceback (most recent call last):
  File "/home/milk/.local/pipx/venvs/mitmproxy/lib/python3.6/site-packages/mitmproxy/master.py", line 86, in run_loop
    loop()
  File "/home/milk/.local/pipx/venvs/mitmproxy/lib/python3.6/site-packages/urwid/main_loop.py", line 286, in run
    self._run()
  File "/home/milk/.local/pipx/venvs/mitmproxy/lib/python3.6/site-packages/urwid/main_loop.py", line 384, in _run
    self.event_loop.run()
  File "/home/milk/.local/pipx/venvs/mitmproxy/lib/python3.6/site-packages/urwid/main_loop.py", line 1484, in run
    reraise(*exc_info)
  File "/home/milk/.local/pipx/venvs/mitmproxy/lib/python3.6/site-packages/urwid/compat.py", line 58, in reraise
    raise value
  File "/usr/lib/python3.6/asyncio/events.py", line 145, in _run
    self._callback(*self._args)
  File "/home/milk/.local/pipx/venvs/mitmproxy/lib/python3.6/site-packages/urwid/raw_display.py", line 404, in <lambda>
    event_loop, callback, self.get_available_raw_input())
  File "/home/milk/.local/pipx/venvs/mitmproxy/lib/python3.6/site-packages/urwid/raw_display.py", line 502, in parse_input
    callback(processed, processed_codes)
  File "/home/milk/.local/pipx/venvs/mitmproxy/lib/python3.6/site-packages/urwid/main_loop.py", line 411, in _update
    self.process_input(keys)
  File "/home/milk/.local/pipx/venvs/mitmproxy/lib/python3.6/site-packages/urwid/main_loop.py", line 511, in process_input
    k = self._topmost_widget.keypress(self.screen_size, k)
[...]

PR #3956, opened 2020-04-25
console error

sarthak212:
#3949


sarthak212:
`unquote` function remove quote when executing command So i think we can apply `?` mark to last quotation.Is there any problem with this approach @mhils

mhils:
Ok, now I think I understand the issue. We failed when we ended with an incomplete escape sequence before. @sarthak212: If you could add a test this can be merged. Thanks!

sarthak212:
Thanks @mhils I have done some changes, Update me on further changes.

mhils:
Awesome, thanks!

PR #3958, opened 2020-04-26
Change options and command execution order

sarthak212:
Some options in Mitmproxy only display information like --help, --version, --options, --commands in which help and version works fine in the situation when there is already mitmproxy server running and you try to run these command in seperate window but with --options and --commands is shows `Port already in use` error because there execution is done later in code after some operations which are not necessary for these commands.

### DIFF
--- a/mitmproxy/tools/_main.py
+++ b/mitmproxy/tools/_main.py
@@ -38,8 +38,7 @@ def process_options(parser, opts, args):
     if args.version:
         print(debug.dump_system_info())
         sys.exit(0)
-    if args.quiet or args.options or args.commands:
-        # also reduce log verbosity if --options or --commands is passed,
+    if args.quiet:
         # we don't want log messages from regular startup then.
         args.termlog_verbosity = 'error'
         args.flow_detail = 0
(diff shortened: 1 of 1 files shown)


Prinzhorn:
Nice catch, I never ran into this situation.
Maybe it makes sense to move it below here, since this is somewhat similar to `--version` (print+exit)?

https://github.com/mitmproxy/mitmproxy/blob/e9d5b6d9f590e514f27cc32abf001a694a3e8795/mitmproxy/tools/_main.py#L38-L40

I think this comment and the two parts in the if condition could then also be removed?

https://github.com/mitmproxy/mitmproxy/blob/e9d5b6d9f590e514f27cc32abf001a694a3e8795/mitmproxy/tools/_main.py#L41-L43

I don't think it's necessary if we exit the process right away anyway.

But let's wait and see if @mhils agrees.

sarthak212:
@mhils any review on this PR?

mhils:
This is fixed in 8.0 now, sorry for the slow turnaround! 🍰

PR #3961, opened 2020-04-28
added add-ons that enhance the performance of web application scanners.

anneborcherding:
Co-authored-by: weichweich <14820950+weichweich@users.noreply.github.com>

We added different add-ons that enhance the performance of web application scanners, especially if they are used against industrial control systems:

* **watchdog add-on**: This add-on checkts if the device under test still responds correctly to the scanner's requests. We used it in combination with an security testing framework and used it to automatically restart the device if the watchdog has been triggered. With this, web application scans can be finished even if the device crashes one or more times during the scan.
* **urlinjection add-on**: This add-on injects given URLs into a web page. With this, the web application scanner can be directed to sub-pages it wont find with its own crawling capabilities. As a result, the crawling capabilities of the web application scanner are improved. We call the URLs to be injected an URL index. It can be generated with the urlindex add-on.
* **urlindex add-on**: This add-on creates the URL index that can be used for the urlinjection add-on. The work flow could either be to explore the web page manually or to run different scanners while using mitmproxy with the urlindex add-on. With this, either the URLs visible during a manual exploration or the combined crawling capabilities of the used scanners are reflected in the generated URL index. Surely, combinations or other work flows are also possible.
* **mapping add-on**: This add-on sets contents of the web page to a fixed value. It is possible to map any css-selectable element to any given string. With this, many false positives of web application scanners can be avoided. The mapping information should be given in JSON format.


mhils:
Thanks for the PR! This looks like very well-written Python code, I'm happy to see that you seem to get tremendous use out of mitmproxy. 😃 Code-wise I really haven't much to complain about and test coverage also looks good. 👍 

However, I do feel your addons are too specialized to belong into mitmproxy's core. There definitely is substantial value for the subset of people using mitmproxy with web application scanners, but subjectively not enough for us to maintain 1,300 lines of code. I'd be happy to move this into [examples/complex](https://github.com/mitmproxy/mitmproxy/tree/master/examples/complex) instead, where we don't give any guarantees of maintaining it. In the long term, we hopefully reach a point where people can install addons as separate Python packages, which would also be ideal for this.

In any case, I really do appreciate you contributing your changes. If there's anything we can do on our end to make your life easier, please let us know!

anneborcherding:
Thanks for your fast reply! I would be happy if the code would be included in examples/complex! Would it be easier for you if I would move it there or if you would move it yourself?

mhils:
If you could move it here that would be great. Thanks! 🍰

anneborcherding:
I just moved it, I hope you can work with that! Please let me know if I can do anything more to make it easier for you! 👍

mhils:
Awesome, thanks! 🍰

PR #3970, opened 2020-05-01
Integrate contentviews to TCP flow details

nikitastupin:
Hey! Just did Step 2 of https://github.com/mitmproxy/mitmproxy/issues/1020#issuecomment-609906840. Here is how it looks like on a real MQTT connection.

Raw contentview:

Screenshot 2020-05-01 at 20 55 33

Hex contentview:

Screenshot 2020-05-01 at 20 55 43

### DIFF
--- a/mitmproxy/tools/console/flowview.py
+++ b/mitmproxy/tools/console/flowview.py
@@ -115,26 +115,21 @@ class FlowDetails(tabs.Tabs):
         if not flow.messages:
             return searchable.Searchable([urwid.Text(("highlight", "No messages."))])
 
-        from_client = None
-        messages = []
-        for message in flow.messages:
-            if message.from_client is not from_client:
-                messages.append(message.content)
-                from_client = message.from_client
-            else:
-                messages[-1] += message.content
+        viewmode = self.master.commands.call("console.flowview.mode")
 
-        from_client = flow.messages[0].from_client
         parts = []
-        for message in messages:
-            parts.append(
-                (
-                    "head" if from_client else "key",
-                    message
-                )
-            )
-            from_client = not from_client
-        return searchable.Searchable([urwid.Text(parts)])
+        for message in flow.messages:
+            _, lines, _ = contentviews.get_tcp_content_view(viewmode, message.content)
+
+            for line in lines:
+                if message.from_client:
+                    line.insert(0, "--> ")
+                else:
+                    line.insert(0, "<-- ")
+
+                parts.append(urwid.Text(line))
+
+        return searchable.Searchable(parts)
 
     def view_details(self):
         return flowdetailview.flowdetails(self.view, self.flow)
--- a/mitmproxy/tools/console/flowview.py
+++ b/mitmproxy/tools/console/flowview.py
@@ -157,7 +157,6 @@ class FlowDetails(tabs.Tabs):
                     "from_client": message.from_client,
                 })
 
-
         widget_lines = []
 
         widget_lines.append(self._contentview_status_bar(viewmode.capitalize(), viewmode))
--- a/mitmproxy/tools/console/flowview.py
+++ b/mitmproxy/tools/console/flowview.py
@@ -139,38 +139,32 @@ class FlowDetails(tabs.Tabs):
 
         # Merge adjacent TCP "messages". For detailed explanation of this code block see:
         # https://github.com/mitmproxy/mitmproxy/pull/3970/files/469bd32582f764f9a29607efa4f5b04bd87961fb#r418670880
-        merged_messages = []
+        from_client = None
+        messages = []
         for message in flow.messages:
-
-            if not merged_messages:
-                merged_messages.append({
-                    "content": message.content,
-                    "from_client": message.from_client,
-                })
-                continue
-
-            if merged_messages[-1]["from_client"] == message.from_client:
-                merged_messages[-1]["content"] += message.content
+            if message.from_client is not from_client:
+                messages.append(message.content)
+                from_client = message.from_client
             else:
-                merged_messages.append({
-                    "content": message.content,
-                    "from_client": message.from_client,
-                })
+                messages[-1] += message.content
 
         widget_lines = []
 
-        widget_lines.append(self._contentview_status_bar(viewmode.capitalize(), viewmode))
-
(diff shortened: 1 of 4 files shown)


mitmproxy/tools/console/flowview.py
```diff
@@ -115,26 +115,21 @@ def view_tcp_stream(self) -> urwid.Widget:
...
             return searchable.Searchable([urwid.Text(("highlight", "No messages."))])
 
-        from_client = None
-        messages = []
-        for message in flow.messages:
-            if message.from_client is not from_client:
-                messages.append(message.content)
-                from_client = message.from_client
-            else:
-                messages[-1] += message.content
```
mhils, line 125: The way we store TCP flow contents as "messages" is a bit disconnected from reality. In practice, TCP is just a bidirectional stream with no defined messages boundaries. A "message" in mitmproxy is essentially just "stuff we happened to read from the socket in a single syscall". Given this arbitrariness, I think it's correct to keep the code here which concatenates adjacent messages sent by the same side. Any objections?

mitmproxy/tools/console/flowview.py
```diff
@@ -115,26 +115,21 @@ def view_tcp_stream(self) -> urwid.Widget:
...
-                )
-            )
-            from_client = not from_client
-        return searchable.Searchable([urwid.Text(parts)])
+        for message in flow.messages:
+            _, lines, _ = contentviews.get_tcp_content_view(viewmode, message.content)
+
+            for line in lines:
+                if message.from_client:
+                    line.insert(0, "--> ")
```
mhils, line 126: How about we use something like `⇒` and ` ⇐`?

We can integrate this here:
https://github.com/mitmproxy/mitmproxy/blob/2774928319b706d8b6f85919f8811776342986c9/mitmproxy/tools/console/common.py#L100-L113

Also, it would be super awesome if we can color-code the arrows as blue and red. That helps a bit to determine message boundaries 😃 #3929 shows how you can define palette entries. I very hackily used "head" and "key" before.

mhils:
Some final thought: It would be very good to have a status line that display the current content view. :)

mitmproxy/tools/console/flowview.py
nikitastupin: At first I just didn't got the purpose of this lines after some time struggling to understand them I've tried to remove them and it worked fine. So I've came to conclusion that we can remove them 😄 

Your explanation definitely makes sense, no objections! I'll resurrect this code!

mitmproxy/tools/console/flowview.py
nikitastupin: This will make arrows more compact which will free space for an actual messages! Totally agree, thanks!

mitmproxy/tools/console/flowview.py
```diff
@@ -139,6 +159,9 @@ def view_tcp_stream(self) -> urwid.Widget:
 
 
         widget_lines = []
+
+        widget_lines.append(self._contentview_status_bar(viewmode.capitalize(), viewmode))
```
nikitastupin, line 162: I have doubts about `viewmode.capitalize()`. Originally there should be `description` of `_get_content_view` (e.g. current contentview's pretty-printed name or error an message if contentview can't be applied). However TCP has several messages as opposed to HTTP (where we have only one message) so it's kind of strange to print errors here because part of messages may parse correctly and part may parse incorrectly.

I propose to leave this stub for now and create an issue in which we'll discuss what's better to be displayed here. One of possible scenarios is to show how much messages (in percentage of total) produced a decode error.

@mhils, what do you think?

nikitastupin:
I've done some job on what we've discussed above. Now TCP flow details looks like this:

Screenshot 2020-05-02 at 13 10 44

mitmproxy/tools/console/flowview.py
nikitastupin: However I've decided to rewrite this part in a way that I believe is a little bit more simple for unfamiliar code reader 😃

nikitastupin:
https://github.com/mitmproxy/mitmproxy/pull/3970/checks?check_run_id=638574199#step:5:11 - So basically I don't know how to easily resolve this problem meanwhile don't see the reason for doing so. I suggest to ignore this one because if `viewmode` will be `None` after first line it'll be `auto` after `if not viewmode`:

```
viewmode = get(viewname)
if not viewmode:
    viewmode = get("auto")

description, lines, error = get_content_view(viewmode, data)
```

---

Also I'm not sure how to correctly type this one https://github.com/mitmproxy/mitmproxy/pull/3970/checks?check_run_id=638574199#step:5:12.

---

https://github.com/mitmproxy/mitmproxy/pull/3970/checks?check_run_id=638574199#step:5:13, https://github.com/mitmproxy/mitmproxy/pull/3970/checks?check_run_id=638574199#step:5:14 and https://github.com/mitmproxy/mitmproxy/pull/3970/checks?check_run_id=638574199#step:5:15 looks like false positives because `merged_messages` is not type of `TCPMessage`

---

I need your help @mhils 😃

mhils:
> https://github.com/mitmproxy/mitmproxy/pull/3970/checks?check_run_id=638574199#step:5:11

Yeah - the type system isn't adding much benefit here, other than pointing out that the whole architecture with `get` is a bit weird and should be refactored for good. The problem is that `get()` returns an `Optional[View]`. You can either add an assert statement in the function, add a `# type: ignore` comment. This is an example of a snippet where mypy does not provide any value, but we still need to fix it so that CI passes.

> https://github.com/mitmproxy/mitmproxy/pull/3970/checks?check_run_id=638574199#step:5:12.

You could either use a typing.NamedTuple instead of a dict, or revert to the old implementation while keeping your explanatory comment (it's a simple List[bytes] there). I personally prefer the old implementation, but that's of course subjective and I don't feel too strongly. I definitely do agree that it would have needed a comment FWIW. :-)

nikitastupin:
I've decided to revert to the old implementation of merging TCP "messages" because declaring `typing.NamedTuple` for this case seems like a little bit of overhead.

I suppose now we're ready to merge this and move to tackling interception and modification 🚀

mhils:
Awesome! 🎉

PR #3973, opened 2020-05-03
webaddons: Respect the user’s preferred web browser (if it’s graphical)

andersk:
The `open_browser` function used a hard-coded preference list to make sure we don’t open a terminal browser like `lynx` on a headless server, as this would suspend the process.  But that hard-coded list also made us open the wrong browser if, e.g., both Chrome and Firefox are installed and the user prefers Firefox.  Instead, always get the user’s preferred browser, then check that it’s a browser that will run in the background.

### DIFF
--- a/mitmproxy/tools/web/webaddons.py
+++ b/mitmproxy/tools/web/webaddons.py
@@ -35,7 +35,8 @@ class WebAddon:
 def open_browser(url: str) -> bool:
     """
     Open a URL in a browser window.
-    In contrast to webbrowser.open, we limit the list of suitable browsers.
+    In contrast to webbrowser.open, we only open a browser if it will run in
+    the background.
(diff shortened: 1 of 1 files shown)


andersk:
You’re right. I’ve rewritten the function using a whitelist; I check the type using `isinstance` rather than `b.name` because there are many fewer cases to consider that way.

mhils:
Thanks again and sorry for the slow turnaround! 🐌 
We had some WSL-related changes here recently which make the approach I proposed fail because we need to syntesize from `wslview %s`. A much simpler fix for your issue seems to be to just [include `x-www-browser` and `gnome-open`](https://github.com/mitmproxy/mitmproxy/blob/c339c4f82a4d1892a257c83651d159edcaa870af/mitmproxy/tools/web/webaddons.py#L49) in the list. Some initial testing seems to confirm tha this works as intended. Can you confirm?

Issue #3946, opened 2020-04-17
Use markdown code blocks for documentation


Prinzhorn:
**Is your feature request related to a problem? Please describe.**
The `{{< highlight none >}} :~q:foo:bar {{< / highlight >}}` blocks don't work on GitHub. It's confusing and makes the code examples almost useless on GitHub itself. And also in your code editor if you have Markdown syntax highlighting with embedded highlighting for nested code.

#### Describe the solution you'd like
It'd be neat if we could switch to using the standard triple backticks. I didn't search for it, I guess this has to do with how the website is generated from the markdown files. Can this be updated?

mhils:
We use Hugo for the website. It looks like this can be done by setting `pygmentsCodefences = true` in config.toml. I haven't looked more into it, but I'd be happy to merge a PR! Instructions for getting the docs up and running locally are at https://github.com/mitmproxy/mitmproxy/blob/master/docs/README.md.

Prinzhorn:
Nice. I won't jump onto that right now, so if anyone is interested feel free to look into it.

deepanshululla:
I would like to take this on.

deepanshululla:
is this all we need to do
```
baseURL = ""
languageCode = "en-us"
title = "mitmproxy.org docs"
theme = "mitmproxydocs"
publishDir = "../public"
RelativeURLs = true
googleAnalytics = "UA-4150636-13"
pygmentsCodefences = true
[indexes]
  tag = "tags"
```

deepanshululla:
Hi, I created a PR for this issue. I would appreciate a review here for the same. @mhils @Prinzhorn  Thanks a lot for your time.

PR #3977, opened 2020-05-05
enable flag pygmentsCodefences = true

deepanshululla:
As per the comments on 
https://github.com/mitmproxy/mitmproxy/issues/3946


mhils:
I think #3946 is more about converting the existing blocks to markdown syntax, but yeah this is the first step. Thanks!

General note: It's absolutely fine if you just open a PR and link to the issue, you don't need to leave an additional comment there. Not a big problem, but that helps a bit to cut down on noise. Thanks!

Issue #3972, opened 2020-05-02
Crash when opening TCP flow view


mhils:
#### Problem Description

```
Traceback (most recent call last):
  File "/mnt/c/Users/user/git/mitmproxy/mitmproxy/master.py", line 86, in run_loop
    loop()
  File "/home/user/venv/lib/python3.8/site-packages/urwid/main_loop.py", line 287, in run
    self._run()
  File "/home/user/venv/lib/python3.8/site-packages/urwid/main_loop.py", line 385, in _run
    self.event_loop.run()
  File "/home/user/venv/lib/python3.8/site-packages/urwid/main_loop.py", line 1494, in run
    reraise(*exc_info)
  File "/home/user/venv/lib/python3.8/site-packages/urwid/compat.py", line 58, in reraise                                                                                   [*:8080]
    raise value
  File "/usr/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/home/user/venv/lib/python3.8/site-packages/urwid/raw_display.py", line 416, in <lambda>
    wrapper = lambda: self.parse_input(
  File "/home/user/venv/lib/python3.8/site-packages/urwid/raw_display.py", line 515, in parse_input
    callback(processed, processed_codes)
  File "/home/user/venv/lib/python3.8/site-packages/urwid/main_loop.py", line 412, in _update
    self.process_input(keys)
  File "/home/user/venv/lib/python3.8/site-packages/urwid/main_loop.py", line 513, in process_input
    k = self._topmost_widget.keypress(self.screen_size, k)
  File "/mnt/c/Users/user/git/mitmproxy/mitmproxy/tools/console/window.py", line 312, in keypress
    return self.master.keymap.handle(
  File "/mnt/c/Users/user/git/mitmproxy/mitmproxy/tools/console/keymap.py", line 144, in handle
    return self.executor(b.command)
  File "/mnt/c/Users/user/git/mitmproxy/mitmproxy/tools/console/commandexecutor.py", line 18, in __call__
    ret = self.master.commands.execute(cmd)
  File "/mnt/c/Users/user/git/mitmproxy/mitmproxy/command.py", line 266, in execute
    return self.call_strings(command_name, args)
  File "/mnt/c/Users/user/git/mitmproxy/mitmproxy/command.py", line 252, in call_strings
    return self.commands[command_name].call(args)
[...]

PR #3980, opened 2020-05-05
fix console tab error

sarthak212:
#3972

### DIFF
--- a/mitmproxy/tools/console/tabs.py
+++ b/mitmproxy/tools/console/tabs.py
@@ -50,7 +50,7 @@ class Tabs(urwid.WidgetWrap):
         headers = []
         for i in range(len(self.tabs)):
             txt = self.tabs[i][0]()
-            if i == self.tab_offset or (len(self.tabs) <= self.tab_offset and i == 0):
+            if i == self.tab_offset % len(self.tabs):
                 headers.append(
                     Tab(
                         i,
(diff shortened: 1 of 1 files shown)


mitmproxy/tools/console/tabs.py
```diff
@@ -50,7 +50,7 @@ def show(self):
         headers = []
         for i in range(len(self.tabs)):
             txt = self.tabs[i][0]()
-            if i == self.tab_offset:
+            if i == self.tab_offset or (len(self.tabs) <= self.tab_offset and i == 0):
```
mhils, line 53: Style tip for this type of code: At the moment it's really hard to parse what is actually happening here. You can alleviate this by introducing a variable with a speaking name:
```python
is_focused = (
    i == self.tab_offset
    or (i == 0 and self.tab_offset >= len(self.tabs)):
)
if is_focused:
    # ...
```

I like that you don't modify `.tab_offset` if we have less tabs: That would be annoying when tabbing HTTP flow -> TCP flow -> HTTP flow as it would reset the offset. We could however simplify the implementation a bit and use `self.tab_offset % len(self.tabs)` everywhere.

mitmproxy/tools/console/tabs.py
```diff
@@ -70,7 +70,7 @@ def show(self):
                 )
         headers = urwid.Columns(headers, dividechars=1)
         self._w = urwid.Frame(
-            body = self.tabs[self.tab_offset][1](),
+            body = self.tabs[self.tab_offset if len(self.tabs) < self.tab_offset else 0][1](),
```
mhils, line 73: `<` or `<=`? (or `% len(self.tabs)` as mentioned above :))

mitmproxy/tools/console/tabs.py
sarthak212: Thanks for the tip

Kriechi:
@mhils good to go?

mhils:
Thanks! 🍰

PR #3985, opened 2020-05-08
Add a space after prefilled console command string

KevCui:
When opening command console by tapping a shortcut, there is no space after prefilled command string, the user has to type a `space` first then user input. For example:

- Type `w` in mitmproxy flow view, see console:

```
: save.file @shown█
```
(█ is current cursor position)

To enter the path `~/test`, user has to enter a `space` first:

```
: save.file @shown ~/test█
```
It's a bit inconvenient every time to type the additional `space`...

After this commit, a `space` is inserted after command string and the initial cursor position is:

```
: save.file @shown █
```


mhils:
I absolutely agree that this is a UX issue we need to fix. I think your patch is good enough for now, the long term solution should probably be that `console.command` accepts a single string as input (which then has a space at the end).

mhils:
Thanks!

PR #3989, opened 2020-05-08
Add minimal TCP interception and modification!

nikitastupin:
Hi team, 

In this pull request I've implemented MVP for Step 3 of https://github.com/mitmproxy/mitmproxy/issues/1020#issuecomment-609906840. Right now it allows to intercept and modify TCP messages. However I suppose that we need to add several cosmetic features to this pull request which will greatly increase usability:

- [x] Highlight intercepted TCP messages which are waiting for acceptance with `a`.
- [ ] Highlight modified TCP messages.
- [x] Remove unrelated list items from `e` menu (`cookies`, `url`, etc) for TCP flow details.
- [ ] Highlight currently selected TCP message.
- [ ] Show *all* intercepted TCP messages, not only one that is going to be accepted. Also user should be able to skip some intercepted messages and don't ever accept them.

@mhils , any additional features from your side? Feedback on above list of features will be helpful too.

P.S. Slowly but surely #1020 moves to an end 😃

### DIFF
--- a/mitmproxy/addons/intercept.py
+++ b/mitmproxy/addons/intercept.py
@@ -51,4 +51,5 @@ class Intercept:
         self.process_flow(f)
 
     def tcp_message(self, f):
-        f.intercept()
+        if self.filt:
+            f.intercept()
--- a/mitmproxy/addons/intercept.py
+++ b/mitmproxy/addons/intercept.py
@@ -49,3 +49,6 @@ class Intercept:
 
     def response(self, f):
         self.process_flow(f)
+
+    def tcp_message(self, f):
+        f.intercept()
--- a/mitmproxy/addons/intercept.py
+++ b/mitmproxy/addons/intercept.py
@@ -51,5 +51,5 @@ class Intercept:
         self.process_flow(f)
 
     def tcp_message(self, f):
-        if self.filt:
+        if self.filt and ctx.options.intercept_active:
             f.intercept()
--- a/mitmproxy/tools/console/consoleaddons.py
+++ b/mitmproxy/tools/console/consoleaddons.py
@@ -398,6 +398,7 @@ class ConsoleAddon:
             "status_code",
             "set-cookies",
             "url",
+            "tcp-message"
         ]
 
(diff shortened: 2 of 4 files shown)


mhils:
Awesome, thanks!

> Highlight intercepted TCP messages which are waiting for acceptance with a.

Yes, agreed.

> Highlight modified TCP messages.

We don't do that for HTTP flows either, so I think that's fine for now. I'd be open to consider this as a general addition, but let's keep this simple here. :)

> Remove unrelated list items from e menu (cookies, url, etc) for TCP flow details.

Yes. You _should_ be able to access the currently selected flow somehow and then return different options.

> Highlight currently selected TCP message.

Yeah, things can get very tricky very quickly here: Ideally we'd want to have a list of TCP messages, where the user can select individual messages and accept/drop/split/merge them. That requires tons of changes in lots of places though. Let's keep it very simple for now, and make only the last message editable/interceptable. At the moment, the whole connection stalls when intercepting a single message anyways. :)

nikitastupin:
So I've added highlighting of intercepted TCP messages and cleaned up edit options for TCP flows. I suppose now it's ready to become a pull request rather than draft!

nikitastupin:
Hey @mhils, I'm happy to hear you again! :)

I'll need some time to get into code coverage topic because I didn't really work with it before. If you have some examples and / or documentation about how it's implemented in `mitmproxy` project I'd like to see those!

mhils:
Sure thing! 😃 In short, we have CI in place which checks that the tests in `test/mitmproxy/addons/foo` cover 100% of the lines in `mitmproxy/addons/foo`. You can do that check yourself by running `tox -e individual_coverage`. It's rightfully complaining about your additions to the intercept addon [here](https://github.com/mitmproxy/mitmproxy/pull/3989/checks?check_run_id=721230088#step:5:25). :)

nikitastupin:
Thanks for the explanation!

Actually this was easier than I thought! Also it uncovered one bug 😄

mhils:
Awesome, thanks for the quick turnaround! 🍰

PR #3996, opened 2020-05-12
First step to fix #848 by adding a relax_http_form_validation option.

### LINKED ISSUE #848: Requesting complete URL over https proxy results in "Invalid request scheme: https"
1. Set up an HTTPS mitmproxy
2. Run the following (minimized) code against it:

``` python
import httplib
conn = httplib.HTTPSConnection('<your proxy IP>',<your proxy port>)
conn.set_tunnel('www.google.com',443)
conn.request('GET','https://www.google.com:443/')
resp = conn.getresponse()
print resp.read()
```

``` html
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
<A HREF="http://www.google.com/">here</A>.
</BODY></HTML>
```
##### What went wrong?

``` html
<html>
            <head>
                <title>400 Bad Request</title>
            </head>
            <body>HttpException('Invalid request scheme: https',)</body>
        </html>
```
##### Any other comments?

I believe it's caused by a combination of https://github.com/mitmproxy/mitmproxy/blob/master/libmproxy/protocol/http.py#L544-L545 and https://github.com/mitmproxy/mitmproxy/blob/master/libmproxy/protocol/http.py#L502-L503. Indeed, if you modify my `GET` request above to ask for just `/` instead of `https://www.google.com:443/`, everything works fine.

I'm not sure if it's sensible to support absolute-form requests over https tunnels, but I'm getting this behavior in real code (boto for python AWS) so I don't think it's very a contrived situation.

---

mitmproxy version: 0.14.0

MrNbaYoh:
Add an option to bypass the call to `validate_request_form` when activated, as suggested in #848.
It's probably not enough to fix the whole issue, but it was enough to solve my issue and I guess that even though it's incomplete, adding a flag disabled by default is okay.

### DIFF
--- a/mitmproxy/options.py
+++ b/mitmproxy/options.py
@@ -181,5 +181,11 @@ class Options(optmanager.OptManager):
             TLS key size for certificates and CA.
             """
         )
+        self.add_option(
+            "relax_http_form_validation", bool, False,
+            """
+            Disable HTTP form validation.
+            """
+        )
 
         self.update(**kwargs)
--- a/mitmproxy/proxy/protocol/http.py
+++ b/mitmproxy/proxy/protocol/http.py
@@ -267,10 +267,12 @@ class HttpLayer(base.Layer):
                     self.send_error_response(400, msg)
                     return False
 
-            validate_request_form(self.mode, request)
+            if not self.config.options.relax_http_form_validation:
+                validate_request_form(self.mode, request)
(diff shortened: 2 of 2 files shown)


mhils:
Thanks! 🍰

PR #4000, opened 2020-05-16
Fix #3975 by enabling inline HTML in Markdown docs

### LINKED ISSUE #3975: [DOC]  https://docs.mitmproxy.org/stable/concepts-filters page doesn't show "View Flow Selectors" section correctly
Markdown is right here (https://github.com/mitmproxy/mitmproxy/blob/c16b4887875bcdecdf18355f28347a48ba4297fd/docs/src/content/concepts-filters.md) but when rendered to HTML, this section is completely gone. 
Thanks!

thunghan:
Setting `unsafe=true` as suggested.

### DIFF
--- a/docs/src/config.toml
+++ b/docs/src/config.toml
@@ -9,3 +9,8 @@ pygmentsCodefences = true
 
 [indexes]
   tag = "tags"
+
+[markup]
+  [markup.goldmark]
+    [markup.goldmark.renderer]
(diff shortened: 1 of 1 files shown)


mhils:
Awesome thanks! Not sure why GitHub is not registering this properly, but this has been merged: https://github.com/mitmproxy/mitmproxy/commits/master 🍰

PR #4009, opened 2020-05-20
Commands, keybindings and FlowList extended for halfpage scrolling (like ctrl-u and ctrl-d in vim)

lusceu:


lusceu:
Oh sorry, i submitted the pull request via hub and missed the description. 

Which functionality is added with this pull request ? 
I love the vim shortcuts (e.g. hjkl) in the console interface of mitmproxy. I've seen, that pagewise "scrolling" via "ctrl+b" and "ctrl+f" is also implemented for the flowlist-console ui.
However as a vim user, i'm used to scroll big documents with the shortcuts "ctrl+u" and "ctrl+d", which scrolls a half page. 
In my personal opinion, half page scrolling is much more readable for the end user.
So this pull requests aims to implement half page scrolling for the flowlist-ui via "ctrl+u" and "ctrl+d".

mhils:
Thanks! Let me try to understand this better - are you specifically looking for a _half_ page up? or are your primarily looking to keep some flows in the view for context? I think we could consider changing the behavior of page up/down to keep one row in view.

lusceu:
Hi @mhils , 
in my opinion, *half* page scrolling allows a much easier orientation in case there are a lot of flows or content to check. So i would prefer the half page solution, just like scrolling big code listings in VIM, that's where i borrowed the idea from. 

In the end, it's probably a matter of taste, which may be the reason why VIM has integrated shortcuts for both half page / page wise scrolling.
Thank you for asking!

mhils:
Thanks. I really appreciate the PR, but this is probably too niche for us to maintain. I think there would be a point in keeping a row in view (similar to page up/down in Chrome, for example), but I see very few people actually benefiting from the added complexity of half page scrolling. Sorry!

PR #4023, opened 2020-06-05
HTTP2 response reason is None, render as '' in property.

rbdixon:
Fixes an error triggered when displaying an HTTP2 response loaded from a file.

Using this script saved as `reason_none.py`:

```
from mitmproxy.io import protobuf

def response(flow):
    repr(protobuf.loads(protobuf.dumps(flow)))
```

Run:

```
$ mitmdump --save-stream-file google.flow
<in another terminal>
$ curl -k --http2 --proxy localhost:8080 https://google.com
```

Crashes:

```
$ mitmdump --rfile google.flow --script reason_none.py 
Loading script reason_none.py
[::1]:51048: GET https://google.com/ HTTP/2.0
          << 301  220b
  File "reason_none.py", line 4, in response
    repr(protobuf.loads(protobuf.dumps(flow)))
  File "/Users/rbdixon/src/mitmproxy/mitmproxy/http.py", line 174, in __repr__
    return s.format(flow=self)
  File "/Users/rbdixon/src/mitmproxy/mitmproxy/net/http/response.py", line 66, in __repr__
    reason=self.reason,
  File "/Users/rbdixon/src/mitmproxy/mitmproxy/net/http/response.py", line 128, in reason
    return self.data.reason.decode("ISO-8859-1", "surrogateescape")
AttributeError: 'NoneType' object has no attribute 'decode'
```

### DIFF
--- a/mitmproxy/net/http/response.py
+++ b/mitmproxy/net/http/response.py
@@ -122,10 +122,14 @@ class Response(message.Message):
     def reason(self):
         """
         HTTP Reason Phrase, e.g. "Not Found".
-        This is always :py:obj:`None` for HTTP2 requests, because HTTP2 responses do not contain a reason phrase.
+        HTTP2 responses do not contain a reason phrase and self.data.reason will be :py:obj:`None`.
+        When :py:obj:`None` return an empty reason phrase so that functions expecting a string work properly.
         """
         # Encoding: http://stackoverflow.com/a/16674906/934719
-        return self.data.reason.decode("ISO-8859-1", "surrogateescape")
+        if self.data.reason is not None:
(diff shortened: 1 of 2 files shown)


Kriechi:
In general this makes sense!
Please update the comment block just above your change to reflect that it now rightfully always returns a string, and in HTTP/2 it is an empty string.

It would be nice if you could add a super simple test for both HTTP/1 and /2.

rbdixon:
Thank you for taking a look. I'll update the PR shortly.

rbdixon:
ok, I think that does it. Let me know if I can help with any other changes. Thank you!

Kriechi:
Perfect - thanks! 🎉 👍

PR #4041, opened 2020-06-25
Added OpenSSL.crypto.X509.get_version()

Ronophobia:


mhils:
Hi, may I ask what's the purpose of this addition? We don't use `.version` in mitmproxy code, so that just means more maintenance cost for us.

mhils:
wrong button, oops.

mhils:
Closing this because of inactivity.

Issue #3948, opened 2020-04-17
Unify Replacement UX


mhils:
#### Problem Description

The replacement addon currently calls `flow.replace()`, which performs replacements in URL, HTTP Headers, and request/response body. This is messy and terrible because there is no way to control in which part of the flow replacements should be done - it does many things, and none of them well. It's often luck that replacements work. For example, in many cases, replacements like `/~s/response-body` "work" because URL/header replacements are not performed if they don't match syntactically:

https://github.com/mitmproxy/mitmproxy/blob/3131160e1c53a55845bb824038e602c14d92cf74/mitmproxy/net/http/headers.py#L172-L177

I think there is some opportunity for improvement here. 😄 

#### Proposed Changes

With the advent of a "map local" editor in #3940, I think it is time to clean up things here and implement the following changes:

1. The replacements addon is for HTTP body replacements (and later on TCP/WebSocket messages). We can possibly rename it to make this more clear (ReplaceContent? ReplaceData?). This also means we can implement it in the addon and get rid of `HTTPFlow.replace` and its descendants.
2. `SetHeaders` is for header modifications.
3. Map remote/map local addons are for request line modification/redirection to localhost.

This also means we can unify the expression syntax for these four addons:

|              | old syntax                            | new syntax                            |
|--------------|---------------------------------------|---------------------------------------|
| Map Local    |                                       | `[/filter]/url/file-or-directory`     |
| Map Remote   |                                       | `[/filter]/url-regex/replacement`     |
| SetHeader    | `/filter/name/value`                  | `[/filter]/header-name/[@]header-value`             |
| Replacements | `/filter[/regex]/[@]replacement` | `[/filter]/body-regex/[@]replacement` |

@mitmproxy/devs, @Prinzhorn, @naivekun, any thoughts? 😃

Prinzhorn:
I'm all for making stuff more consistent and well-defined. I didn't know about the _interesting_ `replace` semantics as I've only worked with addons so far to modify `content` etc.

I'm not saying what Burp does is perfect or even good, but it's worth taking a look at

![Selection_611](https://user-images.githubusercontent.com/679144/79575131-9e3c0b00-80c1-11ea-8cc8-2a4d0954e028.png)
![Selection_612](https://user-images.githubusercontent.com/679144/79575142-a005ce80-80c1-11ea-8cc5-e47f5a7fffe7.png)

[...]

PR #4043, opened 2020-06-25
Improve SetHeaders addon

mplattner:
This PR improves the `SetHeaders` addon with 3 changes:

1) The flow-filter is now the last parameter in the configuration pattern, as it is optional.
2) The user can now remove existing headers by specifying an empty `header-value`.
3) Updated docs to reflect changes.

refs #3948

### DIFF
--- a/docs/src/content/overview-features.md
+++ b/docs/src/content/overview-features.md
@@ -131,11 +131,11 @@ then the respective recorded responses are simply replayed by mitmproxy.
 Otherwise, the unmatched requests is forwarded to the upstream server. If
 forwarding is not desired, you can use the --kill (-k) switch to prevent that.
 
-## Set Headers
+## Modify Headers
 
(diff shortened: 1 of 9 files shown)


mhils:
Looks good to me! Just changing the parameter order is a bit of an ugly silent change. Shall we rename it to `ModifyHeaders` right away? I think there was some consensus on that. :)

mhils:
Other than that, LGTM! 🍰 😃 

Final note, you can add `refs #XXX` in your PR description so that you don't need to comment in that thread. :)

mplattner:
Thanks. Fully agree that this silent breaking change is not ideal.
Good idea, I'll rename it. (Didn't do it as I thought `SetHeaders` is fine without explicit `addheader`, `removeheader` etc. options.
Cool, will use `refs #X` for future PRs.

PR #4046, opened 2020-06-26
Replace replacements addon with new modify body addon

mplattner:
The new addon only replaces data in the body using `--modify-body`.

refs #3948

### DIFF
--- a/docs/src/content/overview-features.md
+++ b/docs/src/content/overview-features.md
@@ -88,17 +88,16 @@ New headers can be added, and existing headers can be overwritten or removed.
 A `modify_headers` expression looks like this:
 
 {{< highlight none  >}}
-/name/value[/filter-expression]
+[/patt]/name/value
 {{< / highlight >}}
 
-Here, **name** and **value** are the header name and the value to set respectively,
-e.g., ``/Host/example.org``. An empty **value** removes existing headers with
-**name**, e.g., ``/Host/``. The optional **filter-expression** is a mitmproxy
-[filter expression]({{< relref "concepts-filters">}}) that defines
-which flows to modify headers on, e.g., only on responses using ``~s``.
-Existing headers are overwritten by default.
-This can be changed using filter-expressions, e.g., ``!~h Host:`` to ignore
-requests and responses with an existing ``Host`` header.
+Here, **patt** is a mitmproxy [filter expression]({{< relref "concepts-filters">}})
+that defines which flows to modify headers on, e.g., only on responses using ``~s``.
+The parameters **name** and **value** are the header name and the value to set
+respectively, e.g., ``/Host/example.org``. An empty **value** removes existing
+headers with **name**, e.g., ``/Host/``. Existing headers are overwritten by
+default. This can be changed using filter-expressions, e.g., ``!~h Host:`` to
+ignore requests and responses with an existing ``Host`` header.
 
 
 ## Proxy Authentication
--- a/docs/src/content/overview-features.md
+++ b/docs/src/content/overview-features.md
@@ -11,10 +11,10 @@ menu:
 
 - [Anticache](#anticache)
 - [Client-side replay](#client-side-replay)
+- [Modify Body](#modify-body)
+- [Modify Headers](#modify-headers)
 - [Proxy Authentication](#proxy-authentication)
-- [Replacements](#replacements)
 - [Server-side replay](#server-side-replay)
-- [Set Headers](#set-headers)
 - [Sticky Auth](#sticky-auth)
 - [Sticky Cookies](#sticky-cookies)
 - [Streaming](#streaming)
@@ -41,30 +41,22 @@ conversation, where requests may have been made concurrently.
 You may want to use client-side replay in conjunction with the `anticache`
 option, to make sure the server responds with complete data.
 
-## Proxy Authentication
 
-Asks the user for authentication before they are permitted to use the proxy.
-Authentication headers are stripped from the flows, so they are not passed to
-upstream servers. For now, only HTTP Basic authentication is supported. The
-proxy auth options are not compatible with the transparent, socks or reverse
-proxy mode.
+## Modify Body
 
-
-## Replacements
-
-The `replacements` option lets you specify an arbitrary number of patterns that
-define text replacements within flows. A replacement pattern looks like this:
+The `modify-body` option lets you specify an arbitrary number of patterns that
+define replacements within bodies of flows. A replacement pattern looks like this:
 
 {{< highlight none  >}}
-/patt/regex/replacement
+[/patt]/regex/replacement
 {{< / highlight >}}
 
-Here, **patt** is a mitmproxy filter expression that defines which flows a
-replacement applies to, **regex** is a valid Python regular expression that
-defines what gets replaced, and **replacement** is a string literal that is
-substituted in. The separator is arbitrary, and defined by the first character.
-If the replacement string literal starts with `@`, it is treated as a file path
-from which the replacement is read.
+Here, **patt** is a mitmproxy [filter expression]({{< relref "concepts-filters">}})
+that defines which flows a replacement applies to, **regex** is a valid Python
+regular expression that defines what gets replaced, and **replacement** is a string
+literal that is substituted in. The separator is arbitrary, and defined by the
+first character. If the replacement string literal starts with `@`, it is treated
+as a file path from which the replacement is read.
 
 Replace hooks fire when either a client request or a server response is
 received. Only the matching flow component is affected: so, for example,
@@ -76,7 +68,7 @@ to create a script using the replacement API on Flow components.
 
 ### Examples
 
-Replace `foo` with `bar` in requests:
+Replace `foo` with `bar` in bodies of requests:
 
 {{< highlight none  >}}
 :~q:foo:bar
@@ -85,8 +77,37 @@ Replace `foo` with `bar` in requests:
 Replace `foo` with the data read from `~/xss-exploit`:
 
 {{< highlight bash  >}}
-mitmdump --replacements :~q:foo:@~/xss-exploit
+mitmdump --modify-body :~q:foo:@~/xss-exploit
 {{< / highlight >}}
+
+
+## Modify Headers
+
+The `modify_headers` option lets you specify a set of headers to be modified.
+New headers can be added, and existing headers can be overwritten or removed.
+A `modify_headers` expression looks like this:
+
+{{< highlight none  >}}
+/name/value[/filter-expression]
+{{< / highlight >}}
+
+Here, **name** and **value** are the header name and the value to set respectively,
+e.g., ``/Host/example.org``. An empty **value** removes existing headers with
+**name**, e.g., ``/Host/``. The optional **filter-expression** is a mitmproxy
+[filter expression]({{< relref "concepts-filters">}}) that defines
+which flows to modify headers on, e.g., only on responses using ``~s``.
+Existing headers are overwritten by default.
+This can be changed using filter-expressions, e.g., ``!~h Host:`` to ignore
+requests and responses with an existing ``Host`` header.
+
+
+## Proxy Authentication
+
+Asks the user for authentication before they are permitted to use the proxy.
+Authentication headers are stripped from the flows, so they are not passed to
+upstream servers. For now, only HTTP Basic authentication is supported. The
+proxy auth options are not compatible with the transparent, socks or reverse
+proxy mode.
 
 
 ## Server-side replay
@@ -131,24 +152,6 @@ then the respective recorded responses are simply replayed by mitmproxy.
 Otherwise, the unmatched requests is forwarded to the upstream server. If
 forwarding is not desired, you can use the --kill (-k) switch to prevent that.
 
-## Modify Headers
-
-The `modify_headers` option lets you specify a set of headers to be modified.
-New headers can be added, and existing headers can be overwritten or removed.
-A `modify_headers` expression looks like this:
-
-{{< highlight none  >}}
-/name/value[/filter-expression]
-{{< / highlight >}}
-
-Here, **name** and **value** are the header name and the value to set respectively,
-e.g., ``/Host/example.org``. An empty **value** removes existing headers with
-**name**, e.g., ``/Host/``. The optional **filter-expression** is a mitmproxy
-[filter expression]({{< relref "concepts-filters">}}) that defines
-which flows to modify headers on, e.g., only on responses using ``~s``.
-Existing headers are overwritten by default.
-This can be changed using filter-expressions, e.g., ``!~h Host:`` to ignore
-requests and responses with an existing ``Host`` header.
 
 ## Sticky auth
 
@@ -158,6 +161,7 @@ seen. This is enough to allow you to access a server resource using HTTP Basic
 authentication through the proxy. Note that <span
 data-role="program">mitmproxy</span> doesn't (yet) support replay of HTTP Digest
 authentication.
+
 
 ## Sticky cookies
 
@@ -175,6 +179,7 @@ Sticky cookies are especially powerful when used in conjunction with [client
 replay]({{< relref "#client-side-replay" >}}) - you can record the
 authentication process once, and simply replay it on startup every time you need
 to interact with the secured resources.
+
 
 ## Streaming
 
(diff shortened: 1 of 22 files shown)


mplattner:
Forgot to update the docs. Will update the PR soon.

mitmproxy/utils/arg_check.py
```diff
@@ -129,11 +131,15 @@ def check():
...
+            if isinstance(REPLACEMENTS.get(option), list):
+                new_options = REPLACEMENTS.get(option)
+            else:
+                new_options = [REPLACEMENTS.get(option)]
             print(
                 "{} is deprecated.\n"
                 "Please use `{}` instead.".format(
                     option,
-                    REPLACEMENTS.get(option)
+                    "` or `".join(new_options)
```
mhils, line 142: This fails for the existing options, see e.g. `" or ".join("--intercept")`.

mitmproxy/addons/modifybody.py
```diff
@@ -0,0 +1,134 @@
...
+import re
+import typing
+
+from mitmproxy import exceptions
+from mitmproxy import flowfilter
+from mitmproxy import ctx
+from mitmproxy.utils import strutils
+
+
+def parse_modify_body(s):
```
mhils, line 11: This looks suspiciously similar to `parse_modify_headers` now. We can probably just merge those two?

 - We can raise a plain `ValueError` here and re-raise an OptionsError in each addon with an individual error message.
 - Let's also include `strutils.escaped_str_to_bytes()` and `flowfilter.parse` here.

mitmproxy/utils/arg_check.py
mplattner: If there is only a single option as `str` in `REPLACEMENTS`, the string is put into a list in line 137.
I've tested it (now again) and it does work. Am I missing something?

```
❯ mitmproxy -i
usage: mitmproxy [options]
mitmproxy: error: unrecognized arguments: -i

-i is deprecated.
Please use `--intercept` instead.
```

mitmproxy/addons/modifybody.py
mplattner: Good idea. What's the preferred approach here? Have the function in one addon file and import it in the other?

Regarding the bullet points:
1) You mean a `ValueError` in the parse method and an `OptionsError` in the `Configure` method? Does that make sense here, i.e., having 2 subsequently thrown exceptions for the same error? (Trying to learn here, not to correct you :))
2) These methods are now applied in `replace()`. You suggest to move them to the parse method, right?

mitmproxy/utils/arg_check.py
mhils: Yeah, I can't read.

mitmproxy/addons/modifybody.py
mhils: > Have the function in one addon file and import it in the other?

yup, let's keep it simple. :)

> You mean a ValueError in the parse method and an OptionsError in the Configure method? Does that make sense here, i.e., having 2 subsequently thrown exceptions for the same error? (Trying to learn here, not to correct you :))

Yes. The very Pythonic way would be to re-raise as follows:

```python
def foo():
    raise ValueError("qux")

def bar():
    try:
        foo()
    except ValueError as e:
        raise BarError("cannot foo() today") from e
```

This ensure the following: 1) You can make sure that bar() only raises BarErrors. 2) You get a very nice chained traceback: 

```
  File "<pyshell#55>", line 3, in bar
    foo()
  File "<pyshell#49>", line 2, in foo
    raise ValueError("qux")
ValueError: qux

The above exception was the direct cause of the following exception:

  File "<pyshell#56>", line 1, in <module>
    bar()
  File "<pyshell#55>", line 5, in bar
    raise BarError("cannot foo() today") from e
BarError: cannot foo() today
```

> These methods are now applied in replace(). You suggest to move them to the parse method, right?

Yes. Reading from file should stay inside replace() however, it can be very useful that file contents are not cached. :)

docs/src/content/overview-features.md
```diff
@@ -41,30 +41,22 @@ conversation, where requests may have been made concurrently.
...
-proxy auth options are not compatible with the transparent, socks or reverse
-proxy mode.
-
 
-## Replacements
+## Modify Body
 
-The `replacements` option lets you specify an arbitrary number of patterns that
-define text replacements within flows. A replacement pattern looks like this:
+The `modify-body` option lets you specify an arbitrary number of patterns that
```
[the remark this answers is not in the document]
mplattner: `modify-body` is correct. `argparse` replaces `_` with `-`.

mitmproxy/addons/modifybody.py
mplattner: Done. Please check.

I haven't moved `flowfilter.parse` to the `parse_modify_hook` method as then the thrown `ValueError` is ambigous, i.e., when executing `parse_modify_hook` again in the `configure` method to re-raise the exception, we'd have to check the error message to determine whether the `ValueError` was caused by an invalid "/.../.../..." option or by an invalid filter.
Hope this makes sense.

test/mitmproxy/addons/test_modifyheaders.py
```diff
@@ -7,19 +7,21 @@
...
+            modifyheaders.parse_modify_hook("/")
 
     def test_configure(self):
         sh = modifyheaders.ModifyHeaders()
         with taddons.context(sh) as tctx:
+            with pytest.raises(Exception, match="Invalid modify_headers option"):
+                tctx.configure(sh, modify_headers = ["/"])
             with pytest.raises(Exception, match="Invalid modify_headers flow filter"):
                 tctx.configure(sh, modify_headers = ["/~b/one/two"])
             tctx.configure(sh, modify_headers = ["/foo/bar/voing"])
```
mplattner, line 27: Does this line do anything useful?

docs/src/content/overview-features.md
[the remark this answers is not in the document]
mhils: We auto-translate `_` to `-` for command line arguments, yes. However, `modify_body` is what you would set in your config.yaml or in mitmproxy's options editor, and what we consistently use in our docs (see https://docs.mitmproxy.org/stable/concepts-options/). You could also do `--set modify_body=...` for that matter. The inconsistency is a bit ugly, but there really aren't many better options around.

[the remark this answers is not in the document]
mhils: (if you have any ideas on how we could have this more consistent, I'd love to hear them - what we currently have is the best solution we came up with so far)

test/mitmproxy/addons/test_modifybody.py
```diff
@@ -1,25 +1,28 @@
...
+        assert x == ("foo", b"bar", b"vo/ing/")
+        x = parse_modify_hook("/bar/voing")
+        assert x == (".*", b"bar", b"voing")
+        with pytest.raises(Exception, match="Invalid modify_\\* specifier"):
+            parse_modify_hook("/")
 
     def test_configure(self):
         mb = modifybody.ModifyBody()
         with taddons.context(mb) as tctx:
             tctx.configure(mb, modify_body=["one/two/three"])
```
mplattner, line 23: Does this line do anything useful? Same for the last line of this test.

mhils: Generally speaking, having this line here asserts that it runs without error. In an ideal world, you would probably also check that the addon has parsed it properly, but "doesn't error" can sometimes be good enough.

mitmproxy/addons/modifybody.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+
+    def configure(self, updated):
+        """
+            .modify_body is a list of tuples (flow_filter_pattern, regex, repl):
+
+            flow_filter_pattern: a string specifying a flow filter pattern.
+            regex: a regular expression, as string.
+            repl: the replacement string
+        """
```
mhils, line 32: Let me be a bit pendantic here - this comment doesn't make too much sense to me:

 - `.modify_body` would indicate that there's a modify_body attribute. As far as I can see, there isn't?
 - If we want to document the type of `.lst` (that thing could probably be renamed, yes...), it would make sense to do this at the definition and not here. I'll add some ideas on how we can do this better in the next comment.

test/mitmproxy/addons/test_modifybody.py
mplattner: Makes sense, thanks.

mitmproxy/addons/modifyheaders.py
```diff
@@ -2,28 +2,29 @@
...
 from mitmproxy import flowfilter
+from mitmproxy.utils import strutils
 from mitmproxy import ctx
 
 
-def parse_modify_headers(s):
+def parse_modify_hook(s):
     """
-        Returns a (header_name, header_value, flow_filter) tuple.
+        Returns a (flow_filter, header_name, header_value) tuple.
```
mhils, line 11: If we re-use this (which is good 👍), we should probably signal somehow that it's not just headers. We can also add a bit more type information to make things easier to grasp:

```python
class ModifySpec(typing.NamedTuple):
    matches: flowfilter.TFilter,
    match_str: str
    key: bytes  # those could potentially have better names if you can think of some
    replacement: bytes


def parse_modify_spec(s: str) -> ModifySpec:
    # ...
    filter = flowfilter.parse(filter_str)
    return ModifySpec(filter, filter_str, header_name, header_value)
```

then, below:
```python
class ModifyHeaders:
    def __init__(self):
        replacements: List[ModifySpec] = []    # replaces self.lst
    
    # ...
    
    def run(self, f, hdrs):
        for spec in self.replacements:
            if spec.matches(f):
                hdrs.pop(spec.key, None)
```

I realize this is just a very rough sketch. Does it make sense anyway?

docs/src/content/overview-features.md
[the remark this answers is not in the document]
mplattner: Thanks for clarification.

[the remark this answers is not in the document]
mplattner: It might make sense to inform the user when underscore-formatted options are used as a cli argument, similar to the messages on deprecated commands.
At the moment it outputs: `mitmproxy: error: unrecognized arguments: --modify_body asdf`.

mitmproxy/addons/modifybody.py
```diff
@@ -0,0 +1,98 @@
...
+        if "modify_body" in updated:
+            lst = []
+            for rep in ctx.options.modify_body:
+                try:
+                    flow_filter_pattern, regex, repl = parse_modify_hook(rep)
+                except ValueError as e:
+                    raise exceptions.OptionsError(
+                        "Invalid modify_body option: %s" % rep
```
mhils, line 40: Make this `f"Cannot parse modify_body option {rep}: {e}"` to include the ValueError's message, then move the checks below into `parse_modify_hook` and make their messages generic (`"Invalid modify_body flow filter: %s" % flow_filter_pattern` -> `Invalid filter pattern: {pattern}`). This should solve the problem mentioned in https://github.com/mitmproxy/mitmproxy/pull/4046#discussion_r447724841, or am I misunderstanding things? :)

docs/src/content/overview-features.md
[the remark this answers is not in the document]
mhils: Agreed, that's actually a really good point! Let's track this in #4054.

mitmproxy/addons/modifybody.py
mplattner: I agree. This comment existed in the existing `Replace` addon and I didn't change it properly.

mitmproxy/addons/modifyheaders.py
mplattner: Great idea, thanks 🙌

mitmproxy/addons/modifybody.py
mplattner: Yep I guess this makes sense :)

mitmproxy/addons/modifybody.py
```diff
@@ -23,62 +22,42 @@ def load(self, loader):
...
         if "modify_body" in updated:
-            lst = []
-            for rep in ctx.options.modify_body:
+            self.replacements = []
+            for option in ctx.options.modify_body:
                 try:
-                    flow_filter_pattern, regex, repl = parse_modify_hook(rep)
+                    spec = parse_modify_spec(option)
+                    try:
+                        # We should ideally escape here before trying to compile
```
mplattner, line 31: I'm not sure what should be escaped here. Comments seems to be from you, @mhils

mhils:
Awesome! Next stop: map remote editor 🎉

mhils:
(@Kriechi's draconian individual coverage checks are still complaining, but that should be a very easy fix)

mplattner:
Test coverage is fixed.
What's left are failing unit tests on Windows for both `ModifyHeader` and `ModifyBody`. I guess the conversion to `bytes` causes some trouble with the path separator `\` on Windows.

Just did some testing in ipython:
```
In [40]: s="C:\\asd\\bsd"

In [41]: print(codecs.escape_decode(s)[0])
b'C:\x07sd\x08sd'

In [42]: print(codecs.escape_decode(s)[0].decode("utf-8"))
C:ssd
```

Why do we convert the given strings to bytes at all? Is that even reliably possible without assuming an encoding, e.g., UTF8?

mplattner:
@mhils, in fe662b4 my VSCode markdown extension messed up the table of contents of the docs page.
For such a minor error, can I force push to the branch and overwrite the existing commit, or should I restrain from force pushing at all?

mhils:
🎉 🍰

robjtede:
The 5.2 minor release broke scripts using Headers.replace. What's the suggested replacement?

mhils:
@robjtede: What were you using `Headers.replace` for?

robjtede:
```python
import re

from mitmproxy import http
from mitmproxy.http import HTTPFlow, HTTPResponse


def add_same_site_none(match):
    no_ss = match.group(0).decode('utf8')
    no_ss = re.sub('[ ;]+SameSite=\w+', '', no_ss)
    no_ss = re.sub('[ ;]+Secure', '', no_ss)
    return bytes(f'{no_ss}; SameSite=None', 'utf8')


def response(flow: HTTPFlow):
    headers = flow.response.headers
    headers.replace('Set-Cookie: (.*)', add_same_site_none)
    headers.replace('set-cookie: (.*)', add_same_site_none)

```

mhils:
The short story is that `Headers.replace` did some weird stuff internally and we found actually very few use-cases for it.

For example, you could just do the following instead:

```python
if "set-cookie" in headers:
    headers["set-cookie"] = re.sub(headers["set-cookie"], add_same_site_none)
```

`Headers` is case-insensitive, so that already covers both `Set-Cookie` and `set-cookie`. The assignment also retains case.

robjtede:
Ok, that fixes that one. Thank you. 🎉 

The other breaking change I'm trying to figure out a replacement for is the removal of the -R flag
```bash
-R "|~hs Set-Cookie|$ENDPOINT|localhost"
```

changing it to --modify-headers doesnt seem to apply the transformation as it did in 5.1
```bash
--modify-headers "|~hs Set-Cookie|$ENDPOINT|localhost"
```

mhils:
Right - @mplattner split out and expanded our replacement options into modify-headers, modify-body, map-local and map-remote for the 5.2 release ([docs](https://docs.mitmproxy.org/stable/overview-features/#map-local)). There should also be a blog post on this in the next few days. 😃 

One of the few things we dropped with this release were Header string replacements, simply because we couldn't come up with a lot of good use-cases. The new API allows you to add, overwrite, or remote headers only. I do think your example is a very valid one. If we find that this is a highly requested feature, we may add it back again.

For now, my recommendation would be to write a simple addon for this as well. If you do need the source the endpoint from the command line, you can have your addons [define their own options](https://github.com/mitmproxy/mitmproxy/blob/master/examples/addons/options-simple.py).

robjtede:
Okay I shall try to follow your advice.

I do feel that a breaking change like this should not have been made on a point release. I don't know if Python ecosystem follows semver but where I come from (JS/Rust) it's very much expected.

Just some honest feedback from a guy with colleagues telling me my script doesn't work.

mplattner:
hey @robjtede, sorry for the hiccup that this breaking change caused. We'll try to do better next time.

robjtede:
FWIW I really like the new addon system and was able to recreate the tool in a couple hours.

PR #4056, opened 2020-07-01
Update changelog and add PR template

Kriechi:
as discussed with @mhils, we should keep the changelog up2date at all times to avoid this daunting task before cutting a release.

This PR adds all recent PRs since the last release to the changelog.
This PR adds a PR template to instruct users to add new changelog entries directly into their newly sent PRs.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,7 +1,8 @@
-This PR fixes #<issue-numer>.
+#### Description
 
-This PR introduces a new feature to ...
+<!-- describe your changes here -->
 
-**Please add a new entry for the changes in this PR to the CHANGELOG in the unreleased section at the top!**
+#### Checklist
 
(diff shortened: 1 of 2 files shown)


mhils:
I'm a bit torn on this to be honest. This is nice for a slightly easier release process, but it also adds more friction and iteration to PRs. We can try this, but I feel like it's a PITA to have PRs which are ready to merge but miss a changelog entry. Is going through the commit log once really that bad?

Kriechi:
I'm not planning on forcing this - but my hope is, that if it's part of the template, users will do it. If not, then it should be on us maintainers to either add a new commit to the PR directly, or after merging, we should immediately push a new commit to master to add the entry.

Yes, IMO going through 500+ commits before a release to write the changelog is a heavy task, which at least for me, creates a high mental threshold on actually cutting a release!

mhils:
Fair enough. I've reduced the word count a bit, does it work for you this way?

PR #4058, opened 2020-07-02
List was not cycled right for tab auto-completion

Vane11ope:
#### Description

Options was not cycled right for tab auto-completion. Using tabs and shift-tabs mixed up when you are in cut.clip for instance. You would see the options just don't go back right when you press shift-tab, and it wouldn't go forward right for some occasion caused by the same bug. 

The code before did not even care if tab or shift tab was pressed and set the previously calculated position as the new position of the options, then calculate the next position after it. 

So fix was fairly simple. Check if tab or shift tab was pressed and calculate the new position according to it. A little difficult part was how to deal with the first time the tab or shift tab was pressed, but it could be handled by setting sys.maxsize as the initial value for the position.

Probably the code itself would explain you better :) 

#### Checklist

### DIFF
--- a/mitmproxy/tools/console/commander/commander.py
+++ b/mitmproxy/tools/console/commander/commander.py
@@ -28,15 +28,17 @@ class ListCompleter(Completer):
             if o.startswith(start):
                 self.options.append(o)
         self.options.sort()
-        self.offset = 0
+        self.pos = None
 
     def cycle(self, forward: bool = True) -> str:
         if not self.options:
             return self.start
-        ret = self.options[self.offset]
-        delta = 1 if forward else -1
-        self.offset = (self.offset + delta) % len(self.options)
-        return ret
+        if self.pos is None:
+            self.pos = 0 if forward else len(self.options) - 1
+        else:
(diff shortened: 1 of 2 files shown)


Vane11ope:
@mhils  Could you please review this because the bug is there and I want that fixed.

mhils:
Thanks @Vane11ope!

Could you please

1. Add some tests in https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/tools/console/test_commander.py#L28?
2. Use something else for the "not set" offset value? I would favor `None` or `-1`. Semantically, sys.maxsize has nothing to do with that, so it's a bit of an odd choice.

Vane11ope:
@mhils 

thank you for the review :)

the initial value was NONE actually as you can tell that from my first commit, yet some CI complained about inserting NONE to int.

so I would probably go with -1. actually I really did not like to import sys just for this.

i am adding test too!

Vane11ope:
@mhils  done.

mhils:
Awesome, thanks! 🍰

PR #4060, opened 2020-07-03
add mapremote addon to modify request URLs

mplattner:
#### Description

Added a new addon (`MapRemote`) to rewrite URLs of requests before the request is performed.

refs #3948 

#### Checklist

### DIFF
--- a/docs/src/content/overview-features.md
+++ b/docs/src/content/overview-features.md
@@ -9,23 +9,17 @@ menu:
 # Mitmproxy Core Features
 
 
-- [Mitmproxy Core Features](#mitmproxy-core-features)
-  - [Anticache](#anticache)
-  - [Client-side replay](#client-side-replay)
-  - [Modify Body](#modify-body)
-    - [Examples](#examples)
-  - [Modify Headers](#modify-headers)
-    - [Examples](#examples-1)
-  - [Proxy Authentication](#proxy-authentication)
-  - [Server-side replay](#server-side-replay)
-    - [Response refreshing](#response-refreshing)
-    - [Replaying a session recorded in Reverse-proxy Mode](#replaying-a-session-recorded-in-reverse-proxy-mode)
-  - [Sticky auth](#sticky-auth)
-  - [Sticky cookies](#sticky-cookies)
-  - [Streaming](#streaming)
-    - [Customizing Streaming](#customizing-streaming)
-    - [Websockets](#websockets)
-  - [Upstream Certificates](#upstream-certificates)
+- [Anticache](#anticache)
+- [Client-side replay](#client-side-replay)
+- [Map Remote](#map-remote)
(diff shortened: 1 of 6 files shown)


mitmproxy/addons/mapremote.py
```diff
@@ -0,0 +1,66 @@
...
+            path = os.path.expanduser(repl[1:])
+            try:
+                with open(path, "rb") as f:
+                    repl = f.read()
+            except IOError:
+                ctx.log.warn("Could not read replacement file: %s" % repl)
+                return
+
+        replacements = 0
+        obj.url, replacements = re.subn(search, repl, strutils.escaped_str_to_bytes(obj.pretty_url), flags=re.DOTALL)
```
mhils, line 65: The URL is surrogate-escaped UTF-8, so we should do `obj.pretty_url.encode("utf8", "surrogateescape")`. We're doing a bit of a weird encoding dance here that's tricky to explain, but `escaped_str_to_bytes` would interpret escape sequences in a URL as raw bytes, which would fail in some edge cases.

docs/src/content/overview-features.md
```diff
@@ -71,17 +70,17 @@ The _separator_ is arbitrary, and is defined by the first character.
...
 ### Examples
 
-Replace `example.org` in all request URLs to with `mitmproxy.org`:
+Map all requests ending with `.jpg` to `https://placedog.net/640/480?random`:
 
-{{< highlight none  >}}
-/example.org/mitmproxy.org
-{{< / highlight >}}
+```
+:.*\.jpg$:https://placedog.net/640/480?random
```
mplattner, line 76: Two problems here:
* pattern is parsed incorrectly due to the separator `:` together with the skipped flow filter and the `:` in `https:`. it should instead be `"|.*\.jpg$|https://placedog.net/640/480?random"`
* using a image placeholder service that does not support http2 does not work for websites that do use http2. I debugged this for a while yesterday and then decided to ditch the image placeholder example but use `example.org/image.jpg`. we might want to put a note in the docs (or try to fix the issue, albeit I don't think this problem is a common use case.)

mplattner: other than that, great improvements to the docs :)

mhils: Good catch on the pattern. 👍 

The HTTP2 issue is a super annoying one, which we'll hopefully be able to fix with sans-io. Let's put in a general note for the addon for now.

mitmproxy/addons/mapremote.py
mplattner: Do you think this is what causes the failing unit test on windows? (e.g. `TestMapRemoteFile`; see https://github.com/mitmproxy/mitmproxy/pull/4046#issuecomment-652938102)

PR #4069, opened 2020-07-07
MapLocal addon

mplattner:
#### Description

Draft for the `MapLocal` addon.
`MapLocal.file_candidates()` generates possible paths to serve instead of the original resource, also see the tests to understand what it does.
`subject` (regex pattern on the url) is only used to determine whether a request should be modified, but `subject` does not affect the file candidates currently.
Hope this makes sense, here is an example:

Request URL = https://example.org/img/topic/subtopic/test.jpg
Map Local option = `:/img/jpg:~/local`
File candidates are:

* ~/local/img/topic/subtopic/test.jpg
* ~/local/img/topic/test.jpg
* ~/local/img/test.jpg
* ~/local/test.jpg

Basic functionality is already working, e.g. with `mitmproxy --map-local ":jpg:~/img"` to try to serve all URLs containing `jpg` from `~/img`.
However, overall this is just an early draft that needs several improvements and clean-up.

#### Checklist


mplattner:
I added mimetype guessing [as it is implemented in Flask](https://github.com/pallets/flask/blob/master/src/flask/helpers.py#L578).
What would you do if the mimetype cannot be determined? Currently then I used `text/plain`.

Another question is what we do in case non of the candidate files is found in the given local directory. Currently the request fires unchanged and the original resource is served. We could also send a 404 error. It depends on the use-case: for a rule like "serve all URLs that contain css from ~/local-css", e.g., `:css:~/local-css:` it might be nice to fallback to the original resource. Then the user can only put CSS files that need modification in ~/local-css and all other files keep being served from the original destination.

mitmproxy/addons/maplocal.py
```diff
@@ -0,0 +1,110 @@
...
+                except ValueError as e:
+                    raise exceptions.OptionsError(f"Cannot parse map_local option {option}: {e}") from e
+
+                self.replacements.append(spec)
+
+    def construct_candidate_path(self, base_path, path_components, filename):
+        candidate_path = base_path.joinpath("/".join(path_components + [filename]))
+        return str(candidate_path)
+
+    def sanitize_candidate_path(self, candidate_path, base_path):
```
mhils, line 42: Here's some existing work on that problem:

 - I believe this is what Flask/werkzeug is doing: https://github.com/pallets/werkzeug/blob/0d0d1e7263013c4e14750d8fb91984cf1584e0e0/src/werkzeug/security.py#L202
https://github.com/pallets/werkzeug/blob/09a248649069780b4adfb910fb69f6fb304f2ea3/tests/test_security.py#L69-L82
 - And here's aiohttp: https://github.com/aio-libs/aiohttp/blob/2a924cae6c19dbbfde220bbd461fc446764ad156/aiohttp/web_urldispatcher.py#L585

mitmproxy/addons/maplocal.py
```diff
@@ -0,0 +1,110 @@
...
+                    break
+
+                path_components.pop()
+
+        return candidates
+
+    def get_mime_type(self, file_path):
+        mimetype = (
+            mimetypes.guess_type(file_path)[0]
+            or "text/plain"
```
mhils, line 84: `'application/octet-stream'` is a bit more useful. The stuff that generally isn't recognized is binary blobs, which should trigger a download. :)

mitmproxy/addons/maplocal.py
mhils: Looking closer at this, it might actually be the best approach to just call werkzeug's [safe_join](https://werkzeug.palletsprojects.com/en/1.0.x/utils/?highlight=safe_join#werkzeug.security.safe_join). We could then also already call sanitization in `file_candidates` so that it only returns a list of safe paths. What do you think?

mitmproxy/addons/maplocal.py
```diff
@@ -0,0 +1,110 @@
...
+        return None
+
+    def file_candidates(self, url: str, spec: ModifySpec) -> typing.List[Path]:
+        replacement = spec.replacement
+        candidates = []
+
+        if replacement.is_file():
+            candidates.append(replacement)
+
+        elif replacement.is_dir():
```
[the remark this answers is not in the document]
mplattner: Good idea!

mplattner:
Your comments all make sense, thanks. 🙌

mitmproxy/addons/maplocal.py
mplattner: Do we consider a symlink in the user-provided base-dir to a target outside of base-dir a problem?
E.g., the user provides `/local` and the symlink /local/mylink -> /etc/. To consider this we _do_ need I/O access to sanitize paths.

test/mitmproxy/addons/test_maplocal.py
```diff
@@ -0,0 +1,112 @@
...
+        file_candidates = ml.file_candidates(url, spec)
+        assert file_candidates[0] == str(tmpdir) + "/img/topic/subtopic/index.html"
+        assert file_candidates[1] == str(tmpdir) + "/img/topic/index.html"
+        assert file_candidates[2] == str(tmpdir) + "/img/index.html"
+        assert file_candidates[3] == str(tmpdir) + "/index.html"
+
+        url = "https://example.org"
+        spec = parse_modify_spec(":org:" + str(tmpdir), True, True)
+        file_candidates = ml.file_candidates(url, spec)
+        assert file_candidates[0] == str(tmpdir) + "/index.html"
```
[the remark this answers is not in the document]
mplattner: The parametrization is really cool, thanks!

mitmproxy/addons/maplocal.py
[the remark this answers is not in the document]
mplattner: Hm, I see your point in removing I/O, but doesn't this complicate testing due to different path separators being used on Unix vs. Windows? I.e., on Windows we might want to test `:jpg:c:\\repl`, while on Linux it would be `:jpg:/repl`.
Thus, to have real cross-platform tests, we need to have OS-level parametrization. Maybe I'm over-complicating stuff.
Will look at this more closely in the evening.

mitmproxy/addons/maplocal.py
mhils: I think that's fine. If a user explicitly symlinks that, it may as well work. No need to check the filesystem.

mitmproxy/addons/maplocal.py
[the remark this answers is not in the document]
mhils: We can probably use `./static` as our test path. That should behave the same way cross-platform.

mitmproxy/addons/maplocal.py
mplattner: Yes, I agree. :+1:

mitmproxy/addons/maplocal.py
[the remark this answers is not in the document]
mplattner: Sure. I didn't know if we aim for tests that explicitly test differences between Unix and Windows.

mitmproxy/addons/modifyheaders.py
```diff
@@ -81,7 +82,7 @@ def parse_modify_spec(option, subject_is_regex: bool, replacement_is_path: bool)
     if replacement_is_path:
         path = Path(replacement)
         try:
-            replacement = path.expanduser().resolve(strict=True)
+            replacement = str(path.expanduser().resolve(strict=True))
```
mplattner, line 85: Hm, Windows and OSX tests for `MapLocal` fail due to the `resolve()` and the obviously non-existing `/tmp` path that the `test_file_candidates` test uses. We do this similarly for the other "replacement" addons when an option with `@path` is provided (see https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/addons/modifyheaders.py#L26), but `parse_modify_spec()` is never tested with an `@path` option. Therefore it's not an issue there.

Should I move the `resolve()` check outside of `parse_modify_spec()` or use the `tmpdir` of pytest?

mitmproxy/addons/maplocal.py
```diff
@@ -0,0 +1,104 @@
...
+
+            if spec.matches(flow) and re.search(spec.subject, url.encode("utf8", "surrogateescape")):
+                replacement_path = None
+                if base_path.is_file():
+                    replacement_path = base_path
+                elif base_path.is_dir():
+                    candidates = file_candidates(url, str(base_path))
+                    for candidate in candidates:
+                        # check that path is not outside of the user-defined base_path
+                        if candidate.is_file() and base_path in candidate.parents:
```
mhils, line 89: Is that check necessary here? I thought file_candidates gives us sanitized paths.

mitmproxy/addons/modifyheaders.py
mhils: I would probably try `tmpdir`, but either option works for me :)

test/mitmproxy/addons/test_maplocal.py
```diff
@@ -0,0 +1,77 @@
...
+
+from mitmproxy.addons.modifyheaders import parse_modify_spec
+
+
+@pytest.mark.parametrize(
+    "url,spec,expected_candidates", [
+        (
+            "https://example.org/img/topic/subtopic/test.jpg",
+            ":example.com/foo:/tmp",
+            ["/tmp/img/topic/subtopic/test.jpg", "/tmp/img/topic/test.jpg", "/tmp/img/test.jpg", "/tmp/test.jpg"]
```
mhils, line 15: I'm not sure if I understand this pattern. Why are we supposed to match here if `/foo` is in the subject?

test/mitmproxy/addons/test_maplocal.py
```diff
@@ -0,0 +1,77 @@
...
+        ),
+        (
+            "https://example.org/img/topic/subtopic/",
+            ":/img:/tmp",
```
mhils, line 19: I'm really not convinced that this form of patterns make much sense, same goes for the one below. What's the use case here? IMO the main point here is to specify URL prefixes, mapping all URLs that match "org" onto a folder seems really far fetched. Am I missing something?

test/mitmproxy/addons/test_maplocal.py
```diff
@@ -0,0 +1,77 @@
...
+            ["/tmp/img/topic/subtopic/index.html", "/tmp/img/topic/index.html", "/tmp/img/index.html", "/tmp/index.html"]
+        ),
+        (
+            "https://example.org",
+            ":org:/tmp",
+            ["/tmp/index.html"]
+        ),
+    ]
```
mhils, line 27: We should have a better test coverage here. How about varying prefix subjects (`example.com`, `example.com/foo`, `example.com/foo/`) and URLs including query strings, for example?

mitmproxy/addons/maplocal.py
```diff
@@ -0,0 +1,104 @@
...
+        mimetypes.guess_type(file_path)[0]
+        or "application/octet-stream"
+    )
+    return mimetype
+
+
+def file_candidates(url: str, base_path: str) -> typing.List[Path]:
+    candidates = []
+    parsed_url = urllib.parse.urlparse(url)
+    path_components = parsed_url.path.lstrip("/").split("/")
```
mhils, line 23: If we want to operate on path components, there's already `flow.request.path_components`. Intuitively I feel that this may work better matching on substrings, but if we get tests passing with this, we're good of course. :)

mitmproxy/addons/maplocal.py
```diff
@@ -0,0 +1,104 @@
...
+
+def get_mime_type(file_path: str) -> str:
+    mimetype = (
+        mimetypes.guess_type(file_path)[0]
+        or "application/octet-stream"
+    )
+    return mimetype
+
+
+def file_candidates(url: str, base_path: str) -> typing.List[Path]:
```
mhils, line 20: nit: let's keep accepting a Path here and only `str()` before we `safe_join`. We really want to work on Path objects, hopefully safe_join does support them at some point as well.

mitmproxy/addons/maplocal.py
mplattner: Oh, you're right, the `base_path in candidate.parents` is not needed as `safe_join` should only return paths that are within `base_path`.
The `candidate.is_file()`, however, is needed as we perform no I/O to check if this file exists up to this point.

test/mitmproxy/addons/test_maplocal.py
mplattner: You're right, this wouldn't match in the `request` method, but `file_candidates` does not consider `subject`, so it's technically a valid test. It is confusing though, I'll change it (also see next comment).

mitmproxy/addons/maplocal.py
mhils: Yes, we're on the same page then. 👍 😃

test/mitmproxy/addons/test_maplocal.py
mplattner: I agree, I didn't use particular useful or common use-case patterns, but just some that test the functionality.
Will replace them with more useful patterns, e.g., the ones I used in the docs.

mitmproxy/addons/maplocal.py
mplattner: Good idea, will change it.

mitmproxy/addons/maplocal.py
```diff
@@ -1,52 +1,86 @@
...
+
 from werkzeug.security import safe_join
 
-from mitmproxy import ctx, exceptions, http
-from mitmproxy.addons.modifyheaders import parse_modify_spec, ModifySpec
+from mitmproxy import ctx, exceptions, flowfilter, http
+from mitmproxy.addons.modifyheaders import parse_spec
+
+
+class MapLocalSpec(typing.NamedTuple):
```
mplattner, line 12: Having separate specs is a good idea, it got a bit messy trying to use the `ModifySpec` for the `Map*` addons. :+1:

mitmproxy/addons/maplocal.py
```diff
@@ -1,52 +1,86 @@
...
+        raise ValueError(f"Invalid file path: {replacement} ({e})")
+
+    return MapLocalSpec(filter, regex, path)
 
 
 def get_mime_type(file_path: str) -> str:
     mimetype = (
-        mimetypes.guess_type(file_path)[0]
-        or "application/octet-stream"
+            mimetypes.guess_type(file_path)[0]
```
mplattner, line 36: Why the double indentation?

mitmproxy/addons/maplocal.py
```diff
@@ -1,52 +1,86 @@
...
-            )
 
-        if not path_components:
-            break
+def file_candidates(url: str, spec: MapLocalSpec) -> typing.List[Path]:
+    candidates = []
 
-        path_components.pop()
+    m = re.search(spec.regex, url)
+    assert m
```
mplattner, line 60: `assert m` instead of a `if m:` block?

mitmproxy/addons/maplocal.py
```diff
@@ -1,52 +1,86 @@
...
+    if m.groups():
+        suffix = m.group(1)
+    else:
+        suffix = re.split(spec.regex, url, maxsplit=1)[1]
+        suffix = suffix.split("?")[0]  # remove query string
+
+    suffix = re.sub(r"[^0-9a-zA-Z-_.=(),/]", "_", suffix.strip("/"))
```
mplattner, line 67: Most filesystems support more characters than `r"[^0-9a-zA-Z-_.=(),/]"`, but this should be fine for most use cases and common file names ;)
Doesn't the `-` in `Z-_` need escaping with a `\`?

mitmproxy/addons/modifyheaders.py
```diff
@@ -7,92 +7,68 @@
...
+        return flow_filter, subject, replacement
     else:
         raise ValueError("Invalid number of parameters (2 or 3 are expected)")
 
-    subject = strutils.escaped_str_to_bytes(subject)
+
+class ModifySpec(typing.NamedTuple):
+    matches: flowfilter.TFilter
+    subject: bytes
+    replacement_str: str
```
mplattner: Minor: You changed this back from `replacement` to `replacement_str`.
Why include the type only for this one, and not for e.g. the subject? In `MapRemoteSpec` we use `replacement`.

test/mitmproxy/addons/test_maplocal.py
```diff
@@ -1,35 +1,59 @@
...
-    "url,spec,expected_candidates", [
-        (
-            "https://example.org/img/topic/subtopic/test.jpg",
-            ":example.com/foo:/tmp",
-            ["/tmp/img/topic/subtopic/test.jpg", "/tmp/img/topic/test.jpg", "/tmp/img/test.jpg", "/tmp/test.jpg"]
-        ),
+    "url,spec,expected_candidates",
+    [
+        # trailing slashes
+        ("https://example.com/foo", ":example.com/foo:/tmp", ["/tmp/index.html"]),
```
mplattner, line 16: Shouldn't this allow that `example.com/foo`is a file that I want to map to /tmp/foo?

mitmproxy/addons/modifyheaders.py
mhils: Ideally I would call it `_replacement` but NamedTuple doesn't like that. The idea is that you don't want to access `replacement_str` directly, but always use `read_replacement()`.

mitmproxy/addons/maplocal.py
mhils: This works fine, but probably a good call to add a `\` for clarity.

mitmproxy/addons/maplocal.py
mhils: We can assert here because we already check for a match in `request`, but I'll quickly refactor this.

mitmproxy/addons/maplocal.py
mhils: PyCharm Bug. 🤷

test/mitmproxy/addons/test_maplocal.py
mhils: Then you should specify `:example.com/foo:/tmp/foo` or  `:example.com/:/tmp/foo`, no?

mplattner:
Nice, LGTM! Thanks! :+1:

mitmproxy/addons/maplocal.py
```diff
@@ -1,52 +1,86 @@
...
+def file_candidates(url: str, spec: MapLocalSpec) -> typing.List[Path]:
+    candidates = []
 
-        path_components.pop()
+    m = re.search(spec.regex, url)
+    assert m
+    if m.groups():
+        suffix = m.group(1)
+    else:
+        suffix = re.split(spec.regex, url, maxsplit=1)[1]
```
mplattner, line 67: Here you assume that the user provides a regex that starts with the path (and an optional hostname).
Assumeing the URL is `https://example/foo/bar/baz.jpg` and subject is `:/bar/baz:/tmp`, then the determined suffix is`.jpg`, which does not make sense. Are we ok with that and state in the docs that the regex should start with `hostname/path-prefix`?

test/mitmproxy/addons/test_maplocal.py
mplattner: I think it should be possible to specify `:/foo/bar/baz.jpg:/tmp/baz.jpg` to map `https://example.com/foo/bar/baz.jpg` to `/tmp/baz.jpg`.

mhils: Why doesn't that work? That's a single file, so it shouldn't go through file_candidates at all.

mplattner: My bad, you're right. I tested the patterns with the `test_file_candidates` method.

mitmproxy/addons/maplocal.py
mhils: I think the best way to document that is state that the pattern splits the URL, and the part right of the pattern is appended to the directory. It would be cool to have two simple colored examples like this:

```
HTTP Request:  https://example.com/foo/bar/qux.jpg?timestamp=123
                              ▲             │
URL Regex: example.com/foo   ─┘             │
                                            ▼
Served File:                  ~/static/bar/qux.jpg
```

```
HTTP Request:  https://example.com/view?id=1
                                 ▲     │
URL Regex: example.com/(.+)     ─┘     │
                                       ▼
Served File:              ~/static/view_id=1
```

Unfortunately I can't color here, but hopefully it gets the point across. :-)

mhils: Talking about examples:

- we should probably list one example with `:~m GET:` to illustrate the point of the flow filter.
- We could consider discussing regex lookaheads here. I'm not sure what's simpler, do we prefer `:example.com/(?=.+\.jpg$):~/images` or `:~u .jpg$:example.com:~/images`? I guess both would get the job done. :)

mplattner: Good idea. For your second example, the regex should be `example.com/` or `example.org` I think. Will put a colored ASCII example in the docs, or did you think of an image?
Regarding examples: I think lookaheads often make regexes hard to read. Using the flow filter might be confusing as well, the user might wonder what's the purpose of the url-regex. But yes, both variants get the job done. Will think about it.

mhils:
This looks really good to me now! 🍰 🚀 😃 

LGTM once tests pass. No hurries, take care of your arm.

@mitmproxy/devs, @Prinzhorn, etc.: If you have any feedback on this, we're all ears! :)

docs/src/content/overview-features.md
```diff
@@ -51,22 +127,19 @@ The substituted URL is fetched instead of the original resource
...
 needs to support HTTP2 as well, otherwise the substituted request may fail.
-`map_remote` patterns looks like this:
+`map_remote` patterns look like this:
 
 ```
-|flow-filter|regex|replacement
-|flow-filter|regex|@file-path
-|regex|replacement
-|regex|@file-path
+|flow-filter|url-regex|replacement
```
Prinzhorn, line 133: What about putting the optional filter at the end? Not sure how the discussion in #3948 ended, it's the same in #4060

mitmproxy/addons/maplocal.py
```diff
@@ -0,0 +1,135 @@
+import mimetypes
+import re
+import typing
+from pathlib import Path
+
+from werkzeug.security import safe_join
+
+from mitmproxy import ctx, exceptions, flowfilter, http
+from mitmproxy.addons.modifyheaders import parse_spec
```
Prinzhorn, line 9: Should we move `parse_spec` to a utility or something? It's semi-odd to import something from an unrelated addon, or is it just me?

Prinzhorn:
The first thing I was checking out was local file inclusion, but nice work with the `safe_join`. Looks good on first glance.

mitmproxy/addons/maplocal.py
```diff
@@ -0,0 +1,135 @@
...
+        raise ValueError("Untrusted paths.")
+    return Path(joined)
+
+
+def file_candidates(url: str, spec: MapLocalSpec) -> typing.List[Path]:
+    """
+    Get all potential file candidates given a URL and a mapping spec ordered by preference.
+    This function already assumes that the spec regex matches the URL.
+    """
+    m = re.search(spec.regex, url)
```
Prinzhorn, line 54: Is there a way to make `filé.jpg` or `spa ce.jpg` work? I'm having a folder with three files, those two and `file.jpg` (without the special character). Only `file.jpg` works even though both my filesystem and the url allow special characters or spaces in names.

I guess at some point `url` or at least the path needs to be URL decoded.

mitmproxy/addons/maplocal.py
```diff
@@ -0,0 +1,135 @@
...
+    assert m
+    if m.groups():
+        suffix = m.group(1)
+    else:
+        suffix = re.split(spec.regex, url, maxsplit=1)[1]
+        suffix = suffix.split("?")[0]  # remove query string
+
+    suffix = re.sub(r"[^0-9a-zA-Z\-_.=(),/]", "_", suffix.strip("/"))
```
Prinzhorn, line 62: URL decode `suffix` here? What exactly is the intent of removing the special characters?

Prinzhorn: This causes

```
spa%20ce.jpg
spa_20ce.jpg
```

instead of `spa ce.jpg`

mhils: urldecode: yes. Let's fix that and add a test.

Special characters: Some characters are not valid in filenames, hence we want to escape them so that those files can be represented (e.g., `?` on Windows). This very much varies across OSes unfortunately, I think we don't feel like special-casing everything here. One thing we could do is add the unescaped version and the escaped version to our candidate list. What do you think?

mitmproxy/addons/maplocal.py
mhils: Ok, let's move this to `mitmproxy/utils/spec.py`

mitmproxy/addons/maplocal.py
Prinzhorn: I see, I didn't think about the case where you want to replace a file that _your_ file system does not support. Without removing them you couldn't do it at all.

> One thing we could do is add the unescaped version and the escaped version (if it differs) to our candidate list.

Yes I like that. The original version should be first in the list which will still allow you to use special characters in 99% of cases (like my two examples). And for this one edge case where there's a character that your filesystem cannot handle then you're still good.

docs/src/content/overview-features.md
```diff
@@ -43,6 +44,81 @@ You may want to use client-side replay in conjunction with the `anticache`
...
+Pattern | Description
+------- | -----------
+`\|example.com/main.js\|~/main-local.js` | Replace `example.com/main.js` with `~/main-local.js`.
+`\|example.com/static\|~/static` | Replace `example.com/static/foo/bar.css` with `~/static/foo/bar.css`.
+`\|example.com/static/foo\|~/static` | Replace `example.com/static/foo/bar.css` with `~/static/bar.css`.
+`\|~m GET\|example.com/static\|~/static` | Replace `example.com/static/foo/bar.css` with `~/static/foo/bar.css` (but only for GET requests).
+
+### Details
+
+If *local-path* is a file, this file will always be served. File changes will be reflected immediately, there is no caching.
```
Kriechi, line 79: *immediately* as in: the next time the file is requested?
Or are you actively re-opening the file in the middle of an ongoing transfer?

docs/src/content/overview-features.md
```diff
@@ -43,6 +44,81 @@ You may want to use client-side replay in conjunction with the `anticache`
...
+                            │
+                            │    URL is split here
+                            ▼            ▼
+HTTP Request URL: https://<span style="color:#f92672">example.com/css</span><span style="color:#66d9ef">/print/main.css</span><span style="color:#bbb">?timestamp=123</span>
+                                               │                ▼
+                                               ▼              query string is ignored
+Served File:      Preferred: <span style="color:#82b719">~/static-css</span><span style="color:#66d9ef">/print/main.css</span>
+                   Fallback: <span style="color:#82b719">~/static-css</span><span style="color:#66d9ef">/print/main.css</span>/index.html
+                  Otherwise: 404 response without content
+</pre>
```
Kriechi, line 99: The inline span tags make the arrows not really nice to read in the raw `.md` file.
Any chance you could include dummy tags or inline comments to pad all the arrows/lines correctly?

docs/src/content/overview-features.md
mhils: There are three arguments in favor of keeping them at the start:

- It matches the structure of natural language, e.g. "Only for flows matching X replace Y with Z".
- For some reason we often use `/` as the separator in examples - maybe it's inspired by sed, I don't know. In any case, this clashes with replacement paths that contain slashes ( `/~m GET/foo/@~/foo.bin` works as expected, if we change argument order, `/foo/@~/foo.bin/~m GET` would be parsed wrongly)
- Existing patterns continue to work as-is.

None of these arguments is particularly strong, but neither are the arguments for moving them to the end. I don't have a very strong opinion here in general, if you all feel strongly we can move them.

docs/src/content/overview-features.md
mhils: immediately as in the next time it is requested. I thought that'd be sufficiently clear here, isn't it?

Prinzhorn: Maybe the second sentence is unnecessary? Why bother the reader with a behavior (caching) that we _didn't_ implement. It behaves as one would expect and everyone is happy.

docs/src/content/overview-features.md
Prinzhorn: No feelings at all, just making sure it wasn't overlooked as you were originally [in favor of it](https://github.com/mitmproxy/mitmproxy/issues/3948#issuecomment-648932541)

mitmproxy/addons/maplocal.py
```diff
@@ -0,0 +1,135 @@
...
+
+                if spec.local_path.is_file():
+                    local_file = spec.local_path
+                elif spec.local_path.is_dir():
+                    for candidate in file_candidates(url, spec):
+                        if candidate.is_file():
+                            local_file = candidate
+                            break
+
+                headers = {}
```
Prinzhorn, line 122: How about we add a `Server: mitmproxy/MapLocal` (or whatever) header? When looking at the response in dev tools or similar it would make it super clear what's happening. Makes it a no-brainer to check if the mapping is correctly working.

mhils: Good point, we have [`version.MITMPROXY`](https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/version.py#L7) for this.

mitmproxy/addons/maplocal.py
```diff
@@ -0,0 +1,135 @@
...
+                if local_file:
+                    flow.response = http.HTTPResponse.make(
+                        200,
+                        local_file.read_bytes(),
+                        headers
+                    )
+                    # only set flow.response once, for the first matching rule
+                    return
+        if any_spec_matches:
+            flow.response = http.HTTPResponse.make(404)
```
mhils, line 135: We should also consider adding a debug log entry here ([like this](https://github.com/mitmproxy/mitmproxy/blob/master/examples/addons/log-events.py)) that explicitly names the files we tried. This involves a bit of refactoring to keep the list of everything we tried, but should work quite well. :)

docs/src/content/overview-features.md
Kriechi: What happens if the file being served is open in an editor and it gets saved in the middle of a transfer?

mitmproxy/addons/maplocal.py
mplattner: Good idea, done.

docs/src/content/overview-features.md
mplattner: Yeah I brought this up but agree with Max that its better to leave it as it is.

mitmproxy/addons/maplocal.py
mplattner: :+1:, done

docs/src/content/overview-features.md
mhils: We just do `file.read_bytes()`, everything that breaks that we don't care about. Which reminds that we should have a test for file read errors. We shouldn't silently break on that and serve a 502/400 instead. :)

docs/src/content/overview-features.md
mplattner: That's a great idea, thanks. Done.

mitmproxy/addons/maplocal.py
mplattner: Good idea, I added it.

mitmproxy/addons/maplocal.py
```diff
@@ -133,11 +133,15 @@ def request(self, flow: http.HTTPFlow) -> None:
...
-                        headers
-                    )
+                    try:
+                        flow.response = http.HTTPResponse.make(
+                            200,
+                            local_file.read_bytes(),
+                            headers
+                        )
+                    except IOError as e:
+                        ctx.log.warn(f"Could not read file: {e}")
```
mplattner, line 143: Hey @mhils, any idea how to test this code. Above we perform `is_dir()` and `is_file()`, with no chance to delete the file in between. Can we trigger the `IOError` somehow?

mhils: How about using pytest's monkeypatch functionality?

mplattner: Great, thanks.

mplattner:
Happy to see this being merged. Thanks.

Issue #4072, opened 2020-07-09
Problems with DNS Rebind Protection and IPv6


tunnelpr0:
Hi,
it seems that somebody has forgotten, that IPv6 is using a hexadecimal system and didnt implemented the hexadecimal numbers 10 - 15 in the DNS Rebind regex. So maybe the currently used pattern at the bottom of this text should be changed to something which is really matching IPv4 and IPv6 addresses... what about:

```
^localhost$|^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$|^\[((?:[0-9a-fA-F]{1,4}\:){7}[0-9a-fA-F]{1,4}|\:\:(?:[0-9a-fA-F]{1,4}\:){0,6}[0-9a-fA-F]{1,4}|[0-9a-fA-F]{1,4}\:\:(?:[0-9a-fA-F]{1,4}\:){0,5}[0-9a-fA-F]{1,4}|[0-9a-fA-F]{1,4}\:[0-9a-fA-F]{1,4}\:\:(?:[0-9a-fA-F]{1,4}\:){0,4}[0-9a-fA-F]{1,4}|(?:[0-9a-fA-F]{1,4}\:){0,2}[0-9a-fA-F]{1,4}\:\:(?:[0-9a-fA-F]{1,4}\:){0,3}[0-9a-fA-F]{1,4}|(?:[0-9a-fA-F]{1,4}\:){0,3}[0-9a-fA-F]{1,4}\:\:(?:[0-9a-fA-F]{1,4}\:){0,2}[0-9a-fA-F]{1,4}|(?:[0-9a-fA-F]{1,4}\:){0,4}[0-9a-fA-F]{1,4}\:\:(?:[0-9a-fA-F]{1,4}\:)?[0-9a-fA-F]{1,4}|(?:[0-9a-fA-F]{1,4}\:){0,5}[0-9a-fA-F]{1,4}\:\:[0-9a-fA-F]{1,4}|(?:[0-9a-fA-F]{1,4}\:){0,6}[0-9a-fA-F]{1,4}\:\:)\]$
```
  
  
Otherwise the DNS Rebind would catch any IPv6 addresses - quite uncool in a IPv6 only network. :P

[...]

PR #4073, opened 2020-07-09
Fixes issue #4072 (correct IPv6 regex)

ylmrx:
#### Description

Fixes issue #4072 : Replace the IP regex so it covers all hex range (0-9 -> 0-f)

#### Checklist


mitmproxy/tools/web/app.py
```diff
@@ -499,7 +499,7 @@ def __init__(self, master: "mitmproxy.tools.web.master.WebMaster", debug: bool)
         self.add_handlers("dns-rebind-protection", [(r"/.*", DnsRebind)])
         self.add_handlers(
             # make mitmweb accessible by IP only to prevent DNS rebinding.
-            r'^(localhost|[0-9.:\[\]]+)$',
+            r'^(localhost|[\da-f.:\[\]]+)$',
```
Kriechi, line 502: I think this would allow `abcdef.ac` which is a valid domain + TLD - and therefore not really what we want here?

ylmrx: It's incomplete (it validated "999...1234.123" as an IP for example) as it was before.
A full RE (OP provided one in #4072), which reliably validates all IPv4 and 6, will be really complicated 
(imho, it shouldn't rely on a regex for this purpose).

tunnelpr0: I also think that this is quite a bullshit and you are transforming a "normal" issue to a security issue...

Please use the longer regex pattern - i am pretty sure that it wont cause issues!

ylmrx: I get this is the first time you submit an issue. You should know this language is way outside the line.
Feel free to review and put the big RE. 
I'm out.

ylmrx:
duplicate. follow-up in #4074

mhils:
Thanks @ylmrx! 

@tunnelpr0, we all try to be friendly here. Please turn down your language.

PR #4074, opened 2020-07-09
Secure fix for the DNS Rebind Protection issue from #4072

tunnelpr0:
#### Description

I changed the problematic RegEx Pattern that i critized earlier in #4072 

####  Checklist

### DIFF
--- a/mitmproxy/tools/web/app.py
+++ b/mitmproxy/tools/web/app.py
@@ -499,8 +499,7 @@ class Application(tornado.web.Application):
         self.add_handlers("dns-rebind-protection", [(r"/.*", DnsRebind)])
         self.add_handlers(
             # make mitmweb accessible by IP only to prevent DNS rebinding.
-            # IPv(4|6) pattern from https://riptutorial.com/regex/example/14146/match-an-ip-address, slightly adjusted to mitmproxy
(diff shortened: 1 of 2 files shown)


mitmproxy/tools/web/app.py
```diff
@@ -499,7 +499,8 @@ def __init__(self, master: "mitmproxy.tools.web.master.WebMaster", debug: bool)
         self.add_handlers("dns-rebind-protection", [(r"/.*", DnsRebind)])
         self.add_handlers(
             # make mitmweb accessible by IP only to prevent DNS rebinding.
-            r'^(localhost|[0-9.:\[\]]+)$',
+            # IPv(4|6) pattern from https://riptutorial.com/regex/example/14146/match-an-ip-address, slightly adjusted to mitmproxy
+            r'^localhost$|^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$|^\[((?:[0-9a-fA-F]{1,4}\:){7}[0-9a-fA-F]{1,4}|\:\:(?:[0-9a-fA-F]{1,4}\:){0,6}[0-9a-fA-F]{1,4}|[0-9a-fA-F]{1,4}\:\:(?:[0-9a-fA-F]{1,4}\:){0,5}[0-9a-fA-F]{1,4}|[0-9a-fA-F]{1,4}\:[0-9a-fA-F]{1,4}\:\:(?:[0-9a-fA-F]{1,4}\:){0,4}[0-9a-fA-F]{1,4}|(?:[0-9a-fA-F]{1,4}\:){0,2}[0-9a-fA-F]{1,4}\:\:(?:[0-9a-fA-F]{1,4}\:){0,3}[0-9a-fA-F]{1,4}|(?:[0-9a-fA-F]{1,4}\:){0,3}[0-9a-fA-F]{1,4}\:\:(?:[0-9a-fA-F]{1,4}\:){0,2}[0-9a-fA-F]{1,4}|(?:[0-9a-fA-F]{1,4}\:){0,4}[0-9a-fA-F]{1,4}\:\:(?:[0-9a-fA-F]{1,4}\:)?[0-9a-fA-F]{1,4}|(?:[0-9a-fA-F]{1,4}\:){0,5}[0-9a-fA-F]{1,4}\:\:[0-9a-fA-F]{1,4}|(?:[0-9a-fA-F]{1,4}\:){0,6}[0-9a-fA-F]{1,4}\:\:)\]$',
```
Kriechi, line 503: This is quite a mouth full...
How about we try and split this long regex into multiple lines based on their main cases it covers:
* localhost
* IPv4 address
* IPv6 address
  * maybe even split it into one line per `::` sub-cases

Furthermore, I would suggest simplifying the IPv4 case to the [first example](https://riptutorial.com/regex/example/14146/match-an-ip-address), since the only thing we care about here in this function is that it "is not a valid domain name".

I not sure what your modifications to the IPv6 case are, but I don't spot the case with plain octets, i.e., without the `[...]`.

Maybe this is out-of-scope, or even irrelevant, but do we consider IPv4-mapped IPv6 address, e.g., `::ffff:c000:0280` is written as `::ffff:192.0.2.128`?

At this point, there is actually a lot of logic and considerations in this seemingly easy code snippet. I feel we should start covering this with a few basic unit tests (postive matching, as well as negative).

tunnelpr0: You are right - tomorrow i will think about and try to find another solution, that i will suggest afterwards here.

mhils: Ignoring IPv4-mapped IPv6 addresses for now (not sure if they are applicable here), is there anything wrong with `^(localhost|[0-9.]+|[0-9a-f:\[\]]+)$`? Again, the point is to have an 1) easy-to-understand expression that 2) matches all IPs 3) but no domains. The point is _not_ to correctly validate IP addresses.

tunnelpr0: Yeah, but your pattern allows standalone hostnames in local networks (example with a host called "beef" --> hhttps://regex101.com/r/bKvMJb/1). 

But theres one thing that could help - the brackets at a IPv6 address aint somewhere in the address, they are everytime at the begin and the end of the address. And, another thing, the letters in an IPv6 address could be also uppercase.

So what about this pattern? `^(localhost|[0-9.]+|\[[0-9a-fA-F:]+\])$` (https://regex101.com/r/MK9LSy/2)

mhils: Looks good to me!

tunnelpr0: I changed the Pattern according to my suggestion in my Pull Request and all tests were successful. :)

PR #4075, opened 2020-07-10
Command that can generates errors in the documentation

sidahmed-malaoui:
### Description

In the title _Work-around to redirect traffic originating from the machine itself_, the command `sudo` is used without the flag `-H`. 

As stated by `sudo` manual :
> -H, --set-home
                 Request that the security policy set the HOME environment variable to the home directory specified by
                 the target user's password database entry.  Depending on the policy, this may be the default behavior.

Without this flag, in some Linux distros (I tested on Ubuntu 16.04), pip will try to install `mitmproxy` in the home directory of the user launching the command, and not in the directory of the user _mitmproxyuser_

#### Checklist

### DIFF
--- a/docs/src/content/howto-transparent.md
+++ b/docs/src/content/howto-transparent.md
@@ -94,7 +94,7 @@ Create a user to run the mitmproxy
 
 ```bash
 sudo useradd --create-home mitmproxyuser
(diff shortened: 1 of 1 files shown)


mhils:
That sounds reasonable. Thanks for the PR! 🍰

sidahmed-malaoui:
@mhils You're welcome. Thank you for this wonderful and amazing `mitmproxy`

PR #4080, opened 2020-07-12
Fix deadlock caused by the "replay.client.stop" command

### LINKED ISSUE #4037: replay.client.stop crashes the client
#### Problem Description
Running replay.client.stop (with or without pending replays) will crash the client and will require a full restart.

Versions:
MacOS 10.15.5 (19F96)
Python 3.7.7 (default, Mar 10 2020, 15:43:33)

The same also occurred on Ubuntu 20.04.

1. Launch mitmproxy with given version
2. Run :replay.client.stop
3. Whole client crashes

#### System Information
Mitmproxy: 5.1.1
Python:    3.7.7
Platform:  Darwin-19.5.0-x86_64-i386-64bit

gorogoroumaru:
Fix #4037 

When we execute the command "replay.client.stop", this function is called.
```@command.command("replay.client.stop")
def stop_replay(self) -> None:
    with self.q.mutex:
        lst = list(self.q.queue)
        self.q.queue.clear()
        for f in lst:
            f.revert()
        ctx.master.addons.trigger("update", lst) 
    ctx.log.alert("Client replay queue cleared.")
```

In this function, "ctx.master.addons.trigger("update", lst)" is called inside "with self.q.mutex".
This causes a deadlock.
So, I moved "ctx.master.addons.trigger("update", lst)" outside of "with self.q.mutex"

### DIFF
--- a/mitmproxy/addons/clientplayback.py
+++ b/mitmproxy/addons/clientplayback.py
@@ -180,7 +180,7 @@ class ClientPlayback:
             self.q.queue.clear()
             for f in lst:
                 f.revert()
-            ctx.master.addons.trigger("update", lst)
+        ctx.master.addons.trigger("update", lst)
         ctx.log.alert("Client replay queue cleared.")
 
     @command.command("replay.client")


mhils:
Awesome, thanks for tracking this down! ❤️ 🍰

PR #4088, opened 2020-07-16
HTTPRequest -> http.Request, add request.authority

### LINKED ISSUE #4055: How to handle conflicting HTTP Host information?
Let's assume the following HTTP/1 exchange with a regular HTTP proxy:

```http
CONNECT example.com:80 HTTP/1.1
Host: example.org:80 

HTTP/1.1 200 OK

GET http://foo.bar/ HTTP/1.0
Host: qux.dev
Host: qux.xyz
```

In this example, we observe five conflicting destinations: example.com, example.org, foo.bar, and qux.dev, and qux.xyz. This is of course a bit contrived, but should illustrate the problem. For a more practical example, let's think of CONNECT + an HTTP/2 request with `:authority` and `host` headers.

### Current Behavior

We first ignore the Host header in the CONNECT request entirely. Then we fail because we receive an absolute-form request while expecting a relative-form one. Absolute-form is valid HTTP, so we shouldn't continue doing this.

If we assume `GET / HTTP/1.1` instead, mitmproxy will set `.host` is `example.com` and `.pretty_host` to `qux.dev, qux.xyz`.

### Desired Behavior

I think some of these problems are relatively easy:

- **Host header in CONNECT requests:** Host headers never leave the proxy here, we can just keep ignoring these.
- **Conflicting Host headers:** This problem is a bit of a tangent here - we can just forward those as-is if other host information is present.

This leaves us with `example.com`, `foo.bar` and `qux.dev`. What do we want to do here? Some ideas:

- Introduce an additional `request.authority` attribute. `.pretty_host` would then try `.authority`, host header and `.host` (in that order). `.host` would update `.authority` if it is present.
- Raise if `:authority` and `Host` are both present and do not match. A bit of an open question is how we determine what to send upstream?
- Use `.host` for `:authority`. Among other things, this'd mean that we have spoofable information in `.host`. Not what we want I guess.

@mitmproxy/devs, any ideas?

See also: 
  - https://www.icir.org/vern/papers/host-of-troubles.ccs16.pdf
  - https://github.com/nghttp2/nghttp2/issues/317

refs #848

mhils:
This PR is a bit of a monster, but necessary:

 - Introduce `flow.request.authority`, which captures the authority portion of an HTTP request and the `:authority` header for HTTP/2. That's necessary so that we can have a clean h1<->h2 interop with sans-io. It also means that `first_line_format` is a computed property now, we also kind of fix #4055.
 - Remove the `HTTPRequest`/`HTTPResponse` wrappers for `http.Request` and `http.Response`. This involved shifting some attributes (request.is_replay/response.is_replay -> flow.is_replay), but generally simplifies code.

### DIFF
--- a/mitmproxy/net/http/http1/assemble.py
+++ b/mitmproxy/net/http/http1/assemble.py
@@ -45,31 +45,26 @@ def _assemble_request_line(request_data):
     Args:
         request_data (mitmproxy.net.http.request.RequestData)
     """
-    form = request_data.first_line_format
-    if form == "relative":
+    if request_data.method.upper() == b"CONNECT":
+        return b"%s %s %s" % (
+            request_data.method,
+            request_data.authority,
+            request_data.http_version
+        )
+    elif request_data.authority:
+        return b"%s %s://%s%s %s" % (
+            request_data.method,
+            request_data.scheme,
+            request_data.authority,
+            request_data.path,
+            request_data.http_version
+        )
+    else:
         return b"%s %s %s" % (
             request_data.method,
             request_data.path,
             request_data.http_version
         )
-    elif form == "authority":
-        return b"%s %s:%d %s" % (
-            request_data.method,
-            request_data.host,
-            request_data.port,
-            request_data.http_version
-        )
-    elif form == "absolute":
-        return b"%s %s://%s:%d%s %s" % (
-            request_data.method,
-            request_data.scheme,
-            request_data.host,
-            request_data.port,
-            request_data.path,
-            request_data.http_version
-        )
-    else:
-        raise RuntimeError("Invalid request form")
 
 
 def _assemble_request_headers(request_data):
--- a/test/mitmproxy/addons/test_session.py
+++ b/test/mitmproxy/addons/test_session.py
@@ -160,6 +160,7 @@ class TestSession:
         assert len(s._view) == 4
 
     @pytest.mark.asyncio
+    @pytest.mark.skip
     async def test_storage_flush_with_specials(self):
         s = self.start_session(fp=0.5)
         f = self.tft()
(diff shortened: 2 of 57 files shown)


mitmproxy/net/http/http1/assemble.py
```diff
@@ -45,31 +45,26 @@ def _assemble_request_line(request_data):
...
-    elif form == "absolute":
-        return b"%s %s://%s:%d%s %s" % (
+    else:
+        return b"%s %s %s" % (
             request_data.method,
-            request_data.scheme,
-            request_data.host,
-            request_data.port,
             request_data.path,
             request_data.http_version
```
Kriechi, line 66: Any chance we can use fancy new Python f-strings?

test/mitmproxy/addons/test_session.py
```diff
@@ -160,6 +160,7 @@ def test_storage_filter(self):
         assert len(s._view) == 4
 
     @pytest.mark.asyncio
+    @pytest.mark.skip
```
Kriechi, line 163: Is this `skip` a debug left-over?

mitmproxy/net/http/http1/assemble.py
mhils: Unfortunately no, they do not work for bytestrings.

test/mitmproxy/addons/test_session.py
mhils: No, this is intentional. We currently don't use the session/db/protobuf code and it's quite a big maintenance burden. The alternative would be removing it from the source entirely, which I'd also be fine with.

Adjusting this is easy when we decide to adopt it, keeping it up-to-date for every small change is burdensome.

PR #4089, opened 2020-07-16
match HTTP-WebSocket handshake flow for ~websocket filter

### LINKED ISSUE #3990: ~websocket filter when reading from a file does not work
#### Problem Description
It appears that the `~websocket` filter, which is supposed to match all WebSocket flows per <https://docs.mitmproxy.org/stable/concepts-filters/>, does not match anything when reading from an existing dump.

1. Capture WebSocket traffic: `mitmdump -w infile ~websocket`
2. Apply a filter of `~websocket`: `mitmdump -nr infile -w outfile ~websocket`

The result is a non-empty `infile`, which contains the expected WebSocket messages, and an **empty** `outfile` even though they should contain the same data.

In addition, I also noticed that `mitmdump -nr infile` produces no output whatsoever, even with `-v` to increase the verbosity. I'm aware that the UI does not properly support WebSocket (#1020), but I'd expect it to at least report that messages exist as it does during dumping (i.e. `127.0.0.1:57466 -> WebSocket 1 message -> URL` etc.). This does not happen when the initial capturing is done without a filter, i.e. those "message" log lines appear in that case. This might point at a more fundamental issue with reading back the WebSocket data than just the filtering.

#### System Information
```
$ mitmproxy --version
Mitmproxy: 5.1.1 (+2, commit a8ca63f)
Python:    3.8.2
Platform:  Linux-5.6.0-1-amd64-x86_64-with-glibc2.29
```

(Regarding the version number, see #3987; this is a clean 5.1.1 installation.)

Kriechi:
#### Description

fixes #3990 by also matching HTTP-WebSocket handshake flow with `~websocket`.

Also add py38 to the default tox env list.

#### Checklist

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -20,6 +20,7 @@ Unreleased: mitmproxy next
     * Add support for HTTP Trailers to the HTTP/2 protocol (@sanlengjingvv and @Kriechi)
     * Fix certificate runtime error during expire cleanup (@gorogoroumaru)
     * Fixed the DNS Rebind Protection for secure support of IPv6 addresses (@tunnelpr0)
+    * WebSockets: match the HTTP-WebSocket flow for the ~websocket filter (@Kriechi)
 
     * --- TODO: add new PRs above this line ---
 
--- a/mitmproxy/flowfilter.py
+++ b/mitmproxy/flowfilter.py
@@ -43,6 +43,7 @@ from mitmproxy import flow
 from mitmproxy import http
 from mitmproxy import tcp
 from mitmproxy import websocket
+from mitmproxy.net import websockets as net_websockets
 
 
 def only(*types):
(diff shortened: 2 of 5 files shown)


Kriechi:
@JustAnotherArchivist does this fix your issue from #3990? Unfortunately existing dump files are "corrupt" as they did not save the handshake flow. Please create a new dump file.

JustAnotherArchivist:
@Kriechi It does, thank you! Using the steps from #3990 with 50a1f6bd, the two files produced are identical.

As I understand it from your commit, the corruption on existing data should only affect files that were initially captured with a filter because those would not include the initial handshake. So captures without filters (or less restrictive ones that don't exclude the handshake) should be fine, right? I'm asking because it doesn't appear to work that way. Specifically, if I `mitmdump -w full.dump` with version 5.1.1 and then filter that with `mitmdump -nr full.dump -w websockets.dump ~websocket` using the current commit (50a1f6bd), the resulting `websockets.dump` file only contains the handshake (with type `http`) but not the WebSocket data (with type `websocket`). I have confirmed that `full.dump` contains the WebSocket data as well.

Kriechi:
> As I understand it from your commit, the corruption on existing data should only affect files that were initially captured with a filter because those would not include the initial handshake.
Correct.

> So captures without filters (or less restrictive ones that don't exclude the handshake) should be fine, right?
Should be, yes.

I now run the following test procedure:
`git checkout 2dfcb537f27f38fb3631af835784753a490cc414 && mitmdump -w full.dump`
I started my Firefox which byitself does a ton of Google Analytics and other house-keeping requests. Then I visited the websocket echo server, connected, sent two messages, and disconected. All traffic is visible in mitmdump and written to the `full.dump`.

Then filter with the recent fix applied: 
`git checkout 50a1f6bdd1a711f7aa222e261966ddc94a94c193 && mitmdump -nr full.dump -w websockets.dump "~websocket"`
Which, as expected, only shows the websocket messages and their initial handshake flow.

Then just for double-confirmation: `mitmdump -nvvr websockets.dump` prints exactly the same flows again.

Is this different from what you expect to see? And what did you get actually stored in the dump files?

PR #4096, opened 2020-07-21
Add a msgpack content viewer.

tasn:
#### Description

This is a msgpack content viewer.
Is this of interest? If so, what kind of tests should I add if at all?

Thanks.

#### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -71,6 +71,7 @@ setup(
         "hyperframe>=5.1.0,<6",
         "kaitaistruct>=0.7,<0.9",
         "ldap3>=2.6.1,<2.8",
+        "msgpack>=1.0.0, <1.1.0",
         "passlib>=1.6.5, <1.8",
         "protobuf>=3.6.0, <3.12",
         "pyasn1>=0.3.1,<0.5",
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,8 @@ Unreleased: mitmproxy next
 
 ** Full Changelog **
 
+    * Add MsgPack content viewer (@tasn)
+
     * --- TODO: add new PRs above this line ---
 
     * ... and various other fixes, documentation improvements, dependency version bumps, etc.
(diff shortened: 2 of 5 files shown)


Kriechi:
Yes - I think this would be a nice feature to have!

You can find similar tests for other contentviews under https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/contentviews/test_json.py and related files in this folder.

setup.py
```diff
@@ -71,6 +71,7 @@
         "hyperframe>=5.1.0,<6",
         "kaitaistruct>=0.7,<0.9",
         "ldap3>=2.6.1,<2.8",
+        "msgpack>=1.0.0",
```
Kriechi, line 74: Please add a `<2` upper bound, assuming they follow semver?

tasn:
I can't run the tests locally for some reason. I'm getting:
```
pkg_resources.DistributionNotFound: The 'msgpack>=1.0.0' distribution was not found and is required by mitmproxy
ERROR: InvocationError for command /home/tom/projects/tmp/mitmproxy/.tox/py/bin/mitmdump --version (exited with code 1)
```

So I have no idea why the rest of the tests fail. If you have any idea, or you could maybe fix it, that would be great. I tried running mypy but already saw errors in JSON, and the errors it's complaining about in msgpack are not really errors. Maybe just missing typings?

setup.py
tasn: Done.

Kriechi:
code coverage is complaining: https://github.com/mitmproxy/mitmproxy/pull/4096/checks?check_run_id=898723405#step:5:58

Otherwise, I think this looks good.
Could you maybe post a screenshot of an example here?

tasn:
Fixed.

mhils:
Awesome, thanks! 🍰

tasn:
Great, thanks!

PR #4097, opened 2020-07-21
Fix incompatibility reading version 7 files when flow has no response.

rbdixon:
A flow can be stored without a response. The compatibility functions
for version 8 and 9 assumed that all flows had a response. This changes
fixes that assumption.

I discovered this problem when opening save files that I had recorded on prior versions of mitmproxy.


Kriechi:
While I totally agree with the reasoning, I'm a bit puzzled because we have similar compatibility functions in previous versions... was this maybe always broken?

@mitmproxy/devs ?

mhils:
Do we? I don't see any for `data["response"]`.

Whatsoever, we need to add a regression test here! This conversion code is super error-prone because there are none of th usual type checks.

Kriechi:
Similar cases in previous converters:
https://github.com/mitmproxy/mitmproxy/blob/3e7f9e488a8d7dd1e86b9a355cda3630d8f372ef/mitmproxy/io/compat.py#L29-L32
https://github.com/mitmproxy/mitmproxy/blob/3e7f9e488a8d7dd1e86b9a355cda3630d8f372ef/mitmproxy/io/compat.py#L46

We do have some tests - but none of these dump files have missing responses:
https://github.com/mitmproxy/mitmproxy/blob/master/test/mitmproxy/io/test_compat.py

rbdixon:
I pulled out the flow that caused the crash. The test case fails on master and passes on this branch.

I also merged what would have been three duplicated tests.

Issue #3728, opened 2019-11-25
mitmproxy 5.0.0 crashes on macOS 10.12 (ok on 10.13)


tomlabaude:
1. Downloaded https://snapshots.mitmproxy.org/branches/master/mitmproxy-master-osx.tar.gz
2. Crashes on macOS 10.12.6 with error

```
./mitmdump --version
[787] Error loading Python lib '/var/folders/sj/f0p38wnd2dz1483n0zlns51w0000gn/T/_MEIJdk2Qd/Python': dlopen: dlopen(/var/folders/sj/f0p38wnd2dz1483n0zlns51w0000gn/T/_MEIJdk2Qd/Python, 10): Symbol not found: _futimens
  Referenced from: /var/folders/sj/f0p38wnd2dz1483n0zlns51w0000gn/T/_MEIJdk2Qd/Python (which was built for Mac OS X 10.13)
  Expected in: /usr/lib/libSystem.B.dylib
 in /var/folders/sj/f0p38wnd2dz1483n0zlns51w0000gn/T/_MEIJdk2Qd/Python
```

-> It runs without problem on 10.13.6

mhils:
Yeah, this is a bit annoying. GitHub Actions currently only provides runners for macos-latest. PyInstaller [wants us to compile on the oldest supported macOS version](https://pythonhosted.org/PyInstaller/usage.html#building-mac-os-x-app-bundles). Two options:

1. We declare that our binaries only support the latest macOS release.
2. We fix compatibility with older versions.

For Option 2, I see the following possibilities:

- Use self-hosted Github CI runners with an old OSX version. This is suuuuper painful, see e.g. https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-self-hosted-runners#self-hosted-runner-security-with-public-repositories
- Bring back Travis CI to build binaries. Re-introducing another CI provider is kind of annoying, but should work. We definitely want to exclude that from Pull Requests to keep build times short.

I have no idea how bad Option 1 would be. @mitmproxy/devs, any opinions?

Kriechi:
Option 1 doesn't sound too bad...

mhils:
Ok. Let's try 1 and reopen here if we receive complaints.

tomlabaude:
> 1. We declare that our binaries only support the latest macOS release.

Where/how can we know which macOS versions are supported for a specific mitmproxy release ?

[...]

PR #4104, opened 2020-07-27
CI Improvements

mhils:
- Pin binary-generating CI jobs to to lowest available macOS and Windows versions (see https://github.com/mitmproxy/mitmproxy/issues/3728). Those were unavailable when we set up CI.
- Add jobs for Python 3.6 and 3.7. We're deliberately not running the full matrix (Windows/macOS) because that takes too much time.
- Fix compatibility with Python 3.6 (incompatibilities were introduced in #4088, after 5.2)
- Simplify CI script with recent GitHub changes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,30 +42,33 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, windows-latest, ubuntu-latest]
+        include:
+          - os: windows-latest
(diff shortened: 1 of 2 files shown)


mhils:
Sending this PR from my fork wasn't a good idea as that doesn't trigger artifact download I think. I've pushed it to the fix-ci branch here: https://github.com/mitmproxy/mitmproxy/tree/fix-ci

Issue #4112, opened 2020-07-29
Is it possible to load folder with scripts?


VladimirPodolyan:
#### Is your feature request related to a problem? Please describe.
I'm always frustrated when I should describe scripts via `-s path/to/scripts/1.py -s path/to/scripts/2.py etc.`. 
Is it possible to load folder with scripts? I didn't found info in docs and after running `mitmproxy --help` command

#### Describe the solution you'd like
`-sf path/to/scripts` that load all files from folder. (`-sf` is `script folder` ) 

#### Describe alternatives you've considered
`-s path/to/scripts/*.py` that load all python files as scripts

Kriechi:
Sounds like a nice and simple feature! PRs are welcome!
I would prefer simply using the "-s" CLI argument. We can check if the given path is a file or directory and act accordingly.
We should be explicit in only loading ".py" files (and potentially symlinks which get auto-resolved) from the given directory to avoid confusion.

PR #4121, opened 2020-07-30
Add wild card support for --scripts option

gorogoroumaru:
#### Description

issue #4112 

When we execute scripts, we have to specify scripts one by one with "--scripts" option.
I added wild card support for --scripts  option ,so that we can specify scripts all at once.

### DIFF
--- a/mitmproxy/addons/script.py
+++ b/mitmproxy/addons/script.py
@@ -19,8 +19,10 @@ import mitmproxy.types as mtypes
 
 
 def get_scripts():
-    scripts = list(chain.from_iterable([glob(re) for re in ctx.options.scripts]))
-    return scripts
+    if getattr(ctx.options, "scripts", None):
+        scripts = [script for re in ctx.options.scripts for script in glob(re)]
+        return scripts
+    return []
 
 
 def load_script(path: str) -> typing.Optional[types.ModuleType]:
@@ -144,13 +146,16 @@ class ScriptLoader:
     def __init__(self):
         self.is_running = False
         self.addons = []
+        self.scripts = get_scripts()
 
     def load(self, loader):
         loader.add_option(
             "scripts", typing.Sequence[str], [],
             """
-                Execute a script. The script name may include wild card. If you include wild card,
-                don't forget to enclose the script name in single or double quotes. Example: mitmproxy -s "some/folder/*.py".
+                Execute a script. 
+                Experimental: If the script name includes a wild card, all matching scripts are executed.
+                Addons created after startup are not picked up.
+                Example: mitmproxy -s "some/folder/*.py".
             """
         )
 
@@ -180,8 +185,16 @@ class ScriptLoader:
                     for evt, arg in eventsequence.iterate(f):
                         ctx.master.addons.invoke_addon(mod, evt, arg)
 
+    @command.command("script.length")
+    def script_length(self):
+        """
+            Return the number of scripts
+        """
+        return len(self.scripts)
+
     def configure(self, updated):
         if "scripts" in updated:
+            # The order of this operation is O(n^2)
             scripts = get_scripts()
             for s in scripts:
                 if scripts.count(s) > 1:
--- a/mitmproxy/addons/script.py
+++ b/mitmproxy/addons/script.py
@@ -122,10 +122,10 @@ class Script:
             try:
                 mtime = os.stat(self.fullpath).st_mtime
             except FileNotFoundError:
-                ctx.log.info("Removing script %s" % self.path)
                 scripts = list(chain.from_iterable([glob(re) for re in ctx.options.scripts]))
                 scripts.remove(self.path)
                 ctx.options.update(scripts=scripts)
+                ctx.log.info("Removed script %s" % self.path)
                 return
             if mtime > last_mtime:
                 self.loadscript()
--- a/mitmproxy/addons/script.py
+++ b/mitmproxy/addons/script.py
@@ -150,8 +150,7 @@ class ScriptLoader:
             "scripts", typing.Sequence[str], [],
             """
                 Execute a script. The script name may include wild card. If you include wild card,
-                don't forget to enclose the script name in single or double quotes. 
-                Example: mitmproxy -s "some/folder/*.py"
+                don't forget to enclose the script name in single or double quotes. Example: mitmproxy -s "some/folder/*.py".
             """
         )
 
--- a/mitmproxy/addons/script.py
+++ b/mitmproxy/addons/script.py
@@ -7,7 +7,6 @@ import types
 import typing
 import traceback
 from glob import glob
-from itertools import chain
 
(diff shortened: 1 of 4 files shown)


mitmproxy/addons/script.py
```diff
@@ -142,7 +144,10 @@ def __init__(self):
     def load(self, loader):
         loader.add_option(
             "scripts", typing.Sequence[str], [],
-            "Execute a script."
+            """
+            Execute a script. The script name may include wild card. If you include wild card,
+            don't forget to enclose the script name in single or double quotes.
+            """
```
Kriechi, line 150: Please add a simple example here.
I assume something like: `some/folder/*.py`

mitmproxy/addons/script.py
```diff
@@ -195,7 +201,9 @@ def configure(self, updated):
 
             ordered = []
             newscripts = []
-            for s in ctx.options.scripts:
+            for s in scripts:
+                if s[-2:] != "py":
+                    continue
```
Kriechi, line 206: Any reason to hard-code and enforce only `.py` files?
I could imagine users want to write scripts (without extensions) and be able to load them into mitmproxy at the same time?

mitmproxy/tools/console/statusbar.py
```diff
@@ -281,7 +283,8 @@ def get_status(self):
         if self.master.options.mode != "regular":
             r.append("[%s]" % self.master.options.mode)
         if self.master.options.scripts:
-            r.append("[scripts:%s]" % len(self.master.options.scripts))
+            scripts = list(chain.from_iterable([glob(re) for re in self.master.options.scripts]))
```
Kriechi, line 286: This snippet seems to appear the second time now - maybe split it off into a function and reuse it?

gorogoroumaru:
Thanks, I applied your suggested changes.

mitmproxy/tools/console/statusbar.py
Kriechi: Is the diff wrong - or did you want to you the new function?

Kriechi:
Please add a changelog entry - as decribed in the Pull Request template.

gorogoroumaru:
Thanks, I added a changelog entry.

mitmproxy/tools/console/statusbar.py
gorogoroumaru: Sorry, I forgot to fix this part. 
Now, it's fixed.

mitmproxy/tools/console/statusbar.py
```diff
@@ -281,7 +284,8 @@ def get_status(self):
         if self.master.options.mode != "regular":
             r.append("[%s]" % self.master.options.mode)
         if self.master.options.scripts:
-            r.append("[scripts:%s]" % len(self.master.options.scripts))
+            scripts = script.get_scripts()
```
mhils, line 287: This is nonobvious, but this change violates some invariants we have. 1) This depends on `ctx` to be populated here, which I think we only guarantee for addons. 2) this change also means we run `glob` on every redraw, which hurts performance because it requires I/O. 3) We also crucially don't want to have a direct dependence on the addon (which we have by importing it).

Long story short, the correct solution here would be for the addon to a) store the list of scripts, 2) expose a command and have that called, see
```
creplay = self.master.commands.call("replay.client.count")
```
above.

mitmproxy/addons/script.py
```diff
@@ -142,7 +148,10 @@ def __init__(self):
     def load(self, loader):
         loader.add_option(
             "scripts", typing.Sequence[str], [],
-            "Execute a script."
+            """
+                Execute a script. The script name may include wild card. If you include wild card,
+                don't forget to enclose the script name in single or double quotes. Example: mitmproxy -s "some/folder/*.py".
```
mhils, line 153: We should explain here what's the purpose of having a wildcard, the fact that it needs to be quoted depends on your console. How about something like this?
```
                Execute a script. 

                Experimental: If the script name includes a wild card, all matching scripts are executed.
                Addons created after startup are not picked up.
                Example: mitmproxy -s "some/folder/*.py".
```

mitmproxy/addons/script.py
```diff
@@ -121,8 +128,7 @@ async def watcher(self):
                 mtime = os.stat(self.fullpath).st_mtime
             except FileNotFoundError:
                 ctx.log.info("Removing script %s" % self.path)
-                scripts = list(ctx.options.scripts)
-                scripts.remove(self.path)
+                scripts = get_scripts()
                 ctx.options.update(scripts=scripts)
```
mhils, line 132: This updates the option to the explicit values, is that intentional?

gorogoroumaru: Yes, I thought it was okay to do so.

mitmproxy/addons/script.py
gorogoroumaru: It looks great.
I will apply this.

test/mitmproxy/addons/test_script.py
```diff
@@ -224,6 +224,9 @@ def test_simple(self, tdata):
 
     def test_dupes(self):
         sc = script.ScriptLoader()
+        with open("one", 'w') as f:
+            f.write("foo")
+
```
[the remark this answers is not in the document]
gorogoroumaru: If there is no "one" file, glob returns an empty list [].
So, the exceptions.OptionsError("Duplicate script") is not raised and test fails.

mitmproxy/addons/script.py
```diff
@@ -16,6 +18,11 @@
 import mitmproxy.types as mtypes
 
 
+def get_scripts():
+    scripts = list(chain.from_iterable([glob(re) for re in ctx.options.scripts]))
```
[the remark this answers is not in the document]
gorogoroumaru: Very nice solution!

mitmproxy/addons/script.py
```diff
@@ -173,12 +182,13 @@ def script_run(self, flows: typing.Sequence[flow.Flow], path: mtypes.Path) -> No
 
     def configure(self, updated):
         if "scripts" in updated:
-            for s in ctx.options.scripts:
-                if ctx.options.scripts.count(s) > 1:
+            scripts = get_scripts()
+            for s in scripts:
+                if scripts.count(s) > 1:
                     raise exceptions.OptionsError("Duplicate script")
```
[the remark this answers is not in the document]
gorogoroumaru: OK, I will add it.

gorogoroumaru:
Thanks, I applied your suggested changes.

gorogoroumaru:
@mhils Could you please review this?

gorogoroumaru:
@mhils Could you please review this?

mhils:
I'm quite busy at the moment, sorry. It definitely needs a dedicated test, and we should use pytest's tmpdir when writing files. I can take a closer look once that's in starting Friday.

PR #4125, opened 2020-08-02
Add option to control stripping newlines from edited bodies

capt8bit:
#### Description

Currently, we strip trailing newlines from edited request/response bodies. But, some Web Apps and API are very picky about trailing newlines being preserved. (I run across a few each year that will view the request body as invalid, if it is missing the trailing newlines.)

This patch adds a console option to configure if the trailing newlines should be stripped or not. I've set the default to "True", because this has been the default behavior for the last 3 years.

#### Suggestion

I feel that it would be more ideal to have the default behavior not make changes to data without the user's knowledge. Otherwise they end up like me, being perplexed why it failed, comparing in the HEX view, and rechecking vim's `endofline` settings over and over. :sweat_smile:

But, I can appreciate preserving the current user experience, so I kept the default as "True".


mhils:
Thanks! 🍰 

So it looks like I introduced that stripping six years ago in https://github.com/mitmproxy/mitmproxy/commit/4767b83726eb82d657825bcf8c00871667d73081, of which I of course don't remember _that_ much. 😅 I think we both agree that _not_ silently modifying contents sounds like a smart thing to do. Do you want to change the default to `False` and add a CHANGELOG entry? 😃  If we find that this breaks lots of workflows, we can still revert the default value (but have documentation why this time).

mhils:
Somehow missed your push, thanks again! 🍰

Issue #3142, opened 2018-05-20
Documentation lacks basic usage examples


stellarpower:
Hi,

mitmproxy looks like a really useful tool but I feel it is really let down by a lack of decent basic documentation. It's great that you go into detail on filter expressions, different modes of operation, and all that, but every section is lacking just a basic introduction and there's no basic user guide on how to use the software. 

For example, the manual page on mitmproxy is just a guide on the keybindings. When I first start using a program, learning how to change the default keybindings is not on my mind. I would just want to see a basic example of how to create a simple setup and perform basic tasks in that setup. So a beginner's guide into starting up the proxy, filtering some content, intercepting and replaying, etc. would be a lot more useful. Similarly, the mitmweb documentation is just a screenshot, and the filter expressions page doesn't even explain how to use them, I'm just presented with a list of flows and no idea how to do anything with them. I'd been using the web interface but atm it's running so slowly on a very powerful laptop that I'm going back to the cli tool.

I know you're pressed for developer time but I really think ten hours spent on documenting how to use the tool for beginners would go a lot further than ten hours of improvements in the codebase at this time.

Thanks

cortesi:
Hi there. Thanks for your feedback - we need to hear this sort of thing to keep improving. I think your critique boils down to a lack of basic tutorials and tool usage in our docs. This is a fair, and something we're thinking about. The type of detailed usage documentation you describe hard to write and maintain. User interfaces change constantly, and whenever you put a screenshot or specific step-by-step instructions into a document, you can expect to have to update it every few releases. So there's a balance to find - ideally we would have just enough information to get people bootstrapped to the point where the technical documentation make sense to them, and where they can begin to rely on the (very complete) built-in documentation for commands, options and the like.

Regarding mitmweb - it's still in beta, and for anything complicated you should still be using the console application. 

I know this is an irritating response to receive to a bug report - but we're always looking for contributors, and we'd be very happy to accept PRs that address some of the specific shortcomings here. 

I've edited the title to this report to be more specific.

stellarpower:
Hi,

Don't worry, it's not irritating, even a basic introduction would be good if everything else is changing so quickly. It's just I feel I'm dropped in and don't know where to begin, I think once you know the basics you can work the rest out. I understand the web interface is still new, I'm just curious as to why it's so slow as I'm running it on a very fast machine. But now I've worked out how to do the basics in the CLI I can use that instead. 

[...]

Issue #2826, opened 2018-01-30
Better UI Testing


mhils:
We're currently doing a bad job at testing the mitmproxy console UI. While there's an argument to be had that test coverage for UI code quickly yields diminishing returns, we should certainly do better than we do right now. It would be very cool if we had a basic framework to drive UI interactions for testing. I'd love to be able to write a test that roughly looks like this:

```python
def test_edit_setcookies():
    ui = TestUI()
    ui.run("console.edit.focus set-cookies")
    ui.press("a foo <tab> bar <tab> <enter>")
    # set a cookie attribute
    assert ui.focus_flow.cookies == ...
```

The syntax is obviously up for debate, but it should get the general idea across. I'm sure @Kriechi would be happy to extend his testing empire to mitmproxy.console as well. :smile:

kajojify:
I will take a look at it :)

mhils:
Fixed: https://github.com/mitmproxy/mitmproxy/blob/main/test/mitmproxy/tools/console/test_integration.py

PR #4127, opened 2020-08-03
tutorial addon to improve onboarding new users

mplattner:
#### Description

Added a draft of the tutorial addon.
To test the current state: start mitmproxy and open http://tutorial.mitm.it

refs #3142 

#### Tasks

- [x] Tutorial layout
- [x] Sample voting app
- [x] Tutorials: View flows / UI intro
- [x] Tutorials: Interception /  Modify
- [x] Tutorials: Replay
- [ ] Tutorials should be valid for `mitmproxy` and `mitmweb`
- [ ] Add tests

#### PR Checklist


mplattner:
@mhils, it currently is not possible to replay the requests to the local voting app (mitmproxy says `Warn: Can't replay live flow.`)
Haven't investigated that yet though.

mplattner:
I played around with how to best format the docs and added a couple of prototypes to explain screenshots in the docs:

* Statically annotated images using a graphics editor: easy to create, hard to maintain (everybody needs the same image editor?)
* Dynamically annotated images using HTML (Bootstrap Tooltips): easy to create, can be versioned in git and are easy to change 
* `asciinema`: just tested it briefly, probably not suitable due to a lack of annotations/comments

@mhils, @Kriechi, @Prinzhorn, the current tutorial app is hosted here such that you can have a look without pulling my branch: http://plattner.me:5000/
(Please note that all users share the same state (vote count) in this setup.)

I'm aware that those options are more for the docs than the tutorial app itself, but can be used for both I guess.
What do you think?

mhils:
This looks like a good first draft. Let's fill it with functionality! :)

I'm on mobile, so abbreviated comments for today:

- the demo app needs more styling to emphasize what it is. Think of something like https://www.w3schools.com/howto/howto_css_browser_window.asp
- When you start writing text, make sure it's super concise. 
- I will fix the WSGI functionality to display a flow, you don't need to take care of that. :)
- Both bootstrap tooltips and static annotations (SVG with image in the background should be shareable/versionable) would work for me. I guess it comes down to matter of taste if you prefer the always-extended text or the clickable/governable tooltip, which allows for longer descriptions.

mplattner:
Cool, thanks for the feedback! Sounds good.

> "_SVG with image in the background_"

Oh wow, I was not seeing the forest for the trees yesterday...

mplattner:
@mhils, I pushed the first tutorial content. Also available via http://plattner.me:5000/

mitmproxy/addons/tutorialapp/templates/mitmproxy/index.html
```diff
@@ -0,0 +1,210 @@
...
+            <li>View HTTP flows and their details, such as HTTP headers.</li>
+            <li>Intercept and modify requests, before passing them to the server.</li>
+            <li>Replay previous requests, i.e., send them to the server again.</li>
+        </ul>
+
+        <p>
+            The tutorial is served by your local mitmproxy instance.
+            This implies that you have already set up your browser to use mitmproxy as a proxy server.
+            Thus, you have already completed the first step 🎉.
+        </p>
```
mhils, line 35: Let's cut this paragraph. That is something we can list in next steps together with the next point, it doesn't matter for the tutorial itself.

mitmproxy/addons/tutorialapp/templates/mitmproxy/index.html
```diff
@@ -0,0 +1,210 @@
...
+            We registered Lucy for the famous global <em>Shocked Animal Challenge</em>.
+            The organizers use a custom web application that allows people to vote for the participants.
+            This voting application is embedded in the right sidebar of this page.
+            Let's see if we can use mitmproxy to get Lucy some extra votes.
+            <!-- Obviously we want Lucy to win. -->
+
+            <!-- One of our clients is the host of the global <em>Shocked Animal challenge</em>.
+            We are the organizers of the global Shocked Animal challenge.
+            Thus, we have created a fictitious web application -->
+        </p>
```
mhils, line 51: This is funny, but too long. Alternative suggestion: _For this tutorial, we will look at the HTTP requests generated by our demonstration voting app on the right. Let's see if you can get your favorite pet some extra votes!_

mitmproxy/addons/tutorialapp/templates/mitmproxy/index.html
```diff
@@ -0,0 +1,210 @@
...
+
+        <p>
+            <a class="btn btn-sm btn-outline-secondary" data-toggle="collapse" href="#collapse-how-tutorial-app-implemented" role="button" aria-expanded="false" aria-controls="collapse-how-tutorial-app-implemented">
+              Optional: How is this tutorial implemented?
+            </a>
+        </p>
```
mhils, line 57: Let's move this to "What's next?". They don't know yet how to start mitmproxy, so that may be overwhelming 😄

mitmproxy/addons/tutorialapp/templates/mitmproxy/index.html
```diff
@@ -0,0 +1,210 @@
...
+        </p> -->
+
+        <h2 id="view-flows" class="pt-3 pb-3 mb-3 border-bottom">mitmproxy's User Interface</h2>
+
+        <!-- <p>We suggest that you position the windows next to ...</p> -->
+
+        <p>
+            First of all,we needs to get familiar with mitmproxy's user interface.
+            Open the terminal window in which you started mitmproxy.
+            You should see the <span class="badge badge-info">Default View</span> of mitmproxy, which shows a table-like list of flows.
```
mhils, line 87: Nit: I think using a badge here doesn't really work out. Maybe just `text-primary`?

mitmproxy/addons/tutorialapp/templates/mitmproxy/index.html
```diff
@@ -0,0 +1,210 @@
...
+
+        <ol>
+            <li>Use the voting app on the right to generate a vote for Lucy.</li>
+            <li>Watch how mitmproxy creates a new line for every vote you submit.</li>
+        </ol>
+
+        <figure class="figure">
+            <img src="/static/images/mitmproxy_flow_list.png" alt="..." class="figure-img img-fluid rounded">
+            <figcaption class="figure-caption text-center">mitmproxy's <span class="badge badge-info">Default View</span></figcaption>
+        </figure>
```
mhils, line 131: Where does the huge black border come from? If it's not too much effort I'd suggest we have no border here (and add an adjustable one using CSS if it looks better)

mitmproxy/addons/tutorialapp/templates/mitmproxy/index.html
```diff
@@ -0,0 +1,210 @@
...
+
+        <p>
+            mitmproxy is controlled using keyboard shortcuts.
+            Every shortcut is internally bound to a command.
+            Advanced users can easily re-map shortcuts using the configuration file.
+            For now, let's focus on some basic shortcuts:
+        </p>
+
+        <ol>
+            <li>Use your arrow keys <span class="badge badge-secondary">🠗</span> and <span class="badge badge-secondary">🠕</span> to move the focus (<code>&gt;&gt;</code>) to the flow requesting <code>/votes</code>.</li>
```
mhils, line 141: Let's use `<kbd>`: https://getbootstrap.com/docs/4.5/content/code/#user-input

If we dislike the style we can change that with CSS, but that's much easier if we use a uniform semantic element. :)

mitmproxy/addons/tutorialapp/templates/mitmproxy/index.html
```diff
@@ -0,0 +1,210 @@
...
+            <img src="/static/images/mitmproxy_flow_list.png" alt="..." class="figure-img img-fluid rounded">
+            <figcaption class="figure-caption text-center">mitmproxy's <span class="badge badge-info">Default View</span></figcaption>
+        </figure>
+
+        <p>
+            mitmproxy is controlled using keyboard shortcuts.
+            Every shortcut is internally bound to a command.
+            Advanced users can easily re-map shortcuts using the configuration file.
+            For now, let's focus on some basic shortcuts:
+        </p>
```
mhils, line 138: Let's go from concrete to the abstract: We can cut this paragraph and move the re-map part either to the end of the section, to "What's next?", or omit it entirely.

mplattner:
Thanks for all your comments. 🙌

mitmproxy/addons/tutorialapp/templates/mitmproxy/index.html
mplattner: Good idea 👍

mitmproxy/addons/tutorialapp/templates/mitmproxy/index.html
mplattner: I'll check what works best. I agree that a badge might emphasize it too much and might be misleading.

mitmproxy/addons/tutorialapp/templates/mitmproxy/index.html
mplattner: Will replace screenshots with cleaner versions at the end. Border comes from setting `padding` to 30 in Windows Terminal - this was needed for my previous examples with tooltips (at the edge of the image for the focussed flow marker `>>`).

mitmproxy/addons/tutorialapp/templates/mitmproxy/index.html
mplattner: Great suggestion!

mitmproxy/addons/tutorialapp/templates/mitmproxy/index.html
mplattner: Fully agree, its a bit "thrown in" there and breaks the didactic "flow".

Prinzhorn:
Thanks you working on this! The following is just the way I had imagined it, feel free to implement whatever you agree with:

My first impression is that there is _way_ too much text. I had also imagined more like a wizard-type of flow instead of the long page. There could be a next/previous button and like a counter or something. The very first screen is a super short introduction and you click like "Start tutorial". And then it starts with the first step. This gives the user a better sense of progress. It would also get things a lot cleaner, because you can remove the menu on the left and the app on the right. These two are not needed for the welcome-screen and only add to distraction when you first open the app. The voting should only be there when it is needed.

> Thus, you have already completed the first step

This is the type of spirit I want :smile: . But instead of having this buried in all the text it could be like a checked off / crossed out item giving the user the impression of having already accomplished something. So like the menu on the left could stay I guess but the items before the current one are checked off or something.

Also not sure what's up with the arrow keys here (I'm on Ubuntu)

---

Selection_643

mplattner:
A new version with a rough draft for intercepting requests is online at: http://plattner.me:5000/mitmproxy/overview
Make sure to refresh without browser cache to reload the CSS file.
For testing purposes I added a GIF for this step. Probably not a good option for several reasons (e.g., user doesn't know where it starts/ends/repeats), but may also have advantages to see something interactive. What do you think of using GIFs? (In general, not the one I added, this one needs improvement ;))

@Prinzhorn, thanks a lot for your feedback!
I have implemented the step-by-step format, added a progess bar, and shortened the text here and there. 👍
Hiding left and right sidebar is something I've now left as it is - this can be optimized at the end if needed.

@mhils, I have resolved most of your comments. :)
My arm is still causing trouble when I work for several hours. Sorry for the slow progress...

Prinzhorn:
Nice! I like the progress bar and the steps.

> Btw, due to the step-by-step format on separate pages, the user now sees additional requests to `tutorial.mitm.it` during the tutorial. We might think of hiding those, not sure yet.

Either show all of them or none. I think it'd be more confusing to only hide some of the requests.

> What do you think of using GIFs? (In general, not the one I added, this one needs improvement ;))

I've looked into asciinema as well. I think it would be great, because of smaller file size, interactivity, crispyness and being able to select/copy text. You said there where no annotation/comments. Could you elaborate how that differs from jpeg/gif/video? We could use the [JavaScript API](https://github.com/asciinema/asciinema-player#controlling-the-player-programmatically) to dynamically show/hide annotations at different timestamps. You can wrap the player with a `postion: relative` element and then place the tooltips absolutely on top where you need them. If you need help with the frontend JavaScript I could write a script that reads `data-annotate-from` and `data-annotate-to` attributes on the annotations and hooks it to the asciinema progress. I guess that would work? So the user can pause and view the annotation.

```html
<style>
  .asciinema {
    position: relative;
  }

  .annotation {
    position: absolute;
    z-index: i-guess;
  }
</style>

<div class="asciinema">
  <asciinema-player src="/demo.cast"></asciinema-player>
  
  <!--Show this annotation from 2s to 6s in the "video"-->
  <div class="annotation" style="left: 30px; top: 130px;" data-from="2" data-to="6">Here should be the dot indicator with the tooltip</div>

  <!--Show this annotation from 8.3s to 16s in the "video"-->
  <div class="annotation" style="left: 30px; top: 130px;" data-from="8.3" data-to="16">Here should be the dot indicator with the tooltip</div>
</div>
```

mplattner:
@Prinzhorn, thanks again.

> Either show all of them or none. I think it'd be more confusing to only hide some of the requests.

Sure, if we do hide `tutorial.mitm.it` flows, we need to be consistent.

> Could you elaborate how that differs from jpeg/gif/video?

It doesn't really differ, the annotation issue is largely the same with jpeg/gif/video. It might be "easier" to manually put an annotation on an image or video. (easier as in 'no/less coding involved'.)

Having asciinema with annotations would be cool indeed. However, for now I'll finish the tutorial with static images.
I guess I should use my already very limited daily working time for mitmproxy, instead of extending other projects. ;)
We can always plan to add dynamic asciinema captures in the future.
What do you think, @mhils and @Prinzhorn?

mplattner:
I have now implemented annotations in asciinema. See the following examples (refresh your cache first):

1. http://plattner.me:5000/mitmproxy/intercept_requests: Prototype
2. http://plattner.me:5000/mitmproxy/modify_requests: more advanced example that also highlights the instruction in the list above.

I think overall it works well and it might be a good idea to keep it - thanks again for the hints, @Prinzhorn.
However, I'm not sure yet how it affects the flow and structure of the existing content. What do you think of having text (paragraphs) and instruction boxes (including the asciinema "video") interleaved, i.e., instruction boxes have asciinema below such that users see the instructions in action. This way, users have to choice whether they want to read and/or watch the instructions.

mitmproxy/addons/tutorialapp/static/tutorial.js
```diff
@@ -0,0 +1,76 @@
...
+    console.log("playing at " + this.currentTime);
+    var playStartPosition = this.currentTime;
+    var playStartDate = new Date();
+    var annotations = this.annotations;
+    var instructions = this.instructions;
+
+    timer = setInterval(function(){
+        var now = new Date();
+        updateAnnotations(
+            playStartPosition + (now - playStartDate) / 1000,
```
Prinzhorn, line 30: You can just pass `this.currentTime` as first parameter and remove these calculations. No need to manually track time.

Prinzhorn:
>     2\. http://plattner.me:5000/mitmproxy/modify_requests: more advanced example that also highlights the instruction in the list above.

That's sick. What if we do this:

1. Add timestamps to the instructions (e.g. `2. Generate some flows (00:06).`). Clicking `00:06` sets `currentTime` of asciinema, like timestamps work on YouTube.
2. Turn the list items into `<details>` elements and if you want to you can expand the item and get the detailed instructions as text right there. I think this might be a good way to mix asciinema and text and let the user decide what route they want to go?

What I don't like about 2 is that the `<details>` element hides the text initially. I mean that's the point but it makes the instructions less scannable and I think it breaks ctrl+f search as well.

So maybe we have the asciinema part with short instructions and timestamps at the top (in a box just like http://plattner.me:5000/mitmproxy/modify_requests) and then below we have regular flowing text with headlines that basically match the list items. So if you know what you're doing you get everything you need right at the top with a sick annotated asciinema, but if you need more details or want to quickly ctrl+f it's right there below.

mhils:
I should briefly disappear more often - very very cool. This looks remarkably awesome. 🍰 

Now you two have to stop me, but how about we take this one step further and automate the generation of these asciinema records?

```python
def record_modify_response(r: Recorder):
    r.step("Start mitmproxy.")  # not shown in the capture, but used to generate the "Instructions" thing.
    r.type("mitmproxy\n")  # this sleeps between keystrokes to make it look human
    r.step("Generate some flows.") 
    r.sleep(1)  # artificial sleep
    r.step("Press <kbd>i</kbd> to prepopulate mitmproxy's command prompt with <code>set intercept ''</code>.") 
    r.type("~u /vote/\n")
    # ...
```

There would be two recorders:

1. The first recorder uses this as a playbook to play inputs to asciinema, i.e. when we call `type`, it plays that as stdin input to asciinema with proper timings.
2. The second recorder would record all sections and their timings to generate the annotations with proper timestamps.

Long story short, this would make a small little tool that generates annotated asciinema records from playbooks. This does sidetrack us a bit, but would also result in a tremendously useful little tool. For example, if we change the look of mitmproxy's flowlist, we don't have to regenerate our tutorial and adjust timestamps manually.

There also is some prior work in that area: https://github.com/redhat-cip/spielbash/tree/master/spielbash - Python 2 unforunately and without annotations. 😄 

@mplattner, thoughts? :)

mplattner:
Many thanks, @mhils and @Prinzhorn, I'm happy you like it 🎉 

@Prinzhorn, I like the idea of clickable timestamps.

> What I don't like about 2 is that the `<details>` element hides the text initially. I mean that's the point but it makes the instructions less scannable and I think it breaks ctrl+f search as well.

Why use `<details>` at all if it should be shown initially? To allow the user hiding the instructions?

> [...]and then below we have regular flowing text (and screenshots? Or maybe a non interactive paused version of the relevant asciinema, which we already have anyway?) with headlines that basically match the list items.

Hm, I'm not sure if we need the same content once more below the asciinema. What could work well is a short intro (text) to five the user some context, followed by the asciinema. Alternatively, we could break up the instructions in multiple asciinemas, e.g., `[intro], [asciinema], more context, [asciinema], ...`. I haven't thought about it enough though, but most probably, just one contiguous asciinema per section is better...

@mhils, oh yeah, programmatically generating the asciinema recordings would be super cool, I'll look into this. 👍

Prinzhorn:
> Long story short, this would make a small little tool that generates annotated asciinema records from playbooks. This does sidetrack us a bit, but would also result in a tremendously useful little tool. For example, if we change the look of mitmproxy's flowlist, we don't have to regenerate our tutorial and adjust timestamps manually.

@mhils my first thought was that this path might lead us to integration/ui tests as well? Not sure if any exist right now, not familiar with the test setup.



> Why use `<details>` at all if it should be shown initially? To allow the user hiding the instructions?

@mplattner sorry I should not have written down my thought process, just the result.

> What could work well is a short intro (text) to five the user some context, followed by the asciinema.

Yes, that's what I was thinking but have these in `<details>` maybe? So that above the asciinema there's just a short list of steps that you can expand if you want. But it's nowhere near the amount of text and instructions that you can get below the asciinema. This is basically for two types of users or two types of learning-styles.

mhils:
> > Long story short, this would make a small little tool that generates annotated asciinema records from playbooks. This does sidetrack us a bit, but would also result in a tremendously useful little tool. For example, if we change the look of mitmproxy's flowlist, we don't have to regenerate our tutorial and adjust timestamps manually.
> 
> @mhils my first thought was that this path might lead us to integration/ui tests as well? Not sure if any exist right now, not familiar with the test setup.


Definitely something we should do, but don't do yet (#2826).

> 
> > Why use `<details>` at all if it should be shown initially? To allow the user hiding the instructions?
> 
> @mplattner sorry I should not have written down my thought process, just the result.
> 
> > What could work well is a short intro (text) to five the user some context, followed by the asciinema.
> 
> Yes, that's what I was thinking but have these in `<details>` maybe? So that above the asciinema there's just a short list of steps that you can expand if you want. But it's nowhere near the amount of text and instructions that you can get below the asciinema. This is basically for two types of users or two types of learning-styles.

I'd generally be in favor to keep tutorial text very short, but no strong opinions. :)

mplattner:
I was fiddling around with [redhat-cip/spielbash](https://github.com/redhat-cip/spielbash), and some similar tools:

* https://github.com/Malinskiy/spielbash: Ruby-based version of redhat-cip/spielbash with more features, e.g., a `pre-run` step, for example to start a clean Ubuntu in Docker to run all commands, see [an example](https://github.com/Malinskiy/spielbash/blob/31aa33c0bf38bb9999865c51d8513dcaa2dac7cc/spec/fixtures/files/scenario_1.yaml). Seems to be the most sophisticated tool, however, I don't think that we want to add Ruby to our stack... Plus, I've never written Ruby and it might take a while to add the "instructions" recorder. ;)
* https://github.com/stonecharioteer/asciinema-director: simple implementation, based on Python's `pexpect`
* https://github.com/paxtonhare/demo-magic: just a bash script. not enough features and flexibility. i think it does not support sending arrow keys etc.
* https://github.com/ficoos/termscript
* https://stackoverflow.com/a/63080929/2026651: suggests some other tools that I have briefly looked at

I was having some issues in Windows (WSL2) that the beginning of the recording is messed up. Might be due to my tmux config, will further test that. Will try to setup a Docker container with a Python3-compatible version of [redhat-cip/spielbash](https://github.com/redhat-cip/spielbash), as custom shell configs (fonts with icons, tmux) do cause trouble. Do you have a setup in mind how the generation of recordings should work?

One (among other) open issue is that we need to generate flows during the recording, i.e., spawn a recording-"external" `curl`.
(Or something like `lynx`, if we want that all requests to `tutorial.mitm.it` (incl. JS/CSS etc.) are in the flow list to match the experience that the user has while taking the tutorial.)

```python
def record_modify_response(r: Recorder):
    # [...]
    r.sleep(1)  # artificial sleep
    r.step("Press <kbd>i</kbd> to prepopulate mitmproxy's command prompt with <code>set intercept ''</code>.") 
    r.type("~u /vote/\n")
    r.external("curl --proxy localhost:8080 http://tutorial.mitm.it/vote/cat") # <-- something like this is needed
```

That said, automating this in a way that the recordings match the quality/flexibility of a manual recording that can be tweaked/adjusted probably isn't trivial. Taking into account the remaining period of my GSoC I don't know what's the best way forward.

Suggestions? Thoughts? Thanks!

mitmproxy/addons/tutorialapp/static/tutorial.js
mplattner: You are right, that's an unnecessarily complex implementation :) Thanks and sorry, I forgot to reply to this comment.

mplattner:
Regarding the layout, I thought something like this might work well:
image

mplattner:
Brief update: new layout is implemented, working on the Docker container to generate the recordings at the moment. Hope to show you something tomorrow.

mplattner:
I have added a draft for the recording tool. I guess it's best explained by an example:

```
~/mitmproxy/mitmproxy/addons/tutorialapp> ./record_tutorials.py
Recording intercept_response.cast: Done

~/mitmproxy/mitmproxy/addons/tutorialapp> asciinema upload intercept_response.cast
View the recording at: https://asciinema.org/a/Jj7Sv8oY69zZu5nxZE0xooYQh
```

The resulting recording can be found at https://asciinema.org/a/Jj7Sv8oY69zZu5nxZE0xooYQh

The recording "script" see [record_tutorials.py](https://github.com/mitmproxy/mitmproxy/blob/c13abfe9da2a088d40c4ab365abaae0acc3bcf95/mitmproxy/addons/tutorialapp/record_tutorials.py) file.

It should be trivial to add the instructions timing recorder using `CliRecorder.current_time`, e.g. something like `instructions.append([currtime, title, instruction])` would work.

What do you think? How would we best integrate this into mitmproxy?
Should I create a PyPi package with only the `CliRecorder` (this part might be helpful to others), and then add it as a dependency to mitmproxy and put the actual scripts & the instruction recorder into the mitmproxy repo? Does that make sense?

mhils:
Nice! This is super awesome. 

> What do you think? How would we best integrate this into mitmproxy?

The easiest and fastest approach would probably be to move this into [docs/scripts](https://github.com/mitmproxy/mitmproxy/tree/master/docs/scripts) for now. This makes it easy because we don't introduce any additional dependency management.

If you want to brush up your GitHub profile you can also move this to your your own repo + PyPI package and reference that in mitmproxy, but let's first get the GSoC-relevant bits done. :)

mplattner:
Sounds good, thanks! 😄
I found that tmux has a `display-message` command to show text in the status bar, and since version 3.2-rc2 there is even a `display-popup` command to show an overlay over the current view. I think this would be quite cool to integrate. It tried it this afternoon but there seems to be an issue that the output is not recorded correctly. I hope to get this working soon, otherwise I'll continue with the current feature set.

image

mplattner:
Phew, finally I got this working. 🎉 Using a docker container we can now create something like the recording below. Unfortunately I don't think this works "headless", i.e., a graphical environment is needed. Without one I always had issues with the recording.

Recording: https://asciinema.org/a/w9ybHUlUCYb8zfxctNsXjBDnZ

Code:
```python
def record_advanced(d: CliDirector):
    tmux = d.start_session()
    window = tmux.attached_window
    pane = tmux.attached_pane

    window.set_window_option("window-status-current-format", "mitmproxy Tutorial: Advanced Demo")

    d.start_recording("intercept_response_advanced.cast")
    d.pause(2)
    d.message("Step 1: Start mitmproxy.")
    d.exec("mitmproxy")
    d.pause(4)
    d.popup(
"""#[fg=red]# mitmproxy Tutorial

This is the default view of mitmproxy.
Here you see a list of flows. [...]
"""
    )
    d.message("Step 2: Press '?' to display keyboard shortcuts.")
    d.type("?")
    d.pause(3)
    d.press_key("Right")
    d.pause(2)
    d.message("Step 3: Press 'q' to go back to the default view.")
    d.pause(1)
    d.press_key("q")
    d.message("Testing fast refeshes: open and close help.")
    d.press_key("?")
    d.press_key("q")
    d.press_key("?")
    d.press_key("q")
    d.press_key("?")
    d.press_key("q")
    d.press_key("?")
    d.press_key("q")
    d.press_key("?")
    d.press_key("q")
    d.message("Step 4: Press 'q' to exit mitmproxy.")
    d.type("q")
    d.message("Step 5: Press 'y' to exit mitmproxy.")
    d.type("y")
    d.pause(2)
```

mplattner:
I further improved the code and added recording+timing of HTML instructions.

I find it hard to come up with a sensible way to have the video & separate instructions (as we discussed) so that the user can decide whether to read or/and watch. If video and text do significantly differ, it becomes really hard to follow, as you can see in http://plattner.me:5000/mitmproxy/user_interface. Please note that the timing of the instructions is not ideal here at the moment.

Thus, I created a version with less text and *the same* instructions written *and* in the video, see http://plattner.me:5000/mitmproxy/whats_next. I think this variant is comprehensible way better. Is that the amount of text you expect per page/step?
However, a drawback is that reading the instructions *alone* is probably not enough to make sense of it.

For both variants the asciinema generation, instruction timing, and the integration into the tutorial is fully automated.

@mhils, would be nice to have a call soon to plan the final submission for GSoC. Let's discuss that today. 😃

mplattner:
I added tutorial content (ascii cast + instructions) for `Step 1: User Interface`, `Step 2: Intercept Requests`, and `Step 3: Modify Requests`. Available at http://plattner.me:5000/mitmproxy/overview (Cache refresh is needed.)

mplattner:
New version is available at http://plattner.me:5000/mitmproxy/overview (Cache refresh is needed.)
I feel that the mitmproxy tutorial is now kind of complete. (Excl. polishing, e.g., improve instruction texts and timings)
Happy to hear your thoughts. 😃

Prinzhorn:
That looks amazing. I'll just take a quick look at everything and dump my thoughts (including polish) here

The same id is used twice, I assume this messes with your JS. Or at least it currently does the correct thing by chance.

Selection_650
Selection_651

On the overview page there is no `#asciinema-player`

Selection_652

I'm testing on my vertical monitor (still 1440 wide!) and this should probably be responsive and move the text below asciinema. Maybe to make more room the left hand navigation could be at the top and vertical above the progress bar? E.g. "Overview > 1. User Interface > 2. Intercept Requests > ...". Or could we use a Bootstrap collapsible and let the user collapse the sidebar? You can navigate back/forward using the buttons any

Selection_653

Well, that wasn't much ;)

mplattner:
Sorry @Prinzhorn, I only saw your feedback today. Thanks a ton 🙌 
@mhils and me decided to go a different route and ditch the dedicated tutorial app (incl. the votbing app).
Instead, we add the tutorials directly into the main docs. This should significantly increase the number of users that benefit from it.
Please see #4194 for the new version. I'll close this PR.

Issue #4109, opened 2020-07-28
intercept_active toggle is not logical (and does not work)


rugk:
#### Problem Description
1. `intercept_active` has no documentation.
  It's documented [as](https://docs.mitmproxy.org/stable/concepts-options/):
   > Intercept toggle
   
   This is no doc, because it merely repeats the option's name.

   So I _guess_ it is supposed to disable all proxing and just forward requests (which is useful to do temporarily) – I do set `allow_hosts`  to `[*]` usually, because that works.

2. But when changing `intercept_active` to `true` in the web view, nothing visibly changes. It still intercepts everything and shows it to me…
3. The name is not logical – or the default value. If `intercept_active` defaults to _false_, it would mean it "intercept is not active", while `true` then is "intercept is active". As intercept is, however, _always active by default_ the default should be `true`, I guess…

## STR

transparent mode
`mitmweb --anticomp --anticache --mode transparent --web-port 1234 --save-stream-file "+$FLOWSAVE" --web-host 0.0.0.0 ` (where `$FLOWSAVE` is a file path)

#### System Information
```
$ mitmproxy --version
Mitmproxy: 5.2
Python:    3.7.3
OpenSSL:   OpenSSL 1.1.1d  10 Sep 2019
Platform:  Linux-5.4.51-v7l+-armv7l-with-debian-10.4
```

mhils:
Thanks! That's definitely a documentation issue. `intercept_active` toggles if the `intercept` option is in effect, we need a better dosctring here. :)

rugk:
I don't understand that explanation – yet again hehe, but I guess you mean the `intercept` filter (option `intercept`) documented one line before that actual option.

PR #4133, opened 2020-08-06
Fix intercept toggeling doc

rugk:
Fixes https://github.com/mitmproxy/mitmproxy/issues/4109

#### Description

I hope this also fixes the docs, because it is the only instance where I've found that string. The [docs](https://github.com/mitmproxy/mitmproxy/tree/master/docs) seem to somehow autogenerate the options tabel/text, so I don't know what else may need editing.
Note you also have no `CONTRIBUTING.md` or so that could explain it to me.

#### Checklist

 - [ ] I have updated tests where applicable. N/A
 - [ ] I have added an entry to the CHANGELOG. N/A (for such a small change IMHO)

### DIFF
--- a/mitmproxy/tools/console/defaultkeys.py
+++ b/mitmproxy/tools/console/defaultkeys.py
@@ -26,7 +26,7 @@ def map(km):
     km.add("ctrl f", "console.nav.pagedown", ["global"], "Page down")
     km.add("ctrl b", "console.nav.pageup", ["global"], "Page up")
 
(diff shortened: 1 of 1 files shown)


mhils:
Thanks for the PR!

FWIW, our contribution instructions are at https://github.com/mitmproxy/mitmproxy#contributing, but don't cover this detail. https://github.com/mitmproxy/mitmproxy/blob/master/docs/src/content/concepts-options.md#available-options points to  https://github.com/mitmproxy/mitmproxy/blob/master/docs/scripts/options.py, so that would be one way to determine how it's built.

Issue #4128, opened 2020-08-04
get [Errno 92] Protocol not available when use --mode transparent


cw4t:
#### Steps to reproduce the behavior:
1. mitmdump --mode transparent --listen-host 10.1.10.160 --listen-port 8080
2. WebBrowser enterer: http://10.1.10.160:8080 
3. 10.1.10.160 Occured:  

```
csock.getsockopt(socket.SOL_IP, SO_ORIGINAL_DST, 16)
Traceback (most recent call last):
  File "/home/saicops/.pycharm_helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<input>", line 1, in <module>
OSError: [Errno 92] Protocol not available
```

#### System Information
Paste the output of "mitmproxy --version" here.
Mitmproxy: 5.1.1
Python:    3.6.8
OpenSSL:   OpenSSL 1.1.0j  20 Nov 2018
Platform:  Linux-3.10.0-957.12.2.el7.x86_64-x86_64-with-centos-7.6.1810-Core

cw4t:
```
Please lodge a bug report at: https://github.com/mitmproxy/mitmproxy
10.191.6.187:64479: Traceback (most recent call last):
  File "/opt/yearn/lib64/python3.6/site-packages/mitmproxy/proxy/server.py", line 121, in handle
    root_layer()
  File "/opt/yearn/lib64/python3.6/site-packages/mitmproxy/proxy/modes/transparent_proxy.py", line 22, in __call__
    layer()
  File "/opt/yearn/lib64/python3.6/site-packages/mitmproxy/proxy/protocol/tls.py", line 285, in __call__
    layer()
  File "/opt/yearn/lib64/python3.6/site-packages/mitmproxy/proxy/protocol/http1.py", line 83, in __call__
    layer()
  File "/opt/yearn/lib64/python3.6/site-packages/mitmproxy/proxy/protocol/http.py", line 190, in __call__
    if not self._process_flow(flow):
  File "/opt/yearn/lib64/python3.6/site-packages/mitmproxy/proxy/protocol/http.py", line 321, in _process_flow
    f.request.host = self.__initial_server_address[0]
TypeError: 'NoneType' object is not subscriptable

10.191.6.187:64479: clientdisconnect
```

Prinzhorn:
[...]

PR #4135, opened 2020-08-07
Prevent transparent mode from connecting to itself in the basic cases, references #4128

Prinzhorn:
#### Description

This is obviously not a security feature, it just prevents the most obvious shoot-yourself-in-the-foot type of thing. For example if I don't specify listen_host and then connect to the proxy via it's local IP (192.168.0.6 right now) it cannot detect that this is actually it's own. We might be able to extend this later tho if DoS is a relevant concern. Right now a single `curl` command will drain all resources and for example result in "OSError: [Errno 24] Too many open files". You can even exit the curl process and mitmproxy will keep doing it's endless loop.

Are there integration tests for this kind of thing? Might be worth setting up using Docker containers (maybe using compose) to test different network combinations. I don't really feel comfortable modifying these core things and don't know if it actually still works as before.

#### Checklist

### DIFF
--- a/mitmproxy/proxy/protocol/base.py
+++ b/mitmproxy/proxy/protocol/base.py
@@ -107,9 +107,14 @@ class ServerConnectionMixin:
         """
         address = self.server_conn.address
         if address:
+            forbidden_hosts = ["localhost", "127.0.0.1", "::1"]
+
+            if self.config.options.listen_host:
+                forbidden_hosts.append(self.config.options.listen_host)
+
             self_connect = (
                 address[1] == self.config.options.listen_port and
-                address[0] in ("localhost", "127.0.0.1", "::1")
+                address[0] in forbidden_hosts
             )
             if self_connect:
                 raise exceptions.ProtocolException(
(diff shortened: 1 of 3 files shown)


mitmproxy/proxy/protocol/base.py
```diff
@@ -107,9 +107,14 @@ def __check_self_connect(self):
         """
         address = self.server_conn.address
         if address:
+            forbidden_hosts = ["localhost", "127.0.0.1", "::1"]
+
+            if self.config.options.listen_host:
+                forbidden_hosts.append(self.config.options.listen_host)
```
mhils, line 113: Could we not just always include `self.config.options.listen_host`? If that is empty, the `address[0]` check should also never be true.

Prinzhorn: That's preference I guess and it was my very first thought as well. But I don't like having an empty string in the `forbidden_hosts` list. It just doesn't feel right. But sure the check would still work because `address[0]` won't be an empty string. I just like to write very explicit code that doesn't require a second thought when looking at it. E.g. when someone later looks at it they might think "wait a second, can't listen_host be empty?" and they need to put work into understanding that for the end result it doesn't matter. Basically I try to prevent humans from playing computer in their heads.

mhils: Fair enough, I'm convinced. Thanks! 🍰

Issue #4092, opened 2020-07-18
[mitmweb][Request] Timestamp column + sorting + keep scrolling to bottom


procule:
#### Is your feature request related to a problem? Please describe.
- Getting the timestamp of a request requires to select the flow and go the Details tab
- If column is selected for sorting, there is no way to get back to a timeline related sorting
- Following the flows requires to always scroll to the bottom to get the newest.

#### Describe the solution you'd like
- Add a Timestamp column related to the `First request byte` (origin of the timing of the flow)
- Add a sorting functionality for that column
- Add an option to keep the scroll bar at the bottom or a reverse order for the flows (added at the top instead of the bottom)

#### Describe alternatives you've considered
- Injecting Javascript to modify the scrollbar with the `scrollTo(x,y)` method from the `flow-table` div element. Tested via the JS console and working. But since the web interface is based on React, that would require modifying the source code directly.

#### Additional context
I am adding `mitmweb` as an extra feature for an existing transparent proxy web interface. It is mostly for realtime analysis of API calls. Multiple people will need to have access with their own environment (and filtering). Using the provided release binaries is a must to manage those environment without dealing with Python venv. I tried to modify the React code but I didn't find a way to recreate binaries like you provide in your releases.

Since the analysis needed are based on time and realtime calls, having to always scroll down manually to follow the new flows and also having to select flows one by one to get the timestamp, those feature requests would be nice additions to your software.

Actually, I was reading the code and the options to be sure it was not already there since for me and I guess a lot else, it is a must have feature ;-)

Thanks,
O.

rugk:
I agree a timestamp column (maybe in human-readable or as you say from the first byte) would be very useful in `mitmweb`.

mhils:
Feel free to send a PR and you'll have it in the official binaries. PRs are welcome! :) 

Regarding our binaries, see https://stackoverflow.com/questions/62334867/how-are-the-pre-compiled-mitmproxy-binaries-actually-built.

rugk:
As for "keep scrolling to bottom" AFAIK `console_focus_follow` should do it. The doc is sparse, but you can enable it in the web interface and it worked in my test.
Enabling that from the command line, did not – for whatever reason: https://github.com/mitmproxy/mitmproxy/issues/4108

mellester:
adding 'Timestamp' to web_collums gives me the timestamp of the `First request byte`
however I cant sort the collum

HoffmannP:
[...]

PR #4138, opened 2020-08-08
Timestamp column + sorting

gorogoroumaru:
#### Description

issue #4092 

mitmweb has no functionality to show timestamp and sort by timestamp.
I added the functionality.


mhils:
Thanks for the PR. I'm sorry if I wasn't clear in #4092, but it is a deliberate design decision to not include a timestamp column in the default view. It just doesn't work as well on smaller displays. If we reach the point where the user can add/remove columns a timestamp column is one of the first things we want to add though!

Long story short, we'd need to following things:

1. Add an option to specify columns here: https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/tools/web/webaddons.py#L6
2. Make the web ui respect that option.
3. Add a timestamp column (off by default)
4. Add an easy way to dynamically add/remove columns from mitmweb.

mhils:
Closing this in favor of #4188. 😃

Issue #4132, opened 2020-08-06
Mitmweb prevents all kinds of hot keys


rugk:
#### Problem Description
In mitmweb you cannot use your "usual" hot key features anymore or so, as they all seem to be catched by mitmproxy – even if mitmproxy does not even do anything else with that.

#### Steps to reproduce the behavior:
1. Press F5 in your browser with mirmweb to reload, or Ctrl+L to focus the location bar.
-> It does not work.

Using the Firefox debugger I can see it somehow catches all keyboard events. (I pressed Ctrl+F5 there, which really should nearly never be catched by a website.)
![grafik](https://user-images.githubusercontent.com/11966684/89574248-22aac900-d82c-11ea-9749-23121aa81c1e.png)

#### System Information
Firefox 79

Mitmproxy: 5.2
Python:    3.7.3
OpenSSL:   OpenSSL 1.1.1d  10 Sep 2019
Platform:  Linux-5.4.51-v7l+-armv7l-with-debian-10.4

mhils:
> In mitmweb you cannot use your "usual" hot key features anymore or so

Could you clarify what that means? Does that mean hotkeys don't work for you at all (e.g., you cannot press `r` to replay a selected flow?), or only some hotkeys (in particular custom ones) don't work?

rugk:
Only the ones of mitmproxy work, all browser hot keys are overwritten – like F5 for reloading or F11 for fullscreen.
Actually testing some hot keys I know, it seems the shortcuts with <kbd>Ctrl</kbd> work, but the function keys (F…) are certainly all overwritten.

Prinzhorn:
I guess we should only `preventDefault` if we actually consume the keydown. Currently it prevents F5, F11 etc.

https://github.com/mitmproxy/mitmproxy/blob/218e69ddb2aecd2fcd224b06bf6bbde23156fd3e/web/src/js/ducks/ui/keyboard.js#L7-L15

rugk:
Well, that should be easy to fix.
My try in https://github.com/mitmproxy/mitmproxy/pull/4140 (untested)

PR #4140, opened 2020-08-09
Only preventDefault if key is actually handled

rugk:
#### Description

Fixes https://github.com/mitmproxy/mitmproxy/issues/4132

**This is untested**

Note I do not know why you need to return a lambda function there, but I guess this is some React-thing I don't know.

#### Checklist

### DIFF
--- a/web/src/js/ducks/ui/keyboard.js
+++ b/web/src/js/ducks/ui/keyboard.js
@@ -6,13 +6,8 @@ import * as modalActions from "./modal"
 
 export function onKeyDown(e) {
     //console.debug("onKeyDown", e)
-    if (e.ctrlKey || e.metaKey) {
-        return () => {
-        }
-    }
     let key      = e.keyCode,
         shiftKey = e.shiftKey
-    e.preventDefault()
     return (dispatch, getState) => {
 
         const flows = getState().flows,
@@ -124,9 +119,12 @@ export function onKeyDown(e) {
                 break
             }
 
-
+            // ignore all other keys
             default:
                 return
         }
+        
+        // prevent key press if it was handled
+        e.preventDefault()
     }
 }


web/src/js/ducks/ui/keyboard.js
```diff
@@ -124,9 +119,12 @@ export function onKeyDown(e) {
...
             }
 
-
+            // ignore all other keys
             default:
                 return
         }
+        
+        // prevent key press if it was handled
+        e.preventDefault()
```
rugk, line 128: Oh I see the problem here. `e` is not defined in this scope as lambda functions create a new scope IIRC.

Okay, dunno how to solve this. Pass `e` into the function or just remove this lambda function somehow and get the other variables somewhere else?

Prinzhorn:
> Note I do not know why you need to return a lambda function there, but I guess this is some React-thing I don't know.

Same. It's a redux thing and I can't wrap my head around it.

I assume it happens asynchronously. `e` is available in this scope, but you need to `preventDefault` synchronously, so it's too late inside the lambda function.

As a quick fix for your issue we could just add the function keys and call it a day, not sure if @mhils would be excited about that. Something like this:

```js
if (e.ctrlKey || e.metaKey || /^F[0-9]+$/.test(e.key)) {
```

rugk:
Ohh yet another workaround. Whitelisting ctrl and meta is already a workaround… :see_no_evil:

mhils:
Yes, welcome to the world of redux, redux-thunk, and pain. Lot of the latter I guess 😄 

The problem is that we need the application state in the key handler, and redux-thunk requires that callback. One workaround would be to pass the entire redux store to [ProxyApp](https://github.com/mitmproxy/mitmproxy/blob/master/web/src/js/components/ProxyApp.jsx) as a prop, but we need to make sure that this doesn't cause a repaint on every state change. It's honestly a long time since I looked at any of this. If someone wants to take a look at this, please do!


I'm closing this here for now, but we can merge a PR that deals with function keys as a workaround if that helps your particular use case. :)

PR #4144, opened 2020-08-11
Support filters for request or response body

sanlengjingvv:
#### Checklist


mhils:
Thanks! I appreciate the effort, but the point here is that we exactly _don't_ want to send the entire message body over the API. We want body search to happen on the backend, which is what makes it nontrivial.

PR #4145, opened 2020-08-11
[Quick Help Wanted] mitmweb: use native_web_app to open browser

mhils:
### TL;DR: How you (yes you, you reading this!) can help

1. Download the snapshots for your platform: https://snapshots.mitmproxy.org/branches/native-web-app/.
2. Run the mitmweb binary.
3. Post a comment here with your OS, browser version, and mitmweb screenshot.

#### Detailed Description

I have a new weekend project that makes it possible to build Electron-style apps without shipping the behemoth that is Electron: https://pypi.org/project/native-web-app/. This makes mitmweb look much more like a proper app on my system:

<p align="center">

</p>

Now that makes me very excited, but correctness highly depends on platform/browser version and we need more testing before we can confidently ship this. If you read this, please follow the instructions at the top! 😃 

#### Checklist

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,7 @@ Unreleased: mitmproxy next
     * Fix links to anticache docs in mitmweb and use HTTPS for links to documentation (@rugk)
     * Updated typing for WebsocketMessage.content (@prinzhorn)
     * Prevent transparent mode from connecting to itself in the basic cases (@prinzhorn)
+    * Make mitmweb feel more like a native app, hide browser controls (@mhils)
 
     * --- TODO: add new PRs above this line ---
 
--- a/mitmproxy/tools/web/webaddons.py
+++ b/mitmproxy/tools/web/webaddons.py
@@ -1,5 +1,4 @@
-import webbrowser
-
+import native_web_app
 from mitmproxy import ctx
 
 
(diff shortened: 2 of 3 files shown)


mplattner:
I've previously used the browser's "app" mode to have separate windows for apps like Overleaf or Todoist.
Your use for Electron-less views with the `native web app` is super cool. 👍 
I hope this can be implemented across OSes and browsers.

It worked instantly for me, using
* Browser: Chrome 84.0.4147.105 (my default browser is Firefox, but Chrome was started)
* OS: Windows 10 Pro Version 2004 (Build 19041.388)

image

Prinzhorn:
Nothing happens for me on Ubuntu 20.04.1. I can launch `/usr/bin/google-chrome` just fine though. I can also open `google-chrome --app=https://www.prinzhorn.it/` fine.

```
 $ ./mitmweb 
Web server listening at http://127.0.0.1:8081/
/usr/bin/google-chrome: /tmp/_MEIgR25Bn/libz.so.1: version `ZLIB_1.2.9' not found (required by /lib/x86_64-linux-gnu/libpng16.so.16)
```

Selection_645

I honestly don't see a benefit in doing this, but I'm neutral to adding it to mitmweb. I don't see a reason to remove the browser UI and make something look like a desktop app if it doesn't use any features of a native app. But I'm also not a mitmweb user and just want to make sure this doesn't degrade the UX (does it improve it?). Users might either think something is broken (having a browser-like window without the common interface) or they might expect native-like experience (e.g. context menus). However I'm not against web based desktop apps at all (I'm working on an Electron app myself) and I've looked into projects like https://github.com/zserge/lorca

Also I just realized that you cannot dock the developer tools in app mode. It's always in a new window.

Prinzhorn:
Had breakfast, am less grumpy now. I think this might actually improve UX for newcomers if it works.

Not sure what edge case I'm running into here, but the process is actually there but not really. Maybe there's something that can be done on the Python end?

```
14651 ?        Zs     0:00 [chrome] <defunct>
```

https://askubuntu.com/questions/201303/what-is-a-defunct-process-and-why-doesnt-it-get-killed

mhils:
> Had breakfast, am less grumpy now. I think this might actually improve UX for newcomers if it works.

Well we definitely do not improve usability if it looks like on your system. 😄 I'm a bit at a loss here - I'm not doing much more than calling the executable in `native_web_app`: https://github.com/mhils/native_web_app/blob/6dd65a22b644bb6c6bd1bba250e2241ae31aee05/native_web_app.py#L91-L93

It works fine on my Ubuntu VM. Let me implement some other changes and see if that changes it as a byproduct.

Also, thanks for the Lorca link - I might steal some ideas there!

mhils:
Ok - I'm putting this on hold, at least temporarily. The short summary is that bugs can probably be fixed, but getting a consistent UX here requires much more complexity than what we gain from this. We want to close our app automatically when the app windows is closed, but can't do that when users open (multiple) tabs, etc. It may be worth revisiting this once Firefox supports app mode properly, then we could hard-fail if no appropriate browser is found.

PR #4146, opened 2020-08-12
Revamp onboarding app

mhils:
#### Description

Our onboarding app accumulated quite a bit of inconsistencies, so here's an attempt at fixing that:

- improve and simplify layout
- revise install instructions
- ditch fontawesome to reduce file size

Here's what it looks like now:

image

#### Checklist


mplattner:
Very cool! 👍 

Minor remarks:
* I think @Prinzhorn had [issues with Unicode arrows on Ubuntu](https://github.com/mitmproxy/mitmproxy/pull/4127#issuecomment-670059244)  We might want to check if the icons work cross-platform, but I don't think those common icons cause problems.
* Maybe add some margin or a `<hr>` between the platform rows.

mhils:
Both Emojis are part of Unicode 6, so we should be good here. Emojis are a bit less font-dependent than arrows. I'll add some margin tomorrow and then we get this shipped. 🚀

Prinzhorn:
Definitely an improvement, I didn't like that the old version just downloaded the thing when clicking the icon. I like the explicit buttons now.

The section that interested me the most seems to be gone? Or is this no longer relevant?

Selection_646

mhils:
@Prinzhorn: Thanks for catching that. I was naively under the impression that Chrome would honor the system certificate store (as it does in Windows). I'll send a PR that fixes it.

PR #4147, opened 2020-08-12
add ASGI support for embedded apps

mhils:
## Description

This commit replaces our WSGI implementation with a new ASGI one,
which then uses `asgiref`'s compatibility mode to still support WSGI applications.
The ASGI implementation is a bit bare-bone, but good enough for our purposes.

The major changes are:

  - We now support ASGI apps.
  - Instead of taking connections out of mitmproxy's normal processing,
    we now just set flow.response and let things continue as usual.
    This allows users to see responses in mitmproxy, use the response hook
    to modify app responses, etc. Also important for us,
    this makes the new implementation work for shenanigans like sans-io.

@mplattner: this should fix the onboarding app to show proper flows. Can you test that everything works for you? 😃 

#### Checklist


mplattner:
Nice, thanks. Will check and let you know! 😄

mplattner:
Only got to test it now. Replay works 👍  Intercepting a requests causes this error:

```
error: Addon error: Traceback (most recent call last):
  File "/home/martin/mitmproxy/mitmproxy/addons/intercept.py", line 48, in request
    self.process_flow(f)
  File "/home/martin/mitmproxy/mitmproxy/addons/intercept.py", line 43, in process_flow
    f.intercept()
  File "/home/martin/mitmproxy/mitmproxy/flow.py", line 172, in intercept
    self.reply.take()
  File "/home/martin/mitmproxy/mitmproxy/controller.py", line 92, in take
    "Reply is {}, but expected it to be start.".format(self.state)
mitmproxy.exceptions.ControlException: Reply is taken, but expected it to be start.
```

Other question:
To test this with my tutorial app, I *temporarily* cherry picked your two commits.
I wonder what's the cleanest/preferred approach to get your commits into my tutorial branch and continue working on the branch?
I see a couple of options:

* Rebase such that my tutorial commits branch off HEAD of the current master branch. Con: pushing the tutorial branch "invalidates" all current commits in the PR (comments etc.)
* Cherry pick your two commits only. Con: we have the same commits twice once we merge. (Unless we squash/rebase first and remove the duplicate commits)
* Merge the current master branch in my tutorial branch.
* *Temporarily* cherry-pick for the test, immediate reset after testing. Pro: clean history, con: cumbersome

Maybe I'm overthinking this, but I like to have a clean history and don't want to cause a mess. 🤷‍♂️ What do you suggest?

mhils:
Uh oh. Interception is actually a really annoying issue and nontrivial to fix: We essentially have two parties competing over control of the `flow.reply` object in the `request` hook. Probably needs some coordination, I'll take a look.

> Maybe I'm overthinking this, but I like to have a clean history and don't want to cause a mess. 🤷‍♂️ What do you suggest?

One of these two:

> Rebase such that my tutorial commits branch off HEAD of the current master branch. Con: pushing the tutorial branch "invalidates" all current commits in the PR (comments etc.)

> Merge the current master branch in my tutorial branch.

I would not recommend cherry-picking, I don't see any advantages over a merge commit while it comes with some ugly side-effects. Other than that I don't mind, I take a rather pragmatic stance when it comes to git history. :)

Issue #4107, opened 2020-07-28
Mitmweb: Make columns configurable/customizable


rugk:
#### Is your feature request related to a problem? Please describe.
E.g. if I have a lot of clients connected, I want to see the source IP directly in one column.

#### Describe the solution you'd like
* I want to right-click on the column header, and have a list of checkboxes there, so I can select one to show it.
* Additionally, I want to be able to switch the order of the columns via drag & drop and resize each column, so I can adjust it so I like it.
* That may be saved in local storage (maybe?), so I can have a per-device profile.

#### Describe alternatives you've considered
Always show the source IP as a column?
But this is annoying if you only have one device.

_And_ by making it customizable we could satisfy many more use cases.

#### Additional context
Again see how browser devtools do it.

mhils:
Sure, we'd be happy to merge a PR for this!

mhils:
refs #4092

huss22us:
do you still need help with this ?

mhils:
@gorogoroumaru's PR looks excellent, I just need some time to get this shipped. See #4188 :)

sleeyax:
I think this issue can be closed now that https://github.com/mitmproxy/mitmproxy/pull/4533 is merged?

EDIT: never mind, it's still not there. Not sure what the other PR is about then.

PR #4160, opened 2020-08-17
mitmweb: Make columns configurable/customizable

gorogoroumaru:
#### Description

issue #4107 and issue #4092 

I made columns in mitmweb configurable and customizable using option.


mitmproxy/tools/web/webaddons.py
```diff
@@ -21,6 +22,10 @@ def load(self, loader):
             "web_host", str, "127.0.0.1",
             "Web UI host."
         )
+        loader.add_option(
+            "columns", Sequence[str], ["", "Path", "Method", "Status", "Size", "Time"],
```
mhils, line 26: Can we call this web_columns (or web_flowlist_columns?) please to make clear that this is about mitmweb only for now? Also let's make the help text a bit more descriptive. How about "Columns to show in the flow list."?

web/src/css/flowtable.less
```diff
@@ -114,16 +114,19 @@
         }
     }
     .col-method {
-        width: 60px;
+        width: auto;
```
mhils, line 117: What's the motivation for adjusting the width here? IMO this looks less desirable on certain screen sizes now.

web/src/js/__tests__/ducks/tutils.js
```diff
@@ -77,6 +77,13 @@ export function TStore(){
                 type: "str",
                 value: "b"
             },
+            columns: {
+                choices: null,
+                default: ["Path"],
+                help: "Display specified columns",
+                type: "sequence of str",
+                value: ["", "Path", "Method", "Status", "Size", "Time"]
```
mhils, line 85: The empty string for the TLS indicator is really awful here. Can we please use some other identifer? How about `column.name.slice(0, -6).toLowerCase()`? Alternative, we can add an explicit one.

web/src/js/ducks/options.js
```diff
@@ -6,7 +6,11 @@ export const RECEIVE = 'OPTIONS_RECEIVE'
 export const UPDATE = 'OPTIONS_UPDATE'
 export const REQUEST_UPDATE = 'REQUEST_UPDATE'
 
-const defaultState = {}
+const defaultState = {
+    columns: {
+        value: ["", "Path", "Method", "Status", "Size", "Time"]
+    }
```
mhils, line 12: Let's keep this empty here (and just have the FlowRow fall back to defaults if the option is not set)

mhils (commented):
Thanks! This looks very nice 😃 
I have a few smaller comments on the code, but overall I'd be very happy to get this in. 😊

gorogoroumaru:
Sorry, I lost my local repository.
I will create another PR.

PR #4184, opened 2020-08-28
new command line argument --cert-passphrase added.

mirosyn:
For decrypting the private key that is provided with the --certs command

#### Description

New command line argument added "--cert-passphrase", OpenSSL.crypto.load_privatekey already supports the passphrase, and it was only missing the arguments to pass the passphrase.

#### Checklist


test/mitmproxy/test_proxy.py
```diff
@@ -42,11 +42,12 @@ def test_simple(self):
...
-            "--cert",
-            tdata.path("mitmproxy/data/testkey.pem"))
-        with pytest.raises(Exception, match="does not exist"):
-            self.p("--cert", "nonexistent")
+        with pytest.raises(Exception, match="ambiguous option"):
+            self.assert_noerr(
+                "--cert",
+                tdata.path("mitmproxy/data/testkey.pem"))
+            with pytest.raises(Exception, match="does not exist"):
+                self.p("--cert", "nonexistent")
```
mhils, line 50: Is this change intentional? The double-indented `.raises` doesn't make much sense to me.

mitmproxy/certs.py
```diff
@@ -196,7 +196,7 @@ def load_dhparam(path):
             return dh
 
     @classmethod
-    def from_store(cls, path, basename, key_size):
+    def from_store(cls, path, basename, key_size, passphrase: str = None):
```
mhils, line 199: Let's make this argument bytes right away and call `options.cert_passphrase.encode()` when calling it. The earlier we convert this, the more likely we are to not run into encoding issues. This also removes the slightly ugly `None if passphrase is None else ` :)

```suggestion
    def from_store(cls, path, basename, key_size, passphrase: typing.Optional[bytes] = None):
```

mitmproxy/options.py
```diff
@@ -47,6 +47,10 @@ def __init__(self, **kwargs) -> None:
             certificate as the first entry.
             """
         )
+        self.add_option(
+            "cert_passphrase", Optional[str], None,
+            "Passphrase for decyrpting the private key provided in the --cert setting."
```
mhils, line 52:
```suggestion
            "Passphrase for decrypting the private key provided in the --cert option."
```

mhils (commented):
This looks like a solid PR overall, thanks! 🍰 😃

test/mitmproxy/test_proxy.py
mirosyn: Indeed double .raises doesn't make sense, removed the original, since the exception is now different, "ambiguous option" --certs and --cert-passphrase resembles each others and it now complains about that.

mitmproxy/certs.py
mirosyn: Changed it, but we still need one "ugly" if statement in config.py

mitmproxy/options.py
mirosyn: Done

mirosyn:
ping @mhils . did some changes based on the review comments.

mhils:
Awesome, thanks! 🍰

PR #4210, opened 2020-09-26
Make it possible to set sequence options

### LINKED ISSUE #3015: No way to set a sequence option on the command line

Yopi:
Attempts to fix #3015 through looking at whether or not the option is
of the type Sequence[str].

#### Description

Looks at the type of the option and allows multiple to be set if it is a Sequence. For the unknown options we don't know the type yet, so all of them are assumed to be sequences until the option has been added, and if it's not a sequence we pick the first entry instead.

#### Checklist


Prinzhorn:
I can't really judge the implementation itself, but I don't think this is generic enough. What about `Sequence[int]` for example?

Yopi:
> but I don't think this is generic enough. What about Sequence[int] for example?

Currently there is no support for any other Sequence types other than Sequence[str]. As can be seen in optmanager.py:335-367 https://github.com/mitmproxy/mitmproxy/blob/3f60e4d5dbdf2180dec8d75ca0f5a2953fd0ba4b/mitmproxy/optmanager.py#L335-L367

So I don't think this PR needs to be support it either.

Prinzhorn:
I didn't know that. Does that mean the following shouldn't work right now (it does)?

```py
import typing

from mitmproxy import ctx

class MyAddon:
    def load(self, loader):
        loader.add_option(
            name = "my_option",
            typespec = typing.Sequence[int],
            default = [1, 2, 3],
            help = "Example",
        )

        print(ctx.options.my_option)

        ctx.options.update(
            my_option = [7, 8, 9]
        )

        print(ctx.options.my_option)

addons = [
    MyAddon()
]
```

Yopi:
Sorry for the delayed response! I realised that this PR was not working exactly correctly, and I believe it wasn't testing exactly the correct thing. I've gone back and tried a slightly different approach. 

> But still this might be the perfect opportunity to clean this up, make it consistent and fix the multiple options issue. But that's just my wishful thinking 😉

Not sure exactly what you want to be cleaned up. I'm not too familiar with this project to say whether or not it makes sense to add Sequence[bool] or Sequence[int]s before there is an actual need to do so. 

But please do take another look at this PR. I think both the tests and the code makes a bit more sense this time around.

mhils:
Thanks for this really excellent PR and sorry that it took us so long! 😃 🍰

PR #4042, opened 2020-06-25
support HTTP2 trailers

sanlengjingvv:
Do not raise exception when TrailersReceived

Issue #2474, opened 2017-07-26
Support HTTP trailers


mhils:
We currently do not support HTTP/1.x or HTTP/2.0 trailers. HTTP trailers are basically additional headers sent after the body. https://discourse.mitmproxy.org/t/http-2-trailers-not-supported/557/2 mentions that gRPC is doing this, so this seems like a reasonable feature we should support. I think the rough implementation plan looks as follows:

- Add `HTTPFlow.trailers` attribute (which is a `http.Headers` instance or None)
- Modify HTTP/2.0 layer to support trailers.
- Wait if anyone ever requests HTTP/1.0 trailer support.

Kriechi:
hyper-h2 already provides this information as event:
https://python-hyper.org/h2/en/stable/api.html#h2.events.TrailersReceived

Mengjianhua-c:
Dear developer, is there any plan to support HTTP trailers? I also encountered this problem when trying to intercept grpc, very much hope that the development team solves this problem.

emaayan:
hi.. so does this mean i cannot use MITMProxy with GRPC?

nzinfo:
I've been add an experimental trailers header support in my personal branch, see. https://github.com/coreseekdev/mitmproxy   branch t/coreseek , commit cc6ae764250bff92ea4f5ae0231a7c9b3c09659f & d57eb720af767441fdedc89c3d9d9b0dfa726106 . It was mixed with some notes, so make a pull request seems a bit difficult to me .

mhils:
Very cool, thanks for linking this here @nzinfo! This looks like a good start. We'd need to extend this to also store trailers as part of a flow. If anyone wants to give this a try, I'd be happy to help with a PR.

mhils:
As @Prinzhorn pointed out in #4211, the first steps are completed here, now it's mostly a matter of exposing it in the UI. 😃

Kriechi:
This needs to be reevaluated due to the new sans-io protocol stack.

mhils:
Yes - we've just merged our new proxy core, which currently does not support trailers yet. I'm sorry about the regression, but the benefits in other areas clearly outweigh this and we need to ship at some point. If anyone wants to give this a stab again, I'd be happy to help!

PR #4211, opened 2020-09-26
Update concepts-protocols.md

Prinzhorn:
There is support for HTTP/2 trailers since 5.2

I was confused for a moment because there is https://github.com/mitmproxy/mitmproxy/blob/46a0f694852fdb73e8b4c8971b41ee7fa96afe01/examples/addons/http-trailers.py but it looks like this just hasn't been updated after #4042

Also #2474 could be updated or even closed (until "anyone ever requests HTTP/1.0 trailer support.")? Or I guess they're still missing from the mitmproxy ui? Haven't checked.


mhils:
Good catch, thanks! 🍰 

I've updated the issue - mitmdump and mitmproxy integration are still missing.

Issue #4205, opened 2020-09-19
In UI, clicking Details shows a blank page, console output attached


asheroto:
#### Problem Description
Using version 5.2 on Windows.

When on the UI page at localhost:8081, click on any entry and click "Details" on the right side and it takes me to a blank white page.

Page: http://localhost:8081/#/flows/<redacted>/details

![image](https://user-images.githubusercontent.com/49938263/93689194-c1644f80-fa91-11ea-8a86-e3b225b44627.png)

#### Steps to reproduce the behavior:
Explained above.

#### System Information
```
Mitmproxy: 5.2 binary
Python:    3.7.8
OpenSSL:   OpenSSL 1.1.1g  21 Apr 2020
Platform:  Windows-10-10.0.17763-SP0
```

mhils:
This seems to only affect flows which do not have a server address set, most prominently flows to `mitm.it`.

Samhita-sopho:
I have been reproduce the behavior and I see this for the flows that include mitm.it static resources. I am working to see how or what would fix it. Where would I discuss this issue further? @mhils

Yoshio123:
> #### Problem Description
> 
> Using version 5.2 on Windows.
> 
> 
> 
> When on the UI page at localhost:8081, click on any entry and click "Details" on the right side and it takes me to a blank white page.
> 
> 
> 
> Page: http://localhost:8081/#/flows/<redacted>/details
> 
> 
> 
> ![image](https://user-images.githubusercontent.com/49938263/93689194-c1644f80-fa91-11ea-8a86-e3b225b44627.png)
> 
> 
> 
> #### Steps to reproduce the behavior:
> 
> Explained above.
> 
> 
> 
> #### System Information
> 
> ```
> 
> Mitmproxy: 5.2 binary
> 
> Python:    3.7.8
> 
> OpenSSL:   OpenSSL 1.1.1g  21 Apr 2020
> 
> Platform:  Windows-10-10.0.17763-SP0
> 
> ```
> 
>

Yoshio123:
> #### Problem Description
> 
> Using version 5.2 on Windows.
> 
> 
> 
> When on the UI page at localhost:8081, click on any entry and click "Details" on the right side and it takes me to a blank white page.
> 
> 
> 
> Page: http://localhost:8081/#/flows/<redacted>/details
> 
> 
> 
[...]

PR #4224, opened 2020-10-08
Fix for details page in web UI coming up as blank

Samhita-sopho:
#### Description

For issue #4205, for a flow that connects to mitm.it the details throws an error. Removed that and added an extra case in Details.js

#### Checklist

### DIFF
--- a/web/src/js/__tests__/components/FlowView/DetailsSpec.js
+++ b/web/src/js/__tests__/components/FlowView/DetailsSpec.js
@@ -1,10 +1,11 @@
 import React from 'react'
+import { render } from 'react-dom'
 import renderer from 'react-test-renderer'
 import Details, { TimeStamp, ConnectionInfo, CertificateInfo, Timing } from '../../../components/FlowView/Details'
-import { TFlow } from '../../ducks/tutils'
+import { TFlow, TFlowServerAddressNull } from '../../ducks/tutils'
 
 let tflow = TFlow()
-
+let tflowServerAddressNull = TFlowServerAddressNull()
 describe('TimeStamp Component', () => {
     it('should render correctly', () => {
         let timestamp = renderer.create(<TimeStamp t={1483228800} deltaTo={1483228700} title="foo"/>),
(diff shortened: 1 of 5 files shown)


Samhita-sopho:
I am not sure where to add tests if any and looked at the failed tests , I am confused on how they are related to the web UI.

web/src/js/__tests__/components/FlowView/DetailsSpec.js
```diff
@@ -46,5 +47,8 @@ describe('Details Component', () => {
         let details = renderer.create(<Details flow={tflow}/>),
             tree = details.toJSON()
         expect(tree).toMatchSnapshot()
+        details = renderer.create(<Details flow={tflowServerAddressNull}/>),
+            tree = details.toJSON()
+        expect(tree).toMatchSnapshot()
```
mhils, line 52: Let's move this into a seperate test, i.e. `it('should render if the server connection is missing', ...)`. That makes things clearer.

web/src/js/__tests__/ducks/_tflowServerAddressNull.js
```diff
@@ -0,0 +1,114 @@
+export default function(){
```
mhils, line 1: This is a lot of code for what essentially is TFlow without a server address. I would propose we simply instantiate TFlow in our test function and manually unset the address attribute before we render the component?

web/src/js/components/FlowView/Details.jsx
```diff
@@ -136,8 +139,12 @@ export default function Details({ flow }) {
...
             <ConnectionInfo conn={flow.client_conn}/>
 
-            <h4>Server Connection</h4>
-            <ConnectionInfo conn={flow.server_conn}/>
+            {flow.server_conn.address && [
+                    <div>
+                    <h4>Server Connection</h4>
+                    <ConnectionInfo conn={flow.server_conn}/>
+                    </div>
+            ]}
```
mhils, line 147: Nit: We can use [`React.Fragment`](https://reactjs.org/docs/fragments.html) and don't need a new div. :)

mhils (commented):
Thanks for the PR!

This looks great overall, but there are a few improvements we can make. Comments below. 😃

PR #4225, opened 2020-10-09
Send http/1 request and response trailer headers

bburky:
#### Description

Add support for sending request and response trailer headers. Support for _parsing_ http/1 is not added (I'm not actually sure where that code goes). This support is sufficient for injecting trailer headers into http/1 requests or responses.

Partial completion for "Wait if anyone ever requests HTTP/1.0 trailer support." in #2474

#### Checklist

### DIFF
--- a/examples/addons/http-trailers.py
+++ b/examples/addons/http-trailers.py
@@ -16,9 +16,17 @@ def request(flow: http.HTTPFlow):
         print("HTTP Trailers detected! Request contains:", flow.request.trailers)
 
     if flow.request.path == "/inject_trailers":
-        if not flow.request.is_http2:
-            # HTTP 1.0 requires transfer-encoding: chunked to send trailers
+        if flow.request.is_http10:
+            # HTTP/1.0 doesn't support trailers
+            return
+        elif flow.request.is_http11:
+            if not flow.request.content:
+                # Avoid sending a body on GET requests or a 0 byte chunked body with trailers.
+                # Otherwise some servers return 400 Bad Request.
+                return
+            # HTTP 1.1 requires transfer-encoding: chunked to send trailers
             flow.request.headers["transfer-encoding"] = "chunked"
+        # HTTP 2+ supports trailers on all requests/responses
 
         flow.request.headers["trailer"] = "x-my-injected-trailer-header"
         flow.request.trailers = Headers([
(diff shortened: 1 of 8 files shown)


examples/addons/http-trailers.py
```diff
@@ -15,14 +15,29 @@ def request(flow: http.HTTPFlow):
     if flow.request.trailers:
         print("HTTP Trailers detected! Request contains:", flow.request.trailers)
 
+    if flow.request.path == "/inject_trailers":
+        if not flow.request.is_http2:
```
Kriechi, line 19: With HTTP/3 being on the horizon, I would prefer making this an explicit check for HTTP/1.1.
And does HTTP/1.0 support trailers at all? Should we restrict it to HTTP/1.1?
Please add a new `is_http11` function similar to the existing `is_http2` one to `mitmproxy/net/http/message.py`.

mitmproxy/net/http/http1/assemble.py
```diff
@@ -29,13 +29,18 @@ def assemble_response_head(response):
...
 
-def assemble_body(headers, body_chunks):
+def assemble_body(headers, body_chunks, trailers):
     if "chunked" in headers.get("transfer-encoding", "").lower():
         for chunk in body_chunks:
             if chunk:
                 yield b"%x\r\n%s\r\n" % (len(chunk), chunk)
-        yield b"0\r\n\r\n"
+        if trailers:
+            yield b"0\r\n%s\r\n" % (trailers,)
```
Kriechi, line 38: I think this could be simplified to `yield b"0\r\n%s\r\n" % trailers`
Or is there a reason you chose the tuple-style syntax here?

Kriechi:
Please check the test coverage:
`FAIL: mitmproxy/net/http/http1/assemble.py      45      1    98%   43`

Overall, I think this is a step in the right direction.

examples/addons/http-trailers.py
bburky: Chunked transfer encoding was introduced in 1.1, so yes this won't work in 1.0. I'll look at adding an `is_http11` function.

bburky: Added both `is_http10` and `is_http11`.  mitmproxy does not appear to support HTTP/0.9, so I didn not include it.

examples/addons/http-trailers.py
```diff
@@ -15,14 +15,41 @@ def request(flow: http.HTTPFlow):
     if flow.request.trailers:
         print("HTTP Trailers detected! Request contains:", flow.request.trailers)
 
+    print(repr(flow.request.http_version))
+    if flow.request.path == "/inject_trailers":
+        if flow.request.is_http10:
+            # HTTP/1.0 doesn't support trailers
+            return
+        elif flow.request.is_http11:
+            if not flow.request.content:
```
bburky, line 24: A lot of servers really dislike an unexpected body on GET requests so I added this check to avoid that. Also, a 0 byte chunked encoded body with trailers is legal, but odd. This ends up blocking those too. If someone wants to allow those types of requests they can tweak this script manually.

PR #4234, opened 2020-10-15
Add support to echo http trailers in dumper addon

### LINKED ISSUE #4213: Echo HTTP trailers in addons/dumper
As the title says, the dumper addon should also print HTTP trailers. This should be quite straightforward to add, so I'm leaving this for someone less experienced with mitmproxy development to pick up. :)

If you're at it, please add a test!

shiv6146:
#### Description

Changed the dumper to echo the http trailers iff trailers are being sent along with the response obj to fix #4213 

#### Checklist

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -18,6 +18,7 @@ Unreleased: mitmproxy next
     * Add interactive tutorials to the documentation (@mplattner)
     * Support `deflateRaw` `Content-Encoding`s (@kjoconnor)
     * Fix broken requests without body on HTTP/2 (@Kriechi)
+    * Add support for sending (but not parsing) HTTP Trailers to the HTTP/1.1 protocol (@bburky)
 
     * --- TODO: add new PRs above this line ---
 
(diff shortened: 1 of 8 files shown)


mitmproxy/addons/dumper.py
```diff
@@ -225,6 +226,8 @@ def echo_flow(self, f):
             self._echo_response_line(f)
             if ctx.options.flow_detail >= 2:
                 self._echo_headers(f.response.headers)
+                if f.response.trailers is not None and isinstance(f.response.trailers, http.Headers):
+                    self._echo_headers(f.response.trailers)
```
Kriechi, line 230: Do you have a sample output how this will look like?
Is there a clear indication that these are trailers? Maybe add a divider line or prefix?

Kriechi: I suppose logically the trailers should be printed after the message body (with `flow_detail >= 3` below)?

PR #4236, opened 2020-10-15
[#4235] Automatic view mode based on should_render method

### LINKED ISSUE #4235: Recognize `+json` content type suffix
Currently mitmproxy uses json view automatically only when request/response content type is exactly `application/json`. Mitmproxy uses default raw view mode, when content mode has vendor-specific media-type with json suffix, e.g. `application/vnd.my.company+json`. User has to change view mode to json each time he/she opens the flow.

Mitmproxy should recognize `+json` suffix in request/response content types.

kam800:
#### Description

This fixes #4235.

This is second attempt to support `+json` content type suffix. The first attempt in #3353 was declined. I applied authors suggestions regarding using `should_render` method for automatic view selection.

#### Checklist


mitmproxy/contentviews/image/view.py
```diff
@@ -45,3 +35,6 @@ def __call__(self, data, **metadata):
         else:
             view_name = "Unknown Image"
         return view_name, base.format_dict(multidict.MultiDict(image_metadata))
+
+    def should_render(self, content_type):
+        return content_type.startswith("image/") and content_type != "image/svg+xml"
```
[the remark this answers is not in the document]
mhils: I actually appreciate this change, less pointless maintenance for us.

mitmproxy/contentviews/msgpack.py
```diff
@@ -48,3 +48,6 @@ def __call__(self, data, **metadata):
         data = parse_msgpack(data)
         if data is not PARSE_ERROR:
             return "MsgPack", format_msgpack(data)
+
+    def should_render(self, content_type):
+        return content_type in self.__content_types
```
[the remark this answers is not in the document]
mhils: Not inlining probably reduces the number of allocations, but this is not a hot path and we shouldn't care. Either option is fine.

More practically speaking, we should consider keeping an documented `content_types` ClassVar and have a default `View.should_render` implementation that checks just that. This way we could retain backward-compatibility with third-party views.

mitmproxy/contentviews/base.py
```diff
@@ -34,6 +33,18 @@ def __call__(self, data: bytes, **metadata) -> TViewResult:
         """
         raise NotImplementedError()  # pragma: no cover
 
+    def should_render(self, content_type) -> bool:
```
mhils, line 36: While we are improving the API here, how would you all feel about having this function return an `int` as priority, where a value > 0 signals that the view can render this?

We could then selected the preferred view like this:

```python
priority, view = max(
    (v.render_priority(ct), v)
    for v in contentviews.views
)
```

A major benefit of this approach would be that I can write a custom contentview addon that pretty-prints data in a specific way for me. For example, let's say I'm only interested in a specific subproperty of a 2MB JSON blob, I could write an addon with a higher priority that only displays that part. I would have found that useful a couple of times in the past.

mhils (commented):
👍

mitmproxy/contentviews/msgpack.py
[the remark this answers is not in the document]
kam800: Well, I thought it could be misleading to have both `content_types` and `should_render`. `should_render` could do more than `content_types` does. Keeping `content_types` visible would encourage 3rd party users to still used, which could not be appropriate all the time. But the argument about keeping 3rd party views worries me and I am not sure how to handle it.

So, do you think we should have both `content_types` and `should_render` visible even when they are potentially incompatible?

test/mitmproxy/contentviews/image/test_view.py
```diff
@@ -15,3 +15,14 @@ def test_view_image(tdata):
...
+
+def test_should_render():
+    v = image.ViewImage()
+    assert v.should_render("image/png")
+    assert v.should_render("image/jpeg")
+    assert v.should_render("image/gif")
+    assert v.should_render("image/vnd.microsoft.icon")
+    assert v.should_render("image/x-icon")
+    assert v.should_render("image/webp")
+    assert not v.should_render("image/svg+xml")
```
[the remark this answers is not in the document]
kam800: Good idea, done.

mitmproxy/contentviews/base.py
kam800: Great idea. But this would require changing `should_render` interface even further – passing the request/response contents as an argument so the extension could check if the JSON subproperty exists. I am afraid that this is going to grow the PR.

On the other hand, if we implemented your proposition, we could minimise the number of incompatible changes for 3-rd party views.

Do you think we should discuss the `should_render` interface now, or should we skip it for now?

mhils:
@kam800: Thanks again for this PR! I've just pushed a proposal on top of your work that implements `render_priority`. If we break the API we should not do it twice. This allows us to do better specialization, such as @homburg's GraphQL PR at #4396. 😃 

Feedback welcome!

mitmproxy/addons/dumper.py
```diff
@@ -165,8 +170,8 @@ def _echo_request_line(self, flow: http.HTTPFlow) -> None:
 
         http_version = ""
         if (
-                flow.request.http_version not in ("HTTP/1.1", "HTTP/1.0")
-                or flow.request.http_version != getattr(flow.response, "http_version", "HTTP/1.1")
+            flow.request.http_version not in ("HTTP/1.1", "HTTP/1.0")
+            or flow.request.http_version != getattr(flow.response, "http_version", "HTTP/1.1")
```
Kriechi, line 174: Please use the new fairly new helper functions: `flow.request.is_http11` and similar.
(also below again)

mitmproxy/http.py
```diff
@@ -9,6 +9,7 @@
 
 HTTPRequest = http.Request
 HTTPResponse = http.Response
+HTTPMessage = http.Message
```
Kriechi, line 12: @mhils didn't we try to get rid of these (now empty) wrapper classes?

mhils: Yes, not sure what I was thinking here. 👍

kam800:
@mhils thank you for helping me with this PR. 👏🏻 Can’t wait to see this released!

Issue #4204, opened 2020-09-18
Allow user agent by regex. Same logic as allow_hosts, just for user agent


outsourcestudio:
I think Mitmproxy has cool features ignore_hosts and allow_hosts
Need same logic for user agent?
is it possible?

I can try to develop it myself. Need support. Copy logic from allow_hosts or ignore_hosts is possible way??
Extend ignore_hosts and allow_hosts options?

mhils:
This is not going to work easily. The point of ignore_hosts is to _not_ parse HTTP, so you don't have access to headers. Maybe you can get away filtering by client IP?

outsourcestudio:
> This is not going to work easily. The point of ignore_hosts is to _not_ parse HTTP, so you don't have access to headers. Maybe you can get away filtering by client IP?

Thanks @mhils . Seems you're right.

PR #4238, opened 2020-10-16
#4204 Updated for server details to not show for mitm.it

Samhita-sopho:
#### Description

Server section does not render if server address is null ( as seen in mitmproxy ) . Added a tflow with null server address to check if that fixes the code coverage issue as in the previous PR

#### Checklist


Kriechi:
I think you are missing some commit in this PR - I only see the updated changelog but no actual code changes...
ref #4204

CHANGELOG
```diff
@@ -2,6 +2,7 @@ Unreleased: mitmproxy next
 
 ** Full Changelog **
 
+    * UI appears blank in mitmweb for mitm.it flow(@samhita-sopho)
```
Kriechi, line 5: Please rephrase to something "that you fixed".
"fixed xyz" or "added feature xyz"

Samhita-sopho: Thank you for pointing it , after updating the same I found that the test for web ui fails so working on it

Samhita-sopho:
@Kriechi I have tested with npm test on my local machine and have seen this output 
Test Suites: 59 passed, 59 total
Tests:       316 passed, 316 total
Snapshots:   103 passed, 103 total
Time:        4.856s , 
**Summary of changes**: Made a null check to details.jsx and added a test flow with different parameters(_tflowServerAddressNull.js) , added a snapshot to test against and updated detailsSpec.js as well. Could you please let me know there is something else that needs to be done here

mhils:
This seems to be a somewhat chaotic duplicate of #4224?

Closing this here for now.

PR #4246, opened 2020-10-24
Collect all flow filter matches before modifying headers, fixes #4245

### LINKED ISSUE #4245: v4 --replacements vs v5 --modify-headers
I'm trying to replace the `User-Agent` request header if it contains a certain string.

This works with "mitmproxy-4.0.4-linux":

```
./mitmproxy --replacements ":~hq User-Agent:Mozilla(.+):CUSTOMAGENT"
```

With "mitmproxy-5.2-linux", this at least replaces the `User-Agent`, but is missing my "certain string condition":

```
./mitmproxy --modify-headers "|~hq .+|User-Agent|CUSTOMAGENT"
```

How do I add my `Mozilla` condition in v5?

None of these work:

```
./mitmproxy --modify-headers "|~hq ^(.*?)Mozilla(.*?)$|User-Agent|CUSTOMAGENT"

./mitmproxy --modify-headers "/~hq .*?Mozilla.*?/User-Agent/CUSTOMAGENT"

./mitmproxy --modify-headers "|~hq Mozilla|User-Agent|CUSTOMAGENT"

./mitmproxy --modify-headers "|~hq User-Agent: Mozilla|User-Agent|CUSTOMAGENT"

./mitmproxy --modify-headers "|~hq \"^(.*?)Mozilla(.*?)$\"|User-Agent|CUSTOMAGENT"
```

I've been trying for hours, and I feel like I've tried every variation under the sun. There's a very small chance it's a bug, but most likely I'm just doing it wrong. If it matters, this system is Ubuntu 16.04.

Prinzhorn:
#### Description

The loop was modifying the flow and later iterations (and the second loop) would call `spec.matches` against the now modified flow.

I think this is probably the most common use-case but it was just overlooked in the tests. Existing tests were only using `~q` or `~s`, which isn't affected by modifying the flow.

Not a Python expert, is the way I've used the list and `enumerate` ok?
 
#### Checklist


Prinzhorn:
I've also noticed another thing from the tests: the docs don't mention anything about duplicate headers. From the tests I can see that you can use modify headers to set the same header with multiple values.

https://github.com/mitmproxy/mitmproxy/blob/a203b69ba97eab47302edbb9b378eff353550132/test/mitmproxy/addons/test_modifyheaders.py#L58-L69

I'm not sure if this works properly for all cases, as I haven't tested it. For example what if I have a filter expression `~hq` or `~hs` but there are two headers with the name, let's say `Set-Cookie`. The addon would replace both, not just the matching one, right? Maybe the addon should draw a clear boundary and explicitly make multiple headers unsupported.

Prinzhorn:
Call me crazy, but is this thread safe? No sure how Python works in this regard, but AFAIK the `request` and and `response` events come from a different thread (every flow is handled in one). Could `configure` occur while `run` is executing? Can `run` be preempted? And then we get an index mismatch with the `matches` list? Or are threads in Python cooperative?

Kriechi:
Every flow might be a different thread, but this still leaves request + response on the same (as they are part of one flow).
Yes, I think `configure` can be called at any time - but `run` would not be preempted. The Python GIL still holds.

mhils:
All addon hooks are processed sequentially on the main thread, so that is not a concern here.

Other than that this looks good to merge except for the minor merge conflicts? 😃

Prinzhorn:
> except for the minor merge conflicts?

which conflicts?

Prinzhorn:
oh noes, the markdown was formatted :facepalm:

mhils (approved):
Ah, yes. Looks excellent. Thanks for nagging. :)

PR #4268, opened 2020-11-03
detangle custom websockets implementation and replace it with mostly …

Kriechi:
This PR moves most of the custom websocket stack into pathod because it is not used in the main mitmproxy protocol stack.


mhils (approved):
👍

PR #4294, opened 2020-11-17
Add a switch for sans-io proxy core

mhils:
This PR adds a simple switch for sans-io, which is currently hardcoded to off.

@Kriechi: I've though a bit about we could handle the coexistence of both stacks. It's a bit tricky because we need to know at import time for some stuff. I think one approach that works is to define out-of-band settings for `mitmproxy.util.compat.new_proxy_core` to be true. For example, we could check for the presence of `$MITMPROXY_NEW_CORE` on startup and then act accordingly. What do you think?

### DIFF
--- a/mitmproxy/addons/dumper.py
+++ b/mitmproxy/addons/dumper.py
@@ -181,19 +181,20 @@ class Dumper:
             replay_str = ""
             replay = ""
 
-        code = flow.response.status_code
+        assert flow.response
+        code_int = flow.response.status_code
         code_color = None
-        if 200 <= code < 300:
+        if 200 <= code_int < 300:
             code_color = "green"
-        elif 300 <= code < 400:
+        elif 300 <= code_int < 400:
             code_color = "magenta"
-        elif 400 <= code < 600:
+        elif 400 <= code_int < 600:
             code_color = "red"
         code = click.style(
-            str(code),
+            str(code_int),
             fg=code_color,
             bold=True,
-            blink=(code == 418)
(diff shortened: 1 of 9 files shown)


mitmproxy/addons/dumper.py
```diff
@@ -154,46 +155,54 @@ def _echo_request_line(self, flow):
             url = flow.request.pretty_url
         else:
             url = flow.request.url
-        terminalWidthLimit = max(shutil.get_terminal_size()[0] - 25, 50)
-        if ctx.options.flow_detail < 1 and len(url) > terminalWidthLimit:
-            url = url[:terminalWidthLimit] + "…"
+
+        if ctx.options.flow_detail <= 1:
```
Kriechi, line 159: Did this condition change from `<` to `<=`?

mhils: Yes. `<` does not make any sense because flow_detail = 0 means quiet...

mitmproxy/addons/dumper.py
```diff
@@ -204,23 +213,25 @@ def _echo_response_line(self, flow):
             size = human.pretty_size(len(flow.response.raw_content))
         size = click.style(size, bold=True)
 
+        http_version = ""
+        if (
+                flow.response.http_version not in ("HTTP/1.1", "HTTP/1.0")
```
Kriechi, line 218: how about using the fancy new `is_http1x` properties?

Kriechi (approved):

PR #4298, opened 2020-11-19
Use OpenSSL's keylog callback for SSLKEYLOGFILE (#3994)

### LINKED ISSUE #3994: SSLKEYLOGFILE is not containing TLSv1.3 secrets
#### When using mitmproxy with the SSLKEYLOGFILE environment variable TLSv1.3 keys are not exported or correctly labeled. 

I want to analyze and decrypt TLSv1.3 traffic of an application with mitmproxy and Wireshark.
I configured a gateway running mitmproxy in transparent mode and inside mitmproxy the traffic gets decrypted but Wireshark can not decrypt the captured data using the keylogfile provided by mitmproxy. 
After some research I found [this presentation](https://lekensteyn.nl/files/wireshark-tls-debugging-sharkfest19us.pdf) regarding the decryption of TLSv1.3 traffic with Wireshark. On Slide 9 there is a keylogfile example for decrypting TLSv1.3. 
In the  keylogfile provided by mitmproxy I can't find any lines starting with CLIENT_HANDSHAKE_TRAFFIC_SECRET, CLIENT_TRAFFIC_SECRET_0 nor EXPORTER_SECRET but only ones starting with CLIENT_RANDOM.

1. Export the SSLKEYLOGFILE environment variable
2. Setup mitmproxy in transparent monde
3. Open a website using TLSv1.3
4. Check the keylogfile

#### System Information
Mitmproxy: 5.1.1 binary
Python:    3.7.6
OpenSSL:   OpenSSL 1.1.1f  31 Mar 2020
Platform:  Linux-5.5.0-kali2-amd64-x86_64-with-debian-kali-rolling

mhils:
This PR fixes #3994, but requires a new pyOpenSSL release to be shipped first. Putting this here because I depend on it for sans-io debugging.

### DIFF
--- a/mitmproxy/net/tls.py
+++ b/mitmproxy/net/tls.py
@@ -86,36 +86,17 @@ class MasterSecretLogger:
     # required for functools.wraps, which pyOpenSSL uses.
     __name__ = "MasterSecretLogger"
 
-    def __call__(self, connection, where, ret):
-        done_now = (
-            where == SSL.SSL_CB_HANDSHAKE_DONE and ret == 1
-        )
-        # this is a horrendous workaround for https://github.com/mitmproxy/mitmproxy/pull/3692#issuecomment-608454530:
-        # OpenSSL 1.1.1f decided to not make connection.master_key() fail in the SSL_CB_HANDSHAKE_DONE callback.
-        # To support various OpenSSL versions and still log master secrets, we now mark connections where this has
-        # happened and then try again on the next event. This is ugly and shouldn't be done, but eventually we
-        # replace this with context.set_keylog_callback anyways.
-        done_previously_but_not_logged_yet = (
-            hasattr(connection, "_still_needs_masterkey")
-        )
-        if done_now or done_previously_but_not_logged_yet:
-            with self.lock:
-                if not self.f:
-                    d = os.path.dirname(self.filename)
-                    if not os.path.isdir(d):
-                        os.makedirs(d)
-                    self.f = open(self.filename, "ab")
-                    self.f.write(b"\r\n")
-                try:
-                    client_random = binascii.hexlify(connection.client_random())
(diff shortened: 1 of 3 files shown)


mitmproxy/net/tls.py
```diff
@@ -86,36 +86,17 @@ def __init__(self, filename):
...
-                    connection._still_needs_masterkey = True
-                else:
-                    self.f.write(b"CLIENT_RANDOM %s %s\r\n" % (client_random, masterkey))
-                    self.f.flush()
-                    if hasattr(connection, "_still_needs_masterkey"):
-                        delattr(connection, "_still_needs_masterkey")
+    def __call__(self, connection, keymaterial):
+        with self.lock:
+            if not self.f:
+                d = os.path.dirname(self.filename)
```
Kriechi, line 92: should we `expanduser` here?

test/mitmproxy/net/test_tls.py
```diff
@@ -43,7 +43,7 @@ def test_log(self, tmpdir):
 
             tls.log_master_secret.close()
             with open(logfile, "rb") as f:
-                assert f.read().count(b"CLIENT_RANDOM") >= 2
+                assert f.read().count(b"SERVER_HANDSHAKE_TRAFFIC_SECRET") >= 2
```
Kriechi, line 46: Does this mean the content changed with the new pyopenssl version?
Is Wireshark already compatible with this new format?

mhils: It definitely changed for TLS 1.3, not sure what we'd get for older versions. It doesn't really matter though, the point of the new API is just that: here are some opaque bytes you should save as-is in your keylog file.

And yes, works fine with WireShark, I used this extensively for sans-io :-)

Issue #1775, opened 2016-11-22
Sans-IO Proxy Implementation


mhils:
We had some discussions about the shortcomings of our current protocol layer implementation and I started working on a [prototype](https://github.com/mitmproxy/mitmproxy/compare/master...mhils:sans-io) of an alternative sans-io architecture to potentially replace that at some point. The purpose of this work-in-progress issue is to collect some general thoughts on what we want to achieve. Some of the problems below are immediate consequences of our current architecture, others could in theory also be solved with what we have, but are still listed here to be considered when we work on that anyway.

**See progress here:** https://github.com/mitmproxy/mitmproxy/projects/4

### Problems with the current implementation

1. Protocols are thread-based, which yields a variety of subtle bugs and race conditions in the test suite in particular.
2. No separation between IO and data processing, which makes testing and connection management hard.
3. Poor isolation between layers. (this can be improved, but it is debatable by how much)
4. Inadequate handling of timeouts. If clients or servers don't disconnect, connections may stay up infinitely.
5. The socket object is exposed to layers and dynamically wrapped there (e.g. for TLS), which changes the semantics in subtle ways (see e.g. [`ssl_read_select`](https://github.com/mitmproxy/mitmproxy/blob/92516a3b5c3e3ea8e2c7d5a92373f3ebbc4aa681/mitmproxy/net/tcp.py#L372) ). 
6. No support for TLS-over-TLS, which would be required for what the Chromium guys call a ["Secure Web Proxy"](https://www.chromium.org/developers/design-documents/secure-web-proxy).
7. No support for server-side greetings (#1774)
8. No support for changing the request destination when using HTTP2
9. We're always in upstream or non-upstream mode. (https://github.com/mitmproxy/mitmproxy/issues/1821)
10. The relationship between WebSocketFlows/HTTP Flows is not really clear (we probably just want to display one of them in the console)
11. No clear strategy determining when to establish an upstream connection. For example, for server replay it is undesirable to establish a connection right away. On the other side, for server-side greetings, connections need to be established immediately (see e.g. #2464)
12. No support for sending transparent requests to an upstream proxy (see #2813)
13. No clear strategy determining if we should be as transparent as possible (e.g. mirror cipher suites offered by the client) or if we should be as secure as possible (only offering ciphersuites we deem secure)

mhils:
Checklist for Merge (WIP):

[...]

PR #4299, opened 2020-11-19
Added addon to suppress the sending of 502 Bad Gateway response…

anneborcherding:
…s if the server is not responding

#### Description

My usecase is to use the proxy in between of a web application scanner and a web application under test. If the web application under test crashes, mitmproxy will send 502 Bad Gateway responses. These responses are irritating the web application scanner since they obfuscate the actual problem.

I propose an additional option with which a user can suppress the 502 error responses. For this, I added teh necessary option. Since the 502 responses are sent deeper in the code, I changed the HTTP code such that it will no longer send the 502 responses itself but it will now re-raise the corresponding exception. The exeption handlig will be done in the ProxyServer as before.

Is this a suitable way to reach that goal? Do you have any suggestions or comments on that?

#### Checklist

### DIFF
--- a/examples/contrib/suppress_error_responses.py
+++ b/examples/contrib/suppress_error_responses.py
@@ -1,0 +1,16 @@
+"""
+This script suppresses the 502 Bad Gateway messages, mitmproxy sends if the server is not responsing correctly.
+For example, this functionality can be helpful if mitmproxy is used in between a web scanner and a web application.
+Without this script, if the web application under test crashes, mitmproxy will send 502 Bad Gateway responses.
+These responses are irritating the web application scanner since they obfuscate the actual problem.
+"""
+from mitmproxy import http
+from mitmproxy.exceptions import HttpSyntaxException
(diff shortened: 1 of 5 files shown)


mhils:
That generally sounds like a valid use case we should support!

Regarding the specific implementation, I think the better approach would be to write an addon that calls `flow.kill()` in the `error` hook:

```python
def error(flow):
    flow.kill()
```

Would that work for you?

I'm not sure if that works properly on master to be honest, but I have just verified that it works on our new sans-io proxy core (#1775), which will likely ship soon.

anneborcherding:
Yes, I think that additional addon would work for this usecase!

Which directory would you prefer for the addon?

mhils:
Great, https://github.com/mitmproxy/mitmproxy/tree/master/examples/contrib please, thanks! 😃

mhils:
Awesome, thanks for the quick turnaround!

PR #4301, opened 2020-11-20
pyupgrade --py36-plus mitmproxy/**/*.py

Kriechi:
#### Description

Let the future begin!
`pyupgrade --py36-plus`


mitmproxy/contrib/kaitaistruct/gif.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, _io, _parent=None, _root=None):
         self.hdr = self._root.Header(self._io, self, self._root)
         self.logical_screen_descriptor = self._root.LogicalScreenDescriptorStruct(self._io, self, self._root)
         if self.logical_screen_descriptor.has_color_table:
-            self._raw_global_color_table = self._io.read_bytes((self.logical_screen_descriptor.color_table_size * 3))
+            self._raw_global_color_table = self._io.read_bytes(self.logical_screen_descriptor.color_table_size * 3)
```
mhils, line 33: We probably want to exclude `contrib`? At least kaitaistruct classes are autogenerated.

mitmproxy/tools/main.py
```diff
@@ -3,7 +3,6 @@
 DO NOT use type annotations or other python3.6-only features that makes this file unparsable by older interpreters!
 """
 
-from __future__ import print_function  # this is here for the version check to work on Python 2.
```
mhils, line 6: This might be debatable, but I think we're at the point where we can stop caring about py2 even here 👍 
(But please remove the comment above)

mhils (commented):
Looks good! The only change I'm not strictly positive about is the removal of the default `"r"` mode for `open()`, but I can certainly live with either version. Overall this is a neat automated improval. 👍

mitmproxy/contrib/kaitaistruct/gif.py
Kriechi: True - I'll revert for the whole sub-dir.

mitmproxy/tools/main.py
Kriechi: I was actually going to put it back in - just because its not a real pain for us, and py27 EOL was just this year...

PR #4307, opened 2020-11-23
Use original flow host instead of IP when exporting to curl/httpie.

dkasak:
#### Description

Unless this is done, the SNI server name will not be sent, often making
the curl/httpie command have different behaviour than the original
request (most often in the form of failing to establish a TLS
connection).

With this change, we always use the original host, fixing this failure.
However, if the original host is a domain, it may sometimes resolve to
a different IP address later on. In curl, we solve this problem by
forcing it to connect to the original IP using `--resolve`. For httpie
there is currently no easy solution (see:
https://github.com/httpie/httpie/issues/414).

#### Checklist

(Note: Ugh, ignore the name change. I sent the initial PR by accident, where I was using the showhost for controlling this, before I decided this has nothing to do with showhost.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased: mitmproxy next
 
 * Fix SNI-related reproducibility issues when exporting to curl/httpie commands. (@dkasak)
+* Add option `export_preserve_original_ip` to force exported command to connect to IP from original request. Only supports curl at the moment. (@dkasak)
 
 ### New Proxy Core (@mhils)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,6 @@
 
 ## Unreleased: mitmproxy next
 
-* Fix SNI-related reproducibility issues when exporting to curl/httpie commands. (@dkasak)
-* Add option `export_preserve_original_ip` to force exported command to connect to IP from original request. Only supports curl at the moment. (@dkasak)
-
 ### New Proxy Core (@mhils)
 
 Mitmproxy has a completely new proxy core, fixing many longstanding issues:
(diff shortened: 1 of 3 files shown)


mhils:
Thanks for the PR! I generally agree that we want to be using `pretty_url`, but I don't think that adding `--resolve` is a smart idea. What's the use case for ensuring that we have exactly the same IP address? I think it would be more intuitive and useful if we'd just connect to whatever IP the domain resolves to at this point.

dkasak:
We could certainly do that too.

My thinking with `--resolve` was to increase reproducibility. I've previously encountered cases (with Burp, in particular) where a domain resolved to multiple hosts but only one of the hosts was vulnerable to an issue. Later on when I tried copying the request as a curl command in order to inspect it in the terminal, I couldn't reliably reproduce it because it was resolving to another host.

Another case where this may occur is if a vulnerable host is later disassociated from a domain afterwards, but is not taken offline. This one is rarer, but I've seen this in the wild too.

All in all, it seemed like it would be a nice correctness edge for mitmproxy over Burp. If not made the default, perhaps it would make sense to have it behind an option?

mhils:
Yes, that's definitely something I'd be happy to have as an option!

We have an example on how you can define options at https://github.com/mitmproxy/mitmproxy/blob/master/examples/addons/options-simple.py. Please pick the option name that you think works best, ideally we have it start with `export_`. 😃

dkasak:
The latest commit seems to work for me, but please give it a glance if you get the time.

I've had some troubles with making tests work, but realized that the ingredient that I was missing was a `taddons.context().configure(addon_instance)` call. Only after this call does the test see the new option.

dkasak:
Rebased against the latest master to fix conflicts.

Anything I can do to help this get merged so it doesn't bitrot? 😇

mhils (approved):
You can't, this looks great. :) Thanks again! 🍰

Issue #4054, opened 2020-06-30
Inform the user when underscore-formatted options are used as a cli argument


mhils:
It might make sense to inform the user when underscore-formatted options are used as a cli argument, similar to the messages on deprecated commands.
At the moment it outputs: `mitmproxy: error: unrecognized arguments: --modify_body asdf`.

_Originally posted by @mplattner in https://github.com/mitmproxy/mitmproxy/pull/4046_

MartinNuc:
Alright for all who ends up here like I do now :-) For CLI args you have to use dash: `--modify-body` instead of `--modify_body`

PR #4309, opened 2020-11-24
Inform user when underscore-formatted options are used.

jrblixt:
#### Description

Informs the user when an underscore-formatted option is used. Currently, only an error is thrown. Replaces the underscore option and gives the correct format.

#### Checklist

 - [*] I have updated tests where applicable.
 - [*] I have added an entry to the CHANGELOG.


jrblixt:
Issue #4054

mhils:
This looks about right, thanks!!

Issue #4316, opened 2020-11-29
Is there any method for cashing Proxy ?


dev-chulbuji:
#### Is your feature request related to a problem? Please describe.
No. but i wanna any method for cashing proxy for any exceptions in `addons`

#### Describe the solution you'd like
I hope there is a way to stop the proxy itself when there is any problem in `addons`
```
class TestAddons:
    def __init__(self):
        test_env = os.environ.get('TEST_ENV', None)
        if test_env is None:
            raise exceptions.ProxyException("test_env is must set!")
```
This `mitmproxy.exceptions` just stops addon logic, not proxy itself.

#### Describe alternatives you've considered
i donot have any idea. but wanna ask to any proper method for me

Kriechi:
How about just calling `sys.exit`?

dev-chulbuji:
@Kriechi 
Thx, close issue.

dev-chulbuji:
@Kriechi 
```
from mitmproxy import ctx, http
import re, json, sys, signal, os

class AwsApiCallProxy:
    def __init__(self):
        # graceful shutdown
        signal.signal(signal.SIGINT, self.signal_handler)

        test_env = os.environ.get('TEST_ENV', None)
        if kabang_env is None:
            self.print_log("TEST_ENV must be set!", None, None)
            sys.exit()
```

`sys.exit()` does not seem to kill the mitmproxy process.. :(

Kriechi:
@dev-chulbuji please see https://github.com/mitmproxy/mitmproxy/blob/master/examples/addons/shutdown.py

PR #4317, opened 2020-11-29
add example to shutdown the proxy master

Kriechi:
ref #4316


mhils:
I'd say there is no shutdown-complex, so how about we just call this shutdown.py or shutdown-proxy.py?

No strong feelings though, in any case LGTM 🍰

Issue #4318, opened 2020-11-30
Replay of http2 flows seems to be broken


rozenbom:
#### Problem Description
When http2 flow is replayed "host"-header is missed which leads to a server error. The root cause is probably the following:
1. During flow construction ":authority"-header is removed: https://github.com/mitmproxy/mitmproxy/blob/880cea66bba2bd126b353d2822466300861f1289/mitmproxy/proxy/protocol/http2.py#L511
2. During replay a value for "host"-header is sourced from ":authority"-header: https://github.com/mitmproxy/mitmproxy/blob/880cea66bba2bd126b353d2822466300861f1289/mitmproxy/addons/clientplayback.py#L208.

#### Steps to reproduce the behavior:
1. Perform a http2 request
`curl --http2 https://httpbin.org/get -k -x localhost:9099`
2. Replay the flow.
3. Observe 400-error.

#### System Information
```
Mitmproxy: 5.3.0
Python:    3.8.6
OpenSSL:   OpenSSL 1.1.1h  22 Sep 2020
Platform:  macOS-10.15.7-x86_64-i386-64bit
```

mhils:
Thanks for the report!

refs #1775

mhils:
The sans-io implementation has been fixed in 81ed5c4, thanks again!

PR #4323, opened 2020-12-03
fix http2 request replay missing host header

Granitosaurus:
#### Description

Http2 replay is broeken as per: https://github.com/mitmproxy/mitmproxy/issues/4318 
this patch fixes this.

Tested with 
```
curl --http1.1 https://httpbin.org/get -k -x localhost:8080
# then replay in mitmweb
curl --http2 https://httpbin.org/get -k -x localhost:8080
# then replay in mitmweb
```

#### Checklist

### DIFF
--- a/mitmproxy/addons/clientplayback.py
+++ b/mitmproxy/addons/clientplayback.py
@@ -205,6 +205,7 @@ class ClientPlayback:
             # https://github.com/mitmproxy/mitmproxy/issues/2197
             if hf.request.http_version == "HTTP/2.0":
                 hf.request.http_version = "HTTP/1.1"
+                hf.request.headers.pop(":authority", None)
                 host = hf.request.host
                 if host is not None:
                     hf.request.headers.insert(0, "host", host)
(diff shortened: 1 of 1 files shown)


Kriechi:
Please take care of the failing test cases - which might give additional insight about the expected or correct behaviour.

Granitosaurus:
Fixed the tests though I'm really not sure why do we have explicitly remove `:authority` header? seems the tests are mocked incorrectly here as `:authority` header would never end up in this point of code in the first place as it's getting popped in http2 handler:  https://github.com/mitmproxy/mitmproxy/blob/880cea66bba2bd126b353d2822466300861f1289/mitmproxy/proxy/protocol/http2.py#L511 (as pointed out in the related issue)

mhils:
Thanks @Granitosaurus! 🍰 

I think we've been a bit inconsistent on the `:authority` header in the past and there might be a mismatch between tests and actual implementation. If your implementation fixes the issue for the current version, let's just go with this.
I'll keep #4318 open so that we have this fixed for our proxy core refactor (#1775), which swaps out the client replay implementation entirely.

PR #4324, opened 2020-12-03
ASGIApp should ignore the HTTP flows loaded from somewhere

linw1995:
#### Description

Reading some flows that processed by `Onboarding.request` from a file will cause the mitmproxy crashing.

Way to reproduce this problem.

1. run `mitmdump -w test`
2. run `http_proxy=http://localhost:8080 curl http://mitm.it`
3. exit the mitmdump process
4. run `mitmproxy -n -r test` and you will see it crashing. But it takes some time to crash.

The way to fix this is that the `ASGIApp` needs to check the flow with or without `DummyReply` before serving it.

#### Checklist


Kriechi (approved):

mhils:
Thanks for the PR! The issue description is excellent, this also looks like a reasonable fix. 👍 
I will do some more testing tomorrow and then merge.

mhils (approved):
Thanks again! This looks good for now. We may want to think about actually re-running ASGIApps in the long term, or at least making it configurable. For now, this is a good fix!

linw1995:
@mhils Thank you for taking the time and effort into this.

PR #4326, opened 2020-12-05
Update PyInstaller, require Python 3.8

### LINKED ISSUE #4303: Upgrade to Python 3.8+
similar to #2266

This issue tracks all roadblocks we need to overcome to make use of Python 3.8-exclusive features. There's not immediate need to upgrade right away, but it's useful to have an overview of what's left nonetheless.

### Windows
✔️ Windows users can just install Python 3.8
✔️ Windows Subsystem for Linux currently contains Ubuntu 20.04

### macOS
✔️ Homebrew and the mitmproxy formulae will work just fine with Python 3.8 or 3.9
✔️ macOS Catalina ships with Python 3.8.2

### Ubuntu
✔️ Groovy 20.10 has Python 3.8.6
✔️ Focal 20.04 LTS has Python 3.8.5
❌ Eoan 19.10 has Python 3.7.5
❌ Disco 19.04 has Python 3.7.3
❌ Cosmic 18.10 has Python 3.6.7
❌ Bionic 18.04 LTS has Python 3.6.9
❌ Artful 17.10 has Python 3.6.3
❌ Zesty 17.04 has Python 3.5.3
❌ Xenial 16.04 LTS has 3.5.1

### Debian
✔️ Debian bullseye has Python 3.8.6 - **unreleased as of 2020-11-21**
❌ Debian 10 (buster) has Python 3.7.3
❌ Debian 9 (stretch) has Python 3.5.3

### Arch Linux
✔️ rolling release: has Python 3.8.6 as _python_ package in _Extra_

### Fedora
✔️ Fedora 32 has Python 3.8.6 (released 2020-10)
✔️ Fedora 32 has Python 3.8.6 (released 2020-04)
❌ Fedora 31 has Python 3.7.9
...
❌ Fedora 27 has Python 3.6.3
❌ Fedora 26 has Python 3.6.3
❌ Fedora 25 has Python 3.5.4

### Kali Linux
✔️ kali-rolling has Python 3.8.6 (as of 2020-11-21) - rolling release since January 2016

### PyInstaller
✔️ PyInstaller supports Python 3.8 since PyInstaller 4.1.

mhils:
This PR updates PyInstaller to Python 3.9 and drops our support for Python 3.6 and 3.7.

fixes #4303 

#### Checklist


setup.py
```diff
@@ -65,6 +63,7 @@
             "pathoc = pathod.pathoc_cmdline:go_pathoc"
         ]
     },
+    python_requires='>=3.8',
```
mhils, line 66: @Kriechi: Is there a reason why we didn't do this before?

See https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires.

This is checked since pip 9, released 2016-11-02.
Do we want to 1) forcibly downgrade pip (which may bring its own issues) to make installs succeed, or 2) drop the test?

Kriechi: Previously we still wanted to "support" py27 and py35 to at least try and install the package, and then tell them with our own  custom message to upgrade.
Back then, pip was still fairly new, and we wanted to supported older versions as well.
Now that pip 9 is almost 4 years old, I think it is safe to bump this, and get rid of our own checks around this.

mitmproxy/tools/main.py
```diff
@@ -1,22 +1,173 @@
 """
-This file must be kept in a python2.7 and python3.5 compatible syntax!
-DO NOT use type annotations or other python3.6-only features that makes this file unparsable by older interpreters!
+This file now contains python3.8+ syntax!
```
Kriechi, line 2: Does it matter anymore? IMO can be removed...

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py35, py36, py37, py38, py39, flake8, filename_matching, mypy, individual_coverage, docs
+envlist = py35, py38, py39, flake8, filename_matching, mypy, individual_coverage, docs
```
Kriechi, line 2: `s/py35, //`

Kriechi (approved):

Issue #4336, opened 2020-12-11
Mitmproxy as a windows service fails to SSL


illera88:
#### Problem Description
Mitmproxy works fine if run directly but fails to intercept SSL requests when run as a windows service. I run the code as Administrator. 

#### Steps to reproduce the behavior:
This code works fine:
```
class AddHeader:
    def __init__(self):
        self.num = 0

    def response(self, flow):
        self.num = self.num + 1
        flow.response.headers["count"] = str(self.num)

if __name__ == '__main__':
    options = Options(listen_host='127.0.0.1', listen_port=8081)
    config = ProxyConfig(options)
    master = ProxyMaster(options, with_termlog=DEBUG, with_dumper=DEBUG)
    master.server = ProxyServer(config)
    master.addons.add(AddHeader())
    master.run()
```

But this code that uses mitmproxy within a service does not work and produces an error:

```
import socket

import win32serviceutil

import servicemanager
import win32event
import win32service

from mitmproxy.options import Options
from mitmproxy.proxy.config import ProxyConfig
from mitmproxy.proxy.server import ProxyServer
from mitmproxy.tools.dump import DumpMaster

class SMWinservice(win32serviceutil.ServiceFramework):
    '''Base class to create winservice in Python'''

    _svc_name_ = 'pythonService'
    _svc_display_name_ = 'Python Service'
    _svc_description_ = 'Python Service Description'

    @classmethod
    def parse_command_line(cls):
        '''
        ClassMethod to parse the command line
        '''
        win32serviceutil.HandleCommandLine(cls)

    def __init__(self, args):
        '''
        Constructor of the winservice
        '''
        win32serviceutil.ServiceFramework.__init__(self, args)
        self.hWaitStop = win32event.CreateEvent(None, 0, 0, None)
        socket.setdefaulttimeout(60)

    def SvcStop(self):
        '''
        Called when the service is asked to stop
        '''
        self.stop()
        self.ReportServiceStatus(win32service.SERVICE_STOP_PENDING)
        win32event.SetEvent(self.hWaitStop)

    def SvcDoRun(self):
        '''
        Called when the service is asked to start
        '''
        #self.ReportServiceStatus(win32service.SERVICE_RUNNING)
        self.start()
        servicemanager.LogMsg(servicemanager.EVENTLOG_INFORMATION_TYPE,
                              servicemanager.PYS_SERVICE_STARTED,
                              (self._svc_name_, ''))
        self.main()

    def start(self):
[...]

PR #4341, opened 2020-12-12
Workaround pyopenssl issue that prevents mitmproxy to be used as a Service

illera88:
#### Description

This a fix for https://github.com/mitmproxy/mitmproxy/issues/4336 that prevents mitmproxy to be used as a service.

#### Checklist

### DIFF
--- a/mitmproxy/net/tcp.py
+++ b/mitmproxy/net/tcp.py
@@ -392,7 +392,7 @@ class TCPClient(_Connection):
             self.sni = sni
             self.connection.set_tlsext_host_name(sni.encode("idna"))
         self.connection.set_connect_state()
-        
+
         count = 0
         while True:
             try:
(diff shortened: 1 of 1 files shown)


illera88:
The only test that fails is `test/pathod/test_pathod.py::TestDaemonSSL::test_ssl_conn_failure`. I've been debugging the test to understand why it fails but I don't get it.

I've compare the execution of both my branch and yours and they execute exactly the same lines of code. Here is the test:
```
class TestDaemonSSL(CommonTests):
    ssl = True

    def test_ssl_conn_failure(self):
        c = tcp.TCPClient(("localhost", self.d.port))
        c.rbufsize = 0
        c.wbufsize = 0
        with c.connect():
            c.wfile.write(b"\0\0\0\0")
            with pytest.raises(exceptions.TlsException):
                c.convert_to_tls()
            l = self.d.last_log()
            assert l["type"] == "error"
            assert "SSL" in l["msg"]
```
It calls `convert_to_tls()` which is one of the functions I've modified but both my branch and yours ends up calling

```
raise exceptions.TlsException("SSL handshake error: %s" % repr(v))
```
which is what it should do but for some reason in my branch `l = self.d.last_log()` returns `None`.

Could you please take a look at it?

Thanks

mhils:
Hey, thanks for looking so deeply into this. I do _really_ appreciate your work here, but as mentioned in https://github.com/mitmproxy/mitmproxy/issues/4336#issuecomment-743413990 the proxy core you are looking at is going away after the next release. In fact, the pain points you are running into here are a good illustration of why we are dumping the whole approach of wrapping socket objects.

This brings me in the uncomfortable position that I don't want to merge any non-trivial changes (like this one) into the current implementation. We haven't seen any major complaints with the current implementation at the moment (unless you wrap it as a Windows service I guess...). Even with extensive testing, I can't really predict that everything keeps working properly with your changes on all supported platforms, OpenSSL versions, ..., etc. We've been bitten a few times here and it's a pain to debug.

My proposal would be you try out the sans-io branch for your Windows service. It's in a relatively stable state except for the few features listed at https://github.com/mitmproxy/mitmproxy/projects/4. If you run into any errors like #4336 there, I'd be more than happy to help you with debugging and getting this going. We also have a development Slack for shorter turnaround time. :-)

mhils:
We've resolved this with the new proxy core, which has just been merged. Closing. 😃

PR #4343, opened 2020-12-12
New Proxy Core

mhils:
This PR brings our new sans-io proxy core. 🥳 

For details see #1775 and https://github.com/mitmproxy/mitmproxy/projects/4. 
We'll have a blog post up in the next few days outlining some of the changes.

For now this PR is intended for initial review. 😃


mitmproxy/addons/__init__.py
```diff
@@ -21,7 +23,13 @@
 from mitmproxy.addons import stickycookie
 from mitmproxy.addons import streambodies
 from mitmproxy.addons import save
+from mitmproxy.addons import tlsconfig
 from mitmproxy.addons import upstream_auth
+from mitmproxy.utils import compat
+
+if compat.new_proxy_core:  # pragma: no cover
```
Kriechi, line 30: Is the still needed with the updated migration plan?

mitmproxy/addons/proxyserver.py
```diff
@@ -0,0 +1,137 @@
...
+    def log(self, message: str, level: str = "info") -> None:
+        x = log.LogEntry(self.log_prefix + message, level)
+        x.reply = controller.DummyReply()  # type: ignore
+        asyncio_utils.create_task(
+            self.master.addons.handle_lifecycle("log", x),
+            name="ProxyConnectionHandler.log"
+        )
+
+
+class Proxyserver:
```
Kriechi, line 61: nitpick: this should probably be `ProxyServer`?

mitmproxy/addons/proxyserver.py
```diff
@@ -0,0 +1,137 @@
...
+            if self.server:
+                await self.shutdown_server()
+                self.server = None
+            if ctx.options.server:
+                self.server = await asyncio.start_server(
+                    self.handle_connection,
+                    self.options.listen_host,
+                    self.options.listen_port,
+                )
+                addrs = {f"http://{human.format_address(s.getsockname())}" for s in self.server.sockets}
```
Kriechi, line 116: Do we already support the HTTPS secure proxy format?

mitmproxy/proxy2/layers/http/_hooks.py
```diff
@@ -0,0 +1,66 @@
+from mitmproxy import http
+from mitmproxy.proxy2 import commands
+
+
```
Kriechi, line 4: This file should probably be rendered in our docs: https://docs.mitmproxy.org/master/addons-events/

test/filename_matching.py
```diff
@@ -9,7 +9,7 @@
 def check_src_files_have_test():
     missing_test_files = []
 
-    excluded = ['mitmproxy/contrib/', 'mitmproxy/io/proto/',
+    excluded = ['mitmproxy/contrib/', 'mitmproxy/io/proto/', 'mitmproxy/proxy2/layers/http',
```
Kriechi, line 12: 👀

test/mitmproxy/proxy2/layers/test_modes.py
```diff
@@ -0,0 +1,275 @@
...
+        assert (
+                playbook
+                << TcpMessageHook(flow)
+                >> reply()
+                << SendData(tctx.server, data)
+        )
+        assert data() == b"\x01\x02\x03"
+    else:
+        if connection_strategy == "lazy":
+            (
```
Kriechi, line 190: What is the difference between this block and the one below with assert? Is the assert statement not necessary?

test/mitmproxy/proxy2/test_context.py
```diff
@@ -0,0 +1,82 @@
+from mitmproxy.proxy2 import context
+from mitmproxy.test import tflow, taddons
+
+
+class TestConnection:
+    def test_basic(self):
+        c = context.Client(
+            ("127.0.0.1", 52314),
+            ("127.0.0.1", 8080),
+            1607780791
```
Kriechi, line 10: No easter egg timestamp? I'm disappointed...

Kriechi (approved):
woohoo! 👍 🎉 🥳

mitmproxy/addons/__init__.py
mhils: No, we can kick this all out once merged. :)

mitmproxy/addons/proxyserver.py
mhils: We now do! Let's do some testing how that works out. Practically speaking your client needs to trust the cert for this already, so that's a bit of a chicken-egg problem for mitm.it.

mitmproxy/proxy2/layers/http/_hooks.py
mhils: Yes. We are collecting all hooks [in commands.py](https://github.com/mitmproxy/mitmproxy/pull/4343/files#diff-95a362d2a1ab18ce23d76ccbdb6638bfce2255455f228134a51d86488185fb29R125), there's already a comment that we should autogenerate docs from that.

test/filename_matching.py
mhils: Yes, the structure of the test files doesn't match the structure of the code. Coverage is pretty good, maybe we can eventually fix this.

test/mitmproxy/proxy2/layers/test_modes.py
mhils: The assert statement starts the actual evaluation of the playbook. You can prepare a playbook for evaluation like this, but if you'd assert here you'd get an error: the final `>> reply(None)` triggers the `<< TlsStartHook(Placeholder())` below, which would not be covered here.
Unasserted playbooks automatically raise an error on deletion, so you can't forget to evaluate them. :)

test/mitmproxy/proxy2/layers/http/test_http_fuzz.py
```diff
@@ -0,0 +1,400 @@
...
+from test.mitmproxy.proxy2.tutils import Placeholder, Playbook, reply, _TracebackInPlaybook, _eq
+
+opts = options.Options()
+Proxyserver().load(opts)
+
+request_lines = sampled_from([
+    b"GET / HTTP/1.1",
+    b"GET http://example.com/ HTTP/1.1",
+    b"CONNECT example.com:443 HTTP/1.1",
+    b"HEAD /foo HTTP/0.9",
```
bburky, line 30: These HTTP/0.9 request lines are a bit nonsensical, actual HTTP/0.9 requests only support GET and look like:
```
GET /foo
```

Does mitmproxy actually support HTTP/0.9 now? I didn't [include a `request.is_http09` property](https://github.com/mitmproxy/mitmproxy/blob/b7efe9b2d4b986933f904912324b770dfb3e3da4/mitmproxy/net/http/message.py#L73-L83) because it didn't appear to be supported.

Code later on appears to assume that a version is always present, but with `HTTP/0.9` the version is missing from the request. So I don't think HTTP/0.9 is supported.

mhils: We don't support HTTP/0.9, I was just making up values for fuzzing here. To be honest I wasn't aware that HTTP/0.9 doesn't always include the version, the more you learn... 🙂

PR #4352, opened 2020-12-13
fix long description content type

Kriechi:
see https://packaging.python.org/guides/making-a-pypi-friendly-readme/


mhils:
Good catch!

PR #4353, opened 2020-12-13
Add a new "verbose" log level for per-connection messages.

rbdixon:
#### Description

I'm writing plugins more frequently now for various hacks. The per-connection client/server connect and disconnect messages fill up the event log and make it hard to monitor what my plugins are doing.

One solution, in the PR, is to add a new "verbose" log level so that these messages are hidden but can be revealed to aid troubleshooting.

Another solution would be to add a new bit of UI just for scripts to use. Maybe that would be a "script log" or something. There are probably other good ideas... I'm open to suggestions that help solve the usability problem I have now.

oh... and if this is just a "bad idea" or not right for the project that's ok, too.

#### Checklist

### DIFF
--- a/mitmproxy/log.py
+++ b/mitmproxy/log.py
@@ -65,10 +65,11 @@ LogTierOrder = [
     "error",
     "warn",
     "info",
+    "verbose",
     "alert",
     "debug",
 ]
 
 
 def log_tier(level):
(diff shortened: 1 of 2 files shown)


rbdixon:
I think I realize now that this would simply be better as a plugin. I'll write up something for the contrib scripts.

PR #4365, opened 2020-12-21
Proposal: Use black for code formatting

Kriechi:
#### Description

This PR changes our code formatting / code style / linting from primarily flake8 to [black](https://black.readthedocs.io/).
Black.

Black is highly opinionated and has no configuration options. There are some nice testimonials: https://black.readthedocs.io/en/stable/#testimonials

There are nice IDE/editor plugins to streamline running black on file saves: https://black.readthedocs.io/en/stable/editor_integration.html

This PR might seem huge - but is only the applied diff from a black run.


Kriechi:
As discussed on Slack, for now we are happy enough with flake8 and friends.

PR #4366, opened 2020-12-21
Cleanup TLS Implementation

mhils:
This PR cleans up our TLS implementation to remove non-sans-io cruft.

TODO:

 - [x] Monkeypatch pyOpenSSL so that we don't depend on a new release
 - [x] Fix end-to-end SSLKEYLOGFILE tests


mitmproxy/addons/tlsconfig.py
```diff
@@ -161,45 +158,40 @@ def create_proxy_server_ssl_conn(self, tls_start: tls.TlsStartData) -> None:
...
                 server_name: str = (server.sni or server.address[0].encode("idna")).decode()
-                path = os.path.join(client_certs, f"{server_name}.pem")
-                if os.path.exists(path):
-                    client_cert = path
-
-        args["cipher_list"] = ':'.join(server.cipher_list) if server.cipher_list else None
-        ssl_ctx = net_tls.create_client_context(
-            cert=client_cert,
-            sni=server.sni.decode("idna") if server.sni else None,  # TODO: Should pass-through here.
+                p = (client_certs / f"{server_name}.pem")
```
Kriechi, line 173: I see this code just moved down a bit, but just to raise the question: is `server_name` always a valid file name? Win/Linux/Mac filesystem restrictions and allowed characters and length?

mitmproxy/net/tls.py
```diff
@@ -1,266 +1,147 @@
...
-    @staticmethod
-    def create_logfun(filename):
-        if filename:
-            return MasterSecretLogger(filename)
-        return None
+
+def make_master_secret_logger(filename: Optional[str]) -> Optional[MasterSecretLogger]:
+    if filename:
+        return MasterSecretLogger(Path(filename))
+    return None
```
Kriechi, line 79: Just for fun: this might be a good place for a Null-Object:
https://en.wikipedia.org/wiki/Null_object_pattern

mitmproxy/net/tls.py
```diff
@@ -1,266 +1,147 @@
...
-
-    context = _create_ssl_context(
-        verify=verify,
-        verify_callback=verify_callback,
-        **sslctx_kwargs,
-    )
-
-    if sni:
+    context.set_verify(verify.value, None)
+    if sni is not None:
```
Kriechi, line 144: empty strings are fine now?

mitmproxy/net/tls.py
```diff
@@ -367,42 +252,13 @@ def is_tls_record_magic(d):
     # TLS ClientHello magic, works for SSLv3, TLSv1.0, TLSv1.1, TLSv1.2
```
Kriechi, line 252: please add `1.3` as another valid magic:
https://tls13.ulfheim.net/

Kriechi (approved):

mitmproxy/addons/tlsconfig.py
mhils: I'm sure you can find an edge case, but this should be fine. We'd need a user who needs client certificates for a _really_ weird domain.

mitmproxy/net/tls.py
mhils: You mean having a noop function? I decided against that because we can avoid the entire C -> Python -> C roundtrip, which probably isn't too cheap.

PR #4372, opened 2020-12-28
cleanup old proxy server

mhils:
1000 more lines 🙌


mitmproxy/addons/tlsconfig.py
```diff
@@ -214,6 +214,14 @@ def configure(self, updated):
             key_size=ctx.options.key_size,
             passphrase=ctx.options.cert_passphrase.encode("utf8") if ctx.options.cert_passphrase else None,
         )
+        if self.certstore.default_ca.has_expired():
+            ctx.log.warn(
+                "The mitmproxy certificate authority has expired!\n"
+                "Please delete all CA-related files in your ~/.mitmproxy folder.\n"
+                "The CA will be regenerated automatically after restarting mitmproxy.\n"
+                "Then make sure all your clients have the new CA installed.",
```
Kriechi, line 222: Let's point them to the docs to use mitm.it et al.

mitmproxy/exceptions.py
```diff
@@ -1,4 +1,10 @@
 """
+
+Edit 2020-12 @mhils:
+    The advice below hasn't paid off in any form. We now just use builtin exceptions and specialize where necessary.
+
+---
```
Kriechi, line 6: Why not? Was it too hard to maintain? Or did our community not adopt them?
If mitmproxy were a library, I would have a stronger position on this - but since we do not promote this way of using mitmproxy, I'm fine with relaxing the exception types a bit. But IMO every exception that gets thrown "outside" of mitmproxy, should have a proper mitmproxy-specific subclass. One might argue this should apply to exception that get thrown "into" a user addon as well...

mhils: (Most exceptions removed in this PR were used in the old proxy core for control flow, i.e. they'd be artificially raised to "travel" multiple layers in the stack. This kind of spaghetti code has been completely eliminated in the new core, so we don't need Socks5ProtocolException and friends. This is independent of what we are discussing here I think)

My general feeling is that we hurt ourselves more than anything with the specialization we introduce here. For example, `HttpSyntaxException` is really just a fancy way to say `ValueError`, but due to the current structure it's not even a subclass of ValueError. Not introducing the additional class and using ValueError directly would just be easier to understand. There are some justified cases which we should keep, but generally it's a lot of abstraction for no particular benefit. If we ever need to distinguish stuff, we can always add subclasses on-demand. There are no subdependencies we need to cater for.

Issue #4368, opened 2020-12-22
OpenSSL.crypto.X509Name cannot be garbage collected


linw1995:
#### Problem Description

If run mitmdump for a long time, its memory will keep increasing. I manage to narrow down the suspected memory leak is, this line of code in the `dummy_cert` function from `mitmproxy/certs.py` file.

https://github.com/mitmproxy/mitmproxy/blob/554674f37c731d92b0c497611bceefe17c9c514d/mitmproxy/certs.py#L108 

This line will keep creating many `OpenSSL.crypto.X509Name` objects. And they are cannot be garbage collected.

#### Steps to reproduce the behavior:

Run the below script to reveal this problem quickly.

```python
import secrets

import objgraph

from mitmproxy.certs import CertStore

store = CertStore.from_store(path="~/.mitmproxy/", basename="mitmproxy", key_size=2048)

print("before")
objgraph.show_growth()

for _ in range(10000):
    random_string = secrets.token_hex(16)
    commonname = random_string.encode("utf-8") + b".com"
    store.get_cert(commonname=commonname, sans=[], organization=None)

print("after")
objgraph.show_growth()
```

**Output**

```
before
function                       5791     +5791
dict                           2809     +2809
tuple                          2571     +2571
builtin_function_or_method     2052     +2052
weakref                        1594     +1594
wrapper_descriptor             1283     +1283
getset_descriptor              1117     +1117
method_descriptor               953      +953
type                            850      +850
cell                            542      +542
after
dict                    13419    +10610
X509Name                10100    +10100
list                      537      +200
_X509NameInvalidator      202      +200
X509                      101      +100
__CDataGCP                103      +100
CertStoreEntry            100      +100
Cert                      100      +100
function                 5800        +9
weakref                  1599        +5
```

#### System Information

Mitmproxy: 7.0.0.dev (+221, commit 6123675)
Python:    3.9.1
OpenSSL:   OpenSSL 1.1.1i  8 Dec 2020
Platform:  macOS-10.15.7-x86_64-i386-64bit

#### Temporary Fix

`OpenSSL.crypto.X509` uses `self._issuer_invalidator = _X509NameInvalidator()` to invalidate `X509Name` object 
(in this problem it is issuer name). It holds them forever even the dummy certs are no longer in usage.

Use `X509.from_cryptography` and `X509.to_cryptography` to create a new `X509` object to resolve this problem.

[...]

PR #4374, opened 2020-12-29
Use cryptography for certificate generation

mhils:
Looking at #4368 turned up a mess of memory leaks in pyOpenSSL. However, [pyca/pyopenssl](https://github.com/pyca/pyopenssl) prominently states that "for anything other than making a TLS connection you should move to cryptography and drop your pyOpenSSL dependency." 

As such, instead of fixing pyOpenSSL, we just migrate and get rid of our problems. :) The cryptography API is much nicer, so we should use it everywhere where we currently use the pyOpenSSL objects.

@linw1995, does this fix the memory leaks you are seeing? 😃


linw1995:
@mhils Yeah! It works well now. Thank you for the fantastic work 👍

Prinzhorn:
When I read the title I was like "What did you use before using cryptography? Rocks and sticks? What's next? Use blockchain for certificate generation?". Now I understand the package is literally called cryptography :smile: 

Nice work!

mitmproxy/addons/tlsconfig.py
```diff
@@ -126,17 +128,17 @@ def create_client_proxy_ssl_conn(self, tls_start: tls.TlsStartData) -> None:
...
+            chain_file=entry.chain_file,
             request_client_cert=False,
             alpn_select_callback=alpn_select_callback,
             extra_chain_certs=server.certificate_list,
             dhparams=self.certstore.dhparams,
         )
         tls_start.ssl_conn = SSL.Connection(ssl_ctx)
         tls_start.ssl_conn.set_app_data(AppData(
-            server_alpn=server.alpn,
+            server_alpn=always_bytes(server.alpn, "utf8", "replace"),
```
Kriechi, line 141: Technically ALPN does not say anything about UTF-8 - just a plain byte string that is registered with IANA.
`replace` in this context sounds wrong/dangerous - but very unlikely to cause a problem in the real world.

mitmproxy/proxy/layers/http/__init__.py
```diff
@@ -351,7 +351,7 @@ def check_killed(self, emit_error_hook: bool) -> layer.CommandGenerator[bool]:
                 yield HttpErrorHook(self.flow)
             # For HTTP/2 we only want to kill the specific stream, for HTTP/1 we want to kill the connection
             # *without* sending an HTTP response (that could be achieved by the user by setting flow.response).
-            if self.context.client.alpn == b"h2":
+            if self.context.client.alpn == "h2":
```
Kriechi, line 354: See above comment: technically this should be a byte string. There is no guarantee that bytes->utf8->bytes conversions will match again. Not only replace-encoding, but also different utf-8 code points etc.

mitmproxy/proxy/server.py
```diff
@@ -427,7 +427,8 @@ def tls_start(tls_start: tls.TlsStartData):
                 tls_start.ssl_conn.set_accept_state()
             else:
                 tls_start.ssl_conn.set_connect_state()
-                tls_start.ssl_conn.set_tlsext_host_name(tls_start.context.client.sni)
+                if tls_start.context.client.sni is not None:
```
Kriechi, line 430: Do we still support handshakes without SNI present?

mitmproxy/tools/console/flowdetailview.py
```diff
@@ -40,64 +39,37 @@ def flowdetails(state, flow: mitmproxy.flow.Flow):
...
         )
 
-        c = sc.cert
+        c = sc.certificate_list[0]
         if c:
             text.append(urwid.Text([("head", "Server Certificate:")]))
             parts = [
                 ("Type", "%s, %s bits" % c.keyinfo),
-                ("SHA1 digest", c.digest("sha1")),
+                ("SHA256 digest", c.fingerprint().hex()),
```
Kriechi, line 63: Nice!

mitmproxy/proxy/server.py
mhils: Yes, we do. The relevant logic is here: https://github.com/mitmproxy/mitmproxy/blob/7d7c2908b2b150e719e3f3c2f086e9290dc0603b/mitmproxy/addons/tlsconfig.py#L99-L106

Kriechi (approved):
Looks good - my only concern is that ALPN is *not* UTF-8. To be fully correct this should stay as byte strings - but I admit it is unlikely to cause problems in the real-world, unless somebody does crazy custom stuff.

mitmproxy/addons/tlsconfig.py
mhils: Fixed. 👍

Issue #4371, opened 2020-12-25
Create a new API endpoint "/download-trafffic" which allows to download traffic on a specific start and end time.


bansalp004:
#### Problem Description
Currently there are few applications which integrates MITM to provide download traffic capture for a specific start and end time. Right now there is no such API which allows such time based  filtering and traffic download.

#### Proposal
Add a new endpoint /download-trafffic?startFrom=timestamp&endTime=timestamp.
use same mitm default port - so the API endpoint below can work 
http://mitm.it?downloadjson?startTimestamp=startTime&endTimestamp=endTime

mhils:
Thanks again for the suggestion! I don't see this as something that is very much core to mitmproxy, so I'll close this here.

FYI, mitmweb already exposes `http://127.0.0.1:8081/flows.json`. 😃

PR #4375, opened 2020-12-29
#4371 Create a new API endpoint "/download-trafffic" which allows to …

bansalp004:
https://github.com/mitmproxy/mitmproxy/issues/4371

#### Description

a new endpoint /download-trafffic?startFrom=timestamp&endTime=timestamp has been added which download traffic capture for a given start and end time
ex. usage of the API
http://mitm.it?downloadjson?startTimestamp=startTime&endTimestamp=endTime

### DIFF
--- a/mitmproxy/addons/onboardingapp/__init__.py
+++ b/mitmproxy/addons/onboardingapp/__init__.py
@@ -1,6 +1,7 @@
 import os
-
-from flask import Flask, render_template
+import urllib
+import requests
+from flask import Flask, render_template,request, Response
 
 from mitmproxy.options import CONF_BASENAME, CONF_DIR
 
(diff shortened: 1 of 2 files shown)


Prinzhorn:
Correct me if I'm wrong, but you've added an API to the `mitm.it` service which requires the mitmweb service on a hard coded port (8081)? That doesn't look right to me.

I can't really comment on the implementation details or the usefulness. But what worries me is that everyone connected to the proxy can download everybody else's data. I know this is not the main purpose of mitmproxy, but some people run mitmproxy and have it accessible to multiple others.

I haven't looked closely at the code, that was just my first impression. Maybe someone else can comment if and how this could instead be implemented as a user addon (can the integrated things be extended? @mhils )

mhils:
Thanks for the PR @bansalp004!

As @Prinzhorn already remarked, this is nothing we want as part of mitm.it. If you take a look at mitmweb, you will in fact discover that it alreadys has an API to obtain all flows as JSON. I think you are best served with a custom addon here, this isn't particularly useful to the majority of users.

PR #4377, opened 2021-01-01
Add mitmproxy.Events

### LINKED ISSUE #4369: sans-io is missing addon events: clientconnect, clientdisconnect, serverconnect, serverdisconnect
#### Problem Description

First try of running sans-io with my addon. I think the connection events are missing completely.

The signature was also not working because

```py
def clientdisconnect(self, layer: mitmproxy.proxy.protocol.Layer):

def serverconnect(self, conn: mitmproxy.connections.ServerConnection):
```
> in script addon.py: No module named 'mitmproxy.connections'
> in script addon.py: module 'mitmproxy.proxy.protocol' has no attribute 'Layer'

While searching for the new names I couldn't find the hooks at all. Maybe they have been overlooked? They exist in tests `/test/mitmproxy/test_flow.py` and `/test/mitmproxy/tservers.py` but are never triggered I think. Are the tests even running?

Compare 6 vs master

`mitmdump --scripts conn.py --quiet`

```python
def clientconnect(layer):
    print("clientconnect")

def clientdisconnect(layer):
    print("clientdisconnect")

def serverconnect(conn):
    print("serverconnect")

def serverdisconnect(conn):
    print("serverdisconnect")
```

I couldn't test sans-io any further because the missing connection events make my foreign key constraints fail :shrug: 

#### System Information
```
Mitmproxy: 7.0.0.dev (+462, commit 554674f)
Python:    3.8.5
Platform:  Linux-5.4.0-58-generic-x86_64-with-glibc2.29
```

mhils:
New year, new PR! 😃 🎆 

fixes #4369


docs/scripts/events.py
```diff
@@ -0,0 +1,142 @@
...
+)
+
+category(
+    "Advanced Lifecycle",
+    [
+        layer.NextLayerHook,
+        events.UpdateEvent,
+        log.AddLogEvent,
+    ]
+)
```
Kriechi, line 133: Are these only specific events we want to document, or could we auto-generate this, e.g., something like https://stackoverflow.com/a/1796247 ?

mitmproxy/addonmanager.py
```diff
@@ -77,11 +79,11 @@ def add_option(
...
-                existing.name == name and
-                existing.typespec == typespec and
-                existing.default == default and
-                existing.help == help and
-                existing.choices == choices
+                    existing.name == name and
+                    existing.typespec == typespec and
+                    existing.default == default and
+                    existing.help == help and
+                    existing.choices == choices
```
Kriechi, line 86: Does the diff just look weird, or is this now 8 spaces indented? (same question for above)

mitmproxy/events.py
```diff
@@ -0,0 +1,85 @@
...
+        # define a custom hash and __eq__ function so that events are hashable and not comparable.
+        cls.__hash__ = object.__hash__
+        cls.__eq__ = object.__eq__
+
+
+all_events: Dict[str, Type[MitmproxyEvent]] = {}
+
+
+@dataclass
+class ConfigureEvent(MitmproxyEvent):
```
Kriechi, line 49: I not sure I particularly like the choice of naming strategy:
* `...Event` suffix is redundant - we already know these are events from the import path
* events are something that have happened - meaning past tense
* unless we have a hierarchy or other grouping, they should be descriptive

IMO this matches what h2 is doing: https://github.com/python-hyper/hyper-h2/blob/master/src/h2/events.py
and to some degree also wsproto: https://github.com/python-hyper/wsproto/blob/master/src/wsproto/events.py (although wsproto is considering this "active tense", because events in wsproto are also used to produce output)

For example, I would propose:
* `ConfigureEvent` -> `ConfigurationChanged`
* `DoneEvent` -> `EventLoopTerminated`
* `RunningEvent` -> `StartupCompleted`
* `UpdateEvent` -> `FlowChanged`

(just proposals to demonstrate my idea - feel free to suggest different ones)

mitmproxy/events.py
```diff
@@ -0,0 +1,85 @@
...
+from typing import ClassVar, Any, Dict, Type, Set, List, TYPE_CHECKING, Sequence
+
+import mitmproxy.flow
+
+if TYPE_CHECKING:
+    import mitmproxy.addonmanager
+    import mitmproxy.log
+
+
+class MitmproxyEvent:
```
Kriechi, line 12: nitpick: personally I always try to avoid writing `mitmproxy` with a capital letter. It looks odd and somehow doesn't fit. 

Also, what purpose serves adding "Mitmproxy..." here? The full import path is now dual redundant:
`mitmproxy.events.MitmproxyEvent`

https://xkcd.com/244/

mitmproxy/proxy/commands.py
```diff
@@ -87,45 +86,16 @@ def __init__(self, connection: Connection, half_close: bool = False):
         self.half_close = half_close
 
 
-class Hook(Command):
+class Hook(Command, mitmproxy.events.MitmproxyEvent):
```
Kriechi, line 89: nitpick: I think everywhere else we use sub-import, `from mitmproxy import events`, instead of the full path...

mitmproxy/proxy/context.py
```diff
@@ -118,6 +122,15 @@ def __init__(self, peername, sockname, timestamp_start):
         self.sockname = sockname
         self.timestamp_start = timestamp_start
 
+    def __str__(self):
+        if self.alpn:
+            tls_state = f", alpn={self.alpn.decode(errors='replace')}"
```
Kriechi, line 127: pedantic nitpick from previous discussion, again we probably will never encounter this:
instead of replacing or hiding characters, it would make more sense to either display ASCII (if everything is ASCII), or the hex bytes.

docs/scripts/events.py
mhils: These are all¹ events, we assert that two lines later. I do like to list/group them manually, this way we can easily get them in a useful order.

¹) all events that were imported at runtime (notice the `import mitmproxy.addons.next_layer  # noqa` at the top). The magic happens [here](https://github.com/mitmproxy/mitmproxy/blob/1158f4e972bd5606bd3d3a38d996c1a8faaf66f9/mitmproxy/events.py#L38). I think this should be good enough. We could conceivably walk the mitmproxy directory for all modules and import them all, but in practice we pretty much achieve full coverage by importing the NextLayer addon.

mitmproxy/addonmanager.py
mhils: This is just auto-formatted with PyCharm... Does your vscode/black/$IDE undo that? If yes, then let's undo it. If no, let's keep it as the pragmatic choice.

mitmproxy/events.py
mhils: Right, I'm not super happy with the naming either so far. Some more notes:

1. The event name/name of the addon function is currently automatically derived from the class name (FooBarEvent -> foo_bar method in the addon), which is why they are named the way they are. I'd suggest we keep this general approach, it makes things rather obvious. Of course we could drop the "Event" or replace it with something else.
2. `...Event` is not always redundant, for example `log.LogEvent`. FWIW I initially had them all in mitmproxy/events.py, but that led to really ugly circular imports. All proxy core hooks (which subclass from Hook, which subclasses from MitmproxyEvent) are spread in the individual protocol layers already anyways.
3. I went with event here because that's what we've called them in the docs so far. Do we have a better idea? Maybe calling them `Hook` instead? 

**Pro:** 
  - Consistency with the proxy core
  - maybe conveys that you can change stuff here (compared to reacting to an event)
  
**Con:** 
  - Change an existing established name
  - Hook is a bit less mainstream/obvious/easy than "event"

If you have other ideas, please shoot!

mitmproxy/proxy/commands.py
mhils: Yes. I made it explicit here because the _proxy core events_ have nothing to do with the _mitmproxy events_, which unfortuately is really ugly. (see my comment at https://github.com/mitmproxy/mitmproxy/pull/4377#discussion_r550805197)

mitmproxy/events.py
mhils: Picked this to have a distinct class name from `mitmproxy.proxy.events.Event`. Agreed on everything else, let's unify the discusson above.

mitmproxy/proxy/context.py
mhils: This function is only every visible when debugging, so we should be good here.

mitmproxy/events.py
Kriechi: Seems like a tricky one indeed.

About Hooks and Events: whats you current definition? Are they the same? Is a subset "user facing" and the others are more "proxy core"? One "bubble" up, the others "down"? There were also Commands...? I think we might need a syllabus soon :-P

About the event name + function name: that is indeed a nice pattern. I still don't like the class names though. Just throwing a bigger thought at it: did you consider ditching the "addons with functions" pattern and adopt a "receiving an object" style that we see with wsproto/h2/h11? We could provide a default dispatch table that would call the "legacy functions", or our users can use the "new addon API" by working with the new objects directly (and overriding the default dispatch table or function with their own)...
Another option, but less preferred even by me, would be to split these two things: Class name (with the style I proposed above), and "Addon API function name". More complexity and potential for confusion though... but a clean separation.

About the LogEvent: nitpick or technicality is that we have "log" as verb and as a noun: `LogLineAdded`, or `AddLogEvent` or `AddLogEntry`?

mhils: > About Hooks and Events: whats you current definition? Are they the same? Is a subset "user facing" and the others are more "proxy core"? One "bubble" up, the others "down"? There were also Commands...? I think we might need a syllabus soon :-P

The proxy core uses a fairly consistent model:

 - Events come "from the outside" and are generally unpredictable. For example, we have the "DataReceived" event, which signals the arrival of bytes on the network.
 - Commands are (immediate) deterministic reactions to events. The are emitted from the proxy core. For example, the "SendData" command instructs the I/O handler to send out some bytes.
 - A Hook is a Command. For example, we may receive some bytes for an HTTP request (a DataReceived event). The proxy core then decides that `request()` needs to be called, so it emits a <s>`Hook`</s>`HookStart` command. When the command has been finished (this is again happening "outside" and is unpredictable), the I/O handler will feed a <s>`HookReply`</s>`HookCompleted` _event_  into the proxy core.

TL;DR: Events are passed into the proxy core, commands come out of the proxy core. "Go call the request method on all addons" is a command (`HookStart`); "we have finished doing that" is an event (`HookCompleted`).

Events "bubble into the core", commands "bubble out". 😛 

> Just throwing a bigger thought at it: did you consider ditching the "addons with functions" pattern and adopt a "receiving an object" style that we see with wsproto/h2/h11?

That would mean that most addons are essential one function with a giant if-else statement. I don't think that is really what we want, no?

> I still don't like the class names though.

Right. I have just pushed a revised version to unify everything a bit. In a nutshell:

- `MitmproxyEvent` -> `EventHook`
- All new subclasses accordingly. For example, the `configure()` method is now represented as `ConfigureEventHook`. We can also take out "Event" entirely if you prefer.
- In the proxy core: 
  - `Hook` -> `StartHook(Command, EventHook)`
  - `HookReply` -> `HookCompleted(Event)`
- In the (user-facing) docs, we call them "Event Hooks" now. I feel that's a fairly understandable term, what do you think?

PR #4379, opened 2021-01-01
Reverse proxy: do not assume TLS for non-standard ports

mhils:
When specifying a reverse proxy, the user can supply [scheme://]host[:port]. We previously always assumed scheme="https" if no scheme was given, but that is ugly when proxying non-HTTP protocols. For example, `mitmproxy --mode reverse:smtp.gmail.com:25` would falsely assume TCP-over-TLS and then fail. This PR changes it so that we only assume HTTPS if no custom port has been provided by the user.

### DIFF
--- a/test/mitmproxy/net/test_server_spec.py
+++ b/test/mitmproxy/net/test_server_spec.py
@@ -3,16 +3,21 @@ import pytest
 from mitmproxy.net import server_spec
 
 
-def test_parse():
-    assert server_spec.parse("example.com") == ("https", ("example.com", 443))
-    assert server_spec.parse("example.com") == ("https", ("example.com", 443))
-    assert server_spec.parse("http://example.com") == ("http", ("example.com", 80))
-    assert server_spec.parse("http://127.0.0.1") == ("http", ("127.0.0.1", 80))
-    assert server_spec.parse("http://[::1]") == ("http", ("::1", 80))
-    assert server_spec.parse("http://[::1]/") == ("http", ("::1", 80))
-    assert server_spec.parse("https://[::1]/") == ("https", ("::1", 443))
-    assert server_spec.parse("http://[::1]:8080") == ("http", ("::1", 8080))
+@pytest.mark.parametrize("spec,out", [
+    ("example.com", ("https", ("example.com", 443))),
+    ("http://example.com", ("http", ("example.com", 80))),
+    ("smtp.example.com:25", ("http", ("smtp.example.com", 25))),
+    ("http://127.0.0.1", ("http", ("127.0.0.1", 80))),
+    ("http://[::1]", ("http", ("::1", 80))),
+    ("http://[::1]/", ("http", ("::1", 80))),
+    ("https://[::1]/", ("https", ("::1", 443))),
+    ("http://[::1]:8080", ("http", ("::1", 8080))),
+])
+def test_parse(spec, out):
+    assert server_spec.parse(spec) == out
 
+
+def test_parse_err():
     with pytest.raises(ValueError, match="Invalid server specification"):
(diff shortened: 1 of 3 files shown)


test/mitmproxy/net/test_server_spec.py
```diff
@@ -3,16 +3,21 @@
...
-    assert server_spec.parse("http://127.0.0.1") == ("http", ("127.0.0.1", 80))
-    assert server_spec.parse("http://[::1]") == ("http", ("::1", 80))
-    assert server_spec.parse("http://[::1]/") == ("http", ("::1", 80))
-    assert server_spec.parse("https://[::1]/") == ("https", ("::1", 443))
-    assert server_spec.parse("http://[::1]:8080") == ("http", ("::1", 8080))
-
+@pytest.mark.parametrize("spec,out", [
+    ("example.com", ("https", ("example.com", 443))),
+    ("http://example.com", ("http", ("example.com", 80))),
+    ("smtp.example.com:25", ("http", ("smtp.example.com", 25))),
```
Kriechi, line 9: Somehow this still feels wrong to me...
Assuming `http` here, while we clearly do not plan on using HTTP, but SMTP (or any other unsupported protocol.

mhils: I agree. Fixing that takes some more work, we'd need to make `.scheme` an `Optional[str]` and generally be smarter about it. Let's keep that out of here, it's not much of a concern until we support additional upstreams (SOCKS for example).

Kriechi (approved):

test/mitmproxy/net/test_server_spec.py
Prinzhorn: Can we get rid of the scheme completely and KISS? Or is it used for more than guessing the port / assuming the default? Or else we'll end up adding a list of all the schemes and default ports just for the user to type more characters then they would by specifying the port.

mhils: @Prinzhorn: That won't really work I think. We want users to be able to specify `--mode reverse:https://example.com`, which makes it clear that it's 1) TLS and 2) port 443. Neither part is really optional - `--mode reverse:example.com:443` would mean we'd need to guess the protocol based on the port, which is wrong. Thinking ahead, we also want to support other schemes, for example `--mode upstream:socks5://upstream.proxy`.

Prinzhorn: @mhils maybe there's different way instead of putting everything in a single option? Say I want to do smtp (or some custom protocol) over TLS, right now I'd have to do `https://example.com:465` to specify TLS. And I don't think we want to have a list of all variants like `smtps://`. Or do we? Maybe support everything that curl does for #ux?

Prinzhorn: also I didn't follow the source fully, is it possible to upstream non-HTTP over port 80/443? Or would mitmproxy assume http because it's default port? What if I `ncat --ssl --listen 443` and it's not HTTP?

Kriechi: this might fit the discussion: https://docs.syncthing.net/users/config.html#listen-addresses
e.g., introducing high-level pseudo-schemes:
* `tcp://`
* `tls://`
* `quic://`

Prinzhorn: I like that, because mixing schemes like `http://` and `socks5://` seems weird to me. But I don't know enough about the internals to know if that's a feasible solution. What about a potential for `udp://` tho? That would make `tcp://` vs `tls://`. I think of `tls` more like a flag than a protocol. It's hard to express that with a scheme, unless we just append `s` to everything :smile:

PR #4381, opened 2021-01-01
Remove unused options, warn about nonfunctioning ones.

mhils:


mitmproxy/addons/streambodies.py
```diff
@@ -19,13 +19,6 @@ def load(self, loader):
             Understands k/m/g suffixes, i.e. 3m for 3 megabytes.
             """
         )
-        loader.add_option(
-            "stream_websockets", bool, False,
-            """
-            Stream WebSocket messages between client and server.
-            Messages are captured and cannot be modified.
-            """
-        )
```
Kriechi, line 28: Do we currently support this?
if not, we should use the same "currently non functioning" message as above.

mitmproxy/options.py
```diff
@@ -105,13 +101,6 @@ def __init__(self, **kwargs) -> None:
             "Enable/disable HTTP/2 support. "
             "HTTP/2 support is enabled by default.",
         )
-        self.add_option(
-            "http2_priority", bool, False,
-            """
-            PRIORITY forwarding for HTTP/2 connections. Disabled by default to ensure compatibility
-            with misbehaving servers.
-            """
-        )
```
Kriechi, line 114: Is the current behaviour documented? I don't mind removing it - just want to make sure we publicly state somewhere what happens with PRIORITY frames and how mitmproxy behaves when (not-)forwarding them.

mhils: Fixed 👍

mitmproxy/addons/streambodies.py
mhils: I want "currently non functioning" for features we generally plan to implement again. Redirecting people to the issue allows us to gauge the importance of particular features.

IMO stream_websockets is very much an edge case we should drop altogether. If you feel this is something that's actually actively used we can make it a log message as well and see what comes back. What do you think?

Kriechi: Why do you feel this is an edge case?
In the absence of a `body_size_limit`-like option for WebSockets, streaming large WS messages is important, otherwise we immediately run out of memory.

mhils: I think it's an edge case because WS messages aren't typically that large. A few MBs sure, but you never really have a 1GB transmission over WebSockets? If someone comes up with an example in practice I'm happy to reconsider.

PR #4393, opened 2021-01-11
Fix dependency typo?

cbracketdash:
I'm not exactly sure this is correct but I just wanted to shed light on it.


Kriechi:
Thanks for the PR - it's not a typo though. Both "spellings" are valid, it's just shorter without the extra space. So, just a style question.

cbracketdash:
Ah I see, thanks for the clarification.

PR #4395, opened 2021-01-15
Upon --replace, suggest --modify-body / --modify-headers

nomeata:
Else the user sees
```
--replace is deprecated.
Please use `--replacements` instead.
mitmdump: error: unrecognized arguments: --replace
```
and then
```
mitmdump: error: unrecognized arguments: --replacements
--replacements is deprecated.
Please use `--modify-body` or `--modify-headers` instead.
```
and feels silly


mhils:
Excellent find, thanks! 🍰

Issue #4030, opened 2020-06-10
GraphQL query view


homburg:
#### Is your feature request related to a problem? Please describe.
GraphQL request query bodies are technically just json, but the query field is very hard to read with escaped newlines.

#### Describe the solution you'd like
I'd like to inspect a GraphQL request directly in mitmweb + mitmproxy with a readable query presentation

#### Additional context
I'm currently using my own [custom content-view](https://github.com/homburg/mitmproxy/blob/feature/graphql-content-view/mitmproxy/contentviews/graphql.py). I print the query field with expanded newlines below the rest of the body as json.

Raw body
```
{
    "operationName": null,
    "query": "{\n  stationWithEvaId(evaId: 8000105) {\n    name\n    location {\n      latitude\n      longitude\n    }\n    picture {\n      url\n    }\n  }\n}\n",
    "variables": null
}
```

GraphQL content-view:
```
{
  "variables": null,
  "operationName": null
}
---
{
  stationWithEvaId(evaId: 8000105) {
    name
    location {
      latitude
      longitude
    }
    picture {
      url
    }
  }
}
```

I'd like the graphql content-view to be auto-selected, but there's no graphql-specific headers to match on.

mhils:
This sounds reasonable. I would be happy to merge a PR that employs some heuristics to detect GraphQL (https://graphql.org/learn/serving-over-http/#post-request), and then changes the formatting accordingly! :smiley:

homburg:
Fixed by #4396

PR #4396, opened 2021-01-15
GraphQL content-view

homburg:
Add GraphQL content-view and auto-selection-logic. See #4030

### Single query

```
{
  "variables": {
    "id": 8000105
  },
  "operationName": "S",
  "query": "..."
}
---
query S($id: Int!) {
  stationWithEvaId(evaId: $id) {
    name
    location {
      latitude
      longitude
    }
    picture {
      url
    }
  }
}
```

### Batched/list of queries

```
--- 0/1
{
  "variables": null,
  "operationName": null,
  "query": "..."
}
---
{
  stationWithEvaId(evaId: 8000105) {
    name
    location {
      latitude
      longitude
    }
    picture {
      url
    }
  }
}
--- 1/1
{
  "variables": null,
  "operationName": null,
  "query": "..."
}
---
{
  stationWithEvaId(evaId: 8000105) {
    name
    location {
      latitude
      longitude
    }
    picture {
      url
    }
  }
}
```

#### Checklist


mhils (commented):
Thanks for the PR! 🍰 

This looks good overall. Am I right to assume that this change pretty much bypasses our JSON view entirely, because all JSON is handled by the GraphQL view which then potentially falls back to `format_json`?

mhils:
Quick update: We'll get #4236 into master in the next day or so. Contentviews then have a method to return render priority, which should fix the issue mentioned in my previous comment. :)

mhils:
Another quick update: #4236 has just landed. Would you mind implementing a `render_priority` method on your contentview? I'd propose something along these lines:

1. Stick an `@lru_cache(1)` decorator on `mitmproxy.contentviews.json.parse_json` and use that in your graphql module
2. In your `render_priority` check for the appropriate JSON content type and only then if the JSON blob is graphql. The important part is to not feed things into `parse_json` all the time, which can be pretty slow.
3. Return a priority >1 if it is GraphQL.

Hope that makes sense. If not, happy to answer any questions!

homburg:
@mhils Updated with `render_priority`. Much better than "wrapping" the json view 👍

mitmproxy/contentviews/graphql.py
```diff
@@ -0,0 +1,60 @@
+import json
+
+import typing
+
+from mitmproxy.contentviews import base
+from mitmproxy.contentviews.json import parse_json
+
+
+PARSE_ERROR = object()
```
mhils, line 9: This should be imported and not redefined here, or am I missing something?

mhils (commented):
Excellent! Thanks for the high-quality PR. Let's get PARSE_ERROR imported and a one-line test for invalid JSON added and this is good to go! :cake:

mitmproxy/contentviews/graphql.py
homburg: Wasn't sure :-) imported

mhils: The idea here is to have a dedicated object representing "null"/error ([a sentinel object](https://python-patterns.guide/python/sentinel-object/#sentinel-objects)). We can't use `None` as `"null"` is valid JSON but also evaluates to `None`. And we don't want to raise because then it's not cached :) 
Long story short, to identify this it needs to be the same unique object. 😄

mhils:
Thanks again for the excellent PR! 😃 🍰

homburg:
@mhils you're welcome! Thank you for the guidance, very pleasant pr experience :-)

PR #4399, opened 2021-01-18
Make development setup instructions more clear

mhils:
Our dev.sh/dev.ps1 scripts feel like an unnecessary layer of abstraction.
The revised docs make the install process transparent to experienced users,
and can also be easily reused for other projects (hi, pdoc!).


CONTRIBUTING.md
```diff
@@ -14,32 +14,49 @@ forward, please consider contributing in the following areas:
...
 ```shell
+# 1) Verify that these commands work:
 python3 --version
 python3 -m pip --help
 python3 -m venv --help
+# 2) Install:
+git clone https://github.com/mitmproxy/mitmproxy.git
+cd mitmproxy
+python3 -m venv venv
+venv/bin/pip install -e .[dev]
```
Kriechi, line 30: Why did you choose to deep-ref the pip binary, instead of activating the venv?

Kriechi:
Is https://github.com/mitmproxy/mitmproxy/blob/master/requirements.txt still needed after this? 
Might be cleaner to just delete it as well?

CONTRIBUTING.md
mhils: I feel that it's simpler this way because we don't need to handle virtualenv activation here and then again at the end of the section.

mhils:
Agreed, let's ditch it as well.

PR #4406, opened 2021-01-21
Clean docs folder before building, fixes #4405

### LINKED ISSUE #4405: Remove old documentation pages
#### Problem Description

There are three pages in the documentation that seem to be old as the menu bar on the left they contain differs from the rest of the page. These are https://docs.mitmproxy.org/stable/tools-mitmproxy/, https://docs.mitmproxy.org/stable/tools-mitmdump/ and https://docs.mitmproxy.org/stable/tools-mitmweb/. These three pages are the only ones that contain the "tools" section and the "feature" link in the "overview" secteion. I assume that they are just old.

mhils:
This fixes #4405. @Kriechi, please double check that we don't rm -rf too much. 😄

### DIFF
--- a/docs/build.sh
+++ b/docs/build.sh
@@ -15,5 +15,6 @@ for script in scripts/*.py ; do
   "${script}" > "${output}"
 done
 
+rm -rf ./public
 cd src
 hugo
--- a/docs/upload-stable.sh
+++ b/docs/upload-stable.sh
@@ -7,7 +7,7 @@ set -o nounset
 
 aws configure set preview.cloudfront true
 aws --profile mitmproxy \
-    s3 sync --acl public-read ./public s3://docs.mitmproxy.org/stable
+    s3 sync --delete --acl public-read ./public s3://docs.mitmproxy.org/stable
 aws --profile mitmproxy \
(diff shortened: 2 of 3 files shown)


mhils:
(I have downloaded a backup for the S3 bucket just in case, we should be good here)

docs/build.sh
```diff
@@ -15,5 +15,6 @@ for script in scripts/*.py ; do
   "${script}" > "${output}"
 done
 
+rm -rf ./public
```
Kriechi, line 18: nitpick: no real need for the dot-slash

docs/upload-stable.sh
```diff
@@ -7,7 +7,7 @@ set -o nounset
 
 aws configure set preview.cloudfront true
 aws --profile mitmproxy \
-    s3 sync --acl public-read ./public s3://docs.mitmproxy.org/stable
+    s3 sync --delete --acl public-read ./public s3://docs.mitmproxy.org/stable
```
Kriechi, line 10: Mental note for future us: DO NOT place "unmanaged" files into the buckets. Everything has to go through the process + hugo.

Kriechi (approved):

PR #4412, opened 2021-01-24
fix --certs argument

zivillian:
the help output claims that --certs is correct


mhils:
Awesome, thanks! 🍰

docs/src/content/concepts-certificates.md
```diff
@@ -93,7 +93,7 @@ The files created by mitmproxy in the .mitmproxy directory are as follows:
 
 ## Using a custom server certificate
 
-You can use your own (leaf) certificate by passing the `--cert
```
CrimegabberGER, line 96: @zivillian

PR #4413, opened 2021-01-25
fix conversion of old flows failing due to missing array keys

r00t-:
#### Description

i have some old dumps (created on 2019-09-17) that failed to load,
and this is what was required to fix it.

it does seem weird to me that those checks check for the value being None,
without checking for the key existing.

i can provide a sample file if desired, but i'm thinking this is either too obscure to bother or simple enough to just merge.

#### Checklist


mhils:
This is super helpful, thanks! 🍰 

Do you really have flows without a `request`, or do you "only" have flows lacking a response?

r00t-:
i have gigabytes of dumps, and only the one from that day has the issue, but does have it all over it.    
i can't reconstruct which version i was using to create them.    
the requests in question are websocket traffic (from facebook messenger, back before it used certiificate pinning?) so they have neither request nor response.

i'd love to provide a sample file, is there maybe an existing script to strip private data from dumps?

r00t-:
here's an affected flow, pretty-printed from FlowReader::stream() right after loading it:
```
{'client_conn': {'address': [<removed...>, 0, 0],
                 'alpn_proto_negotiated': b'http/1.1',
                 'cipher_name': 'TLS_AES_256_GCM_SHA384',
                 'clientcert': None,
                 'id': <removed>,
                 'mitmcert': b'-----BEGIN CERTIFICATE-----\n'<removed>,
                 'sni': 'edge-chat.facebook.com',
                 'timestamp_end': None,
                 'timestamp_start': 1568671224.3427272,
                 'timestamp_tls_setup': 1568671225.0254073,
                 'tls_established': True,
                 'tls_extensions': [[<removed>]],
                 'tls_version': 'TLSv1.3'},
 'client_extensions': 'permessage-deflate',
 'client_key': '<removed>',
 'client_protocol': None,
 'close_code': 1000,
 'close_message': '(message missing)',
 'close_reason': 'unknown status code',
 'close_sender': 'client',
 'error': {'msg': 'WebSocket connection closed unexpectedly by server: '
                  'TcpDisconnect("(32, \'EPIPE\')")',
           'timestamp': 1568671225.856294},
 'id': '<removed>',
 'intercepted': False,
 'marked': False,
 'messages': [[<removed>]],
 'metadata': {'websocket_handshake': <removed>},
 'server_accept': <removed>,
 'server_conn': {'address': ['edge-chat.facebook.com', 443],
                 'alpn_proto_negotiated': b'http/1.1',
                 'cert': b'-----BEGIN CERTIFICATE-----'<removed>,
                 'id': <removed>,
                 'ip_address': ['2a03:2880:f036:13:face:b00c:0:2', 443, 0, 0],
                 'sni': 'edge-chat.facebook.com',
                 'source_address': ['2a02:8109:0:36:1982:507:8630:e8c3',
                                    38061,
                                    0,
                                    0],
                 'timestamp_end': 1568671225.8632703,
                 'timestamp_start': 1568671224.7704172,
                 'timestamp_tcp_setup': 1568671224.8157454,
                 'timestamp_tls_setup': 1568671224.9579084,
                 'tls_established': True,
                 'tls_version': 'TLSv1.3',
                 'via': None},
 'server_extensions': 'permessage-deflate',
 'server_protocol': None,
 'type': 'websocket',
 'version': 7}
 ```

**Comment by mhils** [comment]
Oh thanks, this is really good context - I didn't consider WebSocket flows here I think. This PR is good as-is, but we should add a WebSocket flow to the test suite. 😄  Thanks again!

**Comment by mhils** [comment]
I have deleted your previous post containing a WebSocket flow because you removed the target address at some places but not at all. I don't think there was anything sensitive in there, but given your obvious intent I felt I should respect it. :)

**Comment by r00t-** [comment]
great that this is of use.

(the server address was ok to leave in.
i already wrote it was facebook messenger, and if they have logs old and detailed enough and bother to trace that to my account, i wish them fun with that.)

File: mitmproxy/io/compat.py [hunk 1]
```diff
@@ -174,19 +174,22 @@ def convert_6_7(data):
...
     return data
 
 
 def convert_8_9(data):
     data["version"] = 9
-    data["request"].pop("first_line_format")
-    data["request"]["authority"] = b""
-    is_request_replay = data["request"].pop("is_replay", False)
+    is_request_replay = False
+    if "request" in data:
```
r00t-, line 187: should this also have the check for "not None" for consistency?
(or remove it everywhere? i can't tell if there could ever be an existing key with "None" value).

mhils (approved):

mitmproxy/io/compat.py
mhils: Let's keep it as-is. Eventually we need to tighten down things in a future flow version. :)

mhils:
We now have a proper test case for WebSocket flows as well, thanks again! 🍰

r00t-:
great, glad that i could help!

PR #4422, opened 2021-02-02
Docs: Add API Reference Using Pdoc

mhils:
This draft PR adds API Reference Documentation for mitmproxy using [pdoc](https://pdoc.dev).

**Demo:** https://mitmproxy-pdoc.netlify.app/api/events.html

There are still a few things to do. Most importantly, we need to figure out what to do with class aliases. I think I introduced this rather unfortunate pattern in the codebase where we often expose imported classes as an alias. For example, [`mitmproxy.http.HTTPRequest`](https://github.com/mitmproxy/mitmproxy/blob/a793a6256a23924d08dba82ae3a53779829afb6e/mitmproxy/http.py#L10) is just an alias for  `mitmproxy.net.http.Request`, which again is just an alias for `mitmproxy.net.http.request.Request`. To cut a long story short, pdoc really does not like that. On https://mitmproxy-pdoc.netlify.app/api/mitmproxy/http.html for example, the `HTTPFlow.request` attribute is annotated with `mitmproxy.net.http.request.Request` instead of linking to `HTTPRequest` in the same module. We cannot fix this easily on the pdoc side. 

Two approaches:

 - [x] <s>We just live with the broken links.</s>
 - [x] We move the definitions of `HTTPFlow`, `http.Request`, `http.Response`, `http.Message`, and `http.Headers` into `mitmproxy/http.py`. This would make the file quite a bit longer, but all the user-facing HTTP stuff is in a single place. It may actually be easier for newcomers to grasp. 

@mitmproxy/devs, opinions on that?

Then there are a few other things that still need to be handled:

 - [x] Ship a pdoc release with the latest fixes needed for this.
 - [x] Determine which parts we actually want to document.
   - some potential candidates: `mitmproxy.contentviews`, `mitmproxy.net.http.headers`
   - @mitmproxy/devs, anything you want here in particular?
 - [x] Go over existing docstrings and expand where necessary.


Kriechi:
Approach 2 works for me. 👍 

I feel like we should include a few sentences about what the API docs are covering, and what level of support / long-term stability of this API we are aiming for. This should avoid confusion and complaints when we introduce changes.
Maybe https://mitmproxy-pdoc.netlify.app/api/ would be a good entry page for such a paragraph. Essentially I want to establish a solid line (as good as possible at least) between mitmproxy public-facing stable API, and the private internals.

Prinzhorn:
Nice work :smile: . For me there are two points missing:

1. How will this be integrated with the addon documentation? E.g. will the arguments here be linked https://mitmproxy-pdoc.netlify.app/addons-events/ ? Because from a user perspective the addons are the entry point into the API and I (as a consumer of the API) need to understand what the arguments of the event hooks do. Which leads to 2.
2. @Kriechi already mentioned that there should be some sort of paragraph. To me the most important sentence in this paragraph would be that mitmproxy is not meant to be (can't be?) used as a library and that the scripting API +  event hooks are the way to interact with the API. Otherwise API docs like this might cause confusion that you can import these classes and use them in a Python project. Which as far as I understand is currently not supported and not on the roadmap.

mhils:
Those are great suggestions, thank you two!

> How will this be integrated with the addon documentation?

This is an excellent idea. I've hacked the pdoc template a bit to get this working, and I feel it's super nice overall:
https://mitmproxy-pdoc.netlify.app/api/events.html

I am not happy yet with the mismatch between the example addon at the top (class-based) and the listing of events, which show plain functions as you would have them in a script, let's see. I'll ponder a bit more.

mhils:
Some more progress! The hook listings are consistent now (all class-based), I went over most of the exposed documentation and revised it. 😄  I still need to write an introductory paragraph and make mypy happy, other than that this is getting into shape.

Prinzhorn:
This looks amazing!

I think the first paragraph and "Anatomy of an addon" sample [here](https://mitmproxy-pdoc.netlify.app/addons-overview/ ) are now somewhat duplicate? I guess we could keep this page somewhat high level (with the info that mitmproxy core uses addons itself). And then link to the API section for the fun part.

mhils:
This is ready for final review! I won't have much mitmproxy time in the next few weeks, so please only focus on major blockers. We can improve/expand individual sections anytime later, contributions welcome of course. 😋

Prinzhorn:
The demo link looks great and I can't wait to get back into addon development now that I don't have to browse the source for every bit of information :smile: 
I don't have the mental capacity right now to look into the actual changes/diff.

abitrolly:
I still miss the order in which addon methods are supposed to be called. While I was debugging test, I had to experiment, for example, with if `load()` is called before `configure()` or after.

docs/scripts/api-render.py
```diff
@@ -0,0 +1,66 @@
...
+title: "{module}"
+url: "{filename}"
+
+menu:
+    addons:
+        parent: 'Event Hooks & API'
+---
+
+{{{{< readfile file="/generated/{filename}" >}}}}
+""")
```
Kriechi, line 64: nitpick: wrapping this with `textwrap.dedent` and indenting it properly might make this easier to read or look at.

docs/scripts/pdoc-template/module.html.jinja2
```diff
@@ -0,0 +1,62 @@
...
+            {% endif %}
+        {% elif doc.modulename == "mitmproxy.proxy.server_hooks" %}
+            {% if doc.qualname.startswith("ServerConnectionHookData") and doc.name != "__init__" %}
+                {{ default_is_public(doc) }}
+            {% endif %}
+        {% else %}
+            {{ default_is_public(doc) }}
+        {% endif %}
+    {% endmacro %}
+{% endif %}
```
Kriechi, line 62: meta-comment: why bother with a templating language if this file consists to 99% out of control structures? 😄

Kriechi (approved):

docs/scripts/pdoc-template/module.html.jinja2
mhils: just shut up 😁

mhils:
@abitrolly: that's a good suggestion, thanks. I absolutely agree, but let's move this to a new issue. My time is super limited at the moment and I want to get this shipped.

PR #4424, opened 2021-02-03
Fix location for contrib har_dump.py in its test

abitrolly:
To run the test, I had to copy `conftest.py` to examples.
```bash
cp test/conftest.py examples
pytest examples/contrib/test_har_dump.py
```
This fixes the `fixture 'tdata' not found` error.

#### Checklist


mhils:
Thanks :)

Issue #4390, opened 2021-01-08
mitmweb opens chrome even if other web browser is set as default


tmladek:
#### Problem Description
`mitmproxy` (`mitmweb`), by default, opens `chromium` on my system.

I don't want to open chrome or chromium, I use a different browser, which I also set as the default for my system (so e.g. `xdg-open` correctly resolves it).

If I'm reading correctly, the `browser` addon doesn't even check for any other browser than Chrome or chromium, and just fails if it cannot find them: https://github.com/mitmproxy/mitmproxy/blob/af3dfac85541ce06c0e3302a4ba495fe3c77b18a/mitmproxy/addons/browser.py

I don't really see a reason for this behavior, the `mitmweb` UI works just fine in, for instance, Firefox, and other Blink-based browsers (such as Brave or Opera) would not be found either using this method.

I would suggest using the [webbrowser](https://docs.python.org/3/library/webbrowser.html) module (present in the Python standard library), which not only does the right thing if chromium *is* present but a different browser is set as default, it also simplifies the whole script and removes the need to update it, in case the install path for Chrome/ium were to change in the future.

#### Steps to reproduce the behavior:
1. Run `mitmweb`
2. Chromium opens

#### System Information
Mitmproxy: 6.0.2
Python:    3.9.1
OpenSSL:   OpenSSL 1.1.1i  8 Dec 2020
Platform:  Linux-5.10.5-arch1-1-x86_64-with-glibc2.32

mhils:
The code you are looking at is only used if you manually run the `browser.start` command. It must be Chrome/Chromium there as we instruct it with specific command line flags.

You are probably talking about the browser automatically opened on startup? That is handled mitmproxy.tools.web.webaddons: https://github.com/mitmproxy/mitmproxy/blob/af3dfac85541ce06c0e3302a4ba495fe3c77b18a/mitmproxy/tools/web/webaddons.py#L35-L61.

IIRC we don't have `xdg-open` in the list there because that used to open lynx on older Ubuntu servers? I'm not exactly sure. Given that it's not documented, I'd be happy to merge a PR that just adds it. There is a residual risk that someone will come and complain and we have to undo it, but I'm willing to try that out. :)

PR #4433, opened 2021-02-08
Issue 4390: Changed webaddons.py as per suggestion

alt-glitch:
As per @mhils suggestion in #4390, added `webbrowser.open_new(url)` so that running `mitmweb` opens in default browser.

### DIFF
--- a/mitmproxy/tools/web/webaddons.py
+++ b/mitmproxy/tools/web/webaddons.py
@@ -47,8 +47,7 @@ def open_browser(url: str) -> bool:
         webbrowser.open_new(url)
         return True
     except webbrowser.Error:
-        return False
-    
+        return False   
     """
     As mentioned in issue #4390, adding webbrowser.open_new(url) so that the default browser opens on running mitmweb
     """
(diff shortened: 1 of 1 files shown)


mhils (changes requested):
I appreciate your enthusiasm, but this is not very helpful. Did you read the docstring at the beginning of the function? There is a reason why we don't call `webbrowser.open` directly.

That being said, this change is particularly hard to test for. If you submit a PR for this, I expect you to be on a system which exhibits the problem mentioned in #4390, then add `xdg-open`, and confirm that it solves the problem.

alt-glitch:
> I appreciate your enthusiasm, but this is not very helpful. Did you read the docstring at the beginning of the function? There is a reason why we don't call `webbrowser.open` directly.
> 
> That being said, this change is particularly hard to test for. If you submit a PR for this, I expect you to be on a system which exhibits the problem mentioned in #4390, then add `xdg-open`, and confirm that it solves the problem.

I did read the docstring but I assumed that you wanted to try and go ahead with adding it. I misunderstood the situation. I shall make the desirable changes with the testing and documentation you mentioned.

PR #4435, opened 2021-02-08
Issue 4390: Using xdg-open to open default browser

alt-glitch:
As referenced in #4390, I have confirmed that running `mitmweb` opens the Chromium browser:
image

On using `xdg-open` for this, `mitmweb` opens the default browser; in this case Brave:
image

I have used the subprocess module to run `xdg-open`. On an unsuccessful run of the `xdg-open` function (returns a non-zero integer); the code would default to the previous way of opening the browser.


mhils:
Thanks for verifying this, that's the important part. Please just add `xdg-open` to the list similar to how `wslview`, `gnome-open`, and friends are invoked. :)

alt-glitch:
@mhils: I've added `xdg-open` to the list like you asked and verified it once again for good measure :)

mhils:
Thanks!

Issue #4438, opened 2021-02-10
Way to check if script loading is failed


abitrolly:
I spent quite a few days to debug a problem with missing import in HAR script. Now I need to write a test to fail when any exception occurs during the load, and I can find to detect if there was an exception.

For example, if `../examples/contrib/har_dump.py` is invalid, the following snippet of code will only fail if there is no `a.addons` field. Logs are empty for some reason.

```
      def test_simple(self, tmpdir, tdata):
          with taddons.context() as tctx:
              fullpath = tdata.path("../examples/contrib/har_dump.py")
              a = script.load_script(fullpath)
              assert tctx.master.logs == []
              assert a.addons
```

#### Proposal

A way to check for errors during script load. Maybe there is bug in logger. I don't know.

mhils:
I'm not sure I understand what the problem is here. Can you give an example of what doesn't work?

FWIW, they way I develop addons is that I start `mitmproxy -s ...`, open the event log in a second pane (`-`), and then try out the functionality. If stuff doesn't work, you should be able to edit the script file and it should be automatically reloaded. Once I am happy with my addon, I write some tests that make sure it doesn't break in the future. If I understand you correctly you are working more with a test-driven workflow, is that correct?

abitrolly:
Yep. For integration tests it is important to track API breaks. When script loading starts to fail, I expect to see the exact exception in tests.

Prinzhorn:
Would a new flag make sense like `--scripts-mandatory` (or whatever) that will crash the mitmproxy process when an addon crashes? This might be useful for cases where you absolutely need all-or-nothing and don't want mitmproxy to continue without an addon loaded? The current architecture is great to develop addons but in production one might not want mitmproxy to "silently" continue without a given addon. Or am I misunderstanding the issue?

> I write some tests that make sure it doesn't break in the future

I think you mean to make it _less likely_ to break in the future :smile:

abitrolly:
Not sure about the new option for `mitmproxy`. I guess it is fine if the loading error is present in the event log, but in tests the `ctx.master.logs` was empty even if script failed to parse.

So I'd expect the traceback/crash mode to be on by default for tests and `mitmdump -s <script>`. For interactive tool like `mitmproxy` it is probably not a good idea to crash.

mhils:
We need to distinguish a bit between mitmproxy itself and our (internal) tests here.

[...]

PR #4439, opened 2021-02-11
Fix HAR dump

### LINKED ISSUE #3377: Har file is not being written to disk when using har dump addon with Mitmdump 4.0.4
artifacts:
1. mitmdump 4.0.2 - downloaded release binaries from the site marked as 4.0.4 (which is a bit odd)
2. har_dump example addon from github repository from '4.x' branch.

1. Try to run 'mitmdump' from cmd, with the mitmdump example addon.
2. No har file is created on exit 

**also reproduced on MacOS.

Works perfectly with mitmdump 3.0.4

##### System information
Mitmproxy: 4.0.2 binary
Python:    3.6.5
Platform:  Windows-10-10.0.17134

### LINKED ISSUE #3855: HAR file does not appears when I try to use har_dump.py script
HAR file does not appears when I try to use har_dump.py script calling from command line

1. Type **mitmdump -s har_dump.py --set hardump=dump.har**
2. Press Enter
3. Do something in browser via proxy
4. Press Ctr+C

File dump.har doesn't present in my hard disc at all

System Information:

Mitmproxy: 5.0.1
Python:    3.6.3
OpenSSL:   OpenSSL 1.1.0j  20 Nov 2018
Platform:  Windows-10-10.0.18362-SP0

abitrolly:
The reason was an invalid import of `connections` which is gone in d159897d98d47034827714bd0298268d1b9af6f0, but the loading error was silently ignored (#4438) resulting in a misleading exception at a later stage.

The way to run tests from examples.

    ln -sr test/conftest.py examples
    pytest examples/contrib/test_har_dump.py -k simpl

The way to see the execution flow with silenced exceptions.

    from hunter import trace, Q
    trace(Q(stdlib=False), ~Q(filename_contains='site-packages'))

#### Checklist

### References

Fixes #3377
Fixes #3855

### DIFF
--- a/examples/contrib/har_dump.py
+++ b/examples/contrib/har_dump.py
@@ -20,7 +20,7 @@ from datetime import timezone
 
 import mitmproxy
 
-from mitmproxy import connections  # noqa
+from mitmproxy import connection
 from mitmproxy import version
 from mitmproxy import ctx
 from mitmproxy.utils import strutils
@@ -30,7 +30,7 @@ HAR: typing.Dict = {}
 
 # A list of server seen till now is maintained so we can avoid
 # using 'connect' time for entries that use an existing connection.
-SERVERS_SEEN: typing.Set[connections.ServerConnection] = set()
+SERVERS_SEEN: typing.Set[connection.Server] = set()
 
 
 def load(l):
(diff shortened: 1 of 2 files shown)


mhils:
Fantastic, thank you so much! 🍰

PR #4441, opened 2021-02-12
Try to test examples in the GitHub Actions

abitrolly:
Examples with tests are broken. Probably other examples are broken too. Could be a great source for tracking API changes and documenting the upgrade path.

#### Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -64,6 +64,18 @@ jobs:
         with:
           file: ./coverage.xml
           name: ${{ matrix.os }}
+
+  test_examples:
+    runs-on: ubuntu-latest
+    steps:
+      - run: printenv
+      - uses: actions/checkout@v2
+      - run: pip install pipenv
+      - run: pipenv install
+      - run: |
+          ln -sr test/conftest.py examples/
+          pipenv run pytest examples/
+
   build-wheel:
     runs-on: ubuntu-latest
     env:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,11 +70,10 @@ jobs:
     steps:
       - run: printenv
       - uses: actions/checkout@v2
-      - run: pip install pipenv
-      - run: python -m pipenv install -e .
+      - run: pip install pytest -e ".[dev]"
       - run: |
           ln -sr test/conftest.py examples/
-          python -m pipenv run pytest examples/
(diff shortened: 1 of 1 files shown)


mhils:
We do have some tests for some examples at https://github.com/mitmproxy/mitmproxy/tree/master/test/examples, I absolutely agree that we should ideally test more here.

The story is a bit different when we look at `examples/contrib`: The addons in there often have niche use cases or do not meet our general code standards. I have tried to maintain that in the past, but practically it's been slowing us down more than anything. There is a note in https://github.com/mitmproxy/mitmproxy/blob/master/examples/contrib/README.md that we don't maintain whatever is in there at all, do we need to make that more explicit/prominent?

<small>(As mentioned elsewhere, HAR Support is important enough to move to mitmproxy/addons once it's polished, where we 1) maintain 100% test coverage and 2) maintain it.)</small>

abitrolly:
Maybe it doesn't worth time to maintain examples in the sense of porting them to the latest features, but at least providing backward compatibility API so that they won't break would be beneficial for those who are dealing with their own or legacy extensions.

abitrolly:
I might fix a few tests as a result of this PR.

abitrolly:
Actually, I don't have much time to go deep into it right now. Some of the errors are import errors and I am not sure how to provide right imports for each example to test in automated way. So I better leave it for now and concentrate on HAR when/if I get more time. It may worth to note how to actually run tests in `example/` as the trick with copying fixtures it is not trivial.

The current results just for the history, because GitHub will clean up the build log.
```
=========================== short test summary info ============================
FAILED examples/contrib/webscanner_helper/test_urlindex.py::TestUrlIndexAddon::test_init_append
FAILED examples/contrib/webscanner_helper/test_urlindex.py::TestUrlIndexAddon::test_response
FAILED examples/contrib/webscanner_helper/test_urlindex.py::TestUrlIndexAddon::test_done
ERROR examples/contrib/test_jsondump.py
ERROR examples/contrib/test_xss_scanner.py
ERROR examples/contrib/webscanner_helper/test_mapping.py
ERROR examples/contrib/webscanner_helper/test_proxyauth_selenium.py
ERROR examples/contrib/webscanner_helper/test_watchdog.py
==================== 3 failed, 40 passed, 5 errors in 1.80s ====================
Error: Process completed with exit code 1.
```

PR #4446, opened 2021-02-13
Fixes #4416

### LINKED ISSUE #4416: Omitting a forward slash from http domain produces Bad HTTP request line
#### Problem Description
Running a GET request for "http://wikipedia.com" will produce the following output:
`35.176.198.120:43784: clientconnect
 << Bad HTTP request line: b'GET http://wikipedia.com HTTP/1.1'
::ffff:35.176.198.120:43784: request
  -> HTTP protocol error in client request: Bad HTTP request line: b'GET http://wikipedia.com HTTP/1.1'
35.176.198.120:43784: clientdisconnect`

The request status, headers and body are as follows for the client:
`400
<CIMultiDictProxy('Server': 'mitmproxy 5.3.0', 'Connection': 'close', 'Content-Length': '306', 'Content-Type': 'text/html')>
b'<html>\n            <head>\n                <title>400 Bad Request</title>\n            </head>\n            <body>\n            <h1>400 Bad Request</h1>\n            <p>HttpSyntaxException(&quot;Bad HTTP request line: b&#x27;GET http://wikipedia.com HTTP/1.1&#x27;&quot;)</p>\n            </body>\n        </html>`

Running a GET request for "http://wikipedia.com/" will not produce this error. Nor will "https://wikipedia.com".
Curl runs the problematic URL fine as it appends the missing trailing slash before running the request which can be seen by printing flow.request.url.

1. Start mitmproxy using "mitmdump -s proxy_server.py --set block_global=false"
2. Run a GET request with mitm proxy as the request proxy for "http://wikipedia.com".

#### System Information
Mitmproxy: 5.3.0
Python:    3.7.9
OpenSSL:   OpenSSL 1.1.1h  22 Sep 2020
Platform:  Linux-4.14.203-156.332.amzn2.x86_64-x86_64-with-glibc2.2.5

dlbas:
Fix ValueError when splitting on a request URI without a path part.

Consulted [HTTP 1.1 spec](https://tools.ietf.org/html/rfc2616#section-5.1.2) where it says that proxy should append "/" to URI's path if it already doesn't have one.

#### Checklist

### DIFF
--- a/mitmproxy/net/http/http1/read.py
+++ b/mitmproxy/net/http/http1/read.py
@@ -111,8 +111,8 @@ def _read_request_line(line: bytes) -> Tuple[str, int, bytes, bytes, bytes, byte
                 raise ValueError
         else:
             scheme, rest = target.split(b"://", maxsplit=1)
-            authority, *path_ = rest.split(b"/", maxsplit=1)
-            path_ = path_[0] if path_ else b""
+            authority, *paths_ = rest.split(b"/", maxsplit=1)
+            path_ = paths_[0] if paths_ else b""
             path = b"/" + path_
             host, port = url.parse_authority(authority, check=True)
             port = port or url.default_port(scheme)
--- a/mitmproxy/net/http/http1/read.py
+++ b/mitmproxy/net/http/http1/read.py
@@ -111,7 +111,8 @@ def _read_request_line(line: bytes) -> Tuple[str, int, bytes, bytes, bytes, byte
                 raise ValueError
         else:
             scheme, rest = target.split(b"://", maxsplit=1)
-            authority, path_ = rest.split(b"/", maxsplit=1)
+            authority, *path_ = rest.split(b"/", maxsplit=1)
+            path_ = path_[0] if path_ else b""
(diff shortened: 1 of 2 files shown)


mitmproxy/net/http/http1/read.py
```diff
@@ -111,7 +111,8 @@ def _read_request_line(line: bytes) -> Tuple[str, int, bytes, bytes, bytes, byte
                 raise ValueError
         else:
             scheme, rest = target.split(b"://", maxsplit=1)
-            authority, path_ = rest.split(b"/", maxsplit=1)
+            authority, *paths_ = rest.split(b"/", maxsplit=1)
+            path_ = paths_[0] if paths_ else b""
```
mhils, line 115: Can you use `authority, _, path = rest.partition(b"/")` here? That should make stuff a bit cleaner.

dlbas: Absolutely! Never thought of it, for some reason :)

mhils:
Excellent, thanks!

Issue #3712, opened 2019-11-17
Docker: Move away from Alpine


mhils:
On my quest to shorten CI times, building Docker images has become a major bottleneck. A major issue here is that we use Alpine Linux, which means we can't use precompiled manylinux wheels (Alpine uses musl...).

@mitmproxy/devs: any opinions on moving to a minimal Ubuntu image? I feel that should be just fine.

Kriechi:
👍

ylmrx:
:+1:

mhils:
I'm closing this for now as the switch to GitHub actions means we don't need this for PRs. The pressure for master branch builds is less severe, to say the least.

mhils:
`cryptography` now includes Rust code, which makes an Ubuntu-based image that can consume prebuilt wheels much more appealing.

alt-glitch:
Hi, I'd like to work on this issue. If I understand correctly, we are trying to move from an Alpine Linux Docker Image for mitmproxy to a small Ubuntu/Debian one right? And the reasoning for this is to shorten the Docker build times if I'm not wrong? How should I go about building the image and what steps am I supposed to follow?

mhils:
Yes! It's probably the easiest to base it off [python:3.9-buster](https://hub.docker.com/_/python). You basically need to install mitmproxy with pip and make sure that it starts properly. :)

hazcod:
@mhils honest question, why not python:3.9-slim-buster ?

mhils:
@hazcod: From https://hub.docker.com/_/python

> Unless you are working in an environment where only the python image will be deployed and you have space constraints, we highly recommend using the default image of this repository.

mhils:
@hazcod: I retract my previous statement. Looking at the Python Dockerfile, the slim variants are based upon debian:-slim, which is perfectly fine I think. The image is _much_ smaller, let's just use it.

PR #4450, opened 2021-02-15
Add 'tzdata' package to Docker image

AltoNyan:
#### Description

Add `tzdata` package to Docker image by appending `tzdata` to Dockerfile's `apk add ...` section.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ If you depend on these features, please raise your voice in
 * `--modify-headers` now works correctly when modifying a header that is also part of the filter expression (@Prinzhorn)
 * Fix SNI-related reproducibility issues when exporting to curl/httpie commands. (@dkasak)
 * Add option `export_preserve_original_ip` to force exported command to connect to IP from original request. Only supports curl at the moment. (@dkasak)
+* Add `tzdata` package to Docker image  (@AltoNyan)
 * --- TODO: add new PRs above this line ---
 * ... and various other fixes, documentation improvements, dependency version bumps, etc.
 
--- a/release/docker/Dockerfile
+++ b/release/docker/Dockerfile
@@ -21,6 +21,7 @@ RUN addgroup -S mitmproxy && adduser -S -G mitmproxy mitmproxy \
         openssl-dev \
         python3 \
         python3-dev \
+        tzdata \
     && python3 -m ensurepip --upgrade \
     && pip3 install -U pip \
     && LDFLAGS=-L/lib pip3 install -U /home/mitmproxy/${WHEEL_BASENAME_MITMPROXY} \


mhils:
Thanks for the PR! How does the lack of tzdata influence mitmproxy?

AltoNyan:
Lack of `tzdata` makes we can't able to set a timezone in the Docker container.

Of course, mitmproxy working without a timezone well, but we can not sync timezone with host.

And, this makes some inconvenience with tasks like scheduling.

mhils:
So this affects your particular project, but not mitmproxy users in general?

AltoNyan:
No, I think timezone support is a common thing in Docker container, and this is why I make a pull request.

This is not for my particular project, I just think this is better.

Of course, if you think it's useless or unnecessary it's okay to ignore this PR.


ps. Make a pull request is the first time for me. So, I'm sorry if something is wrong or weird.

mhils:
No worries, sorry if I've come off harsh! I'm just trying to understand the use case here. 😃 

Can you give me a concrete example of some use case that only works when tzdata is installed in the mitmproxy docker image? When/why do you want to set a timezone in a Docker container, and how? What kind of scheduling are you doing here?

mhils:
Side note: we also want to switch from Alpine to a debian/ubuntu base image, probably https://hub.docker.com/_/python (#3712). Not sure what's already installed there.

AltoNyan:
> No worries, sorry if I've come off harsh! I'm just trying to understand the use case here. 😃
> 
> Can you give me a concrete example of some use case that only works when tzdata is installed in the mitmproxy docker image? When/why do you want to set a timezone in a Docker container, and how? What kind of scheduling are you doing here?

Thanks for the kind reply. 😃 

Without tzdata, the Alpine Docker container cannot set timezone. So, its docker logs and the other time-related records are based on UTC only.
But, there is nothing precisely wrong or not working with it.  Just inconvenience and some efforts are needed to correct it.


When the first time I want to set timezone on the container is that I'm reading logs on the docker container.
Its timestamps are based on UTC. So, it's different from the current time.

Basically, there's two common way to sync timezone with the host.
1. Bind mount `/etc/localtime:/etc/localtime:ro`
2. Set `TZ` environment variable on docker run.

But it's not working with the Alpine-Linux image. (because lack of the `tzdata`). So I just install `tzdata` package inside of the container.

Schedule an update of some resources for cache based on days. And it has to convert its time based on UTC.

AltoNyan:
> Side note: we also want to switch from Alpine to a debian/ubuntu base image, probably https://hub.docker.com/_/python (#3712). Not sure what's already installed there.


Wow, now I checked both images 'buster' and 'slim-buster' on https://hub.docker.com/_/python.

And, it perfectly works with the 'TZ' environmental variable to set timezone.

mhils:
Thanks for the explanation! Let's focus on moving the image to `python:3.9-buster` then, PRs welcome!

PR #4462, opened 2021-02-20
HTTP/2: Show human-readable error messages

mhils:
As reported in https://github.com/mitmproxy/mitmproxy/issues/4451#issuecomment-782576657: When encountering an error in the *mitmproxy - server* connection, we previously just reset the matching *client - mitmproxy* HTTP/2 stream. This does not work as intended, as the user is left in the dark what the actual error cause is. This PR fixes this by sending an error document if headers have not been sent yet.

The diff is a bit messier than usual as I configured my IDE to automatically sort imports, this should settle over time. Again, I strongly feel this convention is nothing we should enforce in CI because it adds completely unnecessary friction to PRs, but we sure can do a best (automated) effort here. 😃


r00t-:
yay, those error messages are a good thing!

not an issue for me (and also applicable to http/1):
i wonder if it's in all cases ok to disclose to the client that their request has been intercepted. an application could try to connect to a known bad host and find they get an unexpected response. some researchers with focus on security might want a "stealth mode". but then, certificate pinning has the same effect already.

mhils:
> i wonder if it's in all cases ok to disclose to the client that their request has been intercepted. an application could try to connect to a known bad host and find they get an unexpected response. some researchers with focus on security might want a "stealth mode". but then, certificate pinning has the same effect already and is already quite common.

There is a fine line between maximum compatibility with clients and servers and "stealth mode". We want to keep requests and responses as-is when they pass through mitmproxy, so we don't add a `X-Forwarded-For` header for example. At the same time, we explicitly don't want a stealth mode. Mitmproxy's default CA certificate always has "mitmproxy" as the organization (despite repeated requests to make it configurable), and if we have an error, we error right into the client's face. Users should absolutely know that their traffic is being intercepted with mitmproxy. We need to be pretty mindful here there's potential for abuse, we want to err on the side of being a bit less stealthy.

r00t-:
perfectly valid reasoning.
i was thinking of reverse-engineering hostile applications (malware, or commercial phone apps ),
but that's only a remote edge case compared to possible abuse, yes.

PR #4464, opened 2021-02-21
Speculative HTTP/2 fixes, refs #4451

mhils:
@r00t-'s comment in #4451 pointing to Facebook was a useful one - their custom HTTP/2 stack exposed some interesting race conditions in ours, which I believe I fixed all now.

@r00t-: Could you check out this branch and see if that fixes your Facebook POST requests in particular? 🍰

### DIFF
--- a/mitmproxy/proxy/layers/http/__init__.py
+++ b/mitmproxy/proxy/layers/http/__init__.py
@@ -461,7 +461,7 @@ class HttpStream(layer.Layer):
             )
 
         if 200 <= self.flow.response.status_code < 300:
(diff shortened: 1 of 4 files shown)


r00t-:
guess i should have reported this earlier...
i've had those issues uploading to facebook for a long time.
but i've all the time been assuming that it's a plain http/1 POST, and that the connection being closed could only be due to the server timing out because the request is taking too long :\

r00t-:
i did try uploading a few large posts with this branch, and had a 100% success rate, as opposed to near zero previously :+1:

PR #4471, opened 2021-02-27
Don't stream bodies when content length is missing

Prinzhorn:
#### Description

`content-length` is optional and if it's missing `expected_http_body_size` returns `-1` which made the falsy check for `expected_size` true. I've simplified the condition. Not sure for what use-case `raw_content` is checked? There was no test for that case.

I think having some false negatives (buffering a body that is larger than we want) is way better than having some false positives (missing out on bodies that would've been small enough). I guess that larger bodies will also have a content-length more commonly.

#### Checklist


Prinzhorn:
Arguably returning `-1` is fine here, but what about the `None` here? https://github.com/mitmproxy/mitmproxy/blob/4bd7b6c4eadeaca712f63e0e73f20bcf6aadbffb/mitmproxy/net/http/http1/read.py#L78-L79

This sure makes for some more weird edge cases. Maybe I just broke streaming for chunked encoding with this PR, not sure how `>` behaves in Python when it's compared with `None`.

PR #4481, opened 2021-03-07
Minor fixes

mhils:


mhils:
@Kriechi: Does 47982ca work for you? I know you may have opinions. 😃

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py38, py39, flake8, filename_matching, mypy, individual_coverage, docs
+envlist = py, flake8, mypy
```
Kriechi, line 2: this list only affects what is run into actively for us developers?
Or does this also affect what is run in CI?

mhils: My understanding is that this only affects what is run when you just type `tox` on the command line, see the commit message.

Kriechi (approved):

Issue #4131, opened 2020-08-06
Additional `text` property for WebSocketMessage or whatever


Prinzhorn:
#### Is your feature request related to a problem? Please describe.

Looking at the type declaration of `content` here

https://github.com/mitmproxy/mitmproxy/blob/b4bd1741eb48971a62b26d366bc6e2768350783e/mitmproxy/websocket.py#L19-L21

I was assuming `content` would always be of type `bytes`. However, it is of type `str` if the opcode is TEXT. The issue lies here
https://github.com/mitmproxy/mitmproxy/blob/70777a1b6ed64af9cafcdef223a8a260ecc96864/mitmproxy/proxy/protocol/websocket.py#L101-L110

I know it's kind of common to do that in a scripting language, but it's also harder to work with.

#### Describe the solution you'd like

I'm not sure if we should make a breaking change and introduce `text` similar to how it works with HTTP MessageData? And then keep `content` for the raw bytes? I'm open for anything, I was just irritated that the type is not consistent.

#### Describe alternatives you've considered

I'm using gRPC and had the type declared as `bytes`. I want the raw bytes. I guess I could do the conversion on my end and get the bytes of `content` if the opcode is TEXT. Working around it is not the issue, but I'd rather have consistency.

mhils:
> I'm not sure if we should make a breaking change and introduce text similar to how it works with HTTP MessageData? And then keep content for the raw bytes? I'm open for anything, I was just irritated that the type is not consistent.

I really haven't looked at this for more than a second, but I remotely recall something about WebSocket TEXT frames having a very strictly defined encoding? It looks like we just forward what we get from wsproto. @Kriechi?

The second much simpler issue is that the type annotation is wrong in any case and needs to be fixed. :)

Prinzhorn:
> I really haven't looked at this for more than a second, but I remotely recall something about WebSocket TEXT frames having a very strictly defined encoding? It looks like we just forward what we get from wsproto. @Kriechi?

I think so too (it looks it it's [strictly UTF-8](https://stackoverflow.com/a/43530268/1422124)) and wasn't implying that we need to do the encoding dance in the text getter/setter. I would be fine with keeping just `content` the way it currently is. It does make sense since it's _either_ text or binary and not like with HTTP bodies where you don't know for sure sometimes and want to string replace on jpeg meta data or whatever. I was just tired yesterday.

> The second much simpler issue is that the type annotation is wrong in any case and needs to be fixed. :)

[...]

Issue #3396, opened 2018-11-29
Injecting multiple websocket messages freezes mitmproxy


andresriancho:
https://github.com/mitmproxy/mitmproxy/pull/3119 added the feature to inject new messages into existing websocket connections.

The injected messages get stored in two queues, defined [here](https://github.com/mitmproxy/mitmproxy/pull/3119/files#diff-0de40cbc6458548283e26e099bf4edccR84). The queues have `maxsize=1`.

During an application security assessment I was using `mitmproxy` and noticed that when I wanted to inject more than one message the whole `mitmproxy` (actually mitmdump but I guess it is the same) froze while trying to `put()` the second message to the queue.

After a little bit of debugging I changed the size of those queues (to 100), and my code started to work.

##### Steps to reproduce the problem:
Create a script like:

```python
def websocket_message(flow: websocket.WebSocketFlow) -> None:
    flow.inject_message(flow.server_conn, '123')
    flow.inject_message(flow.server_conn, '123')
```

Run it and the code will freeze in the second call to `inject_message`

##### Any other comments? What have you tried so far?
Not sure if I'm doing something wrong: maybe I should do this in an async way? Maybe I should wait until the queue is empty?

Anyways, wanted to report this because it might be a bug that others might hit too.

##### System information
Latest mitmproxy from github.

CapitanShinChan:
Quick question, can you update the messages in in mitm proxy before sending them to the destination?

andresriancho:
AFAIK, no.

Kriechi:
@CapitanShinChan @andresriancho: updating a message before it is sent to the destination should be simple: see this example https://github.com/mitmproxy/mitmproxy/blob/db658b12edf9a44e40ca79209a652e839ffa78dd/examples/simple/websocket_messages.py

@andresriancho for the actual issue you reported here:
Thanks for opening this issue - unfortunately I don't remember why I chose `max_size=1`. 
I think the main idea of limiting the queue size is, to allow the connection to do protocol-level tasks, such as ping+pong frames, without overwhelming it with 1000's of injected messages.

I would be happy to merge a PR that increases the queue size to a reasonable amount, including some form of comment/documentation on the drawbacks and your original findings (i.e., freezing issues with too small queues).

CapitanShinChan:
@Kriechi I tried what you suggest, but the content of the message doesn't get updated.

[Issue with the information](https://github.com/mitmproxy/mitmproxy/issues/3391)

[...]

PR #4486, opened 2021-03-08
Merge WebSocketFlow into HTTPFlow, add WebSocket UI

### LINKED ISSUE #4425: Combine WebSocketFlow and HTTPFlow
HTTPFlow and WebSocketFlow currently have a really ugly relation where one depends on the other, they appear separately in dumps, etc. To fix this madness, I would propose we merge WebSocketFlow and HTTPFlow: HTTPFlow just gets a `.websocket: Optional[WebSocketData]` attribute, which contains everything related to the websocket connection (messages, close reason, etc). There also is prior art for this approach: For example, the Chrome Devtools use exactly that data model.

@mitmproxy/devs, thoughts?

mhils:
This PR fixes #4425 and then adds WebSocket Message display in mitmprox as a cherry on top! 🍒 

image


mitmproxy/addons/dumper.py
```diff
@@ -277,37 +277,37 @@ def error(self, f):
...
-            "Error in WebSocket connection to {}: {}".format(
-                human.format_address(f.server_conn.address), f.error
-            ),
+            f"Error in WebSocket connection to {human.format_address(f.server_conn.address)}: {f.error}",
             fg="red"
         )
 
-    def websocket_message(self, f):
+    def websocket_message(self, f: http.HTTPFlow):
+        assert f.websocket is not None
```
Prinzhorn, line 287: Are these assertion necessary? If `websocket_message` is called without a websocket then the add-on core is broken, right?

mitmproxy/addons/save.py
```diff
@@ -88,28 +89,24 @@ def tcp_end(self, flow):
...
 
-    def request(self, flow):
+    def request(self, flow: http.HTTPFlow):
         if self.stream:
             self.active_flows.add(flow)
 
-    def response(self, flow):
-        if self.stream:
+    def response(self, flow: http.HTTPFlow):
+        if self.stream and flow.websocket is None:
```
Prinzhorn, line 105: Does this mean the HTTP 101 flow is not saved and lost? But it contains interesting data such as authentication tokens. Or why was that change needed?

mitmproxy/websocket.py
```diff
@@ -1,168 +1,136 @@
...
     from_client: bool
     """True if this messages was sent by the client."""
-    content: Union[bytes, str]
+    is_text: bool
+    """
+    True if the message is a text message, False if the message is a binary message.
+
+    In either case, mitmproxy will store the message contents as *bytes*.
+    """
+    content: bytes
```
Prinzhorn, line 40: What's the reason we're switching to bytes for text messages as well? We know that WebSocket text is UTF-8, doesn't that make it more awkward to work with? Or is that my lack of Python?

mitmproxy/websocket.py
```diff
@@ -1,168 +1,136 @@
...
 
     def set_state(self, state):
-        self.type, self.from_client, self.content, self.timestamp, self.killed = state
-        self.type = Opcode(self.type)  # replace enum with bare int
+        self.from_client, self.is_text, self.content, self.timestamp = state
 
     def __repr__(self):
-        if self.type == Opcode.TEXT:
-            return "text message: {}".format(repr(self.content))
+        if self.is_text:
```
Prinzhorn, line 68: Currently in my add-on I'm explicitly checking both opcodes because the docs say:

> Currently there are two types of messages, corresponding to the BINARY and TEXT frame types.

and I didn't want to run in a future scenario were my `else` would catch others in a later version.

Does this change conclude that we'll never expose PING/PONG/CLOSE? I guess that's the only reason the opcode was exposed originally?

mitmproxy/websocket.py
```diff
@@ -1,168 +1,136 @@
...
             "WebSocketMessage.kill is deprecated, set an empty content instead.",
             DeprecationWarning,
             stacklevel=2,
         )
-        # empty str or empty bytes.
-        self.content = type(self.content)()
+        self.content = b""
 
+    @property
+    def killed(self) -> bool:  # pragma: no cover
```
Prinzhorn, line 87: I only just realized that `kill` is also deprecated. I disagree with that decision, but I don't know what lead to that? Can you link to some discussion?

If I set the message to an empty string the original content is lost. In the UI it's just an empty message? But I want the user to know "Hey this message _would_ have been sent, but wasn't". In the same way we don't empty all HTTP headers and body if a flow is dropped. I don't want to lose any information.

mitmproxy/websocket.py
mhils: I haven't really made my mind up here to be honest. The previous implementation was annoying me quite a bit because `content` was `Union[str,bytes]`, and you always had to add a special case and check. Now my hope is that it's just more consistent to deal with, and if you need text you can just call `.decode()` yourself. @Kriechi, any opinions from your end?

mitmproxy/addons/save.py
Prinzhorn: For a moment I must've forgotten that `WebSocketFlow` is gone, makes sense now

mitmproxy/websocket.py
Prinzhorn: Déjà vu

https://github.com/mitmproxy/mitmproxy/issues/4131

> I remotely recall something about WebSocket TEXT frames having a very strictly defined encoding? It looks like we just forward what we get from wsproto. 

---

> and you always had to add a special case and check

How is that special case gone? You still need to decide when to `.decode()` and when not to.

I'm not favoring one solution or the other, but given that I've brought this exact topic up before maybe we can finally settle this :wink:

Prinzhorn: I swear this is the last comment: maybe my original idea of a `text` property kind of makes sense? It would just be a getter/setter that does the decode/encode for you. Otherwise every user would require the knowledge that WebSocket text frames are always utf-8 to manually do the `decode` of the bytes.

If `not is_text` the `text` property should crash or something tho? Because reading `text` for arbitrary `content` won't always work.

mitmproxy/websocket.py
mhils: Good point, we may want to leave that door open. I'll change it back to using the opcode. 👍

mitmproxy/websocket.py
mhils: There is none, but there probably should have been some - I was just ripping out as much complexity as I could find. You mention a valid use case though. Going back to the drawing board, what would you expect `kill` to do? Promptly abort the connection, or just suppress the current message? I just now realize that we also don't have a good story for interception either.

mitmproxy/proxy/layers/http/__init__.py
```diff
@@ -308,6 +309,25 @@ def send_response(self, already_streamed: bool = False):
         """We have either consumed the entire response from the server or the response was set by an addon."""
         assert self.flow.response
         self.flow.response.timestamp_end = time.time()
+
+        is_websocket = (
+            self.flow.response.status_code == 101
+            and
+            self.flow.response.headers.get("upgrade", "").lower() == "websocket"
+            and
+            self.flow.request.headers.get("Sec-WebSocket-Version", "") == "13"
```
Kriechi, line 318: Should we use wsproto constants here instead?

test/mitmproxy/proxy/layers/test_websocket.py
```diff
@@ -95,12 +94,13 @@ def test_upgrade(tctx):
...
-    assert len(flow().messages) == 2
-    assert flow().messages[0].content == "hello world"
-    assert flow().messages[0].from_client
-    assert flow().messages[1].content == b"hello back"
-    assert flow().messages[1].from_client is False
+    assert len(flow().websocket.messages) == 2
+    assert flow().websocket.messages[0].content == b"hello world"
+    assert flow().websocket.messages[0].from_client
+    assert flow().websocket.messages[0].is_text
+    assert flow().websocket.messages[1].content == b"hello back"
```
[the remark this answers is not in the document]
Prinzhorn: >  How about a `text` property that either works, or raises when called on a Binary message?

That's what I meant above with

> If `not is_text` the `text` property should crash or something tho?

What I don't like about this is that this means an imaginary user's addon works for five weeks and then it crashes because it sees a binary WebSocket for the first time. But on the other hand that's actually great, because if you don't check `is_text` then that's your fault and with the current design it would already result in unexpected behavior.

So yes, I think a `text` property that transparently handles the `encode`/`decode` and crashes when `not is_text` (both when reading and writing) sounds good to me.

Is it currently possible to change `is_text` (or the opcode rather, if this property is removed again)? I think it should be and the user instead should drop the message and inject a new one. Because convert from one to the other is not possible in all cases.

mitmproxy/websocket.py
Prinzhorn: > Going back to the drawing board, what would you expect `kill` to do? Promptly abort the connection, or just suppress the current message? I just now realize that we also don't have a good story for interception either.

First of all we might want to call it `drop` and not introduce more violence :wink: 

Calling `messages[-1].drop()` should not forward the current message to the other end but keep it in the `messages` list with `dropped == True`.

This also touches another area that I always found awkward: when handling the `websocket_message` event I care about the message that this event is about. But it doesn't have a special place and I have to access the messages list. This feels weird to me and even though there probably can't be any race conditions with the current design (?) I always wonder if `messages[-1]` is actually guaranteed to be the message that triggered `websocket_message`.

Maybe we can pass the message as an additional parameter to the event? That would make for a much nicer API because that's what the user cares about.

mhils: `drop` is an excellent suggestion. 👍 

Do we need separate `.drop()` and `.dropped`? Or does a single `.drop` property do the job?



I agree on your point regarding `messages[-1]`, even though there are no race conditions at the moment. :) I think adding an extra argument is good, that would also allow us to make `websocket_message` non-blocking, i.e. we can continue to handle messages while one is intercepted. But that is future work, let's get the basics down here.

Prinzhorn: > Do we need separate `.drop()` and `.dropped`? Or does a single `.drop` property do the job?

I thought it should be consistent with `flow.kill()` but can you even detect if a flow has been killed currently? I just peeked into the flow.py and there is no flag?

To me `recent_message.drop = True` feels less like it actually does something than doing `recent_message.drop()`, which is an imperative. But I guess we already use that pattern with `flow.stream = True`. So maybe we should turn this around and update `flow.drop = True` in a future change and remove `flow.kill()`?

Here's one counter argument: if there is just a single `message.drop` property you don't know if it has already been dropped or not, it depends on the context. Inside of `websocket_message` it means that it will be dropped after the event (or all events) has been processed, but if you inspect any older message in the list then `drop == True` means it had been dropped. The benefit of a property (see streaming) is that multiple chained events can work on it.

So maybe just a `drop` property and an additional new `active` (or `live`?) so that you know exactly what's up.

Prinzhorn: > But that is future work, let's get the basics down here.

Future work definitely references #3396 :wink:

mhils:
Thanks you two for the excellent feedback, that was really helpful. I propose we move the two outstanding discussion points -- drop/kill/... and `.text` -- to new issues. Any objections? This is good to go for my end. 😃

mitmproxy/proxy/layers/http/__init__.py
Kriechi: https://github.com/python-hyper/wsproto/blob/55bf7a4ee1e85e3ebc93f04708327b0d0f193925/src/wsproto/handshake.py#L36

mhils: Yes that's good. We leave it hardcoded in the tests, so we have an obvious breakage if one of them ever changes.

mitmproxy/websocket.py
```diff
@@ -1,168 +1,126 @@
...
+    """
+    A data container for everything related to a single WebSocket connection.
+    This is typically accessed as `mitmproxy.http.HTTPFlow.websocket`.
+    """
 
+    messages: List[WebSocketMessage]
+    """All `WebSocketMessage`s transferred over this connection."""
 
-class WebSocketFlow(flow.Flow):
+    close_by_client: Optional[bool] = None
```
Prinzhorn, line 98: I know the uniform `close_` prefix is nice but `the close was done by the client` reads odd. How about `closed_by_client` which is short for `the WebSocket was closed by the client`

Prinzhorn:
The WebSocket view keeps jumping to the top every time a new message arrives. This makes it basically impossible to work with while the connection is open. I hold down arrow -> it scrolls a bit -> message arrives -> I'm back at the top

mitmproxy/proxy/layers/http/__init__.py
```diff
@@ -308,6 +309,25 @@ def send_response(self, already_streamed: bool = False):
...
+            self.context.options.websocket
+        )
+        if is_websocket:
+            # We need to set this before calling the response hook
+            # so that addons can determine if a WebSocket connection is following up.
+            self.flow.websocket = WebSocketData()
+            if self.flow.request.scheme == "http":
+                self.flow.request.scheme = "ws"
+            elif self.flow.request.scheme == "https":
+                self.flow.request.scheme = "wss"
```
Prinzhorn, line 329: I have hard feelings about this. `flow` is an HTTPFlow that happens to have a WebSocket using the same connection. If this is needed for the UI then the UI layer needs to handle this logic. But the scheme should never change.

```py
def requestheaders(flow):
    if(flow.request.pretty_host == 'echo.websocket.org'):
      print('requestheaders')
      print(flow.request.scheme)

def request(flow):
    if(flow.request.pretty_host == 'echo.websocket.org'):
      print('request')
      print(flow.request.scheme)

def responseheaders(flow):
    if(flow.request.pretty_host == 'echo.websocket.org'):
      print('responseheaders')
      print(flow.request.scheme)

def response(flow):
    if(flow.request.pretty_host == 'echo.websocket.org'):
      print('response')
      print(flow.request.scheme)
```

```
requestheaders
https

request
https

responseheaders
https

response
wss
```

giphy

Prinzhorn: Maybe a `flow.websocket.scheme` property?

mitmproxy/websocket.py
```diff
@@ -1,168 +1,126 @@
...
-        It will not be sent to the other endpoint. This has no effect in streaming mode.
-        """
-        warnings.warn(
-            "WebSocketMessage.kill is deprecated, set an empty content instead.",
-            DeprecationWarning,
-            stacklevel=2,
-        )
-        # empty str or empty bytes.
-        self.content = type(self.content)()
+class WebSocketData(stateobject.StateObject):
```
Prinzhorn, line 89: What do you think about `WebSocketDetails`? `data` has a certain connotation

Prinzhorn:
Despite my comments this looks amazing :tada:
I'm done for now

mitmproxy/websocket.py
mhils: I picked "Data" over "Details" as I felt it's really not just the details, it's also the main part. 😛 
What connotation do you have for `data`?

Other alternatives:

 - WebSocketConnection: Bad idea IMO, that clashes with `mitmproxy.connection`.
 - WebSocketFlow: Bad idea IMO, this is not a top-level flow.
 - WebSocketPart
 - WebSocketExchange
 - WebSocketInfo
 - WebSocketMessages

Nothing really convinces me, I feel that "data" as in dataclass is the most straightforward.

Prinzhorn: `WebSocketData` sounds great :+1: :smile:

mhils:
> The WebSocket view keeps jumping to the top every time a new message arrives. This makes it basically impossible to work with while the connection is open. I hold down arrow -> it scrolls a bit -> message arrives -> I'm back at the top

Why do attempt actual real-world usage instead of synthetic tests with only one or two messages? Stop being weird. 🤷 

Jokes aside, yes this is bad. 😄 Agree that we need to fix it (likely also for TCP), but not as part of this PR.

Prinzhorn:
> Why do attempt actual real-world usage instead of synthetic tests with only one or two messages?

We could also put "UI with support for up to five WebSocket messages" on the box and make it sound like that's a good thing :smile:

mitmproxy/proxy/layers/http/__init__.py
mhils: I don't think it's that bad, but I don't feel as strongly. I've changed it to the UI layer. 👍 
(`flow.websocket.scheme` is not really necessary)

Issue #4348, opened 2020-12-13
New Proxy Core: Vote on Feature Priority!


mhils:
mitmproxy 7 will ship with a completely rewritten proxy core, that has been merged into master today (see [CHANGELOG.md](https://github.com/mitmproxy/mitmproxy/blob/master/CHANGELOG.md)). However, a few features from the old proxy core did not make it to the new implementation yet.

 We want to get an idea what we should tackle first, so please vote on the comments below to indicate what we should focus on.

As always, we very much welcome pull requests on all of the features. Please feel free to ping me on our developer chat if you have implementation questions! 😃

mhils:
# HTTP Trailers

mhils:
### WebSocket/TCP Message Injection

mhils:
### SOCKS 5 Proxy Mode

mhils:
### HTTP/2 Push Promises

mhils:
### body_size_limit option

mhils:
### upstream_auth option

mhils:
### upstream_bind_address option

DiskGetor:
Translate to golang with fyne ui

jenniejj:
The ability to easily configure/start/stop a mitmproxy while being able to get the requests/responses in real-time from within a python script without the need for running subprocesses of mitmdump would be lovely.

mhils:
I intended this issue to only track regressions from 6.x here, so I've taken the liberty to mark some comments as off-topic. :)

wildwind123:
https://github.com/mitmproxy/mitmproxy/issues/3457

wildwind123:
Usefull
https://github.com/mitmproxy/mitmproxy/issues/3457

mhils:
Thanks y'all for the feedback! With the exception of HTTP/2 push promises all missing parts have been implemented and I'm looking forward to shipping mitmproxy 7 soon! 😃

PR #4489, opened 2021-03-10
support http/2 response trailers

sanlengjingvv:
#### Checklist

### DIFF
--- a/mitmproxy/proxy/layers/http/_http2.py
+++ b/mitmproxy/proxy/layers/http/_http2.py
@@ -17,7 +17,7 @@ from mitmproxy.connection import Connection
 from mitmproxy.net.http import status_codes, url
 from mitmproxy.utils import human
 from . import RequestData, RequestEndOfMessage, RequestHeaders, RequestProtocolError, ResponseData, \
-    ResponseEndOfMessage, ResponseHeaders, ResponseProtocolError
+    ResponseEndOfMessage, ResponseHeaders, ResponseTrailers, ResponseProtocolError
 from ._base import HttpConnection, HttpEvent, ReceiveHttp, format_error
 from ._http_h2 import BufferedH2Connection, H2ConnectionLogger
 from ...commands import CloseConnection, Log, SendData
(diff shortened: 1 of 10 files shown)


mhils (changes requested):
Hi @sanlengjingvv, thanks for the PR! I'm sorry that your previous HTTP/2 implementation got taken out when we switched to the new proxy core. I appreciate the effort here, but the implementation you are proposing here has the issue that trailers between concurrent flows may be mixed up. We need to define RequestTrailers/ResponseTrailers in [_events.py](https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/proxy/layers/http/_events.py), emit them when receiving trailers, extend [HttpStream](https://github.com/mitmproxy/mitmproxy/blob/a00f247f0edfb303bd54cb1f0b7b4e97c4a56fd0/mitmproxy/proxy/layers/http/__init__.py#L93) to accomodate for that, and then add tests. I do have that on my radar, but it's admittedly not a high priority item at the moment (https://github.com/mitmproxy/mitmproxy/issues/4348). If you would like to give it a stab I'd be more than happy to help and explain things, feel free to ping me on Slack as well!

mitmproxy/proxy/layers/http/_http2.py
```diff
@@ -448,6 +452,10 @@ def handle_h2_event(self, event: h2.events.Event) -> CommandGenerator[bool]:
             self.streams[event.stream_id] = StreamState.HEADERS_RECEIVED
             yield ReceiveHttp(ResponseHeaders(event.stream_id, response, bool(event.stream_ended)))
             return False
+        elif isinstance(event, h2.events.TrailersReceived):
+            pseudo_trailers, trailers = split_pseudo_headers(event.headers)
```
mhils, line 456: https://httpwg.org/specs/rfc7540.html#rfc.section.8.1.2.1:

> Pseudo-header fields MUST NOT appear in trailers. 

So we can skip that here :)

mhils (changes requested):
This looks great, thanks for the PR! 🍰 Some comments below. 😃

mhils (commented):
Thanks for the quick iteration! I have played around with this a bit further, added some streaming tests and then fixed a few streaming-related bugs. 

The most important change is that I have removed the requesttrailers/responsetrailers hooks. Instead, `request` and `response` now always fire after trailers have been received, but before they are forwarded. Can you test if things work for you as expected?

sanlengjingvv:
Replaying a request with trailers will get an error
> Unexpected event during replay: <module 'mitmproxy.proxy.events' from '/Users/niconi/mitmproxy/mitmproxy/proxy/events.py'>

mhils:
Fixed for me locally, does it work for you as well now? 😃

sanlengjingvv:
> Fixed for me locally, does it work for you as well now? 😃

Yes, it works for me now.

mhils (approved):

mhils:
Excellent. Thanks again for the initial PR and getting this going! 😃 🍰

PR #4493, opened 2021-03-11
Switch Docker release to be based on Debian instead of Alpine

### LINKED ISSUE #3712: Docker: Move away from Alpine

PeterDaveHello:
#### Description

As @mhils invited, help rewrite the Docker image to replace Alpine Linux. The final image size is about 179MB, compared with original 123MB, should be not too big :D

It's a large rewrite and I'm not very familiar with mitmproxy, so please help test its functionality to be as expected, thanks.

fix #3712

#### Checklist

### DIFF
--- a/release/docker/Dockerfile
+++ b/release/docker/Dockerfile
@@ -1,36 +1,17 @@
-FROM alpine:3.12
-
-ENV LANG=en_US.UTF-8
+FROM python:3.9-slim-buster
 
 ARG WHEEL_MITMPROXY
 ARG WHEEL_BASENAME_MITMPROXY
 
+RUN useradd -mU mitmproxy
 COPY $WHEEL_MITMPROXY /home/mitmproxy/
 
-# Add our user first to make sure the ID get assigned consistently,
-# regardless of whatever dependencies get added.
-RUN addgroup -S mitmproxy && adduser -S -G mitmproxy mitmproxy \
-    && apk add --no-cache \
-        su-exec \
-        git \
-        g++ \
-        libffi \
-        libffi-dev \
-        libstdc++ \
-        openssl \
-        openssl-dev \
-        python3 \
-        python3-dev \
-    && python3 -m ensurepip --upgrade \
-    && pip3 install -U pip \
-    && LDFLAGS=-L/lib pip3 install -U /home/mitmproxy/${WHEEL_BASENAME_MITMPROXY} \
-    && apk del --purge \
-        git \
-        g++ \
-        libffi-dev \
-        openssl-dev \
-        python3-dev \
-    && rm -rf ~/.cache/pip /home/mitmproxy/${WHEEL_BASENAME_MITMPROXY}
+RUN pip3 install --no-cache-dir -U /home/mitmproxy/${WHEEL_BASENAME_MITMPROXY} \
+    && rm -rf /home/mitmproxy/${WHEEL_BASENAME_MITMPROXY}
+
+RUN apt-get update \
+    && apt-get install -y gosu \
+    && rm -rf /var/lib/apt/lists/*
 
 VOLUME /home/mitmproxy/.mitmproxy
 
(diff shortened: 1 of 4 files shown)


release/docker/Dockerfile
```diff
@@ -1,36 +1,18 @@
...
-    && pip3 install -U pip \
-    && LDFLAGS=-L/lib pip3 install -U /home/mitmproxy/${WHEEL_BASENAME_MITMPROXY} \
-    && apk del --purge \
-        git \
-        g++ \
-        libffi-dev \
-        openssl-dev \
-        python3-dev \
-    && rm -rf ~/.cache/pip /home/mitmproxy/${WHEEL_BASENAME_MITMPROXY}
+RUN python3 -m ensurepip --upgrade \
```
mhils, line 9: We used to upgrade pip here because old pip versions were annoying to deal with, but everything from the last few years is good enough really. I think we can skip updating pip here, especially considering that the [examples on DockerHub](https://hub.docker.com/_/python) don't do that either. :-)

mhils (commented):
I was about to comment on the mkdir/chown move, but I see you just fixed that. We can skip the pip upgrade, but other than that this looks fabulous. Thanks! 🍰

release/docker/Dockerfile
PeterDaveHello: Cool, removed!

PeterDaveHello (commented):
Fixed 🎉

release/docker/Dockerfile
```diff
@@ -1,36 +1,16 @@
...
-    && LDFLAGS=-L/lib pip3 install -U /home/mitmproxy/${WHEEL_BASENAME_MITMPROXY} \
-    && apk del --purge \
-        git \
-        g++ \
-        libffi-dev \
-        openssl-dev \
-        python3-dev \
-    && rm -rf ~/.cache/pip /home/mitmproxy/${WHEEL_BASENAME_MITMPROXY}
+RUN pip3 install --no-cache-dir -U /home/mitmproxy/${WHEEL_BASENAME_MITMPROXY} \
+    && rm -rf /home/mitmproxy/${WHEEL_BASENAME_MITMPROXY}
```
PeterDaveHello, line 13: @mhils maybe `rm -rf /home/mitmproxy/${WHEEL_BASENAME_MITMPROXY}` can also be removed as it won't save any disk space, because it's not added in this image layer, unless we dynamically download add add it, what do you think?

mhils: I think we really want to use the local wheel, so downloading is bad. I'm indifferent regarding this line. Space-wise it doesn't matter, we either save a) one line in the Dockerfile or b) one file in the container filesystem. Either option is good with me, whatever you prefer. I think you are way more familiar with Docker conventions. :)

mhils:
And shipped. Thanks again, this greatly simplifies our entire Docker setup! 🍰

PeterDaveHello:
@mhils do you think the mkdir/chown thing can just be done in the Dockerfile? I don't think it should be done everytime mitmproxy is called.

mhils:
@PeterDaveHello: I vaguely recall we had multiple users complaining about poor usability (https://github.com/mitmproxy/mitmproxy/issues?q=docker+permission) with this. I'd basically like
```
docker run --rm -it -v ~/.mitmproxy:/home/mitmproxy/.mitmproxy mitmproxy:dev
```
to work on a clean system. If I remember correctly, that just fails when we have the chmod/chown stuff in the Dockerfile. Last time I checked there was no viable workaround, I only found postgres [doing something similar](https://github.com/docker-library/postgres/blob/7bd41786539082857396f4d1b4f1cb326ebee8de/13/docker-entrypoint.sh#L35-L61).

PeterDaveHello:
Oh okay, cool, so that's because it's mounted, I got it.

PR #4495, opened 2021-03-11
Add `.dockerignore` to speed up Docker build image process

PeterDaveHello:
#### Description

This will boost the process of sending build context.

#### Checklist

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+*
+!release/docker/docker-entrypoint.sh
+!**/*.whl


mhils:
👍  for the speedup, 👎  for the extra file in the root directory. Can we instead copy the wheel into `release/docker/` [here](https://github.com/mitmproxy/mitmproxy/blob/27ba85bc30ab60b9db6a885743215df839253c62/release/cibuild.py#L277), build in that directory, and then also remove one of the build args (because the location is fixed)?

PeterDaveHello:
Well... It'll work, but lose the flexibility to build from the root of the project with the same speed, also I'm not a fan of doing too many things in another script, at least I don't know that many project using python to control and build it, doing too many things in it will make it harder for a contributor to contribute.

Anyway I think this file is harmless, instead, it's a very common implementation.

mhils:
#4498 is the cleaner solution I think, but thank you very much for the pointer! 😃

PR #4497, opened 2021-03-11
Improve Shell Script syntax

PeterDaveHello:
#### Description

- Double quote variable to prevent globbing and word splitting
- Use `./*.ksy` instead of `*.ksy` so names with dashes won't become options
- Use `-n` instead of `! -z` double negative condition for better readability and understanding

I believe there isn't a problem here right now as they're already working for a long time, but a better styling can help prevent problem and help improve the maintainability of the scripts ;)

#### Checklist


mhils:
Thanks! :)

PR #4498, opened 2021-03-11
Docker ctx

mhils:
@PeterDaveHello: Alternative to #4495 - what do you think? 😃

### DIFF
--- a/release/cibuild.py
+++ b/release/cibuild.py
@@ -272,18 +272,20 @@ def build_wheel(be: BuildEnviron) -> None:  # pragma: no cover
 
 
 def build_docker_image(be: BuildEnviron) -> None:  # pragma: no cover
-    whl, = be.dist_dir.glob('mitmproxy-*-py3-none-any.whl')
-    whl = whl.relative_to(Path(".").absolute())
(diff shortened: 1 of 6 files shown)


release/cibuild.py
```diff
@@ -272,18 +272,20 @@ def build_wheel(be: BuildEnviron) -> None:  # pragma: no cover
 
 
 def build_docker_image(be: BuildEnviron) -> None:  # pragma: no cover
-    whl, = be.dist_dir.glob('mitmproxy-*-py3-none-any.whl')
-    whl = whl.relative_to(Path(".").absolute())
     click.echo("Building Docker images...")
+
+    whl, = be.dist_dir.glob('mitmproxy-*-py3-none-any.whl')
```
PeterDaveHello, line 277: This line looks like changed by accident?

release/docker/docker-entrypoint.sh
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
-# WARNING: do not change the shebang - the Docker base image might not have what you want!
```
PeterDaveHello, line 2: also not related here

PR #4502, opened 2021-03-12
Add WebSocket/TCP Message Injection

mhils:
<s>I still need to fix test coverage for proxyserver.py,</s> but I figured this can already take some review. 😃 

This PR introduces the `inject` command, which can be used to inject messages into running TCP and WebSocket connections. The implementation here is _very_ different from the previous `flow.inject_message` API as we want to keep our sans-io proxy sans-io. 😄 In a nutshell, injected messages are modeled as an external Event, similar to DataReceived/ConnectionClosed. They get passed into the proxy core from the outside (`.server_event`), passed through to the target layer, and then processed there. Most notably, injected messages now also trigger the associated `*_message` hooks.

### DIFF
--- a/examples/addons/websocket-inject-message.py
+++ b/examples/addons/websocket-inject-message.py
@@ -1,0 +1,31 @@
+"""
+Inject a WebSocket message into a running connection.
+
+This example shows how to inject a WebSocket message into a running connection.
+"""
+import asyncio
+
+from mitmproxy import ctx, http
+
+
+# Simple example: Inject a message as a response to an event
+
+def websocket_message(flow):
+    last_message = flow.websocket.messages[-1]
+    if b"secret" in last_message.content:
+        last_message.kill()
+        ctx.master.commands.call("inject", [flow], not last_message.from_client, "ssssssh")
+
+
+# Complex example: Schedule a periodic timer
+
+async def inject_async(flow: http.HTTPFlow):
+    msg = "hello from mitmproxy! "
+    while flow.server_conn.connected:
+        ctx.master.commands.call("inject", [flow], False, msg)
+        await asyncio.sleep(1)
+        msg = msg[1:] + msg[:1]
+
+
+def websocket_start(flow: http.HTTPFlow):
+    asyncio.create_task(inject_async(flow))
--- a/examples/addons/websocket-inject-message.py
+++ b/examples/addons/websocket-inject-message.py
@@ -14,15 +14,16 @@ def websocket_message(flow):
     last_message = flow.websocket.messages[-1]
     if b"secret" in last_message.content:
         last_message.kill()
-        ctx.master.commands.call("inject", [flow], not last_message.from_client, "ssssssh")
+        ctx.master.commands.call("inject.websocket", flow, last_message.from_client, "ssssssh")
 
 
 # Complex example: Schedule a periodic timer
 
 async def inject_async(flow: http.HTTPFlow):
     msg = "hello from mitmproxy! "
+    assert flow.websocket  # make type checker happy
     while flow.websocket.timestamp_end is None:
-        ctx.master.commands.call("inject", [flow], False, msg)
+        ctx.master.commands.call("inject.websocket", flow, True, msg)
         await asyncio.sleep(1)
         msg = msg[1:] + msg[:1]
 
--- a/examples/addons/websocket-inject-message.py
+++ b/examples/addons/websocket-inject-message.py
@@ -21,7 +21,7 @@ def websocket_message(flow):
 
 async def inject_async(flow: http.HTTPFlow):
     msg = "hello from mitmproxy! "
-    while flow.server_conn.connected:
+    while flow.websocket.timestamp_end is None:
         ctx.master.commands.call("inject", [flow], False, msg)
         await asyncio.sleep(1)
         msg = msg[1:] + msg[:1]
--- a/mitmproxy/addons/proxyserver.py
+++ b/mitmproxy/addons/proxyserver.py
@@ -6,8 +6,9 @@ from mitmproxy import command, controller, ctx, flow, http, log, master, options
 from mitmproxy.flow import Error
 from mitmproxy.proxy import commands, events
 from mitmproxy.proxy import server
+from mitmproxy.proxy.layers.tcp import TcpMessageInjected
 from mitmproxy.proxy.layers.websocket import WebSocketMessageInjected
-from mitmproxy.utils import asyncio_utils, human
+from mitmproxy.utils import asyncio_utils, human, strutils
 from wsproto.frame_protocol import Opcode
 
 
@@ -151,24 +152,23 @@ class Proxyserver:
             raise ValueError("Flow is not from a live connection.")
         self._connections[flow.client_conn.peername].server_event(event)
 
-    @command.command("inject.text")
+    @command.command("inject")
     def inject(self, flows: Sequence[flow.Flow], from_client: bool, message: str):
+        message_bytes = strutils.escaped_str_to_bytes(message)
+        event: events.MessageInjected
         for f in flows:
             if isinstance(f, http.HTTPFlow):
                 if f.websocket:
-                    event = WebSocketMessageInjected(
-                        f,
-                        websocket.WebSocketMessage(
-                            Opcode.TEXT, from_client, message.encode()
-                        )
-                    )
+                    msg = websocket.WebSocketMessage(Opcode.TEXT, from_client, message_bytes)
+                    event = WebSocketMessageInjected(f, msg)
                 else:
                     ctx.log.warn("Cannot inject messages into HTTP connections.")
                     continue
             elif isinstance(f, tcp.TCPFlow):
-                raise NotImplementedError
+                event = TcpMessageInjected(f, tcp.TCPMessage(from_client, message_bytes))
             else:
-                raise NotImplementedError
+                ctx.log.warn(f"Cannot inject message into {f.__class__.__name__}, skipping.")
+
             try:
                 self.inject_event(f, event)
             except ValueError as e:
--- a/mitmproxy/addons/proxyserver.py
+++ b/mitmproxy/addons/proxyserver.py
@@ -166,7 +166,7 @@ class Proxyserver:
                     continue
             elif isinstance(f, tcp.TCPFlow):
                 event = TcpMessageInjected(f, tcp.TCPMessage(from_client, message_bytes))
-            else:
+            else:  # pragma: no cover
                 ctx.log.warn(f"Cannot inject message into {f.__class__.__name__}, skipping.")
 
             try:
--- a/mitmproxy/addons/proxyserver.py
+++ b/mitmproxy/addons/proxyserver.py
@@ -168,6 +168,7 @@ class Proxyserver:
                 event = TcpMessageInjected(f, tcp.TCPMessage(from_client, message_bytes))
             else:  # pragma: no cover
                 ctx.log.warn(f"Cannot inject message into {f.__class__.__name__}, skipping.")
+                continue
 
             try:
                 self.inject_event(f, event)
--- a/mitmproxy/addons/proxyserver.py
+++ b/mitmproxy/addons/proxyserver.py
@@ -1,12 +1,14 @@
 import asyncio
 import warnings
-from typing import Optional
+from typing import Dict, Optional, Sequence, Tuple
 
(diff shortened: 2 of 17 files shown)


Prinzhorn:
Nice :heart_eyes: 

I haven't looked into the PR itself but I saw flashing warning signs when I read

> Most notably, injected messages now also trigger the associated `*_message` hooks.

Can we get a read only `injected` property on `WebSocketMessage` to prevent endless (async) recursions/loops? E.g. if in your example addon if you inject something can contains `secret` as well.

1. `mitmproxy --scripts inject.py`

```py
from mitmproxy import ctx, http

def websocket_message(flow):
    last_message = flow.websocket.messages[-1]
    if b"secret" in last_message.content:
        last_message.kill()
        ctx.master.commands.call("inject", [flow], not last_message.from_client, "super secret")
```

2. `wscat --proxy http://localhost:8080 --ca ~/.mitmproxy/mitmproxy-ca-cert.pem --connect wss://echo.websocket.org`
3. Send a single "secret" using wscat
4. Endless loop:

Selection_830

5. Even better: killing wscat does an oopsie

```
error: 127.0.0.1:37422: mitmproxy has crashed!
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/server.py", line 276, in server_event
    for command in layer_commands:
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layer.py", line 129, in handle_event
    command = command_generator.send(send)
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layer.py", line 223, in handle_event
    yield from self._handle(event)
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layer.py", line 129, in handle_event
    command = command_generator.send(send)
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layers/http/__init__.py", line 589, in _handle_event
    yield from self.event_to_child(stream, event)
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layers/http/__init__.py", line 625, in event_to_child
    for command in child.handle_event(event):
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layer.py", line 129, in handle_event
    command = command_generator.send(send)
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layers/http/__init__.py", line 507, in passthrough
    for command in self.child_layer.handle_event(event):
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layer.py", line 129, in handle_event
    command = command_generator.send(send)
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/tunnel.py", line 85, in _handle_event
    yield from self.event_to_child(event)
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/tunnel.py", line 104, in event_to_child
    for command in self.child_layer.handle_event(event):
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layer.py", line 129, in handle_event
    command = command_generator.send(send)
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/tunnel.py", line 85, in _handle_event
    yield from self.event_to_child(event)
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/tunnel.py", line 104, in event_to_child
    for command in self.child_layer.handle_event(event):
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layer.py", line 129, in handle_event
    command = command_generator.send(send)
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layers/http/__init__.py", line 589, in _handle_event
    yield from self.event_to_child(stream, event)
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layers/http/__init__.py", line 625, in event_to_child
    for command in child.handle_event(event):
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layer.py", line 129, in handle_event
    command = command_generator.send(send)
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layers/http/__init__.py", line 507, in passthrough
    for command in self.child_layer.handle_event(event):
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layer.py", line 117, in handle_event
    yield from self.__continue(event)
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/layer.py", line 194, in __continue
    command_generator = self._handle_event(ev)
  File "/home/alex/src/forks/mitmproxy/mitmproxy/proxy/utils.py", line 23, in _check_event_type
    raise AssertionError(
AssertionError: Unexpected event type at WebsocketLayer.done: Expected DataReceived|ConnectionClosed, got WebSocketMessageInjected(flow=<HTTPFlow?
  request = Request(GET echo.websocket.org:443/)?
  response = Response(101, no content)?
  websocket = <WebSocketData (884 messages)>?
  client_conn = Client(127.0.0.1:37422, state=closed, tls)?
  server_conn = Server(echo.websocket.org:443, state=closed, tls, src_port=37204)>, message='super secret').

```

Prinzhorn:
Honestly triggering `*_message` is somewhat odd and an `injected` flag only kind of solves it. E.g. for injected messages `from_client` needs be neither `True` nor `False`. Because technically it didn't originate from either server or client. But I guess if `injected == True` you can infer that `from_client` just gives you the direction (docs :wink: )

examples/addons/websocket-inject-message.py
```diff
@@ -0,0 +1,31 @@
...
+from mitmproxy import ctx, http
+
+
+# Simple example: Inject a message as a response to an event
+
+def websocket_message(flow):
+    last_message = flow.websocket.messages[-1]
+    if b"secret" in last_message.content:
+        last_message.kill()
+        ctx.master.commands.call("inject", [flow], not last_message.from_client, "ssssssh")
```
Prinzhorn, line 17: Can we hide this API and expose it as `flow.websocket.inject`? This would be much nicer for add-on authors.

mhils:
> Even better: killing wscat does an oopsie

Good catch, you found the race condition. I shall fix.

> Can we get a read only injected property on WebSocketMessage to prevent endless (async) recursions/loops?

Yes, I was thinking about this as well. We're actually getting quite a zoo of those properties: killed (soon-to-be dropped), injected, client-replayed, server-replayed. And of course all that for HTTP requests and responses, TCP messages, and WebSocket messages. I wonder if we can unify that somehow and if that would help. Does anyone have suggestions?

examples/addons/websocket-inject-message.py
mhils: I see where you are coming from. `flow.websocket.inject` unfortunately is annoying for practical reasons. We need a reference to `flow`, which would introduce an unnecessary reference cycle between `HTTPFlow` and `WebSocketData`. We also need a reference to `master`, which we could get from the magic `ctx`, but again we are introducing a bit of spaghetti here.

I'd definitely be open to improve the API to call other commands. This really should be a first-class use case.

mhils:
Catching up on your edits:

> Edit: also looking at the inject arguments it feels weird to pass in from_client. Because at this point you are sending/injecting the message to the other end. Maybe we can have a to_client (virtual) property that is just !from_client? This is confusing 😅

Yeah, you have a point here. Let's make the argument to_client instead.

> Edit2: can you kill injected message inside the event? This is confusing. Did I say that already?

Of course you can, it's treated just like any other message. But I agree that you have a confusing situation going on if you plan on killing injected messages. 😅

From a practical perspective we need to process them like regular messages, this makes sure that the UI is updated properly.

> Edit3: we definitely need an injected flag or else if the UI gets support for intercepting websocket message the user wouldn't know if it originate from client/server or an add-on.

Yes, same for replay and kill, see my previous comment right above.

mitmproxy/addons/proxyserver.py
```diff
@@ -121,15 +129,48 @@ async def shutdown_server(self):
...
+        finally:
+            del self._connections[peername]
+
+    def inject_event(self, flow: flow.Flow, event: events.Event):
+        if flow.client_conn.peername not in self._connections:
+            raise ValueError("Flow is not from a live connection.")
+        self._connections[flow.client_conn.peername].server_event(event)
+
+    @command.command("inject")
+    def inject(self, flows: Sequence[flow.Flow], from_client: bool, message: str):
```
Prinzhorn, line 156: What is the intention of having `Sequence[flow.Flow]`? Is there a remotely realistic scenario where someone injects the exact same thing into multiple flows? That would require keeping references for more than one flow, when usually you process them individually in add-ons. I think this makes the call to `inject` awkward.

mitmproxy/addons/proxyserver.py
```diff
@@ -121,15 +129,48 @@ async def shutdown_server(self):
...
+            await handler.handle_client()
+        finally:
+            del self._connections[peername]
+
+    def inject_event(self, flow: flow.Flow, event: events.Event):
+        if flow.client_conn.peername not in self._connections:
+            raise ValueError("Flow is not from a live connection.")
+        self._connections[flow.client_conn.peername].server_event(event)
+
+    @command.command("inject")
```
Prinzhorn, line 155: How do you feel about having separate `inject_websocket` and `inject_tcp`? This makes it very explicit on the caller side and decreases the surface for errors.

mitmproxy/addons/proxyserver.py
```diff
@@ -121,15 +129,48 @@ async def shutdown_server(self):
...
+    def inject(self, flows: Sequence[flow.Flow], from_client: bool, message: str):
+        message_bytes = strutils.escaped_str_to_bytes(message)
+        event: events.MessageInjected
+        for f in flows:
+            if isinstance(f, http.HTTPFlow):
+                if f.websocket:
+                    msg = websocket.WebSocketMessage(Opcode.TEXT, from_client, message_bytes)
```
Prinzhorn, line 162: `Opcode.TEXT`: for a PoC this worked great. See my comment above about separating ws and tcp injection. Then we can have two separate signatures and go back to injecting bytes into ws and not strings.

Prinzhorn:
> We're actually getting quite a zoo of those properties: killed (soon-to-be dropped), injected, client-replayed, server-replayed. And of course all that for HTTP requests and responses, TCP messages, and WebSocket messages. I wonder if we can unify that somehow and if that would help. Does anyone have suggestions?

I don't see a problem with all these flags, it's just what it is. One alternative would be sth. like `SyntheticWebSocketMessage` etc. classes. Or maybe this train of just just lead to a somewhat better solution: a `synthetic` flag and then a source (enum?) e.g. "injection" or "replay".

examples/addons/websocket-inject-message.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+# Complex example: Schedule a periodic timer
+
+async def inject_async(flow: http.HTTPFlow):
+    msg = "hello from mitmproxy! "
+    while flow.server_conn.connected:
```
Prinzhorn, line 24: I know this is just an example add-on, but does this work? I think it does? My thoughts go in a direction of `keep-alive` etc. But WebSockets are always HTTP 1 and I assume once it's upgraded the connection will only be used for the WebSocket. But I think it gets recycled if the WebSocket is closed and another connection to the same host is needed?

`WebSocketData` has three properties that concern `close` but isn't it lacking the actual information if the WebSocket is still open?

mitmproxy/proxy/layer.py
```diff
@@ -111,7 +121,34 @@ def handle_event(self, event: events.Event) -> CommandGenerator[None]:
...
+                command = command_generator.send(send)
+            except StopIteration:
+                return
+
+            while True:
+                if self.debug is not None:
+                    if not isinstance(command, commands.Log):
+                        yield self.__debug(f"<< {command}")
+                if command.blocking is True:
+                    command.blocking = self  # assign to our layer so that higher layers don't block.
```
Prinzhorn, line 138: This just threw me off. Could you quickly explain why this is `blocking: Union[bool, "mitmproxy.proxy.layer.Layer"]`? I've seen at least one occurrence of a truthy check `if command.blocking:` but no place where `blocking` is actually used as `Layer`?

mhils: @sanlengjingvv also mentioned that, I expanded the documentation a bit. :)

examples/addons/websocket-inject-message.py
mhils: There is no recycling, but your overall point is good. Next to the close properties we have, I'll add a `websocket.timestamp_end`.

PR #4520, opened 2021-03-22
Set jest time zone, otherwise the test will fail in other time zones

sanlengjingvv:
#### Checklist


mhils:
Thanks! 🍰

PR #4535, opened 2021-03-30
master branch -> main branch

mhils:
Changing branch names doesn't fix any problems, but it serves as a reminder about implicit biases we have that contribute to the lack of diversity in tech. You may say that "master" is harmless, yet this change brings repeated awareness to an important topic. Sorry if we temporarily break your workflow. :)


abitrolly:
Yes. It looks most unfortunate that US schools still teach people this meaning of `master` to have them offended.

PR #4537, opened 2021-04-01
docs(concepts-modes): add explanation about socks5

aesyondu:
#### Description

Add link to wikipedia page and brief explanation about SOCKS5.

While reading the mitm proxy documentation, I got curious by the mention of SOCKS5 proxy. I only briefly touched upon it when using ssh tunnel to forward all ports. I find it hard to understand the various explanations about what SOCKS or SOCKS5 proxy is.

#### Checklist

### DIFF
--- a/docs/src/content/concepts-modes.md
+++ b/docs/src/content/concepts-modes.md
@@ -209,4 +209,8 @@ tests).
 
 ## SOCKS Proxy
 
-In this mode, mitmproxy acts as a SOCKS5 proxy.
+In this mode, mitmproxy acts as a [SOCKS5 proxy](https://en.wikipedia.org/wiki/SOCKS).[^1]
+
+### Footnotes
+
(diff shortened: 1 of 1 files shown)


mhils:
Thanks! I appreciate the intent to improve our docs here, I agree they could benefit from some expansion.

That being said, no details is better than misleading details. SOCKS brings no security benefits (it's plaintext), in fact HTTPS ["Secure Web Proxies"](https://www.chromium.org/developers/design-documents/secure-web-proxy) are much better (and will be supported with mitmproxy 7). The page you linked is misleading/inaccurate in a number of places, so I don't feel comfortable linking to it. Wikipedia's comparison is much better, but still confusing.

TL;DR: We should expand here, but I don't feel comfortable to just add a link to a questionable source.

aesyondu:
Got it, thanks for the response!

PR #4539, opened 2021-04-02
Explain how to run tests in examples/

abitrolly:
Too minor the for CHANGELOG.

#### Checklist

### DIFF
--- a/examples/contrib/README.md
+++ b/examples/contrib/README.md
@@ -1,4 +1,12 @@
 # Community-Contributed Examples
 
 Examples in this directory are contributed by the mitmproxy community.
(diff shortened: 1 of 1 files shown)


mhils:
How much work would it be to remove the dependency on `conftest.py`? Ideally the tests here should work without any other dependencies. The only reason they depend on conftest at the moment is that they used to live in test/. My guess would be that they depend on tdata only, which can be easily copied into the test_ files.

abitrolly:
I have no idea how tests work. I mean they fail, because nobody haven't run them for a long time (I feel somewhat strange about my English here).

mhils:
I don't have any idea about how the tests in examples/contrib work either, but running them should not require copying a file first. Feel free to open a new PR if you have concrete fixes. :)

PR #4543, opened 2021-04-04
docker: avoid additional apt packages by `--no-install-recommends`

PeterDaveHello:
#### Description

This will help we keep the image not too heavy that caused by those
additional, suggestion apt packages.

Only install the packages we explicitly want.

#### Checklist


PeterDaveHello:
Interesting... in this case, `gosu` doesn't have any suggested packages, which shouldn't make any difference, I'll take a look...

mhils:
The Coverage CI is just flaky, this seems good. Thanks!

PeterDaveHello:
Looks so, thanks!

PR #4557, opened 2021-04-08
Fix parsing of certificate issuer/subject with escaped special chars

### LINKED ISSUE #4555: v7: Cert issuer/subject values contain escapes sequences
#### Problem Description

Escape sequence is not the proper term I guess. I've noticed that commas are prefixed with a backslash.

```py
def server_disconnected(data):
  server = data.server
  cert = (server.certificate_list[0] if server.certificate_list else None)

  if cert:
    print(dict(cert.issuer).get('O'))
```

`mitmdump --quiet --scripts cert.py`

`curl --proxy "localhost:8080" https://www.github.com`

```
DigiCert\, Inc.
```

#### System Information

```
Mitmproxy: 7.0.0.dev (+421, commit 72b00b5)
Python:    3.8.5
Platform:  Linux-5.8.0-48-generic-x86_64-with-glibc2.29
```

Prinzhorn:
(accidentally on the wrong branch, but the keyinfo typing is already merged)

#### Description

Fixes #4555 

This comes from the serialization to `rfc4514_string` which we didn't undo https://github.com/pyca/cryptography/blob/f0b2269f7221f06aacf883b58577ffc9627e7522/src/cryptography/x509/name.py#L52-L73

`cryptography.x509.NameAttribute` has `value`, which contains the correct thing. But it only has `oid` which is `cryptography.x509.ObjectIdentifier` which does not give us `CN` etc. only `<ObjectIdentifier(oid=2.5.4.3, name=commonName)>`.

cryptography does not expose a nicer way

https://github.com/pyca/cryptography/blob/f0b2269f7221f06aacf883b58577ffc9627e7522/src/cryptography/x509/name.py#L39-L49

Is importing `_NAMEOID_TO_NAME` bad?

Also

> Technically, a Name is a list of sets of attributes, called Relative Distinguished Names or RDNs, although multi-valued RDNs are rarely encountered.

Does that mean `List[Tuple[str, str]]` does not cut it 100% of the time?

#### Checklist


Prinzhorn:
I'll have to add tests, will do later.

mhils:
Thanks, excellent! 🍰 

Let's see what https://github.com/pyca/cryptography/pull/5959 yields, I'm always happy to merge this with `.rfc4514_string().partition("=")[0]`. The cryptography folks rightfully like to break their private APIs, so we should avoid depending on that. :)

Prinzhorn:
> Does that mean `List[Tuple[str, str]]` does not cut it 100% of the time?

So I'm still not sure how that works:

https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-adts/3c96b56d-d7a7-46f1-9883-7d031f9fa01e
https://github.com/pyca/cryptography/issues/3199

We definitely want a test with a certificate that does such a thing. `List[Tuple[str, str]]` should be safe, but not what I'm doing in my addon (`dict(cert.issuer)` could lose information)

mhils:
https://github.com/pyca/cryptography/pull/5969 is in, we can update here. 😃 

I'd propose something like
```python
try:
    k = attr.rfc4514_attribute_name 
except AttributeError:
    k = attr.rfc4514_string().partition("=")[0]
```
    

Other than that the current approach seems fine.

> Does that mean List[Tuple[str, str]] does not cut it 100% of the time?

Technically we have List[List[Tuple[str,str]]], practically speaking the inner list is almost always of length 1. Long story short, iterating over name iterates over all entries, so we should be good. :)

Prinzhorn:
> I'd propose something like
> 
> ```python
> try:
>     k = attr.rfc4514_attribute_name 
> except AttributeError:
>     k = attr.rfc4514_string().partition("=")[0]
> ```

When would `attr.rfc4514_attribute_name ` raise? Or is this for backwards compat? I thought we ship with all dependencies, or is their a way of distribution that works differently?

> Technically we have List[List[Tuple[str,str]]], practically speaking the inner list is almost always of length 1. Long story short, iterating over name iterates over all entries, so we should be good. :)

I'll definitely add a test with a cert that has multiple because I still don't understand what this will actually result in. And I love my edge cases almost as much as race conditions :shrug:

Prinzhorn:
> I'll definitely add a test with a cert that has multiple because I still don't understand what this will actually result in. And I love my edge cases almost as much as race conditions

Alright, I don't know how to integrate that into the tests, but using this from the cryptography tests

```py
subject = x509.Name([
      x509.RelativeDistinguishedName([
          x509.NameAttribute(NameOID.TITLE, u'Test'),
          x509.NameAttribute(NameOID.COMMON_NAME, u'Multivalue'),
          x509.NameAttribute(NameOID.SURNAME, u'RDNs'),
      ]),
      x509.RelativeDistinguishedName([
          x509.NameAttribute(NameOID.ORGANIZATION_NAME, u'PyCA')
      ]),
  ])
```

and passing it to `_name_to_keyval` 

yields this with the current PR changes

```
[('2.5.4.12', 'Test'), ('CN', 'Multivalue'), ('2.5.4.4', 'RDNs'), ('O', 'PyCA')]
```

and this with the old implementation

```
[('2.5.4.12', 'Test+CN=Multivalue+2.5.4.4=RDNs'), ('O', 'PyCA')]
```

so the information about the multi-value is lost, but to me this looks better?

However if there are actual duplicates (organization)

```py
subject = x509.Name([
      x509.RelativeDistinguishedName([
          x509.NameAttribute(NameOID.TITLE, u'Test'),
          x509.NameAttribute(NameOID.COMMON_NAME, u'Multivalue'),
          x509.NameAttribute(NameOID.SURNAME, u'RDNs'),
          x509.NameAttribute(NameOID.ORGANIZATION_NAME, u'TSLA'),
      ]),
      x509.RelativeDistinguishedName([
          x509.NameAttribute(NameOID.ORGANIZATION_NAME, u'PyCA')
      ]),
  ])
```

it's

```
[('2.5.4.12', 'Test'), ('CN', 'Multivalue'), ('2.5.4.4', 'RDNs'), ('O', 'TSLA'), ('O', 'PyCA')]

vs

[('2.5.4.12', 'Test+CN=Multivalue+2.5.4.4=RDNs+O=TSLA'), ('O', 'PyCA')]
```

which confirms what I thought that passing subject/issuer to `dict()` is not safe, one of the `O` is lost. But that's my problem, not that of mitmproxy. And I can live with that. I don't even know of an actual real-world certificate can contain multi-valued entries, I assume this is just something that the spec allows?

mhils:
This PR somehow slipped through, sorry!

> When would attr.rfc4514_attribute_name raise? Or is this for backwards compat? I thought we ship with all dependencies, or is there a way of distribution that works differently?

Backwards compatibility only. Specifically I'd like to not bump our dependency beyond >=3.3 for a few more months. 3.4 has awesome Rust capabilities, but let's give distros some time to adjust.

> I don't even know if an actual real-world certificate can contain multi-valued entries, I assume this is just something that the spec allows?

My understanding is that in practice noone uses multi-valued entries, but I don't know. Honestly I think we're good enough with the edge cases here. What you've shown in https://github.com/mitmproxy/mitmproxy/pull/4557#issuecomment-820999595 all looks good! 😃

Prinzhorn:
>  What you've shown in [#4557 (comment)](https://github.com/mitmproxy/mitmproxy/pull/4557#issuecomment-820999595) all looks good! smiley

I agree, I just need to make a test to make sure it keeps looking good

Prinzhorn:
Not sure if it's ok to test `_name_to_keyval` but now I feel comfortable shipping this

Prinzhorn:
> mitmproxy/certs.py:151: error: "NameAttribute" has no attribute "rfc4514_attribute_name"

How do I make mypy happy here?

mhils:
> > mitmproxy/certs.py:151: error: "NameAttribute" has no attribute "rfc4514_attribute_name"
> 
> How do I make mypy happy here?

Add a  `  # type: ignore` at the end of the line. :)

And please specify which cryptography version we're aiming to keep backward compatibility with. That makes it easier to remove the statement in a couple of years. :)

Prinzhorn:
Done

In case you get as confused as I was, the next cryptography release will be 35.0.0, not 3.5.0 https://github.com/pyca/cryptography/pull/5825

mhils:
Your commit messages make me want to non-squash this. 😂   Thanks again! 🍰

PR #4560, opened 2021-04-11
docs: add more precise instructions on docs setup

jannst:
#### Description

After only installing the "normal" version of hugo the ./build.py script failed because of some SCSS errors. After installing the extended version using snap, everything worked fine.
I also didn't know that the venv had to be activated manually, so I put a small comment, pointing that out.

#### Checklist


mhils:
Fantastic, thanks! 🍰

PR #4562, opened 2021-04-12
Draft: Update ws/tcp injection and add ws.text property

### LINKED ISSUE #4558: V7: WebSocket injection broken
#### Problem Description

WebSockets messages are either binary or utf-8. That's it. Currently the injected string is passed into `strutils.escaped_str_to_bytes` which breaks it (I have a real world WebSocket message that contains backslashes, but not after injection which breaks the application protocol which is some weird JSON based RPC). This is mostly a mental note to me. I want to refactor the injection. I personally would like it to accept a single `websocket.WebSocketMessage` argument to make things explicit and similar to how you can build requests/responses.

I think for `inject_tcp` having a string as argument is not very useful. The meaning of the TCP bytes is entirely application dependent.

```
["{\"method\":\"updateRoomCount\",\"data\":{\"model_name\":\"mitmproxy\",\"private_room\":false}}"]
```

Prinzhorn:
#### Description

This is still broken. Just me saying that I'm working on this. Or I'm trying.

1. Fix #4558 by not using `strutils.escaped_str_to_bytes`
2. Added the `text` property to WebSocketMessage. This makes manipulating messages in `websocket_message` nicer, but creating/injecting messages still only allows `content` bytes because we wanted to avoid type problems
3. Removed Opcode and use `is_text`. @mhils originally did this in another PR but I wanted to keep the Opcodes around because we might add ping/pong later. But if we ever do so, then we definitely do not want to mix them with `websocket_message` or `WebSocketMessage` because that's even more confusing. By only having `is_text` add-on authors also don't need to import from `wsproto.frame_protocol`.

The mix of `to_client` and `from_client` bools when injecting is _really_ confusing. This will be a lot nicer with source/target when you can just `target=msg.source` to bounce something back.

Missing:

* Tests (there are currently no inject_websocket tests and I don't know how they work)
* Migrations for the `WebSocketMessageState`? At least I think I've seen something like this in past PRs

#### Checklist

### DIFF
--- a/examples/addons/websocket-inject-message.py
+++ b/examples/addons/websocket-inject-message.py
@@ -5,16 +5,22 @@ This example shows how to inject a WebSocket message into a running connection.
 """
 import asyncio
 
-from mitmproxy import ctx, http
+from mitmproxy import ctx, http, websocket
 
 
 # Simple example: Inject a message as a response to an event
 
-def websocket_message(flow):
-    last_message = flow.websocket.messages[-1]
-    if b"secret" in last_message.content:
-        last_message.kill()
-        ctx.master.commands.call("inject.websocket", flow, last_message.from_client, "ssssssh")
+def websocket_message(flow: http.HTTPFlow):
+    recent_message = flow.websocket.messages[-1]
+    if recent_message.is_text and "secret" in recent_message.text:
+        recent_message.kill()
+        ctx.master.commands.call(
+          "inject.websocket",
+          flow,
+          to_client=recent_message.from_client,
+          is_text=True,
+          content=b"ssssssh"
+        )
 
 
 # Complex example: Schedule a periodic timer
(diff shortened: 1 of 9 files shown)


Prinzhorn:
> Could not load command inject.tcp: Argument content has an unknown type (<class 'bytes'>) in <bound method Proxyserver.inject_tcp of ProxyServer(stopped, 0 active conns)>.
> Could not load command inject.websocket: Argument content has an unknown type (<class 'bytes'>) in <bound method Proxyserver.inject_websocket of ProxyServer(stopped, 0 active conns)>.

Why would `content: bytes` cause a problem here? It works in other places, doesn't it? What's special about commands?

Prinzhorn:
@mhils this is essentially blocked by the fact that `inject.websocket` is only available as a command and hence I cannot pass bytes to it. Moving forward it'd be great to have a different way of injecting into the ProxyServer and then consume that API inside the `inject.websocket` command. We could then either only allow text injection inside the UI or handle the translation inside the command. The lower level API that is currently missing needs to accept raw bytes.

Prinzhorn:
Superseded by #4650

Issue #4563, opened 2021-04-12
Memory overflow when proxying large files very fast


hazcod:
#### Problem Description
When downloading a gigabyte file (e.g. 10GB) through `mitmdump`, the memory usage suddenly increases to +1GB when I reach around 106MB downloaded of the file.

A 1GB file will complete because of my container limit of 3GB, but a 10GB will fail.

This is rather difficult to debug since I can't replicate this locally on Docker for Mac / Kubernetes on Docker for Mac, but it does happen on GKE on Google Cloud.

The command I use to replicate:
```shell
% curl -k -v -x my-gke-ip:9300 --output /dev/null https://speed.hetzner.de/1GB.bin
```

This is what happens for a 1GB test file:
```
Every 3.0s: kubectl --context gcloud -n dev-proxy top pods                                                                                                                     hyde.local: Mon Apr 12 10:38:55 2021

NAME                     CPU(cores)   MEMORY(bytes)
proxy-57b7db5746-frcvp   326m         952Mi
```

And how I see the memory usage for a 10GB file:
```shell
% watch -n3 kubectl --context gke -n proxy top pods                                                         Mon Apr 12 10:35:50 2021

NAME                     CPU(cores)   MEMORY(bytes)
proxy-57b7db5746-frcvp   1082m        1832Mi
```

And I see it's an Out-Of-Memory:
```shell
% k --context gcloud -n dev-proxy get pod -o go-template='{{range.status.containerStatuses}}{{"Container Name: "}}{{.name}}{{"\r\nLastState: "}}{{.lastState}}{{end}}' proxy-57b7db5746-frcvp
Container Name: proxy
LastState: map[terminated:map[containerID:containerd://0e2b5fcd51f42fe2e3871a335aec5b331489c777ef533454c2493126d4d1c424 exitCode:137 finishedAt:2021-04-12T08:35:51Z reason:OOMKilled startedAt:2021-04-12T08:33:52Z]]%      
```

#### System Information

mitmproxy version:
```
Mitmproxy: 7.0.0.dev
Python:    3.9.2
OpenSSL:   OpenSSL 1.1.1k  25 Mar 2021
Platform:  Linux-5.10.25-linuxkit-aarch64-with-glibc2.31
```

deployment:
```
Name:         proxy-57b7db5746-cf9h2
Namespace:    dev-proxy
Priority:     0
Node:         gk3-innovation-autopilot-nap-ybmap30b-546b1215-60jj/10.10.0.10
[...]

PR #4564, opened 2021-04-12
Fix: check pending writes before requesting more from server

hazcod:
#### Description

Fixes https://github.com/mitmproxy/mitmproxy/issues/4563

This fix ensures that first we write out all pending bytes to the client before requesting more from the server.
Potentially this impacts performance.

#### Checklist

CC @mhils

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ If you depend on these features, please raise your voice in
 
 ### Full Changelog
 
+* Check pending writes to client before requesting more from server (@hazcod)
 * New Proxy Core based on sans-io pattern (@mhils)
 * mitmproxy's command line interface now supports Windows (@mhils)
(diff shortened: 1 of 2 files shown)


hazcod:
@mhils I was thinking, would it be more sensible to do this check only for when the `Content-Length` is >100mb?
Or should we check the size of outstanding chunks to send?

mhils:
> I was thinking, would it be more sensible to do this check only for when the Content-Length is >100mb?

Completely different layer here, that's infeasible. We'd need someone to benchmark this for large requests. Small payloads we can test with [proxybench](https://github.com/mitmproxy/proxybench).

mhils:
I just did some benchmarking and this all seems to work out. 😃  Thanks again!

hazcod:
@mhils sorry for the long delay, this was on my backlog.. I've been running it and works fine tough!
Thanks for the help!

Issue #4415, opened 2021-01-27
Errors While Decoding Response Text Using mitmdump


omarmhaimdat:
#### Problem Description
When I visit websites with Arabic characters using a small addon script with mitmdump, and extract the response text I get the following error:

`Traceback (most recent call last):
  File "main.py", line 36, in response
    response_text = flow.response.text
  File "c:\users\evead-61\appdata\local\programs\python\python38\lib\site-packages\mitmproxy\net\http\message.py", line 232, in get_text
    return cast(str, encoding.decode(content, enc))
  File "c:\users\evead-61\appdata\local\programs\python\python38\lib\site-packages\mitmproxy\net\http\encoding.py", line 76, in decode
    raise ValueError("{} when decoding {} with {}: {}".format(
ValueError: UnicodeDecodeError when decoding b'GIF89a\x with 'UTF-8': UnicodeDecodeError('utf-8', b'GIF89a\x01\x00\x01\x00\xf0\x00\x00\x00\x00\x00\x00\x00\x00!\xf9\x04\x01\x00\x00\x00\x00,\x00\x00\x00\x00\x01\x00\x01\x00\x00\x02\x02D\x01\x00;`

#### Steps to reproduce the behavior:
1. Write a small addon that assigns the HTTPResponse flow response text from the "response()" method
2. Assign _flow.response.text_ to a variable
3. Run using `mitmdump -s main.py --anticomp` (assuming your file is called _main.py_)
4. You can try it on this website [chouftv.ma](https://chouftv.ma/press/341484.html)

#### System Information
Paste the output of "mitmproxy --version" here.
Mitmproxy: 6.0.2
Python:    3.8.7
OpenSSL:   OpenSSL 1.1.1i  8 Dec 2020
Platform:  Windows-10-10.0.17763-SP0

Prinzhorn:
> When I visit websites with Arabic characters using a small addon script with mitmdump

Looking at the error you seem to be accessing the `text` property of an _image_ response. ` b'GIF89a\x` are the magic bytes of a gif image. But it shouldn't crash, I need to look into why our guess_encoding thinks this is UTF-8 (should fall back to `latin-1` for binary bodies).

In the meantime, could you test with the latest master? Snapshots are here https://mitmproxy.org/downloads/#branches/master/

Prinzhorn:
I'm able to reproduce this on github.com with their tracking pixel, so it's definitely unrelated to Arabic

[...]

PR #4566, opened 2021-04-15
Fixed encoding guessing: only search for meta tags in HTML bodies

Prinzhorn:
#### Description

Real world issue: https://www.youtube.com/s/desktop/245f415e/jsbin/desktop_polymer_inlined_html_polymer_flags.vflset/desktop_polymer_inlined_html_polymer_flags.js

This JavaScript file contains a matching meta tag and the encoding ended up being a literal backslash `\` which I passed along to `codecs.lookup(enc).name`. The added check will catch html and xhtml.

I think we want to rewrite/improvie the guessing anyway with better tests, e.g. #4415

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 * Fix processing of `--set` options (#5067, @marwinxxii) 
 * Lowercase user-added header names and emit a log message to notify the user when using HTTP/2 (#4746, @mhils)
 * Exit early if there are errors on startup (#4544, @mhils)
+* Fixed encoding guessing: only search for meta tags in HTML bodies (##4566, @Prinzhorn)
 
 ## 28 September 2021: mitmproxy 7.0.4
 
--- a/mitmproxy/http.py
+++ b/mitmproxy/http.py
@@ -414,9 +414,10 @@ class Message(serializable.Serializable):
             if "json" in self.headers.get("content-type", ""):
                 enc = "utf8"
         if not enc:
-            meta_charset = re.search(rb"""<meta[^>]+charset=['"]?([^'">]+)""", content, re.IGNORECASE)
-            if meta_charset:
-                enc = meta_charset.group(1).decode("ascii", "ignore")
+            if "html" in self.headers.get("content-type", ""):
+                meta_charset = re.search(rb"""<meta[^>]+charset=['"]?([^'">]+)""", content, re.IGNORECASE)
+                if meta_charset:
+                    enc = meta_charset.group(1).decode("ascii", "ignore")
         if not enc:
             if "text/css" in self.headers.get("content-type", ""):
                 # @charset rule must be the very first thing.
--- a/test/mitmproxy/test_http.py
+++ b/test/mitmproxy/test_http.py
@@ -1098,6 +1098,7 @@ class TestMessageText:
     def test_guess_meta_charset(self):
         r = tresp(content=b'<meta http-equiv="content-type" '
                           b'content="text/html;charset=gb2312">\xe6\x98\x8e\xe4\xbc\xaf')
+        r.headers["content-type"] = "text/html"
         # "鏄庝集" is decoded form of \xe6\x98\x8e\xe4\xbc\xaf in gb18030
         assert "鏄庝集" in r.text
 


mhils:
I don't think we should rely on a content-type header being set, that's precisely the point here - the content type is specified in the HTML only. To me it looks like we should much rather improve our detection heuristic to not trigger on your file?

While we're at it, we should also reduce the search to the first 1024 bytes by using `re.match(r".{0,1024}")`... :)

Prinzhorn:
> I don't think we should rely on a content-type header being set, that's precisely the point here - the content type is specified in the HTML only

Could you point to some real world test-case here? I thought we're only looking at the meta tag because of the encoding, not because of the content-type. I don't think any browser beyond Internet Explorer does any content-type sniffing aynmore so if there is no content-type header it will never be upgraded to HTML and hence the meta tag does not have any effect. But maybe I'm missing a use-case here?

> To me it looks like we should much rather improve our detection heuristic to not trigger on your file?

I'd say it has to be `html in content_type or content_type is None`, that should catch all cases? In my case it's correctly served as JavaScript so there is absolutely no reason to sniff a different content-type (which this code doesn't, it just tries to guess the encoding).

> While we're at it, we should also reduce the search to the first 1024 bytes by using `re.match(r".{0,1024}")`... :)

Yes, especially since that YouTube JavaScript has 7.7 MB :smile:

mhils:
> Could you point to some real world test-case here? I thought we're only looking at the meta tag because of the encoding, not because of the content-type. I don't think any browser beyond Internet Explorer does any content-type sniffing aynmore so if there is no content-type header it will never be upgraded to HTML and hence the meta tag does not have any effect. So why should mitmproxy do it? But maybe I'm missing a use-case here?
> 
> >

If you have a proper content-type header, that header could/should already set the charset (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type#directives). So assuming a proper HTML content type is slightly optimistic. That being said, I have no idea what modern browsers are doing.

I know that [requests](https://pypi.org/project/requests/) has given up at some point and now exclusively relies on [chardet](https://pypi.org/project/chardet/). Not sure what's the current state, but that used to have terrible performance. Chromium apparently does something similar with https://opensource.google/projects/ced. Firefox has an even newer thing: https://hsivonen.fi/chardetng/.

> It could be html in content_type or content_type is None, that should catch all cases?

That sounds good to me! We can then also do the same "is None" for the CSS rule right below. :)

Prinzhorn:
> that's precisely the point here - the content type is specified in the HTML only

That's not necessarily true, e.g. `<meta charset="utf-8">` without `http-equiv`.

Prinzhorn:
I think one option that https://github.com/mitmproxy/mitmproxy/issues/4415#issuecomment-768345919 is missing is to `try` decoding inside `text` and if it fails transparently fall back to `latin-1`. Does that make sense? This would be useful regardless of what method we're going to use to detect the charset to make sure that `text` always works. And if the charset was correct or if you never even access `text` then there is no downside.

Prinzhorn:
One issue with the chardet approach that neither Google, nor Firefox or request have: in mitmproxy bodies are not immutable. So there might be a content-type header with a funny charset but the body does not use funny characters. Now chardet boldly claims this is latin-1 or utf-8. And now someone uses the `text` setter and we update `content` with utf-8 bytes and mitmproxy forwards the HTML to the browser. The browser sees the content-type header and might break the utf-8 characters inside the body, depending on what they actually do :sweat_smile: 
I think other projects don't suffer from that, because the bodies are immutable. If you're scraping or analyzing dumps you only care about being able to _read_ the bodies properly, not write them.

mhils:
> > that's precisely the point here - the content type is specified in the HTML only
> 
> That's not necessarily true, e.g. `<meta charset="utf-8">` without `http-equiv`.

My previous comment wasn't phrased well. What I meant is: Ideally everyone would also have proper Content-Type headers, but in practice we check for the meta tag because people can't get their headers right.

You may also "enjoy" reading https://www.w3.org/International/questions/qa-html-encoding-declarations.en.

> I think one option that [#4415 (comment)](https://github.com/mitmproxy/mitmproxy/issues/4415#issuecomment-768345919) is missing is to `try` decoding inside `text` and if it fails transparently fall back to `latin-1`. Does that make sense?

That's actually a neat idea. 👍 Latin1 is quite terrible, but in contrast to UTF-8 it's wonderfully bijective. 😁 

We do need to make sure that we also re-encode it properly in `set_text` then. For all instances where we fall back to decoding with latin1, we need to also then re-encode with latin1. That probably means refactoring _guess_encoding to not take arguments and always return an encoding that can be used to decode the current body.

> One issue with the chardet approach that neither Google, nor Firefox or request have: in mitmproxy bodies are not immutable. So there might be a content-type header with a funny charset but the body does not use funny characters. Now chardet boldly claims this is latin-1 or utf-8. And now someone uses the `text` setter and we update `content` with utf-8 bytes and mitmproxy forwards the HTML to the browser. The browser sees the content-type header and might break the utf-8 characters inside the body, depending on what they actually do 😅

Yes, that's also a good point. I'd generally be leaning heavily towards not using chardet anyways, it's just another dependency for relatively little benefit.

Prinzhorn:
> I don't think we should rely on a content-type header being set, that's precisely the point here - the content type is specified in the HTML only

Since this just came up again, could you please double check what you said there? I think the point here is that the `content-type` header does not specify a charset. Not that the `content-type` is missing completely. Without a `content-type` header we should just straight up assume `application/octet-stream` and force `latin-1`. Anything else would be insane. Unless you have a convincing argument and a real world URL to proof it (one you didn't host yourself or a someone your know or is related to you).

Here's how I'm currently wrapping `_guess_encoding` in my addon. I need to normalize the encoding name and make sure it doesn't crash. I _never_ use `text` at all, Node.js only passes `content` back and forth, that's why I need to make sure I can properly decode it on the other end.

```py
def __guess_encoding(self, message: mitmproxy.http.Message, content: bytes) -> str:
    enc = message._guess_encoding(content)

    try:
        # Normalize the name, e.g. latin-1 becomes iso8859-1 or UTF8 becomes utf-8.
        enc = codecs.lookup(enc).name

        # If we detected a more specific encoding than iso8859-1, make sure the bytes actually decode with it.
        if enc != "iso8859-1":
            try:
                codecs.decode(content, enc)
            except ValueError:
                enc = "iso8859-1"
    except LookupError:
        # Could not look up and normalize the name, fall back to iso8859-1.
        enc = "iso8859-1"

    return enc
```

I am happy with this so far, as it is robust (e.g. the GIF with charset=UTF-8 will fall back to iso8859-1). But now we need to improve our heuristics to get some better results.

Ideas:

1. Missing `content-type` -> early exit `latin-1`
2. Content-type has charset -> done, we trust it (but make sure it actually decodes, evil UTF-8 GIFs beware)
3. Otherwise check the content-type for known types and apply heuristics, e.g. for HTML look at meta tags, for CSS look at @charset, for JSON force UTF-8, for other known TEXT (xml, svg, plain, ...) types look at BOM. 
4. If the above fails add some defaults, e.g. JavaScript as UTF-8 (but make sure it decodes and doesn't crash)

Prinzhorn:
> Without a `content-type` header we should just straight up assume `application/octet-stream` and force `latin-1`

I will soften this up a bit, as it makes sense to look for BOM. Chances that it is a text file is much higher than that your random binary blob starts with BOM.

But it's complicated.

Also see https://github.com/codeprentice-org/sniffpy

Which doesn't concern charset/encoding at all, just the mime type. Did I say it's complicated? We need a third party.

Prinzhorn:
Refs https://github.com/mitmproxy/mitmproxy/issues/5152

PR #4586, opened 2021-05-07
Metadata and replay filter syntax

rbdixon:
#### Description

Flows may be filtered based on metadata and replay status.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ If you depend on these features, please raise your voice in
 * Multiple Browsers: The `browser.start` command may be executed more than once to start additional
   browser sessions. (@rbdixon)
 * Improve readability of SHA256 fingerprint. (@wrekone)
+* Metadata and Replay Flow Filters: Flows may be filtered based on metadata and replay status. (@rbdixon)
 * --- TODO: add new PRs above this line ---
 * ... and various other fixes, documentation improvements, dependency version bumps, etc.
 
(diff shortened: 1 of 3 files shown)


mitmproxy/flowfilter.py
```diff
@@ -382,6 +381,40 @@ def __call__(self, f):
...
+        return f.is_replay == 'response'
+
+
+class FMeta(_Rex):
+    code = "meta"
+    help = "Flow metadata"
+    flags = re.MULTILINE
+
+    def __call__(self, f):
+        m = http.Headers(**{key: repr(value) for key, value in f.metadata.items()})
```
mhils, line 414: That's quite an abuse of our headers class here. Can we search on `f"{key}: {value}"` strings instead please?

mhils (commented):
Thanks! This looks good overall. My main point is that I'm not too happy with `~r`, `~rc`, and `~rs`. I feel that the slightly longer `~replay` would be more natural. We should also use `~replayq` and  `~replays`, which is more in line with the other q/s  specialization. Does that make sense? 😃

CHANGELOG.md
```diff
@@ -58,10 +58,9 @@ If you depend on these features, please raise your voice in
 * Add option `export_preserve_original_ip` to force exported command to connect to IP from original request. Only supports curl at the moment. (@dkasak)
 * Major proxy protocol testing (@r00t-)
 * Switch Docker image release to be based on Debian (@PeterDaveHello)
-* Multiple Browsers: The `browser.start` command may be executed more than once to start additional
```
mhils, line 61: Please re-add, otherwise LGTM. :-)

rbdixon: Cleaned up.

mhils:
Excellent, thanks! 🍰

PR #4587, opened 2021-05-07
Use emoji and characters as markers. Add ~marker filter syntax.

rbdixon:
#### Description

Allows a per-flow custom marker to be set. Can either be a character or emoji (`:bomb:`, for example). Adds a filter syntax for searching flows with a specific marker (`~marker`). The `flow.mark` command has an optional third argument, `marker`, which can be used to set the marker on a flow.

None of this changes the standard mark behavior. Only the presentation of the marker in the view is different. The filter `~marked` will match _all_ marked flows regardless of what marker is being used. The filter `~marked :grapes:` will only match the flows that are marked and have the `marker` attribute set to 🍇.

I had to bump the file format and add an importer for that.

#### Checklist


mhils:
Thanks for the really interesting PR! I was very skeptical initially, but I think overall is a smart addition. 😃 I'm not so sure about 🍇, but being able to distinguish between 🟠,🟣,⚫,🔵,🟤,🟡,🟢, and ⚪ seems rather useful!

<details><summary>1. I'm hitting a traceback when opening mitmproxy ->  `n` -> `enter` -> `m`.</summary>
<pre>
  File "/mnt/c/Users/user/git/mitmproxy/mitmproxy/master.py", line 54, in run_loop
    loop()
  File "/home/user/venv/lib/python3.9/site-packages/urwid/main_loop.py", line 287, in run
    self._run()
  File "/home/user/venv/lib/python3.9/site-packages/urwid/main_loop.py", line 385, in _run
    self.event_loop.run()
  File "/home/user/venv/lib/python3.9/site-packages/urwid/main_loop.py", line 1494, in run
    reraise(*exc_info)
  File "/home/user/venv/lib/python3.9/site-packages/urwid/compat.py", line 58, in reraise
    raise value
  File "/usr/lib/python3.9/asyncio/events.py", line 80, in _run
  File "/home/user/venv/lib/python3.9/site-packages/urwid/main_loop.py", line 1454, in faux_idle_callback
    callback()
  File "/home/user/venv/lib/python3.9/site-packages/urwid/main_loop.py", line 574, in entering_idle
    self.draw_screen()
  File "/home/user/venv/lib/python3.9/site-packages/urwid/main_loop.py", line 588, in draw_screen
    canvas = self._topmost_widget.render(self.screen_size, focus=True)
  File "/home/user/venv/lib/python3.9/site-packages/urwid/widget.py", line 145, in cached_render
    canv = fn(self, size, focus=focus)
  File "/home/user/venv/lib/python3.9/site-packages/urwid/container.py", line 1089, in render
    body = self.body.render((maxcol, maxrow-ftrim-htrim),
  File "/home/user/venv/lib/python3.9/site-packages/urwid/widget.py", line 145, in cached_render
    canv = fn(self, size, focus=focus)
  File "/home/user/venv/lib/python3.9/site-packages/urwid/decoration.py", line 226, in render
    canv = self._original_widget.render(size, focus=focus)
  File "/home/user/venv/lib/python3.9/site-packages/urwid/widget.py", line 145, in cached_render
    canv = fn(self, size, focus=focus)
  File "/home/user/venv/lib/python3.9/site-packages/urwid/container.py", line 1089, in render
    body = self.body.render((maxcol, maxrow-ftrim-htrim),
  File "/home/user/venv/lib/python3.9/site-packages/urwid/widget.py", line 145, in cached_render
    canv = fn(self, size, focus=focus)
  File "/home/user/venv/lib/python3.9/site-packages/urwid/listbox.py", line 439, in render
    middle, top, bottom = self.calculate_visible(
  File "/home/user/venv/lib/python3.9/site-packages/urwid/listbox.py", line 325, in calculate_visible
    focus_widget, focus_pos = self._body.get_focus()
  File "/mnt/c/Users/user/git/mitmproxy/mitmproxy/tools/console/flowlist.py", line 61, in get_focus
    f = FlowItem(self.master, self.master.view.focus.flow)
  File "/mnt/c/Users/user/git/mitmproxy/mitmproxy/tools/console/flowlist.py", line 12, in __init__
    w = self.get_text()
  File "/mnt/c/Users/user/git/mitmproxy/mitmproxy/tools/console/flowlist.py", line 23, in get_text
    return common.format_flow(
  File "/mnt/c/Users/user/git/mitmproxy/mitmproxy/tools/console/common.py", line 735, in format_flow
    return render_func(
  File "/mnt/c/Users/user/git/mitmproxy/mitmproxy/tools/console/common.py", line 596, in format_http_flow_table
    items.append(format_right_indicators(
  File "/mnt/c/Users/user/git/mitmproxy/mitmproxy/tools/console/common.py", line 389, in format_right_indicators
    indicators.append(("mark", render_marker(marker)[0]))
  File "/mnt/c/Users/user/git/mitmproxy/mitmproxy/tools/console/common.py", line 168, in render_marker
    if len(marker) == 1:
TypeError: object of type 'NoneType' has no len()
</pre>
</details>

2. I would be very inclined to not introduce additional dependencies here. Can you maybe add `mitmproxy/utils/emoji.py` with something like this?
```python
#!/usr/bin/env python3

# auto-generated. run this file to refresh.

emoji = {
# execute once to fill
}

if __name__ == "__main__":
    import requests
    import io
    import re
    from pathlib import Path

    out = io.StringIO()

    r = requests.get("https://api.github.com/emojis")
    for name, url in r.json().items():
        codepoints = url.rpartition("/")[2].partition(".png")[0].split("-")
        try:
            emoji = "\u200d".join(chr(int(c, 16)) for c in codepoints)
        except ValueError:
            continue  # some GitHub-specific emojis, e.g. "atom".
        print(f'    "{name}": "{emoji}",', file=out)

    Path(__file__).write_text(
        re.sub(r"(?<={\n)[\s\S]*(?=}\n)", lambda x: out.getvalue(), Path(__file__).read_text("utf8"), 1),
        "utf8"
    )

```
3. Is there a specific reason why you kept `marker` and `marked` separate? I'd say we just make `marked` a str, where a truthy value denotes that it is marked and an empty string denotes that it is not marked. That way they can never go out of sync. :)

rbdixon:
The new update:

- Removes the dependency on `emoji` in favor of your proposal.
- Keeps the `:emoji_name:` syntax to avoid confusion between the letter `X` and the emoji named `X`.
- Drops the `Flow().marker` attribute as suggested. `Flow().marked` may either be a bool or str.
- The `marked` attribute will render as either an emoji, single char (A-Za-z0-9) or the default "red ball" otherwise.

mitmproxy/addons/core.py
```diff
@@ -103,15 +103,15 @@ def resume(self, flows: typing.Sequence[flow.Flow]) -> None:
 
     # FIXME: this will become view.mark later
     @command.command("flow.mark")
-    def mark(self, flows: typing.Sequence[flow.Flow], boolean: bool) -> None:
+    @command.argument("marker", type=mitmproxy.types.Marker)
+    def mark(self, flows: typing.Sequence[flow.Flow], marker: typing.Union[bool, str]) -> None:
```
mhils, line 107: Is there a reason why you renamed the command? I don't particularly mind the change, but we should also strive to not break compatibility too badly.

mitmproxy/flow.py
```diff
@@ -82,9 +82,18 @@ class Flow(stateobject.StateObject):
     We're waiting for a user action to forward the flow to its destination.
     """
 
-    marked: bool
+    marked: typing.Union[bool, str]
```
mhils, line 85: Let's make this `str`. `""` denotes "no marker" and implicitly evaluates to `False`, so we can do keep saying `if flow.marker: ...`. Having this as a single type makes it much easier to work with than having to check if it's a bool or a str all the time.

mitmproxy/io/compat.py
```diff
@@ -300,6 +300,11 @@ def convert_11_12(data):
     return data
 
 
+def convert_12_13(data):
+    data["version"] = 13
+    return data
```
mhils, line 305: Here we then need to do a 
```python
if flow["marked"]:
    flow["marked"] = SYMBOL_MARK
else:
    flow["marked"] = ""
```
conversion to convert from bool to str when importing old dumpfiles.

mitmproxy/utils/emoji.py
```diff
@@ -0,0 +1,1808 @@
...
+    ":zero:": "0‍⃣",
+    ":zimbabwe:": "🇿‍🇼",
+    ":zipper_mouth_face:": "🤐",
+    ":zombie:": "🧟",
+    ":zombie_man:": "🧟‍♂",
+    ":zombie_woman:": "🧟‍♀",
+    ":zzz:": "💤",
+}
+
+if __name__ == "__main__":
```
mhils, line 1788: You can add a  `# pragma: no cover` to the end of this line so that our coverage checks are not complaining. :)

mitmproxy/addons/core.py
```diff
@@ -103,15 +103,15 @@ def resume(self, flows: typing.Sequence[flow.Flow]) -> None:
...
         """
             Mark flows.
         """
         updated = []
         for i in flows:
-            if i.marked != boolean:
-                i.marked = boolean
-                updated.append(i)
+            i.marked = marker
```
mhils, line 113: Have you considered doing the `:grape:` -> 🍇 conversion here, i.e. directly apply `render_marker` here?
We could then also fail here if someone specifies an invalid emoji, instead of just silently assuming the default one.

mhils (commented):
This looks really excellent now without the additional dependency. Really nice work. I have a few smaller comments below, thanks for bearing with me. 😃

mitmproxy/addons/core.py
rbdixon: That is a goof! Good catch. I'll fix that.

mitmproxy/flow.py
rbdixon: I've made the `marker` attribute a `str` type, only, everywhere. This propagated a bit into the test code, too.

I added a `:default:` marker that provides `SYMBOL_MARK` as a convenience for those who want the exact same marker as before.

The `Marker` type will still accept `true` and `false` for when the user issues commands or for use in keystroke bindings. This just seemed a useful accommodation.

mitmproxy/io/compat.py
rbdixon: Done.

mitmproxy/addons/core.py
rbdixon: I decided not to do this because I assumed that, perhaps, `mitmweb` would render markers differently at some point. Since `mitmweb` isn't showing markers at all (right?) I can leave this abstraction out and just handle it right here as you suggest.

rbdixon: Upon further consideration... the other difficulty of rendering the value early is filtering. It is convenient to just search for `~marker circle` and return all flows marked with a circle regardless of the color.

rbdixon: I did add a check that raises in that function if the marker is not known. I also supplemented `emoji.py` with the single char `a-zA-Z0-9` values for completeness and to simplify checks elsewhere in the code.

mhils:
Excellent! This looks good to merge except for two minor issues:

 - We have a pretty cruel CI that mandates test files for all modules: https://github.com/mitmproxy/mitmproxy/pull/4587/checks?check_run_id=2657505739#step:5:10
    Can you please just add an empty test_emoji.py? We only define a constant table, nothing to test here.
 - Is looks like dumpfile-13.mitm is a leftover that can be removed?

rbdixon:
Both fixed.

mhils:
Exellent, thanks!

It looks like our `individual_coverage` script fails on modules that have no test code. I know I said the opposite thing before, but could you add a simple dict lookup test maybe? 😅

mhils:
🚀

PR #4590, opened 2021-05-09
[WIP] Harden CI Pipeline

mhils:
This PR has some attempts to harden our CI pipeline, in particular by strictly pinning dependency versions.


.github/workflows/main.yml
```diff
@@ -1,41 +1,64 @@
...
 jobs:
   lint-pr:
     if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
-      - uses: TrueBrain/actions-flake8@v1.4.1
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
```
mhils, line 15: Without setting this, every action below has implict access to $GITHUB_TOKEN. That's not necessary.

See https://github.com/actions/checkout/issues/485

.github/workflows/main.yml
```diff
@@ -1,41 +1,64 @@
...
+      - uses: TrueBrain/actions-flake8@9a43ff1b2c7b96f3edffc48a49973ce3de116ba1
+        # mirrored at https://github.com/mitmproxy/mitmproxy/settings/actions
```
mhils, line 17: We now pin all non-GitHub actions to a specific revision, and also maintain a whitelist at https://github.com/mitmproxy/mitmproxy/settings/actions.

.github/workflows/main.yml
```diff
@@ -126,66 +114,137 @@ jobs:
...
-      - uses: codecov/codecov-action@v1
+      - uses: codecov/codecov-action@a1ed4b322b4b38cb846afb5a0ebfa17086917d27
+        # mirrored above and at https://github.com/mitmproxy/mitmproxy/settings/actions
         with:
           file: ./web/coverage/coverage-final.json
           name: web
 
   docs:
+    if: github.repository != 'mitmproxy/mitmproxy' || github.ref != 'refs/heads/main'
+    # mirrored below >>>
```
mhils, line 125: This situation here is a bit unfortunate: We only want regular commits on the main branch to have access to deployment secrets, but when we set `environment` for a job, it does not run without that environment being present at all. So we effectively duplicate this job, and have one version with the environment and one without.

Alternatively we could call actions/upload-artifacts here and then download them in a separate action. That also adds a lot of complexity (e.g. we currently grab all artifacts in the deploy step, we'd then need to manually list them), so the current way is probably more pragmatic.

.github/workflows/main.yml
```diff
@@ -126,66 +114,137 @@ jobs:
...
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
+      - run: python release/pins/pip-install-pinned.py
+      - run: |
+          wget -q https://github.com/gohugoio/hugo/releases/download/v0.83.1/hugo_extended_0.83.1_Linux-64bit.deb
+          echo "9487ea3b80f8ddd0ba600d42850b96b6a8b0bb9b41bc08cb285635ebbd41328d hugo_extended_0.83.1_Linux-64bit.deb" | sha256sum -c
```
mhils, line 137: We verify the hash here now. I've tested that it fails early if the hash is different.

.github/workflows/main.yml
```diff
@@ -126,66 +114,137 @@ jobs:
...
+          - image: windows-2019
+            platform: windows
+          - image: ubuntu-16.04  # Old Ubuntu version for old glibc
+            platform: linux
+    runs-on: ${{ matrix.image }}
+    env:
+      CI_BUILD_WHEEL: ${{ matrix.platform == 'linux' }}
+      CI_BUILD_PYINSTALLER: 1
+      CI_BUILD_WININSTALLER: ${{ matrix.platform == 'windows' }}
+      CI_BUILD_KEY: ${{ secrets.CI_BUILD_KEY }}
```
mhils, line 184: I've merged the build-binaries and build-wheels jobs, with the linux build creating the wheel now as well. That makes things easier to handle overall.

.github/workflows/main.yml
```diff
@@ -126,66 +114,137 @@ jobs:
...
       CI_BUILD_WHEEL: 1
       CI_BUILD_PYINSTALLER: 1
       CI_BUILD_WININSTALLER: 1
       TWINE_USERNAME: mitmproxy
-      TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
-      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      #TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
+      #AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      #AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
mhils, line 243: TODO: Figure out if this needs to be explicitly specified if `environment` is set.

.github/workflows/main.yml
```diff
@@ -197,5 +256,5 @@ jobs:
       # and then delete the empty folders
       - run: find release/dist -type d -empty -delete
       - run: ls release/dist
-      - run: pip install tox
-      - run: tox -e cibuild -- upload
+      - run: python release/pins/pip-install-pinned.py
+      - run: python release/cibuild.py upload
```
mhils, line 260: We do not use tox anymore for all jobs that eventually produce release assets. The reason for that is that we want to install all dependencies with their hashes pinned and then install mitmproxy as editable, and that is not (easily) possible with tox.

release/cibuild.py
```diff
@@ -382,6 +384,7 @@ def build_wininstaller(be: BuildEnviron) -> None:  # pragma: no cover
     click.echo("Building wininstaller package...")
 
     IB_VERSION = "20.12.0"
+    IB_SETUP_SHA256 = "657f4785c7d70f140468435b99e79ced813e7e051106e7525e0c819efffb40d3"
```
mhils, line 387: We now also verify the hash for InstallBuilder, which we use to build windows installers. :)

release/docker/Dockerfile
```diff
@@ -6,10 +6,12 @@ RUN useradd -mU mitmproxy
...
     && rm -rf /var/lib/apt/lists/*
-    
+
 COPY $MITMPROXY_WHEEL /home/mitmproxy/
-RUN pip3 install --no-cache-dir -U /home/mitmproxy/${MITMPROXY_WHEEL} \
-    && rm -rf /home/mitmproxy/${MITMPROXY_WHEEL}
+COPY requirements.txt /home/mitmproxy/
+RUN pip3 install --no-cache-dir --require-hashes -r /home/mitmproxy/requirements.txt \
+    && pip3 install --no-cache-dir --no-deps /home/mitmproxy/${MITMPROXY_WHEEL} \
+    && rm /home/mitmproxy/${MITMPROXY_WHEEL} /home/mitmproxy/requirements.txt
```
mhils, line 14: Our Docker images now also only install pinned dependencies.

release/pins/pip-install-pinned.py
```diff
@@ -0,0 +1,17 @@
...
+
+pip_install = [
+    sys.executable, "-m",
+    "pip",
+    "install",
+    "--disable-pip-version-check",
+]
+
+check_call([*pip_install, "--require-hashes", "-r", f"requirements-{platform.system().lower()}.txt"], cwd=here)
+check_call([*pip_install, "--no-deps", "-e", "../.."], cwd=here)
```
mhils, line 17: So... when we want to install mitmproxy with pinned dependencies, we first install all pinned dependencies, and then in a second step install mitmproxy as editable, making sure that no additional dependencies are sneaking in. 🙃

setup.py
```diff
@@ -97,16 +97,21 @@
...
             "pytest-asyncio>=0.10.0,<0.14,!=0.14",
             "pytest-cov>=2.7.1,<3",
             "pytest-timeout>=1.3.3,<2",
             "pytest-xdist>=2.1.0,<3",
             "pytest>=6.1.0,<7",
             "requests>=2.9.1,<3",
             "tox>=3.5,<4",
-        ]
+            "twine>=3.4.1,<4",
+            "wheel>=0.36.2,<0.37",
```
mhils, line 114: These dependencies were specified in tox.ini, I've just moved them all in here for simplicity. Otherwise we need to maintain different pinned requirements files etc, it's all pain.

tox.ini
```diff
@@ -36,31 +36,12 @@ commands =
...
-  mitmdump --version
-  python ./release/cibuild.py {posargs}
-
 [testenv:wheeltest]
 recreate = True
 deps =
 commands =
-  pip install {posargs}
+  pip install --require-hashes -r release/pins/requirements-linux.txt
+  pip install --no-deps {posargs}
```
mhils, line 44: Here was another pitfall: cibuild calls `tox -e wheeltest` , which if unchanged would again introduce unpinned dependencies here. 🤷

mhils (commented):
@Kriechi: This is ready for review and comments for someone with a fresh mind! In a nutshell, we're now making sure that we pin all our dependencies in the build process. It's massively painful, and I now see why noone else is doing it. Anyhow, lots of comments below.

Also, @briansmith has done some pretty interesting investigations into the default security of GitHub Actions at briansmith/untrusted#50 (thanks!). I've tried taking the main points into account here, in particular I've also switched our $GITHUB_TOKENs to be read-only.

release/pins/make.py
```diff
@@ -0,0 +1,21 @@
...
+        "pip-compile",
+        "--generate-hashes", "--upgrade", "--allow-unsafe",
+        "-o", outfile,
+        "requirements.in",
+    ], cwd=here)
+
+    out = outfile.read_bytes()
+    out = re.sub(b"-e .+", b"# --- line removed by make.py ---", out)
+    out = re.sub(b"pip-compile --.+", b"./make.py  (on the same platform)", out)
+    outfile.write_bytes(out)
```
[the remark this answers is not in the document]
Kriechi: Why do we install ourself as editable here? Shouldn't we build the wheel, then install it + deps, and then take the requirements.txt?

.github/workflows/main.yml
Kriechi: This action is only for convenience to speed up the dev cycle? or does it provide anything we don't have in tox?

.github/workflows/main.yml
```diff
@@ -126,56 +113,129 @@ jobs:
         run: yarn
       - working-directory: ./web
         run: npm test
-      - uses: codecov/codecov-action@v1
+      - uses: codecov/codecov-action@a1ed4b322b4b38cb846afb5a0ebfa17086917d27
```
Kriechi, line 116: pinning is nice - but who is checking for updates if they fix a critical security issue?
for python deps we have requires.io/dependabot, I'm not aware of similar tooling for github action steps? (next startup idea?)

.github/workflows/main.yml
Kriechi: fail early?

.github/workflows/main.yml
```diff
@@ -126,56 +113,129 @@ jobs:
...
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - image: macos-10.15
+            platform: macos
+          - image: windows-2019
+            platform: windows
+          - image: ubuntu-16.04  # Old Ubuntu version for old glibc
```
Kriechi, line 180: 16.04 LTS is now EOL since a few weeks.
When was the last time we got a user complaining about glibc issues? 
I'd be happy to see this bumped to 18.04 or even 20.04.

.github/workflows/main.yml
```diff
@@ -126,56 +113,129 @@ jobs:
...
+        with:
+          name: binaries.${{ matrix.platform }}
+          path: release/dist
 
   # Separate from everything else because slow.
   build-and-deploy-docker:
     if: github.repository == 'mitmproxy/mitmproxy' && github.event_name == 'push'
     environment: deploy-docker
-    needs: [test, test-web-ui, build-wheel]
+    needs: [ test, test-web-ui, build ]
```
Kriechi, line 213: nitpick: how about a proper itemized list instead?

release/docker/Dockerfile
Kriechi: Should we pin the container itself as well?

release/pins/requirements-linux.txt
```diff
@@ -0,0 +1,811 @@
+#
```
Kriechi, line 1: When during our dev cycle do we check in these files? And when do we update + verify + test them?

release/pins/make.py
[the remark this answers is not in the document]
mhils: Well we can only pin our dependencies, we can't pin mitmproxy itself. That hash changes with every commit. 😄 So in either case we need to doctor ourselves out of the generated file. 

The alternative would be to have all requirements in a separate requirements.txt file and load that file in setup.py, then we could ingest that requirements file here. But that means ugly file processing in setup.py for environment markers. I don't think this would be much better?

setup.py
```diff
@@ -97,16 +97,21 @@
             "pydivert>=2.0.3,<2.2",
         ],
         'dev': [
+            "awscli>=1.19,<2",
```
Kriechi, line 100: I have a slight urge to get v2 - but unfortunately they don't publish it on pypi anymore because it's considered "a stand-alone app". v1 is still fully supported, but let's keep a mental note of this.

Also, I just realized that installing CLI tools as Python dependencies feels dirty. Hugo is installed "normally", so why not awscli? https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-linux.html#cliv2-linux-install

I'll classify twine and tox as edge cases.

setup.py
```diff
@@ -97,16 +97,21 @@
...
             "hypothesis>=5.8,<6.11",
             "parver>=0.1,<2.0",
             "pdoc>=4.0.0",
+            "pip-tools>=6.1.0,<6.2",
```
Kriechi, line 104: nitpick: I tend to use `>=x.y` instead of `>=x.y.z` if there is no specific `z` requirement. Might be nice to unify everything to one style. You already do so for the upper bound.

.github/workflows/main.yml
briansmith: Dependabot apparently will notify you if you have it enabled. I haven't tested it.

PR #4600, opened 2021-05-14
Feature: add ARM64 docker platform variants

hazcod:
#### Description

A very fresh proposal to also build arm64 docker images using buildx from docker.

#### Checklist

### DIFF
--- a/release/cibuild.py
+++ b/release/cibuild.py
@@ -280,7 +280,9 @@ def build_docker_image(be: BuildEnviron) -> None:  # pragma: no cover
     shutil.copy(whl, docker_build_dir / whl.name)
     subprocess.check_call([
         "docker",
+        "buildx"
         "build",
+        "--platform linux/amd64,linux/arm64,darwin/amd64,darwin/arm64",
         "--tag", be.docker_tag,
         "--build-arg", f"MITMPROXY_WHEEL={whl.name}",
         "."


mhils:
That looks fantastically simple. Let's see what CI says. :)

mhils:
CI is currently erroring with 
```
unknown flag: --platform linux/amd64,linux/arm64,darwin/amd64,darwin/arm64
```

If we could get this fixed easily that'd be phenomenal! 😃

hazcod:
Odd, since I'm following the docs at https://docs.docker.com/buildx/working-with-buildx/#build-multi-platform-images

release/cibuild.py
```diff
@@ -280,7 +280,9 @@ def build_docker_image(be: BuildEnviron) -> None:  # pragma: no cover
     shutil.copy(whl, docker_build_dir / whl.name)
     subprocess.check_call([
         "docker",
+        "buildx"
```
siddharths2710, line 283: You've missed the comma separator here. Subprocess gladly issues the command without erroring out since the new PEP guidelines concatenate two strings adjacent to each other. The CI fails since it ran `docker buildxbuild` instead of `docker buildx build` specified in the Docker docs.

mhils: Thanks @siddharths2710, really appreciate it! 🍰 Let's see if CI likes it now. I think this PR is really promising, it would be fantastic if you or @hazcod could push this to a state where it's mergable. 😃

hazcod:
If that would be the issue, that would be so awesome and lame at the same time. 😄

hazcod:
Seems to work now, great! Thanks for the keen eye @mhils

PR #4608, opened 2021-05-24
Add `flow.comment` command and keybinding to add a comment to a flow.

rbdixon:
#### Description

A very simple addition of a useful feature from other proxy tools.

#### Checklist

### DIFF
--- a/mitmproxy/io/compat.py
+++ b/mitmproxy/io/compat.py
@@ -309,6 +309,12 @@ def convert_12_13(data):
     return data
 
 
+def convert_13_14(data):
+    data["version"] = 14
+    data["comment"] = ""
+    return data
+
+
 def _convert_dict_keys(o: Any) -> Any:
     if isinstance(o, dict):
         return {strutils.always_str(k): _convert_dict_keys(v) for k, v in o.items()}
@@ -369,6 +375,7 @@ converters = {
     10: convert_10_11,
     11: convert_11_12,
     12: convert_12_13,
+    13: convert_13_14,
 }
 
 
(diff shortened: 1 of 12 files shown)


mhils:
That's probably a useful feature, but if I understand your PR correctly this comment is not exposed anywhere in the UI at the moment? How are you accessing those comments?

Second note: We're really trying to get away from `.metadata` due to the untyped/unstructured madness, if this makes it in the we should have it as `flow.comment`. :)

mhils:
We don't want to ditch it, it's really quite useful for scripts. But we're trying to not have mitmproxy itself depend on it too much, it's very easy to introduce type bugs here. I'd much rather have a dedicated `flow.comment: str` attribute.

mitmproxy/io/compat.py
```diff
@@ -300,6 +300,12 @@ def convert_11_12(data):
     return data
 
 
+def convert_12_13(data):
```
mhils, line 303: There is a race condition with your other PR here, we now need to transition from 13 to 14 :)

mitmproxy/tools/console/flowdetailview.py
```diff
@@ -33,6 +33,10 @@ def flowdetails(state, flow: mitmproxy.flow.Flow):
         req = None
         resp = None
     metadata = flow.metadata
+    comment = flow.comment
+
+    if len(comment) > 0:
```
mhils, line 38: Nit: `if flow.comment:` is a bit more Pythonic, `bool("")` evaluates to False; every other `str` to True.

mhils (commented):
👍

mhils (approved):

Issue #4610, opened 2021-05-25
iOS 13 TLS server certificates must have a validity period of 825 days


loretoparisi:
#### Problem Description
Certificate installed via mitm.it is expired.

[This](https://alext.mail.at/?p=727) workaround proposes this approach

```
Starting with IOS 13, TLS server certificates must have a validity period of 825 days or fewer and MITMProxy generates certs with an expiration period of 1095 days.
Solution: I changed the py file of MITMProxy to shorten the cert validity, by changing the file /usr/lib/python2.7/dist-packages/netlib/certutils.py
# DEFAULT_EXP = 94608000  # = 24 * 60 * 60 * 365 * 3
DEFAULT_EXP = 31536000  # = 24 * 60 * 60 * 365
```

The problem is that I cannot follow this approach since I'm using the `mitmproxy` docker image (see below)

#### Steps to reproduce the behavior:
1.  docker pull mitmproxy/mitmproxy 
2.  docker run --rm -it -v ~/.mitmproxy:/home/mitmproxy/.mitmproxy -p 8080:8080 mitmproxy/mitmproxy
3.  Go to mitm.it from iOS device
4.  Go to General > Profiles > mitmproxy > the profile is expired 9 March 2020

#### System Information
Dockerhub image 

```
mitmproxy/mitmproxy   latest    37c70985d385   5 months ago   122MB
```

mhils:
You are conflating two issues here:

1. iOS validity limitation: All recent versions of mitmproxy handle this properly.
2. Expired certificates. It looks like your mitmproxy certificate was generate too long ago. 

The fix for the second issue is to delete all CA-related files in your ~/.mitmproxy folder. The CA will be regenerated automatically after restarting mitmproxy (and then needs to be reinstalled on your devices).

There also should be a "The mitmproxy certificate authority has expired" message in your mitmproxy event log somewhere with these steps. :)

loretoparisi:
@mhils thanks, for point 2), since I'm using `docker` I do not have any previous local folder with expired certificates:

```bash
docker run --rm -it -v ~/.mitmproxy:/home/mitmproxy/.mitmproxy -p 8080:8080 mitmproxy/mitmproxy
```

I mount this (new) volume with fresh certificates;
Then I go to mitmproxy.it and I install the (new) certificate, that should be valid, while it is expired.
So I assume in this config, I do not have any previously expired certificates on my host machine...

mhils:
[...]

PR #4611, opened 2021-05-25
Make the CA cert live longer

mhils:
We need to restrict `CERT_EXPIRY` because modern browsers only accept a ~ <= 1year period for certs, but re-reading all prior commits and issues as well as the CA/Browser Forum Baseline Requirements on the topic, it looks like we don't actually need to restrict our CA lifetime by that much. Let's try out a longer validity and see if anyone complains.

### DIFF
--- a/mitmproxy/certs.py
+++ b/mitmproxy/certs.py
@@ -17,7 +17,7 @@ import OpenSSL
 from mitmproxy.coretypes import serializable
 
 # Default expiry must not be too long: https://github.com/mitmproxy/mitmproxy/issues/815
-CA_EXPIRY = datetime.timedelta(days=3 * 365)
+CA_EXPIRY = datetime.timedelta(days=10 * 365)
 CERT_EXPIRY = datetime.timedelta(days=365)
 
 # Generated with "openssl dhparam". It's too slow to generate this on startup.


mhils:
Did you read the PR description? 😅 

I've wondered about this before, and an inquiry on Slack yesterday + #4610 today made look at this properly. The CA/B reqs don't specify anything, git blame only ever lists requirements for leaf certs, the linked threads do not have any other evidence either, and intuitively it doesn't make sense to limit CAs by so much.

Kriechi:
At least https://github.com/mitmproxy/mitmproxy/commit/9d12425d5ee942ee3d954a9324c31b74f466d520 would indicate we should not use 10years.
Or did we mis-interpret the requirements back then already?

mhils:
Yes. And at this point we only had one shared constant for cert + CA. My understanding is we added a second constant when the requirements for certs were reduced to one year, which is why we currently still have the three year CA limit in place.

Kriechi:
Overall, I'm slightly in favour of keeping it short.
With the appropriate amount of tin-foil-hat mindset, one could argue that having a fully-trusted 10y-valid root CA on a random dev machine is not very secure. Laptops get lost, random backups put on shared drives or copied to thumb drives, co-workers might have access to the same user profile (running as root?!)...
So forcing users to re-create it once a year could protect users from themselves to some degree. I'm happy to hear arguments against this nanny state approach.

Either way, we definitely should improve our UI/UX though to make it obvious what the problem is - and provide easy steps to fix it. I thought we have seen PRs for this in the past as well?

mhils:
> With the appropriate amount of tin-foil-hat mindset, one could argue that having a fully-trusted 10y-valid root CA on a random dev machine is not very secure. Laptops get lost, random backups put on shared drives or copied to thumb drives, co-workers might have access to the same user profile (running as root?!)...

I don't think 10 years is meaningfully less secure than 3 years, but it makes it significantly less likely that anyone ever runs into issues with expiring CAs. Even though we already [provide instructions](https://github.com/mitmproxy/mitmproxy/blob/292e3b413cd6cff497e35f151e29c6d786417144/mitmproxy/addons/tlsconfig.py#L242-L248) in case of expiry, this issue has required me to spend some time on it twice in the last 48 hours.

Kriechi:
1. if 10y CAs are accepted by all major modern systems, I'm 👍 for merging.
2. if you needed to "spend some time" twice on providing support, we need to improve the instructions - clearly this should be at most a simple "RTFM with the provided link".
3. Kudos and extra unicorn stickers to those faithful users who used mitmproxy for more than 3 years 🥇

PR #4619, opened 2021-06-03
[mitmweb] Add raw-edit mode to Header editor

gorogoroumaru:
#### Description

We can edit the headers and contents in HTTP request in mitmweb, but we currently can’t edit raw HTTP header like BurpSuite proxy.
I added an ability to edit raw HTTP headers.

#### Checklist


mhils:
We have a vastly improved editor on main now. 🍰

PR #4620, opened 2021-06-03
Update README.md

Prinzhorn:
#### Description

The original docs didn't make sense to me and the CSS part still does not make sense to me. All `modd` does is watch `scripts/**` and then call `build.py`. Why would that be CSS related but not the original instructions to manually call `build.py`?

Maybe you can clear that up :shrug:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -12,7 +12,3 @@ This directory houses the mitmproxy documentation available at <https://docs.mit
  1. Make sure the mitmproxy Python package is installed and the virtual python environment was activated. See [CONTRIBUTING.md](../CONTRIBUTING.md#development-setup) for details.
  2. Run `./build.py` to generate additional documentation source files.
  3. Now you can run `hugo server -D` in `./src`.
-
-## Editing CSS files
-
-This is required to modify CSS files. Instead of manually running `./build.py` and `hugo server -D` you can run `modd`. It will watch for changes.
--- a/docs/README.md
+++ b/docs/README.md
@@ -2,20 +2,17 @@
 
 This directory houses the mitmproxy documentation available at <https://docs.mitmproxy.org/>.
 
-## Quick Start
+## Prerequisites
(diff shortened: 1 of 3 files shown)


mhils:
Ah, yes. The current version is confusing/wrong. The upper section initially didn't mention the need for the extended version of Hugo, the CSS section then explicitly pointed out that you need to install that one to edit styles. Then I recently merged #4560 (which looked good from the diff), but didn't realize that this makes the second section superfluous/confusing.

Let's take out modd from the README. It's one of my favorite tools that I constantly use, but we don't really need to mention it here and it adds friction for new editors.

Prinzhorn:
> Let's take out modd from the README. It's one of my favorite tools that I constantly use, but we don't really need to mention it here and it adds friction for new editors.

To clarify, you use modd when editing the "meta" Python scripts, something that "normal" editing of the docs doesn't require?

mhils (approved):

Prinzhorn:
> To clarify, you use modd when editing the "meta" Python scripts, something that "normal" editing of the docs doesn't require?

@mhils :eyes:

mhils:
@Prinzhorn: missed your question, thanks for the heads-up.

modd is basically just providing "recompile-on-change" functionality. Most of my projects have modd.conf files, so I just enter the directory and type `modd`, without having to remember the particular setup. It doesn't bring much to the table here except consistency (for modd users).

Prinzhorn:
@mhils I'm aware what modd does and already had it installed. I was just making sure I understood what you are using it for in this particular case. So for a regular contributor that just wants to fix a spelling mistake for example it is irrelevant, but when you personally are editing some of the Python scripts it is just convenient to not `./build.py` manually, correct? But these scripts are not something that needs to be touched in general?

mhils:
Yes, correct. (missed the reply here, but better late than never :) )

PR #4624, opened 2021-06-07
[mitmweb] Add hover menu to FlowRow

gorogoroumaru:
#### Description

I added a "quick actions" menu, which appears when hovering each row.
The menu can add an interception filter corresponding to the host and IP address of the flow.

#### Checklist


mitmproxy/flowfilter.py
```diff
@@ -381,6 +381,17 @@ def __call__(self, f):
         return f.server_conn.address and self.re.search(r)
 
 
+class FDstIP(_Rex):
+    code = "ip"
```
mhils, line 385: Is there a reason why we can't use `~dst`? It feels like that could be good enough.I would like to avoid extending the filter syntax unless there's an urgent need. :)

mhils (commented):
This looks good!

- Let's make sure that we hide the second option here if the path is just `/`:
image

I think the next big part here is adding submenus. We should aim for a "Intercept request like this..." and "Intercept response like this..." categories, as in the sketch we discussed on Slack.

mitmproxy/flowfilter.py
gorogoroumaru: The following is the code for ~dst.
This uses f.server_conn.address for filtering, but f.server_conn.address isn't IP address. It's host name.
So I need to add another filter which uses f.server_conn.ip_address.

```
class FDst(_Rex):
    code = "dst"
    help = "Match destination address"
    is_binary = False

    def __call__(self, f):
        if not f.server_conn or not f.server_conn.address:
            return False
        r = "{}:{}".format(f.server_conn.address[0], f.server_conn.address[1])
        return f.server_conn.address and self.re.search(r)```

Comment (mhils) [reply mitmproxy/flowfilter.py:385]: I think I understand that part, but why do you want to filter by IP address and not by host name? Do you have an example where that is advantageous? :)

Comment (gorogoroumaru) [reply mitmproxy/flowfilter.py:385]: As you described in the chart, I added an item “Intercept all requests from this IP”.
Sometimes, more than one host names can have the same ip address.
In such a case, it may be useful.
But if you don’t need it, I will delete it.

Comment (mhils) [reply mitmproxy/flowfilter.py:385]: Dang! Sorry, I didn't realize I actually asked for that. 🙊 I think filtering by hostname is perfectly fine here, let's go with that! 😃

File: web/src/js/components/FlowTable/FlowColumns.jsx [hunk 2]
```diff
@@ -175,7 +176,9 @@ export function QuickActionsColumn({flow}) {
     return (
         <td className="col-quickactions">
             <div>
-                <div className="quickaction"><i className="fa fa-fw fa-ellipsis-h"/></div>
+                {selected ?
```
mhils, line 179: is there a reason why we only want to show the quickactions menu for selected flows?

web/src/js/components/ProxyApp.jsx
```diff
@@ -11,11 +11,11 @@ import Modal from './Modal/Modal'
 
 class ProxyAppMain extends Component {
 
-    componentWillMount() {
+    componentDidMount() {
         window.addEventListener('keydown', this.props.onKeyDown);
     }
 
-    componentWillUnmount() {
+    UNSAFE_componentWillUnmount() {
```
mhils, line 18: Unmount is not prefixed with `UNSAFE_`, is it?

https://reactjs.org/docs/react-component.html#componentwillunmount

web/src/js/components/ProxyApp.jsx
```diff
@@ -11,11 +11,11 @@ import Modal from './Modal/Modal'
 
 class ProxyAppMain extends Component {
 
-    componentWillMount() {
+    componentDidMount() {
```
mhils, line 14: 👍

web/src/js/components/Header/FilterInput.jsx
```diff
@@ -25,7 +25,7 @@ export default class FilterInput extends Component {
         this.selectFilter = this.selectFilter.bind(this)
     }
 
-    componentWillReceiveProps(nextProps) {
+    UNSAFE_componentWillReceiveProps(nextProps) {
```
mhils, line 28: 👍

mhils (commented):
This looks good. Here's my initial set of comments from reading the diff, I haven't checked the code out locally yet to play around with it. Will do that soon and then probably follow up with more questions. :)

web/src/js/components/FlowTable/FlowColumns.jsx
gorogoroumaru: The hover menu adds an interception filter which corresponds to the selected flow.
So, if we hover to a flow which is not selected, the hover menu still shows the interception filter suggestion of the selected flow.
This is very confusing.
So for now, I've made the menu appear only for the selected flow.

web/src/js/components/ProxyApp.jsx
gorogoroumaru: I made a mistake. I'll fix it.

web/src/js/components/FlowTable/FlowColumns.jsx
```diff
@@ -164,29 +165,62 @@ export function TimeStampColumn({flow}) {
 TimeStampColumn.headerClass = 'col-timestamp'
 TimeStampColumn.headerName = 'TimeStamp'
 
-export function QuickActionsColumn({flow, selected}) {
+export function QuickActionsColumn({flow, intercept, updateSettings, removeFlow, exportFlow}) {
+    const InterceptThis = useCallback((example) => {
```
mhils, line 169: Let's use the `useDispatch` hook here: https://react-redux.js.org/tutorials/quick-start#use-redux-state-and-actions-in-react-components. This allows us to keep keep `FlowRow` "clean" and we don't need to specify all actions in FlowRow.jsx. 😃 

For `state.settings.intercept` it's a bit more tricky as we want FlowRow to be cached/memoized. If we use `useSelector` in a child component, my understanding is that the child component will not be rerendered on changes (because the parent FlowRow is cached), so we cannot do that. I think we can however move the logic from `InterceptThis` into an action creator like this:

`ducks/settings.js`
```javascript
export function addInterceptFilter(e) {
    return (dispatch, getState) => {
        let intercept = getState().settings.intercept;
        // logic here
        dispatch(updateSettings({ intercept }));
    }
}    
```

Then we can just `dispatch(addInterceptFilter(...))` here and don't need any state. Does that make sense to you? Otherwise ping me on Slack and I'll try to explain it better. :)

web/src/js/components/FlowTable/FlowColumns.jsx
```diff
@@ -164,29 +165,62 @@ export function TimeStampColumn({flow}) {
...
+
+    const exportAsCURL = useCallback(() => {
+        if (!flow) {
+            return
+        }
+
+        fetchApi(`/flows/${flow.id}/export`, { method: 'POST' })
+        .then(response => response.json())
+        .then(data => {
+            navigator.clipboard.writeText(data.export)
```
mhils, line 189: Really nice. 👍  I didn't know this is so easy nowadays. 😁

mitmproxy/tools/web/app.py
```diff
@@ -266,6 +272,12 @@ def post(self):
         bio.close()
 
 
+class ExportFlow(RequestHandler):
```
mhils, line 275: I think right after this PR is complete we should add an API endpoint to execute commands. We can then get rid of this one as it's basically the same as invoking a particular command. But that's work for the next PR, not this one. 😃

mhils (commented):
Excellent! Can you fix the tests and look at my `useDispatch` suggestion please? Other than that this looks good to merge. 😃

gorogoroumaru:
I applied your suggested changes.
It seems that your changes affected the test failure.
I don't know how to solve it.

Issue #4613, opened 2021-05-28
Unable to edit multipart/form-data via mitmproxy UI


jhulme:
#### Problem Description
Changes made when editing multipart/form-data on an intercepted request are lost when returning to view the flow.

#### Steps to reproduce the behavior:

1. Set intercept for requests and multipart/form-data - `~q & ~t "multipart/form-data"`
![set_intercept_gh](https://user-images.githubusercontent.com/809451/119967488-4cdc9b80-bfa4-11eb-813c-18797040e164.JPG)
2. Generate simple request with multipart form (during testing used [httpie ](https://httpie.io/)tool)
3. Select flow within mitmproxy UI
![flow_details_gh](https://user-images.githubusercontent.com/809451/119967487-4c440500-bfa4-11eb-907f-099a7f41ac73.JPG)
4. Press 'e' to edit and select multipart
![edit_flow_gh](https://user-images.githubusercontent.com/809451/119967482-4c440500-bfa4-11eb-82fe-018eef0f6ca5.JPG)
5. Update some portion of the data
![editing_form_gh](https://user-images.githubusercontent.com/809451/119967479-4bab6e80-bfa4-11eb-8730-860943c6316e.JPG)
6. Press 'esc' to retain changes
7. Press 'q' to return to flow view. Expect to see edited data, instead no data is present and Content-Length is set to 0.
![missing_content_gh](https://user-images.githubusercontent.com/809451/119967476-4b12d800-bfa4-11eb-9691-827cfcce36f9.JPG)

I expect that I am following an incorrect process here in some manner that is resulting in the behaviour, so any guidance would be appreciated.

I have also tried rolling back and testing with v5.3.0 again with the same behaviour.

#### System Information
```
Mitmproxy: 6.0.2 binary
Python:    3.9.1
OpenSSL:   OpenSSL 1.1.1i  8 Dec 2020
Platform:  Linux-4.19.128-microsoft-standard-x86_64-with-glibc2.27
```

roytu:
I'm running into this issue as well.  Seems a few places in the code are dropping the "boundary" field in the Content-Type when it comes to multipart packets.

I found three locations in the code where this assumption is made, and after adding those fixes it seems to (mostly) work.  There's still an issue where the multipart editor appends b'' to all the fields, and editing them runs into a bytes vs. string type error, but it's closer now.

[...]

PR #4625, opened 2021-06-08
Fix multipart forms losing `boundary` values on edit

roytu:
#### Description

This is a fix for issue #4613 .  The problem is that in several places where multipart is used, the code assumes that the header will match `Content-Type: multipart/form-data` *exactly*, whereas the reality often has `Content-Type: multipart/form-data; boundary=---------foo`.

#### Checklist

### DIFF
--- a/mitmproxy/contentviews/__init__.py
+++ b/mitmproxy/contentviews/__init__.py
@@ -101,7 +101,7 @@ def get_message_content_view(
         http_message = message
         if ctype := message.headers.get("content-type"):
             if ct := http.parse_content_type(ctype):
-                content_type = http.assemble_content_type(ct[0], ct[1], ct[2])
+                content_type = f"{ct[0]}/{ct[1]}"
 
     description, lines, error = get_content_view(
         viewmode, content,
--- a/mitmproxy/contentviews/__init__.py
+++ b/mitmproxy/contentviews/__init__.py
@@ -101,7 +101,7 @@ def get_message_content_view(
         http_message = message
         if ctype := message.headers.get("content-type"):
             if ct := http.parse_content_type(ctype):
-                content_type = f"{ct[0]}/{ct[1]}"
+                content_type = http.assemble_content_type(ct[0], ct[1], ct[2])
 
     description, lines, error = get_content_view(
         viewmode, content,
--- a/mitmproxy/contentviews/multipart.py
+++ b/mitmproxy/contentviews/multipart.py
@@ -21,4 +21,7 @@ class ViewMultipart(base.View):
             return "Multipart form", self._format(v)
 
     def render_priority(self, data: bytes, *, content_type: Optional[str] = None, **metadata) -> float:
-        return float(content_type == "multipart/form-data")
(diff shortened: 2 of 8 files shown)


mitmproxy/contentviews/__init__.py
```diff
@@ -101,7 +101,7 @@ def get_message_content_view(
         http_message = message
         if ctype := message.headers.get("content-type"):
             if ct := http.parse_content_type(ctype):
-                content_type = f"{ct[0]}/{ct[1]}"
+                content_type = http.assemble_content_type(ct[0], ct[1], ct[2])
```
mhils, line 104: I believe we never need the content type attributes when selecting a contentview? Unless there's something I am missing, please revert the changes here.

mitmproxy/contentviews/multipart.py
```diff
@@ -21,4 +21,7 @@ def __call__(self, data: bytes, content_type: Optional[str] = None, **metadata):
             return "Multipart form", self._format(v)
 
     def render_priority(self, data: bytes, *, content_type: Optional[str] = None, **metadata) -> float:
-        return float(content_type == "multipart/form-data")
+        if content_type and content_type.startswith("multipart/form-data"):
```
mhils, line 24: see above - this is only used to select the correct view, we don't need to handle the boundary information here.

mitmproxy/http.py
```diff
@@ -944,7 +944,9 @@ def _get_multipart_form(self):
 
     def _set_multipart_form(self, value):
         self.content = multipart.encode(self.headers, value)
-        self.headers["content-type"] = "multipart/form-data"
+        if "content-type" not in self.headers:
+            # Don't overwrite header if it already exists or it will destroy the boundary value
+            self.headers["content-type"] = "multipart/form-data"
```
mhils, line 949: This... looks like it plainly didn't work before? Good catch.

I would suggest we change the logic here to 

1. Check if `self.headers["content_type"].startswith("multipart/form-data")`, and if that's not the case, add a content-type header with a random (?) boundary. The point here is that if someone assigns to `.multipart_form = ` in their addon, the header should be updated accordingly if the request was not multipart before.
2. *After* that, call `self.content = multipart.encode(self.headers, value)` (as we do right now)

mhils (changes requested):
Thanks! This looks like a good catch. I've added some comments below. In short, we can take contentviews out of the picture, and then change the multipart setter to actually work. 😃

roytu:
Thanks for the comments :)

Added commits to revert the `contentviews` stuff and the fix still seems to work, so I believe you're correct that the content type attributes aren't used there.

Also updated the `set_multipart_form` function to generate a boundary if the header doesn't exist (stole it from https://github.com/urllib3/urllib3/blob/2ef05adfcf001c85f1f95ae3fa078d16190851b0/src/urllib3/filepost.py#L20-L24 and added a few dashes for readability).

mhils (approved):
This looks fantastic now, thanks for the quick turnaround! 🍰 😃

PR #4629, opened 2021-06-09
Update Web Dependencies

mhils:
This PR updates our some of our web dependencies, in particular React. 

@gorogoroumaru: `yarn run jest` now emits a bunch of errors because we use `componentWillMount`, `componentWillReceiveProps`, and `componentWillUpdate` at a couple of places (see https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html). Could you help me with this PR and fix the remaining  warnings? In some cases I could trivially rename `componentWillMount` to `componentDidMount`, in more complicated cases I just slapped `UNSAFE_` in front of it. You can probably figure out what's best in each remaining case. Does that make sense?


gorogoroumaru:
It seems that we need to change class components into function components.
Then, we can replace componentWillMount, componentWillReceiveProps, and componentWillUpdate with useEffect hooks.

mhils:
Functional components would also work, yes. We do have some simpler approaches:

For example here I am just replacing `componentWillMount` with `componentDidMount` as the difference does not matter in this case:

https://github.com/mitmproxy/mitmproxy/pull/4629/files#diff-e311666830bf5c7388c50fbe00e5e33b0a0f8ee2734fb359f37d9ec6a381fa59L17-R17

This doesn't work for all components. For example AutoScroll is a bit more advanced, here I just added `UNSAFE_`: https://github.com/mitmproxy/mitmproxy/pull/4629/files#diff-03038a58726440ecb9c82f48a0dc727a03a6f40dbbf85e38cedc4109c22a5e39L11-R11

Long story short, let's find easy fixes for all remaining occasions and we should be good. 😃  I'm merging this PR here, can you follow up with a new PR that fixes the remaining warnings? 😊

PR #4637, opened 2021-06-15
Docker: Add aarch64 Images

mhils:
We now also build aarch64 wheels. 🙌 
From our end we'd also be ready for ARMv6/ARMv7 builds (which would be nice for older Raspberry Pis), but that is currently blocked on https://github.com/rust-lang/cargo/issues/9545. This is too much of a rabbit hole, so 64 bit ARM will do. :)


hazcod:
Super nice! Thanks for getting into this.

PR #4644, opened 2021-06-18
[mitmweb] fix 'hide response type menu' bug

ErfanBahramali:
#### Description

When the answer type list is displayed, part of this list will be hidden if the answer is low

Example:

Screenshot from 2021-06-18 13-22-21

Fixed problem result:

Screenshot from 2021-06-18 13-24-55

#### Checklist

### DIFF
--- a/web/src/css/flowdetail.less
+++ b/web/src/css/flowdetail.less
@@ -20,6 +20,8 @@
             padding: 5px 12px 0;
         }
         >footer {
+            position: absolute;
+            bottom: 0;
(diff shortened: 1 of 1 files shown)


mhils:
Thanks! I appreciate it - please keep mitmweb PRs coming. I think this particular bug is actually inadvertedly fixed by https://github.com/mitmproxy/mitmproxy/pull/4624, but I'll keep this PR open to double-check once that is in. 😃

mhils:
Fixed on main, thanks again! 🍰

PR #4646, opened 2021-06-20
[mitmweb] Add command bar where we can use mitmproxy commands

gorogoroumaru:
#### Description

We haven't been able to use powerful mitmproxy commands in mitmweb.
I added the command bar where we can use mitmproxy commands.

#### Checklist


mitmproxy/tools/web/app.py
```diff
@@ -431,6 +435,15 @@ def get(self, flow_id, message, content_view):
         ))
 
 
+class Commands(RequestHandler):
+    def post(self):
+        result = self.master.commands.execute(self.command)
```
mhils, line 440: Can we just access `request.json` here instead of adding the `.command` property? That's more direct and in line with the other code. We only need that specific getter here.

mhils (approved):
Good stuff! There's obviously more to do, but this is a solid start. Can you implement the minor fix below, and then merge this yourself into the gsoc branch please? 😃

mhils:
The command bar is in the main branch now 🍰

PR #4650, opened 2021-06-22
[WIP] Fix WebSocket/TCP injection

mhils:
@Prinzhorn: Very early WIP - My plan is to change the command arguments to bytes (simple, see below) and handle the conversion as part of the commands processing logic (harder, to be done). Would that generally satisfy your use case? :)

Also thanks again for your `.is_text` input. That was really quite useful. The public API is now just that, while we internally keep the opcode around, which is not much more effort.

### DIFF
--- a/test/mitmproxy/test_websocket.py
+++ b/test/mitmproxy/test_websocket.py
@@ -1,3 +1,5 @@
+import pytest
+
 from mitmproxy import http
 from mitmproxy import websocket
 from mitmproxy.test import tflow
@@ -26,3 +28,18 @@ class TestWebSocketMessage:
         assert not m.killed
         m.kill()
         assert m.killed
+
+    def test_text(self):
+        txt = websocket.WebSocketMessage(Opcode.TEXT, True, b"foo")
+        bin = websocket.WebSocketMessage(Opcode.BINARY, True, b"foo")
+
+        assert txt.is_text
+        assert txt.text
+        txt.text = "bar"
+        assert txt.content == b"bar"
+
+        assert not bin.is_text
+        with pytest.raises(AttributeError, match="do not have a 'text' attribute."):
+            _ = bin.text
+        with pytest.raises(AttributeError, match="do not have a 'text' attribute."):
+            bin.text = "bar"
--- a/test/mitmproxy/test_websocket.py
+++ b/test/mitmproxy/test_websocket.py
@@ -34,7 +34,7 @@ class TestWebSocketMessage:
         bin = websocket.WebSocketMessage(Opcode.BINARY, True, b"foo")
 
         assert txt.is_text
-        assert txt.text
+        assert txt.text == "foo"
         txt.text = "bar"
         assert txt.content == b"bar"
 
--- a/docs/scripts/pdoc-template/module.html.jinja2
+++ b/docs/scripts/pdoc-template/module.html.jinja2
@@ -55,6 +55,10 @@ To document all event hooks, we do a bit of hackery:
             {% if doc.qualname.startswith("ServerConnectionHookData") and doc.name != "__init__" %}
                 {{ default_is_public(doc) }}
             {% endif %}
+        {% elif doc.modulename == "mitmproxy.websocket" %}
+            {% if doc.qualname != "WebSocketMessage.type" %}
+                {{ default_is_public(doc) }}
+            {% endif %}
         {% else %}
             {{ default_is_public(doc) }}
         {% endif %}
--- a/mitmproxy/addons/proxyserver.py
+++ b/mitmproxy/addons/proxyserver.py
@@ -190,15 +190,14 @@ class Proxyserver:
         self._connections[event.flow.client_conn.peername].server_event(event)
 
     @command.command("inject.websocket")
-    def inject_websocket(self, flow: Flow, to_client: bool, message: str, is_text: bool = True):
+    def inject_websocket(self, flow: Flow, to_client: bool, message: bytes, is_text: bool = True):
         if not isinstance(flow, http.HTTPFlow) or not flow.websocket:
             ctx.log.warn("Cannot inject WebSocket messages into non-WebSocket flows.")
 
-        message_bytes = strutils.escaped_str_to_bytes(message)
         msg = websocket.WebSocketMessage(
             Opcode.TEXT if is_text else Opcode.BINARY,
             not to_client,
-            message_bytes
+            message
         )
         event = WebSocketMessageInjected(flow, msg)
         try:
@@ -207,12 +206,11 @@ class Proxyserver:
             ctx.log.warn(str(e))
 
     @command.command("inject.tcp")
-    def inject_tcp(self, flow: Flow, to_client: bool, message: str):
+    def inject_tcp(self, flow: Flow, to_client: bool, message: bytes):
         if not isinstance(flow, tcp.TCPFlow):
             ctx.log.warn("Cannot inject TCP messages into non-TCP flows.")
 
-        message_bytes = strutils.escaped_str_to_bytes(message)
-        event = TcpMessageInjected(flow, tcp.TCPMessage(not to_client, message_bytes))
+        event = TcpMessageInjected(flow, tcp.TCPMessage(not to_client, message))
         try:
             self.inject_event(event)
         except ValueError as e:
--- a/mitmproxy/addons/proxyserver.py
+++ b/mitmproxy/addons/proxyserver.py
@@ -8,7 +8,7 @@ from mitmproxy.proxy import commands, events, server_hooks
 from mitmproxy.proxy import server
 from mitmproxy.proxy.layers.tcp import TcpMessageInjected
 from mitmproxy.proxy.layers.websocket import WebSocketMessageInjected
-from mitmproxy.utils import asyncio_utils, human, strutils
+from mitmproxy.utils import asyncio_utils, human
 from wsproto.frame_protocol import Opcode
 
 
--- a/mitmproxy/command.py
+++ b/mitmproxy/command.py
@@ -106,8 +106,15 @@ class Command:
             raise exceptions.CommandError(f"Command argument mismatch: \n    {expected}\n    {received}")
 
         for name, value in bound_arguments.arguments.items():
-            convert_to = self.signature.parameters[name].annotation
-            bound_arguments.arguments[name] = parsearg(self.manager, value, convert_to)
+            param = self.signature.parameters[name]
+            convert_to = param.annotation
+            if param.kind == param.VAR_POSITIONAL:
+                bound_arguments.arguments[name] = tuple(
+                    parsearg(self.manager, x, convert_to)
+                    for x in value
+                )
+            else:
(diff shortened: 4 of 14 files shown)


Prinzhorn:
Disclaimer: I have never used commands in mitmproxy at all, so I might misunderstand something regarding how they are used.

> My plan is to change the command arguments to bytes

That's exactly what I did https://github.com/mitmproxy/mitmproxy/pull/4562#issuecomment-817948056 and launching your branch currently produces exactly that

> Could not load command inject.tcp: Argument message has an unknown type <class 'bytes'> in <bound method Proxyserver.inject_tcp of ProxyServer(stopped, 0 active conns)>.
> Could not load command inject.websocket: Argument message has an unknown type <class 'bytes'> in <bound method Proxyserver.inject_websocket of ProxyServer(stopped, 0 active conns)>

So my main issue was that I didn't know how to make commands accept bytes. Since commands are tied to the UI I also don't know if that's even useful? What I was thinking was to keep the command and have an injection API that is _not_ a command (which accepts bytes). Allowing the user to inject strings or bytes via the UI belongs to the UI layer, the internal injection code should not be concerned about that at all. If I want to inject raw bytes into a TCP connection I want exactly those bytes. But since it's hard to enter raw bytes in a UI it needs a way to make that possible for the user. And `escaped_str_to_bytes` is one possible solution, _but_ it must only be used for `Opcode.BINARY` if the user want to explicitly inject bytes and use the escaped string syntax for serialization to make bytes human friendly. If the user wants to inject `Opcode.TEXT` then the string that is entered in the UI needs to be literally used without `escaped_str_to_bytes`.

> Would that generally satisfy your use case? :)

Yes, but I want to clarify that this has nothing to do with my use-case. Injection via an API currently treats backslashes in an unexpected way. E.g. in the example https://github.com/mitmproxy/mitmproxy/blob/f135be8e65de7564d628a1535bd198139a153572/examples/addons/websocket-inject-message.py#L17 if `ssssssh` would contain backslashes it would cause subtle bugs because mitmproxy wouldn't forward the string exactly as is. Especially since `escaped_str_to_bytes` is applied regardless of the `is_text` parameter.

I hope that clears things up!

One more thing: I don't know how useful it actually is to have users manually compose escaped strings to inject bytes. Is there anything similar in mitmproxy already? Maybe it would be more useful to accept a hex dump or similar that users edit via an external editor. Or the UI only allows injection of ws `Opcode.TEXT` and for everything else (ws binary and tcp) we refer to the addon docs.

mhils:
> Disclaimer: I have never used commands in mitmproxy at all, so I might misunderstand something regarding how they are used.

The idea is that addons expose useful functions as commands, and users can directly invoke these functions.
For example, you can fire up mitmproxy, press `:` and type `replay.client @all` to replay all flows.

The replay addon only defines a regular Python function that takes a flow:

https://github.com/mitmproxy/mitmproxy/blob/f135be8e65de7564d628a1535bd198139a153572/mitmproxy/addons/clientplayback.py#L198-L199

and mitmproxy does the `@all` -> actual list of flows conversion in the background. There are a bunch of different argument types (int, bool, str, flow, flows, ...), so having commands that accept bytes also make perfect sense. We previously just didn't have a conversion function to convert user input into bytes.


I think this makes most sense if you just try it out. 

Check out this branch and the run the following commands in three console windows:
```
nc -l -p 8082
mitmproxy --mode reverse:localhost:8082 --tcp-hosts .+
nc localhost 8080
```
Now you can talk in windows 1 and 3 by typing text and pressing enter, with everything being logged by mitmproxy in the middle. We can not try the injection here: Press `:`, then type `inject.tcp @focus true hello` and you should see the hello message in window 3. Next up, type `:` `↑` to recall the last command, and then change `hello` to `foo\nbar`, `"foo bar\nbaz"`, or `\x07` and turn on sound 😄

At the same time, the current PR allows you to call `inect.tcp` manually with raw bytes as the message argument. Let me know if that makes sense now! 😃

Prinzhorn:
Thanks for the explanation. I'll definitely play with the commands and this branch in the next few days! There's this one WebSocket thing I wanted to test for a while now but couldn't because of the injection. If you don't hear from me before you want to make the release please ping me, I need it.

Prinzhorn:
So the backslashes still seem to haunt me :sweat_smile: 

What the client actually sends

```
["{\"method\":\"updateRoomCount\",\"data\":{\"model_name\":\"mitmproxy\",\"private_room\":false}}"]
```

what mitmproxy renders in the WebSocket tab (maybe for Opcode.TEXT we can render the original message?)

```
["{\\"method\\":\\"updateRoomCount\\",\\"data\\":{\\"model_name\\":\\"mitmproxy\\",\\"private_room\\":false}}"]
```

do I need _another_ layer of backslashes for the command? Both of the following result in the server closing the connection, so the message must arrive malformed:

```
inject.websocket @focus false '["{\"method\":\"updateRoomCount\",\"data\":{\"model_name\":\"mitmproxy\",\"private_room\":false}}"]'

inject.websocket @focus false '["{\\"method\\":\\"updateRoomCount\\",\\"data\\":{\\"model_name\\":\\"mitmproxy\\",\\"private_room\\":false}}"]'
```

Alright, so even three backslashes will be completely removed. The following injects what I actually need:

```
inject.websocket @focus false '["{\\\\"method\\\\":\\\\"updateRoomCount\\\\",\\\\"data\\\\":{\\\\"model_name\\\\":\\\\"mitmproxy\\\\",\\\\"private_room\\\\":false}}"]'
```

Pls also adjust https://github.com/mitmproxy/mitmproxy/blob/main/examples/addons/websocket-inject-message.py

Prinzhorn:
Anyway, the previous comment was more of a mental note. I'm having a more serious issue, some sort of deadlock :sweat_smile: 
So I'm injecting a message but it doesn't arrive. And from that point on the WebSocket is stalling, but it's still open. Nothing will come through.

I've done some `print()` based debugging and there doesn't seem to be an Exception. It does reach `server_event` and does the following command:

```
mitmdump stdout: WebsocketMessageHook(flow=<HTTPFlow
  request = Request(GET echo.websocket.org:443/?encoding=text)
  response = Response(101, no content)
  websocket = <WebSocketData (3 messages)>
  client_conn = Client([::1]:38502, state=open, tls)
  server_conn = Server(echo.websocket.org:443, state=open, tls, src_port=37964)>)
```

that means the following line does execute but it doesn an ooopsie I guess?

```
asyncio_utils.create_task(
    self.hook_task(command),
    name=f"handle_hook({command.name})",
    client=self.client.peername,
)
```

I wasn't able to reproduce this is isolation. In my setup currently all hooks are `@concurrent` and I'm calling the injection command from inside a gRPC method implementation (which I guess runs in yet another thread?)

I'll look into this some more next week. There is definitely something funny going on.

mhils:
> what mitmproxy renders in the WebSocket tab

Ok, so I agree this is a rendering problem; but not an injection problem?

> Alright, so even three backslashes will be completely removed. The following injects what I actually need:

Injection from commands does what I would expect it to do:

```
inject.tcp @focus true foo 
```
injects `foo`.

```
inject.tcp @focus true \x07 
```
injects a bell character.


```
inject.tcp @focus true \\wat 
```
injects `\wat`.


> I'll look into this some more next week. There is definitely something funny going on.

`asyncio_utils.create_task` may need to be called on the same asyncio thread. `@concurrent` breaks that. If that's not the issue, `proxy_debug` logs FTW. 😋

Prinzhorn:
> Ok, so I agree this is a rendering problem; but not an injection problem?

Yes, but I like to get side-tracked :smile: . I opened https://github.com/mitmproxy/mitmproxy/issues/4662 for that

> Injection from commands does what I would expect it to do:

I think the problem is my lack of understanding what `bytes_to_escaped_str` / `escaped_str_to_bytes` actually does.

Say I'm a mitmproxy user and I want to inject `\"` (a backslash followed by a double quite). Yes, this is a fun edge case but one I'm seeing in the wild in the first WebSocket I ever wanted to inject something into, so yeah.

First I naively try this

```
inject.websocket @focus true \"
```

but this raises (I don't know when the argument to a command needs to be quote, it doesn't contain a space). Ok I try this

```
inject.websocket @focus true '\"'
```

which injects a single double quote. Ok, so I escape the backslash

```
inject.websocket @focus true '\\"'
```

now it _still_ injects a double quote and nothing else. Alright, keep going:

```
inject.websocket @focus true '\\\"'
```

ok, still only a double quote.

```
inject.websocket @focus true '\\\\"'
```

there we go :smile:

But I guess the problem is that first there is regular python string escaping when you parse the command argument and then there's an additional round of `escaped_str_to_bytes`? But since the first three command calls do _exactly_ the same I feel something is not right?

> `asyncio_utils.create_task` may need to be called on the same asyncio thread. `@concurrent` breaks that. If that's not the issue, `proxy_debug` logs FTW. yum

Turns out grpc does support asyncio as well and it should be easy for me to migrate. But yeah not luck so far. But that's the way to go, I wasn't planning on using `@concurrent` in production, I just wanted to see the performance difference.

Prinzhorn:
Alright, I was able to successfully inject things the way I need them now via the API. Deadlock is dead.

There are three things open for me

1. The backslash "issue" I've described above. If you think this works as intended that's fine for me. We might improve the UX at some later point (e.g. have an actual UI for injecting, where you can insert the TEXT message 1:1).
2. Please update the example to use the new API https://github.com/mitmproxy/mitmproxy/blob/main/examples/addons/websocket-inject-message.py
3. We've discussed an `injected` flag before which lead to some lengthy discussion about consolidating all our flags. I'm fine with shipping without it but it's definitely needed. Since the API hides the creation of the `websocket.WebSocketMessage` instance it's impossible for me to use `metadata` or anything to indicate it's an injected message. This will lead to fun bugs and possibly infinite loops, but like I said we can also ship this after the initial v7.

mhils:
Thanks - the backslash example was really helpful. We incorrectly double-unescape for strings delimited with quotes, whereas we correctly escape only once for strings that are not enclosed in quotes. This is something we should fix. :)

mhils:
@Prinzhorn: new version out for you to test :) We now process escape codes later, which should provide a unified experience between quoted and unquoted strings.

Prinzhorn:
Seems to work as expected :rocket:

test/mitmproxy/test_websocket.py
```diff
@@ -26,3 +28,18 @@ def test_basic(self):
         assert not m.killed
         m.kill()
         assert m.killed
+
+    def test_text(self):
+        txt = websocket.WebSocketMessage(Opcode.TEXT, True, b"foo")
+        bin = websocket.WebSocketMessage(Opcode.BINARY, True, b"foo")
+
+        assert txt.is_text
+        assert txt.text
```
Prinzhorn, line 37: I think you meant to compare the `text` property with something here

Prinzhorn: Or maybe you just wanted to see if it raises, but might as well `== 'foo'` to check the conversion?

mhils: Yes, exactly.

PR #4670, opened 2021-07-08
Command Editor

mhils:
This PR contains @gorogoroumaru's work on the mitmweb command input. Posting it as a PR here so that I can comment easier. 😃


mitmproxy/tools/web/app.py
```diff
@@ -453,6 +453,15 @@ def get(self, flow_id, message, content_view):
         ))
 
 
+class Commands(RequestHandler):
+    def post(self):
+        result = self.master.commands.execute(self.json["command"])
+        if result is None:
+            self.write({"result": ""})
+            return
+        self.write({"result": str(result)})
```
mhils, line 462: Is there a reason why we `str` the result? It would be nice to eventually support more datatypes here, so we want to generally aim for arbitrary JSON.

mitmproxy/tools/web/app.py
```diff
@@ -470,6 +479,16 @@ def put(self):
...
 
 
+class CommandArguments(RequestHandler):
+    def get(self):
+        arguments = {}
+        for (name, command) in self.master.commands.commands.items():
+            arguments[name] = []
+            for parameter in command.parameters:
+                arguments[name].append(parameter.name)
+        self.write(arguments)
```
mhils, line 489: This looks good overall. Some ideas:

1. We can move this method in the `Commands` class so that we only need one route. (`GET /commands` to access the available options, `POST /commands` to issue one)
2. Let's append `{"name": parameter.name}` instead of `parameter.name`. We may want to add the parameter type here at some point.
3. `arguments` should probably be named `commands`?

web/src/js/__tests__/components/Command/CommandSpec.js
```diff
@@ -0,0 +1,13 @@
...
+
+describe('CommandBar Component', () => {
+    let commandBar = renderer.create(
+            <CommandBar />),
+        tree = commandBar.toJSON()
+
+    it('should render correctly', () => {
+        expect(tree).toMatchSnapshot()
+    })
+})
```
mhils, line 13: Let's please use react testing library and not `react-test-renderer` for all new tests. Here is a simple example:

https://github.com/mitmproxy/mitmproxy/blob/9b119c3dac3e57b29efd262edbb389e24a3d0214/web/src/js/__tests__/components/Header/MenuToggleSpec.js#L33-L43

mhils (commented):
Really nice prototype! PEG Grammar looks very nice, good work on that. TypeScript Implementation also looks really clean.

#### High level thoughts

 - For `GET /commands` (currently `CommandArguments`), we should also include the command's help description and the command history as it is kept by mitmproxy itself (see the `commands.history.get` command).
 - It would be really nice to have autocompletion. For example, if I type `view.` in mitmproxy's command prompt and then press the tab key repeatedly, I can cycle through `view.clear`, `view.clear_unmarked`, ... and so on.
 - We need some way to display all available commands.
 - In the long term we want something better to present returned values. For example, mitmproxy's console interface opens sort of a popup with a table if necessary.
- The command help should prefix-match, e.g. when I type `scri.`, it should already infer that I want to run `script.run` and display the apporiate help.

#### Immediate Next Steps

I think a good next step would be to transfer the "Argument suggestion" part into a Popper popup (similar to the search/highlight) filters at the top) so that we have more space there. We can then do the following:

1. If the command name is not fully typed in yet, display all available candidates starting with the current prefix. If the input is empty, just display all commands.
2. When the command name is complete, display the command help string.
3. When typing arguments, highlight the current argument in the help string.

mitmproxy/tools/web/app.py
gorogoroumaru: If we don't `str` the result, the state in CommandBar component doesn't store the result. I don't know the reason of this behavior.

mitmproxy/tools/web/app.py
gorogoroumaru: Done!

mitmproxy/tools/web/app.py
mhils: This works fine for me for everything that is json-serializable, e.g. `view.order.options` executes fine. It fails for flows, e.g. `view.flows.resolve`:
```
  File "c:\python39\lib\site-packages\tornado\web.py", line 1702, in _execute                                           
    result = method(*self.path_args, **self.path_kwargs)                                                                
  File "c:\users\user\git\mitmproxy\mitmproxy\tools\web\app.py", line 462, in post                                      
    self.write({"result": result})                                                                                      
  File "c:\users\user\git\mitmproxy\mitmproxy\tools\web\app.py", line 158, in write                                     
    super().write(chunk)                                                                                                
  File "c:\python39\lib\site-packages\tornado\web.py", line 843, in write                                               
    chunk = escape.json_encode(chunk)                                                                                   
  File "c:\python39\lib\site-packages\tornado\escape.py", line 75, in json_encode                                       
    return json.dumps(value).replace("</", "<\\/")                                                                      
  File "c:\python39\lib\json\__init__.py", line 231, in dumps    return _default_encoder.encode(obj)                    
  File "c:\python39\lib\json\encoder.py", line 199, in encode    chunks = self.iterencode(o, _one_shot=True)            
  File "c:\python39\lib\json\encoder.py", line 257, in iterencode                                                       
    return _iterencode(o, 0)                                                                                            
  File "c:\python39\lib\json\encoder.py", line 179, in default                                                          
    raise TypeError(f'Object of type {o.__class__.__name__} 'TypeError: Object of type HTTPFlow is not JSON serializable
```

Long story short, I think it's best if we return a `{value, type}` object here. For example:
```python
# a string is returned, for example from `view.order`
{
  value: "foo"
  type: "str",
}
# nothing is returned, for example from `view.clear`
{
  value: null
  type: "None",
}
# a list of flows is returned, for example from `view.flows.resolve @all`
{
  value: [flowId1, flowId2, ...],
  type: "list[flow]",
}
# a list of strings is returned, for example from flow.encode.options
{
  value: ["foo", "bar"],
  type: "list[str]",
}
```

Does that make sense?

gorogoroumaru: Sorry, it was a bug in my CommandBar component.
It didn't display the result when command returned 0.
Now, I fixed it and applied your suggested change.

gorogoroumaru:
I implemented a  tab completion.
But when we press a tab multiple times, the focused field changes and we are not on the command bar.
What should I do?

mhils:
This continues to look better and better! For the tab completion, I think we can listen to keydown events and call `e.preventDefault()`? I have not tried it, but I think that should work.

gorogoroumaru:
Thank you!
This is the screenshot of the new version.
image

mhils:
Looks excellent! 😃 Do you want to move the "available commands" part into the popover as well? We can display the available commands at first, and then switch to the command one it has been entered completely.

mitmproxy/tools/web/app.py
```diff
@@ -453,6 +451,27 @@ def get(self, flow_id, message, content_view):
...
+        commands = {}
+        for (name, command) in self.master.commands.commands.items():
+            commands[name] = {
+                "args": [],
+                "signature_help": command.signature_help(),
+                "description": command.help
+            }
+            for parameter in command.parameters:
+                commands[name]["args"].append(parameter.name)
+        self.write({"commands": commands})
```
mhils, line 465: Can we add tests for this?

web/src/js/components/ContentView/ContentLoader.tsx
```diff
@@ -45,6 +50,7 @@ export default function withContentLoader(View) {
 
         updateContent(props) {
             if (this.state.request) {
+                console.log("request:",this.state.request)
```
mhils, line 53: leftover :)

web/src/js/components/FlowView/Details.tsx
```diff
@@ -19,7 +26,19 @@ export function TimeStamp({t, deltaTo, title}) {
     )
 }
 
-export function ConnectionInfo({conn}) {
+type ConnectionInfoProps = {
```
mhils, line 29: I think we should use https://github.com/mitmproxy/mitmproxy/blob/29997bca4b6436e5837c7d086d3311d6e79e9ca4/web/src/js/flow.ts#L35 here.

web/src/js/contrib/CodeMirror.tsx
```diff
@@ -52,7 +52,7 @@ export default class CodeMirror extends React.Component<CodeMirrorProps, CodeMir
     }
 
     UNSAFE_componentWillMount() {
-        this.componentWillReceiveProps = _.debounce(this.componentWillReceiveProps, 0);
+        //this.componentWillReceiveProps = _.debounce(this.componentWillReceiveProps, 0);
```
mhils, line 55: intentional change?

web/src/js/ducks/options.ts
```diff
@@ -36,27 +36,27 @@ const reducer: Reducer<OptionsState> = (state = defaultState, action) => {
...
-export function pureSendUpdate(option: Option, value, dispatch) {
+export function pureSendUpdate(option: Option, value) {
     return async dispatch => {
         try {
             const response = await fetchApi.put('/options', {[option]: value});
             if (response.status === 200) {
                 dispatch(optionsEditorActions.updateSuccess(option))
             } else {
-                throw await response.text()
+                throw await response.text
```
mhils, line 46: Is this right?

web/src/js/ducks/options.ts
```diff
@@ -36,27 +36,27 @@ const reducer: Reducer<OptionsState> = (state = defaultState, action) => {
...
         } catch (error) {
             return dispatch(optionsEditorActions.updateError(option, error))
         }
     }
 }
 
-let sendUpdate = _.throttle(pureSendUpdate, 500, {leading: true, trailing: true})
-
 export function update(name: Option, value: any): AppThunk {
+    let sendUpdate = _.throttle(pureSendUpdate(name, value), 500, {leading: true, trailing: true})
```
mhils, line 55: This looks wrong. If we reconstruct `_.throttle` on every invocation, it's not throttling.

mhils (commented):
PR looks excellent. Getting rid of all the `connect()` calls is such a quality-of-life improvement. 🍰 Eventually we should also apply proper typescript to all ducks reducers, but that's not an urgent priority. Some smaller comments below. :)

web/src/js/ducks/options.ts
gorogoroumaru: I see.
But we need option name and value to construct sendUpdate.
How can I get them?

web/src/js/contrib/CodeMirror.tsx
gorogoroumaru: Sorry, I forgot to clean up my debug.

web/src/js/components/FlowView/Details.tsx
gorogoroumaru: The Connection interface didn't contain some fields used in ConnectionInfo component.
I added them to the Connection interface.

web/src/js/components/ContentView/ContentLoader.tsx
gorogoroumaru: Done

mitmproxy/tools/web/app.py
gorogoroumaru: In what file, should I add tests?

web/src/js/ducks/options.ts
mhils: Maybe I'm not understanding the problem. What was wrong with the implementation before your changes?

web/src/js/ducks/options.ts
gorogoroumaru: Sorry, I was doing wrong.
I fixed this.

web/src/js/ducks/options.ts
gorogoroumaru: Sorry, I was doing wrong.
I fixed this.

gorogoroumaru: The following is the previous code.
It's passing dispatch as argument, but this is not corresponding to dispatch argument in the returned function.
So, dispatch function becomes undefined.
When we type in the input field, it doesn't execute dispatch and "intercept" state doesn't change.
Thus, we can't intercept any websites.

`export function pureSendUpdate(option: Option, value, dispatch) {
    return async dispatch => {
        try {
            const response = await fetchApi.put('/options', {[option]: value});
            if (response.status === 200) {
                dispatch(optionsEditorActions.updateSuccess(option))
            } else {
                throw await response.text()
            }
        } catch (error) {
            return dispatch(optionsEditorActions.updateError(option, error))
        }
    }
}`

PR #4672, opened 2021-07-09
fix forward empty ws messages

Prinzhorn:
#### Description

tests are dlc and available soon

#### Checklist

### DIFF
--- a/examples/addons/websocket-inject-message.py
+++ b/examples/addons/websocket-inject-message.py
@@ -10,10 +10,11 @@ from mitmproxy import ctx, http
 
 # Simple example: Inject a message as a response to an event
 
-def websocket_message(flow):
+def websocket_message(flow: http.HTTPFlow):
+    assert flow.websocket is not None  # make type checker happy
     last_message = flow.websocket.messages[-1]
     if b"secret" in last_message.content:
-        last_message.kill()
+        last_message.drop()
         ctx.master.commands.call("inject.websocket", flow, last_message.from_client, "ssssssh")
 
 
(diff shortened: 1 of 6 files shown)


Prinzhorn:
I'm confused since you explicitly wrote a fragmentizer test for empty messages. Am I missing something? I'm seeing them in the wild and other projects seem to support them.

mhils:
Honestly I did not consider that you can (or would want to) send empty WebSocket frames. If this happens in practice that's of course a bug we need to fix. 👍 
Setting `message.content` to `b""` used to be the only way to kill a message until someone smart intervened in #4486 to reintroduce `.killed`. So this seems to be a leftover we just need to eliminate. 😃 

https://github.com/mitmproxy/mitmproxy/blob/a51dc10d8b97d46c77d11a4bbb4d0c393b2c8a39/examples/addons/websocket-simple.py#L19-L21 still has the old way as an example, maybe you can change that to `kill()` in this PR as well please?

Prinzhorn:
> Honestly I did not consider that you can (or would want to) send empty WebSocket frames. If this happens in practice that's of course a bug we need to fix. +1

I think this is mostly people not aware of ping/pong that use it to keep the TCP connection alive. I also encountered it right before the server did an abnormal close.
If the spec and all implementations allow it we should also.

> until someone smart intervened in

You should hire that person

> maybe you can change that to `kill()` in this PR as well please?

Yes and I'd like to rename it to `drop` while I'm at it. We're drastically changing APIs anyway.

PR #4674, opened 2021-07-12
Document what a websocket_error is

Prinzhorn:
Honestly I don't know how I feel about `websocket_error`. Before digging into the code I was expecting this to be an actual error (e.g. the TCP connection dropped unexpectedly or there was a malformed packet).

Why 1000, 1001 or 1005 in particular? Do you have a source why these three are considered success? The entire 4000 - 4999 range is "Available for use by applications" but considered an error by mitmproxy.

Maybe I'm also misunderstanding the use of `WebsocketErrorHook` in websocket.py vs eventsequence.py

Right now this is literally my addon code

```py
  def websocket_error(self, flow: mitmproxy.http.HTTPFlow):
    self.websocket_end(flow)
```

#### Checklist

### DIFF
--- a/mitmproxy/proxy/layers/websocket.py
+++ b/mitmproxy/proxy/layers/websocket.py
@@ -47,7 +47,7 @@ class WebsocketEndHook(StartHook):
 @dataclass
 class WebsocketErrorHook(StartHook):
     """
-    A WebSocket connection has had an error.
+    A WebSocket connection has had an error (a close code other than 1000, 1001 or 1005).
 
     Every WebSocket flow will receive either a websocket_error or a websocket_end event, but not both.
     """


Prinzhorn:
via Slack:

> Taking this one step further, this sounds like we may want to merge websocket_error into websocket_end, and also tcp_error into tcp_end?

@mhils I think this type of discussion would be better on GitHub, because it gives the issues/PRs context for the future if someone searches for them.

I'm not sure about `tcp_error`. For `websocket_error` it appears that `flow.error` is currently only set because of a wrong `code`. Looking at wsproto there doesn't appear to be an "actual" error event but everything is abstracted in ws terminology. E.g. if the TCP connection is lost it is simply exposed as `ABNORMAL_CLOSURE ` (1006). Arguably mitmproxy should then not go ahead and split this into `end` and `error` with (arbitrary) semantics. I think that's something the addon's can handle themselves and we shouldn't set `flow.error` at all. The only downside is logging. Because abnormally closed websockets should appear in error logs, or not? I feel like the ws spec overdid it with the close codes when 99.9999% of use-cases only care about when it was closed at all and maybe if it was closed on purpose (by calling close() on either end) or because of an error.

As for `tpc_error` I lack the understanding. I think there it makes sense to expose the error event (which in case of ws is abstracted away because it's one layer above).

Summary:

1. keep `tpc_error`
2. remove `websocket_error`
3. move the current error handling to dumpery.py `websocket_end`

Does that make sense?

Prinzhorn:
A good comparison would be that for HTTP we don't trigger `error` for status codes like 500, because that's not an error in that sense. Same for ws codes. With the difference that in ws there is no actual error and everything is handled on that layer.

Prinzhorn:
superseded #4675

PR #4675, opened 2021-07-12
remove websocket_error event, fixes #4674

Prinzhorn:
#### Description

I think we need a filter for close_code now.

Original discussion in #4674

#### Checklist

### DIFF
--- a/test/mitmproxy/proxy/layers/test_websocket.py
+++ b/test/mitmproxy/proxy/layers/test_websocket.py
@@ -249,6 +249,9 @@ def test_close_disconnect(ws_testdata):
             >> reply()
             >> ConnectionClosed(tctx.client)
     )
+    # The \x03\xe8 above is code 1000 (normal closure).
+    # But 1006 (ABNORMAL_CLOSURE) is expected, because the connection was already closed.
+    assert flow.websocket.close_code == 1006
 
 
 def test_close_error(ws_testdata):
@@ -265,6 +268,7 @@ def test_close_error(ws_testdata):
             << websocket.WebsocketEndHook(flow)
             >> reply()
     )
+    assert flow.websocket.close_code == 4000
 
 
 def test_deflate(ws_testdata):
--- a/test/mitmproxy/proxy/layers/test_websocket.py
+++ b/test/mitmproxy/proxy/layers/test_websocket.py
@@ -254,7 +254,7 @@ def test_close_disconnect(ws_testdata):
     assert flow.websocket.close_code == 1006
 
 
-def test_close_error(ws_testdata):
+def test_close_code(ws_testdata):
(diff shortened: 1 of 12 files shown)


mhils:
Your argument in https://github.com/mitmproxy/mitmproxy/pull/4674#issuecomment-878297628 sounds good to me. This looks good to merge except for the missing coverage in `dumper.py`. Can you handle that soon-ish? Otherwise I'll take care of this so that we can release 7.0 this week :)

test/mitmproxy/proxy/layers/test_websocket.py
```diff
@@ -245,11 +245,10 @@ def test_close_disconnect(ws_testdata):
             << CloseConnection(tctx.server)
             << SendData(tctx.client, b"\x88\x02\x03\xe8")
             << CloseConnection(tctx.client)
-            << websocket.WebsocketErrorHook(flow)
+            << websocket.WebsocketEndHook(flow)
             >> reply()
             >> ConnectionClosed(tctx.client)
     )
-    assert "ABNORMAL_CLOSURE" in flow.error.msg
```
mhils, line 252: Can we please assert the code here?

test/mitmproxy/proxy/layers/test_websocket.py
```diff
@@ -263,10 +262,9 @@ def test_close_error(ws_testdata):
             << CloseConnection(tctx.server)
             << SendData(tctx.client, b"\x88\x02\x0f\xa0")
             << CloseConnection(tctx.client)
-            << websocket.WebsocketErrorHook(flow)
+            << websocket.WebsocketEndHook(flow)
             >> reply()
     )
-    assert "UNKNOWN_ERROR=4000" in flow.error.msg
```
mhils, line 269: and here?

mhils (approved):
Let's add the assertions and then please feel free to merge yourself. 😃

Prinzhorn:
@mhils please check my recent commit. Did I understand that correctly (I'm still having a hard time understanding playbooks)? I think `test_close_error` is more like `test_application_close_code` now, because we're no longer populating `flow.error` and hence this basically tests if wsproto supports arbitrary codes in the 4xxx range?

mhils:
Yes this looks good to me. `test_close_error` is making sure that we propagate the error code to the other side if one side sends a close. In contrast, `test_close_disconnect` makes sure that we send a disconnect if the peer completely disconnects without sending anything.

Feel free to change the code to something that's definitely an error or rename the function, either is fine with me :)

Prinzhorn:
Welp, first time merging a PR and I was expecting it to squash automatically :shrug:

mhils:
All good - you have to break an egg to make an omelet. :)

PR #4687, opened 2021-07-17
Allow no-op assignments to Server.address when connection open

SaladDais:
First of all, congrats on shipping 7.0! Lots of nice improvements and I'm seeing the benefits of sans-io in making my own mitmproxy integrations more testable :smile:

#### Description

Calling `Flow.set_state(state)` on a flow implicitly calls `Server.set_state()` if there's an existing server connection. In the `response` addon hook the `Flow`'s `server_conn.state == ConnectionState.OPEN`, which causes the `Server.address` mutation check to `raise` if you call `flow.set_state(state)`.

This works around that issue by checking if the `Server.address` attribute's value is actually being changed rather than just checking if it's being set.

#### Checklist

### DIFF
--- a/mitmproxy/connection.py
+++ b/mitmproxy/connection.py
@@ -291,8 +291,11 @@ class Server(Connection):
         return f"Server({human.format_address(self.address)}, state={self.state.name.lower()}{tls_state}{local_port})"
 
     def __setattr__(self, name, value):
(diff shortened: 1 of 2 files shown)


mhils:
Excellent, thanks for the great PR! 😃

PR #4688, opened 2021-07-17
Add option to fall back to checking CN field if SANs are missing

SaladDais:
#### Description

This optionally increases compatibility with upstream certificates generated by legacy upstreams which only specify the `CN` field.

https://github.com/mitmproxy/mitmproxy/commit/e768f5ba83dd92d380f3e03da5da6c68edb3f45b (correctly) disabled falling back to checking the `CN` field in the absence of `subjectAltName` extensions to match browsers' behaviour. I deal with upstream TLS servers that generate certificates with no `subjectAltName` extensions. The clients manually handle peer verification and only look at the `CN` field. Not ideal, but I'm not able to change either the client or the server.

To deal with that I'm currently monkeypatching the `X509_VERIFY_PARAM_set_hostflags` function in an addon to mask off the `X509_CHECK_FLAG_NEVER_CHECK_SUBJECT` flag, this is a configurable way to do the same without having to monkeypatch pyOpenSSL internals.

What would be the best way to make a test for this? Changing [`test/mitmproxy/net/data/verificationcerts/generate.py`](https://github.com/mitmproxy/mitmproxy/blob/main/test/mitmproxy/net/data/verificationcerts/generate.py) to also generate a cert with no SAN seems like a good start, but I don't see any existing testcases for upstream cert verification.

#### Checklist

### DIFF
--- a/mitmproxy/addons/tlsconfig.py
+++ b/mitmproxy/addons/tlsconfig.py
@@ -102,6 +102,12 @@ class TlsConfig:
             choices=[x.name for x in net_tls.Version],
             help=f"Set the maximum TLS version for server connections.",
         )
+        loader.add_option(
+            name="tls_allow_cn_fallback",
+            typespec=bool,
+            default=False,
+            help=f"Allow checking an upstream certificate's CN field if SANs are missing",
+        )
 
     def tls_clienthello(self, tls_clienthello: tls.ClientHelloData):
         conn_context = tls_clienthello.context
@@ -217,6 +223,7 @@ class TlsConfig:
             ca_pemfile=ctx.options.ssl_verify_upstream_trusted_ca,
             client_cert=client_cert,
             alpn_protos=server.alpn_offers,
+            allow_cn_fallback=ctx.options.tls_allow_cn_fallback,
         )
 
         tls_start.ssl_conn = SSL.Connection(ssl_ctx)
--- a/mitmproxy/net/tls.py
+++ b/mitmproxy/net/tls.py
@@ -138,6 +138,7 @@ def create_proxy_server_context(
         ca_pemfile: Optional[str],
         client_cert: Optional[str],
         alpn_protos: Optional[Sequence[bytes]],
+        allow_cn_fallback: bool,
 ) -> SSL.Context:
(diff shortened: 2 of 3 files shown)


mhils:
Thanks for the PR! Your code changes look excellent, but I'm not sure if this belongs into mitmproxy. The kind of TLS verification you are describing here is quite far away from best practices, and seems to be quite rare to me. I feel you for not being able to fix your endpoints, but from the mitmproxy perspective I would really like to keep the cyclomatic complexity of security-critical paths as minimal as possible. 

Could you instead keep this in an addon that does something like this?
```python
from mitmproxy.proxy.layers import tls
from OpenSSL import SSL

def tls_start_server(self, tls_start: tls.TlsStartData):
    context: SSL.Context = tls_start.ssl_conn._context
    param = SSL._lib.SSL_CTX_get0_param(context._context)
    SSL._lib.X509_VERIFY_PARAM_set_hostflags(param, SSL._lib.X509_CHECK_FLAG_NO_PARTIAL_WILDCARDS)
```

I haven't tested this, and it may require us to reorder addons [here](https://github.com/mitmproxy/mitmproxy/blob/2d866ce9919d4e4cd6347ebcaf8d45689cb0a4b3/mitmproxy/addons/__init__.py#L47-L57), but I'd be happy to do that if it helps.

SaladDais:
Oh rad, I hadn't realized there was a proper hook for that in 7.0, that's a lot more elegant!

Your snippet was pretty close to what I needed, I just had to tweak it to do `param = SSL._lib.SSL_get0_param(tls_start.ssl_conn._ssl)` due to OpenSSL copying the SSL context's verify params struct on connection creation https://github.com/openssl/openssl/blob/ca001524971ccd595bc0e9843611e6784adfc981/ssl/ssl_lib.c#L746, which happens before my `tls_start_server()` runs. Setting the flag on the SSL context after the connection had been created resulted in verification errors.

>it may require us to reorder addons here, but I'd be happy to do that if it helps.

I can't speak for anyone else, but this shouldn't be an issue for me. I already subclass `TlsConfig` directly and don't use the standard addon list due to how I'm embedding mitmproxy.

Thank you for your guidance here!

PR #4691, opened 2021-07-19
Add WebSocketMessage.injected flag

Prinzhorn:
#### Description

I don't think we're going to migrate away from flags anytime soon, so let's add the `injected` flag. These changes wor as expected for me. However:

1. There are currently no tests for `inject.websocket`, I might need some assistance
2. I think this needs a migration for the flow format, right? Because WebSocketMessage serializable changed? Not even sure where to look for them, but I wanted to have this work now :smile: 
3. I was confused for a minute why my injected flag was overridden. Couldn't `relay_messages` early exit if `isinstance(event, WebSocketMessageInjected)` and just clone the WebSocketMessage instead of going through all that logic again? E.g. why involve `Fragmentizer` at all for injected messages?

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Release History
 
+## Unreleased: mitmproxy next
+
+* Allow no-op assignments to `Server.address` when connection open (@SaladDais)
+* Added `WebSocketMessage.injected` flag (@Prinzhorn)
+* --- TODO: add new PRs above this line ---
+
 ## 16 July 2021: mitmproxy 7.0
 
 ### New Proxy Core (@mhils, [blog post](https://www.mitmproxy.org/posts/releases/mitmproxy7/))
--- a/mitmproxy/addons/proxyserver.py
+++ b/mitmproxy/addons/proxyserver.py
@@ -197,7 +197,8 @@ class Proxyserver:
         msg = websocket.WebSocketMessage(
             Opcode.TEXT if is_text else Opcode.BINARY,
             not to_client,
-            message
+            message,
+            injected=True
         )
         event = WebSocketMessageInjected(flow, msg)
         try:
--- a/mitmproxy/proxy/layers/websocket.py
+++ b/mitmproxy/proxy/layers/websocket.py
@@ -125,8 +125,10 @@ class WebsocketLayer(layer.Layer):
 
         if isinstance(event, events.ConnectionEvent):
             from_client = event.connection == self.context.client
+            injected = False
         elif isinstance(event, WebSocketMessageInjected):
             from_client = event.message.from_client
+            injected = True
         else:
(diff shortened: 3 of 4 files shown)


Prinzhorn:
I found and updated the tests.

I'm not 100% confident in the compat tests. There don't seem to be unit tests (I added 14_15 test), only integration tests. I haven't actually tested it with a flow dump.

test/mitmproxy/io/test_compat.py
```diff
@@ -24,3 +24,28 @@ def test_cannot_convert(tdata):
         flow_reader = io.FlowReader(f)
         with pytest.raises(exceptions.FlowReadException):
             list(flow_reader.stream())
+
+
+def test_convert_14_15(tdata):
```
mhils, line 29: This test neatly obscures that `convert_14_15` is never called for the integration tests as you did not bump `mitmproxy.version.FLOWFORMAT_VERSION`. If you bump it to 15, the dumpfile-7-websocket test will fail because websocket messages are currently stored as tuples and not as dicts.
Long story short, I'm strongly in favor of removing the unit test here, we need to have full coverage by integration tests only.

mhils (changes requested):
Thanks! Looks good, but doesn't work yet. 😁 

Feel free to add a comment in `compat.py` that points to `version.py`. I'm also open for mechanisms that automatically derive that somehow, anything is fine with me. :)

test/mitmproxy/io/test_compat.py
Prinzhorn: > we need to have full coverage by integration tests only.

Maybe I'm missing something, where are the actual integration tests? From what I see it looks like `test_compat.py` tests if the number of flows matches and that it doesn't raise. But I don't see tests for the actual logic of the migrations.

mhils: Integration is indeed a bit far-fetched, I just adopted your terminology from https://github.com/mitmproxy/mitmproxy/pull/4691#issuecomment-883352513. 😉 I guess my pragmatic point is that they do actually raise here. 😄

Prinzhorn: So how do I proceed? I think having unit tests for all migrations makes it much clearer what is happening. E.g. looking at something like `convert_9_10` it would help a lot to have a unit test so I can look at the before/after to understand what it does. Right now it appears that it's not tested at all what `convert_9_10` does, only that the migrations don't raise.

For `injected` in particular one issue is `dropped`, right? It was never migrated so that means the tuple is already missing something and I assume `dropped` will then be `None` after migrations? So maybe my migration should just push `False, False` at the end of the tuple?

Prinzhorn: Following the tradition of `dropped` maybe I shouldn't add compat for `injected` at all and we separate the whole testing situation from this PR?

PR #4710, opened 2021-07-29
Improve HTTP Stream Modification

mhils:
@cbenien: Here's an experimental patch that makes it possible to return multiple chunks from `.stream`, and also some final ones on messag end. Let me know what you think! 😃


cbenien:
Hi @mhils ,

Was away for the weekend, but I've just tested the patch and it's awesome! I can
* buffer multiple chunks and then merge them into a big one (this works by returning a zero-length Iterable)
* split chunks into multiple smaller chunks

The final chunk with zero length is also a nice touch, that allows to nicely terminate the processing and flush any remaining data from the buffer. 

Thanks a lot!!! This is super useful!

mhils:
Happy to hear that! This will be included in 7.0.1, which will hit in the next few days. :)

PR #4731, opened 2021-08-04
Improve rendering of size column

### LINKED ISSUE #4696: Improve Rendering of the Size Column
#### Problem Description

The size column currently renders really badly:

image

`….64k` clearly is not what the user is interested in, something like `204k` would be much more useful.

1. Access a file with a size that is larger than 10kb, but smaller than 1MB.

#### Proposed fix

In short, we want to have a string here with a maximum length of 5 characters so that it does not get truncated. My proposal would be to just round to the nearest integer, e.g. `204k` instead of `204.64k`.

The relevant code is here:
https://github.com/mitmproxy/mitmproxy/blob/e270399a3eba7b3212bf7325beaa50e159a7ca0a/mitmproxy/utils/human.py#L18-L27
and the accompanying tests are here:
https://github.com/mitmproxy/mitmproxy/blob/0ca458fd6475ee48728147f3b529467a75e912a4/test/mitmproxy/utils/test_human.py#L28-L34

Fixing this should be really straightforward, so I'm leaving this open for new contributors. 😃 

#### System Information

```
Mitmproxy: 8.0.0.dev (+15, commit e270399)
Python:    3.9.4
OpenSSL:   OpenSSL 1.1.1f  31 Mar 2020
Platform:  Linux-4.4.0-19041-Microsoft-x86_64-with-glibc2.31
```

aaron-tan:
Resolves #4696 

If there is any issue, please feel free to tell me and I will fix it.

### DIFF
--- a/mitmproxy/utils/human.py
+++ b/mitmproxy/utils/human.py
@@ -1,30 +1,34 @@
 import datetime
+import functools
 import ipaddress
 import time
-import functools
 import typing
 
-SIZE_TABLE = [
-    ("b", 1024 ** 0),
-    ("k", 1024 ** 1),
-    ("m", 1024 ** 2),
-    ("g", 1024 ** 3),
-    ("t", 1024 ** 4),
-]
 
-SIZE_UNITS = dict(SIZE_TABLE)
+SIZE_UNITS = {
+    "b": 1024 ** 0,
+    "k": 1024 ** 1,
+    "m": 1024 ** 2,
+    "g": 1024 ** 3,
+    "t": 1024 ** 4,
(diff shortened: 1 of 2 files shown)


mhils (approved):
Thanks for the PR, this looks excellent! 🍰 
I've taken the liberty to simplify the existing code a bit, but that was certainly not expected. Thanks again! 😄

aaron-tan:
No problem. Thank you for giving me the opportunity to contribute. 😄

mhils:
Thank you, please keep the contributions coming! 😄

PR #4741, opened 2021-08-08
case insensitive encoding

### LINKED ISSUE #4735: GZIP request body encoding falling back to raw
#### Problem Description
GZIP requests are not decoded correctly.
This only occurs in the request and not in the response. In the response I see the body as uncompromised gzip but not in the request. 

I created a own content view to test the gzip decompression:

```python
from typing import Optional
from mitmproxy import contentviews, flow
from mitmproxy import http
import gzip


class ViewGZIPCase(contentviews.View):
    name = "GZIP"

    def __call__(self, data, **metadata) -> contentviews.TViewResult:
        return "gzip", contentviews.format_text(gzip.decompress(data))

    def render_priority(
            self,
            data: bytes,
            *,
            content_type: Optional[str] = None,
            flow: Optional[flow.Flow] = None,
            http_message: Optional[http.Message] = None,
            **unknown_metadata,
    ) -> float:

        if http_message.headers["content-encoding"] == "GZIP":
            return 1

        else:
            return 0


view = ViewGZIPCase()


def load(l):
    contentviews.add(view)


def done():
    contentviews.remove(view)
```

##### request

with selected self-made view
Screenshot 2021-08-05 at 16 42 57

without any special view
Screenshot 2021-08-05 at 16 42 42

##### response

Mitmproxy standalone decompression and json parsing in the response but why not in the request?
Screenshot 2021-08-05 at 16 47 55

1. start Mitmproxy 
2. wait for requests ( iPhone 12 as client is connected)
3. open the request

#### System Information
Mitmproxy: 7.0.2
Python:    3.9.6
OpenSSL:   OpenSSL 1.1.1k  25 Mar 2021
Platform:  macOS-11.5.1-x86_64-i386-64bit

Mattwmaster58:
#### Description

In theory, fixes #4735. Disclaimer: I'm not very familiar with the codebase

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Release History
 
+## Unreleased: mitmproxy next
+
+* fix some responses not being decoded properly if the encoding was uppercase #4735 (@Mattwmaster58)
+
 ## 4 August 2021: mitmproxy 7.0.2
 
 * Fix a WebSocket crash introduced in 7.0.1 (@mhils)
--- a/mitmproxy/net/encoding.py
+++ b/mitmproxy/net/encoding.py
@@ -52,6 +52,7 @@ def decode(
     """
     if encoded is None:
         return None
+    encoding = encoding.lower()
 
     global _cache
     cached = (
@@ -67,7 +68,7 @@ def decode(
             decoded = custom_decode[encoding](encoded)
         except KeyError:
             decoded = codecs.decode(encoded, encoding, errors)  # type: ignore
-        if encoding in ("gzip", "deflate", "br", "zstd"):
+        if encoding in ("gzip", "deflate", "deflateraw", "br", "zstd"):
             _cache = CachedDecode(encoded, encoding, errors, decoded)
         return decoded
     except TypeError:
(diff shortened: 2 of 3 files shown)


Prinzhorn:
Nice, thanks for contributing. One could argue that this belong to the HTTP layer (https://github.com/mitmproxy/mitmproxy/blob/c4dd46c3db97eabedf14fc54a23dae8d167fdad2/mitmproxy/http.py#L369 and https://github.com/mitmproxy/mitmproxy/blob/c4dd46c3db97eabedf14fc54a23dae8d167fdad2/mitmproxy/http.py#L387) and is not the business of the `encoding` package. 

> In theory

You can add a test here https://github.com/mitmproxy/mitmproxy/blob/c4dd46c3db97eabedf14fc54a23dae8d167fdad2/test/mitmproxy/test_http.py#L938-L1030 

You can read here how to setup your local environment https://github.com/mitmproxy/mitmproxy/blob/main/CONTRIBUTING.md#development-setup This would also allow you to test this with the original iPhone setup and see if it works

mhils:
Thanks @Prinzhorn! I think this is fine though. Encodings don't really need to be case-sensitive. I'd prefer if we just call encoding=encoding.lower() after the initial `content is None` check, that simplifies things. And please also adjust deflatedRaw. :)

mhils (approved):
I've done the last final tweaking, thanks again! 🍰

Issue #4515, opened 2021-03-19
Console: Add Quick Help Bar


mhils:
Some of my favorite command line tools display some sort of quick help at the bottom of the screen:

<details>
<summary>nano</summary>
<img src="https://user-images.githubusercontent.com/1019198/111744014-d841fe00-888a-11eb-9b02-bd208e2d3bef.png">
</details>
<details>
<summary>htop</summary>
<img src="https://user-images.githubusercontent.com/1019198/111744065-e55eed00-888a-11eb-8cfb-c8dd79f28bdf.png">
</details>

I feel that mitmproxy would greatly benefit from some more discoverability as well. The help menu lists a long list of keybindings, many of which just aren't too relevant when you are starting out. How would you all feel about adding a help bar similar to htop/nano? In particular, we could make that view-dependent, so that you get meaningful key shortcut help in flows and options editor as well 😃 

Maybe someone wants to come up with a first prototype? Implementation should't be too hard. You need to get some familiarity with [urwid](http://urwid.org/), and then it's probably enough to wrap the footer [here](https://github.com/mitmproxy/mitmproxy/blob/c5bd7b1ae1eb7b6280b77cab14f173d18a211062/mitmproxy/tools/console/window.py#L137) in an urwid.Pile which also displays the quick help.
I'm happy to help if you get stuck somewhere!

alt-glitch:
@mhils: I'd like to work on this issue. Let me check urwid and window.py out :)

greatvovan:
Additional thoughts:

1. I think it would be nice to utilize the bottom line of the TUI. Currently it is used for commands and messages, maybe we could display the Help Bar when there are no commands and messages (in all conditions when currently the line is empty). This would save one line for data and make the UI lighter (in my opinion).
2. Should we target 80 character width for terminal window? Or it is possible to put a dynamic content into the Help Bar (meaning the wider the window, the more hints we can put)?

mhils:
@greatvovan: Good points. I think we can first start with a prototype that just appends itself at the bottom, that's easier to implement. It should be possible to get the current dimensions with urwid, at least we somehow do that to decide how to display the flow list. 😃

aaron-tan:
Hi, I'd be happy to have a crack at this. Could anyone help point me in the right direction of which part of the code I should look into to implement this? Any pointers would be appreciated.

mhils:
[...]

PR #4743, opened 2021-08-09
Implement quick help bar prototype

aaron-tan:
#### Description
A prototype of a quick help bar as mentioned in issue #4515.

At the moment, it looks similar to the quick help bar seen in nano. Also it is a fixed widget so it will remain at the footer for as long as the mitmproxy is open. Perhaps a future improvement would be to display it until the user presses a key? Or after a certain amount of time has passed?

Please feel free to provide feedback and thoughts, I'd be happy to hear them. :smiley: 

#### Checklist

### DIFF
--- a/mitmproxy/tools/console/window.py
+++ b/mitmproxy/tools/console/window.py
@@ -131,10 +131,30 @@ class WindowStack:
 class Window(urwid.Frame):
     def __init__(self, master):
         self.statusbar = statusbar.StatusBar(master)
+        self.quick_help = urwid.Text([
+            ("heading", u"?"),
+            u" Help ",
+            ("heading", u"k"),
+            u" Up ",
+            ("heading", u"j"),
+            u" Down ",
+            ("heading", u"h"),
+            u" Left ",
+            ("heading", u"l"),
+            u" Right ",
+            ("heading", u"q"),
+            u" Exit current view ",
+            ("heading", u"enter"),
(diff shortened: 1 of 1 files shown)


mhils:
Awesome, this looks like a great start! I'm really looking forward towards this. 

> Perhaps a future improvement would be to display it until the user presses a key? Or after a certain amount of time has passed?

Let's do the same as nano and keep it there permanently. We do need to figure out how to do this though, right now it's too crowded with three bars (statusbar, input prompt/alerts, help). We can potentially work a bit with color, but we need to get this down to fewer components.

Some ideas:

1. Merge current status bar and alerts, i.e. we make alerts overlay the status bar?
2. Merge status bar and help?
3. Merge alerts/prompt and help? i.e. don't show help when prompting, similar to what nano does here:

image


FWIW we can also take some inspiration from `htop`. Looks almost more polished than nano, I like to color accents:

image

I'm on the go at the moment and can't expand more, but please feel free to chime in with your own ideas. This will be really valuable to onboard new users.

mhils:
Ok, I think we have a great first proof-of-concept here, but we first need to figure out what we want the whole thing to look like eventually. Let's continue over at #4515 so that we have everything in a single space. I'll follow up there momentarily. :)

mhils:
Thank you again for the initial prototype, closing this as we have a quick help bar for a while now! 😃

PR #4754, opened 2021-08-14
Add more type hints

povilasb:
#### Description

Added extra type hints while experimenting with the code base.

### DIFF
--- a/mitmproxy/net/tls.py
+++ b/mitmproxy/net/tls.py
@@ -5,7 +5,7 @@ import threading
 from enum import Enum
 from functools import lru_cache
 from pathlib import Path
-from typing import Iterable, Callable, Optional, Tuple, List, Any, BinaryIO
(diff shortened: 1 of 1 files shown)


mitmproxy/net/tls.py
```diff
@@ -5,7 +5,7 @@
 from enum import Enum
 from functools import lru_cache
 from pathlib import Path
-from typing import Iterable, Callable, Optional, Tuple, List, Any, BinaryIO
+from typing import Iterable, Callable, Optional, Tuple, List, Any, BinaryIO, List
```
Kriechi, line 8:
```suggestion
from typing import Iterable, Callable, Optional, Tuple, List, Any, BinaryIO
```

List is already imported.

mhils (approved):

povilasb:
Was pretty busy these days, thanks for fixing it ;)

PR #4760, opened 2021-08-16
Add toggle view filter option

### LINKED ISSUE #4604: There's a toggle intercept shortcut ("I") but no toggle filter shortcut
To create a filter I use the "f" shortcut but there is no way to toggle it with a shortcut instead of removing the filter.

Can you do like you did with intercept with the "I" shortcut to toggle intercepts?

aaron-tan:
#### Description
Closes #4604
- Implemented a toggle filter shortcut
- Added some tests for it as well
#### Checklist

### DIFF
--- a/mitmproxy/addons/view.py
+++ b/mitmproxy/addons/view.py
@@ -167,7 +167,7 @@ class View(collections.abc.Sequence):
 
     def load(self, loader):
         loader.add_option(
-            "view_filter", typing.Optional[str], None,
+            "view_filter", typing.Optional[str], "",
             "Limit the view to matching flows."
         )
         loader.add_option(
@@ -182,6 +182,10 @@ class View(collections.abc.Sequence):
         loader.add_option(
             "console_focus_follow", bool, False,
             "Focus follows new flows."
+        )
+        loader.add_option(
+            "filter_active", bool, False,
+            "Toggle whether the view filter is enabled."
         )
 
     def store_count(self):
@@ -541,14 +545,10 @@ class View(collections.abc.Sequence):
     # Event handlers
     def configure(self, updated):
         if "view_filter" in updated:
-            filt = None
             if ctx.options.view_filter:
-                filt = flowfilter.parse(ctx.options.view_filter)
-                if not filt:
-                    raise exceptions.OptionsError(
-                        "Invalid interception filter: %s" % ctx.options.view_filter
-                    )
-            self.set_filter(filt)
+                ctx.options.filter_active = True
+            else:
+                ctx.options.filter_active = False
         if "view_order" in updated:
             if ctx.options.view_order not in self.orders:
                 raise exceptions.OptionsError(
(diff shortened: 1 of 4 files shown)


mitmproxy/addons/view.py
```diff
@@ -167,7 +167,7 @@ def __init__(self):
 
     def load(self, loader):
         loader.add_option(
-            "view_filter", typing.Optional[str], None,
+            "view_filter", typing.Optional[str], "",
```
mhils, line 170: Is there a reason why you changed the default?

mitmproxy/addons/view.py
```diff
@@ -183,6 +183,10 @@ def load(self, loader):
             "console_focus_follow", bool, False,
             "Focus follows new flows."
         )
+        loader.add_option(
+            "filter_active", bool, False,
```
mhils, line 187: Let's call this `view_filter_active` please so that it's symmetric with `intercept` and `intercept_active`.

mitmproxy/addons/view.py
```diff
@@ -541,14 +545,10 @@ def inbounds(self, index: int) -> bool:
     # Event handlers
     def configure(self, updated):
         if "view_filter" in updated:
-            filt = None
             if ctx.options.view_filter:
-                filt = flowfilter.parse(ctx.options.view_filter)
-                if not filt:
-                    raise exceptions.OptionsError(
-                        "Invalid interception filter: %s" % ctx.options.view_filter
-                    )
```
mhils, line 550: We should keep validating the filter when it is updated, this makes option validation more straightforward.

mhils (commented):
Thanks for the PR. The code changes here generally look uncontroversial, but I'm not so convinced that #4604 is something we should implement. I'm not aware of any other application that has filter table functionality but then also an additional checkbox to enable/disable that filter. Pausing interception intuitively makes sense to me, but I don't get why you want to toggle a filter. Is there a specific use case I am missing? @AdvisedlyAlburn?

mitmproxy/addons/view.py
aaron-tan: There was a bug where if I tried to toggle the filter on it would return an attribute error none type object. Hence why I changed it.
Screenshot_2021-08-20_00-08-40

mitmproxy/addons/view.py
aaron-tan: Ok, that makes sense. I'll wait until we've confirmed that we want this feature before changing this.

mitmproxy/addons/view.py
aaron-tan: I don't really get what you mean by it makes option validation more straightforward.

89z:
My use case it, without something like this, you have to press `f`, then manually delete any existing filter. This can be really annoying if you have to do it repeatedly, or if the current filter is long.

PR #4761, opened 2021-08-17
update ws injection example, fixes #4751

### LINKED ISSUE #4751: Cannot inject websocket with binary or string content.
#### Problem Description
mitmproxy attempts to parse the injected websocket binary data as text, and replaces the binary data if it cannot be decoded into text, changing the injected websocket message. Also, when a string is given instead, there is error:
File "\mitmproxy\proxy\layers\websocket.py", line 230, in msg
    data_str = data.decode(errors="replace")
AttributeError: 'str' object has no attribute 'decode'

1. from mitmproxy.ctx import master
2. master.commands.call("inject.websocket", flow, False, bytes([i for i in range(256)])) # No error, but message contents are changed
3. master.commands.call("inject.websocket", flow, False, "string") # AttributeError

#### System Information
Mitmproxy: 7.0.2
Python:    3.9.6
OpenSSL:   OpenSSL 1.1.1k  25 Mar 2021
Platform:  Windows-10-10.0.19043-SP0

Prinzhorn:


mhils (approved):

mhils:
Thanks! 🍰

Prinzhorn:
@mhils it's irrelevant for the example, but I purposely explicitly used `encode()` so that you can replace the example strings with unicode as well. E.g. now you cannot just do `msg = b"hellö from mitmproxy! "` because "SyntaxError: bytes can only contain ASCII literal characters.". And moving `encode` outside of the loop breaks the strings because `msg = msg[1:] + msg[:1]` would operate on bytes and break the utf-8.

I also thought using `encode()` would be more explicit and makes it more clear that utf-8 bytes are expected. It would be more closely aligned with real world, where `shhhhhhhhhh` wouldn't be hard coded but held as a string variable.

Yes, I know I always get lost in details :smile:

mhils:
Sold. :)

PR #4766, opened 2021-08-20
Update core.py

JmmBite:
When Press "S" to save options to a file, there is no need to always fill the path, and save to the default path: {confdir}/config.yaml

#### Checklist

### DIFF
--- a/mitmproxy/addons/core.py
+++ b/mitmproxy/addons/core.py
@@ -282,6 +282,8 @@ class Core:
             Save options to a file.
         """
         try:
+            if not path:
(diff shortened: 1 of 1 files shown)


mhils:
I think that's not a bad suggestion in general, but I've quickly pulled your changes and they don't seem to work? Arguments are not optional, I now get 
```
error: Command argument mismatch:
Expected: OrderedDict([('path', <Parameter "path: mitmproxy.types.Path">)])
Received: []
```
when pressing `S` and then `Enter`.

PR #4771, opened 2021-08-22
Feat/socks5auth

starplanet:
#### Description

add --socks5auth option to support userpassword authentication for socks5 mode

#### Checklist

### DIFF
--- a/mitmproxy/proxy/layers/modes.py
+++ b/mitmproxy/proxy/layers/modes.py
@@ -3,7 +3,7 @@ import struct
 from abc import ABCMeta
 from typing import Optional
 
-from mitmproxy import platform
+from mitmproxy import platform, ctx
 from mitmproxy.net import server_spec
 from mitmproxy.proxy import commands, events, layer
 from mitmproxy.proxy.layers import tls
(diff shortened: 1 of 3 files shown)


mhils:
Thanks for the PR! This looks excellent so far. As a next step, can you please add tests at https://github.com/mitmproxy/mitmproxy/blob/main/test/mitmproxy/proxy/layers/test_modes.py? Let me know if you need any help with that. I would also like to integrate this with the existing proxyauth addon, but I can take care of that.

starplanet:
I am happy to add some test. The idea that integrating this with the existing proxyauth, I have thought of that, but it seems to be more troublesome. Adding an extra option looks more easier.

mitmproxy/proxy/layers/modes.py
```diff
@@ -128,17 +141,47 @@ def _handle_event(self, event: events.Event) -> layer.CommandGenerator[None]:
...
+                if len(self.buf) < 3 + id_len:
+                    return
+                pw_len = self.buf[2 + id_len]
+                if len(self.buf) < 3 + id_len + pw_len:
+                    return
+                test_id, test_pw = self.buf[2:(2 + id_len)], self.buf[(3 + id_len):(3 + id_len + pw_len)]
+                if self.check_auth(test_id, test_pw):
+                    yield commands.SendData(self.context.client, b"\x01\x00")
+                else:
+                    yield from self.socks_err("authentication failed", 0x01, ver_code=0x01)
```
mhils, line 180: Why do you submit a custom `ver_code` here?
[RFC 1928](https://www.rfc-editor.org/rfc/rfc1928.html#section-6) always has this hardcoded to 0x05.

mhils: I think this is a misunderstanding of the RFCs:
```
                         +----+--------+
                         |VER | METHOD |
                         +----+--------+
                         | 1  |   1    |
                         +----+--------+
```

The 1 in this table only indicates that the version field is one byte long. If my hypothesis is correct please confirm. No need to fix it, I'm working on the code now anyways :)

Issue #4773, opened 2021-08-23
Allow to close WS connections


mario-purestake:
#### Problem Description
We want to close a websocket connection manually from the proxy addons. This is currently not possible, we only close connections when one of the sides has closed it previously (either server or client).

#### Proposal
Make available this functionality through the `inject.websocket` command, adding another option to change the Opcode we want to send. Currently, only `TEXT` and `BINARY` are supported.

#### Alternatives
None, but always open for more ideas :) 

#### Additional context

We want to use the proxy to redirect traffic on the end side of the connection, so in a `mode:reverse` configuration. We are receiving HTTP and WS connections. Redirecting HTTP requests is easy with this tool, we just need to change the `.host` property. But for WS is more difficult for our use-case, because we would like to redirect (websocket) messages to different upstreams depending on the data they bring. In order to do so, our current approach is to close the current WS connection and redirect it after the same user tries re-connecting again. This is why we would need the possibility to close a WS connection from the addons layer.

mhils:
refs #4711

mhils:
Yes, that's something we should support. :)

WilliamChiu:
Hi! Is this something you are still looking to support? I would love to help with this but might need some guidance

mario-purestake:
Hi, I opened this some time ago because I had an idea on how to implement it using the `inject.websocket` command, but it wasn't followed because the command was kind of complex already. The PR with the idea is linked if you want to take a look, but honestly too much time has passed and I'm not using the tool anymore

plongstaff:
I would be interested in having this functionality available.

PR #4774, opened 2021-08-23
Allow to close WS connections

mario-purestake:
#### Description

Addresses #4773.

Adds another option to the `inject` command to close the connection. This parameter is then managed to encode a CLOSE opcode into the flow that closes the WS connection with a user-defined reason (encoded as bytes).

#### Checklist

 - [ ] I have updated tests where applicable. (?) Didn't find any test related, I updated the docs examples though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Improve error message on TLS version mismatch.
 * Windows: Switch to Python's default asyncio event loop, which increases the number of sockets
   that can be processed simultaneously.
+* Allow to close WS connections using `inject.websocket` command
 
 ## 4 August 2021: mitmproxy 7.0.2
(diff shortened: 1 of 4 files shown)


Prinzhorn:
My two cents: I think the `inject.websocket` signature is already complex enough and I'd rather not mix in a new use-case.

mhils:
Thanks for the PR! I think I do agree with @Prinzhorn here though. What we really want is to extend the proxy core's injection mechanism so that we can't just inject messages, but also connection closes for all protocols. That then opens the door for dedicated `kill.websocket`/`kill.http` commands. Does that make sense?

mario-purestake:
Closing this PR because finally I didn't find another solution to force to close the WS connection

Issue #3884, opened 2020-04-02
New event after `clientconnect` as soon as TLS is established


Prinzhorn:
**Is your feature request related to a problem? Please describe.**
`layer.client_conn.timestamp_tls_setup` is `None` in `clientconnect`. I'm storing all mitmproxy data in SQLite and there is no event I can hook into to easily get `timestamp_tls_setup`. I could wait for `clientdisconnect` but I want to have the data as soon as it is available. Same for `cert` inside `serverconnect`.

#### Describe the solution you'd like
An event, maybe we can call it `clientconvert` analogous to `convert_to_tls` method. The event is triggered once a client connection is converted. So obviously not all client connections will be converted to TLS, it will never fire for them.

I'd also need a symmetrical event like `serverconvert` so that I can process `conn.cert` only once and not for every single `request`. Right now inside `serverconnect` the `cert` is `None`. My current workaround is to process the `cert` for every `request` but have a UNIQUE constraint on the cert sha256 and effectively only storing it once.

#### Describe alternatives you've considered
For HTTP data I can use the first `request` event for any given `client_conn` and process it as needed. But that requires additional logic or it will process stuff multiple times if a connection is shared.

#### Additional context
There's also `next_layer`. Could I use that? I'm not familiar enough with what a layer is and how that works. Even if I could use it I guess two new events would be user friendly and explicit.

mhils:
This is a good feature request. In fact, on the sans-io branch I already expose `tls_clienthello` (we just received a ClientHello) and `tls_start` events (we are about to establish TLS with someone). We could conceivably also add `tls_success` and `tls_error` events there.

Workaround for now: I believe what you should be able to do right now is to just update the TLS certificate on `clientdisconnect` and `serverdisconnect`. You should get all connections in their final state that way.


Finally, `next_layer`: [`mitmproxy/proxy/protocol/__init__.py`](https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/proxy/protocol/__init__.py#L1-L29) has a bit of an explanation. It's a very low-level hook to influence which protocol should be next on the stack. It could be abused for your use case (check for calls to next_layer where the current top layer is TLS), but that's nothing I would actively encourage. Using `*disconnect` seems much cleaner to me.

Prinzhorn:
[...]

PR #4790, opened 2021-09-03
TLS: add `tls_handshake` hook, ignore-after-clienthello

### LINKED ISSUE #4702: Make it possible to enable TLS passthrough from `tls_clienthello` hook.
From Slack:

> given that I cannot decrypt the connection, I am  currently attempting to bypass the TLS connections altogether.
by doing the following:
```
def next_layer(self, nextlayer: layer.NextLayer):
    is_tls = is_tls_record_magic(nextlayer.data_client())
    if(is_tls):
        nextlayer.layer = layers.TCPLayer(nextlayer.context, ignore=True)
```
> This allows the device to function without breaking due to TLS cert errors, however if I do this, the tls_clienthello hook does not fire. Inversely, if it fires, I do get the SNI but the connection fails.

----

We should look into what it takes to add a `tls_clienthello.passthrough: bool` attribute, which when turned on, makes us not intercept the connection. I have an idea what to do here and will look into it.

mhils:
This PR adds two new features to our TLS implementation:

1. A new `tls_handshake` hook, which gets fired after a TLS handshake has been completed (successfully or not). I'm not sure yet if the name is perfect, feedback is welcome.
2. Make it possible to stop TLS interception from the `tls_clienthello` event hook.

Additionally, I've revised the `tls_passthrough.py` script as an example for both. Now without monkeypatching. 😄 

To do:

 - [x] code review
 - [x] tests
 - [x] add docs

fixes #4702


Prinzhorn:
Neat, I don't have the mental juice right now to give it a full review though

1. Shouldn't there be a `_server` and a `_client` variant? There already is a hook with a `if isinstance(data.conn, connection.Server)` in this PR
2. This does fix #3884 as well, right? Or is that not the earliest possible hook to get the cert info?
3. Not sure how I feel about having the same hook for success and failure. But I don't know enough about handshakes and the use-cases of this hook to judge

docs/scripts/api-events.py
```diff
@@ -124,6 +124,7 @@ def category(name: str, desc: str, hooks: List[Type[hooks.Hook]]) -> None:
             tls.TlsClienthelloHook,
             tls.TlsStartClientHook,
             tls.TlsStartServerHook,
+            tls.TlsHandshakeHook,
```
Kriechi, line 127: Any proposals how to make the naming somehow include the "completed" idea of this hook?
`TlsHandshakeCompletedHook` or similar? Or using the `...Start/End...` scheme?

Kriechi:
Looks great! Only nitpicking the event name a bit 😄

89z:
https://github.com/mitmproxy/mitmproxy/issues/4877

mhils:
Summarizing the naming feedback:

> @Kriechi: Any proposals how to make the naming somehow include the "completed" idea of this hook? `TlsHandshakeCompletedHook` or similar? Or using the `...Start/End...` scheme?

The problem I'm having with _end_ is that the TLS connection is not over at this point, only the handshake has completed.

> @Prinzhorn: Shouldn't there be a `_server` and a `_client` variant? There already is a hook with a `if isinstance(data.conn, connection.Server)` in this PR

Yes, that makes sense. 👍

> @Prinzhorn: Not sure how I feel about having the same hook for success and failure. But I don't know enough about handshakes and the use-cases of this hook to judge. If it does solve 2. for me I'd only like successes, but sure I can check for those

Thank you for explicitly taking the API consumer point-of-view, that is good. Makes sense to me. 👍

How would you all feel about splitting `tls_handshake` into the following four event names:

- `tls_established_client`
- `tls_established_server`
- `tls_error_client`
- `tls_error_server`

Kriechi:
I like the `_established_` part! 

I know the `_error_` part follows existing patterns, though I feel like it doesn't match "language wise", a past-tense adjective.  How about `_established_` + `_failed_`?

mhils:
I have no strong preference. Let's invoke democracy:

- React with 🎉  to this comment if you prefer `tls_error_client`.
- React with 🚀 to this comment if you prefer `tls_failed_client`.

Prinzhorn:
I have `github.com##.js-reactions-container` in my uBlock Origin filters, so much for accessibility to your corrupt voting system :smile: 

> I have no strong preference.

Same, with a tendency for :rocket: because consistency within the `tls_*` category at least. Or what will happen with the existing three `tls_` hooks? They're all nouns, `tls_start_client` vs `tls_started_client` :sweat_smile:

mhils:
Available on main now: https://docs.mitmproxy.org/dev/api/events.html#TLSEvents.tls_established_client 😃

89z:
?

mhils:
https://docs.mitmproxy.org/dev/api/events.html#TLSEvents.tls_clienthello has all the relevant bits - sorry for the slowish reply, merry christmas!

89z:
Thanks for the info, but I dont know how that link works with the command line tools...

mhils:
You can use the addon API with all mitmproxy tools:
https://docs.mitmproxy.org/dev/addons-overview/
The previous link points to the `tls_clienthello` hook and describes the argument it receives.

PR #4805, opened 2021-09-14
Make Cert.not{before,after} timezone aware

matthewhughes934:
I noticed when running tests the output of
`web/src/js/__tests__/ducks/_tflow.ts` would change depending on how I
set my timezone, e.g.

    $ TZ=America/Los_Angeles pytest --quiet \
        test/mitmproxy/tools/web/test_app.py >/dev/null \
        && grep --extended-regexp 'not(after|before)' web/src/js/__tests__/ducks/_tflow.ts
                "notafter": 2235132207,
                "notbefore": 1604415807,
    $ TZ=Asia/Tokyo pytest --quiet \
        test/mitmproxy/tools/web/test_app.py >/dev/null \
        && grep --extended-regexp 'not(after|before)'  web/src/js/__tests__/ducks/_tflow.ts
                "notafter": 2235074607,
                "notbefore": 1604354607

It looks like this is because the `cert_to_json` function simply calls
`timestamp` the `datetime` object from
`x509.Certificate.not_valid_before`, however, this `datetime` object is
not timestamp aware, from the docs [1]:

> A naïve datetime representing the beginning of the validity period for
the certificate in UTC

So when serializing to JSON, first convert the `datetime` to UTC then
call `timestamp`.

A test was added by inspecting one of the test certs with:

    $ openssl x509 -in test/mitmproxy/net/data/text_cert_2 -text

Extracting the date and asserting on that.

The corresponding test has also been re-run so that `_tflow.ts` was
regenerated with it's correct value. Snapshots were also updated via:

    $(npm bin)/jest --updateSnapshot

[1] https://cryptography.io/en/latest/x509/reference/#cryptography.x509.Certificate.not_valid_after

#### Checklist

 - [x] I have added an entry to the CHANGELOG. (not sure if applicable)

### DIFF
--- a/mitmproxy/certs.py
+++ b/mitmproxy/certs.py
@@ -89,11 +89,13 @@ class Cert(serializable.Serializable):
 
     @property
     def notbefore(self) -> datetime.datetime:
-        return self._cert.not_valid_before
+        # x509.Certificate.not_valid_before is a naive datetime in UTC
+        return self._cert.not_valid_before.replace(tzinfo=datetime.timezone.utc)
 
     @property
     def notafter(self) -> datetime.datetime:
-        return self._cert.not_valid_after
+        # x509.Certificate.not_valid_after is a naive datetime in UTC
+        return self._cert.not_valid_after.replace(tzinfo=datetime.timezone.utc)
 
     def has_expired(self) -> bool:
         return datetime.datetime.utcnow() > self._cert.not_valid_after
--- a/test/mitmproxy/test_certs.py
+++ b/test/mitmproxy/test_certs.py
@@ -1,4 +1,5 @@
 import os
+from datetime import datetime, timezone
 from pathlib import Path
 from cryptography import x509
 from cryptography.x509 import NameOID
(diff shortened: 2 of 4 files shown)


matthewhughes934:
I guess I'll need to also run `$(npm bin)/jest --updateSnapshot` to update the snapshots (assuming my original change is desired)

mhils:
Thanks for the great write-up! 🍰 

This looks all good so far. I'm wondering whether we should just make `Cert.notbefore` return an UTC-aware datetime already, i.e. call `.replace(tzinfo=timezone.utc)` there? That would prevent similar problems in the future. 

I don't really understand why cryptography doesn't return UTC-aware datetimes in the first place. https://github.com/pyca/cryptography/pull/3467#discussion_r107181959 hints that this is a Python 2.7 relict perhaps? @alex / @reaperhulk, can you confirm this?

alex:
Yes, that's right. Unfortunately I'm not sure if we can change it now due
to backwards compatibility.

בתאריך יום ג׳, 14 בספט׳ 2021, 8:21, מאת Maximilian Hils ‏<
***@***.***>:

mhils:
Yes, totally understand that part. Thanks for the quick confirmation! 🍰

mhils:
@matthewhughes934: We can break compatibility in mitmproxy here, that is fine. Let's move the `.replace(tzinfo=timezone.utc)` call into `mitmproxy.certs` 😃

matthewhughes934:
> @matthewhughes934: We can break compatibility in mitmproxy here, that is fine. Let's move the `.replace(tzinfo=timezone.utc)` call into `mitmproxy.certs` smiley

:+1: done with aa5881b. Let me know if it's looking good and I'll tidy up the commit history a bit (i.e. squash them down and clean up the commit)

mhils:
Fantastic, thank you so much! 😃🍰 

Please don't worry about messy commit histories for PRs. Sometimes it's useful to just see the latest changes, we have this really convenient squash button here to fix things up. 😄

matthewhughes934:
> Please don't worry about messy commit histories for PRs. Sometimes it's useful to just see the latest changes, we have this 
really convenient squash button here to fix things up.

:+1: my concern is that I think GH just merges all the commits together to create the squash commit, and the body of the first commit wasn't really valid any more (it would also include my squash commit, which wouldn't be helpful for anyone to read in the future).

Issue #4824, opened 2021-09-24
MITM crash when modifying request with 'chunked' content encoding


matthewhughes934:
#### Problem Description

Traceback:

```
127.0.0.1:55064: mitmproxy has crashed!
<---snip--->
ValueError: Received both a Transfer-Encoding and a Content-Length header, refusing as recommended in RFC 7230 Section 3.3.3. See https://github.com/mitmproxy/mitmproxy/issues/4799 for detail
```

#### Steps to reproduce the behavior:

1. Run either mitmproxy or mitmweb configured to intercept requests from http://anglesharp.azurewebsites.net/Chunked (just a convenient site that should always return chunked responses)
2. Intercept a request from the above URL and modify the response in anyway, e.g. via

```
$ curl --output /dev/null --silent \
    --cacert ~/.mitmproxy/mitmproxy-ca-cert.pem \
    --proxy 127.0.0.1:8080 \
    http://anglesharp.azurewebsites.net/Chunked
```

3. Resume sending the response

#### System Information

Paste the output of "mitmproxy --version" here.

```
$ mitmproxy --version
Mitmproxy: 8.0.0.dev (+114, commit 7e24e77)
Python:    3.8.6
OpenSSL:   OpenSSL 1.1.1l  24 Aug 2021
Platform:  Linux-5.10.0-0.bpo.8-amd64-x86_64-with-glibc2.28
```

#### More Details

Following up from https://github.com/mitmproxy/mitmproxy/pull/4811#issuecomment-925696622.

At a glance, the issue looks to be at https://github.com/mitmproxy/mitmproxy/blob/main/mitmproxy/http.py#L375 where the `content-length` header is unconditionally set whenever content is updated. Perhaps this just needs a simple `if ce != "chunked"` guard? With another glance at [the spec](https://httpwg.org/specs/rfc7230.html#chunked.encoding), `chunked` encoding specifies the chunk size within the body of each chunk and I'm _assuming_ once we have a response all these chunks have already been decoded, so we don't have to worry about updating this for a modified response

mhils:
Thanks, this is an excellent catch. I think the guard you are proposing is one correct piece of the puzzle. The other part is that we should only validate inbound requests, it's fine if we send Content-Length + Transfer-Encoding upstream if a CL header is added by a user.

matthewhughes934:
> I think the guard you are proposing is one correct piece of the puzzle

[...]

PR #4827, opened 2021-09-26
Don't set 'content-length' with 'transfer-encoding'

matthewhughes934:
When updating the response content for a response, avoid adding the
'content-length' header if the response contains a 'transfer-encoding'
header, from the spec [1]:

> When a message does not have a Transfer-Encoding header field, a
Content-Length header field can provide the anticipated size, as a
decimal number of octets, for a potential payload body

Note the 'transfer-encoding' header is not used with HTTP/2

https://httpwg.org/specs/rfc7230.html#header.content-length

#### Checklist

### DIFF
--- a/mitmproxy/http.py
+++ b/mitmproxy/http.py
@@ -372,7 +372,13 @@ class Message(serializable.Serializable):
             # Let's remove it!
             del self.headers["content-encoding"]
             self.raw_content = value
-        self.headers["content-length"] = str(len(self.raw_content))
+
+        if "transfer-encoding" in self.headers:
+            # https://httpwg.org/specs/rfc7230.html#header.content-length
+            # don't set content-length if a transfer-encoding is provided
+            pass
+        else:
+            self.headers["content-length"] = str(len(self.raw_content))
 
     def get_content(self, strict: bool = True) -> Optional[bytes]:
         """
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 * Windows: Switch to Python's default asyncio event loop, which increases the number of sockets
   that can be processed simultaneously (@mhils)
 
+## 28 September 2021: mitmproxy 7.0.4
+
+* Do not add a Content-Length header for chunked HTTP/1 messages (@matthewhughes934)
+
 ## 16 September 2021: mitmproxy 7.0.3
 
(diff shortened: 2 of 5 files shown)


mitmproxy/http.py
```diff
@@ -372,7 +372,8 @@ def set_content(self, value: Optional[bytes]) -> None:
             # Let's remove it!
             del self.headers["content-encoding"]
             self.raw_content = value
-        self.headers["content-length"] = str(len(self.raw_content))
+        if not "transfer-encoding" in self.headers:
```
matthewhughes934, line 375: I'm still not sure if this should be: `if self.is_http2 or not "transfer-encoding" in self.headers:` since we probably still want to set `content-length` if (the unused/unsupported) `"transfer-encoding"` header is given in a HTTP/2 request. What do you think?

mhils:
Thanks for tackling this!

Let's copy this check here so that things are consistent: 
https://github.com/mitmproxy/mitmproxy/blob/0ca458fd6475ee48728147f3b529467a75e912a4/mitmproxy/net/http/http1/assemble.py#L30

A stylistic device I like to use for cases like this is to have an explicit `pass` case with documentation: That emphasizes the intent here and also makes sure that we have test coverage for both cases:

```python3
if "chunked" in headers.get("transfer-encoding", "").lower(): 
    pass  # http://tools.ietf.org/html/rfc7230#section-3.3: do not add a content-length header to a chunked message.
else:
    self.headers["content-length"] = str(len(self.raw_content))
```

mitmproxy/http.py
mhils: HTTP/2 does not have `transfer-encoding` headers, so I think it's really good enough to do the check below.

mhils:
I've pushed the second part for #4824 on top of your commit so that we have everything in one place. :)

matthewhughes934:
Thanks for the review!

> I've pushed the second part for #4824 on top of your commit so that we have everything in one place. :)

Perfect, thanks

> A stylistic device I like to use for cases like this is to have an explicit pass case with documentation

Good idea, done with c5036d795

> Let's copy this check here so that things are consistent:

I think these are separate checks: I said in the issue that this was only for 'chunked' transfer encodings, but I believe it's for any request with a `transfer-encoding` set

mhils:
Ah, yes, you are right. Thanks for checking this so thoroughly! 🍰

PR #4837, opened 2021-10-04
Fix  #4832

### LINKED ISSUE #4832: URL shortened when --flow-detail 1
#### Problem Description

`mitmdump --help` reads:

```
  --flow-detail LEVEL   The display detail level for flows in mitmdump: 0 (almost quiet) to 3 (very verbose). 0: shortened request URL, response status code, WebSocket
                        and TCP message notifications. 1: full request URL with response status code 2: 1 + HTTP headers 3: 2 + truncated response content, content of
                        WebSocket and TCP messages 4: 3 + nothing is truncated
```

However, the request URLs are shortened like `https://firebaselogging-pa.googleapis.com/v1/firelog/le…` when specifying `--flow-detail 1`.

1. Dump flows with `w` command in `mitmproxy`
2. `mitmdump --no-server --flow-detail 1 -r mitmproxy-flows.dump`

#### System Information

```
Mitmproxy: 7.0.4
Python:    3.9.7
Platform:  macOS-11.6-x86_64-i386-64bit
```

brunopcarv:
#### Description

- Do URL shortening for flow-detail < 1 instead of <= 1

#### Checklist

### DIFF
--- a/mitmproxy/addons/dumper.py
+++ b/mitmproxy/addons/dumper.py
@@ -157,7 +157,7 @@ class Dumper:
         else:
             url = flow.request.url
 
-        if ctx.options.flow_detail <= 1:
+        if ctx.options.flow_detail < 1:
             # We need to truncate before applying styles, so we just focus on the URL.
             terminal_width_limit = max(shutil.get_terminal_size()[0] - 25, 50)
             if len(url) > terminal_width_limit:


mhils:
Thanks! This looks good, but it doesn't really solve the inconsistencies. For example, right now the addon doesn't print anything when flow_detail is set to 0 (https://github.com/mitmproxy/mitmproxy/blob/main/mitmproxy/addons/dumper.py#L260), so we now essentially have a dead code branch. I think we should keep the behavior of not printing anything with `flow_detail=0` so that `-q` keeps working. Here's what I would propose:

- Level 0: no output
- Level 1: shortened request URL, response status code
- Level 2: full request URL + headers
- Level 3 + 4: as-is.

Does that make sense?

PR #4845, opened 2021-10-06
Fix websocket view jumps to top bug

### LINKED ISSUE #4491: WebSocket view jumps to top on new message
The WebSocket view keeps jumping to the top every time a new message arrives. This makes it basically impossible to work with while the connection is open. I hold down arrow -> it scrolls a bit -> message arrives -> I'm back at the top

_Originally posted by @Prinzhorn in https://github.com/mitmproxy/mitmproxy/issues/4486#issuecomment-796578909_

aaron-tan:
Resolves #4491 

Websocket view will now jump to bottom every time a new message arrives.

### DIFF
--- a/mitmproxy/tools/console/searchable.py
+++ b/mitmproxy/tools/console/searchable.py
@@ -19,6 +19,7 @@ class Searchable(urwid.ListBox):
     def __init__(self, contents):
         self.walker = urwid.SimpleFocusListWalker(contents)
         urwid.ListBox.__init__(self, self.walker)
+        self.set_focus(len(self.walker) - 1)
         self.search_offset = 0
         self.current_highlight = None
         self.search_term = None


Prinzhorn:
Thanks for looking into this!

> Websocket view will now jump to bottom every time a new message arrives.

But how does that solve the problem? What I (as a user) want is to scroll through the list and not have any jumping at all. Having it always be at the bottom would be great _optional_ feature that I can toggle, if I want to follow the list (tail -f style).

I don't know how the whole rendering works, but either we need to store + restore the scroll position or the rendering needs to be more intelligent and no re-render the whole thing when data arrives.

mhils:
This is clearly better than the previous approach, so I'll merge it right away. I'm sure there is a better way to retain the position, we manage to do that for the flow list after all. 😅

PR #4851, opened 2021-10-10
Grpc contentview

mame82:
#### Description

New gRPC and Protobuf view (handles both).
Customizable with filter-based message definitions (allows custom field names and decodings).
Renders data flattened, which is better suited for displaying messages of unknown scheme (concatenated field tags, which could serve as identifier for custom definitions).

Beside the contentview itself, another contentview which shows how to use custom protobuf definitions was added to `examples/addons` (higher view priority).

The contentview was heavily tested with real-world gRPC traffic, but I have no time left to implement test cases.
(I am able to provide flowdumps for testing, though)

I am filing the PR for further discussions.

# Additional notes

1. The code contains a "hack" to prevent a bug, which crashes mitmproxy if generator-based contentviews throw exceptions, after leaving the edit mode for message body data (data gets invalid). For details see respective comments in code.

2. The contentview uses a table layout (similar to `base.format_pairs`, but with arbitrary number of columns). It should be considered to move the respective function `format_table` to `mitmproxy.contentviews.base`.

3. Large parts of the new contentview code are occupied by the `ProtoParser` class. There is some fine tuning left for me, to fully decouple this class from the view (following MCV model). I'd prefer to place this class in a module in a different location, to keep the code of the actual view more readable and allow to use the parser elsewhere (for example in scripts). I am not sure about the intended place for such functionality, `mitmproxy.contrib` maybe?

# Possible future work:

1. JSON import/export for file based proto definitions. With this in mind, the classes involved into creation of parser rules, have already been layd out as dataclasses. Anyways, it seems to be a better idea to give the "rule definition scheme" a bit of time, to proof itself in real life use. Frequent changes to the rule definition scheme would render an export format useless (a user is still able to implement a file based rule definition parser in an add-on). To parse definitions from real `proto` files would be a totally different story, due to its complexity - so this is not in my scope.

2. If the parser is finalized, it could be considered to add an encoder  (write back modified protobuf data to http messages, while respecting the given wire types. Such a functionality is perfectly usable in scripts. When it comes to interactive editing inside the frontend (mitmproxy / mitmweb) an editor similar to the one for header fields could help. In fact it could be used immediately, if the produced output of the gRPC-View would be reduced to columns (first column holds field tag and other field info, second column holds field value and is editable). Again, this are only ideas, I have no time left to work on it.

3. I was not able to find a way to trigger re-rendering of the view (similar to cache invalidation after entering and leaving edit mode). If that would be possible, it would be a good idea to add a keybinding to cycle through all possible decodings for primitive protobuf types per key press. Think of two data fields which get parsed as fixed64, but represent a pair of doubles containing GPS coordinates. Being able to toggle decoding in realtime, would help to spot such things easily and thus be a great help in protocol reversing (which is almost always the task for protobuf, unless the definition is known)

Let me know what you thinnk and please consider helping with test case implementation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@
 * Windows: Switch to Python's default asyncio event loop, which increases the number of sockets
   that can be processed simultaneously (@mhils)
 * Add `client_replay_concurrency` option, which allows more than one client replay request to be in-flight at a time. (@rbdixon)
-* New content view to handle gRPC/protobuf. Allows to apply custom definitions to visualize different field decodings.
-  Includes example addon with applying custom definitions for selected gRPC traffic (@mame82)
+* New content view which handles gRPC/protobuf. Allows to apply custom definitions to visualize different field decodings.
+  Includes example addon which applies custom definitions for selected gRPC traffic (@mame82)
 
 ## 28 September 2021: mitmproxy 7.0.4
 
--- a/mitmproxy/contentviews/grpc.py
+++ b/mitmproxy/contentviews/grpc.py
@@ -57,11 +57,11 @@ class ProtoParser:
     @dataclass
     class ParserFieldDefinition:
         """
-        Defines how to parse a field (or fields of the same type) in a protobuf messages.
+        Defines how to parse a field (or multiple fields with the same tag) in a protobuf messages.
 
-        This allows to apply an intended decoding (f.e. decode uint4 as double instead) and a descriptive
-        name to a field. Field definitions are aggregated into rules, which also hold filters matching
-        to messages.
+        This allows to apply an intended decoding (f.e. decode uint64 as double instead) and to assign
+        a descriptive name to a field. Field definitions are aggregated into rules, which also holds
+        a filter to match selected HTTP messages.
 
         The most natural way to use this, is to describe known parts of a single protobuf message
         in a set of field descriptors, pack them into a rule and set the filter of the rule in a way,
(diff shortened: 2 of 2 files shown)


mame82:
The duplicated tags are intentional, they represent repeated fields (similar to arrays of the same nested message). The order of repeated fields is given by the order of appearance in the messages.

Still repeated fields could be mixed with other fields in unpredictable ways.

Example message with repeated field:

```
1.1  "a"
1.1 "b"
3.1 "foo"
1.1. "c"
```
Different Transmission, same logical message 
```
1.1  "a"
3.1 "foo"
1.1 "b"
1.1 "c"
```

In both cases this translates to (pseudo code)

```
1.1: [a,b,c]
3.1: foo
```

In other words:
Message field order is random (not consistent between different transmissions). Order 9f repeated fields across the message is strict (while other fields could be placed in between).

When it comes to numbering of fields (tags):

- they could be chosen freely in the proto definition (there could be gaps)
- even fields which are logically part of the message, are not necessarily part of the transmitted wire message

F.e. a boolean field which is not optional in the defined proto message, would only be transmitted if its value is true. If it is not part of the wire message, the receiver knows that the value is false.

If the same boolean would be an optional field, it would also be transmitted if it is false (otherwise the receiver could not distinguish between an omitted field and a field with default value).

Non-optional fields holding default values (0 for most number fields) never get transmitted, to safe space

mame82:
For reference also see here:
https://developers.google.com/protocol-buffers/docs/encoding#order

In short words:

Without knowing the exact proto definition, it is almost impossible to fully reconstruct a protobuf message from the wire.

The wiredata could be parsed to some level, though

mame82:
Screenshot_20211012-014209_Samsung Internet


The image shows a perfect example for something which could not be de-serialized from the wire, without knowing the proto definition.

What would be recognized:

1) Field ID: 4
2) WireType: 2 (length delimited message)
3) varint message length: 6
4) Message content: 03 8e 02 9e a7 05

There is no way to know that the 6 bytes of message content represent packed repeated varints. Parsing the data as varints would be wild guessing and work in many cases, even if it is not meant to be interpreted like this (a `bytes` field holding those 6 bytes would end up in the exact same wire encoding)

PR #4855, opened 2021-10-12
Feature suggestion: Convert binary data to hex and back for external editors

mame82:
As discussed with @mhils recently.

No tests, neither are functions placed in proper modules.

Just a demo on how binary editing could be done with conversion to a readable format, before handing it over to an external editor.

For a quick test implementation, this already works pretty well. Formatting functionality could use some optimization, though.

## Workflow 

1. Binary message body in mitmproxy (hex view)

hex1

2. Entering edit mode with `e` key (data gets converted to readable representation if not valid UTF-8, editor is Nano in this case)

hex2

3. Edit the "hexlified" data as you like (adding 0xdeadbeef here)

hex3

4. After saving in the editor, data gets converted back to binary (comments prefixed with `#` get stripped)

hex4

### DIFF
--- a/mitmproxy/tools/console/master.py
+++ b/mitmproxy/tools/console/master.py
@@ -32,7 +32,8 @@ from mitmproxy.tools.console import keymap
 from mitmproxy.tools.console import palettes
 from mitmproxy.tools.console import signals
 from mitmproxy.tools.console import window
-from mitmproxy.utils.strutils import always_str, escape_control_characters
+from mitmproxy.utils import strutils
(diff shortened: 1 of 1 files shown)


mame82:
Agree on this.

(Although I'd love to see radare2 as inline hexeditor !1elf!!)

Leaving this open for future discussions and prototype for fallback in absence of console hexeditor

mhils:
This was a useful exchange, thanks. I'm closing this here in favor of #5231. :)

PR #4860, opened 2021-10-13
Make scripts in config.yml relative to config.yml

Speedlulu:
#### Description

Make the paths to scripts found in config.yml relative to
the directory where config.yml is instead of where the command
is ran.

#### Checklist

### DIFF
--- a/mitmproxy/optmanager.py
+++ b/mitmproxy/optmanager.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import contextlib
 import copy
-import os
 import pprint
 import textwrap
 import weakref
@@ -10,6 +9,7 @@ from collections.abc import Callable
 from collections.abc import Iterable
 from collections.abc import Sequence
 from dataclasses import dataclass
+from pathlib import Path
 from typing import Any
 from typing import Optional
 from typing import TextIO
@@ -541,31 +541,38 @@ def parse(text):
     return data
 
 
-def load(opts: OptManager, text: str) -> None:
+def load(opts: OptManager, text: str, cwd: Path | str | None = None) -> None:
     """
     Load configuration from text, over-writing options already set in
     this object. May raise OptionsError if the config file is invalid.
     """
     data = parse(text)
+
+    scripts = data.get("scripts")
+    if scripts is not None and cwd is not None:
+        data["scripts"] = [
+            str(relative_path(Path(path), relative_to=Path(cwd))) for path in scripts
+        ]
+
     opts.update_defer(**data)
 
 
-def load_paths(opts: OptManager, *paths: str) -> None:
+def load_paths(opts: OptManager, *paths: Path | str) -> None:
     """
     Load paths in order. Each path takes precedence over the previous
     path. Paths that don't exist are ignored, errors raise an
     OptionsError.
     """
     for p in paths:
-        p = os.path.expanduser(p)
-        if os.path.exists(p) and os.path.isfile(p):
-            with open(p, encoding="utf8") as f:
+        p = Path(p).expanduser()
+        if p.exists() and p.is_file():
+            with p.open(encoding="utf8") as f:
                 try:
                     txt = f.read()
                 except UnicodeDecodeError as e:
                     raise exceptions.OptionsError(f"Error reading {p}: {e}")
             try:
-                load(opts, txt)
+                load(opts, txt, cwd=p.absolute().parent)
             except exceptions.OptionsError as e:
                 raise exceptions.OptionsError(f"Error reading {p}: {e}")
 
@@ -594,15 +601,15 @@ def serialize(
     ruamel.yaml.YAML().dump(data, file)
 
 
-def save(opts: OptManager, path: str, defaults: bool = False) -> None:
+def save(opts: OptManager, path: Path | str, defaults: bool = False) -> None:
     """
     Save to path. If the destination file exists, modify it in-place.
 
     Raises OptionsError if the existing data is corrupt.
     """
-    path = os.path.expanduser(path)
-    if os.path.exists(path) and os.path.isfile(path):
-        with open(path, encoding="utf8") as f:
+    path = Path(path).expanduser()
+    if path.exists() and path.is_file():
+        with path.open(encoding="utf8") as f:
             try:
                 data = f.read()
             except UnicodeDecodeError as e:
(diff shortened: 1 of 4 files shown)


Speedlulu:
Hello, is this change interesting to you guys ?

mitmproxy/optmanager.py
```diff
@@ -612,3 +617,15 @@ def save(opts: OptManager, path: str, defaults: bool = False) -> None:
...
+def compute_path(script_path: str, config_path: str) -> str:
+    """
+    Make relative paths found in config files relative to said config file,
+    instead of relative to where the command is ran.
+    """
+    if os.path.isabs(script_path):
+        return script_path
+    if script_path != os.path.expanduser(script_path):
+        return script_path
+    return os.path.abspath(os.path.join(os.path.dirname(config_path), script_path))
```
mhils, line 631: I'd like to use pathlib as much as possible, even if that comes at the cost of additional Path/str wrapping above. :) We can also make this simpler by making use of the fact that `Path("/foo") / Path("/bar")` resolves to `/bar`. :)

```suggestion
def script_path(script_path: Path, *, relative_to: Path) -> Path:
    """
    Make relative paths found in config files relative to said config file,
    instead of relative to where the command is ran.
    """
    return relative_to / script_path.expanduser().absolute()
```
(untested - relative_to requires a `relative_to or Path.cwd()` at the call site)

mitmproxy/optmanager.py
```diff
@@ -541,12 +541,17 @@ def parse(text):
     return data
 
 
-def load(opts: OptManager, text: str) -> None:
+def load(opts: OptManager, text: str, config_path: str | None = None) -> None:
```
mhils, line 544: Maybe rename `config_path` to `cwd`?

(Not sure, you decide what's best 😃 )

test/mitmproxy/test_optmanager.py
```diff
@@ -469,3 +478,119 @@ def test_set():
     opts.process_deferred()
     assert "deferredsequenceoption" not in opts.deferred
     assert opts.deferredsequenceoption == ["a", "b"]
+
+
+def test_relative_script_path(tmpdir):
```
mhils, line 483: Can we please have one test that verifies one `optmanager.load_paths` invocation with a hardcoded config (no need for yaml dumping at runtime - see https://abseil.io/resources/swe-book/html/ch12.html#donapostrophet_put_logic_in_tests), and then a separate `pytest.mark.parametrize`d test for `script_path`? ☺️

mhils (commented):
Awesome, thanks! 🍰 
I've left some very nitpicky comments below, looks great overall! 😃

mitmproxy/optmanager.py
Speedlulu: I'll get on it right away.
Just a quick question, why make `relative_to` a kw_only argument here ?

mhils:
No super strong opinion here, I just felt that script_path(...,
relative_to=x) is very readable. Feel free to take it as a regular arg if
you prefer that. :)

mitmproxy/optmanager.py
Speedlulu: I had to handle the case where `$HOME` is not set to an absolute path (which is the case in the tox test env).

mitmproxy/optmanager.py
```diff
@@ -541,31 +541,38 @@ def parse(text):
...
     """
     Load configuration from text, over-writing options already set in
     this object. May raise OptionsError if the config file is invalid.
     """
     data = parse(text)
+
+    scripts = data.get("scripts")
+    if scripts is not None and cwd is not None:
+        data["scripts"] = [
+            str(relative_path(Path(path), relative_to=Path(cwd))) for path in scripts
```
mhils, line 554: If cwd is None, `Path(cwd)` will fail I think?
(reviewing from mobile and can't verify)

test/mitmproxy/test_optmanager.py
```diff
@@ -469,3 +477,44 @@ def test_set():
...
+    "script_path, relative_to, expected",
+    (
+        ("~/abc", ".", Path.home().joinpath("abc")),
+        ("/abc", ".", Path("/abc")),
+        ("abc", ".", Path(".").joinpath("abc")),
+        ("../abc", ".", Path(".").joinpath("../abc")),
+        ("~/abc", "/tmp", Path.home().joinpath("abc")),
+        ("/abc", "/tmp", Path("/abc")),
+        ("abc", "/tmp", Path("/tmp").joinpath("abc")),
+        # Edge case for MacOS
```
mhils, line 505: What's the edge case here?

(Let's have a short explanation in the doscstring)

mitmproxy/optmanager.py
```diff
@@ -594,21 +601,33 @@ def serialize(
...
+
+def relative_path(script_path: Path | str, *, relative_to: Path | str) -> Path:
+    """
+    Make relative paths found in config files relative to said config file,
+    instead of relative to where the command is ran.
+    """
+    script_path = Path(script_path)
+    # Edge case when $HOME is not an absolute path
+    if script_path.expanduser() != script_path and not script_path.is_absolute():
+        script_path = script_path.expanduser().resolve()
```
mhils, line 632: Do we need to call .resolve() here? Symlinks may change, and users may change them intentionally. :)
If .resolve() makes things significantly easier I'm happy to live with this constraint though.

mhils (commented):
Thanks! Relative $HOME is cursed, great catch. :sweat_smile:

mitmproxy/optmanager.py
Speedlulu: Yes, but we check that it is not None on [line 552](https://github.com/mitmproxy/mitmproxy/pull/4860/files#diff-6b02b884786fb205ec197276b4fc6434ff3d51f37fea73479f8c71a888565b6aR552).

mitmproxy/optmanager.py
mhils: > won't they need to reload the options object thus re-resolving the new symlink ?

mitmproxy live-reloads scripts, so you don't need to reload options. :) I think `.absolute()` is more accurate here, so let's try to stick with that.

mhils (commented):
Thanks for the quick turnaround! 😃

mhils (approved):
Awesome, fantastic work! 😃 🚀 🚀 🚀

PR #4864, opened 2021-10-14
WIP: hacky support for EC keys

hexploitable:
#### Description

WIP because it's just a hack that I used locally.
There's a more elegant way to do this and we would need to update the type definitions to not just simply be an RSA key etc.
but for now I'll submit the PR to discuss how we might allow this. Oh, and also to aid anyone who is in need of using EC keys too.

I'll omit the tests & change log entry until it's decided.

#### Checklist

### DIFF
--- a/mitmproxy/net/tls.py
+++ b/mitmproxy/net/tls.py
@@ -230,7 +230,7 @@ def create_client_proxy_context(
             NoEncryption,
             PrivateFormat
         )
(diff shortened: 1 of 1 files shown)


mhils:
Thanks. The cryptography folks are fantastic and usually very responsive, so getting this into upstream should be feasible. In the meantime whoever needs this shall cherry-pick this PR, thanks again for sharing! 🍰

PR #4867, opened 2021-10-15
gRPC view: special case decoding

mame82:
#### Description

It appears that Protobuf based communication with Google endpoints uses field values slightly deviating from the specs (created by Google, too).

More specifically cases have been observed, were field values containing nested Protobuf messages got prefixed with an additional length value (encoded as variable-length integer). Such values could not be decoded, as the parser expected raw messages (without prefixes) as payload of length delimited fields.

While this is a special case, it is easy to test for in the current gRPC contentview implementation - with minimal computing overhead and without disrupting expected decoding behavior for other protobuf traffic.

I consider this encoding scheme a "common case" as it is used by Google itself.

The gRPC contentview was extended to handle this case (tests included).

1. The screenshot shows a protobuf message, containing a nested message which can not be decoded without the adjustment (Field `1.5`):

google-special1

2. The screenshot shows an excerpt of the same message, with this patch applied (Field `1.5` got decoded and holds a valid protobuf message):

google-special2

As shown in the screenshot, messages decoded like this are marked with `len_prefixed_message`. This decoding was inserted into the default decoding-failover-strategy. It is still possible to use custom definitions, to enforce decoding as `bytes` for such fields.

Beside this, a missing MIME type for protobuf messages was added.

#### Checklist

### DIFF
--- a/mitmproxy/contentviews/grpc.py
+++ b/mitmproxy/contentviews/grpc.py
@@ -147,7 +147,7 @@ class ProtoParser:
         # helper
         unknown = 18
         # special
-        # googleapis traffic has found to include varint length prefixes to nested messages in some cases
+        # googleapis traffic was found to include varint length prefixes to nested messages in some cases
         len_prefixed_message = 19
 
     class Message:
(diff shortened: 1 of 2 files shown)


mame82:
Closing this PR for rework, as it turned out that the wire data represents `packed, repeated` elements (`message[]` in this case).

Such decoding can not be predicted reliably and has to be added as custom decoding

PR #4872, opened 2021-10-19
gRPC packed repeated encoding

mame82:
#### Description

Added support for packed field decoding in protobuf/gRPC flows.

According to the protobuf specs, repeated fields could be stored with `[packed = true]`.

In such a case, the field preceding `key` - which contains the field tag and the **wire type** is omitted. Instead, a sequence of values without key is serialized into a length delimited sequence and used as field value. Thus, without knowing how values were encoded, the protobuf parser could not make an educated guess on the value type of the individual packed fields, as the wire type is missing.

For this reason custom field definitions for the gRPC/Protobuf view have been extended with support for packed field encodings.

Although, according to the protobuf v2 and v3 specs, packed encoding only applies to scalar field types, support for packed messages has been added, too. This is because they have been spotted in Google's gRPC traffic (see example below).

Also, the Kaitai based parser has been replaced with a built-in parser, because it did not follow the specs (the base128 encoding of the Kaitai decoder only respects variable length integers up to 56bit, while the protobuf specs support variable length integers up to 64bit).

**The patch preserves full backwards compatibility to the previous implementation of gRPC view, while providing full test coverage for the added functionality** 

Example:

1. No packed decoding rules (Google traffic with unidentified field content, decoded as `bytes`):

google-packed1

2. Custom Definition to deal with those fields as "packed, repeated messages" (field tag `1.5`)

google-packed4

3. Decoding the same flow, with custom definitions applied (note that `1.5` appears repeated)

google-packed2

google-packed3

#### Checklist

### DIFF
--- a/mitmproxy/contentviews/grpc.py
+++ b/mitmproxy/contentviews/grpc.py
@@ -257,34 +257,78 @@ class ProtoParser:
 
     @staticmethod
     def read_packed_fields(
-        wt: ProtoParser.WireTypes,
         packed_field: ProtoParser.Field,
     ) -> List[ProtoParser.Field]:
         if not isinstance(packed_field.wire_value, bytes):
+            ctx.log(type(packed_field.wire_value))
             raise ValueError("can not unpack field with data other than bytes")
         wire_data: bytes = packed_field.wire_value
         tag: int = packed_field.tag
         options: ProtoParser.ParserOptions = packed_field.options
         rules: List[ProtoParser.ParserRule] = packed_field.rules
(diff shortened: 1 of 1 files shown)


mame82:
MacOS 3.9 job never started and shows no usable error

mame82:
Thanks for your responsiveness.

Could you please provide info, on when commits are reflected into docker builds (latest)

mhils:
I realistically won't get around to cutting a new release for at least six weeks, but you should be absolutely safe using the `mitmproxy:dev` image until then. I hope to have some more mitmproxy time around Christmas. :)

PR #4882, opened 2021-10-27
Fix #4876 Don't do CONNECT on plaintext HTTP replays via upstream

### LINKED ISSUE #4876: Replay via upstream proxy uses CONNECT instead of original request method for HTTP connections
#### Problem Description
Replaying a request that was routed through a upstream proxy server fails as the preceding CONNECT method is not supported.

1. Run mitmpoxy in upstream mode (e.g. route traffic through tinyproxy server)
2. Execute any request
3. Replay this request
4. The replay fails as it is executed with a preceding CONNECT request

#### System Information
Mitmproxy: 8.0.0.dev (+148, commit 79f464b)
Python:    3.8.10
Platform:  Linux-5.11.0-38-generic-x86_64-with-glibc2.29

HoffmannP:
#### Description

Upstream replays used to do a CONNECT even for plaintext HTTP which some proxies did not understand. This was removed, the test catching this deviation was removed

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
   Includes example addon which applies custom definitions for selected gRPC traffic (@mame82)
 * Fix a crash caused when editing string option (#4852, @rbdixon)
 * Base container image bumped to Debian 11 Bullseye (@Kriechi)
+* Upstream replays don't do CONNECT on plaintext HTTP requests (#4876, @HoffmannP)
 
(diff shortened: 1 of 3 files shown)


HoffmannP:
I fixed the test. As for the assert-in-async-problem I see no "clean" way to do it and any other monkey wrench would still be just that

mhils (approved):
LGTM % the lint error. :)

Issue #4460, opened 2021-02-19
mitmdump: temporarily running out of diskspace corrupts dumpfile


r00t-:
#### Problem Description
ofcourse the world would be a better place if everybody always responsibly managed their filesystem space,
but we have to face that fact that mitmdump, as a non-interactive dumping tool, is often going to be the program that fills up the last available block on a filesystem.

currently it deals with that situation very badly:
* it keeps running, and keeps *trying* to write to the file
  * which in part is good, because it will resume saving data when space becomes available again
* the last flow it tried to write is left in the file half-written and truncated, which due to the length-prefixed data format leads to the file being corrupt and not loadable with the provided tools.

due to this situation i already learned more about the file format than i ever wanted to ;)
the key to recovering data in such a corrupted file is:
```
grep -oE --binary-files=text --byte-offset '[0-9]+:7:version;'
```
(and i wrote a script that automates the process)

but it would be better to fix this in mitmdump itself.
suggestions:
* on a failed write, truncate the file back to it's previous length
* on a failed write, exit with an error
* maybe provide a tool to recover data from corrupted files

#### Steps to reproduce the behavior:
log more data than is good for you

#### System Information
any version i ever used

mhils:
Good point - I think we should fail visibly and exit with an error on a failed write. 

Optionally we can also consider checking `shutil.disk_usage` on top and exit at 99% disk usage already, that should make recovery much easier in many cases. This is not available on all platforms, but we can just ignore the check if we don't get any stats.

sw9719:
Is it good idea to show a warning when the filesystem is say 95% filled up?

r00t-:
@sw9719: that might make sense for interactive usage of mitmproxy,
but for a non-interactive program like mitmdump (which this ticket explicitly mentions), a warning (most likely ending up in some logfile) is unlikely to be noticed.
(also, with today's mult-terabyte size disks, demanding 5% free space is excessive. - maybe create a metric from the rate of the free space, divided by the amount data being logged/time-unit, in relation to the time-units the process is running for? ("you've been logging for six hours, and in two more hours, available space will be exceeded") - wonder where the balance between sensible and excessively complicated lies.)

but really... the only critical part of the issue (that makes mitmdump's behaviour a bug) is that mitmdump's ignoring of the error leads to corrupt files that cannot be read back, as mentioned in the original description.
any further improvements would at best be nice-to-haves on top.

sw9719:
[...]

PR #4888, opened 2021-11-02
mitmdump: temporarily running out of diskspace corrupts dumpfile

sw9719:
#### Description

As mentioned in this [issue](https://github.com/mitmproxy/mitmproxy/issues/4460), mitmdump writing to a file can lead to data corruption when disk becomes full.
To resolve this, this PR adds disk usage checks and exits the program when it exceeds a preset threshold. 

#### Checklist

 - [ ] If this is acceptable, I can write tests for this.

### DIFF
--- a/mitmproxy/addons/save.py
+++ b/mitmproxy/addons/save.py
@@ -9,6 +9,8 @@ from mitmproxy import ctx
 from mitmproxy import flow
 from mitmproxy import http
 import mitmproxy.types
+import sys
+import errno
 
 
 class Save:
@@ -71,10 +73,17 @@ class Save:
         except OSError as v:
             raise exceptions.CommandError(v) from v
         stream = io.FlowWriter(f)
-        for i in flows:
-            stream.add(i)
-        f.close()
-        ctx.log.alert("Saved %s flows." % len(flows))
+        try:
+            for i in flows:
+                stream.add(i)
+            f.close()
+            ctx.log.alert("Saved %s flows." % len(flows))
+        except OSError as e:
+            if e.errno == errno.ENOSPC:
+                ctx.log.error("Exiting due to insufficient space on disk")
+                sys.exit(1)
+            else:
+                raise e
 
     def tcp_start(self, flow):
         if self.stream:
@@ -101,8 +110,15 @@ class Save:
         # websocket flows will receive a websocket_end,
         # we don't want to persist them here already
         if self.stream and flow.websocket is None:
-            self.stream.add(flow)
-            self.active_flows.discard(flow)
+            try:
+                  self.stream.add(flow)
+                  self.active_flows.discard(flow)
+            except OSError as e:
+                if e.errno == errno.ENOSPC:
+                    ctx.log.error("Exiting due to insufficient space on disk")
+                    sys.exit(1)
+                else:
+                    raise e
 
     def error(self, flow: http.HTTPFlow):
         self.response(flow)
--- a/mitmproxy/addons/save.py
+++ b/mitmproxy/addons/save.py
@@ -46,16 +46,16 @@ class Save:
         self.stream = io.FilteredFlowWriter(f, flt)
         self.active_flows = set()
 
-    def add(self,flow):
+    def add(self, flow):
         try:
             self.stream.add(flow)
         except OSError as e:
             if e.errno == errno.ENOSPC:
-               ctx.log.error("Exiting due to insufficient space on disk")
-               if isinstance(self.stream,FilteredFlowWriter)
-                  sys.exit(1)
+                ctx.log.error("Exiting due to insufficient space on disk")
+                if isinstance(self.stream, io.FilteredFlowWriter):
+                    sys.exit(1)
             else:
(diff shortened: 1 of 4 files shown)


sw9719:
Hello, just wanted to check if this is acceptable

mitmproxy/io/tnetstring.py
```diff
@@ -63,7 +66,14 @@ def dump(value: TSerializable, file_handle: typing.IO[bytes]) -> None:
...
     """
-    file_handle.write(dumps(value))
+    try:
+        file_handle.write(dumps(value))
+    except OSError as e:
+        if e.errno == errno.ENOSPC:
+            ctx.log.error("Exiting due to insufficient space on disk")
+            sys.exit(1)
+        else:
+            raise e
```
mhils, line 76: The logic here looks good, but this should really only happen when continuously streaming flows to disk. 

Imagine you have an interactive mitmproxy session, you want to save your flows, there is not enough disk space, and now everything is closed automatically. That would be painful. :) Long story short, let's move this into https://github.com/mitmproxy/mitmproxy/blob/main/mitmproxy/addons/save.py.

mhils (changes requested):

mitmproxy/io/tnetstring.py
sw9719: When running mitmproxy,  if a user decides to stream to a file, it will still exit without a warning because the function executed will be same for both mitmproxy and mitmdump. I think I need to look for a way to detect if the 'save stream to file' was requested by mitmproxy or mitmdump.

mhils: Yes, the save addon handles that for both. :)

sw9719:
Since `stream.add` is called from multiple methods in `addons/save.py`, I added the excpetion handling in `io/io.py`. Also added logic to detect if the `Save` addon was initialized for mitmdump or mitmproxy and tested it successfully.

mhils (changes requested):
`self.stream.add` is called four times in the save addon. I think that's a reasonable amount to cover, feel free to write a helper function there if you want to reuse functionality. The current approach here introduces unnecessary complexity across multiple interfaces.

sw9719:
I am a bit confused. Do you mean create a wrapper function for `stream.add` and add the mitmdump detection logic in that?

mhils:
We don't want to detect the usage of mitmdump, we specifically want to detect the usage of `save_stream_file` (which sets `self.stream` in the Save addon).

mitmproxy/addons/save.py
```diff
@@ -44,6 +46,17 @@ def start_stream_to_path(self, path, flt):
         self.stream = io.FilteredFlowWriter(f, flt)
         self.active_flows = set()
 
+    def add(self, flow):
+        try:
+            self.stream.add(flow)
+        except OSError as e:
+            if e.errno == errno.ENOSPC:
+                ctx.log.error("Exiting due to insufficient space on disk")
+                if isinstance(self.stream, io.FilteredFlowWriter):
```
mhils, line 55: I think we can also remove this check? Is there a case where this method is called and `self.stream` is not a FilteredFlowWriter?

mitmproxy/addons/save.py
```diff
@@ -44,6 +46,17 @@ def start_stream_to_path(self, path, flt):
         self.stream = io.FilteredFlowWriter(f, flt)
         self.active_flows = set()
 
+    def add(self, flow):
+        try:
+            self.stream.add(flow)
+        except OSError as e:
+            if e.errno == errno.ENOSPC:
```
mhils, line 53: Practically speaking it may be useful if we just exit on _every_ OSError? I mean in any case it indicates that we cannot save anymore, which is critical.

mhils (commented):
This looks great now, thanks for incorporating the feedback! :cake:

I have two smaller remarks, other than that we need a test and we should be good! :)

mitmproxy/addons/save.py
sw9719: Thinking about it now, it is not needed. This check is from one of my previous commits as it was breaking on of the tests. Ill remove it and re-run the tests to make sure nothing breaks.

mitmproxy/addons/save.py
sw9719: Okay, but if exit on every OSError, then we might not understand why exactly the call failed. Is it better to add log the error string from OSError and then exit?

mhils: Yes, that sounds great.

sw9719: Hello, I fixed the code as needed. Is this ok? If yes, Ill proceed with tests.

sw9719:
For the test, I was thinking of creating a small file system and mounting it on test directory and filing it. But this require root privileges so is that fine or do we need some other method?

mhils:
I think this is addressed by https://github.com/mitmproxy/mitmproxy/commit/3a5550a09cd40d76acfe71aa45c7a8309525ad51 now. Thanks again! 🍰

Issue #4846, opened 2021-10-08
Upgrade base image to Debian?


roy-work:
#### Problem Description

It appears that the base image for `mitmproxy` is Debian Buster/"oldstable"; would it be possible to update, or offer, a version based on Debian Bullseye/"stable"?

#### Background

The reason I ask is that the version of GnuTLS in Debian oldstable is horribly out of date. [It includes a bug that causes it to incorrectly fail to validate TLS certificates](https://community.letsencrypt.org/t/openssl-client-compatibility-changes-for-let-s-encrypt-certificates/143816) (that link notes that GnuTLS < 3.6.14 26 is affected; [the version in Debian Buster](https://packages.debian.org/buster/libgnutls30) is 3.6.7) where the root has an expired cross-sign, [which is now the case for Let's Encrypt](https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/). `apt` when talking HTTPS uses GnuTLS, so `apt-get install` against an HTTPS source (which is not normal, normally, `apt` does not use encryption when fetching packages; however, security requirements in my industry requires that we vendor packages & use TLS to prevent, e.g., supply-chain attacks; thus, we end up fetching over HTTPS) fails in Buster. We use `mitmproxy`, but also need `nc`. It used to be installed by default, back when things were based on Alpine, but that's no longer the case. We also think that the absolutely ancient version of `mitmproxy` we're using is affected by a similar bug, as it is failing to validate a cert that is actually valid, and the failures started at the time of the cross-signing cert's expiry. But we'd like to upgrade to the newer version, of course, first. But the lack of `nc` in the base image means it doesn't work as a drop in replacement, and the base image of Buster means it can't be installed.

PR #4889, opened 2021-11-03
Use the slim base image for container

czenker:
#### Description

Container sizes went from [`73MB` in Version 7.0.4](https://hub.docker.com/layers/mitmproxy/mitmproxy/7.0.4/images/sha256-36e9f5397302341d55fec03953e04cd69c47038e051d5ccec6d9e84f987527cc?context=explore) to [`359MB` in `dev`](https://hub.docker.com/layers/mitmproxy/mitmproxy/dev/images/sha256-c418f4eda4a30f486b76a3004e6a04e8fd172095ac29b29de8371b06d41aed25?context=explore). Also the image now contains tools like `imagemagick`, `git` and `libbluetooth3`.

The change was introduced in [commit `fffed0cb`](https://github.com/mitmproxy/mitmproxy/commit/fffed0cb3aa121b6c3735ca3f6beeb4b2356b4e6#diff-5ff03f5acddd8edc535d5ec30b70e78468bf1e748e126ba3ac39c67a28a6af40) where the base image was changed from `python:3.9-slim-buster` to `python:3.9-bullseye`. It was probably an oversight that "slim" was dropped.

#### Checklist

 - [x] ~~I have added an entry to the CHANGELOG.~~ Mention of updated base image is already in the changelog and has not yet been released

### DIFF
--- a/release/docker/Dockerfile
+++ b/release/docker/Dockerfile
@@ -4,7 +4,7 @@ ARG MITMPROXY_WHEEL
 COPY $MITMPROXY_WHEEL /wheels/
 RUN pip install wheel && pip wheel --wheel-dir /wheels /wheels/${MITMPROXY_WHEEL}
 
-FROM python:3.9-bullseye
+FROM python:3.9-slim-bullseye
 
 RUN useradd -mU mitmproxy
 RUN apt-get update \


mhils:
That change was actually intentional, see https://github.com/mitmproxy/mitmproxy/issues/4846#issue-1020560165 and then https://github.com/mitmproxy/mitmproxy/issues/4846#issuecomment-938355326. 
https://hub.docker.com/_/python says this:

> This image does not contain the common packages contained in the default tag and only contains the minimal packages needed to run python. Unless you are working in an environment where only the python image will be deployed and you have space constraints, we highly recommend using the default image of this repository.

So yes... 🤷  I agree that the new container size is somewhat silly. @mitmproxy/devs, does anyone have strong opinions here?

mhils:
also /cc @roy-work - this would remove `nc` again.

Kriechi:
Since we can't make everybody happy with a single image, how about two images in the future:
* `mitmproxy:slim-bullseye`, which is also the default versioned and the `latest` tag
* `mitmproxy:bullseye`, or whatever distro we choose in the future

Essentially replicating the python image tags without the python version.

mhils:
Strongly 👎 to that, too much complexity.

mhils:
Seems like noone has strong opinions against this change, it bothered @czenker enough to open this PR, so let's move to `-slim` and see how that works out! 😃 

Thanks again @czenker! 🍰

Issue #4892, opened 2021-11-05
Distribute via source tarball in additional to wheel on PyPI


bollwyvl:
#### Problem Description

I'd like to re-package `mitmproxy` as a package on `conda-forge` from an official source distribution.

#### Proposal

On conda-forge, we prefer to use official source distributions, usually `.tar.gz` from PyPI. We'd love to use (noarch) whls someday, but it will take a while for all the PEPs to die down.

#### Alternatives

We can build directly from a VCS-generated tarball, but those have more of a habit of changing (e.g. the shasum changing) than an artifact generated by the project's owners/automation.

#### Additional context

Thanks for `mitmproxy`!

mhils:
According to https://docs.conda.io/projects/conda-build/en/latest/user-guide/wheel-files.html#building-a-conda-package-from-a-wheel-file, conda seems to be able to build packages just from wheels. 

If someone wants to update our CI to also deploy sdists to PyPI I'd be happy to merge it, but it's nothing I'm planning to address anytime soon. :)

bollwyvl:
Thanks for the super-quick response! 

While it is technically possible (and all but _required_ for monsters like `tensorflow`) it's kind of a last-resort from a conda-forge perspective... especially as there's no funny stuff in here (e.g. some weird compiled `kaitaistruct` stuff), we try to start with source.

I've opened up a nickle fix to explore this on #4893... but even if this _did_ go out, there's still a fair number of missing pieces (e.g `kaitaistruct`!) before i'd be able to get `mitmproxy` into the pipeline.

PR #4893, opened 2021-11-05
[wip] build python sdists along with wheel

bollwyvl:
#### Description

For #4892, this explores what the CI infrastructure _would_ do if there just happened to be an sdist built at the same time as the wheel. From a cursory look at the workflows and scripts, the `dist` folder seems to be handed around, rather than the `.whl` itself, so it's possible everything will "just work" up to deployment to pypi, even if it's not being further tested as an artifact.

#### Checklist

### DIFF
--- a/release/cibuild.py
+++ b/release/cibuild.py
@@ -268,12 +268,12 @@ def build_wheel(be: BuildEnviron) -> None:  # pragma: no cover
         "bdist_wheel",
         "--dist-dir", be.dist_dir,
     ])
+    all_dist = be.dist_dir.glob('*')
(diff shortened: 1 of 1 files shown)


mhils:
I'm closing this as it's marked WIP and CI is failing, please feel free to open a new PR anytime! :)

bollwyvl:
Sdist builds still seem like something someone with a good understanding of the build machinery could add in a few well-placed lines, and would be very helpful to downstreams, though perhaps I'm not the right one to do so.

So, sure, closing is fine, I've been able to solve my immediate proxy needs with something simpler (but more inflexible), and probably won't pick this back up unless the situation changes drastically.

Issue #4831, opened 2021-09-29
Remove old pyOpenSSL cruft


mhils:
The pyOpenSSL folks shipped a new release (#4829), so we can bump the minium version requirement and remove some hacky workarounds in [`mitmproxy.net.tls`](https://github.com/mitmproxy/mitmproxy/blob/main/mitmproxy/net/tls.py) 🙌

PR #4897, opened 2021-11-09
Remove pyopenssl cruft

KarlParkinson:
#### Description

Addresses #4831.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 * Fix a crash caused when editing string option (#4852, @rbdixon)
 * Base container image bumped to Debian 11 Bullseye (@Kriechi)
 * Upstream replays don't do CONNECT on plaintext HTTP requests (#4876, @HoffmannP)
-* Remove old pyOpenSSL cruft (#4831, @KarlParkinson)
+* Remove workarounds for old pyOpenSSL versions (#4831, @KarlParkinson)
 
 ## 28 September 2021: mitmproxy 7.0.4
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Fix a crash caused when editing string option (#4852, @rbdixon)
 * Base container image bumped to Debian 11 Bullseye (@Kriechi)
 * Upstream replays don't do CONNECT on plaintext HTTP requests (#4876, @HoffmannP)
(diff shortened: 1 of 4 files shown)


mitmproxy/net/tls.py
```diff
@@ -21,29 +21,21 @@
 
 # redeclared here for strict type checking
 class Method(Enum):
-    # TODO: just SSL attributes once https://github.com/pyca/pyopenssl/pull/985 has landed.
-    TLS_SERVER_METHOD = getattr(SSL, "TLS_SERVER_METHOD", 8)
-    TLS_CLIENT_METHOD = getattr(SSL, "TLS_CLIENT_METHOD", 9)
+    TLS_SERVER_METHOD = getattr(SSL, "TLS_SERVER_METHOD")
+    TLS_CLIENT_METHOD = getattr(SSL, "TLS_CLIENT_METHOD")
```
mhils, line 25: We don't need to use `getattr` here or below, we can just use `SSL.TLS_SERVER_METHOD` directly. :)

mitmproxy/net/tls.py
```diff
@@ -21,29 +21,21 @@
...
 
 
-# TODO: remove once https://github.com/pyca/pyopenssl/pull/985 has landed.
```
mhils, line 29: I think the comment here got moved somehow. Let's keep the version check, but remove the `setdefault` lines below. They shouldn't be necessary.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 * Fix a crash caused when editing string option (#4852, @rbdixon)
 * Base container image bumped to Debian 11 Bullseye (@Kriechi)
 * Upstream replays don't do CONNECT on plaintext HTTP requests (#4876, @HoffmannP)
+* Remove old pyOpenSSL cruft (#4831, @KarlParkinson)
```
mhils, line 17: Let's call this "remove workarounds for old pyOpenSSL versions", in retrospect I realize that my wording from the issue could be misunderstood. pyOpenSSL is not crufty. :)

mhils (commented):
Thanks for taking care of this! 🍰

mhils (approved):
Fantastic, thanks! 🍰

KarlParkinson:
My pleasure @mhils.

PR #4905, opened 2021-11-17
Fix #4902

### LINKED ISSUE #4902: Crash when invoking replay.server.count from console
#### Problem Description
Invoking `replay.server.count` from the console causes a crash. I don't think it happens all the time, but see below for repeatable reproduction.

1. Start `mitmproxy`
2. Hit `n` to create a new flow
3. Hit Enter
4. Hit `r` to issue the request
5. With the same flow selected, issue the command `:replay.server @focus`
6. Issue the command `:replay.server.count`

Sample stack trace:
```python
  File "/home/elespike/venvs/mitmproxy/lib/python3.9/site-packages/mitmproxy/tools/console/grideditor/editors.py", line 222, in __init__
    if not isinstance(vals[0], list):
TypeError: 'int' object is not subscriptable
```

#### System Information
```
Mitmproxy: 7.0.4
Python:    3.9.2
```

shindexro:
#### Description

Convert all non-list types into list, rather than blindly assuming `vals` must either be a string or a list.

#### Checklist

### DIFF
--- a/mitmproxy/tools/console/grideditor/editors.py
+++ b/mitmproxy/tools/console/grideditor/editors.py
@@ -215,7 +215,7 @@ class DataViewer(base.GridEditor, layoutwidget.LayoutWidget):
                 typing.List[typing.Any],
                 typing.Any,
             ]) -> None:
-        if vals:
+        if vals is not None:
             # Whatever vals is, make it a list of rows containing lists of column values.
             if not isinstance(vals, list):
                 vals = [vals]
--- a/mitmproxy/tools/console/grideditor/editors.py
+++ b/mitmproxy/tools/console/grideditor/editors.py
@@ -213,7 +213,7 @@ class DataViewer(base.GridEditor, layoutwidget.LayoutWidget):
             vals: typing.Union[
                 typing.List[typing.List[typing.Any]],
                 typing.List[typing.Any],
-                str,
+                typing.Any,
             ]) -> None:
         if vals:
             # Whatever vals is, make it a list of rows containing lists of column values.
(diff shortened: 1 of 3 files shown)


shindexro:
Hi, first time contributing here.

Any tips on how to resolve the test coverage? I thought the modified line should still be covered by the original tests (although I have no idea where the specific tests locate.)

mitmproxy/tools/console/grideditor/editors.py
```diff
@@ -217,7 +217,7 @@ def __init__(
             ]) -> None:
```
mhils, line 217: Let's also adjust the somewhat weird type signature here. This probably should be `typing.Any` instead of `str` in the last line if we intend to support ints.

mhils (commented):
Thanks for the PR! 😃🍰  
Coverage is complaining because we did not have test coverage here, and now <s>the evil coverage overlords</s> CI wants to nudge us into adding some.

The reason we don't see this crash on every invocation is https://github.com/mitmproxy/mitmproxy/blob/83f05897ce57dad5c98d787d23dc87ca4b0a663c/mitmproxy/tools/console/commandexecutor.py#L22 This should probably be an `is None` check (right now, a return value of `0` is discarded).

Could you 1) adjust the abovementioned check to `if ret is not None:`, and 2) then add a (then simpler) integration test where you just invoke `server.replay.count` immediately after startup? 

We currently underutilize these sorts of tests, but you can find three examples at the bottom of
https://github.com/mitmproxy/mitmproxy/blob/main/test/mitmproxy/tools/console/test_integration.py. Just adding one below that would be fine. Let me know if something does not make sense. :)

mhils (approved):
Excellent, thank you so much! 😃 🍰

shindexro:
Thanks for the detailed guidance! 😄

PR #4910, opened 2021-11-18
Quote argument of view.flows.resolve

### LINKED ISSUE #4899: Command argument mismatch when executing command:  ":view.flows.remove '~m POST'"
I am trying to delete all the POST flows.

I execute the command

`:view.flows.remove '~m POST'`

And get the following error

```
error: Command argument mismatch:
    Expected: OrderedDict([('flow_spec', <Parameter "flow_spec: str">)])
    Received: ['~m', 'POST']
```

Filtering the flows via `f` keybind works

#### System Information
Mitmproxy: 7.0.4
Python:    3.9.7
Platform:  macOS-11.6-x86_64-i386-64bit

shindexro:
#### Description

Fix #4899.
Quote the argument of internal `view.flows.resolve` calls.

Not sure what is an effective way to test it. The internal `view.flows.resolve` happens after the user invokes `:view.flows.remove ...` Should the test be written in `test_view.py`, `test_types.py` or `test_integration.py`?

#### Checklist

### DIFF
--- a/mitmproxy/types.py
+++ b/mitmproxy/types.py
@@ -369,7 +369,7 @@ class _FlowType(_BaseFlowType):
 
     def parse(self, manager: "CommandManager", t: type, s: str) -> flow.Flow:
         try:
-            flows = manager.execute("view.flows.resolve '%s'" % (s))
+            flows = manager.call_strings("view.flows.resolve", [s])
         except exceptions.CommandError as e:
             raise exceptions.TypeError(str(e)) from e
         if len(flows) != 1:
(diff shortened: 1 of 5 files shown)


mitmproxy/types.py
```diff
@@ -369,7 +369,7 @@ class _FlowType(_BaseFlowType):
 
     def parse(self, manager: "CommandManager", t: type, s: str) -> flow.Flow:
         try:
-            flows = manager.execute("view.flows.resolve %s" % (s))
+            flows = manager.execute("view.flows.resolve '%s'" % (s))
```
mhils, line 372: This looks better than before, but we'll now likely run into issues with `'` characters in the spec. Maybe we can just use `manager.call_strings` instead?

mhils:
> Not sure what is an effective way to test it. The internal `view.flows.resolve` happens after the user invokes `:view.flows.remove ...` Should the test be written in `test_view.py`, `test_types.py` or `test_integration.py`?

`test_types.py` seems most appropriate, I think we already have some tests there that just need additional cases for flowspecs with spaces. 👍

mhils (approved):
Awesome. Thank you very much for the great PRs in the last few days! 🍰 😃

PR #4928, opened 2021-11-20
Add font types to asset filter (~a)

elespike:
#### Description

Add 2 font MIME type patterns to the asset filter (~a) to both `mitmproxy` and `mitmweb`:
- `font/*`
- `application/font-*`

#### Checklist

### DIFF
--- a/mitmproxy/flowfilter.py
+++ b/mitmproxy/flowfilter.py
@@ -15,7 +15,6 @@
                         application/javascript
                         text/css
                         image/*
-                        application/x-shockwave-flash ?
                         font/*
                         application/font-*
         ~h rex      Header line in either request or response
--- a/mitmproxy/flowfilter.py
+++ b/mitmproxy/flowfilter.py
@@ -15,7 +15,9 @@
                         application/javascript
                         text/css
                         image/*
-                        application/x-shockwave-flash
+                        application/x-shockwave-flash ?
+                        font/*
+                        application/font-*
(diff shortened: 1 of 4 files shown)


mitmproxy/flowfilter.py
```diff
@@ -15,7 +15,9 @@
                         application/javascript
                         text/css
                         image/*
-                        application/x-shockwave-flash
+                        application/x-shockwave-flash ?
```
elespike, line 18: Should flash be included in the patterns too? It currently isn't.

elespike:
Looks like docs and mitmweb's JS files won't need manual changes, correct?

Cool to see you use pyparsing! I've used it in one of my tools as well =)

mitmproxy/flowfilter.py
mhils: Thanks for catching this leftover. Flash nowadays would be a rather interesting find than anything else, we removed it from the asset list a while ago.

mhils (approved):
Thanks for the excellent PR. I really appreciate that you properly looked into this, found the Flash leftover, and also updated the tests. 🍰 

I have reverted the test change as it IMO does more harm than good. The current test is short and explicit, which are desirable properties. There is no benefit in automatically checking more regexes in what is more of a self-fulfilling prophecy. I appreciate you looking into this anyway, hope my reasoning makes sense. 😃

elespike:
Makes sense, and my pleasure =)

PR #4943, opened 2021-11-22
Created howto on patching android application to accept user CA

Grijnak:
#### Description

I figured out how to patch an app that would refuse to run on a rooted device to accept the user CA certificate store. I hope others find this useful as well.

#### Checklist

### DIFF
--- a/docs/src/content/howto-make-android-apk-trust-userca.md
+++ b/docs/src/content/howto-make-android-apk-trust-userca.md
@@ -1,0 +1,60 @@
+---
+title: "Make Android app trust user CA"
+menu:
+    howto:
+        weight: 4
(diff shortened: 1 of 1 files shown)


mhils:
Thanks! We already have a short section on certificate pinning here: https://docs.mitmproxy.org/dev/concepts-certificates/#certificate-pinning

How does your approach compare to using apk-mitm or objection?

Grijnak:
I looked into apk-mitm and it is exactly what I describe here but automated... Well at least I know of an even better way now. I hadn't looked into the certificate pinning section, because I understood it to mean certificates that are packaged with the app, which was not relevant in my case.

Issue #2964, opened 2018-03-06
Transparent mode failure: AttributeError("'Resolver' object has no attribute 'wfile'",)


GHXST01:
When trying to start mitmproxy in transparent mode and a custom CA directory as followed:
` .\mitmweb.exe --set cadir=C:\Users\Admin\Documents\_custom_cert`

I get the following output:
```
192.168.178.40:46345: Transparent mode failure: AttributeError("'Resolver' object has no attribute 'wfile'",)
192.168.178.40:46345: clientdisconnect
```

I am forwarding port 80 and 443 from my custom router to mitmproxy listening on port 8080.

Any idea what is causing this?

GHXST01:
To clarify; I was having this issue on windows. I decided to move to Linux and it behaves as expected!

pravic:
That's not the solution! How to workaround this on Windows?

GHXST01:
You install Linux on on a VM or you try the built in Linux kernel.

pravic:
Not an option.

GHXST01:
No solution that I know.

PR #4951, opened 2021-11-24
Fixes AttributeError in transparent mode

lllama:
#### Description

If we haven't connected to the upstream host in transparent mode, then the call to `write` on line 74 will raise an `AttributeError` as we haven't initialised `self.wfile`.

If we add `AttributeError` to the `except` line, then we'll catch the exception and call `_connect`, which will in turn initialise `wfile`.

#### Checklist

### DIFF
--- a/mitmproxy/platform/windows.py
+++ b/mitmproxy/platform/windows.py
@@ -76,7 +76,7 @@ class Resolver:
                 if addr is None:
                     raise RuntimeError("Cannot resolve original destination.")
                 return tuple(addr)
-            except (EOFError, OSError):
+            except (EOFError, OSError, AttributeError):
                 self._connect()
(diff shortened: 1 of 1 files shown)


mhils:
Out of curiosity, how did you manage to trigger this?  😄

mhils (approved):

lllama:
Pretty sure this was just running `mitmdump` on Windows with transparent mode set. I was trying to talk to a host that required clients certs - not sure if that contributed to it. This change fixed the initial error but I then bumped up against another which I fixed by switching to use `ignore_hosts` instead of transparent mode so I didn't dig any further.

lllama:
I think this is actually related to #2964 which got closed before its time.

Issue #4916, opened 2021-11-18
Commands that return a list of flows crash when a flow has been loaded from a file


rbdixon:
#### Problem Description

Execute `view.flows.resolve @focus` under two conditions:

1. With recorded flows: overlay pops up with result
2. With flows loaded from a file using `--rfile test.mitm`: mitmproxy crashes

```
$ mitmproxy --rfile test.mitm 
Traceback (most recent call last):
  File "/Users/rbdixon/src/mitmproxy/mitmproxy/master.py", line 54, in run_loop
    loop()
  File "/Users/rbdixon/projects/library/restrike-project/.direnv/python-3.9.6/lib/python3.9/site-packages/urwid/main_loop.py", line 287, in run
    self._run()
  File "/Users/rbdixon/projects/library/restrike-project/.direnv/python-3.9.6/lib/python3.9/site-packages/urwid/main_loop.py", line 385, in _run
    self.event_loop.run()
  File "/Users/rbdixon/projects/library/restrike-project/.direnv/python-3.9.6/lib/python3.9/site-packages/urwid/main_loop.py", line 1494, in run
    reraise(*exc_info)
  File "/Users/rbdixon/projects/library/restrike-project/.direnv/python-3.9.6/lib/python3.9/site-packages/urwid/compat.py", line 58, in reraise
    raise value
  File "/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/Users/rbdixon/projects/library/restrike-project/.direnv/python-3.9.6/lib/python3.9/site-packages/urwid/raw_display.py", line 416, in <lambda>
    wrapper = lambda: self.parse_input(
  File "/Users/rbdixon/projects/library/restrike-project/.direnv/python-3.9.6/lib/python3.9/site-packages/urwid/raw_display.py", line 515, in parse_input
    callback(processed, processed_codes)
  File "/Users/rbdixon/projects/library/restrike-project/.direnv/python-3.9.6/lib/python3.9/site-packages/urwid/main_loop.py", line 412, in _update
    self.process_input(keys)
  File "/Users/rbdixon/projects/library/restrike-project/.direnv/python-3.9.6/lib/python3.9/site-packages/urwid/main_loop.py", line 513, in process_input
    k = self._topmost_widget.keypress(self.screen_size, k)
[...]

PR #4952, opened 2021-11-24
Add deepcopy to allow view.flows.resolve (issue #4916)

rbdixon:
#### Description

Add a `__deepcopy__()` method so that when an intercepted flow or a flow loaded from a file is displayed using `view.flows.resolve` the flows are able to be deepcopy()'ed.

#### Checklist


mhils:
Thanks! Could you check if my simplified workaround covers all the cases just as well please? :)

rbdixon:
Looks good and you simplified the fix, too. Thanks!

PR #4997, opened 2021-12-17
Example specified incorrect header

jamesyale:
#### Description

Trivial documentation fix, the `User-Agent` header modification [example](https://docs.mitmproxy.org/stable/overview-features/#examples-4) was incorrect, specifying the `Host` header. Copy/paste error by the looks of it. 

#### Checklist


mhils (approved):
Good catch, thanks! 🍰

PR #4621, opened 2021-06-04
Added `wsreplay` as a temporary fix of mitmproxy not being able to replay websockets

KOLANICH:
#### Description

Added a standalone mitigation to mitmproxy not being able to replay websockets

#### Checklist

 - [x] I have updated tests where applicable. - not applicable, `contrib` goes without tests
 - [ ] I have added an entry to the CHANGELOG. - not sure if it should be done

PR #5012, opened 2021-12-23
Created a list of standalone tools that can be considered examples and added `wsreplay` into it.

KOLANICH:
Supersedes: #4621

#### Description
Added a link to a tool by me.
#### Checklist

 - [ ] I have updated tests where applicable - not needed
 - [ ] I have added an entry to the CHANGELOG. - IMHO not needed

### DIFF
--- a/examples/contrib/README.md
+++ b/examples/contrib/README.md
@@ -3,6 +3,10 @@
 Examples in this directory are contributed by the mitmproxy community.  
 If you developed something thats useful for a wider audience, please add it here!
 
+# List of standalone tools that can be considered examples
+
+The list is in [the TSV file](./standalone.tsv).
+
 # Maintenance
 
 
--- a/examples/contrib/standalone.tsv
+++ b/examples/contrib/standalone.tsv
@@ -1,0 +1,2 @@
+link	description
+https://github.com/KOLANICH-tools/wsreplay.py	While mitmproxy is currently unable to replay websocket streams, I have created a simple and incomplete standalone (but depending on mitmproxy) tool to do this.


mhils:
Very cool, thank you very much for sharing that! Let's make sure we get enough visibility for this, currently it feels a bit tucked away in `standalone.tsv`. Do you want to a) rename `standalone.tsv` to `websocket-replay.md` and then add a sentence describing what it does + link to your repo, or b) start a list in the README directly? I'm happy with either option, whatever you think helps discoverability the most. :-)


On that note - thank you for your fantastic work on Kaitai! 🍰 I'm always super happy to see devs working on projects we depend on (Kaitai) make use of mitmproxy! 😃

KOLANICH:
>Let's make sure we get enough visibility for this, currently it feels a bit tucked away in standalone.tsv.

Yeah. The purpose of this to make the list machine-readable, so a fancy table for the website can be generated automatically.

>Do you want to a) rename standalone.tsv to websocket-replay.md and then add a sentence describing what it does + link to your repo, or b) start a list in the README directly?

In previous PR it was proposed to create an md file, but I felt that in a separate file it'd be too little visible, noone would ever read that file. So my second thought was to put it into the ReadMe. But I thought it was not a very good idea, because the purpose of that readme was a bit different. And reading a file for just 1 short line of text - IMHO it is an overkill. So I thought we need a separate page, featuring such examples and tools. So there are options, either to use Markdown + sections and lists, or to generate the page from a more structured description. I felt like the latter is better. Of course we want the base be both human and machine readable and editable. So the choice is between TSV and YAML. I guess YAML may be better if we need more optional fields, but for now we need no, so I decided to keep it simple for now and use TSV.

About putting the impl details here ... the tool is in a separate repo and the description may get obsolete.

mhils:
Fixed with https://github.com/mitmproxy/mitmproxy/commit/648d3bc0f9fea6331976f4f67f51dcf5b26bff92, thanks again! 🍰

PR #5027, opened 2021-12-31
Feature #5019 - allow context configuration changes in tls_* hooks

### LINKED ISSUE #5019: A hook for certificate verification overrides
#### Problem Description

It seems there is currently no way to tell mitmproxy to accept individual invalid server certificates¹, `--ssl-insecure`/`--no-ssl-insecure` is as granular as it currently gets. This is a problem for people that use mitmproxy for casual browsing and occasionally stumble upon some *non-critical* websites (like, say, blogs) with annoyances like slightly outdated certs.

¹ Or maybe even discard valid ones? I'm not sure if anyone would find such feature useful though, so I'll speak about verification overrides for the rest of this post.

#### Proposal

A hook that will let addons choose `Verify.VERIFY_NONE` or `Verify.VERIFY_PEER` for each individual certificate and/or host.

#### Alternatives

- Extend the `ssl_insecure` option so that it could also accept a list of filters of some sort.
  - This will definitely look like a mess if exposed via cmdline.
- Alternatively `ssl_insecure` could accept path to file with a list of overrides.
  - It seems it would make addons that implement "accept the risk" page or web-UI for overrides mgmt, well, a lot more hacky.
  - As a bonus, however, this may allow users to reuse overrides stored in their Firefox profiles (`cert_override.txt`). (Can't say anything about WebKit/Blink-based browsers though.)

#### Additional context

[Gecko's `nsCertOverrideService.cpp`](https://github.com/mozilla/gecko-dev/blob/fce6eb7d8380f3e98429d1f5096d3a0dd0f412d2/security/manager/ssl/nsCertOverrideService.cpp) describes override matching process, as well as format of `cert_override.txt`, which basically boils down to: `host:port`, SHA-256 fingerprint, error kinds (any combination of `M`ismatch, `U`ntrusted, `T`ime), and DB key, which itself [is generated from serial and DN](https://github.com/mozilla/gecko-dev/blob/fce6eb7d8380f3e98429d1f5096d3a0dd0f412d2/security/manager/ssl/nsNSSCertificateDB.cpp#L94-L103).

jsmucr:
#### Description

Creating a new `Context` now clones the `Options` passed to it so that these options can be changed later in the process without changing anything globally.

This makes it possible to set for example `ssl_insecure` for each request individually from within the `tls_start_server` hook.

```python
from mitmproxy import ctx

def tls_start_server(data: any):
    address = data.context.server.address[0] + ":" + str(data.context.server.address[1])
    if address == 'as2.pipechain.com:10443':
        data.context.options.ssl_insecure = True
    ctx.log.info("ssl_insecure[" + address + "] => " + str(data.context.options.ssl_insecure))
```

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * Fix bug that crashed when using `view.flows.resolve` (#4916, @rbdixon)
 * Fix a bug where `running()` is invoked twice on startup (#3584, @mhils)
 * Correct documentation example for User-Agent header modification (#4997, @jamesyale)
+* Allow context configuration changes in `tls_*` hooks (#5019, @jsmucr)
 
 ## 28 September 2021: mitmproxy 7.0.4
 
--- a/mitmproxy/addons/tlsconfig.py
+++ b/mitmproxy/addons/tlsconfig.py
@@ -172,7 +172,9 @@ class TlsConfig:
         server: connection.Server = tls_start.context.server
         assert server.address
 
-        if ctx.options.ssl_insecure:
+        options = tls_start.context.options
+
+        if options.ssl_insecure:
             verify = net_tls.Verify.VERIFY_NONE
         else:
             verify = net_tls.Verify.VERIFY_PEER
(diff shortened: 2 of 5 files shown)


mhils:
Thanks for the PR! The changes look good and you put in effort to update the changelog and tests, which is great. I really do appreciate it. Unfortunately I'm strongly -1 on the architectural changes here. mitmproxy has one global options object shared by all addons and the core. Introducing multiple option states makes it harder to reason about the code and will lead to unpredictable side effects long term. So this is a hard pass - sorry. I'm more than happy to discuss other ways how we can implement certificate verification overrides, let's continue the discussion in #5019.

jsmucr:
No problem @mhils. :-) Thank you for reviewing the PR.

PR #5039, opened 2022-01-07
Help bar added

koknaroglu:
#### Description

I have added the help bar on statusbar.py/redraw().
I'm having problems editing the marked info ([2/3]) that I need to add to the flows header.

#### Checklist


mhils:
Hi there! I appreciate the intent, but this doesn't cross the bar for a useful contribution. It looks like you have merely removed the logic for the footer and inserted some static text instead. I'm afraid we don't have the capacity to mentor you from your current skill level to a successful PR - sorry!

PR #5051, opened 2022-01-12
Non-functional demaonstrator for the concept of a global tweaks instance

EndUser509:
#### Description

A non-functional POC for how a tweaks class could be used for low-level changes at runtime. Intended to be used with `-s mitmproxy/script.py`.


Prinzhorn:
I wonder how tweaks are different from options? You can enable/disable options from within an add-on as well and if I understand correctly tweaks are meant to be _only_ available to add-ons? But what if I want `--set no_strict_http2_headers` without needing a script?

PR #5061, opened 2022-01-15
add keyboard shortcut n to mitmweb (by using runCommand inside the kb shortcut handler)

ianklatzco:
#### Description

Following the behavior of mitmproxy, this adds an `n` keybind to mitmweb, which creates a new flow in the FlowView:
image

Which can then be clicked, edited, and `r`'d to be sent.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 * Fix a bug where `running()` is invoked twice on startup (#3584, @mhils)
 * Correct documentation example for User-Agent header modification (#4997, @jamesyale)
 * Fix random connection stalls (#5040, @EndUser509)
+* Add `n` new flow keybind to mitmweb (#5061, @ianklatzco)
 
 ## 28 September 2021: mitmproxy 7.0.4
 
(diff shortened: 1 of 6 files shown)


ianklatzco:
Last failing check is an unreachable URL validation; currently the /create endpoint doesn't have any way to specify a URL (example.com is hardcoded, and then users edit the desired URL in the web UI).

I'll probably add a way for POST /flows/create to take a URL and then write a test, when I touch this next.

mhils:
👍 on adding this, thanks for the PR! 😃 

We could potentially remove a lot of boilerplate by just using `runCommand("view.flows.create", "get", "https://example.com/")` either in `flows.ts:create` or directly in `keyboard.tsx:onKeyDown`. The functionality to directly call mitmproxy [commands](https://docs.mitmproxy.org/dev/concepts-commands/) via the `POST /commands` endpoint is fairly new, which is why most of the existing code goes through the trouble of defining new routes etc. Alternatively we should at least just invoke the command in your newly defined `CreateFlow` endpoint. Either way is fine for me, let me know what you think. :)

(also feel free to ask stupid questions here or on Slack. Based on your PR I'm assuming quite a bit of familiarity with everything already, happy to explain stuff in more detail :))

ianklatzco:
All zipped up and rebased. Should be good to go! Please merge at your convenience if everything looks okay.

mhils:
This looks super straightforward, excellent! 🎉
It could be useful to add `File` -> `New Flow` as well so that it's easier to discover, but let's leave that for a new PR! 😃

Issue #5010, opened 2021-12-21
String options are read from the config file incorrectly


marwinxxii:
#### Problem Description
In version `7.0.4` options with type `str` and `Optional[str]` are read incorrectly from the config file, only last character is used.

#### Steps to reproduce the behavior:
1. Create add-on `test_option_addon.py`:
```python
from typing import Optional
from mitmproxy import ctx

class OptionTestAddon:
    def load(self, loader) -> None:
        loader.add_option(
            name="test_option",
            typespec=Optional[str],
            default=None,
            help="Will it fail?"
        )

    def configure(self, updates) -> None:
        print("test_option value: " + ctx.options.test_option)

addons = [OptionTestAddon()]
```
2. in `~/.mitmproxy/config.yaml` define:
```yaml
test_option: abcdef
```
3. Run `mitmdump -s test_option_addon.py` and it prints:
```
test_option value: f
Loading script ../../../zmisc/option_test_addon.py
Proxy server listening at http://*:8080
```

while `abcdef` is expected.

I've tried wrapping value in the quotes, but the result is the same.

#### System Information
```
Mitmproxy: 7.0.4
Python:    3.9.1
OpenSSL:   OpenSSL 1.1.1l  24 Aug 2021
Platform:  macOS-10.16-x86_64-i386-64bit
```

mhils:
Thanks for the clear and concise report, clearly looks like a bug. Does specifying them as command line args with `--set` work?

marwinxxii:
@mhils yes, as command line argument it works fine:

`mitmdump -s ../../../zmisc/option_test_addon.py --set test_option=abcdef` (when option in the config file is commented out).

produces:
```
test_option value: abcdef
Loading script ../../../zmisc/option_test_addon.py
Proxy server listening at http://*:8080
```

marwinxxii:
I've tried following test on both `main` and `v7.0.4` and it passes successfully, probably I'm missing something:

`test_optmanager.py`:
```python
def test_read_config(tmp_path):
    config_path = tmp_path / "config.yaml"
    config_yaml_text = """
    three: value_three
    """
    with open(config_path, "w") as f:
        f.write(config_yaml_text)
    manager = TD2()
    optmanager.load_paths(manager, config_path)
    assert manager.three == "value_three"
```

mhils:
```
λ cat ~/.mitmproxy/config.yaml
intercept: foo
```

[...]

Issue #4047, opened 2020-06-27
Addons in mitmweb stops working when an addon adds a boolean option and that option is set in config.yml


andreas-wolf:
#### Problem Description
When an addon uses add_option to add a boolean and that boolean is set in config.yml all addons stop working.
I've tried the same with a string option, that was working ok. 
My assumption is that the yaml file is not parsed correctly. The following entries should be possible for a boolean: https://yaml.org/type/bool.html

Sample Addon optionstest.py:

```python
from mitmproxy import ctx
class OptionsTest:
    def load(self, loader):
        loader.add_option(
            name = "abool",
            typespec = bool,
            default = False,
            help = "A boolean",
        )

    def configure(self, updates):
        ctx.log.info('A boolean: %s' % ctx.options.abool)

    def response(self, flow):
        ctx.log.info('A boolean: %s' % ctx.options.abool)

addons = [
    OptionsTest()
]
```

sample config.yaml
`abool: true`

#### Steps to reproduce the behavior:
1. start mitmweb -s optionstest.py without a config.yaml -> see logmessages
2. add sample config.yaml
3. start mitmweb -s optionstest.py again -> no log messages

#### System Information
Mitmproxy: 5.1.1
Python:    3.8.3
OpenSSL:   OpenSSL 1.1.1g  21 Apr 2020
Platform:  Linux-5.4.0-39-generic-x86_64-with-glibc2.10

mhils:
Thanks for the clean minimal example! This definitely looks like a bug. Is mitmdump/mitmproxy affected, too?

andreas-wolf:
I assume so, since it is the config. I have no time to test more right now.

mhils:
This should be fixed in mitmproxy 8! 😃

PR #5067, opened 2022-01-16
Refactor how we process `--set` options

mhils:
refs #5056

### DIFF
--- a/mitmproxy/addons/core.py
+++ b/mitmproxy/addons/core.py
@@ -50,18 +50,19 @@ class Core:
                     )
 
     @command.command("set")
-    def set(self, option: str, value: str = "") -> None:
+    def set(self, option: str, *value: str) -> None:
         """
             Set an option. When the value is omitted, booleans are set to true,
             strings and integers are set to None (if permitted), and sequences
             are emptied. Boolean values can be true, false or toggle.
             Multiple values are concatenated with a single space.
-            Sequences are set using multiple invocations to set for
-            the same option.
         """
-        strspec = f"{option}={value}"
+        if value:
+            specs = [f"{option}={v}" for v in value]
+        else:
+            specs = [option]
         try:
-            ctx.options.set(strspec)
+            ctx.options.set(*specs)
         except exceptions.OptionsError as e:
             raise exceptions.CommandError(e) from e
 
--- a/mitmproxy/optmanager.py
+++ b/mitmproxy/optmanager.py
@@ -1,14 +1,14 @@
-import collections
 import contextlib
-import blinker
-import blinker._saferef
-import pprint
 import copy
+from dataclasses import dataclass
 import functools
 import os
-import typing
+import pprint
 import textwrap
+import typing
 
+import blinker
+import blinker._saferef
 import ruamel.yaml
 
 from mitmproxy import exceptions
@@ -79,6 +79,11 @@ class _Option:
         return o
 
 
+@dataclass
+class _UnconvertedStr:
+    val: typing.List[str]
+
+
 class OptManager:
     """
         OptManager is the base class from which Options objects are derived.
@@ -92,7 +97,7 @@ class OptManager:
         mutation doesn't change the option state inadvertently.
     """
     def __init__(self):
-        self.deferred: typing.Dict[str, typing.List[str]] = {}
+        self.deferred: typing.Dict[str, typing.Any] = {}
         self.changed = blinker.Signal()
         self.errored = blinker.Signal()
         # Options must be the last attribute here - after that, we raise an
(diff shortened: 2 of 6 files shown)


mitmproxy/addons/core.py
```diff
@@ -50,18 +50,19 @@ def configure(self, updated):
...
     @command.command("set")
-    def set(self, option: str, value: str = "") -> None:
+    def set(self, option: str, *value: str) -> None:
         """
             Set an option. When the value is omitted, booleans are set to true,
             strings and integers are set to None (if permitted), and sequences
             are emptied. Boolean values can be true, false or toggle.
             Multiple values are concatenated with a single space.
-            Sequences are set using multiple invocations to set for
-            the same option.
```
mhils, line 60: This explanation here was bonkers (and never worked), the previous line already explains how to set multiple values.

mitmproxy/optmanager.py
```diff
@@ -288,60 +293,77 @@ def __repr__(self):
...
+        processed: typing.Dict[str, typing.Any] = {}
+        for name in list(unprocessed.keys()):
+            if name in self._options:
+                processed[name] = self.parse_setval(self._options[name], unprocessed.pop(name))
+
+        if defer:
+            self.deferred.update({
+                k: _UnconvertedStr(v)
+                for k, v in unprocessed.items()
+            })
```
mhils, line 320: We now explicitly mark unconverted strings that go into deferred so that they can be parsed when the final option type is known.

mitmproxy/tools/main.py
```diff
@@ -76,6 +76,7 @@ def run(
         sys.exit(1)
 
     try:
+        # TODO: `--set` should take precedence over config.yaml, but we do this first for now as it may modify confdir.
```
marwinxxii, line 79: I've created an issue to track this and not forget https://github.com/mitmproxy/mitmproxy/issues/4047

mitmproxy/addons/core.py
```diff
@@ -50,18 +50,19 @@ def configure(self, updated):
...
         """
-        strspec = f"{option}={value}"
+        if value:
+            specs = [f"{option}={v}" for v in value]
+        else:
+            specs = [option]
```
marwinxxii, line 63: Shouldn't this be `f"{option}"`?

mitmproxy/optmanager.py
```diff
@@ -288,60 +293,77 @@ def __repr__(self):
...
-        if defer:
-            self.deferred.update(unknown)
-        elif unknown:
-            raise exceptions.OptionsError("Unknown options: %s" % ", ".join(unknown.keys()))
-        self.update(**vals)
+                unprocessed.setdefault(spec, [])
 
-    def process_deferred(self):
+        processed: typing.Dict[str, typing.Any] = {}
+        for name in list(unprocessed.keys()):
```
marwinxxii, line 312: maybe it would be easier to understand if this is done in a single loop?

mitmproxy/addons/core.py
mhils: `option` can only be a str, or am I missing something?

mitmproxy/tools/console/options.py
```diff
@@ -181,8 +181,7 @@ def keypress(self, size, key):
                 foc, idx = self.get_focus()
                 v = self.walker.get_edit_text()
                 try:
-                    current = getattr(self.master.options, foc.opt.name)
-                    d = self.master.options.parse_setval(foc.opt, v, current)
+                    d = self.master.options.parse_setval(foc.opt, [v])
```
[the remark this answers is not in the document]
mhils: Not sure if I'm following. Could you explain what you are referring to? This here only handles non-sequence types.

mitmproxy/addons/core.py
```diff
@@ -50,18 +50,19 @@ def configure(self, updated):
...
             Set an option. When the value is omitted, booleans are set to true,
             strings and integers are set to None (if permitted), and sequences
             are emptied. Boolean values can be true, false or toggle.
             Multiple values are concatenated with a single space.
-            Sequences are set using multiple invocations to set for
-            the same option.
         """
-        strspec = f"{option}={value}"
+        if value:
+            specs = [f"{option}={v}" for v in value]
```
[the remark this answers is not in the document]
mhils: This is a good point, thanks - let's add something to OptManager and then do the `key=value` -> dict preproccessing only for CLI args.

mitmproxy/addons/core.py
marwinxxii: right, I thought for a moment that `option` is of Option type, not a name.
I think such a confusion could be avoided if we have a method mentioned in https://github.com/mitmproxy/mitmproxy/pull/5067/files#r786017760

mitmproxy/tools/console/options.py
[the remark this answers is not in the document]
marwinxxii: I was thinking that if there is a `set` method which accepts a name and a value we don't need to use `update`.

In general my first impression is that we could make things consistent if we either always expect the caller to do `parse_setval` or allow two ways of setting options: parsed pair or a raw string (in which case `parse_setval` becomes a private function?)

mitmproxy/optmanager.py
mhils: We could do that, but for string sequences we'd then need to feed the current value into `parse_setval` as well. I've actually did that first (based on the code that was there), but felt it was much harder to reason about. I've now added feedback that hopefully makes the code a bit easier to understand here.

mhils:
I've made some changes based on your feedback - thanks again! We now end up funneling everything through `set=...` specs because I couldn't get the structure equally straightforward otherwise, but this should be good to go now. I will have to take a bit of a mitmproxy hiatus for the next few weeks, but this will go in right after together with your tests! Thanks again! :)

mitmproxy/optmanager.py
```diff
@@ -353,17 +382,9 @@ def parse_setval(self, o: _Option, optstr: typing.Optional[str], currentvalue: t
...
-        elif o.typespec == typing.Sequence[str]:
-            if not optstr:
-                return []
-            else:
-                if currentvalue:
-                    return currentvalue + [optstr]
-                else:
-                    return [optstr]
-        raise NotImplementedError("Unsupported option type: %s", o.typespec)
+        raise NotImplementedError("Unsupported option type: {o.typespec}")
```
marwinxxii, line 387:
```suggestion
        raise NotImplementedError(f"Unsupported option type: {o.typespec}")
```

mitmproxy/optmanager.py
marwinxxii: thanks, now it's straightforward 👍

marwinxxii:
I ran locally the test from https://github.com/mitmproxy/mitmproxy/issues/5010#issuecomment-998960629 and it passed successfully.

marwinxxii (approved):

mhils:
Thank you again for the initial PR and the very helpful reviews! 🍰

PR #5083, opened 2022-01-21
Pass length of hostname.encode() to X509_VERIFY_PARAM_set1_host

pmoulton:
#### Description

Passing zero for the size_t length argument of
X509_VERIFY_PARAM_set1_host causes MITM Proxy to crash when used with
BoringSSL.

https://www.openssl.org/docs/man1.1.1/man3/X509_VERIFY_PARAM_set1_host.html
https://boringssl.googlesource.com/boringssl/

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * Correct documentation example for User-Agent header modification (#4997, @jamesyale)
 * Fix random connection stalls (#5040, @EndUser509)
 * Add `n` new flow keybind to mitmweb (#5061, @ianklatzco)
+* Fix crash from not passing length to OpenSSL library call (pmoulton@)
 
 ## 28 September 2021: mitmproxy 7.0.4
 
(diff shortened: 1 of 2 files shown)


mhils:
Thanks! This is an interesting case in that the OpenSSL docs explicitly allow what we are doing, but BoringSSL is stricter. More than happy to accommodate BoringSSL here of course. I'm wondering if there is a good way for us to make sure that we don't regress in the future, are there cryptography-with-BoringSSL wheels readily available somewhere?

pmoulton:
It seems like BoringSSL support for cryptography is relatively new & experimental.

> 36.0.0 - 2021-11-21¶
Added experimental support for compiling against BoringSSL. As BoringSSL does not commit to a stable API, cryptography tests against the latest commit only. Please note that several features are not available when building against BoringSSL.

https://cryptography.io/en/latest/changelog/#v36-0-0

I couldn't find anything BoringSSL specific in the cryptography releases and BoringSSL didn't seem to have anything cryptography-specific either.

https://pypi.org/project/cryptography/36.0.1/#files

Anywhere else you would recommend looking?

reaperhulk:
Hey folks, I saw boring mentioned in the mitmproxy 8 release notes and went spelunking a bit to find this! We have no plans to ship boring wheels at this time, but if there’s something simple we can do to help you out here let us know. Our CI demonstrates how to build cryptography against boring, although you may not want to carry that in your own configs.

PR #5084, opened 2022-01-21
Parse set header cookies properly

Speedlulu:
#### Description

Currently when an empty cookie attribute (`Secure`, `HttpOnly` ...) is encountered while parsing a `Set-Cookie` header it will create a `CookieAttrs` object containing a (key, value) pair with an empty string for the attribute value ie:

```python
CookieAttrs[('Secure', ''), ('HttpOnly', ''), ('Path', '/')]
``` 
Resulting in an updated `Set-Cookie` header for the `Response` object with invalid values for those empty attributes ie:
```python
(b'SetCookie', b'value=XYZ; Secure=; HttpOnly=; Path=/')
``` 
My browser (Firefox 95.0.1) does not pickup these attributes so the cookie looses them.

______

This fix replaces the empty string attribute for empty cookie attributes by the value `None` ie:

```python
CookieAttrs[('Secure', None), ('HttpOnly', None), ('Path', '/')]
``` 

So that they can be told apart from attributes with intentional empty string values when setting the updated header, which results in a properly formatted header:

```python
(b'SetCookie', b'value=XYZ; Secure; HttpOnly; Path=/')
``` 

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased: mitmproxy next
 
+* Fix empty cookie attributes being set to `Key=` instead of `Key`
+  ([#5084](https://github.com/mitmproxy/mitmproxy/pull/5084), @Speedlulu)
 
 
 ## 14 November 2023: mitmproxy 10.1.5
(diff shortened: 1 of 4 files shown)


bratao:
Just a +1 from me, this fixed an website found on the wild.

mhils (approved):
💯 

Thanks! 🍰

PR #5091, opened 2022-01-25
Add support for grpc viewing and modification using protoc

### LINKED ISSUE #5001: Add support for viewing and editing protobuf content using Google's protoc package
#### Problem Description
As of now, mitmproxy only supports raw deserialization of protobuf data. There is no way to provide a set of proto files in order to get "contextual" deserialization. Editing the content is not possible either, as it  requires proto definition in order to serialize the content.

#### Proposal
I've built a few addons using mitmproxy's public API. The implementation is working based on descriptor files and is running on top of [google's protobuf package](https://googleapis.dev/python/protobuf/latest/) . See [this repo](https://github.com/aarnaut/mitmproxy-grpc) for more info. I've only tested it with the services I use, so this is still experimental. My proposal is to add these addons to the list of default ones and expose it to a wider audience. 

#### Alternatives
An alternative would be to update the existing protobuf implementation that is available in mitmproxy. I found that difficult, as the current one is an implementation of a protobuf serializer and deserializer. I don't have the context of why it was done like this, but on first glance, I find it better to use packages that provide implementation of the protobuf spec, as it requires minimal maintenance. But perhaps there was a good reason to go with a custom implementation.

aarnaut:
#### Description

Fixes #5001.
This PR addresses #5001 with some adjustments. We use text format provided by the protobuf library instead of JSON to present the protobuf content. The reason for that is the lack of ability to represent unknown fields with JSON.

The current implementation for modification only works on console. I've tried to make this more explicit in the name itself.
It relies on a command addon, which is not supported in web. Viewing works both in console and web.

I have a couple of checks failing and I'm unable to resolve them. I would appreciate if you could give me some pointers.

#### Open points
 - Current implementation only allows for uncompressed payload. I don't think there would be issues with adding this in the future. I've checked the existing `grpc` implementation and we could probably do something similar.
 - Message resolution works based on path. For example, HTTP path `example/server/GetUser` will resolve into `example.server.GetUser` and we will do an RPC lookup based on this name. I've only worked on services with this kind of mapping. I haven't managed to find any resources on whether this is a standard, so I'm a bit concerned if this method is correct. Perhaps it would make sense to introduce a way to provide custom mappings.
- I have yet to verify this, but I believe that `modify_body` will not work correctly with this extension. The reason is that protobuf is a binary format. Maybe there's a way to extend this and add this support as well.
- I think it would make sense to mark this addon/contenview as experimental for the time being. I also have to write some docs. What would be the right way to go about this?

#### Checklist


mitmproxy/utils/protoc.py
```diff
@@ -0,0 +1,112 @@
...
+    def __find_method_by_path(self, path: str) -> MethodDescriptor:
+        try:
+            # Drop the first '/' from the path and convert the rest to a fully qualified namespace that we can look up.
+            method_path = path.replace("/", ".")[1:]
+            return self.descriptor_pool.FindMethodByName(method_path)
+        except KeyError as e:
+            raise ValueError("Failed to resolve method name by path") from e
+
+
+serializer = ProtocSerializer()
```
aarnaut, line 112: `ProtocSerializer` is used in both addons and the contentview, so I need to instantiate it earlier. I don't work with python on a daily basis, so I'm unsure if the approach I took is correct. Feel free to propose something better.

aarnaut:
Hey @mhils, should I request a review on this from someone specific? :)

mhils:
@aarnaut: thanks for the great work here. I'm not purposefully ignoring your PR, I'm just in the final stretches of my PhD. 😅 Operations should continue normally next week. Of course if anyone else wants to provide some feedback before that (@mame82, @Prinzhorn), please do. :)

aarnaut:
Hey, no worries! I just wanted to make sure there are no actions left from my side 😄 
Take as long as you need!

Losiowaty:
Hey @mhils @aarnaut 👋 
Sorry for resurrecting this PR out of the blue but I wanted to ask what is the status of it?
I'd really love to see this merged and released (and I guess/hope there may be others like me 😅 ).
Although I don't have much python experience, maybe I could help somehow?

And since I'm already resurrecting this - what do you think about allowing to set multiple descriptor paths?
I'm asking about this, since my use case actually requires that since we have `import "google/protobuf/field_mask.proto";` in our `*.proto` file.
I'm not a grpc/protobuf expert either, but I guess more complicated setups may also have more complicated `import` hierarchies.

I was actually able to patch the [original addon](https://github.com/aarnaut/mitmproxy-grpc) by @aarnaut and it doesn't require that much changes - the core of them being changing
```python
def load(self, loader):
        loader.add_option(
            name = "descriptor",
            typespec = typing.Optional[str],
            default = None,
            help = "Set the descriptor file used for serialiation and deserialization of protobuf content",
        )
```
to
```python
def load(self, loader):
        loader.add_option(
            name = "descriptor",
            typespec = collections.abc.Sequence[str],
            default = [],
            help = "Set the descriptor file used for serialiation and deserialization of protobuf content",
        )
```
and then obviously iterating over the input.

So - what do you think?

Cheers,
Paweł

aarnaut:
Hey, I haven't touched the PR in a long time, so feel free to take it over as well. Your change is a reasonable addition. There are still some open points to talk about before merging. I'm not sure how much of a priority this PR is, but @mhils can provide more info.

mhils:
Sorry for taking so long. #7623 will fix this! 🎉

PR #5097, opened 2022-01-26
Rotate stream files

EndUser509:
#### Description

This PR adds support for rotating saved stream files every hour or day. The path supplied via `--save-stream-file` is now passed through strftime(). Directories in the path are created on the fly. Flows are stored in files according to the end time, not the start time.

#### Checklist


EndUser509:
Any advice on what I can do to pass the two failed checks here?

mhils:
> Any advice on what I can do to pass the two failed checks here?

All files in `mitmproxy/addons` are required to have 100% test coverage by a matching file in `test/mitmproxy/addons`. :)

EndUser509:
OK, I'll leave this open and try to add the missing tests. 
What are the chances that this PR will be accepted once that all checks pass?

mhils:
I thought i left a review, but apparently that got lost:

1. I think I'd be happy to merge a PR that adds support for `strftime`-style placeholders.
2. Ideally we can add that with much less complexity than what is introduced in this PR. 
3. I don't really understand what `save_stream_period` does, but I haven't looked closely.

EndUser509:
I added the missing tests but also used `pragma: no cover` where tests were too complicated. Especially code sections depending on the actual current time. I hope it is enough for your quality control though.

EndUser509:
> I thought I left a review, but apparently that got lost:
> 
>     1. I think I'd be happy to merge a PR that adds support for `strftime`-style placeholders.

That is actually only a means to an end: To save flows in individual stream files for every hour or day.

>     2. Ideally we can add that with much less complexity than what is introduced in this PR.

If you want to get rid of detailed error reporting you can strip it down a little. A python crack can optimize a few statements maybe.
 
>     3. I don't really understand what `save_stream_period` does, but I haven't looked closely.

The point is that every hour or day the flows are saved into a new file. Maybe there is a better term than period, but in the end each stream file contains the flow of one or 24 hours. 

While running I think it is not a good idea to terminate just because you cannot save to a file anymore. If a new stream file cannot be opened it logs an error.

>     4. My mitmproxy capacity will be really limited in the next few weeks, so please don't expect a speedy review.

NP, I leave things like that, take your time. I move on to getting a POC for HTTP/2 PING via times.

mitmproxy/addons/save.py
```diff
@@ -54,6 +108,16 @@ def configure(self, updated):
                     raise exceptions.OptionsError(str(e)) from e
             else:
                 self.filt = None
+        if "save_stream_period" in updated:
+            if ctx.options.save_stream_period:
+                if ctx.options.save_stream_period == 'hour':
```
ericbeland, line 109: Instead of two hard-coded options, what about save_stream_hours as the number of hours for rotation?   24 for a day, or 0.25 for every fifteen minutes? An hour of video might be too much in some scenarios?

EndUser509: That would work too, although it would get tedious for 10 minute intervals. In this case I would suggest to go with `save_stream_minutes` or `save_stream_seconds` to keep the supplied number an integer. We could discuss if sub-minute resolution is really a use case but I would be fine with seconds.

Prinzhorn: There surely is some standard way in Python to have human readable interval format (like CRON but in nice). We do the same in other options, e.g. you can specify `3m` in `stream_large_bodies` for 3 megabytes. I'm sure we're not the first that want something like `24h` oder `15min` and have it parse into a duration/interval.

mhils:
> The point is that every hour or day the flows are saved into a new file. Maybe there is a better term than period, but in the end each stream file contains the flow of one or 24 hours.

Maybe I'm misunderstanding things (I also haven't looked to deeply), but my understanding was that if I specify
``` 
logs/%Y-%m-%d.mitm
```
as the output filename, mitmdump would automatically rotate to `logs/$today.mitm` at midnight. So if I want an hourly file I just put that as part of the filename. What do I need `save_stream_period` for? I think `strftime` is cheap enough to be called on every flow, the rest of the serialization is much more work.

mhils:
I think it would be a good idea to `sys.exit(1)` if we fail to write to the stream file, this is really a scenario where I as a user want to be notified visibly.

> This file may actually remain empty 

So be it, let's not overengineer things. :)

EndUser509:
Just noticed #4888 which concludes to exit on any `OSError` in `save.py`. My current implementation gracefully logs all errors but does not exit. My preference is to keep the proxy running, keeping functionality for clients. But we have two conflicting PRs here.

EndUser509:
OK, then I'll rewrite it with string comparison after `strftime()` as trigger. And with hard failure. It will take a moment to test it on my production system first and update the test cases. I hope `pragma: no cover` is fine for you for the code branches with `exit(1)`. :)

mhils:
You can use purest.raises with SystemExit, and the monkeypatch fixture to trigger an OSError. 😉

EndUser509:
OK, this version is I think ready and does what was discussed above. I had to regroup some functions but I think its compact and robust this way. Tests are also updated.

mhils:
@EndUser509: I've simplified the logic a bit. Could you please review to see if I missed anything?

mhils:
fixes #4888

EndUser509:
> @EndUser509: I've simplified the logic a bit. Could you please review to see if I missed anything?

Looks nice and compact, more than I could have done. I didn't spot anything, the code still seems to do what it is supposed to.
I am running this version of `save.py` now and so far no problems.

PR #5098, opened 2022-01-26
Save streamed data

EndUser509:
#### Description

This is a PR for a new sample addon that saves streamed data in individual files. It complements `save.py` in a way and allows to store all data passing through.
Documentation is in the addon.

#### Checklist


EndUser509:
Here too, some advice on how to pass all test would help. :)

PR #5099, opened 2022-01-26
Save streamed data2

EndUser509:
#### Description

This is a PR for a new sample addon that saves streamed data in individual files. It complements `save.py` in a way and allows to store all data passing through.
Documentation is in the addon.

#### Checklist


EndUser509:
Here you go. I was wondering if I missed some cleanup or error handling. So far I have not noticed that it leaks file handles which would be a lot worse then just a little memory.

vaibhavpacharya (approved):
👍

PR #5100, opened 2022-01-27
an ntlm upstream addon for http \ https

mrsiano:
Ntlm upstream proxy for http https all http methods
a monkey patching, to override the upstream handshake in order to inject the ntlm handshake flow
#### Description

ntlm upstream proxy caps 

#### Checklist

I did not did this further steps


mrsiano:
@mhils ptal

PR #5102, opened 2022-01-31
Pass the flow to the stream modify function

Prinzhorn:
#### Description

I need the flow and I think as second parameter it's not a breaking change and won't hurt?
Do I even need to write a test for this?

I'm currently doing this:

```py
def responseheaders(flow):
    if flow.response.headers.get('content-type', '').startswith('text/event-stream'):
        def modify(data: bytes) -> Union[bytes, Iterable[bytes]]:
            # Do something here that requires the flow to decide what to do with data.
            print(flow.id, data)
            return data

        flow.response.stream = modify
```

which doesn't look Python (JS callbacks anyone?) and also leaks memory because when the flow is done processing `stream` keeps pointing to the function that is never being called any longer.

#### Checklist

### DIFF
--- a/mitmproxy/http.py
+++ b/mitmproxy/http.py
@@ -250,7 +250,7 @@ class Message(serializable.Serializable):
         self.data.set_state(state)
 
     data: MessageData
-    stream: Union[Callable[[bytes], Union[Iterable[bytes], bytes]], bool] = False
+    stream: Union[Callable[[bytes, flow.Flow], Union[Iterable[bytes], bytes]], bool] = False
     """
     This attribute controls if the message body should be streamed.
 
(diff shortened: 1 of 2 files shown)


mhils:
This breaks existing stream functions that only expect one argument to be given. I don't think memory leak is the right characterization. It does consume a bit more memory, but it should be freed together with the flow - which means it does not leak. If you dislike the inline method, you can also use `functools.partial`:

```python
def modify(data: bytes, flow: http.HTTPFlow) -> Union[bytes, Iterable[bytes]]:
    ...

def responseheaders(flow):
    flow.response.stream = functools.partial(modify, flow=flow)
```

Prinzhorn:
TIL about `functools`. At this rate I'll learn Python in no less than two decades :+1:

PR #5106, opened 2022-02-02
Support async hooks. Fixes #4207.

### LINKED ISSUE #4207: async events in addons
It would be a nice feature if the mitm event's can be run as async functions, this will enable developers of addon's to use async methods. 

Make the event's async!

I've tried to schedule futures in the current running eventloop but it seems that it doesn't apply changes to the HTTP flow object. 

#### Additional context
Would love to see async possibilities, that can make some advanced addons more efficient and extensible

nneonneo:
#### Description

Add support for async hooks to AddonManager, enabling the proxy server to call hooks asynchronously. Uses `inspect.isawaitable` to ensure that existing sync hooks continue to work without any change. Fixes #4207.

The forking of `invoke_addon` and `trigger` is a little unfortunate, but both of these get called from sync contexts throughout mitmproxy (in particular, `trigger`s are often called from `command.command` handlers, which are all sync). To ensure that we don't accidentally invoke an async handler in such cases, this patch adds a check to throw an exception if an async handler gets invoked synchronously.

This is a first stab at async hook support. There are more things that *could* be done here, and I'd love to hear if either of them makes sense for this PR, or left for future work:

- ~Eliminate `concurrent` entirely, and possibly even the entire `reply` mechanism. This would be a significant, breaking API change. However, in almost every case, `async` hooks are better: async hooks are properly sequenced (even with respect to other sync hooks), races are essentially eliminated, and the behaviour is much more predictable. (It might be better to just deprecate `concurrent` now for future removal).~ Basically done: `concurrent` is kept, but made async. `reply` is gone.
- Eliminate the sync versions of `trigger` and `invoke_addon` to allow async hooks in all contexts. This is trickier than it seems because of the number of places in which hooks are triggered in a synchronous context; a lot of code would probably need to be updated to be async. While this probably should happen eventually, I'm hoping this much smaller patch can serve 80% of use cases immediately.

There are also a few easier things that could be done:

- Document which contexts will permit async hooks. Lifecycle hooks like `start`, `done`, etc. don't look like they should be async, for example, since they're often called synchronously.
- Add tests, and/or modify existing tests to use async hooks instead.

#### Checklist


mitmproxy/addonmanager.py
```diff
@@ -273,6 +274,38 @@ def invoke_addon(self, addon, event: hooks.Hook):
...
+
+    def invoke_addon(self, addon, event: hooks.Hook):
+        """
+            Invoke an event on an addon and all its children.
+        """
+        for addon, func in self._iter_hooks(addon, event):
+            res = func(*event.args())
+            if isawaitable(res):
+                raise exceptions.AddonManagerError(
+                    f"Async handler {event.name} ({addon}) called from sync context"
```
mhils, line 295: FWWI I think this is quite a substantial limitation. The long-term plan needs to be that all callers (including commands) should be async-capable. That of course gets unwieldy for a single PR, so let's keep this error here for now. Just FYI.

mitmproxy/addonmanager.py
```diff
@@ -273,6 +274,38 @@ def invoke_addon(self, addon, event: hooks.Hook):
...
+                )
+
+    async def async_trigger(self, event: hooks.Hook):
```
mhils, line 298: In line with my previous comment, we eventually want everyone to call this method here instead of the sync `trigger()`.  This means we should give it a more permanent name that does not include async. How about we call this `trigger_event()`, then we can eventually add deprecation warnings for `trigger/invoke_addon` and tell people to migrate to this new (async) API?

examples/addons/nonblocking.py
```diff
@@ -1,17 +1,13 @@
...
+from mitmproxy import ctx
 
 
-@concurrent  # Remove this and see what happens
-def request(flow):
-    # This is ugly in mitmproxy's UI, but you don't want to use mitmproxy.ctx.log from a different thread.
-    print(f"handle request: {flow.request.host}{flow.request.path}")
-    time.sleep(5)
-    print(f"start  request: {flow.request.host}{flow.request.path}")
+# Change this to a sync hook using time.sleep and see what happens
```
mhils, line 9: Nit: Can we add a commented sync version below and have a `# use this version instead and see what happens` comment? I think "changing to a sync hook" is not trivial to do for newcomers. :)

mhils (commented):
Thanks for the very nice PR! 🍰 

> Eliminate concurrent entirely, and possibly even the entire reply mechanism. This would be a significant, breaking API change.

Yes, eliminating reply is something I definitely want to see happening. `@concurrent` is here to stay though, that's widely used by existing addons. It just needs to be changed to spawn a thread, and then await the completion of an `asyncio.Event` which is set by the thread on completion. This way users can continue to use `requests` or other blocking stuff in their hooks. 

If we could tackle the reply-sectomy in this PR that would be great, if you give it a stab and notice it's too much work I won't insist. :)

> Eliminate the sync versions of trigger and invoke_addon to allow async hooks in all contexts. This is trickier than it seems because of the number of places in which hooks are triggered in a synchronous context; a lot of code would probably need to be updated to be async. While this probably should happen eventually, I'm hoping this much smaller patch can serve 80% of use cases immediately.

Yes, 👍 to this. See my inline comments below.

> Document which contexts will permit async hooks. Lifecycle hooks like start, done, etc. don't look like they should be async, for example, since they're often called synchronously.

My naive approach would be that we could just permit all hooks to be async. If you add `asyncio.sleep(5)` to the start hook that's your own problem really, on the other hand it can be useful to have I-know-what-I'm-doing `await`s in there.

examples/addons/nonblocking.py
mhils: Add: We also want to keep `@concurrent` (see main comment), so it make sense to demonstrate 1) async 2) `@concurrent` and 3) the bad blocking sleep() in here. :)

mitmproxy/addonmanager.py
nneonneo: Yes, it is a big limitation. In particular, I noticed that `run.script` will be unable to call async methods, which sucks for replayability :(.

It might be nice to have async commands for 8.0, although that's trickier because callers would then need to be async-aware themselves.

mitmproxy/addonmanager.py
nneonneo: Do you have a suggested name for `invoke_addon`, then?

mhils: `invoke_addon` is really only used by the script addon and the addon manager, happy to break API compatibility here. 🙊

mhils: If you have nice other idea please feel free to go with it. I don't feel strongly about `invoke_addon` at all. :)

nneonneo:
> If we could tackle the reply-sectomy in this PR that would be great, if you give it a stab and notice it's too much work I won't insist. :)

I'd be happy to, but I did just realize one problem: as it currently stands, if a hook can't be made async (because it gets called synchronously somewhere), it can be made `@concurrent` and called synchronously. If we switch now to using `asyncio` internally, this particular workaround goes away. If this isn't a concern (e.g. if 8.0.0 is planned to ship with async everything), then I'll go ahead and make the change.

mhils:
> If we switch now to using asyncio internally, this particular workaround goes away.

Good catch. We don't use `@concurrent` internally, so this should only affect third-party hooks making use of `@concurrent` in unexpected hooks (or invoking them in non-standard ways), so I think that is tolerable. Let's not impede progress. :)

Prinzhorn:
Thank you very much! I've been playing with it this morning and so far haven't run into any issues. It does consistently increase the throughput of my addon quite a bit (between 10% and 300%).

I've been using mostly https://http2.akamai.com/demo to test.

nneonneo:
@mhils: I appended a patch series to this PR which eliminates the `flow.reply` system. There are a few semantic differences to be aware of, but these are hopefully not significant:

- `@concurrent` produces an async hook, so any `@concurrent` hooks need to run in async context. `@concurrent` hooks now run in series with other hooks of the same type on a given flow (but run concurrently with other hooks on other flows).
- Due to the way `intercept` now works, `Flow.resume` is no longer thread-safe, so it cannot be called from a `@concurrent` function. (If you do, it'll *probably* work but there is a potential race with `wait_for_resume`)
- There is no more option to "take" a reply to signal to future hooks, so a few addons had those checks removed. Other than interception, there were no examples of taking a reply in the repo.

mitmproxy/addons/modifybody.py
```diff
@@ -31,12 +31,12 @@ def configure(self, updated):
                 self.replacements.append(spec)
 
     def request(self, flow):
-        if flow.response or flow.error or flow.reply.state == "taken":
+        if flow.response:
```
mhils, line 34: flow.error removed intentionally?

mitmproxy/addons/onboarding.py
```diff
@@ -6,7 +6,7 @@
 APP_PORT = 80
 
 
-class Onboarding(asgiapp.WSGIApp):
+class Onboarding(asgiapp.ASGIApp):
```
mhils, line 9: Our onboarding app is wsgi, so this breaks http://mitm.it/. Just changing it back to WSGI works for me.

mitmproxy/flow.py
```diff
@@ -226,9 +232,8 @@ def resume(self):
         if not self.intercepted:
             return
         self.intercepted = False
-        # If a flow is intercepted and then duplicated, the duplicated one is not taken.
-        if self.reply.state == "taken":
-            self.reply.commit()
+        if self._resume_event is not None:
+            self._resume_event.set()
```
mhils, line 236: There is a race now where we wait indefinitely if we call `intercept()`, then `resume()`, and only then `wait_for_resume()`, no?

mhils (commented):
This looks exceptional - thank you so much! 🍰🎉 I'm in the somewhat dry final write-up stage of my PhD and this PR just made my day. 😃

A few minor comments below. Once those are addressed I'd be happy to get this into main right away so that it can already get some testing before 8.0. There likely are some unintended side-effects we haven't considered yet, but that's okay. Overall this PR is a massive net benefit.

Ah, also for your entertainment:

https://user-images.githubusercontent.com/1019198/152372170-bc1bbcb8-1564-4f51-a16a-417880df11d8.mp4

A mismatch between the actual changes and the rendered diff tickles my spidey sense, but I'll refrain from falling into this rabbit hole now. 😅

mitmproxy/addons/modifybody.py
nneonneo: Oops, that's a typo. Fixed.

mitmproxy/addons/onboarding.py
nneonneo: OK. I don't know why ASGIApp worked for me, but I'll change it back.

mitmproxy/flow.py
nneonneo: This is specifically tested in `test/mitmproxy/test_http.py::test_wait_for_resume`. `_resume_event` is only set if `wait_for_resume` gets called, but `wait_for_resume` won't create `_resume_event` nor wait if `intercepted` is False. (You can also see where this would break if these get called from different threads...)

mhils: Ah true - `wait_for_resume` exits early if intercept is false. Thanks!

mhils (commented):
Excellent, thanks. I'll re-review the changes tomorrow with a fresh mind so that this can go into main right away. 😃

mhils:
Thanks! I'm currently on a final pass locally and will force-push again once I'm done. :)

mhils:
@nneonneo: Done! I think we are in a good state here. The only major issue I discovered was that `Flow.killable` turned true for flows loaded from disk, which is now mitigated by reintroducing `Flow.live`. I think it's also not a bad idea that flow-modifying addons (map local/map remote/modify headers/ASGI addons) are not applied when flows are read from disk, so I've added that to 372a632161dee642d81542069507826e34466ba1. Let me know if you have any comments/thoughts, otherwise this is good to go from my end! 😃

nneonneo:
LGTM! Let's ship it! :)

PR #5110, opened 2022-02-04
await server_connected hook before doing something with the connection, fixes #5108

### LINKED ISSUE #5108: Wait for connection hooks to finish before calling HTTP hooks
#### Problem Description

This issue is in response to your comment https://github.com/mitmproxy/mitmproxy/issues/4207#issuecomment-907014592

> Ah, you are hitting a special case here: We are not waiting for server_connected to complete before notifying the HTTP layer that the connection has been established (I didn't think this would be useful to anyone and it surely isn't faster). If you have a concrete use case here please open a separate issue, this should be straightforward to adjust. In fact at the moment it's a bit more complicated because we don't block.

To summarize: I'm doing rather heavy work in every hook and communicate with a different service (gRPC calls to Node.js that process the hooks). When using `@concurrent` (and probably same with async hooks) hooks such as `server_connected` and `request` are racing each other (I think that's actually the only case). With synchronous hooks this cannot happen because whatever I do in `server_connected` blocks the entire proxy.

```py
import time


@concurrent
def client_connected(data):
    print('>>> client_connected')
    time.sleep(2)
    print('<<< client_connected')


@concurrent
def server_connect(data):
    print('>>> server_connect')
    time.sleep(2)
    print('<<< server_connect')


@concurrent
def server_connected(data):
    print('>>> server_connected')
    time.sleep(2)
    print('<<< server_connected')


@concurrent
def request(flow):
    print('>>> request')
    time.sleep(2)
    print('<<< request')


@concurrent
def response(flow):
    print('>>> response')
    time.sleep(2)
    print('<<< response')


@concurrent
def client_disconnected(data):
    print('>>> client_disconnected')
    time.sleep(2)
    print('<<< client_disconnected')


@concurrent
def server_disconnected(data):
    print('>>> server_disconnected')
    time.sleep(2)
    print('<<< server_disconnected')
```

As you can see `server_connected` and `request` are interleaved.

```
>>> client_connected
<<< client_connected
>>> server_connect
<<< server_connect
>>> server_connected
>>> request
<<< server_connected
<<< request
>>> response
<<< response
>>> client_disconnected
<<< client_disconnected
>>> server_disconnected
<<< server_disconnected
```

#### Proposal

Wait for the connection hooks to finish before continuing with the HTTP layer.

#### Alternatives

I could leave the connection event synchronous so that the always block and finish. But depending on the number of clients and servers this would mostly diminish the effect of making all other hooks async or concurrent.

Prinzhorn:
#### Description

I tried with some live traffic and it behaved as expected. Also the demo add-on no longer logs server_connected and request interleaved.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 * Fix random connection stalls (#5040, @EndUser509)
 * Add `n` new flow keybind to mitmweb (#5061, @ianklatzco)
 * Fix compatibility with BoringSSL (@pmoulton)
+* Await the `server_connected` hook before doing something with the connection, e.g. before HTTP `request` hook (@Prinzhorn)
 
 ## 28 September 2021: mitmproxy 7.0.4
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,8 @@
 * Fix random connection stalls (#5040, @EndUser509)
 * Add `n` new flow keybind to mitmweb (#5061, @ianklatzco)
 * Fix compatibility with BoringSSL (@pmoulton)
-* Await the `server_connected` hook before doing something with the connection, e.g. before HTTP `request` hook (@Prinzhorn)
+* Change connection event hooks to be blocking.
+  Processing will only resume once the event hook has finished. (@Prinzhorn)
 
 ## 28 September 2021: mitmproxy 7.0.4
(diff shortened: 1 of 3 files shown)


mhils (approved):
Excellent, thanks! 🍰

PR #5134, opened 2022-02-15
Add support for XDG standard

jrabinow:
#### Description

Add support for [XDG standard](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)

https://github.com/mitmproxy/mitmproxy/pull/3950 seems to be in limbo, so I have implemented my own version.

I'd like to draw attention to the following points:
- `config.{yaml, yml}` and `keys.yaml` go under `${XDG_CONFIG_HOME:-~/.config}/mitmproxy`. However command history and anything cert-related probably count as data, not as configs, so I made them go under `${XDG_DATA_HOME:-~/.local/share}/mitmproxy`. I have also added a new option `datadir`.
 - xdg directories are purely opt-in: this PR does not move or remove support for `~/.mitmproxy` dir. Users can choose to move their files to `$XDG_CONFIG_HOME/mitmproxy` and `$XDG_DATA_HOME/mitmproxy` if they like. If the user doesn't manually perform the migration, they will see no change.
 - if both `~/.mitmproxy` and either `$XDG_CONFIG_HOME/mitmproxy` or `$XDG_DATA_HOME/mitmproxy` exist, the app prints a warning and uses the XDG directory. This behavior can be changed however you'd like.

Differences compared to the discussion on https://github.com/mitmproxy/mitmproxy/pull/3950:
- I have not exposed a `path.Pathlib` object. It's a bit more work but I don't have any problem with doing so if it's still desirable.
- I have not "killed `~/.mitmproxy` with fire" despite it being suggested in the original discussion, as that would imply figuring out a migration path. I would recommend against breaking users setup on upgrade without warning beforehand, so this migration path would mean at least 2 releases, one with a warning that urges users to perform the migration, and the second which drops support for `~/.mitmproxy` entirely. Then again, if the maintainers prefer migrating all users to xdg directories and would like to just migrate everyone in one go, I can change the code to do that.
- >How about we add a section at the beginning of https://docs.mitmproxy.org/stable/concepts-options/ that looks as follows?

  I have not taken this approach as the code is written in a way to make using xdg directories purely opt-in. As such, I'd say it's reasonable to assume that people who take the manual steps to opt-in will be aware of their config situation. Please let me know if you'd prefer to add a new section at the beginning of `concepts-options.md` despite this.

Questions:
 - As of this moment, the value for `confdir` is hardcoded in `app.js` and I'm wary of writing code in `web/src/js/ducks/_options_gen.ts` as it seems to be autogenerated. What would be the best course of action here if any?

#### Checklist


jrabinow:
I'm not sure why tests are failing on `windows-latest, 3.9`, maybe my patch isn't working on that platform? I would special-case for windows but `windows-2019, windows` tests are running so I'm not sure if that's an option. If anyone had pointers, they would be appreciated

jrabinow:
Hi @mhils , I understand you don't have much time to spare on thorough reviews these days. I'm wondering, do you have any overall  impressions and/or comments that I could address while I'm waiting?

setup.py
```diff
@@ -92,6 +92,7 @@
         "urwid>=2.1.1,<2.2",
         "wsproto>=1.0,<1.2",
         "publicsuffix2>=2.20190812,<3",
+        "xdg>=5.1.1",
```
sujaldev, line 95: Why can't we just hard-code `~/.config/mitmproxy`?

jrabinow:
@mhils to answer your review here:

the reason for 3/ is actually because of 4/ I believe the best migration plan involves giving users time to adapt and move their files on their own. We could remove support for `~/.mitmproxy` at any time, but doing a direct cutover would break backwards-compat exactly as you mention.
This PR keeps `~/.mitmproxy` the default but the default could easily be changed to `$XDG_CONFIG_DIR/mitmproxy` before merging if you'd prefer.
Or I could even do like I suggested initially and entirely drop support for `~/.mitmproxy` in this PR, though I have doubts about this being a good idea.
Happy to go with whatever you prefer here.

As for the new data directory, this is in accordance with the spec. My observation was things were being stored under `~/.config/mitmproxy` which I really wouldn't want committed to a git repo. The data directory is there to help avoid that.

Any pointers re: windows CI failures?

sujaldev:
> My observation was things were being stored under `~/.config/mitmproxy` which I really wouldn't want committed to a git repo.

that could be easily fixed by adding this to the gitignore file:
```gitignore
.config/mitmproxy/*
!.config/mitmproxy/*.yml # or whatever files that would go in the data dir
```

jrabinow:
@sujaldev yes, gitignore is another option but if I forget to add my files in there, now I just published online a way for anyone to MITM all my devices where I installed the private cert. IMO best to completely remove the possibility of those kinds of accidents.

jrabinow:
@sujaldev yes, gitignore is another option but if I forget to add my files in there, now I just published online a way for anyone to MITM all my devices where I installed the private cert. Let's not tempt fate, IMO best to completely remove the possibility of those kinds of accidents from happening in the first place.

mhils:
@jrabinow: Sorry for not getting to your PR earlier, this somehow feel through the cracks!

> I left `~/.mitmproxy` as the default [...] giving users time to adapt and move their files on their own.

Practically speaking I don't expect a significant percentage of users would move stuff on their own. Frankly my hunch is that a lot of less experienced users are unware that `~/.mitmproxy` even exists. So all this does it add an esoteric option that is used by a very small percentage of Linux power users. Not a fan of that at all, sorry.

> but doing a direct cutover would break backwards-compat exactly as you mention.

Yes and no - two ideas how we could not break backward compat:

 1. Move `~/.mitmproxy` and add a symlink instead.
 2. Move `~/.mitmproxy` and add a `~/.mitmproxy/this-directory-has-moved.md` file with a short explanation.

Admittedly I haven't given either of those a lot of thought yet.

> which I really wouldn't want committed to a git repo

`command_history` could maybe default to `$XDG_DATA_DIR` without being configurable at all. Which only leaves certs.

Moving `~/.mitmproxy` to `$XDG_CONF_DIR` but then placing certs in yet another directory (`$XDG_DATA_DIR`) sounds like a massive support headache. There are tons of tutorials out there referencing `~/.mitmproxy`. I do see your security argument, but it's pain. Long story short, the more I think about this the more I'm convinced that we should keep things as-is. I don't think it's enough of an improvement to justify the breakage and dev costs. Again, I'm mostly working on intuitions here and might be wrong, but I don't really see the appeal. This may make sense to incorporate later down the road when we a breaking compatibility somehow anyway, but maybe not now.

jrabinow:
@mhils no worries for the delay, thanks for getting back to me!

> So all this does it add an esoteric option that is used by a very small percentage of Linux power users. Not a fan of that at all, sorry.

I'm sorry to hear that. If it can help you reconsider, it was the solution that [taskwarrior went with](https://github.com/GothenburgBitFactory/taskwarrior/pull/2316#issuecomment-732668710), precisely because it would be a feature which maintainers wouldn't have to worry about unless users had deliberately opted in. At which point they'd be able to handle any breakage on their own, assuming the feature was implemented properly of course.
That being said, if you still feel like the pros are outweighed by the cons here, I get it, I'm sure there's a maintenance burden involved that's invisible to me.

Re backwards-compat, I considered symlinks as well. EOD we're still moving the user's files around without their permission. I'll admit I'm quite conservative when it comes to these things, it's not the first time this approach would have been taken.

Thanks for providing this great software free of charge, it's quite appreciated with our without xdg 🙂

Issue #5046, opened 2022-01-11
Send HTTP/2 PING frames to keep connections alive.


mhils:
### Discussed in https://github.com/mitmproxy/mitmproxy/discussions/5044

<div type='discussions-op-text'>

<sup>Originally posted by **EndUser509** January 10, 2022</sup>
I have noticed that for instance https://heise.de, when proxied, returns a '502' HTTP error to the client when I click on a link after not doing anything (=reading the page) for over 60 seconds. What I see on the outgoing interface is that the remote server resets the connection the moment the connection is used again (TCP RST). 
When not proxied I observe that the client sends a HTTP/2 PING after 60 seconds which seems to keep the connection alive. Varying the idle time from 60 to 120 seconds, I never manage to reset the connection and a link always loads. 
I wonder if there is a way to remedy the proxied behavior. I am not familiar enough with HTTP/2 to see if this PING interval is something that client and server agree to. Or if this is 'best practice' for browsers. Any ideas where to start? Would it make sense that the proxy sends regular PINGs for idle HTTP/2 connections? </div>

EndUser509:
This ca be closed now, thanks for the final changes to the PR and merging it! :100:

PR #5144, opened 2022-02-18
Http2 keepalive

EndUser509:
#### Description

This is a first PR for the implementation of the HTTP/2 keep-alive via ping as discussed in #5044 and the subsequent open issue #5046. 

I am open about naming of options, classes and functions, just tell me what you would like. The RequestKeepAlive command can be extended with a version int to deal with future HTTP versions that require a similar strategy but other intervals and messages.

Questions that I am not sure:
 - Keeping connections alive leads to more concurrently open connections. On my iPad I managed to get beyond 1200 open file handles on my Linux mitmproxy. Is that an issue?
 - Is there any additional safety needed in case somebody toys with layers, like accepting but ignoring events or commands in other layers?
 
#### Checklist


mhils (commented):
This looks fantastic, thanks! 🍰

I don't have time for a proper review in the next few days, but a few high-level points:
 - The `RequestKeepAlive`/`KeepAlive` command/event pair is too specific. Let's aim for a generic timer functionality that can be reused in other protocols. For example, name it `ScheduleWakeup`/`Wakeup` or `SetTimer`/`TimerExpired`. The wakup event would then always be executed after the specified amount of seconds. In the HTTP/2 layer, we keep a timestamp of the last activity. If we are woken up while `last_activity + http2_ping_threshold > time.time()` (because something happened in the meantime), we just reschedule a new timer with the remaining time.
 - I think we should enable this by default with the same 58s setting as Firefox. There's no harm, is there?

mhils:
> Any ideas how get the additional test coverage. Just adding #pragma: no cover for all of them would be lame.

The idle watchdog would go away with the changes proposed in my previous comment. Let's focus on the test coverage of the HTTP/2 bits. :)

mhils:
> I looked for `time` in the code but I could not find this timestamp. Can you point me out which variable it is?

My proposal was to do this, we aren't doing it yet. Sorry, that was a bit unclear. 😅

EndUser509:
OK, next try. I have restructured this as discussed, `RequestWakeup` now starts a one-shot timer that sends `Wakeup`. The timer restart logic is now all in `_http2.py`. I am nut sure if the place where I update `last_activity` is the best choice but it seemed natural.
Making 58s idle PINGs the default breaks a couple of tests. I have not touched these yet.

mitmproxy/proxy/layers/http/__init__.py
```diff
@@ -708,6 +708,8 @@ def _handle_event(self, event: events.Event):
             yield from self.event_to_child(self.connections[self.context.client], event)
             if self.mode is HTTPMode.upstream:
                 self.context.server.via = server_spec.parse_with_mode(self.context.options.mode)[1]
+        elif isinstance(event, events.Wakeup):
+            yield from self.event_to_child(self.connections[self.context.server], event)
```
mhils, line 712: This always forwards the event to the first server, which may not be the one which requested a wakeup. We need to keep track which stream is the correct one here. Fortunately the same machinery is already in place for `CommandCompleted` events.
I think we can just extend the logic at https://github.com/mitmproxy/mitmproxy/blob/01d67951eed021f4de8dc53f29ae3132a121a348/mitmproxy/proxy/layers/http/__init__.py#L765-L769 to not only register blocking commands but also WakeupRequests, and then forward to the correct layer here similar to what we do for `CommandCompleted` right below.
Let me know if that does not make sense, it's a bit tricky. Happy to expand.

mitmproxy/proxy/layers/http/_http2.py
```diff
@@ -407,7 +408,21 @@ def _handle_event(self, event: Event) -> CommandGenerator[None]:
                 yield from self._handle_event(event)
 
     def _handle_event2(self, event: Event) -> CommandGenerator[None]:
-        if isinstance(event, RequestHeaders):
+        if isinstance(event, Wakeup):
+            remaining = self.context.options.http2_ping_threshold - time.time() + self.last_activity
+            if remaining <= 0.5:
+                remaining = self.context.options.http2_ping_threshold
+                self.h2_conn.ping(b"0" * 8)
```
mhils, line 415: I was wondering why you'd add 8 seemingly random bytes here. The RFC had the answer. :)
```suggestion
                # PING frames MUST contain 8 octets of opaque data in the payload. 
                # A sender can include any value it chooses and use those octets in any fashion.
                self.h2_conn.ping(b"0" * 8)
```

mitmproxy/proxy/server.py
```diff
@@ -76,6 +88,8 @@ class ConnectionHandler(metaclass=abc.ABCMeta):
     client: Client
     max_conns: typing.DefaultDict[Address, asyncio.Semaphore]
     layer: layer.Layer
+    wakeuptimer: typing.Optional[WakeupTimer] = None
+    wakeuptimer_task: typing.Optional[asyncio.Task] = None
```
mhils, line 92: We may have multiple timers here, which would currently overwrite each other. Let's make this a simple `wakeup_timer: set[asyncio.Task]`, I don't think we need to store the instances at all. :)

test/mitmproxy/proxy/layers/http/test_http2.py
```diff
@@ -780,3 +780,24 @@ def test_request_smuggling_te(tctx):
...
+            >> DataReceived(tctx.client,
+                            cff.build_headers_frame(example_request_headers, flags=["END_STREAM"]).serialize())
+            << http.HttpRequestHeadersHook(flow)
+            >> reply()
+            << http.HttpRequestHook(flow)
+            >> reply()
+            << OpenConnection(server)
+            >> reply(None, side_effect=make_h2)
+            << RequestWakeup(1)
+            << SendData(server, initial)
```
mhils, line 802: We should emulate a `Wakeup` event here and make sure that it triggers a PING. :-)

```python
def test_request_keepalive(monkeypatch, tctx):
    ...
    assert (playbook ...)
    t = time.time()
    monkeypatch.setattr(time, "time", lambda: t + 60)
    assert (playbook >> Wakeup() ...)
```

mhils (commented):
This looks awesome, thanks! 😃🍰 
I've added a few comments below, let me know if anything is unclear. :)

mitmproxy/proxy/server.py
EndUser509: OK, I will just cancel all elements in `wakeup_timer` on shutdown. The remembered `RequestWakeup` instances in the `command_sources[command]` in `http/__init__.py` do not require any cleanup, right?

test/mitmproxy/proxy/layers/http/test_http2.py
EndUser509: I'll do this at the very end, when the rest is working. :)

mitmproxy/proxy/server.py
mhils: No, all of that is automatically garbage-collected. :)

EndUser509:
The update should implement the proper routing of the `Wakeup` event to the right client. I have used this in production for a week and observed no crashes, errors or other problems.
Quite some tests for HTTP/2 are broken due to the additional `WakeupRequest` command and I am not sure whether I should just disable PING for all test cases except for the ones explicitly testing the functionality. Or fix all broken tests.

mitmproxy/proxy/layers/http/_http2.py
```diff
@@ -407,7 +408,23 @@ def _handle_event(self, event: Event) -> CommandGenerator[None]:
...
+        if isinstance(event, Wakeup):
+            remaining = self.context.options.http2_ping_keepalive - time.time() + self.last_activity
+            if remaining <= 0.5:
+                remaining = self.context.options.http2_ping_keepalive
+                # PING frames MUST contain 8 octets of opaque data in the payload. 
+                # A sender can include any value it chooses and use those octets in any fashion.
+                self.h2_conn.ping(b"0" * 8)
+                yield Log(f"Send HTTP/2 keep-alive PING to {human.format_address(self.conn.peername)}")
+                yield SendData(self.conn, self.h2_conn.data_to_send())
+            yield RequestWakeup(remaining)
```
mhils, line 420: Stupid question: Shouldn't we update `last_activity` after sending a PING?

mitmproxy/proxy/commands.py
```diff
@@ -39,6 +39,16 @@ def __repr__(self):
         return f"{type(self).__name__}({repr(x)})"
 
 
+class RequestWakeup(Command):
+    """
+    Request a `Wakeup` event after the specified amount of seconds.
+    """
+    threshold: float
```
mhils, line 46: Nit: `threshold` is a bit nondescriptive. How about `delay` (matching https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.call_later)? Or `wakeup_after`?

mitmproxy/proxy/server.py
```diff
@@ -63,6 +63,18 @@ def disarm(self):
...
+
+    def __init__(self, callback: typing.Callable[[], typing.Any], request: commands.RequestWakeup):
+        self.callback = callback
+        self.request = request
+
+    async def sleep(self):
+        await asyncio.sleep(self.request.threshold)
+        await self.callback(self)
+
+
```
mhils, line 77:
```suggestion
```

I think we can simplify this a bit (see suggestions below)

mitmproxy/proxy/server.py
```diff
@@ -199,6 +216,11 @@ async def open_connection(self, command: commands.OpenConnection) -> None:
                 command.connection.timestamp_end = time.time()
                 await self.handle_hook(server_hooks.ServerDisconnectedHook(hook_data))
 
+    async def wakeup(self, timer: WakeupTimer) -> None:
+        self.wakeup_timer.discard(timer.handler)
+        timer.handler = None
```
mhils, line 221:
```suggestion
    async def wakeup(self, request: commands.RequestWakeup) -> None:
        await asyncio.sleep(request.threshold)
        self.wakeup_timer.discard(asyncio.current_task())
```

mitmproxy/proxy/server.py
```diff
@@ -297,6 +319,15 @@ def server_event(self, event: events.Event) -> None:
...
                     self.transports[command.connection] = ConnectionIO(handler=handler)
+                elif isinstance(command, commands.RequestWakeup):
+                    timer = WakeupTimer(self.wakeup, command)
+                    handler = asyncio_utils.create_task(
+                        timer.sleep(),
+                        name="wakeup_timer",
+                        client=self.client.peername
+                    )
+                    timer.handler = handler
+                    self.wakeup_timer.add(handler)
```
mhils, line 330:
```suggestion
                    task = asyncio_utils.create_task(
                        self.wakeup(command),
                        name="wakeup_timer",
                        client=self.client.peername
                    )
                    self.wakeup_timer.add(task)
```

mhils (commented):
Looks awesome now! 💯 
I only have some comments that should be very easy to implement.

> Quite some tests for HTTP/2 are broken due to the additional WakeupRequest command and I am not sure whether I should just disable PING for all test cases except for the ones explicitly testing the functionality. Or fix all broken tests.

I think just disabling it for those tests is totally fine. 👍 I'm happy with either approach though, whatever you find is easier. 😃

mitmproxy/proxy/layers/http/_http2.py
EndUser509: Right, that would be more clean.

mitmproxy/proxy/commands.py
EndUser509: Makes sense. I'll rename it to `delay`.

EndUser509:
OK, next update. I had to add `assert` to `wakeup` in `server.py` to get rid of type missmatches involving `Optional`. Maybe you have a more elegant solution again.
Can you please look at the remaining errors?

- The error about `set[asyncio.Task]` only affects 3.8 and the code is valid in 3.9 and newer (PEP 585)
- The coverage errors are for the initializers of `Wakeup` and `RequestWakeup`, both are instantiated during the tests so I do not see what I can do here.
- I am struggling to get the test case to run with the monkeypatched time. Any ideas?

mhils:
> OK, next update. I had to add `assert` to `wakeup` in `server.py` to get rid of type missmatches involving `Optional`. Maybe you have a more elegant solution again. 

No, that's fine. :)

> * I am struggling to get the test case to run with the monkeypatched time. Any ideas?

Do you want to push what you have/what is not working so far? I'm busy with family stuff over easter, but happy to look at this (and your other questions) next week. :)

EndUser509:
No need to hurry. :) My sample  test is already in the commit, the last one in [test_http2.py](https://github.com/mitmproxy/mitmproxy/pull/5144/files#diff-71acfa4101f4c1c88e07eb1f618e0da37361e0f557cb325956ebf9d363212e68). All other tests are fixed. 
I don't understand the test harness too well, but I assume that the `RequestWakeup` instance that is expected first needs to be the same that is again referenced in the `Wakeup` coming out. But that does not work.

mhils (approved):
This looks good to merge now, thank you very much again! 🍰 🚀 😃

PR #5191, opened 2022-03-17
Add devcontainer

o4x:
#### Description

I have added some devcontainer configurations ( Based on #5150 ) for the test of this feature.

#### Checklist


mhils:
Thanks for the PR! I appreciate you building this, but at the same time this will incur lots of maintenance effort down the road. I don't feel comfortable taking that over (in particular as I don't use vscode), so this unfortunately won't make it in. It's way too much overhead for being specific to one editor. Sorry!

PR #5217, opened 2022-03-27
Add domain fronting example

randomstuff:
#### Description

Adds an addon which implements domain fronting. 

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Add flatpak support to the browser addon (#5200, @pauloromeira)
 * Add example addon to dump contents to files based on a filter expression (#5190, @redraw)
+* Add example addon for domain fronting.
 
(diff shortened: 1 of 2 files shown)


mhils:
Thanks for the PR! 🍰 
Is there a reason why this functionality cannot be accomplished using the *Map Remote* option? :) https://docs.mitmproxy.org/dev/overview-features/#map-remote

randomstuff:
So I tried and it is possible indeed to achieve domain fronting using something like this:

~~~sh
--map-remote '|^https://private.example/|https://public.example/' \
--modify-headers '|^https://public.example/|Host|private.example'
~~~

Using this: `https://private.example/` results in:

* TLS SNI = `public.example`
* HTTP Host = `private.example`

A downside of this approach, is that it is necessary to use one public host name per private host name we want to access i.e. it is not possible to access several different host names using a single host name at the TLS (SNI) layer.

PR #5227, opened 2022-03-29
Add `Flow.timestamp_created` to retain list order on replay

mhils:
This PR adds `Flow.timestamp_created`, which is used to make sure that replayed flows retain their position in the flow list. They don't jump to the bottom of the list anymore.

@Prinzhorn / @Kriechi, given that this is a public API change, any comments?


bratao:
@mhils congratulations on the phd thesis submission!

PR #5252, opened 2022-04-07
Add console-based hex editors for binary formats

mohamedmostafadawood:


mitmproxy/tools/console/master.py
```diff
@@ -123,12 +123,30 @@ def get_editor(self) -> str:
         else:
             return "vi"
 
+    def get_hex_editor(self) -> str:
+        if m := os.environ.get("MITMPROXY_EDITOR"):
+            return m
+        if m := os.environ.get("EDITOR"):
+            return m
```
mhils, line 130: Do you think it makes sense to use `$EDITOR` here?

mitmproxy/tools/console/master.py
```diff
@@ -123,12 +123,30 @@ def get_editor(self) -> str:
...
+        for editor in "hexedit", "hexyl":
+            if shutil.which(editor):
+                return editor
+        if os.name == "xxd":
```
mhils, line 134: I don't understand this comparison. Can you explain?

mitmproxy/tools/console/master.py
```diff
@@ -123,12 +123,30 @@ def get_editor(self) -> str:
...
+            return "xxd"
+        else:
+            return "ghex"
```
mhils, line 137: What happens if ghex is not installed?

mitmproxy/tools/console/master.py
```diff
@@ -123,12 +123,30 @@ def get_editor(self) -> str:
...
+
     def spawn_editor(self, data):
+        if strutils.is_mostly_bin(data):
+            isBinary = True
```
mhils, line 141: - Please use `snake_case` for Python variables.
- Is there a reason why you create a variable here instead of directly using `strutils.is_mostly_bin` later on?

mhils (changes requested):
Thanks for the PR!

mitmproxy/tools/console/master.py
mohamedmostafadawood: I assumed that the user may have a preferred utility to edit binary formats.

mitmproxy/tools/console/master.py
mohamedmostafadawood: I got it! Thanks.

mitmproxy/tools/console/master.py
mohamedmostafadawood: I figured out the error :).
I will push the new code.

mitmproxy/tools/console/master.py
mhils: `$EDITOR` is set on most Linux systems, if we always take that as the first choice we would essentially never open a real hex editor. I don't think we should take it into account. We could check for a `$MITMPROXY_HEXEDITOR` variable though.

mitmproxy/tools/console/master.py
```diff
@@ -123,12 +123,32 @@ def get_editor(self) -> str:
...
+            return m
+        for editor in "xxd", "ghex", "hexedit", "hexyl":
+            if shutil.which(editor):
+                return editor
+        if os.name == "nt":
+            for editor in "hexalter", "frhed", "HxD":
+                if shutil.which(editor):
+                    return editor
+                else:
+                    return "notepad"
```
mhils, line 139: This code is clearly untested as it contains a major logic bug. Please don't rely on me catching your errors but test your code yourself. Ideally by writing tests. :)

mhils (changes requested):

mitmproxy/tools/console/master.py
mohamedmostafadawood: Okay, sir. I will add the required test and sorry for the inconvenience.

mohamedmostafadawood:
@mhils  I have tested it and it is now working.
But, on Windows, I faced a problem related to that shutil.which() sometimes cannot locate a hex-editor.
I have made some research related to this problem and found this workaround. But, I think it is dependent on the user machine using the tool and if he has changed the PATHEXT as this workaround suggests or not.
So, what do you think will be the best solution for this to continue my work :) ?
Screenshot from 2022-04-14 21-32-54

mohamedmostafadawood:
> @mhils I have tested it and it is now working. But, on Windows, I faced a problem related to that shutil.which() sometimes cannot locate a hex-editor. I have made some research related to this problem and found this workaround. But, I think it is dependent on the user machine using the tool and if he has changed the PATHEXT as this workaround suggests or not. So, what do you think will be the best solution for this to continue my work :) ? Screenshot from 2022-04-14 21-32-54

references:
https://stackoverflow.com/questions/56399696/shutil-which-not-finding-programs-without-appending-file-extension

https://stackoverflow.com/questions/61097406/jupyter-kernelspec-not-found-while-installing-iqsharp-however-it-exists-on-pat/

mitmproxy/tools/console/master.py
```diff
@@ -13,7 +13,7 @@
 import threading
```
rinsuki, line 14: It is need to use `platform.system()`

```suggestion
import platform
```

mitmproxy/tools/console/master.py
```diff
@@ -123,12 +123,28 @@ def get_editor(self) -> str:
...
+            return m
+        for editor in "ghex", "hexedit", "bless", "wxhexeditor", "nano":
+            if shutil.which(editor):
+                return editor
+        if os.name == "nt":
+            for editor in "HxD", "WinVi", "Cygnus", "Frhed", "notepad":
+                if shutil.which(editor):
+                    return editor
+        else:
+            return "vi"
```
rinsuki, line 137: `bless` command is completely different tool on macOS (in macOS, it is a tool to modify startup disk configuration / set volumes bootability), so it might be good to check Darwin or not.

also I think nano is not suitable for binary editor.

```suggestion
        editors = ["ghex", "hexedit", "wxhexeditor"]
        if platform.system() == "Darwin":
            editors = ["hexf"] + editors
        else:
            editors += ["bless"]
        if os.name == "nt":
            editors += ["HxD", "WinVi", "Cygnus", "Frhed", "notepad"]
        for editor in editors:
            if shutil.which(editor):
                return editor
        return "vi"
```

(note: `hexf` command is shortcut to Hex Fiend.app https://github.com/HexFiend/HexFiend )

Issue #5264, opened 2022-04-16
Mitmweb export copy failed in non-secure domain.


Pactortester:
#### Problem Description

mitmweb export copy failed !

`Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'writeText')`
#### Steps to reproduce the behavior:
1. startup parameters: `mitmweb -p 8888 --web-host 192.168.0.124`
2. mitmweb-->Flow-->Export-->Copy as cURL
<img width="1211" alt="image" src="https://user-images.githubusercontent.com/29191106/163660097-0fbadd9c-73bc-4b8a-a16f-28bc22d6ee67.png">

#### System Information
Mitmproxy: 8.0.0
Python:    3.8.9
OpenSSL:   OpenSSL 1.1.1n  15 Mar 2022
Platform:  macOS-10.16-x86_64-i386-64bit

PR #5265, opened 2022-04-16
Fix mitmweb export copy failed in non-secure domain

Pactortester:
https://github.com/mitmproxy/mitmproxy/issues/5264

#### Description

The browser disables the navigator.clipboard object for non-secure domains.

- startup parameters: `mitmweb -p 8888 --web-host 192.168.0.124`
- browser console error : `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'writeText')`

The security domain includes addresses for local access and TLS security authentication, such as `https` protocol addresses, `127.0.0.1` or `localhost`.

So I submit a PR:
- Use `navigator.clipboard` in the secure domain to improve efficiency, and fall back to `document.execCommand('copy');` in the non-secure domain to ensure that the function is available.

#### Checklist


mhils:
Thank you very much for the PR, much appreciated! 🍰

PR #5318, opened 2022-05-04
Fix terminal breakage if an error occurs at startup

nneonneo:
sys.exit() causes a CancelledError to be thrown. ErrorCheck._shutdown_if_errored
fires sys.exit after self.running() returns, causing the CancelledError to
appear in Event.wait() and preventing Master.done() from being called. This can
result in terminal breakage when using the CLI, and is especially apparent if
any errors occur during startup (e.g. port is not available, script has a syntax
error, etc.).

#### Description

The fix is simply to ensure that Master.done always gets called.

Note that this doesn't handle the case of people calling `sys.exit` or throwing some other `BaseException` from a `running` hook; this ideally should be a very rare occurrence, so it's probably not worth specially handling.

#### Checklist

### DIFF
--- a/mitmproxy/addons/errorcheck.py
+++ b/mitmproxy/addons/errorcheck.py
@@ -1,6 +1,5 @@
 import asyncio
 import sys
-from typing import Optional
 
 from mitmproxy import log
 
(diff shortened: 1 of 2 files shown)


nneonneo:
Not really sure how to test this, given that it's mostly a CLI bug. I guess I could make some kind of test that throws a delayed `sys.exit` in `running` and then test that `done` gets called, but it seems contrived and not really related to the actual bug (terminal state is not restored on exit).

mhils:
> Not really sure how to test this, given that it's mostly a CLI bug. I guess I could make some kind of test that throws a delayed `sys.exit` in `running` and then test that `done` gets called, but it seems contrived and not really related to the actual bug (terminal state is not restored on exit).

Yes, I'm with you here. If this ever comes up again we'll make something up. Thank you again for the excellent PR! 🍰 ❤️

mhils (approved):

PR #5325, opened 2022-05-06
Properly encode hostname for upstream requests.

### LINKED ISSUE #5316: Improper handling of Punycode
#### Problem Description
Can't open an address like `https://стопкоронавирус.рф/` through `mitmproxy` or other applications. My upstream proxy receives a CONNECT request to https://СЃС‚РѕРїРєРѕСЂРѕРЅР°РІРёСЂСѓСЃ.СЂС„:443 instead. As the current run of `mitmproxy` was supposed to be just a test, it was configured only to forward all requests as-is to the upstream proxy, so this rules out any and all issues that could arise from my tinkering. Note: the actual URL that the browser opens is `https://xn--80aesfpebagmfblc0a.xn--p1ai` in this case.

Did it fail to properly encode the resulting authority? My upstream proxy normally has no issues with opening Puny-encoded URLs. I can verify that by opening that URL bypassing `mitmproxy`. It looks like it uses the wrong encoding, as it reminds me of the time when Unicode was not widespread and so this is how text in Russian would display when the text encoding wasn't set correctly.

1. Configure `mitmproxy` to forward all requests as-is to the upstream proxy that optionally can report what requests it receives. This includes no HTTPS decryption.
2. Navigate your browser to `https://стопкоронавирус.рф/`.
3. Check what the authority part of the URL the upstream proxy gets, it should be mangled.

#### System Information

```Mitmproxy: 8.0.0 binary
Python:    3.10.2
OpenSSL:   OpenSSL 1.1.1n  15 Mar 2022
Platform:  Windows-10-10.0.19043-SP0
```

nneonneo:
#### Description

Encode host part of authority using idna in upstream proxy, rather than the system encoding.

Fixes #5316.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@
 * When opening an external viewer for message contents, mailcap files are not considered anymore.  
   This preempts the upcoming deprecation of Python's `mailcap` module. 
   ([#5297](https://github.com/mitmproxy/mitmproxy/issues/5297), @KORraNpl)
+* Fix hostname encoding for IDNA domains in upstream mode.
+  ([#5316](https://github.com/mitmproxy/mitmproxy/issues/5316), @nneonneo)
   
 ## 19 March 2022: mitmproxy 8.0.0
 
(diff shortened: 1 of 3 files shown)


mhils (approved):
Awesome, thanks for fixing this! 😃

PR #5378, opened 2022-05-26
Fix 5363

EndUser509:
#### Description

Proposed fix for #5363. 

Not sure if the action taken when the transport is closing are correct but they looked logical to me.

#### Checklist


EndUser509:
Hmm, seems that the new `DatagramWriter` does not have a `transport` member. What is better, add a fake transport with the needed member function to that class or make a more complex `if` statement?

EndUser509:
I'm a little bit lost how to add the missing tests to get the last failed commit check to pass...  :/

mhils:
Fixed in db9b3c21be9fc3b6fe56b4eed20b5771f72d864b, thanks again! 😃

PR #5393, opened 2022-06-03
Multi-Server Mode

mhils:
This is a very-much-work-in-progress implementation of the multi proxy mode discussed in https://github.com/mitmproxy/mitmproxy/discussions/5288:

```shell
$ mitmdump --mode regular --mode reverse:example.com@8443
HTTP(S) proxy listening at *:8080.
Reverse proxy to example.com listening at *:8443.
```

Tests are still failing, but the invocation above works. There are a few minor changes on the latest dicussions:

 - @Prinzhorn's question, *"what new properties does Flow need so we know where it came from?"*, to which I foolishly replied *"None. If you for some weird reason want different logic here, you can look at `flow.client_conn.sockname`"* was of course spot-on. There are some legitimate cases where we need this (e.g. proxyauth) and `Client.proxy_mode` is a thing now. 😄 
 - After pondering on this a bit, I think it's most sensible if we continue to default to a single HTTP(S) proxy on port 8080. However, we shall have super easy buttons in mitmproxy/mitmweb to quickly start up more services.
 - `dns`/`regular-dns`/... wording isn't fully fleshed out yet, we can make final decisions here once we have a raw UDP layer and/or QUIC.
 - @Meitinger / @decathorpe: Do `mitmproxy/proxy/mode_specs.py` and `mitmproxy/proxy/mode_servers.py` make sense to you? Does that work for WireGuard/QUIC?

TODO:

 - [ ] Fix tests.
 - [ ] Fail on duplicate ports.
 - [ ] Address remaining FIXMEs.
 - [ ] Adjust mitmweb.
 - [ ] Add migration code for `listen_host` and `listen_port`.
 - [ ] Docs.

### DIFF
--- a/mitmproxy/addons/proxyserver.py
+++ b/mitmproxy/addons/proxyserver.py
@@ -1,112 +1,60 @@
+"""
+This addon is responsible for starting/stopping the proxy server sockets/instances specified by the mode option.
+"""
+from __future__ import annotations
+
 import asyncio
-from asyncio import base_events
+import collections
 import ipaddress
-import re
-import struct
+from contextlib import contextmanager
 from typing import Optional
+
+from wsproto.frame_protocol import Opcode
 
 from mitmproxy import (
     command,
     ctx,
     exceptions,
-    flow,
     http,
-    log,
-    master,
-    options,
     platform,
     tcp,
     websocket,
 )
 from mitmproxy.connection import Address
 from mitmproxy.flow import Flow
-from mitmproxy.net import udp
-from mitmproxy.proxy import commands, events, layers, server_hooks
-from mitmproxy.proxy import server
+from mitmproxy.proxy import events, mode_specs, server_hooks
 from mitmproxy.proxy.layers.tcp import TcpMessageInjected
 from mitmproxy.proxy.layers.websocket import WebSocketMessageInjected
-from mitmproxy.utils import asyncio_utils, human
-from wsproto.frame_protocol import Opcode
+from mitmproxy.proxy.mode_servers import ProxyConnectionHandler, ServerInstance, ServerManager
+from mitmproxy.utils import human
 
 
-class ProxyConnectionHandler(server.LiveConnectionHandler):
-    master: master.Master
-
-    def __init__(self, master, r, w, options, timeout=None):
-        self.master = master
-        super().__init__(r, w, options)
-        self.log_prefix = f"{human.format_address(self.client.peername)}: "
-        if timeout is not None:
-            self.timeout_watchdog.CONNECTION_TIMEOUT = timeout
-
-    async def handle_hook(self, hook: commands.StartHook) -> None:
-        with self.timeout_watchdog.disarm():
-            # We currently only support single-argument hooks.
-            (data,) = hook.args()
-            await self.master.addons.handle_lifecycle(hook)
-            if isinstance(data, flow.Flow):
-                await data.wait_for_resume()
-
-    def log(self, message: str, level: str = "info") -> None:
-        x = log.LogEntry(self.log_prefix + message, level)
-        asyncio_utils.create_task(
-            self.master.addons.handle_lifecycle(log.AddLogHook(x)),
-            name="ProxyConnectionHandler.log",
-        )
-
-
-class Proxyserver:
+class Proxyserver(ServerManager):
     """
     This addon runs the actual proxy server.
     """
-
-    tcp_server: Optional[base_events.Server]
-    dns_server: Optional[udp.UdpServer]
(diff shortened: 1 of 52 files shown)


decathorpe:
This looks remarkably similar to what I looked into doing just earlier today.
So from my (WireGuard server mode) POV this looks good, AFAICT 👍🏼

mitmproxy/proxy/mode_servers.py
```diff
@@ -0,0 +1,263 @@
...
+    async def start(self):
+        assert not self.server
+        self.server = await udp.start_server(
+            self.handle_dns_datagram,
+            self.mode.listen_host or "127.0.0.1",
+            self.mode.listen_port,
+            transparent=False
+        )
+        addrs = {f"{human.format_address(s)}" for s in self.listen_addrs}
+        if self.mode.data in ["", "regular"]:
```
Meitinger, line 208: I think this should be `"resolve-local"` to match `mode_specs.py`.

Prinzhorn:
I was thinking what if we deprecate `--mode` with `--listen-port` and `--listen-host`? I don't think re-using `--mode` with the new semantics makes sense (it's more than just mode), maybe call it `--listen` as a new option?

mitmproxy/addons/proxyserver.py
```diff
@@ -222,156 +151,96 @@ def configure(self, updated):
...
-        asyncio_utils.set_task_debug_info(
-            task,
-            name=f"Proxyserver.handle_connection",
-            client=handler.client.peername,
-        )
-        try:
-            await handler.handle_client()
-        finally:
-            del self._connections[connection_id]
+            for ret in await asyncio.gather(*[m.start() for m in new_instances], return_exceptions=True):
```
meitinger, line 220: Could there be the case where a server instance needs to be notified of a configuration change, where the mode stays the same but other settings get changed? If so maybe do something like
```python
            new_instances: list[ServerInstance] = []
            existing_instances: list[ServerInstance] = []
            for spec in new_servers:
                if existing := self.servers.get(spec, None):
                    existing_instances.append(instance)
                    new_servers[spec] = existing
                else:
                    instance: ServerInstance = ServerInstance.make(spec, self)
                    new_instances.append(instance)
                    new_servers[spec] = instance

            for ret in await asyncio.gather(
                *[m.start() for m in new_instances],
                *[m.reconfigure() for m in existing_instances],
                return_exceptions=True,
            ):
                if ret:
                    all_ok = False
                    ctx.log.error(str(ret))
```

mitmproxy/addons/proxyserver.py
```diff
@@ -400,23 +269,28 @@ def inject_tcp(self, flow: Flow, to_client: bool, message: bytes):
...
+    def server_connect(self, data: server_hooks.ServerConnectionHookData):
+        if data.server.sockname is None:
+            data.server.sockname = self._connect_addr
+
+        # Prevent mitmproxy from recursively connecting to itself.
+        assert data.server.address
+        connect_host, connect_port, *_ = data.server.address
+
+        for server in self.servers.values():
+            for listen_host, listen_port, *_ in server.listen_addrs:
```
meitinger, line 281: Ideally the server instance should also have a `transport_protocol` property that get's checked for equality. It could very well be the case, that on the other transport someone else is listening, e.g. mitmproxy QUIC on UDP 443, but apache HTTP on TCP 443).

mitmproxy/proxy/mode_specs.py
```diff
@@ -0,0 +1,199 @@
...
+    The transport protocol used by this mode. Used to detect multiple servers targeting the same proto+port.
+    """
+    default_port: ClassVar[int] = 8080
+    __modes: ClassVar[dict[str, type[ProxyMode]]] = {}
+
+    @abstractmethod
+    def __post_init__(self) -> None:
+        """Validation of data happens here."""
+
+    def listen_host(self) -> str:
```
meitinger, line 53: Nit: Is there a reason why `listen_host` and `listen_port` aren't marked `@property` like `type`? Since `transport_protocol` can also be accesses without `()`, it might be more unified.

meitinger (commented):
Looks great, especially having server instance classes (and proxy modes) with your `__init_subclass__` method will allow for cleaner code and easier maintenance/extensibility, very cool!

Once #5393 lands I'll update #5435 and introduce a common `UdpServerInstance` for `DnsInstance`, `QuicInstance` and other.

mitmproxy/proxy/mode_specs.py
mhils: They now take an argument to make it explicit that they depend on `ctx.master.options`.

Issue #5359, opened 2022-05-20
DTLS Support


MaxBosse:
With the recent addition of UDP (for DNS) I was wondering if supporting DTLS is a possibility.

It is being used by many IoT devices and would help investigations in these areas.

mhils:
Absolutely possible, there should not be any major blockers. We only need someone who wants to give the implementation a stab. :)

kckeiks:
Hi, I'd like to start contributing to this project and take a crack at this. I've read your docs and have a high level understanding of the architecture but now I'm going to dive deep in the code ;). Would it be possible to get a high-level overview of what will need to be updated? Any guidance would be appreciated. Thanks in advance.

mhils:
HI @kckeiks! Sure thing:

1. We first receive individual datagrams [here](https://github.com/mitmproxy/mitmproxy/blob/4a2ca7563108d2256909309c770feb0c5b481397/mitmproxy/addons/proxyserver.py#L320-L328) and put them into the associated connection/session/flow/"DNS layer stack". You need to append this to detect DTLS, then make up a matching connection id (the usual host/port 4-tuple should be good enough).
2. After that, the main part is writing your own DTLS layer (["draw the rest of the fucking owl"](https://knowyourmeme.com/memes/how-to-draw-an-owl)). For a first proof-of-concept, I would recommend you just use the logging mechanism to print raw message contents and forward messages as-is. Three existing layers that would be worth looking at are:
   - https://github.com/mitmproxy/mitmproxy/blob/main/mitmproxy/proxy/layers/dns.py
   - https://github.com/mitmproxy/mitmproxy/blob/main/mitmproxy/proxy/layers/tcp.py
   - https://github.com/mitmproxy/mitmproxy/blob/main/mitmproxy/proxy/layers/tls.py
3. The most tricky part will probably be dealing with (py)OpenSSL to do DTLS. I'm not quite sure what the current state there is. https://github.com/python-trio/trio/issues/2010 has some pointers (we can ignore potential MTU and retransmit issues as the client should be doing that), other than that it's pretty much unchartered territory. 😃 

Draft PRs are of course more than welcome! 🍰

kckeiks:
Looks like pyOpenSSL supports some DTLS features https://github.com/pyca/pyopenssl/pull/1026. I started working on this and I will have bandwidth next week. I'm hoping to have a draft soon.

kckeiks:
This is good to close now.

[...]

PR #5397, opened 2022-06-04
Add dtls support

kckeiks:
#### Description
Issue #5359 

#### Checklist

 - [x] parse client hello
 - [x] add raw UDP layer
 - [x] remove test logging
 - [x] Integrate with multi-server mode.


mhils (commented):
Oh wow - this is surprisingly little additional complexity, fantastic! 🥳 I didn't expect the TLS parts to be _that_ reusable for DTLS, this is great news.

High-level comment: I think we should follow a similar structure as we have for TCP, i.e. we add a *raw UDP layer* akin to [`mitmproxy/proxy/layers/tcp.py`](https://github.com/mitmproxy/mitmproxy/blob/main/mitmproxy/proxy/layers/tcp.py). DTLS connections then have a `Proxy Server -> Server DTLS -> Client DTLS -> Raw UDP` layer stack. Does that make sense?

Second high-level comment: Given the similarity in the code, I wouldn't be opposed to reusing the existing `tls_*` hooks instead of adding `dtls_*` versions. We could probably add an `is_dtls` flag to `TlsData`. Either way is perfectly fine with me, feel free to make a call here.

> I created two separate layers, one for client and one for the server, as we currently do for TLS layers but I'm not sure we need that split at the moment.

I agree that we don't, but it's probably easier to keep doing that if we want to reuse the existing code. Again, happy with either approach. FWIW I had the TLS layers combined initially, but that turned out to be more of a mess than anything else. It might be a different story for DTLS of course.

>  configuring ssl context for cookie verification

I'm not too familiar with DTLS, but if I can help you with any pyOpenSSL stuff please feel free to ask dumb questions here or on Slack.

For the current DTLS layer implementation, I've seen that you've basically removed the ["we need to parse the clienthello first"](https://github.com/mitmproxy/mitmproxy/blob/8c151c66a49e91c9625d88f918273cc8bef7144b/mitmproxy/proxy/layers/tls.py#L462-L509) logic, which I guess makes sense for a first PoC. That part can be quite essential though, in particular we need the ClientHello info to do ALPN with the server. Can we somehow make this ClientHello peeking/reassembly work for UDP?

>  remove hard coded server address (Could we pass this information as an option?)

See comment below.

>  log flows in flowlist

See above - I think a raw UDP layer needs to happen for that. I'm happy to take care of that if you want to focus on DTLS. In that case you can probably just use TCPLayer as a child layer in the meantime. You should get away with ignoring that the connection close logic is incorrect, everything else should be pretty much the same for UDP.

Again, thanks for the fantastic PR! 😃

kckeiks:
> High-level comment: I think we should follow a similar structure as we have for TCP, i.e. we add a _raw UDP layer_ akin to [`mitmproxy/proxy/layers/tcp.py`](https://github.com/mitmproxy/mitmproxy/blob/main/mitmproxy/proxy/layers/tcp.py). DTLS connections then have a `Proxy Server -> Server DTLS -> Client DTLS -> Raw UDP` layer stack. Does that make sense?

I need to look closer at what that layer does but I think I know what you mean. I'd like to try building it but I will reach out if I need help. Thank you.

> Second high-level comment: Given the similarity in the code, I wouldn't be opposed to reusing the existing `tls_*` hooks instead of adding `dtls_*` versions. We could probably add an `is_dtls` flag to `TlsData`. Either way is perfectly fine with me, feel free to make a call here.

I was on the fence about this so I'll change it as you're suggesting.

> I'm not too familiar with DTLS, but if I can help you with any pyOpenSSL stuff please feel free to ask dumb questions here or on Slack.

:+1:

> For the current DTLS layer implementation, I've seen that you've basically removed the ["we need to parse the clienthello first"](https://github.com/mitmproxy/mitmproxy/blob/8c151c66a49e91c9625d88f918273cc8bef7144b/mitmproxy/proxy/layers/tls.py#L462-L509) logic, which I guess makes sense for a first PoC. That part can be quite essential though, in particular we need the ClientHello info to do ALPN with the server. Can we somehow make this ClientHello peeking/reassembly work for UDP?

You're right. I was having some issues getting it to work when parsing the client hello so that is why I put it off for the moment. I'll add this to the checklist so I can address this.

mitmproxy/addons/proxyserver.py
```diff
@@ -97,6 +99,14 @@ def _server_desc(self):
...
             transparent=self.options.dns_mode == "transparent",
         )
+        yield "DTLS", self.dtls_server, lambda x: setattr(
+            self, "dtls_server", x
+        ), True, lambda: udp.start_server(
+            self.handle_dtls_datagram,
+            "127.0.0.1",
+            "8084",
+            transparent=False,
+        )
```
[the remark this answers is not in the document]
kckeiks: Yeah, I saw that but I need to look at it closer. I will circle back about this.

kckeiks:
Per the discussion in #5414,  commits https://github.com/mitmproxy/mitmproxy/pull/5397/commits/9d0ae05888b123913baa253e8d7fee5e77b00a9f and https://github.com/mitmproxy/mitmproxy/pull/5397/commits/ebd2fe082ad9be8e369224987963e0b16e08a79a contain a rough wip idea for the `RelayLayer`, `TCPLayer` and `UDPLayer`. I'll have some bandwidth on Thursday and the weekend so I'll address feedback then and improve the code.

kckeiks:
> configuring ssl context for cookie verification 

I'm starting to think that this [feature](https://datatracker.ietf.org/doc/html/rfc6347#section-4.2.1) may not be needed as it would just involve the proxy and server so we don't need to involve the client. Plus, I'm not sure pyOpenSSL supports it atm.

mhils:
> Do we want to add support for this [feature](https://datatracker.ietf.org/doc/html/rfc6347#section-4.2.1) for reverse proxies at the moment?

If I understand things correctly this only concerns DoS by clients? I'd be fine with ignoring that. Realistically speaking, it will always be easy to overwhelm a mitmproxy instance with requests. We don't intend to compete with nginx.

kckeiks:
Thanks, I'll ignore that feature. Yeah, AFAIK it only concerns DoS by clients.

setup.cfg
```diff
@@ -72,6 +72,8 @@ exclude =
     mitmproxy/proxy/config.py
     mitmproxy/proxy/server.py
     mitmproxy/proxy/layers/tls.py
+    mitmproxy/proxy/layers/dtls.py
+    mitmproxy/tls.py
```
kckeiks, line 76: Both the `.coverage.xml` and running `pytest --cov` locally show that this file has 100% coverage but `individual_coverage` says it's at 98%. I'm not sure what the issue is....

setup.cfg
```diff
@@ -72,6 +72,8 @@ exclude =
     mitmproxy/proxy/config.py
     mitmproxy/proxy/server.py
     mitmproxy/proxy/layers/tls.py
+    mitmproxy/proxy/layers/dtls.py
```
kckeiks, line 75: I am adding this file because the DTLS layers are simply sub-classes of TLS layers with some copy and pasted code. I'm unable to copy the `*TLSLayer` tests because they use the `ssl` module which does not support DTLS. It didn't seem worth the time to write tests (creating another version of `SSLTest` that works for DTLS) for the new layers since they are practically identical to the TLS layers which we are already testing. Open to suggestions of course.

kckeiks:
This is ready for another round of review. I will fix the remaining coverage failures this weekend.

mitmproxy/proxy/layers/dtls.py
```diff
@@ -0,0 +1,175 @@
+import struct
```
kckeiks, line 1: I was considering "merging" the DTLS layers with the TLS layers. We could add a "mode" or flag that tells the TLS layer that it is in DTLS mode. This would required that we use if-statements to call DTLS related code which is almost insignificant compared to all the logic the layers share. Thoughts?

kckeiks: We could have something like in this file:
```python
class ClientDTLSLayer(ClientTLSLayer):
       def __init__(self, context: context.Context):
         super().__init__(context, dtls_mode=True)
```

kckeiks:
@mhils I fixed the conflicts. The windows build is failing on the command `mitmweb --version` with `ValueError: Fernet key must be 32 url-safe base64-encoded bytes.` I'm not sure what that is or if it's a known issue?

mhils:
@kckeiks: Thanks! The Windows build error is on me, I'll handle it. :)

mitmproxy/contrib/kaitaistruct/dtls_client_hello.py
```diff
@@ -0,0 +1,203 @@
+# This is a generated file! Please edit source .ksy file and use kaitai-struct-compiler to rebuild
```
mhils, line 1: Can you please add the corresponding .ksy file, or alternatively get this into https://github.com/kaitai-io/kaitai_struct_formats/tree/master/network (I'm sure they woud appreciate it) and https://github.com/mitmproxy/mitmproxy/blob/main/mitmproxy/contrib/kaitaistruct/make.sh?

We also need to compile the ksy with the latest Kaitai compiler as mitmproxy does not always ship with `pkg_resources` (https://github.com/mitmproxy/mitmproxy/issues/5401).

test/mitmproxy/proxy/test_mode_specs.py
```diff
@@ -54,6 +54,7 @@ def test_parse_specific_modes():
     assert ProxyMode.parse("socks5")
     assert ProxyMode.parse("dns").resolve_local
     assert ProxyMode.parse("dns:reverse:8.8.8.8")
+    assert ProxyMode.parse("dtls:reverse:127.0.0.1:0")
```
mhils, line 57: Let's use a valid destination port here in case we ever get to tightening the check.

setup.cfg
mhils: Given how (fantastically) minimal the changes in the DTLS layer now are I think it would make most sense to just merge the TLS and DTLS layers? I think you should be able to add something like
```python
@property
def is_dtls(self):
    return self.conn.transport_protocol == "udp"
```
and then get by with very few if-else constructs in the layers. This would also get around the test coverage issue. I will leave it up to you if you want to leave the `parse_client_hello` stuff in `layers/dtls.py` (we can add a top-level comment there pointing to `tls.py`) or rename it and integrate it into `tls.py` as well. Both is fine with me. :)

mhils (commented):
Sorry again for the long turnaround, but I'm happy to say this looks fantastic now! My only non-trivial suggestion at this point would be to merge the TLS and DTLS layers, I diffed the functions individually and there are pretty much no substantial differences left. Other than that this looks good to merge!

mhils (approved):
Awesome! 🍰 Thank you very much for contribution this feature, it's a fantastic addition to mitmproxy. 😃

kckeiks:
> Awesome! cake Thank you very much for contribution this feature, it's a fantastic addition to mitmproxy. smiley

My pleasure!

I was just in the process of pushing the recompiled (with 0.10) ksy file to this branch but I was too slow :laughing:. Would you like a separate PR for it?

Issue #5398, opened 2022-06-05
Would it be better to have mitmweb use option "content_view_lines_cutoff" as well?


sanlengjingvv:
#### Problem Description
There is an option "content_view_lines_cutoff" is effective in `mitmproxy`.
But it is not work in `mitmweb`.
The web project uses a new constant in _web/src/js/components/contentviews/useContent.tsx_
`export const SHOW_MAX_LINES = 80;`

#### Proposal
Would it be better to have mitmweb use option "content_view_lines_cutoff" as well?

mhils:
Yes, that sounds reasonable if it can be implemented without too much new complexity. 👍

PR #5404, opened 2022-06-07
Make mitmweb use option "content_view_lines_cutoff"

sanlengjingvv:
#### Description

Make mitmweb use option "content_view_lines_cutoff" (#5398)

#### Checklist

### DIFF
--- a/mitmproxy/addonmanager.py
+++ b/mitmproxy/addonmanager.py
@@ -128,7 +128,7 @@ class AddonManager:
         self.master = master
         master.options.changed.connect(self._configure_all)
 
-    def _configure_all(self, options, updated):
+    def _configure_all(self, updated):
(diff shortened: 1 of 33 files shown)


web/src/js/backends/websocket.tsx
```diff
@@ -35,9 +35,9 @@ export default class WebsocketBackend {
     }
 
     onOpen() {
+        this.fetchData("options")
         this.fetchData("flows")
         this.fetchData("events")
-        this.fetchData("options")
```
mhils, line 40: Doesn't that introduce a new race condition on which request finishes first?

sanlengjingvv: Shall I make `connect()` run after `/options` is completed? I don't quite get it here, any advice from you?

PR #5408, opened 2022-06-12
Allow using Delete and Backspace to delete selected flow

KevinSJ:
#### Description

Allow using Delete and Backspace key to delete selected flow

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased: mitmproxy next
 
+* Support using Backspace or Delete to delete selected flow in mitmweb
+  ([#5396](https://github.com/mitmproxy/mitmproxy/discussions/5396), @kevinsj)
 * Support specifying the local address for outgoing connections
   ([#5364](https://github.com/mitmproxy/mitmproxy/discussions/5364), @meitinger)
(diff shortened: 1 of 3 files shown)


mhils:
Thanks! Is there a special reason/convention I'm unaware of for adding backspace? Feels like that's something that is easily fatfingered, maybe just adding the delete key is better?

KevinSJ:
> Thanks! Is there a special reason/convention I'm unaware of for adding backspace? Feels like that's something that is easily fatfingered, maybe just adding the delete key is better?

I feel like backspace is necessary cause some laptops and compact keyboards don't provide direct acces to del key (namely on MacBooks, you have to press fn and backspace together for del, which I feel is a bit inconvinient). And to me personaly, I almost never use del key. 

Do you think we should give users a choice on this? I can try to add a new option group (keyboard options) and include a checkbox saying 'Use backspace to delete' in the options tab? Let me know what you think.

mhils:
Sorry for not getting back to you here - somehow lost track of this.

Support for "Delete" has been added in https://github.com/mitmproxy/mitmproxy/pull/5896/files. I'm still 👎 on adding Backspace by default. Making this configurable is a bit out of scope for the moment, so "Delete" it is for now! Thanks again for raising this! 🍰

Issue #5283, opened 2022-04-23
UDP Refinements


mhils:
- [ ] Later (not this PR): Add raw UDP layer - will be useful as a fallback for everything.
- [x] Dumper: Only dump on DNS response (as we do for HTTP)
  - [x] This probably means that we should time out DNS reqs timely, ideally using RequestWakeup introduced in https://github.com/mitmproxy/mitmproxy/pull/5144. Or with a shortened timeout watchdog.
- [x] ConnectionProtocol Enum
- [x] BypassStateObject
- [x] Move Resolve Code to resolve addon. Then we can require 100% coverage for mitmproxy/dns.py
- [x] str methods
- [x] timestamp_tcp_setup
- [x] DNS layer: state prefix
- [x] Console UI: don't show contentview selector in request/response tabs.
- [x] We UI: Update filt.peg with filter changes.
- [ ] DNS API docs (docs/modd.conf)
- [x] Invalid CNAME data for `dig @127.0.0.1 -p 5354 [alive.github.com](http://alive.github.com/) TXT`
- [x] `mitmproxy -nr flows.mitm` yields `NOTIMP` instead of NOERROR for DNS flows.

PR #5414, opened 2022-06-14
Add support for raw UDP.

meitinger:
#### Description

To support ignoring connections in QUIC and for #5397 & #5283 a raw UDP layer is needed.
With this small change, the (raw) TCP layer works equally well in the meantime (and has UI support), as half-closing will fail in supported way.

Of course I could also extend this PR and rename the TCPLayer to RAWLayer (or StreamLayer?), TCPFlow to RAWFlow etc., so it can be used for both protocols without confusion about the name. (The TCPLayer could also be duplicated under the name UDPLayer, but that's _a lot_ of duplicated code without real added value.)
User entry points like the 'inject.tcp' command or the 'tcp' filter should be kept and filter for/restrict to the underlying protocol.

Comments are very welcome :)

EDIT: UDP layer and flow have now been implemented without basing it on TCP or unifying RAW hooks.

#### Checklist


Meitinger:
> I've been working on a layer similar to your suggestion of `RawLayer` called `RelayLayer` (I was considering renaming it to `TransportLayer` though). I will push the changes to #5397 for review in a bit. Comments are welcome. :)

That's actually a great name as it clears up what the layer itself is supposed to do.

But I'd still prefer to rename the `TCPFlow` to `RAWFlow`, `StreamFlow` or `MessageFlow` and not `RelayFlow`, as in QUIC (and possibly elsewhere), this kind of flow will be needed in cases where neither a TCPLayer or UDPLayer (i.e. `RelayLayer`) is applicable. Or will your `RelayLayer` come with it's own flow and UI?

Meitinger:
> Please correct me if I'm wrong, the goal is to reuse `TCPFlow` or `RawFlow` for QUIC but QUIC will not use `RelayFlow`? In that case, you're right, `RelayFlow` won't work. I think `RawFlow` would be fine then. I wanted to reuse that instead of building a flow and UI for it.

Actually QUIC needs both: A `RelayLayer` (not `RelayFlow`) is needed if either `--ignore` is specified or an addon decides in `tls_clienthello` that we should ignore a connection. This will set `ignore=True` in the `RelayLayer` constructor and not generate a flow, just like it's done now with `TCPLayer`. Data is then relayed at UDP level.

But if ALPN yields something besides h3, QUIC streams will be show similar to `TCPFlow`s now. While this is in principle also a relay of messages, I'm not convinced the flow should be named `RelayFlow`. For one the `RelayLayer` is not used in this case. Also this flow will need to have a more refined protocol field (not just UDP and TCP), that can also be filtered. Maybe `StreamFlow` might be the most appropriate name, as this is what it displays.

So we both agree that renaming `TCPFlow` to something reusable for multiple protocols is the better choice than reimplementing. I've looked at your latest commits [ebd2fe0](https://github.com/mitmproxy/mitmproxy/commit/ebd2fe082ad9be8e369224987963e0b16e08a79a) and [9d0ae05](https://github.com/mitmproxy/mitmproxy/commit/9d0ae05888b123913baa253e8d7fee5e77b00a9f) in #5397 and I think that's a perfect solution to keep backward compatibility. (Perhaps introducing `Dns(Start|Error|Message|End)Hook` to mirror the functionality might be beneficial though?)

@mhils: Any comments/suggestions on a name for a unified `TCPFlow`, then I'd do the renaming work in this PR :)

mhils:
I need to think about this a bit longer, but here's a short raw brain dump:

 - We mainly should think about this from the perspective of users who user our API hooks. Does a single set of events make sense, or would dedicated `tcp_start` / `udp_start` events make more sense? To me it feels like the latter makes more sense, but maybe our heavy API users have some input (@Prinzhorn?)
 - From a backwards compatibility perspective, it would make most sense to have UDPFlow and TCPFlow. I'm not particularly thrilled about potentially renaming the `tcp_` hooks.
 - One downside of TCP/UDP flows/layers is that it's a bit confusing in the sense that higher-level protocols such as HTTP are also TCP (but not TCP flows). From that perspective a different naming would make sense.
 - I dislike StreamFlow as UDP is not a stream-based protocol. RelayFlow is equally non-descriptive.
 - Maybe RawFlow? Then the question is what event hook names we need, `raw_start` is ugli.
 - I wouldn't mind having a UDP layer that is a copy of the TCP layer minus the connection closing ceremony parts. Sure there is a bit of duplication, but it's not particularly bad.

TL;DR: Currently I'm leaning towards separate UDP/TCP layers/flows/event hooks, but I haven't thought about it enough. More input welcome. :)

Meitinger:
> * We mainly should think about this from the perspective of users who user our API hooks. Does a single set of events make sense, or would dedicated `tcp_start` / `udp_start` events make more sense? To me it feels like the latter makes more sense, but maybe our heavy API users have some input (@Prinzhorn?)
> * From a backwards compatibility perspective, it would make most sense to have UDPFlow and TCPFlow. I'm not particularly thrilled about potentially renaming the `tcp_` hooks.

I agree, and @kckeiks's solution actually keeps the old hooks. `RelayLayer` is given the hook classes it should call on start, message, error and end. Currently for UDP `None` is set, as there are no UDP hooks yet, which I think we should introduce to also have this functionality for UDP.

> * I dislike StreamFlow as UDP is not a stream-based protocol. RelayFlow is equally non-descriptive.
> * Maybe RawFlow? Then the question is what event hook names we need, `raw_start` is ugli.

You're right, RawFlow was also my initial goto name.  We can do the same trick mentioned above for `TCPFlow`, i.e. internally (UI, filter, etc.) rename it to `RAWFlow`, but introduce `TCPFlow` and `UDPFlow` classes that inherit from `RAWFlow` and just set it's protocol property. (For QUIC there could then be a `QuicStreamFlow` doing the same thing.)

As a next step, `RelayLayer` - and every other layer that needs `RAWFlow` - will also call `raw_flow_start`, `raw_flow_message`, etc. together with the protocol-appropriate functions. So in `dumper.py`, `view.py`, `intercept.py` etc. we listen for `raw_flow_*`, while protocol-specific addons get e.g. `tcp_start`, `tcp_message`, etc. (Protocol-agnostic addons could also listen for `raw_flow_*`.)

In `flowfilter.py` and the UI we'll also just use the `RAWFlow` class, but use it's protocol property to filter and display.

> * I wouldn't mind having a UDP layer that is a copy of the TCP layer minus the connection closing ceremony parts. Sure there is a bit of duplication, but it's not particularly bad.
> 
> TL;DR: Currently I'm leaning towards separate UDP/TCP layers/flows/event hooks, but I haven't thought about it enough. More input welcome. :)

Of course copying is also a possibility, but then we'd definitely need two more copies, UDP and QuicStream :)

The advantage of having a base class, however, is that it really makes it easy for developers new to mitmproxy to quickly write  an addon for a new protocol that also can show something in the UI, without going through a lot of mitmproxy's internals making changes. (Which also makes it more cumbersome to keep up with main.)

The downside is, that subtleties, like QUIC's stream ids, will either not be shown in the UI, or get encoded into something else, like a host name that then results in a `host:stream:port` display.

Meitinger:
You're right, there should be different hooks. This, however, does not imply the need for different flows, if the same functionality as `TCPFlow` is needed. Take https://github.com/Meitinger/mitmproxy/blob/d94345b2f32aa04ed57dadac99c1ce9724ec15b2/mitmproxy/proxy/layers/quic.py#L251 for example. Right now it uses `TCPFlow` and calls e.g. `TcpMessageHook`. There should definitely be a `QuicStreamMessageHook` instead, but not necessarily a `QuicStreamFlow`. `TCPFlow` could just be renamed to `RAWFlow` and used in `QuicStreamMessageHook`.

The second part of my suggestion was to then call
```python
    yield RawMessageHook(flow)
    # and
    yield QuicStreamMessageHook(flow)
```
so that in mitmproxy `raw_message` can be used to drive the UI, dumper, etc. QUIC related addons would of course implement `quic_stream_message`.

If we opt for duplication of flow classes, then we'd have to do something like
```python
    def handle_message(flow, proto):
        show_some_stuff(f"{proto}: {flow.messages[-1].content}")

    def quic_stream_message(flow: QuicStreamFlow):
        handle_message(flow, "quic")

    def tcp_message(flow: TCPFlow):
        handle_message(flow, "tcp")

    def udp_message(flow: UDPFlow):
        handle_message(flow, "udp")
```
for _every_  protocol and _everywhere_ in mitmproxy where currently `TCPFlow` is handled.
I personally would prefer to just have:
```python
    def raw_message(flow: RAWFlow):
        show_some_stuff(f"{flow.proto} {flow.messages[-1].content}")
```
but I'd happily do the code duplication as well (as it's really just copy/pasting functions and tests, there is no change in functionality regarding the flows between the protocols).

EndUser509:
> for _every_ protocol and _everywhere_ in mitmproxy where currently `TCPFlow` is handled. I personally would prefer to just have:
> 
> ```python
>     def raw_message(flow: RAWFlow):
>         show_some_stuff(f"{flow.proto} {flow.messages[-1].content}")
> ```
> 
> but I'd happily do the code duplication as well (as it's really just copy/pasting functions and tests, there is no change in functionality regarding the flows between the protocols).

Personally I am in favor of having a `RAWFlow` if it is still possible to determine the underlying transport protocol.

Meitinger:
> Edit: even for driving the UI this would make sense internally to have separate hooks. We wouldn't dump everything in a single table but have different views, right? That would make for a much better UX, because like I said, if you proxy two different things and even protocols, you usually want to do different things with them and look at them separately.

The (Web) UI would look the same in both cases. In the first case you would have

```typescript
export interface RAWFlow extends _Flow {
    type: "tcp" | "udp" | "quicstream"
    messages_meta: RAWMessagesMeta,
}

export type Flow = HTTPFlow | DNSFlow | RAWFlow;
```

while in the second you could write

```typescript
export interface TCPFlow extends _Flow {
    type: "tcp"
    messages_meta: TCPMessagesMeta,
}
export interface UDPFlow extends _Flow {
    type: "udp"
    messages_meta: UDPMessagesMeta,
}
export interface QuicStreamFlow extends _Flow {
    type: "quicstream"
    messages_meta: QuicStreamMessagesMeta,
}

export type Flow = HTTPFlow | DNSFlow | TCPFlow | UDPFlow | QuicStreamFlow;
```

which actually collapses to the first case, since all flows have the same `MessagesMeta`.

But since this PR is now a month old and there are more comments against unification, I'll go ahead and introduce appropriate hooks for UDP and QUIC without unified RAW hooks or flows.
In that case I'll use the opportunity to augment QuicStreamFlow with the stream id in the message meta data.

Meitinger:
`udp_*` hooks, `UDPLayer` and `UDPFlow` have now been implemented.
As soon as this PR gets merged, I'll implement the `quic_stream_*` hooks and `QuicStreamFlow` in #5435, also switching from `TCPFlow` to `UDPFlow` for pass-thru in QUIC. (Which would be the first real use-case for `UDPFlow`.)

@EndUser509 : I'll keep `QuicStreamFlow` similar to `UDPFlow` and `TCPFlow`, so you should be able to solve #5455 by listening to all `*_message` hooks and sum up the bytes.

kckeiks:
@Meitinger @mhils #5397 also adds UDPFlow, hooks, etc. How should we resolve?

Meitinger:
It seems that #5397 doesn't add UDP flows to the web UI and in some other places, like `dumper.py`. Maybe use #5414 for UDP and #5397 for DTLS? Since the classes/functions are named the same and in the same location, it shouldn't be hard to switch to this PR for UDP.

Meitinger:
> @Meitinger Ok, I have a feeling my PR will be blocked by #5393 (?) so this might move faster. PS you could've just let me know that you really wanted to do the `UDPLayer`. It's not like I love spending time copying and pasting code, UI-related components, units tests, ect. 😆

I'm sorry, but from your last comment on [June 14th](https://github.com/mitmproxy/mitmproxy/pull/5414#issuecomment-1155362772), I gathered you were doing a RawLayer (RelayLayer), which got more or less rejected. (At least given the comments on this PR.)
I didn't do UDPLayer until my comment [three days ago](https://github.com/mitmproxy/mitmproxy/pull/5414#issuecomment-1186186525), outlining my plan to go forward and copy TCPFlow to UDPFlow and QuicStreamFlow. If you had told me then, I could have added the missing parts to your PR instead.

Although as you can see, it takes more changes to implement UDPFlow fully (1536) than DTLS with parts of UDPFlow (1371), so maybe it's better to have separate PRs for review. (Also the reason why I was hesitant to implement UDPFlow/QuicStreamFlow in #5435).

But again, sorry for the misunderstanding/miscommunication and the extra work you had to do... 😔

mhils (approved):
Sorry for letting you two hanging for so long here, @meitinger and @kckeiks! 😔 I'll promise that cycle times will be much shorter from my end from now on.

> But since this PR is now a month old and there are more comments against unification, I'll go ahead and introduce appropriate hooks for UDP and QUIC without unified RAW hooks or flows.

My lack of communication was not a statement against unification, but of a lack of sufficiently-large chunks of time to play around with things and think about the edge cases. Sorry for causing extra work. I think a RawFlow class with `raw_flow_start`/... hooks sounds very reasonable long term, but I'm also perfectly happy with the "duplicated" version we have here. I don't think I had considered raw QUIC streams in my initial thinking, which tilts my previous assessment slightly in favor of unification.

In the spirit of unblocking you all, I'm happy to merge either approach for now. We can merge this as-is here (and then later make a decision on whether we want to unify under raw_ only, deprecating `tcp_*` and `udp_*`), or switch to a RawFlow instance right away. @meitinger, feel free to make a judgement call on what's easier for your upcoming QUIC work. :)

meitinger:
@mhils No problem, and your fantastic #5481, i.e. having a signed store app, was definitely worth the wait :)

I think it's best to merge this PR now as-is, so @kckeiks and I can continue building upon it.
As for unification: The code already contains some changes in that direction. If more unification is possible and reasonable, I'll introduce it in #5435. (Of course without any UDP flow/layer API changes, so @kckeiks work isn't affected.)

mhils:
Sounds good to me. Resolving the merge conflicts should be straightforward, otherwise LGTM. You should have the necessary permission bits. :)

PR #5419, opened 2022-06-17
fix the mounted local storage permission issue for the docker release

wisicn:
#### Description

add the capability to specific the UID and GID of the local storage which will be mounted to the docker container by -v.

if we don't specify the UID and GID, there will be a lot of issues like 
Permission denied: '/home/mitmproxy/.mitmproxy/mitmproxy-ca.pem' in this [issue #6](https://github.com/mitmproxy/docker-releases/issues/6)

### DIFF
--- a/release/docker/DockerHub-README.md
+++ b/release/docker/DockerHub-README.md
@@ -5,11 +5,11 @@ Containerized version of [mitmproxy](https://mitmproxy.org/): an interactive, SS
 ## Usage
 
 ```sh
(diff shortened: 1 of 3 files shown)


mhils:
Could you please explain why this is necessary?  We do `chown -R mitmproxy:mitmproxy ~/.mitmproxy` in our entrypoint, which should resolve all permission errors.

mhils:
Superseded by https://github.com/mitmproxy/mitmproxy/pull/5550

Issue #5160, opened 2022-03-01
Required protobuf version range is too strict


ngg:
#### Problem Description

The latest released mitmproxy `7.0.4` version depends on `protobuf>=3.14,<3.19`, which is incompatible with the latest protobuf `3.19.4` release. In my project, I have a strict dependency on `protobuf>=3.19` and I cannot downgrade, thus I cannot use the latest released mitmproxy, only a forked version or the main branch.

The main branch currently depends on `protobuf>=3.14,<3.20`, so even if a new mitmproxy version will be released, this issue will reappear when protobuf `3.20.0` is released.

I think this upper bound is too strict and not necessary as protobuf always kept backwards compatibility within a major version.

#### Proposal
Mitmproxy should depend on `protobuf>=3.14,<4`.

mhils:
AFAIK protobuf does not define a public versioning policy at the moment. The best I could find is a comment from 20 days ago stating that they are about to update their strategy: https://github.com/protocolbuffers/protobuf/issues/9487#issuecomment-1034919068.

We're in this weird in between state where some folks use mitmproxy as a library and most use it as an app. To cut a long story short, we don't want to rush emergency releases because one of our dependencies decides to ship a breaking release which would immediately break `pip install mitmproxy` as well. The current ranges are already a concession to make life easier for people who embed mitmproxy. I don't mind extending the permissible range if the dependency does specify a clear versioning policy, but that's not the case for protobuf.

For your specific case, I'd recommend pinning to a specific commit hash or tracking main - pip should handle Git dependencies just fine. Alternatively, we should see a new mitmproxy release in the coming weeks. :)

ngg:
Do you have an ETA for the 8.0.0 release? Or would it be possible to release a 7.0.5 version that accepts the current latest protobuf? (With a <3.20 range, just like the main branch has)

mhils:
My best ETA is "in the coming weeks". I'd rather spend my time on that than pushing a 7.0.5. release, sorry.

ngg:
Tis issue because this is causing a problem for us yet again. The protobuf dependency was updated in https://github.com/mitmproxy/mitmproxy/commit/85e57a91e0cc96affc282d511ce606782ea2047b 25 days ago, but there were no releases since, so we can't update our project to use protobuf 3.21 while using a released version of mitmproxy.

[...]

PR #5429, opened 2022-06-24
accept newer protobuf versions

ngg:
#### Description

ProtoBuf has changed their versioning scheme and now they guarantee backward compatibility within a major version. See https://developers.google.com/protocol-buffers/docs/news/2022-05-06#versioning and https://github.com/protocolbuffers/protobuf/issues/9872#issuecomment-1135962265 for more details.

I have ran the tests with the current latest version 4.21.1, and now they guarantee there will be no breaking change until version 5, so it is safe to use the `protobuf>=3.14,<5` range instead of `protobuf>=3.14,<3.21`.

Fixes https://github.com/mitmproxy/mitmproxy/issues/5160

#### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,7 +82,7 @@ setup(
         "ldap3>=2.8,<2.10",
         "msgpack>=1.0.0, <1.1.0",
         "passlib>=1.6.5, <1.8",
-        "protobuf>=3.14,<3.21",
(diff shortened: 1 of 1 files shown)


mhils:
Awesome, thanks! 🍰

PR #5430, opened 2022-06-24
fix huge (>65kb) http2 responses corrupted

### LINKED ISSUE #5428: [grpc] mitmproxy corrupts huge (>65kb) responses
#### Problem Description
Grpc requests with huge responses fails with error `io.grpc.StatusException: INTERNAL: Encountered end-of-stream mid-frame` if mitmproxy is used. Requests succeed without mitmproxy.
In my case response was 200k bytes.

1. Just make any grpc request that produces huge response

#### System Information
Mitmproxy: 8.0.0
Python:    3.10.2
OpenSSL:   OpenSSL 1.1.1n  15 Mar 2022

Mitmproxy: 9.0.0.dev (+13, commit 3eab9f0)
Python:    3.9.12
OpenSSL:   OpenSSL 3.0.3 3 May 2022

dhabensky:
#### Description

BufferedH2Connection is able to buffer huge response and send it to client, properly following flow control specs.
But trailers are not subject to flow control and were mistankenly sent immediate.
That led to the situation that mitmproxy worked fine on small responses (that fit in 65k bytes - initial size of flow control window) and broke on huge responses.

With this changes I delay trailers until all data gets sent.
I also altered `end_stream` method because it is hard to expose proper stream state to `Http2Connection.is_closed` now.
Without the change an empty data frame with `end_stream` flag was sent before trailers and it also violated protocol.

Fixes #5428

I will remove excessive logging and add tests/changelog in the next iterations.
@mhils are you ok with the solution?

#### Checklist

### DIFF
--- a/mitmproxy/proxy/layers/http/_http_h2.py
+++ b/mitmproxy/proxy/layers/http/_http_h2.py
@@ -1,5 +1,5 @@
 import collections
-from typing import Any, Dict, NamedTuple
+from typing import Dict, List, NamedTuple, Tuple
 
 import h2.config
 import h2.connection
@@ -34,7 +34,7 @@ class BufferedH2Connection(h2.connection.H2Connection):
     """
 
     stream_buffers: collections.defaultdict[int, collections.deque[SendH2Data]]
-    stream_trailers: Dict[int, Any]
+    stream_trailers: Dict[int, List[Tuple[bytes, bytes]]]
 
     def __init__(self, config: h2.config.H2Configuration):
         super().__init__(config)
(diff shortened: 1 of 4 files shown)


mitmproxy/proxy/layers/http/_http_h2.py
```diff
@@ -34,10 +34,12 @@ class BufferedH2Connection(h2.connection.H2Connection):
     """
 
     stream_buffers: collections.defaultdict[int, collections.deque[SendH2Data]]
+    stream_trailers: Dict[int, Any]
```
mhils, line 37: While we're here, could we make this `List[Tuple[bytes, bytes]]` instead of `Any` please (and add type annotations below)? :)

mitmproxy/proxy/layers/http/_http_h2.py
```diff
@@ -77,8 +79,16 @@ def send_data(
...
+        if self.stream_buffers.get(stream_id, None):
+            # Though trailers are not subject to flow control, we need to queue them and send strictly after data frames
+            self.stream_trailers[stream_id] = trailers
+        else:
+            self.send_headers(stream_id, trailers, end_stream=True)
+
     def end_stream(self, stream_id) -> None:
-        self.send_data(stream_id, b"", end_stream=True)
+        if not self.stream_trailers.get(stream_id):
+            self.send_data(stream_id, b"", end_stream=True)
```
mhils, line 91: This path here wasn't immediately clear to me, let's add a short explainer.
```suggestion
        if stream_id in self.stream_trailers:
            return  # we already have trailers queued up that will end the stream.
        self.send_data(stream_id, b"", end_stream=True)
```

mitmproxy/proxy/layers/http/_http_h2.py
```diff
@@ -147,6 +157,8 @@ def stream_window_updated(self, stream_id: int) -> bool:
             available_window -= len(chunk.data)
             if not self.stream_buffers[stream_id]:
                 del self.stream_buffers[stream_id]
+                if self.stream_trailers.get(stream_id):
```
mhils, line 160:
```suggestion
                if stream_id in self.stream_trailers:
```
(We may have an empty trailers frame, which would otherwise evaluate to False)

mhils (commented):
Great catch, thank you very much! 🍰 

Only a few nits, this generally looks good to merge. :)

dhabensky:
Made all changes you proposed. Should I squash now?

mhils (approved):
Awesome, thanks! 🍰 
No need to squash manually, it's nicer for me to see the incremental diffs and then use GitHub's own "squash" button. :)

Issue #4170, opened 2020-08-26
Support for HTTP/3 and QUIC


Aniketh01:
With the advent of the new transport protocol QUIC, a lot of the network traffic would be shifting to speak HTTP/3 soon. A number of services are already speaking HTTP/3 already. It would be nice to if mitmproxy would support these protocols as well.

The current available python-based QUIC/HTTP3 library to use would be aioquic: https://github.com/aiortc/aioquic/.

Would be happy to discuss more this and the future work this could entail :)

mhils:
That's definitely something that's on our radar. The changes involved are quite nontrivial, but @Meitinger has a working fork with QUIC support at https://github.com/Meitinger/mitmproxy. We will probably take a closer look once #1775 has landed.

Aniketh01:
Thanks for letting me know. @Meitinger Is there a working commit with https://github.com/Meitinger/mitmproxy where I can test, listen to the QUIC and HTTP/3 traffic? So far I have tested, and wasn't able to capture any QUIC traffic. Am I missing something here?

alt-glitch:
@Aniketh01 Did you manage to get this fork to work? Moreover how did you test QUIC traffic? I have installed that fork in an Ubuntu VM to try it out but haven't been able to find a way to generate QUIC traffic.

hellodword:
> @Aniketh01 Did you manage to get this fork to work? Moreover how did you test QUIC traffic? I have installed that fork in an Ubuntu VM to try it out but haven't been able to find a way to generate QUIC traffic.

`curl -i --http3 https://quic.aiortc.org`

But you need to build a cURL with HTTP/3 support.

- [cURL - HTTP3 (and QUIC)](https://github.com/curl/curl/blob/master/docs/HTTP3.md)
- [HTTP/3 test servers](https://bagder.github.io/HTTP3-test/)

hope it is helpful :)

daMatz:
Is there any news about HTTP/3 support? Would love to see it for mitmproxy but as QUIC is based on UDP it seems like a bigger task, right?

ericbeland:
Saw this comment go by, which spurred my curiosity on how popular HTTP/3 is? When this was opened, HTTP/3 was probably 2 or 3 percent of all internet traffic, but now it is up to [about 24 percent](https://w3techs.com/technologies/details/ce-http3).

j-h-a:
[...]

PR #5435, opened 2022-06-27
Add QUIC support.

meitinger:
#### Description

This is the PR for QUIC support in mitmproxy.

#### Checklist

 - [x] Stream and datagram relay layer for arbitrary protocols. (See also #5414)
 - [x] Support `TlsClienthelloHook`, `TlsFailedClientHook` and `TlsEstablishedClientHook`.
 - [x] Use hooks for TLS settings (`QuicTlsStartClientHook` and `QuicTlsStartServerHook`).
 - [x] Implement `Http3Client` and `Http3Server`.
 - [x] Support `NextLayer` decisions.
 - [x] ~~Extended UDP implementation to support connection roaming.~~ [dropped by request]
 - [x] ~~Implement and test upstream mode and `via` handling (if possible).~~ [postponed until UDP/TCP mode merge]
 - [x] Write tests.
 - [x] Add entry to CHANGELOG.


mhils:
I haven't looked through `quic.py` yet, but everything else looks fantastic overall. 🍰 

What's the reason for making `read()` return the address as well?

Meitinger:
One of the remaining question is, how do we implement H3 to H2/1 and vice versa?
I've had a look into doing this automatically, but this might not be what we want: mitmproxy would first connect over TCP (i.e. H1 or H2), look at `alt-svc` and then try to connect over QUIC, falling back to TCP when that fails. This is very likely something that the browser already did or is currently trying to do, so it just introduces additional delay, which actually could lead to a wrong browser decision, i.e. going for H2 instead of H3.
So I'd suggest as a default, let's keep it at the same protocol. Which then raises the question, how do we support overriding that decision?
There are three ways that I can see:
1. Introduce a protocol field in `GetHttpConnection` and `Request`, that is initialized with the original protocol but can be changed by a `request` hook.
2. Introduce a command line option that allows mapping host names to protocols that _must_ be used for that host.
3. Introduce a new hook that the `HttpLayer` calls in `get_connection`, which picks the protocol (defaulting to the one used in the client connection).

Personally, I'd prefer the the first choice...

meitinger:
@mhils: Do you have any preference/comments on the aforementioned protocol override mechanism?

meitinger:
With the integration of #5393 I've opted to specify what HTTP version should be used to connect to upstream in the mode string itself. That allows for quite some flexibility while also being user-controllable (i.e. via command line, not requiring to develop an addon that modifies a `Request`).

mitmproxy/proxy/mode_servers.py
```diff
@@ -340,3 +348,77 @@ def make_connection_id(
...
+        )
+        if header.version is not None and header.version not in supported_versions:
+            transport.sendto(
+                encode_quic_version_negotiation(
+                    source_cid=header.destination_cid,
+                    destination_cid=header.source_cid,
+                    supported_versions=supported_versions,
+                ),
+                remote_addr,
+            )
```
mhils, line 391: This needs to happen here because we otherwise don't get a connection id yet?

mhils (commented):
Fantastic! 🍰 

I haven't fully reviewed `layers/quic.py` and `layers/http/_http3.py`, but that's all very well-isolated complexity. My main gripe at the moment is the addition of `remote_addr` to SendData/DataReceived, DatagramReader/Writer, and `ConnectionHandler`. That's touching a lot of core places I would prefer to keep as simple as they are, but of course that's not a breaking deal if there are no easier alternatives.

Other than that I mostly have dumb questions you can probably answer quickly with a short sentence or reference. :-)

mitmproxy/proxy/mode_servers.py
```diff
@@ -340,3 +348,77 @@ def make_connection_id(
...
+        context.server.transport_protocol = "udp"
+        return layers.QuicLayer(context, self)
+
+    async def handle_udp_connection(self, connection_id: tuple, handler: ProxyConnectionHandler) -> None:
+        layer = cast(layers.QuicLayer, handler.layer)
+        try:
+            return await super().handle_udp_connection(connection_id, handler)
+        finally:
+            # clear up all additional connection ids
+            for cid in layer.connection_ids:
```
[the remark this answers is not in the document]
meitinger: That's due to QUIC allowing multiple connection IDs for the same endpoint. (Also as a privacy measure: https://datatracker.ietf.org/doc/html/rfc9000#section-9.5)

mitmproxy/net/server_spec.py
```diff
@@ -3,12 +3,11 @@
 """
 import re
 from functools import cache
-from typing import Literal
 
 from mitmproxy.net import check
 
 ServerSpec = tuple[
-    Literal["http", "https", "tcp", "tls", "dns"],
+    str,
```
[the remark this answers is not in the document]
meitinger: We're already checking the returned scheme in `mode_specs.py`, so instead of adding `h1`, `h2` and `h3` (which will be checked again in `QuicMode` anyways), I think it's better to make `server_spec.parse` scheme-agnostic.

meitinger:
> My main gripe at the moment is the addition of `remote_addr` to SendData/DataReceived, DatagramReader/Writer, and `ConnectionHandler`. 

I think I can work around that. In this case, the QUIC connection ID won't show up in `ServerManager.connections`, and there will be special `QuicDataReceived` and `QuicSendData`.

meitinger:
@mhils: I've (almost completely) rewritten the QUIC layers. Except for one line, `server.py` stays untouched now. `ClientQuicLayer` and `ServerQuicLayer` inherit from `TunnelLayer` and are basically copies from their TLS counterparts, so the code should be very supportable in the future. (We may even consider doing mixins to deduplicate code...)

Roaming is done by a custom layer. It doesn't support the full range as the old implementation did[^1], but without touching `server.py` and special connection id handling this just isn't possible.

Are the changes to `server.py` and `context.py` okay with you? Otherwise I could work around them and query `proxyserver` directly...

[^1]: For example, multiplexing multiple connections (i.e. not just stream) over the same client ip/port. However, most clients don't do that anyway.

mitmproxy/proxy/layers/quic.py
```diff
@@ -0,0 +1,1052 @@
...
+            return
+
+        # transmit data to the peer
+        yield QuicTransmit(peer_connection)
+
+    @expect(events.Start)
+    def state_start(self, _) -> layer.CommandGenerator[None]:
+        # mark the main flow as started
+        if self.flow is not None:
+            yield UdpStartHook(self.flow)
```
mhils, line 410: Do I get this right that there's a fake top level "UDP flow" that is only use for signalling?

mitmproxy/proxy/layers/quic.py
```diff
@@ -0,0 +1,1052 @@
...
+        from_client: bool,
+    ) -> layer.CommandGenerator[None]:
+        # buffer events if the peer is not ready yet
+        peer_quic = self.quic_server if from_client else self.quic_client
+        if peer_quic is None:
+            (self.buffer_from_client if from_client else self.buffer_from_server).append(event)
+            return
+        peer_connection = self.context.server if from_client else self.context.client
+
+        if isinstance(event, quic_events.DatagramFrameReceived):
```
mhils, line 346: Do I see this correctly that DatagramFrameReceived is completely undocumented in aioquic?

mitmproxy/proxy/layers/quic.py
```diff
@@ -0,0 +1,1052 @@
...
+            yield TcpErrorHook(self.flow)
+
+        # report error-free endings and always clear the live flag
+        if self._ended_client and self._ended_server:
+            if self.flow.error is None:
+                yield TcpEndHook(self.flow)
+            self.flow.live = False
+
+
+class QuicStreamLayer(layer.Layer):
```
mhils, line 294: Name suggestion: RawQuicStreamsLayer?

mitmproxy/proxy/layers/quic.py
```diff
@@ -0,0 +1,1052 @@
...
+    """
+    Layer on top of `ClientQuicLayer` and `ServerQuicLayer`, that simply relays all QUIC streams and datagrams.
+    This layer is chosen by the default NextLayer addon if ALPN yields no known protocol.
```
mhils, line 297: Clarification question: This would also work perfectly fine to proxy HTTP/3 streams, correct?

mitmproxy/proxy/layers/quic.py
```diff
@@ -0,0 +1,1052 @@
...
+            stream = yield from self.get_or_create_stream(event.stream_id)
+            if stream.has_ended(from_client):
+                yield commands.Log(f"Received {len(event.data)} byte(s) on already closed stream #{event.stream_id}.", level="debug")
+                return
+
+            # forward the message allowing addons to change it
+            if stream.flow is not None:
+                tcp_message = tcp.TCPMessage(from_client, event.data)
+                stream.flow.messages.append(tcp_message)
+                yield TcpMessageHook(stream.flow)
```
mhils, line 368: FYI one reason why HttpStream subclasses layer.Layer is that streams can block independently. Here we handle all message hooks in QuicStreamLayer (the HTTP equivalent would be HttpLayer), which means that one intercepted stream blocks all other streams.

This here is definitely good enough for a first iteration though.

mhils (commented):
Preliminary review - very nice work! I still need to grok connection roaming, but I figured I'll leave you with some initial impressions already.

My main question would be if you already have any idea on how we can approach testing best. We definitely want as many playbook tests as possible here to give us some confidence in the code and make sure that we don't regress when changing things.

mitmproxy/proxy/mode_specs.py
```diff
@@ -236,3 +238,15 @@ class DnsMode(ProxyMode):
 
     def __post_init__(self) -> None:
         _check_empty(self.data)
+
+
+class Http3Mode(ProxyMode):
+    """
+    A regular HTTP3 proxy that is interfaced with `HTTP CONNECT` calls (or absolute-form HTTP requests).
```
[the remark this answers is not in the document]
meitinger: Yes, Chromium (with _a lot_ of debugging to get the proper command line):
```shell
chrome.exe \
    --proxy-server="http=quic://localhost:8080;https=quic://127.0.0.1:8080" \  # needs to be set per proto
    --force-fieldtrial-params=QUIC.Enabled:enable_quic_proxies_for_https_urls/true \  # flag buried deep in Finch
    --origin-to-force-quic-on=127.0.0.1:8080 \  # only way to get around a custom root CA, --ignore-certificate-errors-spki-list doesn't work for proxies
    https://mitmproxy.org
```

Currently, as you pointed out, this triggers https://github.com/mitmproxy/mitmproxy/blob/891dae031c65a3268a18056bff9719ac9ee2a66f/mitmproxy/proxy/layers/http/_http2.py#L605 and is not implemented.

mitmproxy/addons/proxyserver.py
```diff
@@ -188,6 +190,12 @@ def load(self, loader):
             None,
             """Set the local IP address that mitmproxy should use when connecting to upstream servers.""",
         )
+        loader.add_option(
+            "quic_connection_id_length",
+            int,
+            8,
+            """The length in bytes of local QUIC connection IDs.""",
+        )
```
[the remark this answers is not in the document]
meitinger: It's the only settings that's actually used in multiple places, but otherwise... I'll remove it :)

mitmproxy/proxy/layers/quic.py
```diff
@@ -0,0 +1,1052 @@
...
+            if self.flow.error is None:
+                yield TcpEndHook(self.flow)
+            self.flow.live = False
+
+
+class QuicStreamLayer(layer.Layer):
+    """
+    Layer on top of `ClientQuicLayer` and `ServerQuicLayer`, that simply relays all QUIC streams and datagrams.
+    This layer is chosen by the default NextLayer addon if ALPN yields no known protocol.
+    It uses `UDPFlow` and `TCPFlow` for datagrams and stream respectively, which makes message injection possible.
```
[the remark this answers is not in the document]
meitinger: Yes, and I think it's also what WebTransport uses (https://web.dev/webtransport/#datagram).
aioquic's own client/server examples support it:
https://github.com/aiortc/aioquic/blob/239f99b8a3d4f5bc88cb280df765f35722cefe57/examples/http3_server.py#L249

mitmproxy/proxy/layers/quic.py
meitinger: And datagrams :)

mitmproxy/proxy/layers/quic.py
meitinger: Yes, I've actually tested that in the previous implementation (before multi-server). This layer is basically TCPLayer/UDPLayer for QUIC.

mitmproxy/proxy/layers/quic.py
meitinger: Well in that case... I was actually exploring the idea of doing `QuicStreamLayer` differently before, namely splitting up the  client connection per stream, and merging it back together at the server. That would allow for doing `NextLayer` per stream (implementing DNS over QUIC would be a breeze), but also ease implementation of CONNECT/upstream support. With the new multi-server mode I dropped it, because of how connections are registered now, since I planned to use proxyserver's connections for routing. But if the additional complexity of layer routing like in HTTP is desired anyways, I'll give it another try :)

mitmproxy/addons/proxyserver.py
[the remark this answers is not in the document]
mhils: Let's just make it a constant somewhere, I don't see why users would want to configure it.

mitmproxy/proxy/layers/quic.py
mhils: Awesome!

mhils:
FYI, I have an experimental branch based on your work at https://github.com/meitinger/mitmproxy/compare/quic...mhils:mitmproxy:pr-5435. I hope I'll find some time over the weekend to continue, at least the weather looks promising. 🙃 Feel free to merge this in here, but I didn't want to impose myself with unfinished ideas that may still change.

The main changes are in https://github.com/mhils/mitmproxy/commit/149303ddc3323c6011ce7d82b9aaf5f5e5c48c2b. Most importantly, I have started the work to transition from sublayers working with QuicConnection objects directly to sublayers receiving (enhanced) Connection Events and emitting (enhanced) SendData events. The main reason for this is testability: I ultimately want to eliminate QuicTransmit and instead have the Http3Connection layers yield raw SendData commands (enhanced with stream id). This allows us to test HTTP/3 without any of the crypto stuff and write proper playbook tests.

To make sure that this is feasible, I have started by refactoring the `QuicStreamLayer` to a demultiplexing `RawQuicLayer` which now has raw UDPLayer/TCPLayer substacks (pretty similar to what we discussed in https://github.com/mitmproxy/mitmproxy/pull/5435#discussion_r965344279). Things are still a bit rough around the edges, but 
```
mitmdump --mode reverse:quic://pgjones.dev@localhost:443 -k --set udp_hosts=pgjones.dev
```
works quite well and 
```
mitmdump --mode reverse:quic://pgjones.dev@localhost:443 -k
```
also _sometimes_ works. 😁

meitinger:
So I merged it and changed a bit, added comments and some not-yet-implemented events and commands.
Having a 1:1 stream id match will not work out properly. We need to keep track of the mapping and connection state.
Using `QuicDatagramLayer` and `QuicStreamLayer` helps `NextLayer` to choose what layer to place on top. `OpenConnection` actually "opens" the stream on the server side (using `OpenQuicStream`, once implemented basically just gets the next free stream id).
By creating new `Client` and `Server` objects, we can set the state independently, make sure we can route injected messages, and also allow layers to set `server` to something else and have a QUIC-to-else bridge.

There's a comment on how to get rid of `QuicTransmit` and fetching `QuicConnection` for H3.

Sorry for all these very rough changes, I just wanted to get these in as early as possible. I'm happy to implement them in full as well, if you agree on the direction they take.

mhils:
This looks fantastic now!

Mostly nits at https://github.com/meitinger/mitmproxy/compare/quic...mhils:mitmproxy:pr-5435. The one architectural change I did was that I replaced `set/get_connection_error` with `ConnectionClosed/CloseConnection` subclasses (and also moved TCP half closes out of the core command - still not sure about this change).

Please feel free to merge that in here, then we need tests and we'll have something mergable! ✨

mhils:
Tests look very good so far. 👍 😃 🍰

mhils (approved):
🎉🎉🎉

NKShukla:
@mhils, have you rolled out this feature in the latest release?

cooperq:
Is there a good way to test this live? I've been trying to test using https://quic.nginx.org on firefox and the facebook messenger app on android but it seems like quic connections are just being blocked. Do I have to enable QUIC somehow? Happy to open a new issue if that is preferred. Using version Mitmproxy: 10.0.0.dev (+177, commit 12960c0)

ShellCode33:
Hey, thanks for your amazing work ! I'm willing to try this out but I'm unable to route UDP trafic through mitmproxy.

I'm running mitmproxy using transparent mode but it doesn't seem to be listening for incoming UDP packets:

```
$ ss -lapute | grep mitmproxy
tcp  LISTEN  0  100   0.0.0.0:http-alt   0.0.0.0:*   users:(("mitmproxy",pid=265930,fd=6)) uid:1000 ino:3299559 sk:5001 cgroup:/user.slice/user-1000.slice/session-3.scope <->
```

Is there any CLI parameter that would trigger UDP listening ?

I'm using the HEAD of the main branch.

mhils:
@ShellCode33: QUIC support is very much experimental/undocumented/opt-in at the moment - there's still a ton of work ahead of us (with no clear timeline yet). 

You can start experimenting with it by running `mitmdump --mode http3` or something like `mitmdump --mode reverse:http3://example.com`. Feedback is welcome. :)

mhils:
You can pass `--mode` multiple times. :)

3052:
https://github.com/mitmproxy/mitmproxy/issues/4170

3052:
does it only work with `mitmdump` at this time? I get this:

~~~
mitmproxy-10.1.5-windows> mitmproxy --mode http3
mitmproxy-10.1.5-windows\mitmproxy.exe: Invalid proxy mode specification: http3
(unknown mode)
~~~

3052:
doesnt work for me:

~~~
mitmproxy-10.1.5-windows> mitmdump --mode http3
mitmproxy-10.1.5-windows\mitmdump.exe: Invalid proxy mode specification: http3
(unknown mode)
~~~

ShellCode33:
IIRC it is not supported. This PR adds some useful primitives to allow mitmproxy to support QUIC at some point, but this is not enough to intercept QUIC traffic transparently yet.

mhils:
See https://mitmproxy.org/posts/releases/mitmproxy10/ for the current state.

3052:
OK so it seems I have to scope the interception to a single domain for time being? but even with that, if I start the proxy and try to enable with Android Studio, it says "Proxy is unreachable", where if I just run the normal `mitmproxy` its fine. so how to I intercept HTTP/3 with MITM Proxy?

ShellCode33:
> and try to enable with Android Studio

I guess you are trying to use the built-in proxy feature of the emulator ?

The emulator does not support UDP proxifying.
 
[From the documentation](https://developer.android.com/studio/run/emulator-networking#proxy) :

> The -http-proxy option forces the emulator to use the specified HTTP or HTTPS proxy for all outgoing TCP connections. Redirection for UDP is not supported.

HTTP3 works over QUIC, which works over UDP. Therefore you either have to setup a wireguard tunnel, or use the [TPROXY feature](https://www.kernel.org/doc/html/latest/networking/tproxy.html) of the Linux kernel.

PR #5441, opened 2022-06-29
 Allow configuring certain CA details, fixes #5386 

### LINKED ISSUE #5386: Make the CA details and filename configurable
#### Problem Description

I'm wrapping mitmproxy and I want to avoid confusion by not calling my root certs `mitmproxy-ca.pem` etc. And also by not having my root cert be issued by `mitmproxy`. So that if you're also using `mitmproxy` it's clear which is which.

#### Proposal

`CONF_BASENAME` looks like it's almost meant to be an option here:

https://github.com/mitmproxy/mitmproxy/blob/8f23a26e355c472725206839125be9d51f548b10/mitmproxy/options.py#L7

and the `organization` and `cn` arguments are already there, but unused in our codebase:

https://github.com/mitmproxy/mitmproxy/blob/8f23a26e355c472725206839125be9d51f548b10/mitmproxy/certs.py#L393

I think having these three things be configurable would already be amazing.

Would we want something like `ca_basename`, `ca_organization` and `ca_cn` in core?

#### Alternatives

I'm aware of https://docs.mitmproxy.org/stable/concepts-certificates/#using-a-custom-certificate-authority but I'd rather have mitmproxy manage the cert for me. It would also still look for the same filename.

Prinzhorn:
#### Description

```sh
mitmdump --set ca_basename=example --set ca_organization=Example --set ca_common_name=example
```

Screenshot from 2022-06-29 11-31-41

So how would I test the "via mitmproxy" part?

Two notes:

1. The onboarding app does not work standalone now, I guess we could do some magic and check for `ctx` and have it still work but with the default "mitmproxy"?
2. Is there a reason only these two options trigger a new CA? If I understand correctly you can change the confdir at runtime but not other aspects such as key_size?

https://github.com/mitmproxy/mitmproxy/blob/af5be0b92817eebb534e05fa0cc45127a70fa113/mitmproxy/addons/tlsconfig.py#L301-L303

#### Checklist

### DIFF
--- a/mitmproxy/addons/tlsconfig.py
+++ b/mitmproxy/addons/tlsconfig.py
@@ -305,7 +305,7 @@ class TlsConfig:
         # If the organization is customized, always append the default to it.
         if ctx.options.has_changed("ca_organization"):
             default_organization = ctx.options.default("ca_organization")
-            organization = f"{ctx.options.ca_organization} (via {default})"
+            organization = f"{ctx.options.ca_organization} (via {default_organization})"
         else:
             organization = ctx.options.ca_organization
 


mitmproxy/addons/onboardingapp/templates/index.html
```diff
@@ -39,29 +39,29 @@ <h5>Manual Installation</h5>
...
+        <li><code>mv {{ ca_basename }}-ca-cert.pem /usr/local/share/ca-certificates/{{ ca_basename }}.crt</code></li>
         <li><code>sudo update-ca-certificates</code></li>
     </ol>
     {% endcall %}
     {% call entry('macOS', 'apple') %}
     <h5>Manual Installation</h5>
     <ol>
         <li>Double-click the PEM file to open the <samp>Keychain Access</samp> application.</li>
-        <li>Locate the new certificate "mitmproxy" in the list and double-click it.</li>
+        <li>Locate the new certificate "{{ ca_basename }}" in the list and double-click it.</li>
```
Prinzhorn, line 58: If it actually the basename that is displayed there?

mitmproxy/addons/onboardingapp/templates/index.html
```diff
@@ -70,14 +70,14 @@ <h5>iOS 13+</h5>
         <li>Use Safari to download the certificate. Other browsers may not open the proper installation prompt.</li>
         <li>Install the new Profile (<samp>Settings -> General -> VPN & Device Management</samp>).</li>
         <li><span class="text-danger"><strong>Important:</strong> Go to <samp>Settings -> General -> About -> Certificate Trust Settings</samp>.
-            Toggle <samp>mitmproxy</samp> to <samp>ON</samp>.</span></li>
+            Toggle <samp>{{ ca_basename }}</samp> to <samp>ON</samp>.</span></li>
```
Prinzhorn, line 73: If it actually the basename that is displayed there?

roniemartinez:
+1 
I need this. Looking forward.

Prinzhorn:
@roniemartinez I'm curious about your use-case

roniemartinez:
@Prinzhorn When distributing an app that uses `mitmproxy`, users are asking why they are downloading a file of another name. I think using `Example (via mitmproxy)` makes it more clearer to the users.

mhils:
@roniemartinez: Out of curiosity, could you provide some details on your app? :)

roniemartinez:
@mhils Sorry but cannot disclose at this moment.

keeely:
Instead of these options, would it not be better to document how to create our own CA toward the bottom of the page on certs:
https://docs.mitmproxy.org/stable/concepts-certificates/
There is mention of the X509v3 extensions required but, I think an example openssl cfg + command(s) would go some way towards removing this requirement.
Also, if you do go ahead with this, I'd request that you don't add '(via mitmproxy)' this should surely be a choice for the user?

Prinzhorn:
> Instead of these options, would it not be better to document how to create our own CA toward the bottom of the page on certs:
> https://docs.mitmproxy.org/stable/concepts-certificates/
> There is mention of the X509v3 extensions required but, I think an example openssl cfg + command(s) would go some way towards removing this requirement.

I would prefer to have this integrated, since generating the cert is only 50% of what needs to be done. E.g. I still want the `mitm.it` cert download links and instructions to work. mitmproxy already contains all the code necessary to generate the certs it needs with the appropriate configuration. I'm shipping mitmproxy as part of a desktop application and I definitely don't want to replicate this logic using sub-processes (cross OS :scream: ) or additional code.

> Also, if you do go ahead with this, I'd request that you don't add '(via mitmproxy)' this should surely be a choice for the user?

No, see https://github.com/mitmproxy/mitmproxy/issues/5386#issuecomment-1144776918 . We don't want to make it trivial to hide the fact that mitmproxy is intercepting the connection.

keeely:
Point noted about the convenience, but I don't really agree with the security aspect.  It's already trivial:

```
openssl genrsa -out myCA.key 2048
openssl req -x509 -new -nodes -key myCA.key -sha256 -days 9999 -out ~/.mitmproxy/mitmproxy-ca-cert.pem
<<< enter your details pretending to be someone else>>>
cp myCA.key ~/.mitmproxy/mitmproxy-ca.pem
cat ~/.mitmproxy/mitmproxy-ca-cert.pem >> ~/.mitmproxy/mitmproxy-ca.pem
```

I suppose now I've posted this it's even more trivial :-/.

PR #5465, opened 2022-07-19
Include information about lazy connection_strategy in related errors.

meitinger:
#### Description

There have been some discussions and issues resolving around `connection_strategy`, where the default eager strategy caused a behavior that might not be obvious to the end user.
This PR augments the error messages in this case to include information about how to switch to the lazy strategy.

#### Checklist


meitinger:
>  If you plan to redirect requests away from this server, consider setting `connection_strategy` to `lazy` to suppress early connections.

That's definitely a better message, as it includes the possible intend of the user. It's been changed in d56df4e0281624bfd88bc11dcfb9f9099ca51ae5.

mhils (approved):

PR #5481, opened 2022-07-24
simplify ci build script, add MSIX installer and Microsoft store.

mhils:
This PR:

 - Replaces our ~700 line `cibuild.py` with a much more straightforward ~300 line `build.py`. The added bonus is that it now also works locally out of the box, which is great for future iterations.
 - Adds an MSIX installer and the Microsoft Store as a deployment target: https://www.microsoft.com/store/productId/9NWNDLQMNZD7
 - Moves Docker builds into its own script without third-party dependencies, which is good for security.
 - Removes test coverage for `release/`. The tests here have never caught anything, but did require the code to be unnecessarily abstract and harder to understand.

There are two changes w.r.t. conventions:

 - Tags are just "1.2.3" in the future, not "v1.2.3". The "v" serves no purpose other that it needs to be removed manually everywhere.
 - For iterating on the CI infrastructure, we now have a `citest` branch. This deploys to https://mitmproxy.org/downloads/#branches/citest/, the `mitmproxy/mitmproxy:citest` Docker tag, and a "flight" (an subset of users that receives different binaries) on the Microsoft Store with currently just me (please add yourself if you see a need).


release/windows-installer/AppxManifest.xml
```diff
@@ -0,0 +1,85 @@
...
+                        <desktop:ExecutionAlias Alias="mitmdump.exe" />
+                    </uap3:AppExecutionAlias>
+                </uap3:Extension>
+            </Extensions>
+        </Application>
+    </Applications>
+    <Extensions>
+        <desktop2:Extension Category="windows.firewallRules">
+            <desktop2:FirewallRules Executable="mitmproxy.exe">
+                <desktop2:Rule Direction="in" IPProtocol="TCP" LocalPortMin="1" LocalPortMax="65535" Profile="all"/>
```
meitinger, line 68: Is there a reason why `LocalPortMin` and `LocalPortMax` are specified? If they are missing, all ports are granted. (In contrast to specific ports with range 1-65535).

Issue #5491, opened 2022-07-26
8.1.1: "OSError: [Errno 48] Address already in use" -- error message could use improvement


jcollum-nutrien:
#### Problem Description

```
~ mitmweb
Traceback (most recent call last):
  File "/opt/homebrew/bin/mitmweb", line 33, in <module>
    sys.exit(load_entry_point('mitmproxy==8.1.1', 'console_scripts', 'mitmweb')())
  File "/opt/homebrew/Cellar/mitmproxy/8.1.1/libexec/lib/python3.10/site-packages/mitmproxy/tools/main.py", line 144, in mitmweb
    run(web.master.WebMaster, cmdline.mitmweb, args)
  File "/opt/homebrew/Cellar/mitmproxy/8.1.1/libexec/lib/python3.10/site-packages/mitmproxy/tools/main.py", line 114, in run
    return asyncio.run(main())
  File "/opt/homebrew/Cellar/python@3.10/3.10.5/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/opt/homebrew/Cellar/python@3.10/3.10.5/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/base_events.py", line 646, in run_until_complete
    return future.result()
  File "/opt/homebrew/Cellar/mitmproxy/8.1.1/libexec/lib/python3.10/site-packages/mitmproxy/tools/main.py", line 111, in main
    await master.run()
  File "/opt/homebrew/Cellar/mitmproxy/8.1.1/libexec/lib/python3.10/site-packages/mitmproxy/master.py", line 48, in run
    await self.running()
  File "/opt/homebrew/Cellar/mitmproxy/8.1.1/libexec/lib/python3.10/site-packages/mitmproxy/tools/web/master.py", line 83, in running
    http_server.listen(self.options.web_port, self.options.web_host)
  File "/opt/homebrew/Cellar/mitmproxy/8.1.1/libexec/lib/python3.10/site-packages/tornado/tcpserver.py", line 151, in listen
    sockets = bind_sockets(port, address=address)
  File "/opt/homebrew/Cellar/mitmproxy/8.1.1/libexec/lib/python3.10/site-packages/tornado/netutil.py", line 161, in bind_sockets
    sock.bind(sockaddr)
OSError: [Errno 48] Address already in use
```

OK that's cryptic. What address? How do I override it? 

[...]

Issue #5376, opened 2022-05-26
`running` hook is triggered too early, before "address already in use" exit


Prinzhorn:
#### Problem Description

Docs say

> Called when the proxy is completely up and running

In reality `running` is triggered despite the port not being available.

For me this causes a flickering in the UI because I use `running` to tell the parent process that the proxy is ready. But shortly after that the process exits and I'm back in the `stopped` state. So the UI will be in the "Proxy is up" state for like 200ms which looks ugli and buggy.

In v7 we didn't properly exit at all when the port was available and it v6 the `running` hook is _not_ called.

#### Steps to reproduce the behavior:
1. `ncat -l 8080` make the port unavaiable
2. `mitmdump --quiet --scripts addon.py`

```py
def running():
    print("running")
```

You get the following output

```
running
[Errno 98] error while attempting to bind on address ('0.0.0.0', 8080): address already in use
```

#### System Information

I downloaded the 8.1.0 because I currently can't run mitmproxy locally (still on Ubuntu 20.04 with Python 3.8) and can't remember how to download the main snapshots.

```
Mitmproxy: 8.1.0 binary
Python:    3.10.4
OpenSSL:   OpenSSL 3.0.3 3 May 2022
Platform:  Linux-5.13.0-44-generic-x86_64-with-glibc2.31

```

PR #5495, opened 2022-07-27
Include server information in bind/listen errors.

meitinger:
#### Description

With multi server comes multi responsibility... for more verbose errors.
As pointed out in #5491, some platforms don't include host and port information in address-in-use and other related errors.
Before #5393 this wasn't an issue, as there were only two possible suspects, now there can be more.
This PR includes name, host and port in `OSError`s.

#### Checklist

 - [x] Check if `# pragma: no error` in web master is okay. EDIT: now covered
 - [x] Handle address-in-use errors separately and make suggestions if the user didn't enter a port.
 - [x] reduce code duplication of UDP servers


mhils:
> With multi server comes multi responsibility

Where can I press the revert button?

Jokes aside, this is a good idea. https://github.com/mitmproxy/mitmproxy/issues/5376 is tangentially related, but does not have to be tackled in the same PR.

meitinger:
> Jokes aside, this is a good idea. #5376 is tangentially related, but does not have to be tackled in the same PR.

Will look into #5376 next, in a separate PR :)

But I've used the opportunity to simplify UDP server instance code in this PR.

mhils (approved):
👍

PR #5504, opened 2022-07-30
add showhost option in har_dump.py

inceabdullah:
#### Description

I have added **showhost** option in har_dump.py script to convert into har file, replacing address of urls with pretty_host.

Use:
```
mitmdump -nC mitm_dump -s har_dump.py --set showhost=true --set hardump=mitm_dump.har
```

as default, urls look like: 
```json
        "request": {
          "method": "GET",
          "url": "https://11.11.11.11/pagead/viewthroughconversion/878852058/?value=0&guid=ON&script=0",
          ...
}
```

with using pretty_host:
```json
        "request": {
          "method": "GET",
          "url": "https://domain.com/pagead/viewthroughconversion/878852058/?value=0&guid=ON&script=0",
          ...
}
```

#### Checklist

### DIFF
--- a/examples/contrib/har_dump.py
+++ b/examples/contrib/har_dump.py
@@ -24,6 +24,19 @@ from mitmproxy import version
 from mitmproxy import ctx
 from mitmproxy.utils import strutils
 from mitmproxy.net.http import cookies
+import re
+import sys
+
+args = sys.argv[1:]
+
+try:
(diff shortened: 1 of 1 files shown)


mhils:
You should be able to access `mitmproxy.ctx.options`, which makes detecting whether showhost is true much simpler. :)

Issue #5198, opened 2022-03-21
[mitmweb] mitmproxy has crashed when scroll down the page


sanlengjingvv:
#### Problem Description
May be related to websocket.

Error message show on page
```
Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
    at checkForNestedUpdates (http://127.0.0.1:8081/static/app.js:19477:25)
    at scheduleUpdateOnFiber (http://127.0.0.1:8081/static/app.js:18176:13)
    at Object.enqueueSetState (http://127.0.0.1:8081/static/app.js:11577:15)
    at _a.Component14.setState (http://127.0.0.1:8081/static/app.js:349:26)
    at _a.onViewportUpdate (http://127.0.0.1:8081/static/app.js:48134:14)
    at _a.componentDidUpdate (http://127.0.0.1:8081/static/app.js:48100:12)
    at _a.componentDidUpdate (http://127.0.0.1:8081/static/app.js:48002:43)
    at commitLifeCycles (http://127.0.0.1:8081/static/app.js:17400:32)
    at commitLayoutEffects (http://127.0.0.1:8081/static/app.js:19200:17)
    at HTMLUnknownElement.callCallback (http://127.0.0.1:8081/static/app.js:5588:24)

Component Stack:
    at AutoScrollWrapper (http://127.0.0.1:8081/static/app.js:47990:55)
    at ConnectFunction (http://127.0.0.1:8081/static/app.js:40488:50)
    at div
    at MainView (http://127.0.0.1:8081/static/app.js:48178:26)
    at div
    at ProxyAppMain (http://127.0.0.1:8081/static/app.js:50253:7)
    at ConnectFunction (http://127.0.0.1:8081/static/app.js:40488:50)
    at Provider (http://127.0.0.1:8081/static/app.js:40280:23)
```

Error message show on chrome dev tools console
```
react-dom.development.js:23803 Uncaught Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
    at checkForNestedUpdates (react-dom.development.js:23803:15)
    at scheduleUpdateOnFiber (react-dom.development.js:21835:3)
    at Object.enqueueSetState (react-dom.development.js:12467:5)
    at _a.Component14.setState (react.development.js:365:16)
[...]

PR #5507, opened 2022-08-01
fix issue #5198: mitmWeb crash

LIU-shuyi:
#### Description

issue #5198 mitmWeb crashed when scroll the window, and get error message like this:
Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

code Logic:
Function "onViewportUpdate()" is used to control how many flows displayed in flow-table, In component FlowTable(components/Modal/FlowTable.jsx).
When user scroll the "flow-table", the procedure is as follows:
  step1 div whose className is"flow-table" onScroll callback onViewportUpdate() is called
  step2 setState({vScroll, viewportTop}) after calcVScroll()
  step3 componentDidUpdate() is called to update component "FlowTable"
  step4 onViewportUpdate() is called again
  step5 normally calcVScroll() will get same result with step2, and setState() is not called

why crash?
Flow-item's default height is '32px', but when some flow-item's height is not '32px', the viewport.scrollTop will be different in step1 and step3. This was because that step3 is after flow-tiem's render(), and the real flow-item's height is used, such as '38.5px' not '32px'. In this case step5 will do setState(), and do step3-step5 again.

As an example: when the scrollbar is at a specific position, calcVScroll() calculates flow-item 32-52 will be displayed,  in step2. Bacause of a heigher flow-item, calcVScroll() may calculates flow-item 30-50 will be displayed, in step5, and do setStete(). In the next loop of step3-step5, flow-item 30-50 do not contain the higher flow-item, calcVScroll() get result of flow-item 32-52 again, and do component upDate again.....
Result in an endless loop, and then you will get the Crash.

which flow-item has a higher？
The flow-item cannot "play" or "repeat", such as TCP and websocket:
image

How to fix?
My way to fix this bug is to change higher flow-items to default height '32px'.
In FlowColumn() (components/FlowTable/FlowColumns.tsx), there was a if {...} else if {...}, but no else {} for default case. I try to add an else {} branch for default case, and mitmWeb do not crash in my computer.

#### Checklist

### DIFF
--- a/web/src/js/__tests__/components/FlowTable/__snapshots__/FlowColumnsSpec.tsx.snap
+++ b/web/src/js/__tests__/components/FlowTable/__snapshots__/FlowColumnsSpec.tsx.snap
@@ -214,9 +214,7 @@ exports[`should render columns: quickactions 1`] = `
       <tr>
         <td
           class="col-quickactions"
-        >
-          <div />
-        </td>
+        />
       </tr>
     </tbody>
   </table>
--- a/web/src/js/components/FlowTable/FlowColumns.tsx
+++ b/web/src/js/components/FlowTable/FlowColumns.tsx
@@ -223,7 +223,7 @@ export const quickactions: FlowColumn = ({flow}) => {
                 {resume_or_replay}
             </div>
         </td>
-    );
+    )
 }
 
 quickactions.headerName = ''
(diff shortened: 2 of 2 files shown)


mhils:
Thanks for the very clear root cause analysis! 😃 I will see if I can adjust the CSS to fix this without adding a dummy element, but that is just details. Thank you for tracking that down! 🍰

mhils:
Ah, even simpler - thanks for updating this! If you could run `npm test -- -u` to update the snapshots this looks good to merge!

mhils:
Thank you again for the contribution! 🍰

PR #5528, opened 2022-08-15
Replace blinker with custom implementation

mhils:
@meitinger: Your comment on signal type hints was too much of a temptation, so here's a relatively simple Signal implementation that requires a function signature for receivers to be passed on creation and does not depend on Blinker anymore. 😃

### DIFF
--- a/mitmproxy/optmanager.py
+++ b/mitmproxy/optmanager.py
@@ -82,12 +82,12 @@ class _UnconvertedStrings:
     val: list[str]
 
 
-def _sig_changed(updated: set[str]) -> None:
-    ...
+def _sig_changed_spec(updated: set[str]) -> None:  # pragma: no cover
+    ...  # expected function signature for OptManager.changed receivers.
 
 
-def _sig_errored(exc: Exception) -> None:
-    ...
+def _sig_errored_spec(exc: Exception) -> None:  # pragma: no cover
+    ...  # expected function signature for OptManager.errored receivers.
 
 
 class OptManager:
@@ -105,13 +105,13 @@ class OptManager:
 
     def __init__(self):
         self.deferred: dict[str, Any] = {}
-        self.changed = signals.SyncSignal(_sig_changed)
-        self.errored = signals.SyncSignal(_sig_errored)
+        self.changed = signals.SyncSignal(_sig_changed_spec)
+        self.changed.connect(self._notify_subscribers)
+        self.errored = signals.SyncSignal(_sig_errored_spec)
+        self._subscriptions: list[tuple[weakref.ref[Callable], set[str]]] = []
         # Options must be the last attribute here - after that, we raise an
         # error for attribute assignment to unknown options.
         self._options: dict[str, Any] = {}
-        self._subscriptions: list[tuple[weakref.ref[Callable], set[str]]] = []
-        self.changed.connect(self._notify_subscribers)
 
     def add_option(
         self,
--- a/mitmproxy/optmanager.py
+++ b/mitmproxy/optmanager.py
@@ -1,15 +1,13 @@
 import contextlib
 import copy
-from collections.abc import Sequence
+import weakref
+from collections.abc import Callable, Iterable, Sequence
 from dataclasses import dataclass
-import functools
 import os
 import pprint
 import textwrap
 from typing import Any, Optional, TextIO, Union
 
-import blinker
-import blinker._saferef
 import ruamel.yaml
 
 from mitmproxy import exceptions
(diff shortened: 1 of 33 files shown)


meitinger:
Awesome! I didn't know `ParamSpec` was available in `typing_extensions` in 3.9... 🤦‍♂️

Since your solution uses template functions, may I suggest something like:

```python
from typing import Callable, Generic, TypeVar
from typing_extensions import ParamSpec, Concatenate


S = TypeVar("S")
P = ParamSpec("P")


class event(Generic[P]):

    def __init__(self, func: Callable[Concatenate[S, P], None]) -> None:
        self.func = func
        self.receivers: list[Callable[P, None]] = []

    def connect(self, receiver: Callable[P, None]) -> None:
        self.receivers.append(receiver)

    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> None:
        for receiver in self.receivers:
            receiver(*args, **kwargs)


class Test:

    def do_something(self):
        index = 1
        result = "string"
        self.something_happened(result, index)

    @event
    def something_happened(self, result: str, index: int):
        ...


def my_callback(result: str, index: int) -> None:
    print(f"{result}: {index}")


test = Test()
test.something_happened.connect(my_callback)
test.do_something()
```

mhils:
> Since your solution uses template functions, may I suggest something like:

I like the proposal in general, but I feel it's not super clear if events are per class or per instance and the implications of that. The current approach is a bit more unwieldly, but clear about that (and directly maps to singletons). I think of signals/events more as "objects you can trigger/subscribe to" than as "functions with a decorator".

meitinger:
> I like the proposal in general, but I feel it's not super clear if events are per class or per instance and the implications of that. 

Totally agree, hence my comment about including `self` as `sender` to ensure per-instance behavior.

> I think of signals/events more as "objects you can trigger/subscribe to" than as "functions with a decorator".

Fair point, I guess it's more suitable for mitmproxy, as `sender` isn't really that relevant for how signals are used.
Maybe it's also more 'pythonic'. My proposal is of course heavily inspired by C# events (and motivated by not just having e.g. `_sig_view_remove` to simply do typing for a signal but turn it into a signal).

mitmproxy/utils/signals.py
```diff
@@ -1,54 +1,134 @@
...
+    def connect(self, receiver: Callable[P, None]) -> None:
+        assert not asyncio.iscoroutinefunction(receiver)
+        super().connect(receiver)
+
+    def disconnect(self, receiver: Callable[P, None]) -> None:
+        super().disconnect(receiver)
+
+    def send(self, *args: P.args, **kwargs: P.kwargs) -> None:
+        for ret in super().notify(*args, **kwargs):
+            assert ret is None or not inspect.isawaitable(ret)
```
meitinger, line 77: Actually we could `assert ret is None`, since the new implementation requires a `None` return type.

mitmproxy/optmanager.py
```diff
@@ -83,11 +82,19 @@ class _UnconvertedStrings:
     val: list[str]
 
 
+def _sig_changed(updated: set[str]) -> None:
```
meitinger, line 85: Maybe it might be a bit cleaner to define signature functions as inner functions in `__init__` before using them in `SyncSignal` constructor?

mitmproxy/utils/signals.py
```diff
@@ -1,54 +1,134 @@
...
+
+class _AsyncSignal(Generic[P], _SignalMixin):
+    def connect(self, receiver: Callable[P, Awaitable[None] | None]) -> None:
+        super().connect(receiver)
+
+    def disconnect(self, receiver: Callable[P, Awaitable[None] | None]) -> None:
+        super().disconnect(receiver)
+
+    async def send(self, *args: P.args, **kwargs: P.kwargs) -> None:
+        await asyncio.gather(*[
```
meitinger, line 88: Depending on how strict we check `SyncSignal`, maybe also add some `assert`s here.

meitinger (approved):
Great work, I especially like the idea of limiting signal receivers to a `None` return type.

mitmproxy/utils/signals.py
mhils: I suspect that folks will be tempted to reuse existing functions with a non-None return value as receivers, in which case we don't want to crash (but just ignore the return value). :)

mitmproxy/optmanager.py
mhils: Unfortunately that looks ugly due to the extensive whitespace around it (which is enforced by black). I've made it a bit clearer what they are for. :)

PR #5531, opened 2022-08-15
Save filtered button

rnbwdsh:
Add "Save filtered" to mitmweb (backend: optional arg + js: added handler)
Use conexts in backend instead of manually closing
Transfer dumps gzipped instead of plain for speedup
Change Save... to Save, as there is no ...

#### Description

image

#### Checklist


rnbwdsh:
The only test failing is the jest snapshot, and that works as intended as the UI does change, compared to the previous version (see screenshot)

web/src/js/ducks/flows.ts
```diff
@@ -100,6 +101,14 @@ export default function reducer(state: FlowsState = defaultState, action): Flows
...
             }
 
+        case REQUEST_ACTION:
+        case REQUEST_ACTION_FILTERED:
+            let args = action.type == REQUEST_ACTION_FILTERED ? ('?filter=' + state.filter) : ''
+            if (process.env.NODE_ENV !== 'test') {
+                window.location.href = '/flows/dump' + args
+            }
+            return state
+
```
mhils, line 111: On closer inspection, I don't quite understand why we want to deal with dispatch/redux here at all. Can we just put the `window.location.href` part into `FileMenu.tsx`?

mhils (commented):
Thanks for the PR! This looks like a great addition, but I still have some comments on the implementation. 😃

mitmproxy/tools/web/app.py
```diff
@@ -308,14 +310,19 @@ class DumpFlows(RequestHandler):
...
-            fw.add(f)
-
-        self.write(bio.getvalue())
-        bio.close()
+        self.set_header("Content-Encoding", "gzip")
+
+        match = lambda _: True  # noqa: E731
+        with contextlib.suppress(KeyError, IndexError):  # skip if ["filter"] or [0] fails
+            match = flowfilter.parse(self.request.arguments["filter"][0].decode())
+            print("FILTER", self.request.arguments["filter"])
```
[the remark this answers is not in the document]
rnbwdsh: I'll use if statements as it's probably the easiest to understand and it seems to be most consistent with existing stuff.

mitmproxy/tools/web/app.py
```diff
@@ -308,14 +310,19 @@ class DumpFlows(RequestHandler):
...
         self.set_header("Content-Type", "application/octet-stream")
-
-        bio = BytesIO()
-        fw = io.FlowWriter(bio)
-        for f in self.view:
-            fw.add(f)
-
-        self.write(bio.getvalue())
-        bio.close()
+        self.set_header("Content-Encoding", "gzip")
```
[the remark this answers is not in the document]
rnbwdsh: Removed. I guess this would be a case for another PR anyways, doing it for all, if Accept-Encoding contains gzip, but it's not that important as everything runs local anyways.

web/src/js/ducks/flows.ts
rnbwdsh: It was nearly that convoluted before, but you are right. I just didn't know how to access the relevant filter state from the component, but the sibling files showed me.

mitmproxy/tools/web/app.py
```diff
@@ -309,13 +309,17 @@ def get(self):
...
-        bio = BytesIO()
-        fw = io.FlowWriter(bio)
-        for f in self.view:
-            fw.add(f)
-
-        self.write(bio.getvalue())
-        bio.close()
+        try:
+            match = flowfilter.parse(self.request.arguments["filter"][0].decode())
+        except (KeyError, IndexError, ValueError):  # Key+Index: ["filter"][0], Value: parsing problem
```
mhils, line 314: We should return an HTTP error if the filter is not valid (and add a test for this), so let's not just catch the ValueError here. :)

mhils (commented):
Looks good now, please address the comment below and update the snapshots (`npm test -- -u` I think), then this should be good to go!

mitmproxy/tools/web/app.py
rnbwdsh: Ok, I split the except up and threw a 400 (without reflecting the filter, as that would be reflected XSS)

And I ran the update command, which updated a HeaderSpec.tsx.map

Just to be sure:
The files app.css, app.js and vendor.css are also dirty but I shouldn't upload them as that's updated by some CI/CD?

rnbwdsh:
I rebased and for me, on my branch, the codecov stuff doesn't run.
But I guess I'd need to add (more) jest tests?

mitmproxy/tools/web/app.py
mhils: Yes. We include the compiled web assets in the repo so that Python devs don't need to deal with an entire Node toolchain. We manually update them before releases to avoid merge conflicts in the meantime.

mhils:
We can ignore the coverage warnings here, but there is still a test that fails: https://github.com/mitmproxy/mitmproxy/runs/8017484011?check_suite_focus=true#step:7:1445

rnbwdsh:
@mhils should be fine now.
Problem was, that a unary ! is interpreted as a regex, and therefore a valid filter.
Made it a "[" which isn't a valid regex.

mhils (approved):

mhils:
Thanks again! 🍰

PR #5534, opened 2022-08-16
Improved word searching in console flow 

Kuba1333:
When searching for a word in flow's request or response (by clicking "/") you can see highlighted word that you typed. Before a searched word was highlighted with whole line, and when lines were long, finding it was problematic.


mhils (commented):
Thanks for the PR! I'm generally happy with the planned change, but this PR still needs some work.

 - Highlighting is off for non-trivial regexes, e.g. `(Foo)` instead of `Foo`. It probably makes sense to use `re.split` or so.
 - From very limited testing this introduces new bugs, e.g. a crash when searching for an empty string.
 - Please add type annotations for all new functions.
 - We should think about how we can make some parts of it testable, and then add tests.

PR #5542, opened 2022-08-21
Simplify transparent mode, fix listening on port 0

mhils:
fat-fingered this in #5541, trying again...

### DIFF
--- a/mitmproxy/proxy/commands.py
+++ b/mitmproxy/proxy/commands.py
@@ -119,15 +119,6 @@ class StartHook(Command, mitmproxy.hooks.Hook):
         return super().__new__(cls, *args, **kwargs)
 
 
-class GetSocket(ConnectionCommand):
-    """
-    Get the underlying socket.
-    This should really never be used, but is required to implement transparent mode.
-    """
-
-    blocking = True
-
-
 class Log(Command):
     message: str
     level: str
--- a/mitmproxy/proxy/events.py
+++ b/mitmproxy/proxy/events.py
@@ -3,7 +3,6 @@ When IO actions occur at the proxy server, they are passed down to layers as eve
 Events represent the only way for layers to receive new data from sockets.
 The counterpart to events are commands.
 """
-import socket
 import warnings
 from dataclasses import dataclass, is_dataclass
 from typing import Any, Generic, Optional, TypeVar
@@ -110,12 +109,6 @@ class OpenConnectionCompleted(CommandCompleted):
 class HookCompleted(CommandCompleted):
     command: commands.StartHook
     reply: None = None
-
-
-@dataclass(repr=False)
-class GetSocketCompleted(CommandCompleted):
-    command: commands.GetSocket
-    reply: socket.socket
 
 
 T = TypeVar("T")
--- a/mitmproxy/proxy/layers/modes.py
+++ b/mitmproxy/proxy/layers/modes.py
@@ -4,7 +4,7 @@ from abc import ABCMeta
 from dataclasses import dataclass
 from typing import Optional
 
-from mitmproxy import connection, platform
+from mitmproxy import connection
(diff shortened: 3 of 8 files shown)


meitinger:
Looks like we both decided to remove `GetSocket`: https://github.com/mitmproxy/mitmproxy/pull/5435/files#diff-69c9809e4653b511d34e90e75c40aa2449f8b3bd2f4c08809eb1ddcd4bd00c0f :)

mhils:
Args, sorry 😅 Needed it for https://github.com/mitmproxy/mitmproxy/pull/5543 because original destination resolution is async  there..

meitinger:
Not a problem, I'm just glad `GetSocket` is finally gone 😅

meitinger:
But there are some issues with the listening on port 0 code:
- the socket for determining the port is IPv4, `bind` is called with the host string, which could be IPv6, or there could be no IPv4 interface at all (in which case `socket.socket` fails)
- if the address is a host name, `bind` will block since it's not using the async resolve (technically also not allowed in the [doc](https://docs.python.org/3/library/socket.html#socket.socket.bind))
- a free IPv4 port doesn't necessarily guarantee that the IPv6 port is also free, which would cause `create_server` to fail

I guess there's a reason why the Python folks decided to document the behavior of listening to port 0 rather than fix it :)

I'd be happy to give it a try, though, and basically do what [base_events](https://github.com/python/cpython/blob/129998bd7b133defa37c7529bfad9052c0022b5c/Lib/asyncio/base_events.py#L1482) does, only with an added first-succeeded port in the loop (and a bit simplified). This would still best effort, but without any of aforementioned `raise`s or blocks.

mhils:
Honestly this should just be properly fixed in CPython. The *"What do you do if a port is bound for IPv4, but is in use for IPv6?"* question seems to be mostly academic to me, at least I get fresh ports for each socket on all my Windows and Linux test machines. The current behavior (where you get different ports) is just painful for everyone.

Apart from that your points are perfectly valid of course, no question about that. I don't want to overinvest into an edge case though. How about we just do the following:

```python
            # workaround for https://github.com/python/cpython/issues/89856:
            # We want both IPv4 and IPv6 sockets to bind to the same port.
            # This may fail for various reasons (https://github.com/mitmproxy/mitmproxy/pull/5542#issuecomment-1222803291),
            # so we try to cover the 99% case and then give up and fall back to what asyncio does.
            if port == 0:
                try:
                    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
                    s.bind(("", 0))
                    port = s.getsockname()[1]
                    s.close()
                    self._server = await self.listen(host, port)
                except Exception:
                    self._server = await self.listen(host, 0)
            else:  
                self._server = await self.listen(host, port)
```

meitinger:
> Honestly this should just be properly fixed in CPython.

Absolutely, `create_server` has some other oddities as well...

> I don't want to overinvest into an edge case though.

Sure, we can keep it simple :)


Just for comparison, that's the shortest code I can come up with that should work in all cases and re-uses the test socket:
```python
    sockets = list[socket.socket]()
    try:
        for family, socktype, proto, _, sockaddr in await loop.getaddrinfo(host, port, proto=socket.IPPROTO_TCP):
            sock = socket.socket(family, socktype, proto)
            sockaddr[1] = port
            try:
                sock.bind(sockaddr)
            except OSError as e:
                if e.errno != errno.EADDRINUSE:
                    raise
            else:
                port = sock.getsockname()[1]
                sockets.append(sock)
                sock.setblocking(False)
    except:
        for sock in sockets:
            sock.close()
        raise
    if not sockets:
        raise ...
    server = base_events.Server(loop, sockets, lambda: StreamReaderProtocol(StreamReader(loop=loop), client_connected_cb, loop=loop))
    await server.start_serving()
    return server
```

PR #5543, opened 2022-08-21
add Windows OS proxy mode

mhils:
This is still very much work-in-progress, but eventually should allow a "capture OS traffic" button in mitmproxy
and the option to intercept specific processes. This currently is very similar to the transparent mode functionality
on Windows, and will eventually replace it.

/cc @meitinger


meitinger:
If the given implementation isn't already set in stone, I'd actually like to try the following (at least for UDP):
1. remove launcher dependencies, link with `/MANIFESTUAC:"level='requireAdministrator'"`
2. use pipes for communication (faster IPC and less traffic to filter out)
3. send entire packets over the pipe, no address mingling

mhils:
Not set in stone at all. :-)

1) I did that [earlier](https://github.com/mitmproxy/windows-transparent-mode/commit/70c79531320cec0f53ff048e218788d2a997896b#diff-d0d98998092552a1d3259338c2c71e118a5b8343dd4703c0c7f552ada7f9cb42L1). The problem is that we need to ShellExecute this, which means `asyncio.subprocess` won't work out of the box.
2) Kind of ties into 1) - if.you have some pointers on how to get this working easily without WinElevator I'm all up for it.
3) That won't immediately work for TCP because we need reassembly there... but I do like the idea. Maybe we'll defer this project here until https://github.com/decathorpe/mitmproxy_wireguard has landed. If our Rust TCP reassembly there works as reliably as I hope, maybe we can just add another WinDivert ingestion there next to WireGuard. 

For completeness, we could also use [WinTun](https://www.wintun.net/) and mirror what the [WireGuard client does](https://github.com/WireGuard/wireguard-windows/tree/master/tunnel/firewall) instead of using WinDivert, not sure if that gets us much though.

meitinger:
> The problem is that we need to ShellExecute this, which means `asyncio.subprocess` won't work out of the box.

True, then again, we'll be using named pipes so no need for stdin/stdout.

> if you have some pointers on how to get this working easily without WinElevator I'm all up for it

Something along the lines:
```python
class IncomingPacketsProto(asyncio.protocols.Protocol):
    ...

pipe_name = f"\\\\.\\pipe\\{uuid.uuid1()}"
loop = asyncio.get_event_loop()  # Proactor on Windows
await loop.start_serving_pipe(lambda: IncomingPacketsProto(), pipe_name)
win32api.ShellExecute(0, None, path_to_exe, pipe_name, os.getcwd(), 0)
```
In the binary the pipe can then be opened with `CreateFile`. I'm not too familiar with Rust, but if `std::fs::File` is just a thin wrapper around Win32, then that should work too.

> That won't immediately work for TCP because we need reassembly there...

That's why I sneakily added "(at least for UDP)" 😅
But for TCP and UDP, maybe do a combination: UDP goes over the pipe, TCP get's redirected to the listen port.

If you'd like, I'll also implement it. But I'd use plain old C and make use of `OVERLAPPED`, as the Win32 File API and WinDivert both support it.

mhils:
> ```python
> loop.start_serving_pipe
> ```

_How_ do you know about these things? 😅  See https://github.com/python/cpython/issues/81998, you're not supposed to know. 😄 

I will readily admit that I have no idea what I'm doing/wasn't aware/didn't think of Named Pipes, but it looks like a much cleaner approach. 👍 

> But for TCP and UDP, maybe do a combination: UDP goes over the pipe, TCP get's redirected to the listen port.

No strong preference here on what we do. The current implementation does not really distinguish between TCP and UDP, but adding such special handling for UDP should be easy. Let's see how well @decathorpe's TCP reassembly for WireGuard works out, if we could just shovel all packets into the same pipeline there that would be neat.

> If you'd like, I'll also implement it. But I'd use plain old C and make use of OVERLAPPED, as the Win32 File API and WinDivert both support it.

This part I have strong opinions on, namely no memory-unsafe language in or around mitmproxy 😛 I don't think this work here is blocking anything (please correct me if I'm wrong), so let's get the WireGuard bits rolling and then decide if we build on top (and do Rust<->Rust IPC with Named Pipes) or adjust the current draft here and do asyncio-with-named-pipes. The super nasty bits (dealing with WinDivert) thankfully stay the same[^1].

[^1]: Or are switched out with WinTun 😬🙃

meitinger:
> This part I have strong opinions on, namely no memory-unsafe language in or around mitmproxy 😛 I don't think this work here is blocking anything (please correct me if I'm wrong), so let's get the WireGuard bits rolling and then decide if we build on top [...] or adjust the current draft here and do asyncio-with-named-pipes

Alright, alright, would modern C++ without any `new` be fine? 😅

All joking aside, I'll look into Rust once the time comes and how to get it to play nice with IOCP and `OVERLAPPED`. That way we can do away with all the spawn-multiple-reader/writer-threads and have the `async` pendant of Win32. [Miow](https://docs.rs/miow/latest/miow/) looks promising...
(In the meanwhile I'll refocus on QUIC :)

mhils:
> 3\. That won't immediately work for TCP because we need reassembly there... but I do like the idea. Maybe we'll defer this project here until https://github.com/decathorpe/mitmproxy_wireguard has landed. If our Rust TCP reassembly there works as reliably as I hope, maybe we can just add another WinDivert ingestion there next to WireGuard.

I've started prototyping this idea over at https://github.com/mhils/mitmproxy_wireguard/tree/windows-proxy. Adapting mitmproxy-wireguard to handle different packet sources (not just WireGuard) worked great, but I still need to write the redirecting server to figure out if this is viable. :)

mhils:
Closing in favor of https://github.com/mhils/mitmproxy_wireguard/tree/windows-proxy

PR #5547, opened 2022-08-22
Magisk module onboarding for Android

jorants:
#### Description

When generating certificate files, a Magisk module is generated as well for easier Android onboarding. To get the certificate accepted system wide on Android it needs to be in the system certificate store, but the `/system` partition is no longer writable. For rooted phones with Magisk a module is an easy way to circumvent this. The module is generated with the correct key, and is then served in the onboarding add-on.

Small other things:
 - The tests for the onboarding add-on test more headers
 - `Content-disposition: inline` was changed to `attachment` as we are serving a download.

I do wonder, it seems certificate generation is not automatic for the test suite when testing the onboarding add-on, should `create_store()` not be run when testing instead of committing the certificates to git? I now just added the Magisk module as well.

#### Checklist


jorants:
I am not sure why the ubuntu-18 check was cancelled (I think that is what is holding this up?)

mhils (commented):
Thanks for the PR! This is a neat addition. 😃 

Three comments: 

 - Good catch on the Ubuntu 18.04 brownouts, but please change it to `ubuntu-20.04` then. As the comment indicates we want the oldest supported version. Binaries built in 22.04 won't work on 20.04.
 - Could you move everything magisk-related into `mitmproxy/utils/magisk.py` please? This is too far removed from certificate-related stuff to belong into `certs.py`.
 - `mitmproxy-ca-cert.zip` doesn't make clear that this is Magisk-specific. Can we rename it to `mitmproxy-ca-cert-magisk-installer.zip` or so please? No strong opinions on the particular name, but it should be clear that this is not a general purpose ZIP archive.

Thanks!

jorants:
Cool, done! I dont know what is going wrong with the coverage?

mhils (approved):
Thanks for the quick adjustments!

Codecov is just doing codecov things, it's unfortunately quite flaky. Coverage is fine, this looks good to merge.

mhils:
Ah, it's not just coverage. The Windows tests are failing: https://github.com/mitmproxy/mitmproxy/runs/7975503357?check_suite_focus=true

mitmproxy/utils/magisk.py
```diff
@@ -0,0 +1,109 @@
...
+    ca = get_ca_from_files()
+    der_cert = ca.public_bytes(serialization.Encoding.DER)
+    with ZipFile(path, "w") as zipp:
+        # Main cert file, name is always the old subject hash with a '.0' added
+        zipp.writestr(f"system/etc/security/cacerts/{subject_hash_old(ca)}.0", der_cert)
+        zipp.writestr("module.prop", MODULE_PROP_TEXT)
+        zipp.writestr("config.sh", CONFIG_SH_TEXT)
+        zipp.writestr("META-INF/com/google/android/updater-script", "#MAGISK")
+        zipp.writestr("META-INF/com/google/android/update-binary", UPDATE_BINARY_TEXT)
+        zipp.writestr("common/file_contexts_image", "/magisk(/.*)? u:object_r:system_file:s0")
```
Jurrie, line 106: Out of curiosity: what do these files do? I'm asking because II wrote a Magisk module and only needed `module.prop`, `updater-script`, `update-binary` and the CA certificate...

mhils: I have no particular Magisk experience, if anything @jorants may be able to answer this :)

(Happy to merge a PR that removes those files if this has been tested to work as well)

PR #5548, opened 2022-08-22
Make mitmweb use option "content_view_lines_cutoff"

sanlengjingvv:
#### Description

Make mitmweb use option "content_view_lines_cutoff" (https://github.com/mitmproxy/mitmproxy/issues/5398)

The previous pr was automatically merged as I tried the sync fork feature of github.
The content_view_lines_cutoff is used when viewing details of the request. so the component needs to wait response before rendering.

#### Checklist

### DIFF
--- a/web/src/js/__tests__/components/ProxyAppSpec.tsx
+++ b/web/src/js/__tests__/components/ProxyAppSpec.tsx
@@ -8,7 +8,7 @@ enableFetchMocks();
 
 test("ProxyApp", async () => {
     const cv: ContentViewData = {lines: [[["text", "my data"]]], description: ""}
-    fetchMock.doMockOnceIf("./flows/flow2/request/content/Auto.json?lines=81", JSON.stringify(cv));
+    fetchMock.doMockOnceIf("./flows/flow2/request/content/Auto.json?lines=513", JSON.stringify(cv));
     render(<ProxyApp/>);
     expect(screen.getByTitle("Mitmproxy Version")).toBeDefined();
     await waitFor(() => screen.getByText("my data"));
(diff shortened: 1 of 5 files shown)


web/src/js/components/MainView.tsx
```diff
@@ -6,11 +6,12 @@ import {useAppSelector} from "../ducks";
...
 export default function MainView() {
     const hasSelection = useAppSelector(state => !!state.flows.byId[state.flows.selected[0]])
+    const connection_established = useAppSelector(state => state.connection.state) === 'CONNECTION_ESTABLISHED'
     return (
         <div className="main-view">
             <FlowTable/>
-            {hasSelection && <Splitter key="splitter"/>}
-            {hasSelection && <FlowView key="flowDetails"/>}
+            {(hasSelection && connection_established) && <Splitter key="splitter"/>}
+            {(hasSelection && connection_established) && <FlowView key="flowDetails"/>}
```
mhils, line 14: This shouldn't be a requirement here. (Why) is this necessary?

mhils (commented):
Thanks for the PR and sorry for not reviewing it earlier! This looks good overall, I just don't understand the `connection_established` part.

web/src/js/components/MainView.tsx
sanlengjingvv: There is a bug without this: 
1. Open mitmweb and record some request.
2. Change the `content_view_lines_cutoff` from default to another value
3. Click on a request  to open the flow details.
4. Refresh page.

Result:
It is possible for `/flows/<flow_id>/request/content/Auto.json?lines=<content_view_lines_cutoff>` to start before `/options` gets response. In this case the `content_view_lines_cutoff` will use the default value.

mhils: I see, thanks. That's not ideal, but also not a dealbreaker. We don't want to make flow display conditional on having a connection though, so let's just accept this edge case and remove the special case here. :) We want to be able to show flows even if the connection is lost.

sanlengjingvv: Sure, removed.

mhils (approved):
Awesome, thank you for bearing with my slow replies! 🍰

PR #5549, opened 2022-08-23
Improve handling of `port=0`

mhils:
This is the most dumb approach I've come up with, let's go with that.


meitinger (approved):
LGTM
(And should users actually complain about the bind-close-bind sequence, we can always fall back to something like https://github.com/mitmproxy/mitmproxy/pull/5542#issuecomment-1223758625 later.)

PR #5550, opened 2022-08-23
Docker: assume uid of who holds permissions

mhils:
This PR changes our Docker entrypoint to not blindly `chown` an existing `~/.mitmproxy` folder, but instead just assume the uid/gid of the existing user. This idea was born out of #5419 to make sure that we don't break setups.

Not sure if this is the right approach, I'd definitely appreciate a second pair of eyes with more Docker experience.

### DIFF
--- a/release/docker/DockerHub-README.md
+++ b/release/docker/DockerHub-README.md
@@ -45,6 +45,7 @@ Proxy server listening at http://*:8080
 [...]
 ```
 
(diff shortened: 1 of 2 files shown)


Kriechi:
LGTM if we document somewhere that mitmproxy container will assume the user permission based on the CA cert file (and not the config file if present, or the folder, or the current user from the host.
Or would the folder be better?

mhils:
Thanks, added a line to the DockerHub readme. `mitmproxy-ca.pem` is what we eventually end up reading, so let's go with that. :)

PR #5555, opened 2022-08-25
Mitmweb: Show Proxy State

mhils:
This PR adds a `/state` API call to mitmweb to expose general mitmproxy state outside of flows/eventlogs/options. So far this mostly is the discussed proxy server instance state. I've also added a quick overview thing utilizing that information which is currently displayed if there are no flows yet. The high-level plan would be to extend that with functionality to dynamically spawn additional servers (see https://github.com/mitmproxy/mitmproxy/discussions/5288). And we probably also want a way to resurrect this when there are flows, but hey it's a start. :)

I added `.to_json()` to `ServerInstance` so that individual server instances can provide additional data. For example, it would be super cool to eventually have a WireGuard mode which shares its configuration this way so that we can display a QR code in mitmweb. 😃


web/src/js/__tests__/ducks/_tbackendstate.ts
```diff
@@ -0,0 +1,60 @@
...
+            "Multipart Form",
+            "Image",
+            "Protocol Buffer",
+            "MsgPack",
+            "gRPC/Protocol Buffer"
+        ],
+        "servers": [
+            {
+                "description": "HTTP(S) proxy",
+                "full_spec": "regular",
```
Prinzhorn, line 26: Looking at this I wonder if we should add a property like `expanded_full_spec` (or call the current just `spec`) that's basically the _actual_ full spec, with all the defaults options filled in. Or maybe also a `human_spec` that `to_json` gives us to show users. Right now `full_spec` isn't the full spec, it's just what the option was set to. I would like both please :smile: . I'll probably just render a list of `<li>{actual_full_spec}</li>` in my UI.

1. Parse spec
2. It sets all the default options not specified in the option
3. Serialize it back with the full options (host, port, etc.) to a spec that would result in the exact same listener but verbose and human friendly

Prinzhorn (changes requested):
I'll comment in a minute, I can't type in this tiny "Finish your review" window

mitmproxy/addons/proxyserver.py
```diff
@@ -31,9 +31,9 @@
 
 class Servers:
     def __init__(self, manager: ServerManager):
-        self.updating = signals.AsyncSignal(lambda old_modes, new_modes: None)
+        self.updating = signals.AsyncSignal(lambda: None)
```
meitinger, line 34: I'm a bit sad to see this go, as any listener will now have to keep a copy of the instances to get a diff, but I guess if it's not needed then it's just needless complexity.

mitmproxy/proxy/mode_servers.py
```diff
@@ -122,6 +122,21 @@ async def stop(self) -> None:
     def listen_addrs(self) -> tuple[Address, ...]:
         pass
 
+    @property
+    @abstractmethod
+    def log_desc(self) -> str:
```
meitinger, line 127: Maybe we should use this opportunity and rename it to `description`, as in it's JSON. (And also because it's not just used in logging anymore.)

Prinzhorn:
Something isn't working with the signaling and `_sig_servers_update`. If you look at the WS communication it is jumping from 8080 to 8082 with `is_running: true`. It's missing the `is_running: false` update for 8082.

```py
import asyncio
from mitmproxy import ctx


async def running():
    await asyncio.sleep(10)
    ctx.options.update(mode=["regular@8082"])
```

image

I think the `self.updated.send()` and `self.updating.send()` are just wrongly placed and don't actually do anything. Or there are calls missing. Because `self._instances` is only set after all listeners are listening. There is never a chance for a `is_running == False`.

meitinger (approved):
Great UX improvement!

Prinzhorn:
FWIW here's how this looks in my product:

image

-----

image

mitmproxy/addons/proxyserver.py
```diff
@@ -31,9 +31,9 @@
...
         """"Notified before any instances are started or stopped"""
-        self.updated = signals.AsyncSignal(lambda old_modes, new_modes: None)
+        self.updated = signals.AsyncSignal(lambda: None)
```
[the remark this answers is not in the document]
meitinger: If I may chip in here: The original idea behind the two in #5512 was to support UI changes, like showing a spinner while updating, for `updating` listeners to alter `new_modes` in order to prevent updates, and to improve performance if someone needed a diff of changed servers.
If that all isn't necessary, one `updated` signal without any argument suffices. (Probably should call it `changed` then, to match `options`.)

meitinger:
> I think the `self.updated.send()` and `self.updating.send()` are just wrongly placed and don't actually do anything. Or there are calls missing. Because `self._instances` is only set after all listeners are listening. There is never a chance for a `is_running == False` inside a signal handler. Before this PR this wasn't an issue, because `old_modes` and `new_modes` was explicitly passed I guess.

That is actually by design. In https://github.com/mitmproxy/mitmproxy/pull/5512#issuecomment-1209136911 there was an agreement that having individual start/stop notifications per server instance was adding too much complexity. (This has actually been implemented once, see https://github.com/mitmproxy/mitmproxy/blob/5b291d6110f93daa834aaf4d84cd01667d14fdd4/mitmproxy/proxy/mode_servers.py )
Now signals are only sent after all changes have been made.

mitmproxy/addons/proxyserver.py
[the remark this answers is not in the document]
mhils: The reason I simplified this was because I modified `new_modes` to be a dict, which then made me trace uses of the signal to adjust those as well. I'm happy to add whatever we need here once there is the first concrete use case in mitmproxy. :)

mitmproxy/addons/proxyserver.py
mhils: See above - absolutely not opposed, happy to add again if we have a use case :smiley:

mitmproxy/proxy/mode_servers.py
mhils: 👍

mhils:
I think @Prinzhorn's point is that we could set _instances before starting the servers so that you'd see the not-yet-running instances?

Then again not sure if necessary at all, starting is just quick. mitmweb's options editor currently (wrongly) updates on every keystroke, still works flawlessly 😅

Prinzhorn:
> I think @Prinzhorn's point is that we could set _instances before starting the servers so that you'd see the not-yet-running instances?

My point was that currently it does nothing because `is_running` is essentially a constant and will _never_ be `false` in mitmweb. That lead me to the conclusion that there must be a bug or else why would `is_running` exist. I think what's happening is that we're now mixing pushing and polling once again, which we've discussed before. I thought we settled purely on polling so that the signal tells us "hey, something changed, please pull the new state". But from what @meitinger said I think that `updating` and `updated` exist to the we can show a spinner. Which comes down to being a very complicated and fragile way to push a boolean to the listener of those to signals. So basically instead of treating both signals identical via `_sig_servers_update` mitmweb would need to toggle a boolean back and forth with the two signals. Which sounds more complicated than having the state per instance, which already works.

> That is actually by design. In [#5512 (comment)](https://github.com/mitmproxy/mitmproxy/pull/5512#issuecomment-1209136911) there was an agreement that having individual start/stop notifications per server instance was adding too much complexity.

We basically already have it per instance now and if the signal would be triggered in a different place then mitmweb would literally be able to show a loading state only for the new listeners, which is the whole idea behind `is_running`.

My conclusion is that if we use a single signal and use it in the right two places then everything works perfectly fine.

1. Have a single `changed` signal
2. Move `self._instances = cast(dict[mode_specs.ProxyMode, ServerInstance], new_instances)` above `for ret in await asyncio.gather(*start_tasks, return_exceptions=True):`
3. Trigger signal once above `for ret in await asyncio.gather(*start_tasks, return_exceptions=True):` and once right below, so that mitmweb will get `is_running` and `last_exception`

Prinzhorn:
If we can get this sorted then everything else looks great to me!

mhils:
Yes, that sounds like a good plan!

meitinger:
Just two quick notes:

> My point was that currently it does nothing because is_running is essentially a constant and will never be false in mitmweb.

`is_running` is `False` if the server instance fails to start. But yes, that's currently the only reason why it's there.

> Which comes down to being a very complicated and fragile way to push a boolean to the listener of those to signals. So basically instead of treating both signals identical via _sig_servers_update, mitmweb would need to toggle a boolean back and forth with the two signals.

The events themselves weren't supposed to push the flag, they were supposed to notify interested parties to check `servers.is_updating`. (Also the reason for this last minute change back then: https://github.com/mitmproxy/mitmproxy/pull/5512/commits/3cd242ed690006fdba6ff69fd90550194b325cfe :)

> My conclusion is that if we use a single signal and use it in the right two places then everything works perfectly fine.

Good idea, but it won't give you a notification about stopping servers, so the UI has to keep track of the modes that are removed and wait/count for the first `changed` event. But I guess that's fine.

Prinzhorn:
> `is_running` is `False` if the server instance fails to start. But yes, that's currently the only reason why it's there.

Right, I forgot about that. But my point about it being a bug holds true, because in this PR the branch that renders the `fa-pause` icon can never ever be executed with the current implementation (the error branch has priority over it). That's the line of code I was basing my whole discussion around.

> Good idea, but it won't give you a notification about stopping servers, so the UI has to keep track of the modes that are removed and wait/count for the first `changed` event. But I guess that's fine.

I'm not sure I can follow the second part. I've checked out the code again (I'm not familiar with this part at all) and I think what you're saying is that we're waiting for all `stop_tasks` to finish stopping before doing anything else. I didn't realize that was happening and that puts things into a different light. Let me go back to my very original point in the other discussion: when I edit listeners in my UI I want _instant_ feedback and the UI needs to immediately update when I click the button (because mitmproxy is the single source of truth it takes a roundtrip and clicking the button doesn't directly do anything in the UI). I'm very latency sensitive and every software ever gets this wrong. And if I can easily avoid even 10ms of delay for user feedback then please let's go for it. That's what I worry about. E.g. you click the "X" button next to a listener to remove it and it takes 100ms (or even more) before the list updates, I don't want that, that would be labeled a bug in my issue tracker.

I really can't help but write walls of text :smile: .

New suggestion: `stop()` does immediately do `self._server = None` (so `is_running == False` again), so if we trigger the signal once more time right before `for ret in await asyncio.gather(*stop_tasks, return_exceptions=True):` then the stopping listeners automatically get the spinner back before they disappear. Does that sound reasonable?

mhils:
I've revised things so that there is only one signal that is fired once before servers are started ("immediately") and then again when servers have been started. Overall restarting servers is really quite fast, so this should be more than enough. Does this look reasonable to everyone? :)

@meitinger: I think this LGTM if you two agree, but I don't want to make your work in #5556 any harder. Feel free to just let this sit for now, I'm happy to fix any merge conflicts here at a later point. :)

meitinger:
> @meitinger: I think this LGTM if you two agree, but I don't want to make your work in #5556 any harder. Feel free to just let this sit for now, I'm happy to fix any merge conflicts here at a later point. :)

LGTM too, and just go ahead and push the big green button if @Prinzhorn agrees as well :)
(I'll have to work on tests in #5556 anyways, so I'd rather see this PR and the UX benefits it brings in main than have it delayed.)

Prinzhorn:
Thank you very much! This seems to do exactly what I had in mind and LGTM.

One last note regarding "but it's fast": for me switching from 8080 to 8082 takes 4ms. So yes, you never actually see the loading state. However, I'm sure tools along the lines of `authbind` that LD_PRELOAD `bind` can have a big enough overhead to matter. And somewhere there's probably a firewall/AV that will prompt the user when a process wants to bind to a port for the first time. And during this time we now have a loading state :tada: . Sorry for being so pedantic :sweat_smile:

Prinzhorn (approved):

PR #5556, opened 2022-08-26
Unify proxy modes, introduce UDP protocol detection.

meitinger:
#### Description

Contains changes for #5535.

1. Removed transparent mode in UDP. The way transparent mode for UDP is implemented is far from ideal: It's only supported on Linux, requires sudo for the whole process, prevents full coverage of `udp.py` and is not even used/enabled.
    Together with #5543, I'll plan to do re-add transparent UDP mode later in a similar fashion, i.e. a separate process, asking for sudo if necessary, setting everything up for the user. In the meantime, `udp.py` has now full coverage.
2. Updated `AsyncioServerInstance` and `ProxyMode` to support both protocols for one `ProxyMode`. In the process, the need for additional `ServerInstances` disappeared, so `ProxyMode`s now fully describe proto, port and layer.
3. Removed `make_connection_id`. Since we don't know on what layer we'll detect what protocol, we can't use the DNS-ID in `ServerManager` connection-ID trick anymore - same as goes for QUIC, if we don't want to alter `SendData` and `DataReceived` (see https://github.com/mitmproxy/mitmproxy/pull/5435#pullrequestreview-1075694658).
4. Changed the proxy mode syntax to what was discussed in #5535. Also DNS has now `--mode reverse:dns://8.8.8.8` if used with an upstream, `--mode dns` will launch a local-resolve server.
5. Adjusted DNS layer to match packets by their ID. As mentioned above, this was previously done through `ServerManager` connection-IDs, but if we detect DNS higher up, or in the future use some transparent mode, this can't be done this way anymore.
6. Copied and adjusted TCP layer decisions for UDP.


meitinger:
@mhils: Before I'm adjusting and (re-)adding tests, could you maybe have a quick look as to whether the way the "new" modes are implemented is something you'd sign off on? Thanks :)

mitmproxy/proxy/layers/dns.py
```diff
@@ -40,40 +39,37 @@ class DNSLayer(layer.Layer):
     Layer that handles resolving DNS queries.
     """
 
-    flow: dns.DNSFlow
+    flows: dict[int, dns.DNSFlow]
```
mhils, line 42: How/why do we end up having multiple DNS flows here now?

mitmproxy/proxy/mode.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def get_state(self):
+        return self.full_spec
+
+    def set_state(self, state):
+        if state != self.full_spec:
+            raise RuntimeError("Proxy modes are frozen.")
+
+
+class AsyncioProxyMode(ProxyMode, metaclass=ABCMeta):
```
mhils, line 162: I feel this complicates things more than it helps. Maybe we can move `description` into ProxyMode itself and then keep the layer information as part of individual server instance subclasses? That avoids the special-casing in `ServerInstance.make`, keeps all modes as immediate subclasses of ProxyMode,  retains a dumb 1-1 mapping between modes and instances, and avoids any circular spec -> layer -> spec dependencies. This may entirely be "code I wrote myself is easier to understand" bias, so feel free to go ahead with this approach if you feel it's more straightforward. Just a subjective impression. :)

mhils (commented):
Overall this looks great, mostly dumb clarification questions below! Please don't read any judgement/sentiment into these comments, reviewing on mobile is tricky and they might be stupid questions. 😅

mitmproxy/proxy/mode.py
```diff
@@ -0,0 +1,173 @@
+"""
```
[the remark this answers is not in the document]
meitinger: Python and it's cyclic dependencies 😅
But I'll restore the 1:1 `ServerInstance`s, which will move the `layers`-dependency to `mode_servers.py` again, so I can move `mode.py` back into `mode_specs.py`.

mitmproxy/proxy/mode.py
meitinger: I guess it's just my code-compressing attitude, but you're right. Having 1:1 `ServerInstance`s removes all the `if TYPE_CHECKING` stuff, so I'll undo the changes.

mitmproxy/proxy/layers/dns.py
meitinger: So the history here is: We had that before, in the DNS draft. Then we moved the ID into the connection_id of proxyserver (i.e. essentially the `dict`). However, if we want to do protocol detection and fancy transparent stuff, DNS might not be at the top layer, so the layer itself will have to do it.

It's the same with QUIC. We've talked about connection migration (which currently also uses connection_id in proxyserver) and the change of `SendData` and `DataReceived` to include the remote address, that ideally should not happen. So there we will also need a layer that does the routing instead, keeping the old `SendData` and not using custom `connection_id`.

And that's actually a good thing: Having uniform connection id's is the only way to reliably do `inject_udp` in the future.

mhils:
Looks fantastic now! 🍰 😃

meitinger:
@mhils This should be ready now :) I've updated the docs as well, that's why I didn't want to merge before you've had the chance to look over the wording. Once we support QUIC and HTTP3, I think it's time to add a table as in #5535.

mhils (approved):
Very nice! 🍰

docs/src/content/concepts-modes.md
```diff
@@ -172,6 +177,19 @@ There are various use-cases:
...
     server. As you see, it's a fairly flexible thing.
+- Want to know what goes on over (D)TLS (without HTTP)? With mitmproxy's raw
+    traffic support you can. Use `--mode reverse:tls://example.com:1234` to
+    spawn a TCP instance that connects to `example.com:1234` using TLS, and
+    `--mode reverse:dtls://example.com:1234` to use UDP and DTLS respectively instead.  
+    Incoming client connections can either use (D)TLS themselves or raw TCP/UDP.
+    In case you want to inspect raw traffic only for some hosts and HTTP for
+    others, have a look at the [tcp_hosts]({{< relref "concepts-options" >}}#tcp_hosts)
+    and [udp_hosts]({{< relref "concepts-options" >}}#udp_hosts) options.
+- Say you want to capture DNS traffic to Google's Public DNS server? Then you
```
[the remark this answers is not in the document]
meitinger: You're right, here's the PR: #5563 :)

PR #5562, opened 2022-08-29
add transparent server mode based on WireGuard

decathorpe:
This is a draft for a transparent mode implementation based on WireGuard. I'm filing this "early" even though some things aren't finished yet, but I'm looking for feedback on whether things are generally looking "OK" or if some things should be done differently.

**New mode spec for a WireGuard mode**:

- simple and extensible key-value based settings syntax (in case more settings need to be specified in the future)
- currently supported settings: `listen_port` to override the default WireGuard port (51820), `name` to override the default filename prefix for generated WireGuard configuration files (`mitmproxy_wireguard`), and `peers` to override the default number of peers (1) for which a configuration will be generated
- unit tests to verify that the new mode spec is parsed correctly

**New WireGuard server mode implementation**:

- new implementation of `WireGuardConnectionHandler` (based on `ProxyConnectionHandler` but using `mitmproxy_wireguard.TcpStream` instead of `asyncio.StreamReader/StreamWriter)
- new implementation of `WireGuardInstance`: based on `TcpServerInstance` / `TransparentInstance` for TCP, and `UdpServerInstance` for UDP (not hooked up yet, better support for UDP connections needs to be implemented in `mitmproxy_wireguard` [first](https://github.com/decathorpe/mitmproxy_wireguard/issues/7))
- small adaptations for accepted types of `reader` / `writer` in `ConnectionIO` (might need to be refactored to be generic, and / or require further adaptations once `mitmproxy_wireguard.UdpStream` is ready for handling UDP packets)
- not done yet: unit tests (maybe something similar to the `test_transparent` test in `tests/mitmproxy/proxy/test_mode_servers.py`?)

**What's not working yet**:

- `TcpStream.get_extra_info("original_addr")` is [not implemented yet](https://github.com/decathorpe/mitmproxy_wireguard/issues/6), but exposing the original destination in this manner should be relatively straightforward in `mitmproxy_wireguard`.
- Running `mitmdump --mode wireguard:[spec]` correctly generates WireGuard configuration files with the given settings, but then crashes with `invalid IP address syntax` and I cannot find where this error message is coming from (grepping mitmproxy source code doesn't yield any hits for this string and relevant substrings of it).


decathorpe:
Re: CI failures:

- No idea why tests fail on macos-latest 3.10 but pass on macos-11, looks like the former can't find a Rust toolchain, but the latter can? Weird ...
- I have never used mypy for incremental type checking in Python, I'll need help figuring out what it doesn't like here.
- Coverage: Yeah this is kind of obvious, not everything new is covered by unit tests yet.

mitmproxy/proxy/mode_servers.py
```diff
@@ -333,6 +387,86 @@ async def handle_udp_connection(self, connection_id: tuple, handler: ProxyConnec
...
+        # stream.feed_data(data, remote_addr)
+
+    async def handle_udp_connection(self, connection_id: tuple, handler: WireGuardConnectionHandler) -> None:
+        with self.manager.register_connection(connection_id, handler):
+            await handler.handle_client()
+
+    @property
+    def listen_addrs(self) -> tuple[Address, ...]:
+        # the WireGuard server listens on exactly one socket
+        return self._server.getsockname(),
```
mhils, line 467: How is IPv4/IPv6 handled right now when you listen on `""`?

mitmproxy/proxy/mode_specs.py
```diff
@@ -209,6 +209,52 @@ def __post_init__(self) -> None:
...
+    default_port = 51820
+    transport_protocol: ClassVar[Literal["tcp", "udp"]] = "udp"
+
+    wireguard_conf: str | None = None
+    wireguard_peers: int | None = None
+
+    # noinspection PyDataclass
+    def __post_init__(self) -> None:
+        # settings are empty or a literal "default":
+        if self.data in ["", "default"]:
```
mhils, line 222: Is there a specific reason for adding an explicit `"default"`? I'd just skip that for brevity.

mitmproxy/proxy/mode_specs.py
```diff
@@ -209,6 +209,52 @@ def __post_init__(self) -> None:
...
+            if not kv:
+                # empty string after a trailing comma
+                break
+
+            try:
+                key, value = kv.split("=", 1)
+            except ValueError:
+                raise ValueError(f"invalid wireguard mode (not a key=value pair: {kv!r})")
+
+            if key not in ["name", "listen_port", "peers"]:
```
mhils, line 239: What is name here? The configuration file path?

mhils (commented):
Nice, very excited for this to land! 😃 

- there's a bit of overlap with https://github.com/mitmproxy/mitmproxy/pull/5556, which I'd like to get in first (@meitinger already had the pleasure of fixing merge conflicts with my stuff).  That shouldn't be a substantial obstacle though. :)
- if possible I'd like to reuse the ProxyConnectionHandler instead of subclassing it. That should become more straightforward with @meitinger's PR, which changes the location of transparent mode address resolution
- The macOS failure we should tackle at its root, i.e. provide matching binary wheels. Building additional Rust code in CI takes too much time.

decathorpe:
> Nice, very excited for this to land!
> 
>     * there's a bit of overlap with #5556, which I'd like to get in first (@meitinger already had the pleasure of fixing merge conflicts with my stuff).  That shouldn't be a substantial obstacle though. :)

I figured as much. I still wanted to file this PR to get some initial feedback rather than waiting for more stuff to land.

>     * if possible I'd like to reuse the ProxyConnectionHandler instead of subclassing it. That should become more straightforward with @meitinger's PR, which changes the location of transparent mode address resolution

Ok, sounds good to me. Hope resolving those merge conflicts won't be too bad.

>     * The macOS failure we should tackle at its root, i.e. provide matching binary wheels. Building additional Rust code in CI takes too much time.

Yeah, I'll try to figure out how to use Github Actions to publish binary wheels to PyPI ... looks like there's something I can use here: https://github.com/messense/maturin-action

mitmproxy/proxy/mode_specs.py
```diff
@@ -209,6 +209,52 @@ def __post_init__(self) -> None:
...
+                raise ValueError(f"invalid wireguard mode (unrecognized option: {key!r}")
+
+            if key == "name":
+                self.wireguard_conf = value
+
+            if key == "listen_port":
```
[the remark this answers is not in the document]
decathorpe: Ah, I see now that generic parsing of the mode spec happens in `ProxyMode.parse` ... This might conflict with the custom format that's used by WireGuardMode right now - might explain why apparently mitmproxy gets fed an invalid IP address when you try to run it :)

Looks like I shouldn't use `:` as separator character in `wireguard:custom=foo,settings=bar,` after all, because this breaks things elsewhere - any suggestions for what I could use instead?

mitmproxy/proxy/mode_servers.py
decathorpe: The WireGuard server listens on on only one socket, so it's either IPv4 *or* IPv6, but never both.

I'm not sure if an empty string is even parsed into a `std::net::SocketAddr` successfully or not (but I'd guess not).

mitmproxy/proxy/mode_specs.py
decathorpe: Right. For some reason I thought that there might need to be a distinction between the two, but right now, this doesn't make sense. I'll remove the "default" option.

mitmproxy/proxy/mode_specs.py
decathorpe: It's the base name / prefix for the names of the configuration files that will be generated. 

For example, if `name = "foo", peers = 2`, then the configuration files will be named `foo.conf`, `foo_peer0.conf`, and `foo_peer1.conf`.

I'm sometimes bad at naming things, so I'm open to suggestions for a better name for this option.

mitmproxy/proxy/mode_specs.py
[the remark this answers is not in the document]
meitinger: `:` should be fine, but if `@` appears anywhere in your settings, the string after `@` will be interpreted as `[host:]port`, which will fail.

@mhils: Maybe we should use regex to ensure `@` only matches if it's followed by an IP and port? (Of course for something like `wireguard:conf_prefix=@7` this would still be a problem, but only people named [**Robert'); DROP TABLE Student;--**](https://xkcd.com/327/) pick a prefix like that ;)

meitinger (commented):
Very well integrated!
Some changes in #5556 might help to simplify some parts even further. (But I think @mhils already mentioned it.)

mitmproxy/proxy/mode_specs.py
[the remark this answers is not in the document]
mhils: > Looks like I shouldn't use `:` as separator character

I think you need to, that's what ProxyMode.parse expects.

> @mhils: Maybe we should use regex to ensure @ only matches if it's followed by an IP and port?

Let's just not support `@` in specs for now, I don't think it's much of a restriction. :)

mitmproxy/proxy/mode_servers.py
mhils: `""` is our default listen_addr, so it would be nice if it works. 😅 
Anyhow - no need to do anything here for now, I was just wondering what happens.

mitmproxy/proxy/mode_specs.py
mhils: From a practical perspective, maybe we don't even need the `peers` parameter? If you want multiple peers, you could just start additional instances on different ports and that would work just as well? Then specs could literally just be `wireguard:path/to/wireguard.conf` or simply `wireguard` (for `~/.mitmproxy/wireguard.conf`), avoiding any key/value parsing. :)

mhils:
> Yeah, I'll try to figure out how to use Github Actions to publish binary wheels to PyPI ... looks like there's something I can use here: https://github.com/messense/maturin-action

Cross compilation is a can of worms. Let's just spawn `macos-latest` and `windows-latest` jobs that build the additional wheels? :)

mitmproxy/proxy/mode_specs.py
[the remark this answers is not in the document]
decathorpe: Yeah, `@` characters aren't expected in any of the settings that the wireguard mode will accept.

Still doesn't help that I have no idea where the `invalid IP address syntax` error comes from then :/

mitmproxy/proxy/mode_servers.py
decathorpe: Right now the `listen_addr` is passed unmodified to `str::parse::<std::net::SocketAddr>()` here:
https://github.com/decathorpe/mitmproxy_wireguard/blob/0.1.0-alpha.3/src/lib.rs#L128-L131

I just checked, and an empty string is not an accepted host address for `SocketAddr`. What does the empty "" address actually do? If it just binds to `0.0.0.0` or `127.0.0.1`, I can use that as the default (or even make that configurable).

mitmproxy/proxy/mode_specs.py
decathorpe: Would that launch multiple processes? I'm not sure if `boringtun` or `smoltcp` have any global variables that might interfere with running multiple instances in one process.

And you'd still need to generate *at least one* configuration for your peer(s) in some easily accessible location, because otherwise you can't configure your client device. Just writing it to `$(cwd)` seemed to be the simplest option ...

mitmproxy/proxy/mode_servers.py
mhils: `asyncio.create_server("", 8080, ...)` binds two sockets: One on `0.0.0.0`, and one on `::` (assuming both are available - otherwise it's just one). We should ultimately do the same for Rust so that things are consistent, but just defaulting to `0.0.0.0`  to get things running is fine with me

mitmproxy/proxy/mode_specs.py
mhils: No, that's all in the same process. I'd expect them to be free of global state though.

Yes, I think the mechanism here should just be to first determine the filename (a custom passed path or - if none is passed - the default location in ~/. mitmproxy), and then create such a config file with exactly one peer if it does not exist yet. The configuration should then be included in ServerInstance.to_json() so that it can be displayed in mitmweb. :)

decathorpe: That won't work though. If we keep only the server configuration, then there's no way to recover the private key for client configurations.

decathorpe:
PS: I managed to build binary wheels for Windows (x64, x86), mac OS (x64, aarch64), and linux (x86_64, i686) by using a GitHub Action: https://github.com/decathorpe/mitmproxy_wireguard/actions/runs/2958218239

Once I figure out how to upload those files to PyPI, that should make things considerably easier. Are there any targets that are important for mitmproxy which are missing from this list?

mhils:
This job here is pretty much best practice: https://github.com/mitmproxy/pdoc/blob/021cbec1424f91caf9dbd35105a7d8a6fbda8f8c/.github/workflows/main.yml#L63

Instead of the `python setup.py` bits you want to download the artifacts of course. Speaking of which, I think you need different names per OS. See https://github.com/mitmproxy/mitmproxy/blob/1706a9b9fe8c8157824df22634d97c3d695a495d/.github/workflows/main.yml#L139.

mitmproxy/proxy/mode_specs.py
mhils: One way around this would be to write out a `.key` file next to the `.conf` file. Lack of such a file would not be a critical failure but only stops us from displaying full configuration details in mitmproxy.

decathorpe:
> Targets are good! Not sure if we even want or need Windows x86.
> 
> This job here is pretty much best practice: https://github.com/mitmproxy/pdoc/blob/021cbec1424f91caf9dbd35105a7d8a6fbda8f8c/.github/workflows/main.yml#L63

Thanks for the link, looks like exactly what I needed.

> Instead of the `python setup.py` bits you want to download the artifacts of course. Speaking of which, I think you need different names per OS. See
> 
> https://github.com/mitmproxy/mitmproxy/blob/1706a9b9fe8c8157824df22634d97c3d695a495d/.github/workflows/main.yml#L139
> .

Looks like this works fine as-is. When downloading the artifact, I got a ZIP archive that's called "wheels" that contained all the files from the different `upload-artifact` actions, which I then uploaded with `twine`, which worked perfectly:

https://pypi.org/project/mitmproxy-wireguard/#files

---

Since that's now squared away, I'll try to actually rebase this PR tomorrow, so we can see whether the binary wheels work for the CI here.

mhils:
> Looks like this works fine as-is. When downloading the artifact, I got a ZIP archive that's called "wheels" that contained all the files from the different `upload-artifact` actions, which I then uploaded with `twine`, which worked perfectly:

Awesome! This used to not work, great to see that they apparently fixed it. 😃

mitmproxy/proxy/mode_specs.py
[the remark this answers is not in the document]
decathorpe: I simplified the mode spec a bit based on feedback here, and by adding better test coverage, I found the error (it's not in my code, but in ProxyMode.parse. For a mode like `wireguard@51821:generate`, it will attempt to parse `51821` as a host address and `generate` as a port number.

decathorpe:
I pushed a rebased version of the "add a wireguard mode spec" commit, and tried to address feedback wrt/ how to locate configuration files. Instead of specifying a "name" (which seemed to be both ambiguous to the user and not clear how to implement for us), the mode spec can optionally specify the *directory* from which configuration can be loaded from if it's already present, or in which configuration files will be written if they don't exist yet (or if the `generate` option was specified). The default directory, if it's not specified, will probably be `~/.mitmproxy`, and the file name is always `mitmproxy_wireguard[_peerN].conf`, to simplify things.

So, for example, when using `--mode wireguard:generate=./test,peers=2`, configuration files would be written to

- `./test/mitmproxy_wireguard.conf`
- `./test/mitmproxy_wireguard_peer0.conf`
- `./test/mitmproxy_wireguard_peer1.conf`

I think this should address the problem of *"where to place files by default*" and "*how to expose this setting to the user*". What do you think?

mitmproxy/proxy/mode_specs.py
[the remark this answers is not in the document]
mhils: The spec is invalid, it should be `wireguard:generate@51821` in your example. The `@listen_port` part always comes last.

mhils (commented):
Thanks for the update! Maybe I have missed this, but have we found a definite use case for a `peers` argument? Do we need to make a distinction betwen load and generate, or could we just do generate-if-not-exists? I'd be happy to keep things as simple as possible for now.

decathorpe:
> Thanks for the update! Maybe I have missed this, but have we found a definite use case for a `peers` argument?

If the `peers` argument is not specified, the default is `1`. If a user later wants to connect more clients than the number of configured peers, they need to regenerate the configurations, invalidating previously generated files and encryption keys.

Keeping valid configurations for existing peers around when adding configuration for more peers would mean I need to write parsing code for WireGuard client configuration files in addition to the already existing server configuration file parser and/or make the code for configuration file handling in`mitmproxy_wireguard` a whole lot more complicated. Allowing the user to specify the expected number of peers *in advance* and overriding existing files when generating configurations for *more peers* looks like the simplest solution to me.

> Do we need to make a distinction between load and generate, or could we just do generate-if-not-exists? I'd be happy to keep things as simple as possible for now.

Yeah, keeping it simple was my intention. Right now, there are three cases:

1. neither `load` nor `generate` are specified: Attempt to load WireGuard server configuration from default location. If no configuration file is present, generate configuration files with default settings.
2. `load` is specified: Attempt to load WireGuard server configuration. If no configuration file is present, fail (since "load" was specified explicitly).
3. `generate` is specified: Generate new WireGuard configuration files, and override files that potentially already exist (since "generate" was specified explicitly).

These three cases should cover a) a sane default (load or generate with defaults), b) an explicit "use this configuration file" case, and c) an explicit "generate new configuration files even if they already exist" case. I think this is a good balance between simplicity and UX.

mitmproxy/proxy/mode_specs.py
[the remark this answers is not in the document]
decathorpe: Oh, that finally makes sense. I had misinterpreted the format.

decathorpe:
Second draft of WireGuard server mode:

- `mitmproxy / mitmdump / mitmweb --mode wireguard` all run without errors
- binary wheels of `mitmproxy_wireguard` were used on all targets :tada:
- good test coverage for WireGuard mode spec
- bad test coverage for WireGuard server mode (not sure if this can even be tested in CI?)
- BUT: it doesn't seem to work yet, there doesn't seem to be any incoming data. not sure why.

decathorpe:
PS: It looks like mitmproxy does not use the "logging" module from the python standard library? mitmproxy_wireguard is currently explicitly compatible with and reliant on Python's "logging" module (with `pyo3-logging`), but if that's not going to work, would it be OK to set up logging on the Rust side instead (i.e. at server initialization)?

mhils:
> PS: It looks like mitmproxy does not use the "logging" module from the python standard library? mitmproxy_wireguard is currently explicitly compatible with and reliant on Python's "logging" module (with `pyo3-logging`), but if that's not going to work, would it be OK to set up logging on the Rust side instead (i.e. at server initialization)?

Yeah, we need to fix interop here. I think the correct way is to actually deprecate our custom logging shenanigans and switch to the builtin logging module. Let's keep using `pyo3-logging` here.

mhils (commented):
The custom DatagramTransport looks good! The lack of incoming data is a WireGuard handshake problem. If I hardcode the configuration:
```python
self._wireguard_conf = wg.WireGuardServerConf.custom(51820, "EG47ZWjYjr+Y97TQ1A7sVl7Xn3mMWDnvjU/VxU769ls=", ["Test1sbpTFmJULgSlJ5hJ1RdzsXWrl3Mg7k9UTN//jE="])
```

I get tons of output. I haven't tracked down what the actual issue is, but I suspect we should vastly simplify configuration parsing. 

- Do we really need to create or parse WireGuard ini files? I don't think we need to. It's perfectly fine for us to have our own custom file format which just contains two private keys as JSON:
   `~/.mitmproxy/wireguard.conf`
   ```json 
   {
        "mitmproxy_key": "...",
        "client_key": "...",
   }
   ```
   We only support one peer until we find a reason later on why we would need more peers on the same port.
 - When `wireguard:~/.mitmproxy/wireguard.conf` is passed to mitmproxy, we do the following: 
   1) Check if `~/.mitmproxy/wireguard.conf` exists, and create one if it doesn't.
       (mitm-wg only provides a `genkey()`  method for this)
   2) Read keys from `foo.conf`, fail if file format is invalid.
 - In `WireGuardServerInstance.start`, we 
   1) Invoke `WireGuardServerConf.custom`  with the keys we have (using mitm-wg's `pubkey(key)` to get public keys). This basically becomes our only interface - we don't even need a WireGuardServerConf object really, we can just pass keys directly to `start_server`.
   2) Log a WireGuard client configuration file to mitmproxy's event log [like we do here](https://github.com/decathorpe/mitmproxy_wireguard/blob/eb6f47fcb8b1f7489b033346ee50ca789b3b860c/echo_test_server.py#L48-L59)[^1]. This means that `mitmdump --mode wireguard` will basically print the configuration to stdout on startup.

This means we don't need to maintain the ~750 lines of conf parsing that are currently in mitm-wg. Does that make sense? Or is that too simple and I'm overlooking something?

[^1]: We later add `WireGuardServerInstance.client_config()` which returns the config as a string. Then we can reuse the same code for `WireGuardServerInstance.to_json()`, which adds the config so that it's available in mitmweb.

mhils:
> * (not sure if this can even be tested in CI?)

This is a very good point. If you look at `boringtun` you find that tests are absent there as well. 🙃  I think we have two approaches:

1. YOLO
2. Write a small standalone (Rust) binary that takes client_key, server_pubkey, listen_port and then performs TCP and UDP echo roundtrips using tokio/boringtun. In the Python test suite we spawn a server and then just shell out to that.

Either is fine with me. The standalone thingy doesn't sound _too bad_ though. :)

decathorpe:
> The custom DatagramTransport looks good! The lack of incoming data is a WireGuard handshake problem. If I hardcode the configuration:
> 
> ```python
> self._wireguard_conf = wg.WireGuardServerConf.custom(51820, "EG47ZWjYjr+Y97TQ1A7sVl7Xn3mMWDnvjU/VxU769ls=", ["Test1sbpTFmJULgSlJ5hJ1RdzsXWrl3Mg7k9UTN//jE="])
> ```
> 
> I get tons of output. I haven't tracked down what the actual issue is, but I suspect we should vastly simplify configuration parsing.

Oh oh ... ok, that explains why "nothing is happening". I'll look into it.

>     * Do we really need to create or parse WireGuard ini files? I don't think we need to. It's perfectly fine for us to have our own custom file format which just contains two private keys as JSON:
>       `~/.mitmproxy/wireguard.conf`
>       ```json
>       {
>            "mitmproxy_key": "...",
>            "client_key": "...",
>       }
>       ```

That *would* be simpler, yes ... *but* (cont'd below)

>       We only support one peer until we find a reason later on why we would need more peers on the same port.

Supporting variable number of peers is easy, not sure it's worth ripping it out again. The default is already *one peer*, so unless somebody would explicitly do something different, this is already what happens.

>     * When `wireguard:~/.mitmproxy/wireguard.conf` is passed to mitmproxy, we do the following:
>       
>       1. Check if `~/.mitmproxy/wireguard.conf` exists, and create one if it doesn't.
>          (mitm-wg only provides a `genkey()`  method for this)
>       2. Read keys from `foo.conf`, fail if file format is invalid.
> 
>     * In `WireGuardServerInstance.start`, we
>       
>       1. Invoke `WireGuardServerConf.custom`  with the keys we have (using mitm-wg's `pubkey(key)` to get public keys). This basically becomes our only interface - we don't even need a WireGuardServerConf object really, we can just pass keys directly to `start_server`.
>       2. Log a WireGuard client configuration file to mitmproxy's event log [like we do here](https://github.com/decathorpe/mitmproxy_wireguard/blob/eb6f47fcb8b1f7489b033346ee50ca789b3b860c/echo_test_server.py#L48-L59)[1](#user-content-fn-1-151b76ec0aa8cb162ee3c47d9763abfc). This means that `mitmdump --mode wireguard` will basically print the configuration to stdout on startup.
>
> This means we don't need to maintain the ~750 lines of conf parsing that are currently in mitm-wg. Does that make sense? Or is that too simple and I'm overlooking something?

My intention with the configuration parser / writer code was to be compatible with official / other WireGuard servers and clients, which all (?) understand the same file format. For example, it should be possible to just drop the generated client configuration file into `/etc/wireguard/` and use it, or import the peer configuration file into the official WireGuard Android client.

If we don't provide a config file, the user would have to either put together a config file themselves, or copy public / private keys and settings into their WireGuard client app to manually configure the tunnel. I thought that providing a ready-to-use / importable config file would be much nicer - and since the biggest benefit of using the WireGuard mode over the transparent mode would be better UX, I thought that this would be important.

> ## Footnotes
> 
>     1. We later add `WireGuardServerInstance.client_config()` which returns the config as a string. Then we can reuse the same code for `WireGuardServerInstance.to_json()`, which adds the config so that it's available in mitmweb. [leftwards_arrow_with_hook](#user-content-fnref-1-151b76ec0aa8cb162ee3c47d9763abfc)

Yeah, extending the `to_json` method to show WireGuard configuration was on my TODO list.

mhils:
> My intention with the configuration parser / writer code was to be compatible with official / other WireGuard servers and clients, which all (?) understand the same file format. For example, it should be possible to just drop the generated client configuration file into `/etc/wireguard/` and use it, or import the peer configuration file into the official WireGuard Android client.

To be clear, I want mitmdump to print out a valid configuration file on startup which can then be pasted into `/etc/wireguard/` or the client. I just don't think we want to write this to disk and maintain a a full parser/serializer.

> Supporting variable number of peers is easy, not sure it's worth ripping it out again. The default is already _one peer_, so unless somebody would explicitly do something different, this is already what happens.

Supporting more than one peer means:

 1) a more complex spec format (key-value vs a single path)
 2) extra UI work to nicely display a variable number of client configurations.
 3) extra handling of mismatches between the specified number of peers and an existing configuration file, etc.

I don'd mind keeping support for multiple peers in the Rust bits, but outside of that it's still quite a bit of unnecessary complexity.

> If we don't provide a config file, the user would have to either put together a config file themselves, or copy public / private keys and settings into their WireGuard client app to manually configure the tunnel. I thought that providing a ready-to-use / importable config file would be much nicer - and since the biggest benefit of using the WireGuard mode over the transparent mode would be better UX, I thought that this would be important.

Here's the kind of UX I'm envisioning:
```
$ mitmdump --mode wireguard
WireGuard server listening at *:51820. 
Client configuration:
───────────────────────────────────────────────────────────────────────────
[Interface]
PrivateKey = qG8b7LI/s+ezngWpXqj5A7Nj988hbGL+eQ8ePki0iHk=
Address = 10.0.0.1/32

[Peer]
PublicKey = mitmV5Wo7pRJrHNAKhZEI0nzqqeO8u4fXG+zUbZEXA0=
AllowedIPs = 0.0.0.0/0
Endpoint = 192.168.0.4:51820
───────────────────────────────────────────────────────────────────────────
```

The benefit of printing this to stdout is that the configuration is right in your face and you don't need to go search for a file. For other devices, the configuration file needs to adjust [the local IP address](https://github.com/mitmproxy/mitmproxy/blob/891dae031c65a3268a18056bff9719ac9ee2a66f/mitmproxy/platform/windows.py#L252-L275) of the endpoint in the configuration, so we want that to be dynamic anyways (and not just written once to a file).

Taking things one step further, it would be awesome to display a QR code in mitmweb which can then just be scanned on Android. Does that make more sense now?

> Writing a standalone test client in Rust shouldn't be hard, but how would you expect to ship it / install it in CI?

Compile it once for Linux and include the compiled binary in the repo. Skip the tests on other OSes.

> With the latest code, I can use the official WireGuard Android VPN to connect to mitmproxy, view `mitm.it` successfully, download a certificate, install it, and see instagram trying to ping facebook servers with tracking data in mitmweb 😉

🎉 🎉 🎉

Prinzhorn:
I really just skimmed over the discussion, I'm very excited for WireGuard support!

> Taking things one step further, it would be awesome to display a QR code in mitmweb which can then just be scanned on Android. Does that make more sense now?

Or in mitmproxy as well? :eyes: https://pythonhosted.org/PyQRCode/rendering.html#terminal-rendering (was the first thing I found)

decathorpe:
> I just don't think we want to write this to disk and maintain a full parser/serializer.

Aaaaaah ... I get it now.

When writing the original implementation, my first thought was to re-use the existing / standard WireGuard configuration file format instead of reinventing the wheel. But we can use a *simpler* serialization format (i.e. JSON), since we don't actually need to *store files on disk* that are compatible with other WireGuard tools, and only need to display / print client configuration as necessary.

I'll need to think about what to rip out of the current implementation and how to implement the new approach ...

decathorpe:
I think this is ready for review now. Configuration handling is much simpler now.

The `Configuration.pretty_print()` method can generate configuration file(s) for WireGuard client(s) if supplied with the appropriate "endpoint" (the public network address of the machine that's running mitmproxy), the "address" (list of IP addresses to bind to the WireGuard interface), and the "allowed_ips" (the list of IP ranges for which traffic will be sent over WireGuard; will likely be "0.0.0.0/0" in most cases).

I also tested generating QR codes from this pretty-printed configuration file (with something like `qrencode -r test.conf -o out.png -s 4`) and was able to imported it successfully into the official WireGuard Android client and connect to mitmproxy this way.

decathorpe:
Note: I don't think adding a test client application is worthwhile. It would almost double the amount of code in `mitmproxy_wireguard`, because the test client application would need to do 1) client-side of the WireGuard protocol, 2) user-space network stack for assembling TCP/IP and UDP/IP packets, handling TCP connections, etc. - so basically the same thing we already have for the server-side (which is already quite a lot of code).

mhils:
> Configuration handling is much simpler now.

Much better, but not quite there yet. I've taken the liberty to remove the load/gen/keyval parsing as suggested above. I'll open a PR with the corresponding Rust changes momentarily. Let me know if I missed something. :-)

mitmproxy/proxy/mode_servers.py
```diff
@@ -12,6 +12,10 @@
 from __future__ import annotations
 
 import asyncio
+import json
+import textwrap
+from pathlib import Path
+
 import errno
```
decathorpe, line 19: Looks like the new imports are not sorted correctly?

mhils: We currently don't insist on a particular import style or order, we only use flake8 to flag the most egregious style violations. The reason for this is that it doesn't provide much value in itself while often getting in the way of merging PRs. It's not a very common thing for Python projects (Rust is very different here), so new contributors are unlikely to sort imports right away.

decathorpe:
I merged your mitmproxy_wireguard PR and pushed some follow-up fixup commits and released v0.1.0 to PyPI so it is usable for this PR: https://pypi.org/project/mitmproxy-wireguard/0.1.0/

mhils:
Excellent, thanks!

I've pushed a minimal test client covering UDP and TCP here: https://github.com/mhils/mitmproxy_wireguard/tree/test-client. 

Running `echo_test_server.py` in a separate window, I get the following output:

```
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
     Running `target/debug/wg-test-client`
Sending packet: 450000218d6600008011307f0a0000010a00002a04d27a69000d253f68656c6c6f
Sending packet: 45000034d97b40008006ffff0a0000010a00002acafe005012345678000000008002faf0ffff0000020405b40103030801010402
Received packet: 4500003400004000400600000a00002a0a0000010050cafefdecf8a2123456798012ffff00000000020405640303010402000000
It's a TCP SYN/ACK.
Sending packet: 45000034d97b40008006ffff0a0000010a00002acafe005012345679fdecf8a350100204ffff000068656c6c6f20776f726c6421
Received packet: 4500002100004000ff1167a10a00002a0a0000017a6904d2000d88dc48454c4c4f
It's the UDP reply we were looking for.
Received packet: 4500002800004000400600000a00002a0a0000010050cafefdecf8a31234568550107ffa00000000
It's a TCP ACK with no data.
Received packet: 4500002800004000400600000a00002a0a0000010050cafefdecf8a3123456855010800000000000
It's a TCP ACK with no data.
Received packet: 4500003400004000400600000a00002a0a0000010050cafefdecf8a312345685501880000000000048454c4c4f20574f524c4421
It's a TCP ACK with the data we were looking for.
All set!
```

Pretty much everything is hardcoded and as simple as possible, but it's good enough and works for a minimal test case. I'd propose we just put the compiled linux binary into the repo here and then use that to test WireGuardServerInstance.

Feel free to integrate the test binary into the main mitmproxy_wireguard repo, I don't mind either way. One conflict at the moment is that test-client is already depending on boringtun 0.5, which uses some other key primitives.

decathorpe:
> Feel free to integrate the test binary into the main mitmproxy_wireguard repo, I don't mind either way. One conflict at the moment is that test-client is already depending on boringtun 0.5, which uses some other key primitives.

Done!
https://github.com/decathorpe/mitmproxy_wireguard/commit/d630007

Though due to a limitation of maturin, the binary can't be included in built binary wheels:
https://github.com/PyO3/maturin/issues/368

decathorpe:
I released v0.1.1 to PyPI which has the requested changes. The test binary can be built with "cargo build --release" into `target/release/mitm-wg-test-client` and included here manually, if needed.

decathorpe:
Note that changes that were not made by me seem to introduce two bugs:

1. `mitmweb --mode wireguard` crashed because `WireGuardServerInstance.to_json()` is called before `WireGuardServerInstance.start()`, which means the `client_key` and `server_key` attributes were not set yet. I moved the configuration handling into `__init__`, which means `client_key` and `server_key` are always set early enough, so this should be fixed now.

2. There's a lot of "client connect", "client disconnect", and "server disconnect" events logged to the console when running `mitmweb --mode wireguard`, but almost no flows are shown in the web interface. This might be a regression from porting to boringtun v0.5? There's many errors like `error establishing server connection: [Errno 113] Connect call failed ('192.168.86.53', 8009)`.

There's also one bug in the UDP handling code that I couldn't figure out how it happens:

```

  File "/usr/lib64/python3.10/asyncio/events.py", line 80, in _run

  File "/home/deca/Workspace/mitmproxy/mitmproxy/proxy/mode_servers.py", line 386, in handle_udp_datagram
    writer = udp.DatagramWriter(transport, remote_addr, reader)

  File "/home/deca/Workspace/mitmproxy/mitmproxy/net/udp.py", line 204, in __init__
    assert isinstance(proto, DrainableDatagramProtocol)

AssertionError

Please lodge a bug report at:
	https://github.com/mitmproxy/mitmproxy/issues
```

I have no idea how any `DatagramTransport` can ever be initialized if not with a `DrainableDatagramProtocol`, but it sometimes seems to happen, and it leads to crashes in `udp.DatagramWriiter.__init__()`.

decathorpe:
> I think the problem here is that you have WireGuard running on the same machine and end up in an infinite loop? This actually is a problem that's really quite tricky for us to detect. Not sure how we can handle it best (or if we need to handle it).

No, this was with `mitmweb --mode wireguard` running on my PC and connecting to it with the Android WireGuard app from my phone.

decathorpe:
Looks like your two most recent commits introduced two new test failures, though I'm not sure how that happened. Looks like tests wait for certain log messages to be written, but they no longer are (or different log messages are written first):

```
_____________________________________________ test_tcp_start_stop ______________________________________________

    async def test_tcp_start_stop():
        manager = MagicMock()
    
        with taddons.context() as tctx:
            inst = ServerInstance.make("regular@127.0.0.1:0", manager)
            await inst.start()
            assert inst.last_exception is None
            assert await tctx.master.await_log("proxy listening")
    
            host, port, *_ = inst.listen_addrs[0]
            reader, writer = await asyncio.open_connection(host, port)
            assert await tctx.master.await_log("client connect")
    
            writer.close()
            await writer.wait_closed()
>           assert await tctx.master.await_log("client disconnect")

test/mitmproxy/proxy/test_mode_servers.py:70: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <mitmproxy.test.taddons.RecordingMaster object at 0x7f2db76174f0>, txt = 'client disconnect'
level = None, timeout = 1

    async def await_log(self, txt, level=None, timeout=1):
        # start with a sleep(0), which lets all other coroutines advance.
        # often this is enough to not sleep at all.
        await asyncio.sleep(0)
        for i in range(int(timeout / 0.01)):
            if self.has_log(txt, level):
                return True
            else:
                await asyncio.sleep(0.01)
>       raise AssertionError(f"Did not find log entry {txt!r} in {self.logs}.")
E       AssertionError: Did not find log entry 'client disconnect' in [LogEntry(HTTP(S) proxy listening at 127.0.0.1:43019., info), LogEntry(127.0.0.1:55052: client connect, info)].

mitmproxy/test/taddons.py:51: AssertionError
___________________________________________ test_transparent[False] ____________________________________________

failure = False, monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f2db7622890>

    @pytest.mark.parametrize("failure", [True, False])
    async def test_transparent(failure, monkeypatch):
        manager = MagicMock()
    
        if failure:
            monkeypatch.setattr(platform, "original_addr", None)
        else:
            monkeypatch.setattr(platform, "original_addr", lambda s: ("address", 42))
    
        with taddons.context(Proxyserver()) as tctx:
            tctx.options.connection_strategy = "lazy"
            inst = ServerInstance.make("transparent@127.0.0.1:0", manager)
            await inst.start()
            await tctx.master.await_log("proxy listening")
    
            host, port, *_ = inst.listen_addrs[0]
            reader, writer = await asyncio.open_connection(host, port)
    
            if failure:
                assert await tctx.master.await_log("Transparent mode failure")
                writer.close()
                await writer.wait_closed()
            else:
                assert await tctx.master.await_log("client connect")
                writer.close()
                await writer.wait_closed()
>               assert await tctx.master.await_log("client disconnect")

test/mitmproxy/proxy/test_mode_servers.py:102: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <mitmproxy.test.taddons.RecordingMaster object at 0x7f2db761b040>, txt = 'client disconnect'
level = None, timeout = 1

    async def await_log(self, txt, level=None, timeout=1):
        # start with a sleep(0), which lets all other coroutines advance.
        # often this is enough to not sleep at all.
        await asyncio.sleep(0)
        for i in range(int(timeout / 0.01)):
            if self.has_log(txt, level):
                return True
            else:
                await asyncio.sleep(0.01)
>       raise AssertionError(f"Did not find log entry {txt!r} in {self.logs}.")
E       AssertionError: Did not find log entry 'client disconnect' in [LogEntry(transparent proxy listening at 127.0.0.1:36767., info), LogEntry(127.0.0.1:52592: client connect, info)].

mitmproxy/test/taddons.py:51: AssertionError
```

mitmproxy/net/udp.py
```diff
@@ -96,7 +126,8 @@ def __init__(
 
     def connection_made(self, transport: asyncio.BaseTransport) -> None:
         if self._transport is None:
-            self._transport = cast(asyncio.DatagramTransport, transport)
+            self._transport = cast(DatagramTransport, transport)
+            self._transport.set_protocol(self)
```
decathorpe, line 130: Is setting `protocol` to `self` correct, though? Looking at the code in the Python `asyncio` module, the `Transport` and `Protocol` interfaces are distinct and serve different purposes.

mitmproxy/net/udp.py
```diff
@@ -77,11 +80,38 @@ async def wait_closed(self) -> None:
...
+        self._server.send_datagram(data, self._local_addr, addr or self._remote_addr)
+
+    def get_extra_info(self, name: str, default: Any = None) -> Any:
+        if name == "sockname":
+            return self._server.getsockname()
+        else:
+            raise NotImplementedError
+
+    def get_protocol(self):
+        return self
```
decathorpe, line 101: Similar problem here. `self` is a `Transport` instance, not a `Protocol` instance.

mitmproxy/net/udp.py
```diff
@@ -175,7 +206,7 @@ def __init__(
         self._transport = transport
         self._remote_addr = remote_addr
         proto = transport.get_protocol()
-        assert isinstance(proto, DrainableDatagramProtocol)
+        assert isinstance(proto, (DrainableDatagramProtocol, WireGuardDatagramTransport))
```
decathorpe, line 209: Same problem here. If I understand correctly, `Protocol` != `Transport`.

mhils:
> Looks like your two most recent commits introduced two new test failures,

They've been there before :)

Protocol and Transport are indeed different things, but we really need neither here except for sending. No particular opinions on how we solve it as long as it works. :)

decathorpe:
> > Looks like your two most recent commits introduced two new test failures,
> 
> They've been there before :)

No, I'm very sure they weren't. The CI was all-green before you pushed the "simplify configuration" commits. I'm very careful not to push changes that fail `tox -e py`, `tox -e mypy`, or `tox -e flake8`.

> Protocol and Transport are indeed different things, but we really need neither here except for sending. No particular opinions on how we solve it as long as it works. :)

🤷🏼‍♂️ Well, if you don't care, then I probably shouldn't either.

mhils:
> No, I'm very sure they weren't.

Seems like I can't GitHub. Sorry. 🫣

> 🤷🏼‍♂️ Well, if you don't care, then I probably shouldn't either.

From mitmproxy's perspective we only really care about DatagramReader/DatagramWriter, which happen to have some internal dependencies on transport/protocol. We shouldn't try to copy all that asyncio complexity, if Reader/Writer are happy we're good. :)

> I bumped the dependency to the just-released mitmproxy_wireguard v0.1.2 which reverts the ChecksumCapability::ignored change. It was the reason why TCP connections were broken in v0.1.1.

Can you elaborate on what exactly was broken without checksums? I'm surprised to see packets with in valid checksums outside of the test client. 

If we need checksums we should revert 80b63bf991e26e2aedde53f3d2ea0cd99a72f801 entirely, otherwise we'll have a sneaky memory leak. Finally we need to adjust the test client to compute proper checksums, but that should be doable.

decathorpe:
> Can you elaborate on what exactly was broken without checksums?

I am not sure why, but the TCP stack just didn't establish connections at all when adding `ChecksumCapabilities::ignored` to the interface. I tested it with mitmweb, and it showed no TCP flows with this change. Removing it again, I see the expected flows again.

mhils:
Ah. Random hypothesis: With `ChecksumCapabilities::ignored`, smoltcp also stops calculating correct checksums, which my client ignores but your client refuses to accept.

> Checking TCP packet payload checksum is independent of this, IIUC.

I think smoltcp uses the device capabilities for all checksums, including TCP:
https://github.com/smoltcp-rs/smoltcp/blob/6b8b667be5878fbf288365df9260996116b81fa6/src/iface/interface.rs#L2620-L2625
https://github.com/smoltcp-rs/smoltcp/blob/6b8b667be5878fbf288365df9260996116b81fa6/src/wire/tcp.rs#L808-L810

I'll adjust the test client to emit correct checksums.

----------------------------

One unrelated question: I noticed the MTU in `test_echo_server` from https://github.com/decathorpe/mitmproxy_wireguard/commit/09325dafcdf78f5f0ab7f54150b7cf36b6ac22ce.
I think the 64k MTU in wg-quick comes from here: https://git.zx2c4.com/wireguard-tools/tree/src/wg-quick/linux.bash#n125
Would it work to just increase our recv/send buffer sizes from 1500 to 64k so what we support all cases? (except for the smoltcp device, which can stay at 1500). I don't have a test setup where I run into such problems.

decathorpe:
> I'll adjust the test client to emit correct checksums.

Great, thanks!

> One unrelated question: I noticed the MTU in `test_echo_server` from [decathorpe/mitmproxy_wireguard@09325da](https://github.com/decathorpe/mitmproxy_wireguard/commit/09325dafcdf78f5f0ab7f54150b7cf36b6ac22ce). I think the 64k MTU in wg-quick comes from here: https://git.zx2c4.com/wireguard-tools/tree/src/wg-quick/linux.bash#n125 Would it work to just increase our recv/send buffer sizes from 1500 to 64k so what we support all cases? (except for the smoltcp device, which can stay at 1500). I don't have a test setup where I run into such problems.

If I remember correctly, this happened when I ran `echo_test_server.py` on `0.0.0.0` and had the wireguard configuration for it connect to it via `0.0.0.0` as well. I don't speak bash, but it seems plausible that wg-quick might set the MTU to 64 KiB if it thinks it's not sending packets over a network interface at all.

The TCP buffer sizes are already 64 KiB:
https://github.com/decathorpe/mitmproxy_wireguard/blob/0.1.2/src/network.rs#L290-L293

The WireGuard and UDP packet buffer sizes are 1500 bytes:
https://github.com/decathorpe/mitmproxy_wireguard/blob/0.1.2/src/wireguard.rs#L120
https://github.com/decathorpe/mitmproxy_wireguard/blob/0.1.2/src/wireguard.rs#L130

I can increase those to 64 * 1024 as well. That should work around the problem with truncated packets when using MTU > 1420?

mhils:
Yes, could you try that and see if it fixes things? :-)

decathorpe:
> Yes, could you try that and see if it fixes things? :-)

Sorry, the edit to my previous commit was probably not the most visible place to put this:

> I don't think that will work. The TCP handling code passes incoming packets to the virtual network device, which cannot handle receiving / sending packets larger than its MTU, and the maximum value accepted by smoltcp is only 9216 bytes (9 KiB ?).

So, if we start to allow packets that are larger than the MTU set in the virtual network device, they will be truncated by smoltcp. It might be possible to dynamically set MTU when initializing the virtual network device, but I don't see an obvious way to determine the "safe" MTU at runtime. So setting it to 1420 is probably the safest and easiest solution.

I also pushed a new release of mitmproxy_wireguard to GitHub (with binaries attached as an asset) for the test client built for linux-x86_64, macos-{x86_64,aarch64}, and windows-x64. These should be usable for testing the WireGuard server mode here?

https://github.com/decathorpe/mitmproxy_wireguard/releases/tag/0.1.4

mhils:
> I also pushed a new release of mitmproxy_wireguard to GitHub (with binaries attached as an asset) for the test client built for linux-x86_64, macos-{x86_64,aarch64}, and windows-x64. These should be usable for testing the WireGuard server mode here?

Absolutely, Honestly I don't see much change for the test binary at all, just take a random x64 Linux binary, copy it into the repo here and build a simple test with it. :)

decathorpe:
Looks like the test *almost* works.

But somehow the `WireGuardServerInstance._server` attribute is `None` in the `echo_udp` function (link below), even though the server *is* running and `assert inst.is_running` succeeds ...

https://github.com/mitmproxy/mitmproxy/pull/5562/commits/6c22471478d0b2955f8fe21f615b09ba27c10da3#diff-1ebfcdb52f4d132251ed110e13c2ce7bf39f0e2d9d39b38ed9f4c742d8ef070dR123

decathorpe:
I finally managed to fix the test client (tracked down a packet that didn't have correct checksums). Not sure what your commits changed about the binaries though? Either way, pushed the latest version of the test client binaries as of 0.1.6.

decathorpe:
Looks like the CI is finally happy, with the exception of some missed lines in test coverage?

decathorpe:
🎉

PR #5563, opened 2022-08-30
Make default_port protocol specific.

meitinger:
#### Description

Fixes https://github.com/mitmproxy/mitmproxy/pull/5556#discussion_r957481469.

### DIFF
--- a/mitmproxy/proxy/mode_specs.py
+++ b/mitmproxy/proxy/mode_specs.py
@@ -51,10 +51,6 @@ class ProxyMode(Serializable, metaclass=ABCMeta):
 
     type: ClassVar[str]  # automatically derived from the class name in __init_subclass__
     """The unique name for this proxy mode, e.g. "regular" or "reverse"."""
-    default_port: ClassVar[int] = 8080
-    """
-    Default listen port of servers for this mode, see `ProxyMode.listen_port()`.
-    """
     __types: ClassVar[dict[str, Type[ProxyMode]]] = {}
 
     def __init_subclass__(cls, **kwargs):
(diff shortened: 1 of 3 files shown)


mitmproxy/proxy/mode_specs.py
```diff
@@ -51,10 +51,6 @@ class ProxyMode(Serializable, metaclass=ABCMeta):
 
     type: ClassVar[str]  # automatically derived from the class name in __init_subclass__
     """The unique name for this proxy mode, e.g. "regular" or "reverse"."""
-    default_port: ClassVar[int] = 8080
```
mhils, line 54: I think this can stay, just as `int` instead of `ClassVar[int]`? Then we don't need the abstract property stuff. It's ok to have a class-level default that is overwritten on a per-instance basis. :)

mhils (approved):

mitmproxy/proxy/mode_specs.py
meitinger: Actually I tried that before in #5556, but with `@dataclass(frozen=True)` it cannot be set in `__post_init__`.

mhils: Was that maybe the case because you had it as a ClassVar[int] instead of int? Because we do set other attrs in `__post_init__`. :-)

meitinger: The `ClassVar`s are still from the/your original code. If an attribute is declared in the subclass (like `address` in `ReverseMode`), setting it in `__post_init__` works. If it's declared in `ProxyMode`, I couldn't get it to be overridden post init.
But I'm more than happy to be proven wrong (my Python-Fu isn't the strongest), then we can turn all `@abstractmethod @property`s into ordinary ones  :)

PR #5590, opened 2022-09-17
Switch to stdlib logging

mhils:
mitmproxy previously used a homegrown logging mechanism based around
`mitmproxy.ctx.log` and the `add_log` hook. This worked well for everything
we control, but does not work outside the mitmproxy universe.
For now we have simply ignored logging in e.g. tornado or h2, but with the
upcoming introduction of mitmproxy_wireguard we now have a dependency
on some Rust/PyO3 code for which we definitely want logs, but which also
cannot easily be changed to use our homegrown logging (PyO3 does the heavy
lifting to add interoperability with stdlib logging). Long story short,
we want to introduce a log handler for stdlib logging.

Now there are two ways how such a handler could operate:

 1. We could build a handler that forwards all stdlib log events
    into our homegrown mechanism.
 2. We embrace stdlib's logging as the correct way to do things,
    and get rid of our homegrown stuff.

This PR follows the second approach by removing the `add_log` hook and
rewriting the `TermLog` and `EventStore` addons to listen for stdlib log records.
This means that all `mitmproxy.ctx.log.info` events are now simply `logging.info` etc.

One upside of this approach is that many parts of the codebase now don't depend
on the existence of `mitmproxy.ctx` and we can use off-the-shelf things like pytest's
`caplog`. We can also now better colorize log output and/or add timestamps.

@mitmproxy/devs: There's a breaking API change in here, feedback is welcome!

### DIFF
--- a/docs/src/content/addons-api-changelog.md
+++ b/docs/src/content/addons-api-changelog.md
@@ -28,7 +28,7 @@ logging.info("hello world")
 ```
 
 
-Accordingly, the `add_log` event has been removed. Developers who rely on log entries need to register their own 
+Accordingly, the `add_log` event has been deprecated. Developers who rely on log entries should register their own 
 `logging.Handler` instead. An example for this can be found in the `EventStore` addon.
 
 ## mitmproxy 7.0
--- a/docs/src/content/addons-api-changelog.md
+++ b/docs/src/content/addons-api-changelog.md
@@ -10,6 +10,27 @@ menu:
 
 We try to avoid them, but this page lists breaking changes in the mitmproxy addon API.
 
+## mitmproxy 9.0
+
+#### Logging
+
+We've deprecated mitmproxy's homegrown logging system in favor of Python's builtin `logging` module.
+This means that addons should now use standard logging functionality instead of `mitmproxy.ctx.log`:
+
+```python
+# Deprecated:
+from mitmproxy import ctx
+ctx.log.info("hello world")
+
+# New:
+import logging
+logging.info("hello world")
+```
+
+
+Accordingly, the `add_log` event has been removed. Developers who rely on log entries need to register their own 
+`logging.Handler` instead. An example for this can be found in the `EventStore` addon.
+
 ## mitmproxy 7.0
 
 #### Connection Events
--- a/examples/contrib/save_streamed_data.py
+++ b/examples/contrib/save_streamed_data.py
@@ -18,16 +18,16 @@ for associating a file with its corresponding flow in the stream saved with
 This addon is not compatible with addons that use the same mechanism to
 capture streamed data, http-stream-modify.py for instance.
 """
+import logging
+import os
+from datetime import datetime
+from pathlib import Path
 from typing import Optional
 
 from mitmproxy import ctx
-from datetime import datetime
-from pathlib import Path
-import os
 
 
 class StreamSaver:
-
     TAG = "save_streamed_data: "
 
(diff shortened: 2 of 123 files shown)


docs/src/content/addons-api-changelog.md
```diff
@@ -10,6 +10,27 @@ menu:
...
+from mitmproxy import ctx
+ctx.log.info("hello world")
+
+# New:
+import logging
+logging.info("hello world")
+```
+
+
+Accordingly, the `add_log` event has been removed. Developers who rely on log entries need to register their own
```
Kriechi, line 31: `We've deprecated ` vs. `has been removed`
Aren't they mutually exclusive? Or are we not following a deprecate-release-remove-release strategy here?

examples/contrib/save_streamed_data.py
```diff
@@ -58,7 +58,8 @@ def __call__(self, data):
             return data
 
         if not self.fh:
-            self.path = datetime.fromtimestamp(self.flow.request.timestamp_start).strftime(ctx.options.save_streamed_data)
+            self.path = datetime.fromtimestamp(self.flow.request.timestamp_start).strftime(
+                ctx.options.save_streamed_data)
```
Kriechi, line 62: is this auto-formatting or did you feel today is a day for shorter lines?

mitmproxy/addonmanager.py
```diff
@@ -88,7 +91,7 @@ def add_option(
             if same_signature:
                 return
             else:
-                ctx.log.warn("Over-riding existing option %s" % name)
+                logger.warning("Over-riding existing option %s" % name)
```
Kriechi, line 94: call-out for the future PRs: let's move towards f-strings when we touch such a line.

mitmproxy/log.py
```diff
@@ -1,7 +1,91 @@
...
+from mitmproxy.utils import human
+
+ALERT = logging.INFO + 1
+"""
+The ALERT logging level has the same urgency as info, but
+signals to interactive tools that the user's attention should be
+drawn to the output even if they're not currently looking at the
+event log.
+"""
+logging.addLevelName(ALERT, "ALERT")
```
Kriechi, line 15: Thoughts on also adding a `logging.alert(...)` function to make it look and behave the same way as built-in levels?
https://stackoverflow.com/a/35804945

Kriechi (approved):
Generally 👍  but only a few questions.

examples/contrib/save_streamed_data.py
mhils: This is auto-formatting...

mitmproxy/log.py
mhils: I've been on that SO page as well, but intentionally decided against this as it does not really play well with mypy. We could add it for quick-and-dirty addons and then not use it ourselves, but not sure if that is particularly useful.

docs/src/content/addons-api-changelog.md
mhils: Right now `ctx.log` is deprecated, but `add_log` has been removed. You're making a reasonable point though, I can add a bit of boilerplate to still trigger `add_log` for now and just emit a deprecation warning.

mitmproxy/log.py
Kriechi: Ok. My only fear is that nobody will use ALERT, because it requires this extra little step - compared to a possibly auto-completed function.

PR #5600, opened 2022-09-21
Update options-simple.py

daycoder:
Usage in the docstring is incorrect. Changes "--set addheader true" to "--set addheader=true"

#### Checklist


daycoder:
I was attempting to run based on the usage in the docstring. Only after coming back to it the next day did I see the correct usage further down in https://docs.mitmproxy.org/stable/addons-options/

mhils (approved):
Thanks! 🍰

PR #5607, opened 2022-09-25
docs: add documentation for wireguard mode

decathorpe:
There's a new section for "wireguard" mode in the "Modes of operation" page, below the existing docs for the "transparent" mode. I included documentation for all configuration knobs (i.e. how to specify the path to the file with encryption keys, and how to set a custom listen port) and a short explanation of the limitations of the new mode.

I also added a short note to the "Transparent HOWTO" page pointing out that the new WireGuard mode might be easier to use and that it works on Windows (which `--mode transparent` apparently doesn't? TIL).

### DIFF
--- a/docs/src/content/concepts-modes.md
+++ b/docs/src/content/concepts-modes.md
@@ -9,6 +9,7 @@ menu:
 
 - [Regular](#regular-proxy) (the default)
 - [Transparent](#transparent-proxy)
+- [WireGuard](#wireguard-transparent-proxy)
 - [Reverse Proxy](#reverse-proxy)
 - [Upstream Proxy](#upstream-proxy)
 - [SOCKS Proxy](#socks-proxy)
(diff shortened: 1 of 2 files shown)


docs/src/content/concepts-modes.md
```diff
@@ -138,6 +139,62 @@ most cases, the configuration will look like this:
...
+
+```shell
+mitmdump --mode wireguard
+```
+
+The WireGuard mode works in the same way as transparent mode, except that setup
+and routing client traffic to mitmproxy are different. In this mode, mitmproxy
+runs an internal WireGuard server, which devices can be connected to by using
+standard WireGuard clients applications, and just by importing the WireGuard
+configuration provided by mitmproxy.
```
mhils, line 152:
```suggestion
standard WireGuard clients applications:

1. Start `mitmweb --mode wireguard`.
2. Install Wireguard on target device.
3. Import the WireGuard config provided by mitmproxy on the device.
```

(documentation consumers generally hate prose and prefer simple lists...)

docs/src/content/concepts-modes.md
```diff
@@ -138,6 +139,62 @@ most cases, the configuration will look like this:
...
+the WireGuard server runs entirely in userspace, so no administrative privileges
+are necessary for setup or operation of mitmproxy in this mode.
+
+### Configuration
+
+By default, the WireGuard server will listen on port `51820/udp`, the default
+port for WireGuard servers, which should not require modifying firewall settings
+in most circumstances. However, changing firewall settings might be necessary
+when specifying a different port (i.e. when using something like
+`--mode wireguard@51821`).
```
mhils, line 166:
```suggestion
port for WireGuard servers. This can be changed by setting the `listen_port` 
option or by specifying an explicit port (`--mode wireguard@51821`).
```

(let's avoid the firewall part here, I'm not sure if anyone is any wiser after reading that changing the firewall may or may not be necessary)

docs/src/content/concepts-modes.md
```diff
@@ -138,6 +139,62 @@ most cases, the configuration will look like this:
...
+
+The encryption keys for WireGuard connections are stored in a file on disk, and
+it is possible to specify its path manually with `--mode wireguard:path`. If no
+path is specified, a fallback path within the default mitmproxy configuration
+directory is used. If no file exists at that path when the WireGuard server is
+instantiated, new keys will be generated and stored there automatically.
```
mhils, line 172:
```suggestion
The encryption keys for WireGuard connections are stored in `~/.mitmproxy/wireguard.conf`. It is possible to specify a custom path manually with `--mode wireguard:path`. New keys will be generated automatically if the specified file does not exist. For example, to connect two clients simultaneously, you can run `mitmdump --mode wireguard:wg-keys-1.conf --mode wireguard:wg-keys-2.conf@51821`.
```

docs/src/content/concepts-modes.md
```diff
@@ -138,6 +139,62 @@ most cases, the configuration will look like this:
...
+
+With the current implementation, it is not possible to proxy all traffic of the
+host that mitmproxy itself is running on, since this would result in outgoing
+WireGuard packets being sent over the WireGuard tunnel themselves.
+
+However, it is possible to limit the IP addresses for which traffic is sent over
+the WireGuard tunnel to specific ranges. In this case, the `AllowedIPs` setting
+in the WireGuard client configuration needs to be changed from `0.0.0.0/0` (i.e.
+"route *all* IPv4 traffic through WireGuard tunnel") to the required ranges of
+IP addresses (this setting allows multiple, comma-separated values).
```
mhils, line 186: That does not really help by itself, no? The problem is that you additionally need to redirect, otherwise it's caught again. I'd just skip this for now. The great plan is to have dedicated "current OS proxy" modes anyhow.

mhils (commented):
Looks good. :+1:

Some suggestions below. Maybe we should add a short comment on the Endpoint IP needing adjustment on more complex network layouts.

docs/src/content/concepts-modes.md
decathorpe: LGTM.

docs/src/content/concepts-modes.md
decathorpe: Makes sense. I'll adapt.

docs/src/content/concepts-modes.md
decathorpe: I still think that setting up multiple peers for one WireGuard server would be better than running multiple WireGuard servers, but for now, this makes sense.

docs/src/content/concepts-modes.md
decathorpe: You're right. I got confused because this is how it works for the echo server, but that of course doesn't help for transparent proxy. I'll remove that paragraph.

decathorpe:
I think I addressed all comments / merged all suggestions with the new commit.

mhils (approved):

PR #5613, opened 2022-09-28
har_dump.py - get real url in transparent mode

BigSully:
Dump the real url rather than the url containing ip address when mitmproxy works in transparent mode.

#### Description

Currently, flow.request.url shows the ip address rather than host name when mitmproxy works on transparent mode.

flow.request.pretty_url can show the correct url.

#### Checklist


mhils:
Thanks!

PR #5650, opened 2022-10-16
Added dark mode for the Web UI

devapro:
#### Description

Added Dark mode for the web interface.

- Added colors variables
- Added colors set for dark mode
- Added settings toggle "Dark mode"

Related issue https://github.com/mitmproxy/mitmproxy/issues/3886

How it looks you can check in this video:

https://user-images.githubusercontent.com/2677592/196033872-74873f19-ed17-4a88-8968-5191d15c581a.mp4

#### Checklist


mhils:
Thank you for the PR! I very much appreciate the work you've put into this and the intent to improve mitmproxy. Nonetheless, I'm relatively strongly against merging this. Dark mode adds a substantial amount of maintenance overhead, which is something I don't want to take on. As the primary maintainer of mitmweb, I don't want the additional burden of testing all changes with multiple color schemes and adding dark mode styles for all new UI elements. We have some long-overdue modernization to do on the CSS front (upgrading from Bootstrap 3), and dark modes just adds more work on top of this. 

Long story short, I'm happy to have mitmweb dark mode stay around as a user style, and I'd be more than happy to link it prominently in `examples/contrib` and refer folks to it. However, unless the rest of @mitmproxy/devs comes out as being strongly in favor of this (with some maintenance commitments), I don't think we'll merge this PR. Of course, I'd be more than happy to chat about other potential mitmweb improvements before you start hacking on stuff!

devapro:
Thank you for your comment! I agree with your point of view - the additional burden in maintaining or testing it is not good, however, in this PR I added an approach with color variables instead of using different color values in the styles. This approach should reduce efforts of maintaining UI in the future, I mean that having a list of the colors is easier than having infinity variants of the colors in the styles. If new UI elements will use colors from the existing set it shouldn't have any impact on maintenance. In case if some new elements will need new colors it is easy to add them to the colors set.

sangsatori:
I made a half-hearted stab at implementing dark mode after being put off by lack of it earlier on, but soon abandoned after finding Bootstrap 3 unpleasant to override.

Just want to say, I find your implementation looking good and code clean. Great job on hunting down colours throughout these stylesheets! :+1:

---

Certainly I agree that Bootstrap 3 needs upgrading from. Though for now going forward with having extracted all colours used in the UI to CSS variables seems _more_ maintainable than before. :thinking:

mhils:
> This approach should reduce efforts of maintaining UI in the future, I mean that having a list of the colors is easier than having infinity variants of the colors in the styles. If new UI elements will use colors from the existing set it shouldn't have any impact on maintenance. In case if some new elements will need new colors it is easy to add them to the colors set.

Thanks - I've looked at the implementation and I'm very familiar with the pattern.[^1] First, I don't think you get around doing additional testing in dark mode, so it definitely comes with a constant maintenance cost. Second, dark mode often is not as easy as "adding new colors to the color set". For example, you can't simply change icon colors unless you construct SVG icons in a very specific way or use icon fonts. Finally, upgrading from Bootstrap 3 will be a major rewrite, and an already-existing, conflicting set of color variables just gets in the way.

> Just want to say, I find your implementation looking good and code clean.

FWIW I absolutely agree that this PR looks great, and I probably should have pointed that out more prominently in my first response. I like the code, but I don't like the long-term maintenance implications. If anyone with a track record of making constant contributions to the project would want this as their pet feature in I'd be the last one to complain, but I simply don't want to spend my time maintaining on dark mode. 
Again, I have nothing against a mitmweb dark mode, but it does not have to live in our main repository, I'd much rather prefer it as a user style maintained somewhere else. :)

[^1]: https://github.com/mitmproxy/pdoc/tree/main/examples/dark-mode

devapro:
Ok, I can move styles to another repository with custom styles, but could you help me with instructions on how custom styles can be added to the existing installation of the mitmproxy?
For example, if I created a repository with custom styles, how other people can use them?

During working on this PR, I thought maybe will be better to have a separate repository for web UI? Probably in this case more people will be involved in working on Ui part. (It is just my thoughts)

mhils:
> Ok, I can move styles to another repository with custom styles, but could you help me with instructions of how custom styles can be applied to the existing installation of the mitmproxy?

I'm afraid we don't have a really good API for this. I was mostly thinking of something like https://userstyles.org/, which works via a browser extension and not a mitmproxy addon.

> During working on this PR, I thought maybe will be better to have a separate repository for web UI? Probably in this case more people will be involved in working on Ui part. (It is just my thoughts)

Can you explain how you think that would help attract contributors? I'm genuinely curious. :)

SekiBetu:
https://github.com/darkreader/darkreader
This extension works fine for me

devapro:
> Can you explain how you think that would help attract contributors? I'm genuinely curious. :)

I mean that a separate repository with only the frontend part will be easier to maintain by people who have experience in frontend development only and it prevents breaking the main project. Maybe in this case more people will add more improvements to UI which is important to make the whole project more popular.

devapro:
I created a new [PR](https://github.com/mitmproxy/mitmproxy/pull/5815)

PR #5661, opened 2022-10-19
fix vScroll calcuation, refs #5658

mhils:
This PR is an alternative fix for #5658: `calcVScroll` now bounds viewportTop so that additional checks should not be necessary.

@LIU-shuyi, can you verify that this works as expected please? 😃

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,8 @@
 * DTLS support ([#5397](https://github.com/mitmproxy/mitmproxy/pull/5397), @kckeiks).
 * Added Magisk module generation for Android onboarding (@jorants).
 * Update Linux binary builder to Ubuntu 20.04, bumping the minimum glibc version to 2.31. (@jorants)
-* "Save filtered" button in mitmweb. ([#5531](https://github.com/mitmproxy/mitmproxy/pull/5531), @rnbwdsh, @mhils)
+* "Save filtered" button in mitmweb.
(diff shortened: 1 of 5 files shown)


LIU-shuyi:
Hi, I tested it, but didn't work well in the scene i described in #5658. Same error as before.

the return value of calcVScroll() in onViewportUpdate() will not influence `viewportTop` directly.
viewport.scrollTop is set to viewport.scrollHeight in `AutoScroll.tsx: componentDidUpdate()`

base on your change , I add code as fallows, and it works:
```
diff --git a/web/src/js/components/FlowTable.jsx b/web/src/js/components/FlowTable.jsx
index 45f12e9d3..6bb1873b9 100644
--- a/web/src/js/components/FlowTable.jsx
+++ b/web/src/js/components/FlowTable.jsx
@@ -85,7 +85,10 @@ class FlowTable extends React.Component {
         })
 
         if (this.state.viewportTop !== viewportTop || !shallowEqual(this.state.vScroll, vScroll)) {
-            this.setState({ vScroll, viewportTop })
+            this.setState({
+                vScroll,
+                viewportTop: Math.min( viewportTop, vScroll.end * this.props.rowHeight),
+            });
         }
     }
```

mhils:
Thanks! 🍰 Seems like I wasn't testing with enough rows. I've been able to reproduce it now, and incorporated your fix. Can you confirm that this works now? 😃

LIU-shuyi:
Aha, the code works well this time.

mhils:
Awesome, thanks for the initial PR and the quick confirmation! 🍰

PR #5666, opened 2022-10-20
Make HAR files compatible with Firefox DevTools

m4ci3k2:
Firefox Developer Tools reject to load HAR files that do not have `har.log.pages` element (see
<https://bugzilla.mozilla.org/show_bug.cgi?id=1691240>). Adding a placeholder entry fixes it.


mhils (approved):
Thanks! 🍰

PR #5687, opened 2022-10-27
Fix missing proxy port when starting a browser.

rbdixon:
#### Description

When starting a browser with `browser.start` and the `listen_port` value set to default the proxy is started with `--proxy-server=127.0.0.1:None`. This causes Chrome to ignore the proxy.

#### Checklist


mhils:
Great catch, thanks! 🍰 😃

PR #5697, opened 2022-10-31
Add support for Firefox

### LINKED ISSUE #5247: Add more commands to start applications with mitmproxy preconfigured
Many mitmproxy users only want to proxy one specific application. We already have a [`browser.start` command](https://github.com/mitmproxy/mitmproxy/blob/ef69701b0a64df97fc92ed568d39eed987fe8bdf/mitmproxy/addons/browser.py#L29-L74) to boot up a Chrome instance, it would be very useful to have similar wrappers for other browsers and applications:

- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] Docker Containers
- [ ] Android Devices

## Implementation Steps

1. Take one of the targets above and figure out how to start it with a proxy server preconfigured.
2. Extend the [browser addon](https://github.com/mitmproxy/mitmproxy/blob/main/mitmproxy/addons/browser.py) with a custom command that invokes the application.
3. Add a test in https://github.com/mitmproxy/mitmproxy/blob/main/test/mitmproxy/addons/test_browser.py. These tests aren't particularly useful as we want to mock the actual application, but we require 100% test coverage for all builtin addons.

mrwacky42:
#### Description

Partial fix for #5247. Add basic support for Firefox

#### Checklist

### DIFF
--- a/mitmproxy/addons/browser.py
+++ b/mitmproxy/addons/browser.py
@@ -50,12 +50,24 @@ def get_chrome_flatpak() -> Optional[str]:
     return None
 
 
+def get_firefox_executable() -> Optional[str]:
+    for browser in (
+        "/Applications/Firefox.app/Contents/MacOS/Firefox",
+        r"C:\Program Files\Mozilla Firefox\firefox.exe",
+        "firefox",
+    ):
+        if shutil.which(browser):
+            return browser
+
+    return None
+
 def get_browser_cmd() -> Optional[list[str]]:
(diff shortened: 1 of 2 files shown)


mhils:
Thanks for the PR! This does not really set proper command line options for Firefox, does it?

mrwacky42:
Hi @mhils - no I guess I wasn't paying very close attention. I'll try to update this in the next day or two.

Issue #5449, opened 2022-07-10
cannot get server connection to ilo3 (unsupported protocol version)


markusdd:
Hi, I am trying to use mitmdump to still access my HP ilo3 management interfaces. These stopped at supporting 
TLS v1.1 and essentially every Browser nowadays refuses to connect, even with the AES ciphers turned on.

From a CentOS 7 machine, I can successfully curl the login page like so:
` curl -k --tlsv1.1 <my ip>`

I start mitmdump like this (setting cipher_server to 'ALL' does not help):
`mitmdump --set tls_version_server_min=TLS1_1  --set tls_version_server_max=TLS1_1 --ssl-insecure -p 9443 --mode reverse:https://192.1xxxxxxxx/`

Wireshark shows this. Could it be, that the Ciphersuite is not supported by mitmdump? I tried setting 'ALL', but also with TLS I had to be very specific as mitmdump would first try TLSv1.0, that would fail and it would not retry the others.
![image](https://user-images.githubusercontent.com/25175069/178127761-063d3add-8308-4693-9629-17c00e5a86b6.png)

I can load the page in firefox after waiving the pinned cert, but then I get the same message as mitmdump prints on the commandline, so mitmdump won't talk to the server:
![image](https://user-images.githubusercontent.com/25175069/178127802-7bb19e0f-11a0-4fba-93b2-3a40c85c6eed.png)

To successfully ssh into the ilo port, I have to use this contraption, but it works:
`ssh -oKexAlgorithms=+diffie-hellman-group14-sha1 -oHostKeyAlgorithms=+ssh-dss -oCiphers=+aes256-ctr`

Any hints which syntax/cipher to use to make this work?

```
Mitmproxy: 8.1.1 binary
Python:    3.10.5
OpenSSL:   OpenSSL 3.0.3 3 May 2022
Platform:  Linux-5.18.9-200.fc36.x86_64-x86_64-with-glibc2.35
```

markusdd:
Ok, some more info:

if I add `--set ciphers_server=TLS_RSA_WITH_AES_128_CBC_SHA ` to the commandline what I get is this:

![image](https://user-images.githubusercontent.com/25175069/178128037-c8addd4d-e7f8-499b-896d-331ad577165f.png)

markusdd:
ok, so I have it working now, albeit with mitmproxy 6.0.2, so an outdated version.

[...]

PR #5698, opened 2022-10-31
Allow unsafe legacy renegotiation

alexyao2015:
#### Description

I had a very similar issue as in #5449 and was able to solve it with this change.

See here for more info https://www.openssl.org/docs/man3.0/man3/SSL_clear_options.html

#### Checklist


mhils:
Thanks for the PR! I'm happy to cater for this use case, but we can't turn on insecure options by default. It needs to be toggleable mitmproxy option.

Issue #4499, opened 2021-03-12
Cannot create a Master object in a sub-thread of the program.


zongrongjin:
#### Problem Description
When my program is running in a subthread, I can't create a new Master or DumpMaster object in that subthread.

#### Steps to reproduce the behavior:
Run this code:
```
import threading
from mitmproxy import proxy, options
from mitmproxy.tools.dump import DumpMaster

class AddHeader:
    def response(self, flow):
        flow.response.headers["test"] = 'success'

class Test(object):
    def mitm_run(self):
        opts = options.Options(listen_host='0.0.0.0', listen_port=8080)
        opts.add_option('mode', str, 'upstream:http://127.0.0.1:1087', '')
        pconf = proxy.config.ProxyConfig(opts)
        mitm_server = DumpMaster(opts, with_termlog=False)
        mitm_server.server = proxy.server.ProxyServer(pconf)
        class_names = []
        mitm_server.addons.add(AddHeader())
        print(f"mitm server is running...")
        mitm_server.run()

    def other_task(self):
        print('other task run...')

    def start(self):
        threads = []

        for i in [self.mitm_run, self.other_task]:
            thread = threading.Thread(target=i)
            thread.start()
            threads.append(thread)

        for thread in threads:
            thread.join()

if __name__ == '__main__':
    test = Test()
    test.start()
```
Then you will get:
```
Traceback (most recent call last):
  File "C:\Users\PC\AppData\Local\Programs\Python\Python37\lib\threading.py", line 926, in _bootstrap_inner
    self.run()
  File "C:\Users\PC\AppData\Local\Programs\Python\Python37\lib\threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "C:/Users/PC/Desktop/tiktok_user_crawler/test.py", line 17, in mitm_run
    mitm_server = DumpMaster(opts, with_termlog=False)
  File "C:\Users\PC\AppData\Roaming\Python\Python37\site-packages\mitmproxy\tools\dump.py", line 24, in __init__
    super().__init__(options)
  File "C:\Users\PC\AppData\Roaming\Python\Python37\site-packages\mitmproxy\master.py", line 48, in __init__
    asyncio.get_event_loop(),
  File "C:\Users\PC\AppData\Local\Programs\Python\Python37\lib\asyncio\events.py", line 644, in get_event_loop
    % threading.current_thread().name)
RuntimeError: There is no current event loop in thread 'Thread-12'.

```
[...]

PR #5790, opened 2022-12-03
Add loop as a parameter in DumpMaster

dcostersabin:
#### Description

added loop parameter to DumpMaster so that loop can be passed to parent class Master which allows to directly inherit DumpMaster. 

#### Checklist


mhils:
I will merge this for now (it's a harmless change that presumably makes your life easier), but this explicitly is not a commitment to maintaining this in the future. See https://github.com/mitmproxy/mitmproxy/issues/4499#issuecomment-797379786.

PR #5845, opened 2023-01-02
Use latest codecov uploader

mhils:
It looks like codecov finally shut down their v1 API, so we need to update to v3. The official codecov action is unusable for us as it cannot pinned, so I've come up with a custom action that is more sane.


mhils:
Ok, seems like it's a backend problem and their v1 API still works. 🤷

PR #5852, opened 2023-01-06
Fix tests after 2037

bmwiedemann:
Background:
As part of my work on reproducible builds for openSUSE, I check that software still gives identical build results in the future.
The usual offset is +16 years, because that is how long I expect some software will be used in some places.
This showed up failing tests in our package build.
See https://reproducible-builds.org/ for why this matters.

#### Description

Make tests pass after 2037

#### Checklist

### DIFF
--- a/test/mitmproxy/net/http/test_cookies.py
+++ b/test/mitmproxy/net/http/test_cookies.py
@@ -170,15 +170,15 @@ def test_parse_set_cookie_header():
             ],
         ],
         [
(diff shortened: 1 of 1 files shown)


mhils:
Awesome, thanks! 🍰

bmwiedemann:
Wanted to mention this in the commit message, but forgot to push it:

    Year 2133 was chosen because it has the same weekdays as 2037
    and it is far enough in the future.

and I tested that it also passes on i586 atm with python3.10

PR #5866, opened 2023-01-13
Adapt contrib/mitmproxywrapper.py to Python 3 strings

seanf:
This fixes errors like this one: 

`TypeError: memoryview: a bytes-like object is required, not 'str'`

#### Checklist


mhils (approved):
Thanks!

PR #5882, opened 2023-01-17
Add meaningful exit codes

### LINKED ISSUE #5696: Add individual exit codes for different errors
#### Problem Description

This came from https://github.com/mitmproxy/mitmproxy/issues/4669 and was attempted in https://github.com/mitmproxy/mitmproxy/pull/5133

Currently every error that causes mitmproxy to go down has an exit code of `1`. We should add more specific error codes.

#### Proposal

Replace all `sys.exit(1)` with more specific exit codes `>=1` so that a parent process gets more information why mitmproxy (likely `mitmdump`) went down.

Example codes I came up with in the failed PR attempt:

https://github.com/mitmproxy/mitmproxy/pull/5133/files#diff-eb2a1155058d5000102ff2c85ec2acb691ca8277539e2d1c9702dbbab583dfa9R1-R12

kiklidash:
#### Add meaningful exit codes, fixes #5696 

Replaced all instances of sys.exit(1) with exit codes that carry a meaning (outlined in exit_codes.py) for more context if the application exits with a non-zero exit code.

#### Checklist

 - [✓] I have updated tests where applicable.
 - [✓] I have added an entry to the CHANGELOG.


Prinzhorn:
Thanks, it's looking good so far!

@mhils can you help with the missing tests? See details here https://github.com/mitmproxy/mitmproxy/issues/5696#issuecomment-1382124358 I'm sure you just need to give some helpful pointers.

PR #5889, opened 2023-01-24
fix: anatomy.py should set basic config log level to info to show logs

giper45:
#### Description

Changed the anatomy.py example in order to show the logs when the script is executed.

- Expected behavior: the mitmdump script should show the flow logging output. 
- Current behavior: it does not show the flow output as the default settings is not configured. 
Tested python version:  3.10.8

### DIFF
--- a/examples/addons/anatomy.py
+++ b/examples/addons/anatomy.py
@@ -4,6 +4,8 @@ Basic skeleton of a mitmproxy addon.
 Run as follows: mitmproxy -s anatomy.py
 """
 import logging
+logging.basicConfig(level=logging.INFO)
+
 
 
 class Counter:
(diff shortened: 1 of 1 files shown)


mhils:
> as the default settings is not configured.

That's not true, the default setting is configured here:
https://github.com/mitmproxy/mitmproxy/blob/4ebccfa2365a15b88d9342641f35d66b53583548/mitmproxy/tools/main.py#L58-L64

mitmdump does print the output on my system. Are you using the latest mitmproxy release?

giper45:
Sorry @mhils , on my latest kali the apt repo is not updated to the latest (version 8). 
I have installed the latest version (9) and the example properly works. 
Anyway, I think that could be useful for backward compatibility.

mhils:
The v8 docs do have a correct v8 example: https://docs.mitmproxy.org/archive/v8/addons-overview/ :)

See also the warning at https://github.com/mitmproxy/mitmproxy/blob/main/examples/README.md - we don't want to maintain backward compatibility here. Thank you for raising this though, I appreciate the intent!

PR #5896, opened 2023-01-30
Add delete shortcut to delete flows in mitmweb

### LINKED ISSUE #5881: (mitmweb) feat: Add the delete key as a shortcut to delete flows.
#### Problem Description
Currently in mitmweb there is an ability to remove flows with 'd', but it's not intuitive compared to a key literally designed to delete things: 'delete'. For a long time I was using mitmweb without even knowing that you can delete flows with a shortcut, even though the hint shows that.

#### Proposal
Add a 'delete' key as an alternative to 'd' to delete flows. 'delete' isn't used for anything else in mitmweb, so it shouldn't cause any issues.

#### Additional context
Shortcut handling for flow deletions happens [here](https://github.com/mitmproxy/mitmproxy/blob/1abb8f69217910c8623bd1339da2502aed98ff0d/web/src/js/ducks/ui/keyboard.tsx#L76).

I will be happy to make a PR for it if you agree with the idea.

sujaldev:
#### Description

Added the `delete` key as a shortcut to delete flows in mitmweb. Fixes #5881

#### Checklist

### DIFF
--- a/web/src/js/ducks/ui/keyboard.tsx
+++ b/web/src/js/ducks/ui/keyboard.tsx
@@ -73,6 +73,7 @@ export function onKeyDown(e: KeyboardEvent) {
                 break
             }
 
+            case "Delete":
             case "d": {
                 if (!flow) {
                     return


mhils (approved):
Thanks! 🍰

Issue #5880, opened 2023-01-16
AssertionError: Unexpected event: <AlternativeServiceAvailable origin:https://kaspersky.com:443, field_value:h3=":443">


JannisBush:
#### Problem Description
Mitmproxy crashes for certain responses, probably related to HTTP3.

#### Steps to reproduce the behavior:
1. Run `mitmproxy`
2. Run `curl --proxy http://localhost:8080 -k https://kaspersky.com`
3. Observe error in mitmproxy
```
error: 127.0.0.1:62672: mitmproxy has crashed!
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/mitmproxy/8.1.1/libexec/lib/python3.10/site-packages/mitmproxy/proxy/server.py", line 331, in server_event
    for command in layer_commands:
  File "/opt/homebrew/Cellar/mitmproxy/8.1.1/libexec/lib/python3.10/site-packages/mitmproxy/proxy/layer.py", line 167, in handle_event
    command = next(command_generator)
  File "/opt/homebrew/Cellar/mitmproxy/8.1.1/libexec/lib/python3.10/site-packages/mitmproxy/proxy/layer.py", line 256, in handle_event
    yield from self._handle(event)
  File "/opt/homebrew/Cellar/mitmproxy/8.1.1/libexec/lib/python3.10/site-packages/mitmproxy/proxy/layer.py", line 167, in handle_event
    command = next(command_generator)
  File "/opt/homebrew/Cellar/mitmproxy/8.1.1/libexec/lib/python3.10/site-packages/mitmproxy/proxy/layers/http/__init__.py", line 846, in _handle_event
    yield from self.event_to_child(stream, event)
  File "/opt/homebrew/Cellar/mitmproxy/8.1.1/libexec/lib/python3.10/site-packages/mitmproxy/proxy/layers/http/__init__.py", line 899, in event_to_child
    for command in child.handle_event(event):
  File "/opt/homebrew/Cellar/mitmproxy/8.1.1/libexec/lib/python3.10/site-packages/mitmproxy/proxy/layer.py", line 167, in handle_event
    command = next(command_generator)
  File "/opt/homebrew/Cellar/mitmproxy/8.1.1/libexec/lib/python3.10/site-packages/mitmproxy/proxy/layers/http/__init__.py", line 743, in passthrough
    for command in self.child_layer.handle_event(event):
[...]

PR #5898, opened 2023-02-01
temporary fix for unhandled AlternativeServiceAvailable event

sujaldev:
#### Description

Temporary fix for #5880. `AlternativeServiceAvailable` event will generate a log command. I'm unsure if the test I've written is correct, please suggest changes if it isn't

#### Checklist

### DIFF
--- a/mitmproxy/proxy/layers/http/_http2.py
+++ b/mitmproxy/proxy/layers/http/_http2.py
@@ -275,6 +275,8 @@ class Http2Connection(HttpConnection):
             # https://http2.github.io/http2-spec/#rfc.section.4.1
             # Implementations MUST ignore and discard any frame that has a type that is unknown.
             yield Log(f"Ignoring unknown HTTP/2 frame type: {event.frame.type}")
+        elif isinstance(event, h2.events.AlternativeServiceAvailable):
+            yield Log("Temporarily ignoring an Alt-Svc header")
         else:
             raise AssertionError(f"Unexpected event: {event!r}")
         return False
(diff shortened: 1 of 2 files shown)


test/mitmproxy/proxy/layers/http/test_http2.py
```diff
@@ -1191,3 +1191,31 @@ def advance_time(_):
...
+        >> reply()
+        << http.HttpRequestHook(flow)
+        >> reply()
+        << OpenConnection(server)
+        >> reply(None, side_effect=make_h2)
+        << SendData(server, initial)
+        >> DataReceived(
+            server, cff.build_alt_svc_frame(0, b"example.com", b'h3=":443"').serialize()
+        )
+        << Log(Placeholder(str), Placeholder(int))
```
mhils, line 1220: We can just use the actual values here `<< Log("Temporarily ignoring an Alt-Svc header")` or whatever it is. The primary purpose of placeholders is that they can be put in places where we don't know the actual value in advance (e.g. an HTTPFlow, which is created during execution). For simple strings, we can just specify the same string and rely on equality. :)

mitmproxy/proxy/layers/http/_http2.py
```diff
@@ -275,6 +275,8 @@ def handle_h2_event(self, event: h2.events.Event) -> CommandGenerator[bool]:
             # https://http2.github.io/http2-spec/#rfc.section.4.1
             # Implementations MUST ignore and discard any frame that has a type that is unknown.
             yield Log(f"Ignoring unknown HTTP/2 frame type: {event.frame.type}")
+        elif isinstance(event, h2.events.AlternativeServiceAvailable):
+            yield Log("Temporarily ignoring an Alt-Svc header")
```
mhils, line 279: Let's reduce this to DEBUG verbosity (most people will not care) and make the message a bit clearer.
```suggestion
            yield Log("Received HTTP/2 Alt-Svc frame, which will not be forwarded.", DEBUG)
```

mhils (commented):
Awesome, thanks for getting this fixed! 🍰 😃

mhils (approved):
Awesome, thanks for handling this! 🍰 🎉 🚀

PR #5904, opened 2023-02-02
Fix server addr issue in tls_passthrough example.

truebit:
#### Description
when peername returned None, the script would not working. use serveral alternatives for `server_address` variable.

#### Checklist

 - [*] I have updated tests where applicable.
 - [*] I have added an entry to the CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased: mitmproxy next
 
+* Fix a bug where peername would be None in tls_passthrough script, which would make it not working.
+  ([#5904](https://github.com/mitmproxy/mitmproxy/pull/5904), @truebit)
+
 * Add experimental QUIC support.
   ([#5435](https://github.com/mitmproxy/mitmproxy/issues/5435), @meitinger)
 * ASGI/WSGI apps can now listen on all ports for a specific hostname. 
--- a/examples/contrib/tls_passthrough.py
+++ b/examples/contrib/tls_passthrough.py
@@ -95,22 +95,26 @@ class MaybeTls:
         else:
             self.strategy = ConservativeStrategy()
 
+    @staticmethod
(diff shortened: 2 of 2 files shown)


mhils:
Could you give an example for where peername is None?

truebit:
Here below is an example log:

> 
> [20:47:34.790][172.19.199.235:58722] client connect
> [20:47:34.803] {'address': ('api.hk.deliveroo.com', 443), 'alpn': None, 'id': '8f3369ea-827b-4971-ae12-b36ef8b12e8f', 'ip_address': None, 'sni': None, 'source_address': None, 'timestamp_end': None, 'timestamp_start': None, 'timestamp_tcp_setup': None, 'timestamp_tls_setup': None, 'tls_established': False, 'tls_version': None, 'via': None, 'state': 0, 'error': None, 'tls': True, 'certificate_list': [], 'alpn_offers': (), 'cipher_name': None, 'cipher_list': (), 'via2': ('http', ('127.0.0.1', 7890))}
> [20:47:34.827][172.19.199.235:58714] client disconnect
> [20:47:34.828][172.19.199.235:58714] server disconnect 127.0.0.1:7890

where upstream proxy is http://127.0.0.1:7890

mhils:
Thanks for clarifying! I've added a note in the docs. If I understand things correctly, `server.address` is always set in this case, or do you have a falsifying example for this as well? :)

truebit:
I should clarify this thoroughly:
1. The current code is `server_address = data.context.server.peername`, which used peername for server address key.
2. in upstream proxy mode, `peername` would return `None`, that would make the tls mapping not working;
3. So I used code `server.peername or server.address or server.sockname` to get the first not None result for the `server_address`

truebit:
Nope, I only encountered the scenario mentioned in this PR :)

mhils (approved):
Thanks again! :cake:

PR #5919, opened 2023-02-08
Reformatted list

pradyotRanjan:
#### Checklist


Prinzhorn:
Thanks for looking into this! I think this could be done a lot simpler:

1. Use `enumerate()` instead of manually counting `i`
2. Compare with `len(data)` instead of iterating just to get `highest`
3. Add an actual test case with repeating numbers. I think the test you updated would've succeeded without your patch. You need to have a list like `[0,0,0]` to actually test repeated items.

pradyotRanjan:
That's great ! Doing the changes now.

mhils (approved):
Thanks! 🍰

PR #5923, opened 2023-02-09
console: fix bug that caused wrong direction indicators in message stream flow view

### LINKED ISSUE #5921: Wrong message direction indicator in console raw TCP view
#### Problem Description

In the flow view of the console tool, for raw TCP flows the `⇒` and `⇐` direction indicators (from client and to client) always alternate and don't represent the correct direction.

1. Start a TCP server: `ncat --listen localhost 2000`
2. Start mitmproxy: `mitmproxy --mode=reverse:tcp://localhost:2000@2001 --tcp-hosts='.*'`
3. Connect to the reverse proxy: `ncat localhost 2000`. Enter multiple lines, e.g. `a`, press enter,  `b`, press enter.
4. In mitmproxy, open the flow view for the new flow. There the two messages have different direction indicators even though they were both sent by the client.

#### System Information

```
Mitmproxy: 9.0.1 binary
Python:    3.11.0
OpenSSL:   OpenSSL 3.0.7 1 Nov 2022
Platform:  Linux-6.1.9-arch1-2-x86_64-with-glibc2.37
```

#### Fix

The bug is in the function [`view_message_stream`](https://github.com/mitmproxy/mitmproxy/blob/main/mitmproxy/tools/console/flowview.py#L248) and was introduced in https://github.com/mitmproxy/mitmproxy/pull/3970/commits/fe75f14ea29d0654c3a688ecc4c3669a946ea2a8. The solution could be the same as in the similar function [`view_websocket_messages`](https://github.com/mitmproxy/mitmproxy/blob/main/mitmproxy/tools/console/flowview.py#L197).

I would like to contribute a fix for this issue.

konradh:
#### Description

Fixes #5921. In the stream view for TCP and UDP flows in the console, the direction indicators (from client, to client) would always alternate and not represent the actual direction. This PR fixes that bug.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased: mitmproxy next
 
+* Fix a bug where the direction indicator in the message stream view would be in the wrong direction.
+  ([#5921](https://github.com/mitmproxy/mitmproxy/issues/5921), @konradh)
 * Fix a bug where peername would be None in tls_passthrough script, which would make it not working.
   ([#5904](https://github.com/mitmproxy/mitmproxy/pull/5904), @truebit)
 
--- a/mitmproxy/tools/console/flowview.py
+++ b/mitmproxy/tools/console/flowview.py
@@ -255,20 +255,16 @@ class FlowDetails(tabs.Tabs):
         viewmode = self.master.commands.call("console.flowview.mode")
 
         widget_lines = []
-
-        from_client = flow.messages[0].from_client
         for m in flow.messages:
             _, lines, _ = contentviews.get_message_content_view(viewmode, m, flow)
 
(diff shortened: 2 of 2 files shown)


mhils (commented):
Thanks for the PR! Pulling in `console` in a test is currently verry expensive in terms of test running time, and it doesn't contribute much to the result here. Could we easily set up a much simpler unit test for just `view_message_stream`? If that's not the case I'd also be happy with not adding a test here.

konradh:
I am not aware of an easy way of testing `view_message_stream` in isolation. I agree that the test doesn't add much value and removed it.

mhils (approved):
Awesome, thanks for handling this so swiftly! 🍰

PR #5953, opened 2023-02-26
fix: add the architecture determination when running the wireguard test

iamtwz:
#### Description

The wireguard test only check the system but didn't check the architecture.

For example, a non-x86 machine running Linux will fail the test.

Because the binary file is hardcoded in the test, I added the architecture check to skip the wireguard test on non-x86 machine.

#### Checklist


mhils (approved):
Awesome, thanks! 🍰

Issue #5973, opened 2023-03-07
Upstream proxy ignored in 10.0.0.dev0


anilatx:
#### Steps to reproduce the behavior:
1. Install current main version (10.0)
2a. set upstream to a closed port
3a. requests are fullfilled despite proxy being invalid
2b. set upstream to another MITM proxy
3b. the requests are not reaching the upstream proxy, but sent directly to the server

#### System Information
Paste the output of "mitmproxy --version" here.
10.0.0.dev0 # c7de03bc3679a8c3a90ad4d5767f450558908341

mhils:
Thanks for testing on main! `git bisect` shows this issue was introduced in https://github.com/mitmproxy/mitmproxy/pull/5967.

![image](https://user-images.githubusercontent.com/1019198/223689251-bd2e2c46-41bf-46ec-a455-0f351d8022c8.png)

mhils:
I linked the wrong issue in 884fd60e26b8beaa3785ea3d708c2d4386707f93, this has been fixed on main. Thanks again! 🍰

PR #5982, opened 2023-03-09
fix #5972

mhils:
Fix now, integration test later. This first needs some additional refactoring.


Prinzhorn:
For future reference it was https://github.com/mitmproxy/mitmproxy/issues/5973 not #5972

PR #5983, opened 2023-03-10
added metadata view in mitm\web

### LINKED ISSUE #5704: Addon: add the possibility to add text/metadata to a connection
#### Problem Description
I modify the flows with a addon, but something the server returns 403, so I want to debug the flows by attaching a text, a comment or a data and see it in the web gui.   

#### Proposal
Create a API `HTTPFlow.setComment(text: str)` that set a text in the flow info, and in the webgui add a tab for seeing the text. 

```
class ModifyRequest:
    def request(self, flow: HTTPFlow):
          flow.setComment("This request was modified by the addon")
```

#### Alternatives
Use ctx.log.info more often

#### Additional context
None

RG7279805:
#### Description
closes #5704

- Added metadata viewer in mitm\web.
#### Checklist

I don't think any test is required in this case

### DIFF
--- a/mitmproxy/flow.py
+++ b/mitmproxy/flow.py
@@ -126,7 +126,7 @@ class Flow(serializable.Serializable):
         self._backup: Flow | None = None
         self.marked: str = ""
         self.is_replay: str | None = None
-        self.metadata: dict[str, Any] = dict()
+        self.metadata: Any = None
         self.comment: str = ""
 
     __types: dict[str, type[Flow]] = {}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,8 @@
   ([#5904](https://github.com/mitmproxy/mitmproxy/pull/5904), @truebit)
 * mitmproxy now requires Python 3.10 or above.
   ([#5954](https://github.com/mitmproxy/mitmproxy/pull/5954), @mhils)
-
+* Added metadata view in mitm web.
+  [#5704](https://github.com/mitmproxy/mitmproxy/issues/5704)
 ### Breaking Changes
 
 * The `onboarding_port` option has been removed. The onboarding app now responds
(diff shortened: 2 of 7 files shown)


Prinzhorn:
I assume you or the editor you're using were doing some automatic code formatting. This makes it hard to focus on the actual changes. We should definitely clean this up at some point (e.g. mixed single/double quotes or missing spaces) but it would be better not to have these changes in this PR.

RG7279805:
I tried to have the proper formatting for web but there is no formatter config available 
Maybe in the web we can use prettier for web

Prinzhorn:
I've opened https://github.com/mitmproxy/mitmproxy/issues/5984 to fix this in the future. For now I recommend disabling any formatting your editor does. That's a recommendation I would give in general when contributing to a project. Formatting changes make PRs unnecessarily noisy. Unrelated code should not be touched.

RG7279805:
Ok i will submit make the changes and push it

RG7279805:
[ignore: to make it easy to compare changes](https://github.com/mitmproxy/mitmproxy/pull/5983/commits/ed064018bd293f1f692c3efb2c75655da913b920)
made this commit to revert the changes to original formatting 

[changes made in mitm web](https://github.com/mitmproxy/mitmproxy/pull/5983/commits/d5a3fab1c90d4c2488bf1c5bb55e641bdc1abea6)
made the changes without formatting

Prinzhorn:
Perfect, I think that's good for a first iteration. Two small changes:

1. We should allow any type of metadata, so I'd check for `null` (`None` in Python) and throw everything else into `JSON.stringify`. Fun fact, strings are already rendered because `Object.keys('foo')` is `[ '0', '1', '2' ]`
2. Indent the JSON, e.g. `JSON.stringify(flow.metadata, null, 4)`

RG7279805:
@Prinzhorn 
i have messed up my fork by clicking on discard button when updating my fork 
which closed this PR do you mind opening it again
I have made the changes you asked in 1st iteration

mitmproxy/flow.py
```diff
@@ -126,7 +126,7 @@ def __init__(
         self._backup: Flow | None = None
         self.marked: str = ""
         self.is_replay: str | None = None
-        self.metadata: Any = None
```
RG7279805: if i change this to a lot of test start failing so should a leave metadata as empty `dict` or it must initialized with `None`
```
self.metadata: Any = None
```

PR #5985, opened 2023-03-11
Add prettier to mitmweb

### LINKED ISSUE #5984: Add prettier to mitmweb
This came up in https://github.com/mitmproxy/mitmproxy/pull/5983 . I didn't realized how inconsistently formatted the mitmweb sources are.

Examples:

Mixed single/double quotes

https://github.com/mitmproxy/mitmproxy/blob/884fd60e26b8beaa3785ea3d708c2d4386707f93/web/src/js/components/FlowView.tsx#L36-L49

Inconsistent indentation:

https://github.com/mitmproxy/mitmproxy/blob/884fd60e26b8beaa3785ea3d708c2d4386707f93/web/src/js/flow.ts#L106-L115

https://prettier.io/ is the de facto standard nowadays and I think we should just let it run over the entire code once and then add it to autofix. Objections @mhils ?

alexgershberg:
#### Description

Closes #5984 

This simple change adds prettier as a dev dependency, a .prettierignore file and a change to autofix.ci to run a prettier check.

#### Checklist


alexgershberg:
I'm not sure what kind of environment GitHub actions run prettier under, not sure if I should call to it using `npx` or `prettier` directly. 
Also, this would also require a commit with linting changes (as CI check will currently fail), but I might do that after an initial code review.

mhils:
Awesome, thanks!

Is npx installing the prettier version from package-lock.json, or is it just getting whatever is latest? I would really like to pin the version in CI to avoid random breakage on prettier updates.

web/src/js/ducks/_options_gen.ts
```diff
@@ -1,94 +1,94 @@
...
-    view_order: string
-    view_order_reversed: boolean
-    web_columns: string[]
-    web_debug: boolean
-    web_host: string
-    web_open_browser: boolean
-    web_port: number
-    web_static_viewer: string | undefined
-    websocket: boolean
+    add_upstream_certs_to_client_chain: boolean;
```
mhils, line 3: This here is a problem because that file is autogenerated when running the Python tests. We need to adjust the code that generates it so that they don't flipflop.

mhils (commented):
web/src/js/filt/*.js probably wants to be excluded as well. :)

alexgershberg:
> Awesome, thanks!
> 
> Is npx installing the prettier version from package-lock.json, or is it just getting whatever is latest? I would really like to pin the version in CI to avoid random breakage on prettier updates.
> 
> Seems like one of the snapshot tests needs to be updated after pretty-printing. :)

Yeah, npx will be getting the latest prettier, I'm really unfamiliar with GitHub actions, so I don't know what kind of environment npx runs in (I'm kind of surprised npx just works out of the box). My CI experience is from Azure, and we have to do an explicit environment set-up there (install node & any other binaries you want to use in your tasks).

 I'll do a bit of reading, but I'm open to suggestions :)

web/src/js/ducks/_options_gen.ts
alexgershberg: > This here is a problem because that file is autogenerated when running the Python tests. We need to adjust the code that generates it so that they don't flipflop.
Weird question, but why would we source control an auto-generated file? 

I'll take a look at where this gets generated, but might take me some time as I'm still coming to grips with he repo.

alexgershberg:
@mhils Making python tests generate valid formatted JS can be an issue of it's own I think. I've spent some time looking into it, but I don't think I have the time available to do it properly, at this time. Besides, I didn't really realise that it was in the scope of this issue.

My suggestion: Ignore these files for now, capture a new issue to re-write the python tests to generate proper JS. 

But do tell me, how you want me to proceed with this.

mhils:
> Yeah, npx will be getting the latest prettier, I'm really unfamiliar with GitHub actions, so I don't know what kind of environment tasks run in (I'm kind of surprised npx just works out of the box). My CI experience is from Azure, and we have to do an explicit environment set-up there (install node & any other binaries you want to use in your tasks).

GitHub CI is Azure Pipelines under the hood, so you can assume most things are the same. The GHA runner just has more stuff preloaded I think? Back to the question, I guess the most straightforward approach would be to run `npm ci` first and then call the prettier binary directly. :)

> My suggestion: Add these files to .prettierignore, capture a new issue to re-write the python tests to generate proper JS.

I'm happy with two approaches:

1. Fix the file generation - from the diff it looks like we only need to add semicolons and that's it?
2. Add it to `.prettierignore` and that's it. I don't think we should create an issue for this, realistically speaking there is no reason to believe someone wants to tackle this in the future if you don't want to do it now.

web/src/js/ducks/_options_gen.ts
mhils: > Weird question, but why would we source control an auto-generated file?

Because this way we can run the web tests without pulling in and setting up an entire Python environment first.

alexgershberg:
> I'm happy with two approaches:
> 
> 1. Fix the file generation - from the diff it looks like we only need to add semicolons and that's it?
> 2. Add it to `.prettierignore` and that's it. I don't think we should create an issue for this, realistically speaking there is no reason to believe someone wants to tackle this in the future if you don't want to do it now.


It's not that I don't want to do it, It would just require a lot more work. 😄 


There are 3 test files which require changing:
`mitmproxy/web/src/js/ducks/_options_gen.ts` - Straightforward, just add the semicolons.

`mitmproxy/web/src/js/__tests__/ducks/_tflow.ts` 
`mitmproxy/web/src/js/__tests__/ducks/_tbackendstate.ts`
The two above are trickier, as they parse their respective data (python dict), and output a JSON, which then just gets written into a javascript file.

I'm not going to go into exhaustive detail of what's different between the JSON and the JS code prettier expects. Long story short, `json.dumps()` doesn't give us the formatting we need.

I have tried replacing `json.dumps()` with `json5.dumps()` from JSON5 library. That has worked, and gives us a result closer to what we want, but still not quite. 
I have searched far and wide to output the format we want, but I don't see any libraries that would do that for us. Writing my own implementation for formatting just seems excessive. 

I think at this point, I will push the remaining changes I have, and add the problematic tests to .prettierignore.

mhils:
Ah yes - I was only looking at the `_options_gen.ts` diff, which I guess was the trivial part where just fixing it is the right approach. I think we can generously put the rest into prettierignore :)

alexgershberg:
> Ah yes - I was only looking at the `_options_gen.ts` diff, which I guess was the trivial part where just fixing it is the right approach. I think we can generously put the rest into prettierignore :)

I've also been meaning to ask, shouldn't we just put `prettier --check` into `main.yml` ci? I think tox linting task runs as part of that CI.
(Note the difference between `prettier --check` and `prettier --write`)
I know the issue mentioned putting it into autofix.ci. 

In my opinion, contributors should be fixing linting errors before committing, rather than CI doing that automatically + pushing a commit on top of their changes. I'm happy to be proven wrong, just want to double-check. :)

mhils:
> I've also been meaning to ask, shouldn't we just put `prettier --check` into `main.yml` ci? I think tox linting task runs as part of that CI. (Note the difference between `prettier --check` and `prettier --write`) I know the issue mentioned putting it into autofix.ci.

The autofix.ci task already updates the pull request if it's not formatted correctly, so adding a second check into `main.yml` feels redundant. I do however think it would be good to add a note about prettier in `web/README.md`.

> In my opinion, contributors should be fixing linting errors before committing, rather than CI doing that automatically + pushing a commit on top of their changes. I'm happy to be proven wrong, just want to double-check. :)

I think that makes sense for a project with a stable set of contributors, but that approach does not work as well for FOSS. Yes, ideally everyone does format their code before sending a pull request, but in practice we regularly had pull requests that were blocked on failing CI due insignificant minutiae, which is highly annoying. Having these things auto-fixed by CI is much easier. We squash small PRs anyways, so it does not end up in the git history. :)

alexgershberg:
> > I've also been meaning to ask, shouldn't we just put `prettier --check` into `main.yml` ci? I think tox linting task runs as part of that CI. (Note the difference between `prettier --check` and `prettier --write`) I know the issue mentioned putting it into autofix.ci.
> 
> The autofix.ci task already updates the pull request if it's not formatted correctly, so adding a second check into `main.yml` feels redundant. I do however think it would be good to add a note about prettier in `web/README.md`.
> 
> > In my opinion, contributors should be fixing linting errors before committing, rather than CI doing that automatically + pushing a commit on top of their changes. I'm happy to be proven wrong, just want to double-check. :)
> 
> I think that makes sense for a project with a stable set of contributors, but that approach does not work as well for FOSS. Yes, ideally everyone does format their code before sending a pull request, but in practice we regularly had pull requests that were blocked on failing CI due insignificant minutiae, which is highly annoying. Having these things auto-fixed by CI is much easier. We squash small PRs anyways, so it does not end up in the git history. :)


Alright! Thank you for all of your help, I really appreciate it.

alexgershberg:
That's all the changes I had, let me know if I have missed anything!

mhils (approved):
Awesome, thanks! 🍰

alexgershberg:
> Awesome, thanks! 🍰

This was my first ever PR to an open source project, I will now be forever chasing this high of having my pull request approved.

mhils:
You're welcome back to seek that any time! 😃

PR #5986, opened 2023-03-13
Add a prettier npm script

Prinzhorn:
I prefer that over `npx`. You can also locally run `npm run prettier` to update all files without knowing the cli arguments for prettier.


Prinzhorn:
I also updated the README, since this has nothing to do with linting (we can look into integrating eslint at some point too).

mhils (approved):
Thanks :)

PR #5989, opened 2023-03-15
Fix #5910 - Spoof ICMP echo replies

### LINKED ISSUE #5910: WireGuard Mode: Spoof ICMP echo replies
Some apps check network connectivity by sending ICMP pings. ICMP traffic is currently swallowed by mitmproxy_rs, which makes them believe that there is no network connectivity.

Correctly forwarding ICMP in both directions is tricky, so a cheap mitigation that probably covers the majority of use cases would be to simply fake ICMP echo replies [somewhere around here](https://github.com/mitmproxy/mitmproxy_rs/blob/459bd9c1da29378b14e7f0230705c892f89668d1/src/network/task.rs#L68).

jetvova:
#### Description

Important: this depends on https://github.com/mitmproxy/mitmproxy_rs/pull/28, which has to be merged first due to interface change.

Some apps check network connectivity by sending ICMP pings. Before this fix, ICMP traffic was swallowed by mitmproxy_rs, which makes them believe that there is no network connectivity. 

Mitmproxy now generates fake ICMP Echo Replies in response to ICMP Echo Requests, and transmits them back to the sender's address. 

#### Checklist

### DIFF
--- a/mitmproxy/proxy/mode_servers.py
+++ b/mitmproxy/proxy/mode_servers.py
@@ -258,7 +258,6 @@ class ServerInstance(Generic[M], metaclass=ABCMeta):
         src_addr: str,
         dst_addr: str,
     ) -> None:
-
         # Some apps check network connectivity by sending ICMP pings. ICMP traffic is currently
         # swallowed by mitmproxy_rs, which makes them believe that there is no network connectivity.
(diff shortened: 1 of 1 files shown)


jetvova:
Closing PR in favor of implementing a more simple approach as advised.

PR #6004, opened 2023-03-21
Script unloading fix - mitm-web

### LINKED ISSUE #6002: Script in mitm-web won't unload
#### Problem Description
We can't remove a loaded script in running mitm-web.
The scripts option for executing a script with mitm-web seems to have a bug.

Note: If we specify multiple scripts to be executed, then if we intend to remove them, one script remains present and cant be unloaded

1. In a running mitm-web instance, go to Options > Scripts > add path to script.
2. Try removing, the script, it won't remove. It simply says "no such script" even when removing.
3. Add more scripts. All can be removed except one.

Applying script
chrome_Qy0rYZw6QZ

Error: No such script, on removing it
chrome_tIZ1HoQ2Km

Should be removed like this: no errors
chrome_4FAF27syxE

#### System Information
Mitmproxy: 9.0.1 binary
Python:    3.11.0
OpenSSL:   OpenSSL 3.0.7 1 Nov 2022
Platform:  Windows-10-10.0.22621-SP0

binbard:
#### Description

Fixed scripts not unloading in mitm-web.

Resolves #6002 

#### Checklist


mhils:
Thanks - I believe this needs to be fixed in mitmweb's options editor, we should continue to raise on an empty string in the Python world.

PR #6021, opened 2023-03-26
Flow multi select to delete

### LINKED ISSUE #3313: [Feature Request] Save filtered flows only in mitmweb
In mitmweb it is able to filter out flows with some search criteria, but it still saves all the flows when clicking on "save" button.  So I have to delete unwanted flows one by one and then save.

It would be great if we can multi-select flows to delete/save.

Rishabhg71:
#### Description
closes #3313
Added support to have multiple flows selected at the same time by pressing CTRL + onClick event on flow row

#### Checklist


RG7279805:
Give me green light on code and functionality and I will update the tests

PR #6038, opened 2023-04-01
Fixed a bug with auto scroll in mitmweb

y52en:
#### Description
Fixed a bug that could prevent mitmweb autoscroll from working properly.
This bug was caused by the fact that clientHeight and scrollHeight returned integers, but scrollTop returned a decimal number.
[https://drafts.csswg.org/cssom-view/#extension-to-the-element-interface](https://drafts.csswg.org/cssom-view/#extension-to-the-element-interface)

screenshot-2023-04-01-143309

#### Checklist


mhils (approved):

mhils:
Thanks!

PR #6080, opened 2023-04-24
DockerHub-README.md: reword ambiguous command

vadcx:
#### Description

Being unfamiliar with the docker options, I thought the square brackets could be part of the command.

Instead, they refer to the sentence *after* the command, that is, they meant "optionally".

To remove the ambiguity, I put it onto different lines with an explanation.

#### Checklist

I think neither are applicable?

### DIFF
--- a/release/docker/DockerHub-README.md
+++ b/release/docker/DockerHub-README.md
@@ -4,9 +4,13 @@ Containerized version of [mitmproxy](https://mitmproxy.org/): an interactive, SS
 
 ## Usage
 
+To launch the terminal user interface of mitmproxy:
+
 ```sh
-$ docker run --rm -it [-v ~/.mitmproxy:/home/mitmproxy/.mitmproxy] -p 8080:8080 mitmproxy/mitmproxy
-[terminal user interface of mitmproxy is launched...]
+$ docker run --rm -it -p 8080:8080 mitmproxy/mitmproxy
+
+# or start it with the volume mounted to your home directory:
+$ docker run --rm -it -v ~/.mitmproxy:/home/mitmproxy/.mitmproxy -p 8080:8080 mitmproxy/mitmproxy
 ```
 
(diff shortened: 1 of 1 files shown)


mhils:
Thanks. That's a great explanation why the current version is bad. 

Invoking mitmproxy without the shared volume can lead to UX problems as well (new CA generated on restart), so maybe we should just remove the square brackets and keep `-v ...` in the first invocation?

vadcx:
(limited understanding because I've actually used mitmproxy for the first time ever): There are only two usage scenarios, right?

1. Use the built-in TUI (sufficed for me)
2. Get the TLS certs for use in Wireshark

If so then they really are equally important, I'd keep the -v command but change the explanation to point it out:

> The `-v` for *volume mount* is optional: It allows to store the generated CA certificates between runs and for you to access them.

Feel free to edit/push without me if it's less hassle.

vadcx:
From commit message:

> - only leave the command with -v as an example
> - grammar
> - httpS link

I hope it's alright now? I really appreciate the communication (we seem to agree anyway), but if there's anything else to do, I'd rather leave it to you. The overhead of discussing a couple of changed readme lines is growing absurdly high :)

mhils (approved):
Looks awesome now, thanks again! :cake:

> The overhead of discussing a couple of changed readme lines is growing absurdly high :)

I kind of agree, but then again it's really valuable for us to have stuff like this written by someone new to the project. I wouldn't have thought of adding something like "To launch the terminal user interface of mitmproxy:", but in retrospect it makes total sense! So it's actually good to spend a few more cycles here then usual. 😃

vadcx:
I see, makes total sense. I've spent a fair share of time adding/writing documentation myself. And while I think I know what to write for a clueless user, it still helps to completely forget the topic of discussion and to look at what you've written with a fresh eye. Cheers!

Issue #4748, opened 2021-08-12
Syntax error in documented config file


ghost:
I'm trying [this](https://docs.mitmproxy.org/stable/howto-transparent/#work-around-to-redirect-traffic-originating-from-the-machine-itself-1) work-around to get the proxy in transparent mode to redirect traffic originating from the machine itself correctly. However, the configuration file gives the following error when executing `sudo pfctl -f pf.conf`:

```
pfctl: Use of -f option, could result in flushing of rules
present in the main ruleset added by the system at startup.
See /etc/pf.conf for further details.

No ALTQ support in kernel
ALTQ related functions disabled
pf.conf:17: syntax error
pfctl: Syntax error in config file: pf rules not loaded
```

**System Information**
```
Mitmproxy: 7.0.2
Python:    3.9.6
OpenSSL:   OpenSSL 1.1.1k  25 Mar 2021
Platform:  macOS-11.4-x86_64-i386-64bit
```

kikuyan:
Are you sure you needed to append a blank line?
I guess the last line simply didn't end with newline character.
- without newline character, get syntax error
```
$ xxd pf.conf | tail -n 2
00000200: 2075 7365 7220 7b20 213d 2024 7470 726f   user { != $tpro
00000210: 7879 5f75 7365 7220 7d                   xy_user }
$ sudo pfctl -f pf.conf
pfctl: Use of -f option, could result in flushing of rules
present in the main ruleset added by the system at startup.
See /etc/pf.conf for further details.

No ALTQ support in kernel
ALTQ related functions disabled
pf.conf:17: syntax error
pfctl: Syntax error in config file: pf rules not loaded
$ 
```
- append newline character, no error
```
$ echo >>pf.conf 
$ xxd pf.conf | tail -n 2
00000200: 2075 7365 7220 7b20 213d 2024 7470 726f   user { != $tpro
00000210: 7879 5f75 7365 7220 7d0a                 xy_user }.
$ sudo pfctl -f pf.conf
pfctl: Use of -f option, could result in flushing of rules
present in the main ruleset added by the system at startup.
See /etc/pf.conf for further details.

No ALTQ support in kernel
ALTQ related functions disabled
$ 
```

PR #6091, opened 2023-04-29
Update file howto-transparent.md as suggested in issue #4748

ThomasBucaioni:
#### Description

File `pf.conf` needs to end with a blank newline for the configuration to be valid. Update the doc according to issue #4748

### DIFF
--- a/docs/src/content/howto-transparent.md
+++ b/docs/src/content/howto-transparent.md
@@ -276,7 +276,7 @@ tproxy_user = "nobody"
 rdr pass proto tcp from any to any port $redir_ports -> $tproxy
 pass out route-to (lo0 127.0.0.1) proto tcp from any to any port $redir_ports user { != $tproxy_user }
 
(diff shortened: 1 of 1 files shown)


docs/src/content/howto-transparent.md
```diff
@@ -275,6 +275,9 @@ tproxy_user = "nobody"
 
 rdr pass proto tcp from any to any port $redir_ports -> $tproxy
 pass out route-to (lo0 127.0.0.1) proto tcp from any to any port $redir_ports user { != $tproxy_user }
+
+# End the file with a blank newline /!\
```
mhils, line 279:
```suggestion
# End the file with a blank newline
```
I'm not too familiar with pf.conf syntax, but does the backslash at the end cancel out the newline? In either case, I don't want to think about it, so let's just remove it.

mhils (approved):
Thanks!

Issue #3752, opened 2019-12-20
Issues with --set


YuriyTigiev:
#### Problem Description
The option --set doesn't work

#### Steps to reproduce the behavior:
1. The command mitmdump.exe -p 8088 --set ssl_version_client=all --set ssl_version_server=all doesn't change configuration
2. mitmdump.exe -p 8088 --options doesn't show changes  
3. mitmdump.exe --options returns  Error starting proxy server: OSError(98, 'Address already in use')

4. A default config.yaml file is not present for linux and windows versions. 
5. Documentation doesn't contain samples of how to properly use --set options with one or a few parameters. 

#### System Information
Paste the output of "mitmproxy --version" here.

(venv) [root@localhost mitmproxy]# mitmproxy --version
Mitmproxy: 6.0.0.dev (+494, commit 67955fa)
Python:    3.6.8
OpenSSL:   OpenSSL 1.1.0j  20 Nov 2018
Platform:  Linux-3.10.0-1062.9.1.el7.x86_64-x86_64-with-centos-7.7.1908-Core

C:\Program Files (x86)\mitmproxy\bin>mitmdump --version
Mitmproxy: 4.0.4 binary
Python:    3.6.6
OpenSSL:   OpenSSL 1.1.0i  14 Aug 2018
Platform:  Windows-10-10.0.17763

mhils:
1. It does change the option for me.

2. This is expected behavior. See `mitmproxy --help`: 

```
--options             Show all options and their default values
```

3. You already have a proxy server running.

4. Options are documented at https://docs.mitmproxy.org/stable/concepts-options/. We're using YAML. The output of `--mitmproxy --options` incidentally gives you the default configuration.

5. Your usage of `--set` seems fine, so I'm not sure what's the issue here.

YuriyTigiev:
@mhils  thank you for your attention

3) No. I have restarted a notebook (windows 10) and run the program the first time after restart. The same issue was on the centos server. But after a few tries now everything ok.

4) I saw the configuration, but It was not clear how to change it.  I didn't find the source config file.
If is it possible to add the file to the install package.

5) The page https://docs.mitmproxy.org/stable/concepts-options/ contains detail information which options admins can use for the run program. If is it possible to add a few examples of how to use these options from the command line below the options' descriptions.  It will be really helpful to who first time trying to use this software.

PR #6094, opened 2023-04-30
Add example, according to issue #3752

ThomasBucaioni:
#### Description

Add a `--set` option example, as suggested in issue #3752

### DIFF
--- a/docs/src/content/concepts-options.md
+++ b/docs/src/content/concepts-options.md
@@ -34,7 +34,10 @@ persistent by saving the settings out to a YAML configuration file (please see
 the specific tool's interactive help for details on how to do this).
 
 For all tools, options can be set directly by name using the `--set`
-command-line option. Please see the command-line help (`--help`) for usage.
+command-line option. Please see the command-line help (`--help`) for usage. Example on Windows:
+```
+mitmdump.exe -p 8088 --set ssl_version_client=all --set ssl_version_server=all
+```
 
 ## Available Options
 


docs/src/content/concepts-options.md
```diff
@@ -34,7 +34,10 @@ persistent by saving the settings out to a YAML configuration file (please see
 the specific tool's interactive help for details on how to do this).
 
 For all tools, options can be set directly by name using the `--set`
-command-line option. Please see the command-line help (`--help`) for usage.
+command-line option. Please see the command-line help (`--help`) for usage. Example on Windows:
+```
+mitmdump.exe -p 8088 --set ssl_version_client=all --set ssl_version_server=all
```
mhils, line 39: Can we please make this not specific to a singel platform, and keep it as concise as possible (i.e. either demonstrate a single option, or demonstrate a single list option with two settings)?

ThomasBucaioni: For example:
```
mitmweb --set  anticomp=true
mitmproxy --set  allow_hosts=[a.b.c.d, e.f.g.h]
```
?

ThomasBucaioni: Corrected. Maybe `<IPv4-1>`, `<IPv4-2>`, `<IPv6-1>`, `<IPv6-2>` would be better, just to say we can mix both?

mhils (approved):
Awesome, thanks! 🍰

PR #6127, opened 2023-05-14
solve #4272 docs bug

### LINKED ISSUE #4272: `view.flows.add` command does not exist but the examples reference it
#### Problem Description

The `view.flows.add` command does not exist but the example `duplicate-modify-replay.py` shows this command being used.

`replay.client` seems to perform both the "add to view" and "replay" function.

emanuele-em:
#### Description

close #4272 with a small docs fix


mhils (approved):
🙌

PR #6170, opened 2023-06-13
Support for selecting an elliptic curve for key exchange

Mike-Ki-ASD:
#### Description

Added option to specify an elliptic curve for key exchange between mitmproxy <-> server

#### Checklist


mitmproxy/net/tls.py
```diff
@@ -106,6 +107,13 @@ def _create_ssl_context(
     # Options
     context.set_options(DEFAULT_OPTIONS)
 
+    # ECDHE for Key exchange
+    if ctx.options.curve is not None:
```
mhils, line 111: The return values of methods in `net/tls.py` should depend on their arguments only, not on externally-defined variables/options. This needs to be accessed in tlsconfig.py and passed in as an argument.

mitmproxy/options.py
```diff
@@ -74,6 +74,13 @@ def __init__(self, **kwargs) -> None:
             None,
             "Set supported ciphers for mitmproxy <-> server connections using OpenSSL syntax.",
         )
+        self.add_option(
+            "curve",
+            Optional[str],
+            None,
+            "Specify an elliptic curve for ECDHE key exchange between mitmproxy <-> server connections using OpenSSL "
+            "syntax."
```
mhils, line 82: Add a curve example here?

mhils (commented):
Thanks for the PR! :)

mitmproxy/options.py
Mike-Ki-ASD: I added 3 example curve, however I was not sure whether I should specify the command `openssl ecparam -list_curve`. This command lists the supported curves, at least for Linux.

mhils (approved):
Thanks again! 🍰

PR #6188, opened 2023-06-23
grpc: enable decompressing zstd responses

tsaaristo:
#### Description

Adds zstd to valid gRPC encoding schemes. Spotted some `grpc-encoding: zstd` in the wild, and this trivial fix was found working.

#### Checklist

   (I assume existing gzip test is acceptable, we're relying on `mitmproxy.net.encoding.decode()` anyway)


mhils (approved):
Thanks for the PR! The CI failures are unrelated (https://github.com/mitmproxy/mitmproxy/pull/6189), this looks good as-is. 😃

PR #6190, opened 2023-06-23
first draft of read har.

stanleygvi:
#### Description

Added readhar to examples/contrib. Reads har files into mitmproxy.


examples/contrib/readhar.py
```diff
@@ -0,0 +1,134 @@
...
+from mitmproxy import http
+from mitmproxy import types
+
+
+class ReadHar:
+    def __init__(self):
+        self.harFile = {}
+        self.flows = []
+
+    def fix_headers(self, request_headers):
```
mhils, line 21: Please add type annotations for all methods. Those are super important to catch errors early. :)

examples/contrib/readhar.py
```diff
@@ -0,0 +1,134 @@
...
+        """Changes provided headers into flow-friendly format"""
```
mhils, line 22: Nit: This method does not _change_ the passed headers, it converts/transforms/... the passed headers. The passed value remains unchanged.

"flow-friendly format" is a bit nondescriptive. Let's state right away that it converts into a list of `(b"header-name", b"header-value")` tuples?

examples/contrib/readhar.py
```diff
@@ -0,0 +1,134 @@
...
+        flow_headers = []
+        for header in request_headers:
+            if type(header) == dict and header.get("name", None):
+                key = header["name"]
+                value = header["value"]
+            else:
```
mhils, line 28: Based on the code I assume that there are HAR files with `{name: ..., value: ...}` dicts as well as HAR files with `[name, value]` arrays? For future readers it's useful to add a rough comment explaining the context. For example, if all HAR files but Firefox ones use the former notation, a `# Firefox's HAR files deviate by doing X...`-style comment would be useful.

examples/contrib/readhar.py
```diff
@@ -0,0 +1,134 @@
...
+class ReadHar:
+    def __init__(self):
+        self.harFile = {}
+        self.flows = []
+
+    def fix_headers(self, request_headers):
+        """Changes provided headers into flow-friendly format"""
+        flow_headers = []
+        for header in request_headers:
+            if type(header) == dict and header.get("name", None):
```
mhils, line 25: Nit: `isinstance(header, dict)` is slightly better here. Two reasons:

1. `type(header) == dict` does not work for dict subclasses. That's not really relevant here at all, but you can avoid any problems in that regard by just consistently using `isinstance`.
2. The `header.get("name", None)` does not save things I believe. If there is no name attribute, `header[0]` in line 29 will also just crash as well.

examples/contrib/readhar.py
```diff
@@ -0,0 +1,134 @@
...
+            flow_headers.append((bytes(key, "utf-8"), bytes(value, "utf-8")))
+
+        return flow_headers
+
+    def request_to_flow(self, request_json):
+        """
+        Creates a HTTPFlow object from a given entry in HAR file
+        """
+
+        requestMethod = request_json["request"].get("method", "")
```
mhils, line 40: Are there cases where "method" is not present? If yes, let's document that (see comment above). If no, we should probably just do `request_json["request"]["method"]`.

examples/contrib/readhar.py
```diff
@@ -0,0 +1,134 @@
...
+                value = header[1]
+            flow_headers.append((bytes(key, "utf-8"), bytes(value, "utf-8")))
+
+        return flow_headers
+
+    def request_to_flow(self, request_json):
+        """
+        Creates a HTTPFlow object from a given entry in HAR file
+        """
+
```
mhils, line 39: For variables and method names, please use `snake_case` convention (see https://peps.python.org/pep-0008/#function-and-variable-names). :)

examples/contrib/readhar.py
```diff
@@ -0,0 +1,134 @@
...
+        requestUrl = request_json["request"].get("url", None)
+        if not requestUrl:
+            requestUrl = request_json["request"].get("host", None)
+
+        requestHeaders = self.fix_headers(request_json["request"].get("headers", []))
+
+        client_conn = connection.Client(
+            peername=("127.0.0.1", 51513),
+            sockname=("127.0.0.1", 8080),
+            timestamp_start=time.time(),
```
mhils, line 50: HAR files have some sort of time information, so let's add a `TODO` comment here to revisit later that we insert something more correct here.

examples/contrib/readhar.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+        server_conn = connection.Server(address=None)
```
mhils, line 52: We should set a server address, no?

examples/contrib/readhar.py
```diff
@@ -0,0 +1,134 @@
...
+
+        newflow = http.HTTPFlow(client_conn, server_conn)
+        try:
+            newflow.request = http.Request.make(
+                requestMethod, requestUrl, "", requestHeaders
+            )
+        except:
```
mhils, line 59: Two notes on style:

1. Bare except statements are evil in Python, see https://www.flake8rules.com/rules/E722.html. If unsure about the error `except Exception` is better to avoid catching `KeyboardInterrupt`. See https://docs.python.org/3/library/exceptions.html#exception-hierarchy.

examples/contrib/readhar.py
```diff
@@ -0,0 +1,134 @@
...
+
+        client_conn = connection.Client(
+            peername=("127.0.0.1", 51513),
+            sockname=("127.0.0.1", 8080),
+            timestamp_start=time.time(),
+        )
+        server_conn = connection.Server(address=None)
+
+        newflow = http.HTTPFlow(client_conn, server_conn)
+        try:
```
mhils, line 55: Practically speaking I would let this method either return a correct HTTPFlow or have it raise an error. You can catch that then below at a single place.

examples/contrib/readhar.py
```diff
@@ -0,0 +1,134 @@
...
+            newflow.request = http.Request.make(
+                requestMethod, requestUrl, "", requestHeaders
+            )
+        except:
+            logging.error("Unable to create flow, please change to a valid HAR format")
```
mhils, line 60: Please use the `logger = logging.getLogger(__name__)` pattern, see https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial.

examples/contrib/readhar.py
```diff
@@ -0,0 +1,134 @@
...
+                        "Unable to read HAR file. Please provide a valid HAR file"
+                    )
+                    return
+
+        for request_json in self.harFile["log"]["entries"]:
+            flow = self.request_to_flow(request_json)
+            if flow:
+                self.flows.append(flow)
+        self.running()
+        self._read_har_task
```
mhils, line 131: This here can be greatly simplified I think. You can just do this here instead and avoid the functions above:
```suggestion
        async def load_flows() -> None:
            for flow in flows:
                await ctx.master.load_flow(flow)
        asyncio.create_task(load_flows())
```

The main underlying problem here is that commands cannot be async yet, otherwise we could just call `await ctx.master.load_flow(flow)` directly.

examples/contrib/readhar.py
```diff
@@ -0,0 +1,134 @@
...
+
+        Args:
+            path (types.Path): Path to HAR file
+        """
+        with open(path) as fp:
+            try:
+                self.harFile = json.load(fp)
+            except:
+                try:
+                    json.load(codecs.open(path, "r", "utf-8-sig"))
```
mhils, line 119: Are we really dealing with utf-8-sig for HAR files? Normal utf8 does not work? (Again this probably needs documentation).

In either case, this is currently not functioning because the result is not used anywhere.

examples/contrib/readhar.py
```diff
@@ -0,0 +1,134 @@
...
+    ) -> None:
+        """
+        Reads a HAR file into mitmproxy. Loads a flow for each entry in given HAR file.
+
+        Args:
+            path (types.Path): Path to HAR file
+        """
+        with open(path) as fp:
+            try:
+                self.harFile = json.load(fp)
```
mhils, line 116: Let's try and avoid all object state - we can keep `harFile` as a local variable.

mhils (commented):
Good stuff! Lots of feedback below, please extrapolate from my comments to the rest of your code! 😃

stanleygvi:
Thanks! Your feedback was super helpful! Here is the updated version of my code.

examples/contrib/readhar.py
```diff
@@ -0,0 +1,141 @@
...
+        for header in request_headers:
+            # Applications that use the {"name":item,"value":item} notation are Brave, Chrome, Edge, Firefox, Charles, Fiddler, Insomnia, Safari
+            if isinstance(header, dict):
+                try:
+                    key = header["name"]
+                    value = header["value"]
+                except KeyError as e:
+                    raise exceptions.OptionsError(str(e)) from e
+
+            # Application that use the [name, value] notation is Slack
```
mhils, line 35: Possibly stupid question: How do you create a HAR file in Slack? :)

examples/contrib/readhar.py
```diff
@@ -0,0 +1,141 @@
...
+        """Converts provided headers into (b"header-name", b"header-value") tuples"""
+        flow_headers = []
+        for header in request_headers:
+            # Applications that use the {"name":item,"value":item} notation are Brave, Chrome, Edge, Firefox, Charles, Fiddler, Insomnia, Safari
+            if isinstance(header, dict):
+                try:
+                    key = header["name"]
+                    value = header["value"]
+                except KeyError as e:
+                    raise exceptions.OptionsError(str(e)) from e
```
mhils, line 33: Why are we raising OptionsError here? Given that the input is malformed, we should a) raise an appropriate error type (ValueError makes most sense I think), or b) not do any error handling and let the KeyError propagate as-is.

examples/contrib/readhar.py
```diff
@@ -0,0 +1,141 @@
...
+    # Don't know how to make a type annotation for the request json
+    def request_to_flow(self, request_json: dict) -> http.HTTPFlow:
+        """
+        Creates a HTTPFlow object from a given entry in HAR file
+        """
+
+        request_method = request_json["request"]["method"]
+        request_url = request_json["request"].get("url", None)
+        if not request_url:
+            request_url = request_json["request"].get("host", None)
```
mhils, line 55: Can you explain this part here? When is `url` unavailable? And does `host` really contain a URL in this case?

examples/contrib/readhar.py
```diff
@@ -0,0 +1,141 @@
...
+
+        client_conn = connection.Client(
+            peername=("127.0.0.1", 51513),
+            sockname=("127.0.0.1", 8080),
+            # TODO Get time info from HAR File
+            timestamp_start=time.time(),
+        )
+
+        # 375:3 is default mitmproxy server_conn when making a new flow.
+        server_conn = connection.Server(address=("375", 3))
```
mhils, line 67: I don't comment and code here. Can you explain? :)

examples/contrib/readhar.py
```diff
@@ -0,0 +1,141 @@
...
+            newflow.response = http.Response.make(
+                response_code, response_content, response_headers
+            )
+        except TypeError as e:
+            logger.error("Failed to create response")
+            raise exceptions.OptionsError(str(e)) from e
+
+        return newflow
+
+    async def load_flows(self) -> None:
```
mhils, line 93: What happened to https://github.com/mitmproxy/mitmproxy/pull/6190#discussion_r1240470653? This seems overly complicated. I think we can avoid all instance state (including `self.flows`) for this addon.

examples/contrib/readhar.py
```diff
@@ -0,0 +1,141 @@
...
+        newflow = http.HTTPFlow(client_conn, server_conn)
+        try:
+            newflow.request = http.Request.make(
+                request_method, request_url, "", request_headers
+            )
+        except TypeError as e:
+            logger.error("Failed to create request")
+            raise exceptions.OptionsError(str(e)) from e
+
+        response_code = request_json["response"].get("status", "")
```
mhils, line 78: The status code should always be an int, never a string.

examples/contrib/readhar.py
```diff
@@ -0,0 +1,141 @@
...
+    ) -> None:
+        """
+        Reads a HAR file into mitmproxy. Loads a flow for each entry in given HAR file.
+
+        Args:
+            path (types.Path): Path to HAR file
+        """
+        with open(path) as fp:
+            try:
+                harFile = json.load(fp)
```
mhils, line 125: snake_case

examples/contrib/readhar.py
```diff
@@ -0,0 +1,141 @@
...
+            except Exception:
+                logger.error("HAR file was unable to be loaded using json.load()")
+                raise (
+                    Exception(
```
mhils, line 129: If you raise a `mitmproxy.exceptions.CommandError`, this should look much nicer in the UI. We also don't need a separate log statement then. :)

mhils (commented):
Progress! 🚀

examples/contrib/readhar.py
```diff
@@ -0,0 +1,122 @@
...
+            except Exception:
+                raise (
+                    exceptions.CommandError(
+                        "Unable to read HAR file. Please provide a valid HAR file"
+                    )
+                )
+
+        for request_json in har_file["log"]["entries"]:
+            flow = self.request_to_flow(request_json)
+            if flow:
```
mhils, line 112: This check shouldn't be necessary, because `request_to_flow` always returns a flow object (or raises). :)

examples/contrib/readhar.py
```diff
@@ -0,0 +1,122 @@
...
+        flows = []
+        with open(path) as fp:
+            try:
+                har_file = json.load(fp)
+            except Exception:
+                raise (
+                    exceptions.CommandError(
+                        "Unable to read HAR file. Please provide a valid HAR file"
+                    )
+                )
```
mhils, line 108: Nit: The outer parentheses are unnecessary, this could just be one line 😛

examples/contrib/readhar.py
```diff
@@ -0,0 +1,122 @@
...
+        server_conn = connection.Server(address=("375", 3))
+
+        new_flow = http.HTTPFlow(client_conn, server_conn)
+        try:
+            new_flow.request = http.Request.make(
+                request_method, request_url, "", request_headers
+            )
+        except TypeError as e:
+            logger.error("Failed to create request")
+            raise exceptions.OptionsError(str(e)) from e
```
mhils, line 71: As with https://github.com/mitmproxy/mitmproxy/pull/6190#discussion_r1243518517, I think we don't need the `try` block here, it does not provide any additional context when it would happen. I'm also not even sure why/how `http.Request.make` would even fail here.

mhils (commented):
🚀

examples/contrib/readhar.py
```diff
@@ -0,0 +1,112 @@
...
+        Creates a HTTPFlow object from a given entry in HAR file
+        """
+
+        request_method = request_json["request"]["method"]
+        request_url = request_json["request"]["url"]
+        server_address = request_json["serverIPAddress"]
+        request_headers = self.fix_headers(request_json["request"]["headers"])
+
+        client_conn = connection.Client(
+            peername=("127.0.0.1", 51513),
```
mhils, line 55: Let's use port 0 as a placeholder, that makes it clear that it's an artificial value. :)

examples/contrib/readhar.py
```diff
@@ -0,0 +1,112 @@
...
+    @command.command("readhar")
+    def read_har(
+        self,
+        path: types.Path,
+    ) -> None:
+        """
+        Reads a HAR file into mitmproxy. Loads a flow for each entry in given HAR file.
+
+        Args:
+            path (types.Path): Path to HAR file
```
mhils, line 90: Good docstring! Feel free to skip the type here, that's a pattern from a time where python didn't have type annotations in the method itself. :)

mhils (commented):
Looks good! Please make sure that CI is passing (https://github.com/mitmproxy/mitmproxy/actions/runs/5405361567/jobs/9820804825), and then this is good to go! 😃

mhils (approved):
🎉

PR #6246, opened 2023-07-13
Import har

stanleygvi:
#### Description

Flow is now created with correct HTTP version from HAR file

Chrome.har is updated to be smaller and include POST data

timeout.har is a har file generated by a internet disconnect error


examples/contrib/import_har/write_to_expected.py
```diff
@@ -0,0 +1,21 @@
...
+    for path_name in file_list:
+        print(path_name)
+
+        flows = file_to_flows(path_name)
+
+        with open(f"har_files/{path_name.stem}.json", "w") as f:
+            json.dump(flows, f, indent=4)
+
+
+if __name__ == "__main__":
```
mhils, line 17: I moved this into test_readhar.py, so no need to add it here again. :)

examples/contrib/import_har/readhar.py
```diff
@@ -72,34 +70,21 @@ def request_to_flow(self, request_json: dict) -> http.HTTPFlow:
...
 
         response_code = request_json["response"]["status"]
 
         # In Firefox HAR files images don't include response bodies
-        response_content = request_json["response"]["content"].get("text", None)
-        content_encoding = request_json["response"]["content"].get("encoding", None)
-        if content_encoding == "base64":
-            response_content = base64.b64decode(response_content)
+        response_content = request_json["response"]["content"].get("text", "")
+
```
mhils, line 81: base64 missing as well :)

mhils (changes requested):

PR #6248, opened 2023-07-13
http version for flow is now based on HAR file

stanleygvi:
#### Description

Added handling for http version when creating flow.
Takes the HTTP version directly from the HAR file for the request and response objects of the flow.


examples/contrib/import_har/readhar.py
```diff
@@ -100,6 +104,9 @@ def request_to_flow(self, request_json: dict) -> http.HTTPFlow:
 
         new_flow.client_conn.timestamp_start = timestamp_start
         new_flow.client_conn.timestamp_end = timestamp_end
+
+        new_flow.request.http_version = http_version_req
```
mhils, line 108: This needs testing. If I look into the HAR files, I see cases where we have `httpVersion` set to `""`, or to `http/2.0`. We typically have something like `HTTP/2` in there though!

Also this should update the test results so that tests pass. :)

mhils (changes requested):

mhils (approved):

PR #6249, opened 2023-07-13
Chrome HAR file updated for more test cases

stanleygvi:
Added entries where:
POST data is included
Internet disconnected error

Also shortened amount of entries to 3.


mhils:
How does this only change the input file, but not the output file? :)

Relatedly, the POST data stuff should be in the POST data PR and not here. :)

PR #6250, opened 2023-07-13
request content handling

stanleygvi:
#### Description

Sets request content of flow object to text value of a requests postData if it exists.


examples/contrib/import_har/readhar.py
```diff
@@ -74,8 +75,11 @@ def request_to_flow(self, request_json: dict) -> http.HTTPFlow:
         new_flow = http.HTTPFlow(client_conn, server_conn)
 
         # FIXME: Handle request body.
```
mhils, line 77: Remove (now outdated) comment

mhils (changes requested):
This looks good, but it's missing a test file! 😃

mhils (commented):
Looks good. Let's make the example HAR a single request so that it's easier to spot issues?

PR #6252, opened 2023-07-13
Created addon for har export

stanleygvi:
Created addon similar to import_har. 
Args: flows, path to write to
Turns flow(s) into HAR file entry then writes HAR file with given flow(s) to given path.
Uses methods from har_dump.py


examples/contrib/export_har/test_exporthar.py
```diff
@@ -0,0 +1,46 @@
...
+@pytest.mark.parametrize(
+    "log_file", [pytest.param(here / "flows/logfile", id="logfile")]
+)
+def test_exporthar(log_file, tmp_path):
+    e = ExportHar()
+    path = open(log_file, "rb")
+    flows = io.FlowReader(path).stream()
+    e.export_har(flows, types.Path(tmp_path / "testing_flow.har"))
+    correct_har = json.load(open("flows/correct_flows.har"))
+    testing_har = json.load(open(tmp_path / "testing_flow.har"))
```
mhils, line 37: Let's not use `open` without a context manager (`with open(...): ...`) . That leaves open file handles around, which will break on Windows. You can run `.read_bytes()` on path objects and then just `json.load`.

examples/contrib/export_har/exporthar.py
```diff
@@ -0,0 +1,199 @@
...
+from mitmproxy.net.http import cookies
+from mitmproxy.utils import strutils
+
+
+logger = logging.getLogger(__name__)
+
+
+class ExportHar:
+    def __init__(self):
+        self.SERVERS_SEEN = set()
```
mhils, line 18: Short explanation what this variable is used for would be handy. Also it should probably be lowercase given that it's a regular attribute. :)

mitmproxy/addons/exporthar.py
```diff
@@ -0,0 +1,210 @@
...
+from mitmproxy.exceptions import CommandError
+from mitmproxy.utils import strutils
+
+logger = logging.getLogger(__name__)
+
+
+class ExportHar:
+    @command.command("exporthar")
+    def export_har(self, flows: Sequence[flow.Flow], path: types.Path) -> None:
+        """Writes provided flows into a HAR file at a given path"""
```
mhils, line 26:
```suggestion
        """Export flows to an HAR (HTTP Archive) file."""
```

mhils (approved):

zewelor:
Is it possible to use it for example with mitmweb to dump flows directly into .har file ? Can't find info about it, or how to do this.

mhils:
@zewelor: This is pretty much work-in-progress, we plan to have a better UX + docs for that in the next few weeks. :)

PR #6271, opened 2023-07-21
Remove incorrect usage of `typing.AnyStr`

sobolevn:
#### Description

Hi! While working on python/mypy#15732 our tools detected a misuse of AnyStr TypeVar (which is quite common). The proper way here is to use `@overload`s :)

#### Checklist


mhils (approved):
Fantastic, thank you so much! 🍰

Issue #5422, opened 2022-06-18
OpenSSL Error: unsafe legacy renegotiation disabled


EndUser509:
#### Problem Description
I have switched from the CentOS AppStream Python 3.9 to a self-compiled Python 3.10.4. With `python.10` I sometimes get this error which I never had with 3.9 (I checked all my old logs):
```
Unable to establish TLS connection with server (OpenSSL Error([('SSL routines', '', 'unsafe legacy renegotiation disabled')])). Trying to establish TLS with client anyway.
```
So this error seems to occur in the server-facing connection where mitmproxy is the client. Renegotiation is AFAIK initiated by the client so it looks like some default has changed for the newer python.  
Before I start digging deeper myself I wonder if this rings a bell with someone.

#### Steps to reproduce the behavior:
1. Set up mitmproxy or mitmdump
2. Go to https://bild.de and click on a few links in a browser
3. 

#### System Information
Mitmproxy: 9.0.0.dev (+11, commit 6110a2a)
Python:    3.10.4
OpenSSL:   OpenSSL 3.0.3 3 May 2022
Platform:  Linux-4.18.0-383.el8.x86_64-x86_64-with-glibc2.28

mhils:
This doesn't seem to be a mitmproxy bug, it looks like you are just dealing with a server that is so outdated that we cannot establish TLS securely.

andrewpedia:
> For those having the same problem, here is my fix: [1ad96dd](https://github.com/mitmproxy/mitmproxy/commit/1ad96dd4b89beb3a71ffaabf959d889d060e7141)
> 
> I originally found the required flag here: https://stackoverflow.com/a/76045866/1584915
> 
> I wonder if this is worthwhile to make a PR, maybe this flag should be added when the flag `ssl_insecure` is passed. This issue is very annoying when the target server uses outdated SSL and is not under control.

how to do this on android ? thanks

PR #6281, opened 2023-07-26
Enable unsafe legacy SSL renegotiation

DurandA:
#### Description

This PR fixes the issue discussed in #5422 when connecting to a server with outdated/misconfigured SSL. This is a draft PR to discuss if a full PR would be accepted. A full PR would only add the `OP_LEGACY_SERVER_CONNECT` SSL context flag if the `ssl_insecure` switch is passed. 

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,8 @@
 * For reverse proxy directly accessed via IP address, the IP address is now included
   as a subject in the generated certificate.
   ([#6202](https://github.com/mitmproxy/mitmproxy/pull/6202), @mhils)
+* Enable legacy SSL connect when connecting to server if the `ssl_insecure` flag is set.
+  ([#6281](https://github.com/mitmproxy/mitmproxy/pull/6281), @DurandA)
 
 ### Breaking Changes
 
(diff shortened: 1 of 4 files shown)


mhils:
Doing this only when ssl_insecure is set sounds good to me! Some notes:

1. This is only required for mitmproxy<->server connections, so we can move it to `create_proxy_server_context`.
2. Let's add a `legacy_server_connect` boolean argument there and pass `ssl_insecure` as the value in the tlsconfig addon.
3. Ideally `OP_LEGACY_SERVER_CONNECT` would be directly exposed by cryptography/pyOpenSSL. I'm happy to handle that part though once this PR is in. :)

DurandA:
@mhils Thanks for the guidelines. Let me know if this is ok as implemented.

> Ideally OP_LEGACY_SERVER_CONNECT would be directly exposed by cryptography/pyOpenSSL. I'm happy to handle that part though once this PR is in. :)

:+1:

mhils (approved):

mhils:
Looks great, thanks! I've filed PRs with cryptography and pyOpenSSL. :)

PR #6283, opened 2023-07-27
ReadHAR addon

stanleygvi:
#### Description
Moved readhar addon to addons
Args: path to import from
Imports flows from a HAR file into mitmproxy


mhils:
Can you rebase your PR cleanly onto main please? The diff on GitHub shows that his PR adds exporthar.py, which does not make sense. :)

PR #6305, opened 2023-08-09
Change default keybind to export flow to "x"

### LINKED ISSUE #6225: Both Edit and Export share the same key binding
#### Problem Description

When in flow view, both `Edit` and `Export` commands are bound to the `e` key by default.

imagen

`Export` could use a different binding by default, like `x`.

#### System Information

Mitmproxy: 9.0.1
Python:    3.11.3
OpenSSL:   OpenSSL 1.1.1u  30 May 2023
Platform:  macOS-12.6.5-arm64-arm-64bit

Llama1412:
#### Description

Changed the default keybinding for exporting flows from "e" (also used by Edit) to "x".
Resolves #6225

#### Checklist


mhils (approved):
Thanks for the PR and sorry this took so long! 😃

PR #6328, opened 2023-08-20
Include sent headers with CONNECT request to upstream proxy

Yopi:
#### Description

Includes the request headers from the client to the upstream proxy in the CONNECT request. Looking at https://github.com/mitmproxy/mitmproxy/discussions/5802, and the relevant events, I don't believe that there is a way today to access the headers that are sent from the client and override the empty headers sent to the upstream proxy.

I don't know if this is the best way to tackle it, or if there is a reason to strip all the headers in the CONNECT request, but for my specific use case of the day, it would be nice to be able to handle the traffic differently based on the user agent of the caller.  

#### Checklist

### DIFF
--- a/mitmproxy/addons/proxy_forward_headers.py
+++ b/mitmproxy/addons/proxy_forward_headers.py
@@ -1,0 +1,34 @@
+import weakref
+from collections.abc import MutableMapping
+
+from mitmproxy import connection
+from mitmproxy import ctx
+from mitmproxy import http
+
+
+class ProxyForwardHeaders:
+    def __init__(self) -> None:
+        self.connection_headers: MutableMapping[
+            connection.Client, tuple[str, str]
+        ] = weakref.WeakKeyDictionary()
+
+    def load(self, loader):
+        loader.add_option(
+            "proxyforwardheaders",
+            bool,
+            False,
+            """
+            Forward CONNECT headers to upstream proxy.
+            """,
+        )
+
+    def http_connect(self, f: http.HTTPFlow) -> None:
+        if ctx.options.proxyforwardheaders:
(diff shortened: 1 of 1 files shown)


Yopi:
Thank you for the feedback. That makes sense, that it might accidentally leak content.

I'll have a look if we can make it as an addon, but when I was playing with the different events it didn't seem like we could actually access the headers sent from the client to mitmproxy on the connect call. What `upstream_auth` is doing, if I'm not mistaken, is adding headers from mitmproxy to the upstream proxy on the request. 

Since I had to make the changes that I did in this PR for the connection between mitmproxy to the upstream proxy for any headers to be available in the flow in this event:
```
    def http_connect_upstream(self, f: http.HTTPFlow):
```

mhils:
> I'll have a look if we can make it as an addon, but when I was playing with the different events it didn't seem like we could actually access the headers sent from the client to mitmproxy on the connect call.

Take a look at the proxyauth addons - that one accesses them. `http_connect` to access, `http_connect_upstream` to set, and a weak key dict in the meantime to remember them.

Yopi:
Yes, you were completely right. Looks like it wasn't too difficult to achieve!

I'll add some tests as well, but I'm not sure what your general policy is on addons. Do you merge them if they are general enough, or are they better left to each user to have on their side? 

Similar question to whether or not it should be a default addon or not?

mhils:
An `forward_connect_headers` option (even if off by default) with a corresponding addon seems to be generic enough to be in mitmproxy by default. We don't have non-default addons. For things that cater very specific use cases or incur significant complexity, examples/contrib is a great breeding ground. :)

mhils:
Any update on this? :)

PR #6335, opened 2023-08-25
added har support in FlowReader class

stanleygvi:
#### Description

Added handling for HAR files in the stream() method of the FlowReader class.

#### Checklist


mitmproxy/io/har.py
```diff
@@ -0,0 +1,162 @@
...
+            case "HTTP/2":
+                new_flow.response.http_version = "HTTP/2"
+            case "HTTP/3":
+                new_flow.response.http_version = "HTTP/3"
+            case _:
+                new_flow.response.http_version = "HTTP/1.1"
+
+        return new_flow
+
+    @command.command("readhar")
```
mhils, line 134: Do we still need this command here? Feels like the normal read command covers everything now. 😃

mitmproxy/io/har.py
```diff
@@ -0,0 +1,162 @@
...
+            flows.append(flow)
+
+        async def load_flows() -> None:
+            for flow in flows:
+                await ctx.master.load_flow(flow)
+
+        asyncio.create_task(load_flows())
+
+
+addons = [ReadHar()]
```
mhils, line 162: This is not an addon anymore, so we can remove this. :)

mitmproxy/io/io.py
```diff
@@ -23,27 +26,41 @@ def add(self, f: flow.Flow) -> None:
...
-                    yield flow.Flow.from_state(compat.migrate_flow(loaded))
-                except ValueError as e:
-                    raise exceptions.FlowReadException(e) from e
-        except (ValueError, TypeError, IndexError) as e:
-            if str(e) == "not a tnetstring: empty file":
-                return  # Error is due to EOF
-            raise exceptions.FlowReadException("Invalid data format.") from e
+
+            for request_json in har_file["log"]["entries"]:
+                har_flow = ReadHar().request_to_flow(request_json)
```
mhils, line 45: What about errors raised here? They should probably also be reraised as a FlowReadException?

mhils (commented):
This looks good so far, but needs adjusted tests as well! 😃

stanleygvi:
Removed read_har() method in har.py because it wasn't used.

stanleygvi:
Also, I made another method to test the har files in test_io.py. It's basically the same as test_fuzz, but it runs it through each test file used in test_readhar.py

mitmproxy/io/io.py
```diff
@@ -10,10 +12,12 @@
 from mitmproxy import flowfilter
 from mitmproxy.io import compat
 from mitmproxy.io import tnetstring
+from mitmproxy.io.har import request_to_flow
 
 
 class FlowWriter:
     def __init__(self, fo):
+        print(type(fo))
```
mhils, line 20:
```suggestion
```

test/mitmproxy/io/test_har.py
```diff
@@ -0,0 +1,64 @@
...
+    actual_flows = file_to_flows(har_file)
+
+    for expected, actual in zip(expected_flows, actual_flows):
+        actual = json.loads(json.dumps(actual))
+
+        assert actual == expected
+
+
+if __name__ == "__main__":
+    for path_name in here.glob("har_files/*.har"):
```
mhils, line 58: files still need to be moved?

mhils (commented):
This still needs the removal of the old readhar addon + moving of test files, otherwise LGTM. 👍

mhils (approved):
🎉

PR #6352, opened 2023-09-04
Improve docstring for `Server.address`

mhils:
refs https://github.com/mitmproxy/mitmproxy/discussions/6351


mitmproxy/connection.py
rosydawn6, line 1: This clearly delineates the conditions under which a domain or ip will be present.  Perhaps match the language/tense (e.g. it will be, vs is always).  You could use: For transparent proxies such as WireGuard mode, this value will be an IP address.
    """

PR #6368, opened 2023-09-16
save HAR addon 

stanleygvi:
#### Description

Addon that writes flows to a HAR file. 

#### Checklist


test/mitmproxy/addons/test_savehar.py
```diff
@@ -0,0 +1,310 @@
...
+
+        for x in io.read_flows_from_paths(
+            [Path(test_dir / "data/flows/websocket.mitm")]
+        ):
+            a.websocket_end(x)
+        a.done()
+
+        with open(path) as inp:
+            har = json.load(inp)
+            har["log"]["creator"]["comment"] = "mitmproxy version 1.2.3"
```
stanleygvi, line 213: Monkeypatch didnt work with the savehar dump approach, so I changed it manually for each case. Is there a better way to do this?

mhils: You mean monkepatch did not work because the top-level `HAR` object was already created? An easy workaround here could be to only store the pages at runtime, and create the "outer frame" within `done`? :)

mitmproxy/addons/savehar.py
```diff
@@ -59,12 +66,12 @@ def flow_entry(self, flow: http.HTTPFlow, servers_seen: set[Server]) -> dict:
             ssl_time = -1.0
         elif flow.server_conn.timestamp_tcp_setup:
             assert flow.server_conn.timestamp_start
-            connect_time = 1000 * (
+            connect_time = (
```
mhils, line 69: Why did we remove the `1000 *` here? IIRC the HAR spec was in milliseconds, or am I mixing up things?

mitmproxy/addons/savehar.py
```diff
@@ -207,4 +231,59 @@ def format_multidict(self, obj: _MultiDict[str, str]) -> list[dict]:
         return [{"name": k, "value": v} for k, v in obj.items(multi=True)]
 
 
-addons = [ExportHar()]
+def load(l):
```
mhils, line 234: What are these top-level functions here doing? They are not part of the addon, so this looks like dead code?

stanleygvi: In the old har dump file they had it as top level functions so I kept it like that. Should I move them to a separate file?

mhils (approved):

PR #6373, opened 2023-09-19
mitmweb: Allow up to 4GB dump to be uploaded

### LINKED ISSUE #4651: Outfile -w cannot be loaded
#### Problem Description
When the dump file is getting bigger, about 100mb it's not loaded anymore.

Make a big outfile and try to open it with an new instance of mitmweb.

#### System Information
Mitmweb Windows 10 6.0.2

### LINKED ISSUE #5609: mitmweb Not loading my saved flow
So I recorded some actions with mitmweb and saved the flow.

Then I closed mitmweb, and reopened it.  Then I went to open the saved flow file (which is 100 megabytes).  But when I open it, the requests and responses do not appear?

t-wy:
#### Description

A temporary fix to allow up to 4GB dump to be uploaded by supplying max_buffer_size of the HTTPServer constructor to be 2**32 (4GB).

(Perhaps it is better to be a configurable item, or change the post mechanism to stream instead if there ever is any security concern implied due to usage or hardware limitations)

By default, if no argument is supplied, the maximum buffer size is 100MB, which causes saved dump file to be unreadable through the web interface.

Reference Documentation: [Link](https://www.tornadoweb.org/en/stable/httpclient.html#:~:text=a%20simple%20mapping.-,max_buffer_size,-(default%20100MB)%20is)

Possibly:
Resolves #4651, resolves #5609

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Add support for reading HAR files using the existing flow loading APIs, e.g. `mitmproxy -r example.har`.
   ([#6335](https://github.com/mitmproxy/mitmproxy/pull/6335), @stanleygvi)
+* Increase maximum dump file size accepted by mitmweb
 
 
 ## 04 August 2023: mitmproxy 10.0.0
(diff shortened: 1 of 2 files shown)


mhils (approved):
Awesome, thanks for digging into this! 🍰

huuck:
Just tried the latest version, still happening with a 400 Mb file

huuck:
My bad, it just takes a really long time to upload it. Sorry for the confusion.

PR #6410, opened 2023-10-18
fix certificates generated by dummy_cert on clients that require AuthorityKeyIdentifier

mmaxim:
#### Description

`Cert.dummy_cert` creates certificates for domains and signs them with the configured CA certificate. Currently, these certificates are generated without the `SubjectKeyIdentifier` or `AuthorityKeyIdentifier` extensions. Normally this is fine, however in OpenSSL 3.x, it is required that these extensions be present if `X509_V_FLAG_X509_STRICT` is set. For such clients, mitmproxy right now serves "broken" certificates and nothing works. With this change, we add those two extensions to get the certificates generated by `dummy_cert` in compliance with those strict checks. 

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
   ([#6389](https://github.com/mitmproxy/mitmproxy/pull/6389), @mhils)
 * Add a contentview for DNS-over-HTTPS.
   ([#6389](https://github.com/mitmproxy/mitmproxy/pull/6389), @mhils)
+* Fix certificate generation to work with strict mode OpenSSL 3.x clients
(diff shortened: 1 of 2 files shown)


mhils:
Thanks for digging into this, fantastic! 🍰  Could you add a changelog entry please? Otherwise this looks good to merge.

mmaxim:
> Thanks for digging into this, fantastic! 🍰 Could you add a changelog entry please? Otherwise this looks good to merge.

No problem, thanks for looking so quickly! I added a changelog item for the fix.

mhils (approved):
Thanks! 🍰

PR #6411, opened 2023-10-18
Make mitmweb use relative paths to connect to its resources.

davet2001:
#### Description

mitmweb currently serves up urls that are relative to the document root of the webserver, e.g. `/static/app.js`
This is fine when running mitmweb in a conventional sense, i.e. running on one machine and clients directly connecting to it.

However, if the mitmweb server is provided as part of another document tree, it may not be at the top level.  This causes connections to `/static/app.js`, as well as all other assets and the `/updates` websocket to not work correctly, as they now point to a page that does not exist on that server

For example:
Working ok:
```
https://example.com/index.html
-->https://example.com/static/app.js
```

Not working:
```
https://example.com/mysubdir/index.html
-->https://example.com/static/app.js
```
File not found.

What we need is for index.html to reference with a relative link, i.e. `static/app.js` instead of `/static/app.js`
This applies to 4 links within index.html, and part of the javascript that generates the websocket connection code that runs on the client.

Proposed fix:
```
https://example.com/mysubdir/index.html
-->https://example.com/mysubdir/static/app.js
```
Working.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
   ([#6389](https://github.com/mitmproxy/mitmproxy/pull/6389), @mhils)
 * Fix certificate generation to work with strict mode OpenSSL 3.x clients
   ([#6410](https://github.com/mitmproxy/mitmproxy/pull/6410), @mmaxim)
+* Fix root-relative URLs so that mitmweb can run in subdirectories.
+  ([#6411](https://github.com/mitmproxy/mitmproxy/pull/6411), @davet2001)
 
 
 ## 27 September 2023: mitmproxy 10.1.1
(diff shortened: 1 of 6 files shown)


davet2001:
I'm currently successfully working around this by [patching the mitmproxy library after install](https://github.com/davet2001/home-assistant-addons/blob/f2b6f4b09df490c88c09a698bdb5a88ab6808ffc/mitmproxy/Dockerfile#L52-L55):

mhils:
I can't guarantee that this will keep working because it's not covered by CI, but happy to merge this as-is. Would you mind adding a changelog entry please?

davet2001:
@mhils I have added at least an initial CI test which would catch a root-relative URL reappearing in index.html, and spotted+removed a few others.

I think catching a root-relative URL generated by the client side .js would be much harder (not done so far).

Note: It looks like app.js is in the git source tree, but I assume it is generated somehow from files including websocket.tsx which is one I modified.  I am not sure how this is done, and I have not regenerated it.  If I understand correctly, that needs doing before the next release.

Changelog updated.

PR #6442, opened 2023-11-01
cheat knowbe4

yaitskov:
#### Description

Avoid boring quiz from knowbe4


mhils:
Thanks! I appreciate the intent, but - while benign - that's not the type of example we want to host. :)

PR #6443, opened 2023-11-02
Feat(*) : Allow Cors requests in web

levyitay:
#### Description

Added the option to allow cors requests for WS and http web requests.
This feature can be turn on with `allow_cors` option.
It will allow us to use a standalone web ui with the MITMProxy backend implementation.
#### Checklist


Prinzhorn:
Not sure why this PR has been lying around for so long. Two thoughts:

1. You deactivate all other security headers, since they're in the `else` branch
2. I would prefer an explizit origin, or else visiting a malicious website would mean they can read all your mitmproxy data. But I guess this has become a non-issue with the introduction of web-password.

levyitay:
Thanks @Prinzhorn 
If you still think this feature is important I can work on it, otherwise we can close the PR

mitmproxy/tools/cmdline.py
```diff
@@ -140,7 +140,7 @@ def mitmweb(opts):
     opts.make_parser(group, "web_open_browser")
     opts.make_parser(group, "web_port", metavar="PORT")
     opts.make_parser(group, "web_host", metavar="HOST")
-
+    opts.make_parser(group, "allow_cors")
```
Prinzhorn, line 143: `web_allow_cors`

mitmproxy/tools/web/app.py
```diff
@@ -290,6 +299,11 @@ def open(self):
     def on_close(self):
         self.connections.discard(self)
 
+    def check_origin(self, origin):
```
Prinzhorn, line 302: Please explain. Doesn't this prevent normal ws connection?

mitmproxy/tools/web/app.py
```diff
@@ -498,8 +512,6 @@ def get(self, flow_id, message):
         cd = f"attachment; filename={filename}"
         self.set_header("Content-Disposition", cd)
         self.set_header("Content-Type", "application/text")
-        self.set_header("X-Content-Type-Options", "nosniff")
```
Prinzhorn, line 501: Why remove?

Prinzhorn (changes requested):
I've added some more comments. In general I see value in adding this. But I'm not sure how you're going to make this work with web password. You definitely need to allow credentials in CORS but not sure how you're going to authenticate the other frontend.

PR #6458, opened 2023-11-04
Handle "path" KeyError in the savehar addon

pogzyb:
Use `.get` when retrieving "path" from the `attrs` cookie dict. This avoids the occasional `KeyError`. 

#### Description

Modifies a line in the `format_response_cookies` function in the `savehar.py` addon. I'm not sure if there is another underlying issue preventing the "path" key from making it into the `attrs` dict, but either way this solved issues I was experiencing.

#### Checklist


stanleygvi:
Thanks for catching this 😄!

Issue #3489, opened 2019-02-27
Retrun 404 if  server_replay_kill_extra flag is set to true


bebef1987:
Background:
Mozilla is using mitmproxy  for performance testing using it's server replay capability 

When replaying our recorded files we need to return a 404 HTTPResponse  if a request is not found in the recording file.

Because of this we need to support and update a separate addon to manage this replay.

Request:
Can we add a option to server replay to return 404  response if the response if not found in the recording?

bebef1987:
Instead of killing the request here:
https://github.com/mitmproxy/mitmproxy/blob/master/mitmproxy/addons/serverplayback.py#L191

add a option to return:
 `http.HTTPResponse.make(404, b'', {'content-type': 'text/plain'})`

PR #6465, opened 2023-11-06
Add option to return various status codes rather than killing during server replay

DKarandikar:
#### Description

Designed to satisfy the requirements of https://github.com/mitmproxy/mitmproxy/issues/3489

Add `server_replay_404_extra` which behaves similarly to the kill flag, but returns 404 responses rather than killing

#### Checklist


nneonneo:
Would it maybe make sense for the status code to be configurable (e.g. to change it to 204, 400, 404, 500, etc.)?

DKarandikar:
That sounds reasonable to me. I don't know if it would be better for the option to take an `int` for any status code you might want, or to enumerate some reasonable ones and use the `choices` option for `add_option`. Thoughts on which would be better?

mhils:
Thanks for the PR! 🍰 How would you folks feel about a single
```
server_replay_extra = "forward" | "kill" | "204" | "400" | "404" | "500"
```
option with "forward" as the default? We could then deprecate `server_replay_kill_extra` eventually. :)

mitmproxy/addons/serverplayback.py
```diff
@@ -30,7 +30,15 @@ def load(self, loader):
             "server_replay_kill_extra",
             bool,
             False,
-            "Kill extra requests during replay (for which no replayable response was found).",
+            "Kill extra requests during replay (for which no replayable response was found)."
+            "[Deprecated, prefer to use server_replay_extra='kill']",
```
mhils, line 34: Awesome! In `configure`, can we emit a `logger.warning` message if server_replay_kill_extra is set so that people start moving over? :)

mitmproxy/addons/serverplayback.py
```diff
@@ -30,7 +30,15 @@ def load(self, loader):
...
+        )
+        loader.add_option(
+            "server_replay_extra",
+            str,
+            "forward",
+            "Behaviour for extra requests during replay for which no replayable response was found.",
```
mhils, line 40: Let's add a second sentence explaining what the numbers do. Something like that?
```
Setting a numeric string value will return an empty HTTP response with the respective status code.
```

mhils (commented):
Awesome, thanks for the quick turnaround! 🍰

Issue #5998, opened 2023-03-19
Changing proxy credentials for upstream proxy. Missing 'Proxy-Authorization' header in HTTPS request


PATAPOsha:
#### Problem Description
I want to dynamically change upstream proxy depending on auth data I connect to my MITM proxy server. Everything seem work, but only for HTTP requests, when I try same with HTTPS requests - there is no 'Proxy-Authorization' header inside request.

#### Steps to reproduce the behavior:
1. Create addon for reading and modifying request proxy auth data:
`my_addon.py`
```
import re
import base64
import mitmproxy.http
from mitmproxy.net.server_spec import ServerSpec


class MyAddon:
    def __init__(self):
        self.num = 0

    def requestheaders(self, flow: mitmproxy.http.HTTPFlow):
        # flow.server_conn.via = ServerSpec('http', ('127.0.0.1', 8994))
        
        # Read proxy auth data passed to connect to MITM proxy server - MITM_USER:MITM_PASS
        proxy_auth_b64 = flow.request.headers["Proxy-Authorization"]
        proxy_auth_b64 = proxy_auth_b64.split("Basic ")[1]
        proxy_auth = base64.b64decode(proxy_auth_b64).decode()
        username, password = proxy_auth.split(":")
        flow.metadata["mitm_auth"] = (username, password)
        print(f"requestheaders(): proxy_auth={proxy_auth}")

    def request(self, flow: mitmproxy.http.HTTPFlow):
        # Read proxy auth data passed to connect to UPSTREAM proxy server - UPSTREAM_USR:UPSTREAM_PASSW
        proxy_auth_b64 = flow.request.headers["Proxy-Authorization"]
        proxy_auth_b64 = proxy_auth_b64.split("Basic ")[1]
        proxy_auth = base64.b64decode(proxy_auth_b64)
        print(f"request(): proxy_auth={proxy_auth}")

        # Set another user/password for connecting to UPSTREAM proxy server
        new_proxy_auth = "UPSTREAM_USR_2:UPSTREAM_PASSW_2"
        new_proxy_auth_b64 = base64.b64encode(new_proxy_auth.encode()).decode()
        flow.request.headers["Proxy-Authorization"] = f"Basic {new_proxy_auth_b64}"
        return
 
 addons = [MyAddon()]
```
2. Start MITM proxy server using script
`run_debug.py`
```
from mitmproxy.tools.main import mitmdump

mitmdump(args=[
    "-s", "my_addon.py",
[...]

PR #6504, opened 2023-11-28
Allow passing a hostname which includes authentication credentials

seidnerj:
#### Description

This commit allows passing a hostname with authentication credentials as an upstream proxy value (e.g. upstream:http://user:pass@default-upstream-proxy:8080/) something which thus far would result in a connection error. This allows the connection address (which includes the credentials) to be used later to and set proxy authentication headers (amongst other things).

In addition, I believe that since the credentials could now be passed as part of the address, different credentials will prevent connection reuse resulting in a new connection when different credentials are used for the same hostname/port combination. I believe this is desired behavior since a proxy server can route traffic differently depending on the provided credentials.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Unreleased: mitmproxy next
 
-
+* Allow passing a hostname which includes authentication credentials.
+    ([#6504](https://github.com/mitmproxy/mitmproxy/pull/6504), @seidnerj)
 
 ## 14 November 2023: mitmproxy 10.1.5
 
--- a/mitmproxy/proxy/server.py
+++ b/mitmproxy/proxy/server.py
@@ -204,16 +204,16 @@ class ConnectionHandler(metaclass=abc.ABCMeta):
             writer: asyncio.StreamWriter | udp.DatagramWriter
             try:
                 command.connection.timestamp_start = time.time()
-                if command.connection.transport_protocol == "tcp":                    
+                if command.connection.transport_protocol == "tcp":
                     host, port = command.connection.address
                     split_host = host.split("@", 1)
-                    
+
                     # this can only be 1 or 2, no other value is possible.
-                    if len(split_host) == 2: 
+                    if len(split_host) == 2:
                         host = split_host[1]
-                        
+
                     reader, writer = await asyncio.open_connection(
-                        host, 
+                        host,
                         port,
                         local_addr=command.connection.sockname,
                     )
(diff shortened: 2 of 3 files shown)


seidnerj:
I'm having trouble finding a way to add a test that covers this change, would appreciate any help 🙏

mhils:
Thank for the PR! I appreciate you raising this, but this is not in a state where I would feel comfortable merging it. [`connection.address`](https://docs.mitmproxy.org/stable/api/mitmproxy/connection.html#Server.address) has a well-defined meaning and there is no world where we will start smuggling authentication credentials in there. If we want to support this, we need to start at the [`UpstreamMode` implementation](https://github.com/mitmproxy/mitmproxy/blob/10.1.5/mitmproxy/proxy/mode_specs.py#L203) and then take over in the existing [`UpstreamAuth` addon](https://github.com/mitmproxy/mitmproxy/blob/10.1.5/mitmproxy/addons/upstream_auth.py). Does that make sense?

mhils:
> Also - what are your thoughts about the potential connection re-use issue I raised?

Ah yes - this is quite annoying. At the moment we really only support one set of credentials in form of the upstream_auth setting. And the approach I described above would not really fix that. To solve this properly, we'd probably need to extend  https://github.com/mitmproxy/mitmproxy/blob/main/mitmproxy/net/server_spec.py to include authentication information. Which is quite a PITA unfortunately - not sure if the solved edge case (I assume it may not be an edge case for you - but this issue hasn't really come up much (at all?) before) makes up for the added complexity.

seidnerj:
Oh yes, I started working on implementing this there but there was so much changes that had to be made that I gave up and monkey patched the code in the same fashion as I proposed in this PR.

My scenario is that I need a client to be able to set the upstream proxy used (via a specialized header that's passed). Now, for HTTP that's not an issue because you can simply add a"Proxy-Authorization" header during the "request" event, but for HTTPS you need to add the "Proxy-Authorization" header during the "http_connect_upstream", trouble is there doesn't seem to be any way to access the original request's headers (or share that information in some other way) during that event.

This proposed change allowed me to share the credentials via the server address, but I definitely understand why this is a less-than-ideal (to say the least) change.

BTW - here are two related issues/discussons in the mitmproxy repo recently:
https://github.com/mitmproxy/mitmproxy/discussions/6471
https://github.com/mitmproxy/mitmproxy/issues/5998

seidnerj:
Will check it out, thanks!

patilabhay679:
```         if server_connection_already_open:
            # server_conn already refers to an existing connection (which cannot be modified),
            # so we need to replace it with a new server connection object.
            flow.server_conn = Server(address = flow.server_conn.address)
        
        # flow.server_conn.via = ServerSpec(("http", address))    
        flow.server_conn.via = ServerSpec(("http", (proxy_host, proxy_port)))
````

using --set upstream_auth=username:password command line flag is working, but

passing flow.request.headers["Proxy-Authorization"] is not working?
looks like this header is stripped away internally.

having proxy scheme http and target scheme http?

any ideas?

PR #6513, opened 2023-12-01
Fix for --allow-hosts/--ignore-hosts options in WireGuard mode (#5930)

dsphper:
#### Description

This PR addresses issue #5930, where using the `--allow-hosts` or `--ignore-hosts` options in WireGuard mode was causing all traffic to break. The issue seems to be related to DNS resolution. 

The fix ensures that DNS traffic is handled correctly when these options are used, allowing users to filter traffic without breaking their network connection. Now, users can use `--allow-hosts` or `--ignore-hosts` to focus on specific traffic, making the network inspection process more efficient. 

This update enhances the usability of mitmproxy in WireGuard mode, especially for users who need to filter out irrelevant traffic.
#### Checklist

### DIFF
--- a/mitmproxy/addons/next_layer.py
+++ b/mitmproxy/addons/next_layer.py
@@ -225,7 +225,7 @@ class NextLayer:
         ) and client_hello.sni:
             hostnames.append(client_hello.sni)
         # If the client data is not a TLS record, try to extract the domain from the HTTP request
-        elif (host := self._extract_http1_host_header(data_client)):
+        elif host := self._extract_http1_host_header(data_client):
             hostnames.append(host)
         if not hostnames:
             return False


dsphper:
@mhils  I have submitted the PR. Please review when you have time. Thanks!

mitmproxy/addons/next_layer.py
```diff
@@ -239,6 +247,12 @@ def _ignore_connection(
         else:  # pragma: no cover
             raise AssertionError()
 
+    @staticmethod
+    def _extract_host(data_client: bytes) -> str:
```
mhils, line 251:
```suggestion
    def _extract_http1_host_header(data_client: bytes) -> str:
```
(I'd like to emphasize this is for HTTP and HTTP/1 only)

mitmproxy/addons/next_layer.py
```diff
@@ -239,6 +247,12 @@ def _ignore_connection(
...
+        pattern = rb"Host:\s+(.*?)\r\n"
```
mhils, line 252:
```suggestion
        pattern = rb"Host:\s+(.+?)\r\n"
```

mitmproxy/addons/next_layer.py
```diff
@@ -220,10 +220,18 @@ def _ignore_connection(
...
+        else:
+            if isinstance(data_client, bytes):
+                host = self._extract_host(data_client)
+                if host:
+                    hostnames.append(host)
         if not hostnames:
             return False
 
+        # Special handling for wireguard mode: if the hostname is "10.0.0.53", do not ignore the connection
+        if "wireguard" in ctx.options.mode and "10.0.0.53" in hostnames:
```
mhils, line 233: We can do a slightly more precise check here, like this:
https://github.com/mitmproxy/mitmproxy/blob/439e5264eb8aaff74aeb80588b86e13b9f0c320e/mitmproxy/addons/dns_resolver.py#L147-L148

flow.client_conn -> context.client and the same for server. :)

mitmproxy can spawn multiple proxy servers now, and this makes sure we only target the correct one.
We can also move this to the top of the method as it does not depend on collecting `hostnames`.

CHANGELOG.md
```diff
@@ -1,8 +1,7 @@
 # Release History
 
 ## Unreleased: mitmproxy next
-
-
+* Fix for --allow-hosts/--ignore-hosts options in WireGuard mode (#5930)
```
mhils, line 4: Let's add a bit more detail here. How about something like this?

- DNS resolution is now exempted from `--ignore-hosts` in WireGuard Mode.
- For plaintext traffic, `--ignore-hosts` now also takes HTTP/1 host headers into account.

Improvements welcome! 😃

mitmproxy/addons/next_layer.py
```diff
@@ -220,10 +220,18 @@ def _ignore_connection(
             client_hello := self._get_client_hello(context, data_client)
         ) and client_hello.sni:
             hostnames.append(client_hello.sni)
-
+        # If the client data is not a TLS record, try to extract the domain from the HTTP request
+        else:
+            if isinstance(data_client, bytes):
+                host = self._extract_host(data_client)
+                if host:
+                    hostnames.append(host)
```
mhils, line 228:
```suggestion
        elif host := self._extract_http1_host_header(data_client):
            hostnames.append(host)
```

`isinstance(data_client, bytes)` should always hold based on the functions type signature and the rest can be shortened a bit :)

mhils (commented):
This is highly appreciated, thanks! I've added a few minor comments below, but this is great overall. 😃

mitmproxy/addons/next_layer.py
dsphper: Hello @mhils,

Thank you for your valuable feedback. I greatly appreciate your suggestions, they make perfect sense.

CHANGELOG.md
dsphper: I wholeheartedly agree with your proposed changes.

dsphper:
> Looks great now, thanks! Let's resolve the instance check question and this is good to go. 😃现在看起来很棒，谢谢！让我们解决实例检查问题，这样就可以了。 😃

?

dsphper:
Hi @mhils ,

The `isinstance` check here is used to handle a special case in our tests. In the test `test_next_layer`, there's a scenario where `data_client` is assigned a MagicMock object, which is not of type `bytes`. Attempting to use a regular expression to match against this MagicMock object will lead to an error.

Here's the relevant code from the test:
Code Line:test/mitmproxy/addons/test_next_layer.py:334
```python
m.layer = None
nl.next_layer(m)
assert m.layer
```

In this scenario, `data_client` ends up receiving a MagicMock object instead of `bytes`. So, the `isinstance` check is used to avoid the error in this situation. 

There are two possible solutions to the issue:

1. Keep the `isinstance` check to handle this special case.
2. Modify the test to ensure `data_client` always receives a `bytes` object. This could be done by adding `monkeypatch.setattr(m, "data_client", lambda: b"example string")` before calling `nl.next_layer(m)`.

```python
m.layer = None
monkeypatch.setattr(m, "data_client", lambda: b"example string")
nl.next_layer(m)
assert m.layer
```

Either solution would resolve the issue, but it's a matter of deciding whether to handle the special case in the function or in the tests.


Given the explanation above for the `isinstance` check, I would like to know your preference on how to proceed. Do we keep the `isinstance` check to handle the special case, or should we modify the test so that `data_client` always receives a bytes object? Alternatively, do you see a better solution to handle this scenario? Your guidance on this would be greatly appreciated.

Best.

dsphper:
> See [#6513 (comment)](https://github.com/mitmproxy/mitmproxy/pull/6513#discussion_r1413331062). Does my question make sense?请参阅#6513（评论）。我的问题有意义吗？

Hi @mhils ,

The `isinstance` check here is used to handle a special case in our tests. In the test `test_next_layer`, there's a scenario where `data_client` is assigned a MagicMock object, which is not of type `bytes`. Attempting to use a regular expression to match against this MagicMock object will lead to an error.

Here's the relevant code from the test:
Code Line:test/mitmproxy/addons/test_next_layer.py:334
```python
m.layer = None
nl.next_layer(m)
assert m.layer
```

In this scenario, `data_client` ends up receiving a MagicMock object instead of `bytes`. So, the `isinstance` check is used to avoid the error in this situation. 

There are two possible solutions to the issue:

1. Keep the `isinstance` check to handle this special case.
2. Modify the test to ensure `data_client` always receives a `bytes` object. This could be done by adding `monkeypatch.setattr(m, "data_client", lambda: b"example string")` before calling `nl.next_layer(m)`.

```python
m.layer = None
monkeypatch.setattr(m, "data_client", lambda: b"example string")
nl.next_layer(m)
assert m.layer
```

Either solution would resolve the issue, but it's a matter of deciding whether to handle the special case in the function or in the tests.


Given the explanation above for the `isinstance` check, I would like to know your preference on how to proceed. Do we keep the `isinstance` check to handle the special case, or should we modify the test so that `data_client` always receives a bytes object? Alternatively, do you see a better solution to handle this scenario? Your guidance on this would be greatly appreciated.

Best.

mhils:
Let's fix the test and remove the check. Thanks! 😃

mhils (approved):
Thank you so much! 😃 🍰

dsphper:
> Thank you so much! 😃 🍰太感谢了！ 😃🍰

😄

Issue #6479, opened 2023-11-12
Warn if both `stream_large_bodies` and `modify_body` are set


mhils:
See below for a user report - we should print a warning if stream_large_bodies and modify_body are set. Contributions welcome! 😃 

--------------------

I found this and a few other links looking for information on this topic.  Several attempts were not producing the expected results.  I was using stream_large_bodies=1 and after rereading the docs more https://docs.mitmproxy.org/stable/concepts-options/#stream_large_bodies found that when using stream_large_bodies=1 the body is not stored and hence cannot be modified.

_Originally posted by @rosydawn6 in https://github.com/mitmproxy/mitmproxy/discussions/5151#discussioncomment-7537706_

rosydawn6:
I can do this.  My initial thought was to add some warning/notice to the options in the documentation.  You seemed to imply printing a warning message upon startup if stream_large_bodies flag is set?  Which would you like me to do - or both?

mhils:
I think it makes most sense to 1) mention it in the docstring for `stream_large_bodies`, and 2) log with ALERT level in the affected addons. Thanks! :)

rosydawn6:
Few comments on the attached PR:
I log with alert in both stream_large_modies addon and in modify_body addon.  If both options are set it will basically duplicate the log message.  Did you envision something else @mhils?

mhils:
Thanks, see PR. :)

PR #6514, opened 2023-12-01
Stream large bodies warn with modify body

rosydawn6:
#### Description

Added alert logging for stream_large_bodies and modify_body and enhance documentation around these options ( #6479 )

#### Checklist

 - [NA] I have updated tests where applicable.
 - [Y] I have added an entry to the CHANGELOG.


mitmproxy/addons/modifybody.py
```diff
@@ -20,7 +21,7 @@ def load(self, loader):
             """
             Replacement pattern of the form "[/flow-filter]/regex/[@]replacement", where
             the separator can be any character. The @ allows to provide a file path that
-            is used to read the replacement string.
+            is used to read the replacement string.  Streamed bodies cannot be modified.
```
mhils, line 24: I think this addition here may add a bit confusion as it's not clear that this refers to mitmproxy's off-by-default streaming functionality and not say any chunked HTTP encoding. I'd be fine with leaving this out here entirely and just document it in the `stream_large_bodies` option.

mitmproxy/addons/modifybody.py
```diff
@@ -30,6 +31,7 @@ def configure(self, updated):
             for option in ctx.options.modify_body:
                 try:
                     spec = parse_modify_spec(option, True)
+                    logging.log(ALERT, "Cannot modify streamed bodies.")
```
mhils, line 34: As you said, this is just duplicating - let's take it out.

mitmproxy/addons/proxyserver.py
```diff
@@ -211,6 +213,7 @@ def configure(self, updated) -> None:
         if "stream_large_bodies" in updated:
             try:
                 human.parse_size(ctx.options.stream_large_bodies)
+                logging.log(ALERT, "Cannot modify streamed bodies.")
```
mhils, line 216:
```suggestion
                logging.log(ALERT, "Both modify_body and stream_large_bodies are active. Streamed bodies will not be modified.")
```

(further suggestions welcome if you feel this is worse - I feel that directly linking this to the respective option names is useful here)

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased: mitmproxy next
 
 
+Enhance documentation and add alert log messages when stream_large_bodies and modify_body are set (#6479)
```
mhils, line 6:
```suggestion
* Enhance documentation and add alert log messages when stream_large_bodies and modify_body are set
```

CI will do the linking, but it needs to be a bullet point for that. :)

mhils (commented):
Thank you for the super nice first PR! It's great to have both docs updates + an interactive warning. 😃 🍰 🚀 

Super small suggestions below. :)

mitmproxy/addons/modifybody.py
rosydawn6: I have removed this and the corresponding import of ALERT

mitmproxy/addons/modifybody.py
rosydawn6: I have reverted this change

mitmproxy/addons/modifybody.py
rosydawn6: Clarifying that I did a git revert

CHANGELOG.md
rosydawn6: Thanks for that tip.

mitmproxy/addons/proxyserver.py
rosydawn6: I like linking to the option names.   One minor question the message now says that modify_body is active, but we are not checking for its presence.  Are we simply assuming that the body modification could happen via a def response lifecycle event or the modify_body option?

rosydawn6:
Thanks for your suggestions.  I overall agree and have made your changes with one clarifying question.

mitmproxy/addons/proxyserver.py
mhils: Ah, thanks for catching this. We should definitely check if modify_body is set, and only then emit an alert! :)

rosydawn6:
I have added the check for both options.  After some testing I saw that I need to check for both options being set in both addsons as the options can come in any order.

mhils (approved):
Thanks! 😃 🍰

PR #6527, opened 2023-12-04
Certstore option

Speedlulu:
#### Description

Add a new option: `certstore`, to store the CA certs somewhere other than `confir`
If the option is not passed the behaviour stays the same and the CA certs is created in `confdir`

#### Checklist


Speedlulu:
The changes mainly concern `mitmproxy/addons/tlsconfig.py` but I saw that `mitmproxy/utils/magisk.py` could also create a certstore.

If we decide that this option has nothing to do with the `magisk` module I could define the option directly in `mitmproxy/addons/tlsconfig.py` which seems to be the recommended way to do things.

Speedlulu:
I also included a fix for the signature of `TlsConfig.configure` but I may be wrong in thinking it was not correct.

Issue #3609, opened 2019-08-05
reopen response body flowfilter


iTosun:
Hi, 

first of all i want to thank the whole community and team that maintain this project. I love mitm and use it for many things during pen-tests. Currently i'm playing around with flowfilters and i noticed that when i try to implement a ~bs (response body) i get a notification in mitmweb that this is not implemented yet with a uri: https://github.com/mitmproxy/mitmweb/issues/10
The mitmweb is archived and i couldn't find it here as a issue. So like to know if this issue is still on the roadmap or not. 

Thanks in advance!

miaomi:
A temp workaround:
Use a script add some headers for matched flows then filter them by header in web UI.
` def response(flow):
    if re.findall(r'foo', flow.response.text):
           flow.response.headers["newheader"] = "foo"`
`

`~hs foo`

iTosun:
Thanks, i've implemented the temp-fix. Like to be updated when the body flow filters are fixed.

mhils:
Thanks for migrating this here, @iTosun! Yes, this is something we'd be happy to have support for, mitmproxy/mitmweb#10 has the relevant discussion. Unfortunately it's not quite trivial, if someone wants to give it a stab please ping us on Slack and I'm happy to help.

iTosun:
It seems that with mitm version 5.2 there are methods in the flow objects that could pose as a solution to this issue. I see that there are now a few functions related to the http request/response body:
- https://github.com/mitmproxy/mitmproxy/blob/v5.2/mitmproxy/net/http/message.py#L77
- https://github.com/mitmproxy/mitmproxy/blob/v5.2/mitmproxy/net/http/message.py#L89
- https://github.com/mitmproxy/mitmproxy/blob/v5.2/mitmproxy/net/http/message.py#L207 

You can use regex or simple string methods to make the ~b filter working properly

guyskk:
I want to use body filters to find out hidden parameters for api request in html or javascript when debug crawler code. But https://github.com/mitmproxy/mitmweb/issues/10 has been archived, the feature not implemented yet.

Or it there a way to parse/inspect the dumped flows file, so I can write python code to find it?

guyskk:
@guyskk I got example at https://docs.mitmproxy.org/stable/addons-examples/#io-read-saved-flows , here is my modified code to filter response body.

```python
import sys

from mitmproxy import http, io


def check_flow(flow: http.HTTPFlow, target: str):
[...]

PR #6538, opened 2023-12-07
Added support for `~b`, `~bq`, and `~bs` filters in `mitmweb`.

dsphper:
#### Description

This PR implements the `~b`, `~bq`, and `~bs` filters in `mitmweb`. These filters allow users to search or highlight response body contents, thereby enhancing the usability and functionality of `mitmweb`.

Previously, a user indicated in issue #3609 that the `~bs` (response body) implementation was not yet available in `mitmweb`. This update addresses that concern and extends the filtering capability to include `~b` and `~bq` filters as well.

This feature will significantly improve user experience, based on response body is crucial for efficient data analysis.

Looking forward to feedback and suggestions for improvement.

#### Checklist


dsphper:
@mhils, could you please review the code in this PR when you have some time? Your guidance and feedback would be greatly appreciated. Thank you.

dsphper:
Hi @mhils,
Could you please take a look and let me know if there are any additional improvements or changes you'd recommend? Any feedback would be greatly appreciated, as I aim to make this feature as useful as possible for the community.

Thank you for your time and consideration.

Best regards,

mhils (commented):
Thanks! This looks good so far, but unfortunately does not support flows as-they-appear. For example, if I specify `~b foo` first and only then a flow with `foo` in the body arrives, it will not be shown until I modify/resubmit the filter. 

Of course, implementing live search is significantly harder (we don't want to issue a new request every time a flow arrives). So we'd need to somehow tell the server via the existing WebSocket connection and then have the server-side mark requests as it transmits them. I know this is a major change, but do you see what I am aiming for?

ddxv:
> Thanks! This looks good so far, but unfortunately does not support flows as-they-appear. For example, if I specify `~b foo` first and only then a flow with `foo` in the body arrives, it will not be shown until I modify/resubmit the filter.
> 
> Of course, implementing live search is significantly harder (we don't want to issue a new request every time a flow arrives). So we'd need to somehow tell the server via the existing WebSocket connection and then have the server-side mark requests as it transmits them. I know this is a major change, but do you see what I am aiming for?

Perhaps a warning about this type of behavior? It's not the best, but it is somewhat understandable.

gjrtimmer:
any process on this?

cornfeedhobo:
@dsphper ping. are you able to continue with this, or do you need someone to pick up where you left off?

cornfeedhobo:
> For historical dumps (my use case) where there is no live traffic, this is more than sufficient and actually makes `mitmweb` filters useful. A warning that it doesn't auto-update (which is reasonable given the perf impact) seems like a pragmatic approach, especially since we already do give tooltip notices saying things like "body searching doesn't work, see this github issue". We could just say "it doesn't live update, see this github issue"?

@mhils Given the time passed, maybe this is a decent middle ground and a follow-up PR could be made?

cornfeedhobo:
> Thanks! This looks good so far, but unfortunately does not support flows as-they-appear. For example, if I specify `~b foo` first and only then a flow with `foo` in the body arrives, it will not be shown until I modify/resubmit the filter.
> 
> Of course, implementing live search is significantly harder (we don't want to issue a new request every time a flow arrives). So we'd need to somehow tell the server via the existing WebSocket connection and then have the server-side mark requests as it transmits them. I know this is a major change, but do you see what I am aiming for?

@mhils Is there any previous art for the functionality you're suggesting? Does ~h do this already?

ammar313-cs:
I’m working on implementing a more dynamic WebSocket filtering system in mitmproxy that supports live updates as flows arrive. Currently, with the standard add-on approach, filters are only applied when they are set or modified. This means if I set a filter for ~b foo, I won’t see any matching flows until the filter is resubmitted or modified.

I’m considering modifying the dumper.py and websockets.py files to allow filters to be passed to the server in real-time via the existing WebSocket connection. The server would then be responsible for marking requests as they are transmitted based on these filters.

Questions:

Would you prefer seeing this functionality as a new add-on, or would a direct modification to the core files (dumper.py, websockets.py) be more appropriate?
How would you recommend handling the passing of filters in real-time to ensure efficiency and minimize performance impact?
Are there any existing mechanisms within mitmproxy that could facilitate this kind of live, server-side filtering that I might have overlooked?
If possible, could you help me better understand the overall flow of the code, particularly how data is processed and passed through the different components? Understanding this would greatly assist me in making the necessary modifications.
Any guidance or suggestions would be greatly appreciated!

mhils:
Hi @ammar313-cs!

> The server would then be responsible for marking requests as they are transmitted based on these filters.

This is the correct approach. I think the best way to implement this is to make WebSocket connections stateful. The client needs to be able to send addFilter and removeFilter commands. Upon receipt of an addFilter message, the server should evaluate the passed filter on all existing flows, notify the client of what matches, and then send metadata for all new flows going forward.

This is all quite complex - I think a good first step would be to come up with a detailed proposal on what the data structures on the server side and on the client side should look like as well as the message format in the middle.


-----------------------

On your other questions:


> Would you prefer seeing this functionality as a new add-on, or would a direct modification to the core files (dumper.py, websockets.py) be more appropriate?

This should probably live in https://github.com/mitmproxy/mitmproxy/tree/main/mitmproxy/tools/web

> How would you recommend handling the passing of filters in real-time to ensure efficiency and minimize performance impact?

See above, that needs to be part of the design doc. 🤷 

> Are there any existing mechanisms within mitmproxy that could facilitate this kind of live, server-side filtering that I might have overlooked?

The view addon has filtering, but we probably shouldn't use that here. Realistically we just want to manually evaluate `flowfilter` on flows.

> If possible, could you help me better understand the overall flow of the code, particularly how data is processed and passed through the different components? Understanding this would greatly assist me in making the necessary modifications.

This should give you a rough idea:
https://github.com/mitmproxy/mitmproxy/blob/v10.4.2/mitmproxy/addons/view.py#L511-L523
https://github.com/mitmproxy/mitmproxy/blob/v10.4.2/mitmproxy/tools/web/master.py#L55-L69
https://github.com/mitmproxy/mitmproxy/blob/v10.4.2/mitmproxy/tools/web/app.py#L282-L316
https://github.com/mitmproxy/mitmproxy/blob/main/web/src/js/backends/websocket.tsx
https://github.com/mitmproxy/mitmproxy/blob/main/web/src/js/ducks/flows.ts

lups2000:
closed by #7704

PR #6550, opened 2023-12-12
simplify cert loading with new cryptography APIs

mhils:
/cc @driuba - does that look good?


driuba:
> /cc @driuba - does that look good?

Looks good to me 👍

PR #6555, opened 2023-12-17
Update cryptography requirement to >=39.0

bakugo:
https://github.com/mitmproxy/mitmproxy/commit/0b5e310881f0116c9e8e2ca5b9da98e258de0a11 uses `load_pem_x509_certificates` which was introduced in cryptography 39.0 but the requirement wasn't updated.


mhils:
Thanks! 🍰

Issue #6526, opened 2023-12-03
mitmweb - disabling the "edit options" button by a cli argument


ariel42:
#### Problem Description

In cases when mitmweb is exposed inside a network, (e.g. for monitoring some network resources and allowing the users to watch the traffic), the network admin may want to configure it in a specific way that fits the environment and the purpose of the installation, but the lack of authentication and permissions mechanisms makes it possible to anyone in the network to change the options and to corrupt the installation. Beside it this is an "alpha" feature that may still contain unknown vulnerabilities that may be exploited by a malicious user in the network.

#### Proposal
It is suggested to add a cli argument that will hide the "Edit options" button, along with all the server-side HTTP requests handlers that support the options there.

#### Alternatives
I didn't find a better idea without 3rd party tools.

PR #6568, opened 2023-12-25
Disabling the edit options button by a cli argument

Peacexoom:
Fixes: https://github.com/mitmproxy/mitmproxy/issues/6526

#### Description
Added a cli argument that will hide the "Edit options" button.

`
mitmweb --set disable_edit_options=true 
` 

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 ## Unreleased: mitmproxy next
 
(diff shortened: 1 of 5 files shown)


mhils:
I'd generally be happy to have something like this, but it needs to be enforced server-side and it needs to be comprehensive (i.e. cover all endpoints that essentially allow RCE). Users will perceive it as a security boundary.

PR #6578, opened 2024-01-04
savehar: Fix timings being set to null

DaniElectra:
#### Description

The HAR spec says that the `connect` and `ssl` timings are set to -1 when they are not applicable. This is the case for HTTP (not HTTPS) requests and for failed CONNECT requests, which are dumped to the HAR file too.

This allows dumps with these requests to be parsed correctly on [harviewer](http://www.softwareishard.com/har/viewer/).

#### Checklist

### DIFF
--- a/mitmproxy/addons/savehar.py
+++ b/mitmproxy/addons/savehar.py
@@ -144,11 +144,11 @@ class SaveHar:
                     - flow.server_conn.timestamp_tcp_setup
                 )
             else:
-                ssl_time = None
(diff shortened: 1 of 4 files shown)


stanleygvi:
Thank you for this PR 😁

DaniElectra:
It looks like the PR got misplaced on the CHANGELOG and it appears on the 10.2.0 release. Could you place it on `mitmproxy next`?

Issue #6579, opened 2024-01-04
mitmproxy saves failed CONNECT request URLs to HAR files incorrectly


DaniElectra:
#### Problem Description

mitmproxy saves failed CONNECT dumps to HAR files without having the protocol on the `url` field:

```
            {
                "startedDateTime": "2024-01-04T11:50:38.052258+00:00",
                "time": 0.37479400634765625,
                "request": {
                    "method": "CONNECT",
                    "url": "dhgyhuim.pretendo.cc:443",
                    "httpVersion": "HTTP/1.0",
                    "cookies": [],
                    "headers": [
                        {
                            "name": "User-Agent",
                            "value": "Opera/9.50 (Nintendo DSi; Opera/507; U; es-ES)"
                        },
                        {
                            "name": "Host",
                            "value": "dhgyhuim.pretendo.cc:443"
                        }
                    ],
                    "queryString": [],
                    "headersSize": 115,
                    "bodySize": 0
                },
                "response": {
                    "status": 502,
                    "statusText": "Bad Gateway",
                    "httpVersion": "HTTP/1.1",
                    "cookies": [],
                    "headers": [
                        {
                            "name": "content-length",
                            "value": "216"
                        }
                    ],
                    "content": {
                        "size": 216,
                        "compression": 0,
                        "mimeType": "",
                        "text": "Cannot connect to dhgyhuim.pretendo.cc:443: [Errno -2] Name or service not known If you plan to redirect requests away from this server, consider setting `connection_strategy` to `lazy` to suppress early connections."
                    },
                    "redirectURL": "",
                    "headersSize": 36,
                    "bodySize": 216
                },
                "cache": {},
                "timings": {
                    "connect": -1.0,
                    "ssl": -1.0,
                    "send": 0,
                    "receive": 0.37479400634765625,
                    "wait": 0
                }
            },
```

This causes visual parsing issues on other HAR parsers such as Firefox or harviewer, and in mitmproxy shows a URL parsing error, and won't try to parse any following dumps on the file:

```
[12:42:38.792] Uncaught exception POST /flows/dump (127.0.0.1)
HTTPServerRequest(protocol='http', host='127.0.0.1:8081', method='POST', uri='/flows/dump', version='HTTP/1.1', remote_ip='127.0.0.1')
Traceback (most recent call last):
  File "/home/daniel/Sources/mitmproxy/mitmproxy/io/io.py", line 53, in stream
    yield request_to_flow(request_json)
[...]

PR #6599, opened 2024-01-13
fix: save failed connect request urls to har files correctly

basedBaba:
#### Description
Fix issue #6579

This saves failed CONNECT request URLs to HAR files correctly.

#### Checklist

### DIFF
--- a/test/mitmproxy/addons/test_savehar.py
+++ b/test/mitmproxy/addons/test_savehar.py
@@ -175,7 +175,9 @@ def test_savehar(log_file: Path, tmp_path: Path, monkeypatch):
 
 def test_flow_entry():
     s = SaveHar()
-    flow = tflow.twebsocketflow(method="CONNECT", host="test.test", port=443, scheme="https")
+    flow = tflow.twebsocketflow(
+        method="CONNECT", host="test.test", port=443, scheme="https"
+    )
     servers_seen: set[Server] = set()
 
     flow_entry = s.flow_entry(flow, servers_seen)
--- a/test/mitmproxy/addons/test_savehar.py
+++ b/test/mitmproxy/addons/test_savehar.py
@@ -174,6 +174,7 @@ def test_savehar(log_file: Path, tmp_path: Path, monkeypatch):
 
 
 def test_flow_entry():
+    """https://github.com/mitmproxy/mitmproxy/issues/6579"""
     s = SaveHar()
     req = Request.make("CONNECT", "https://test.test/")
     flow = tflow.tflow(req=req)
--- a/test/mitmproxy/addons/test_savehar.py
+++ b/test/mitmproxy/addons/test_savehar.py
@@ -173,6 +173,17 @@ def test_savehar(log_file: Path, tmp_path: Path, monkeypatch):
     assert actual_har == expected_har
 
 
+def test_flow_entry():
+    s = SaveHar()
+    flow = tflow.twebsocketflow(method="CONNECT", host="test.test", port=443, scheme="https")
+    servers_seen: set[Server] = set()
+
+    flow_entry = s.flow_entry(flow, servers_seen)
+
+    if flow_entry["request"]["method"] == "CONNECT":
+        assert flow_entry["request"]["url"].startswith("https")
+
+
 class TestHardumpOption:
     def test_simple(self, capsys):
         s = SaveHar()
--- a/test/mitmproxy/addons/test_savehar.py
+++ b/test/mitmproxy/addons/test_savehar.py
@@ -175,9 +175,8 @@ def test_savehar(log_file: Path, tmp_path: Path, monkeypatch):
 
 def test_flow_entry():
     s = SaveHar()
-    flow = tflow.twebsocketflow(
-        method="CONNECT", host="test.test", port=443, scheme="https"
-    )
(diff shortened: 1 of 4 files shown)


DaniElectra:
The fix works as intended :+1: 

Only thing is you have to add a test to cover the problem so that it doesn't happen again

basedBaba:
I have added the relevant tests for the fix which required a few modifications in the sample flow generator `tflow.py`'s `twebsocketflow` function which allowed me to generate a sample flow using the `CONNECT` method and a non-existent domain for this particular case.

The checks also seem to pass now.

basedBaba:
@DaniElectra Hello! Please have a look at this.

test/mitmproxy/addons/test_savehar.py
```diff
@@ -173,6 +173,19 @@ def test_savehar(log_file: Path, tmp_path: Path, monkeypatch):
...
+    s = SaveHar()
+    flow = tflow.twebsocketflow(
+        method="CONNECT", host="test.test", port=443, scheme="https"
+    )
+    servers_seen: set[Server] = set()
+
+    flow_entry = s.flow_entry(flow, servers_seen)
+
+    if flow_entry["request"]["method"] == "CONNECT":
+        assert flow_entry["request"]["url"].startswith("https")
```
DaniElectra, line 186: This if statement feels redundant, as we already know the method will be CONNECT, as per the configuration above. The following should work too:

```suggestion
    assert flow_entry["request"]["url"].startswith("https")
```

DaniElectra (commented):
btw, I'm not a mitmproxy maintainer but a regular user, so we'd have to wait for an actual review by one of the maintainers so that this gets merged

test/mitmproxy/addons/test_savehar.py
basedBaba: That makes sense, thanks for the suggestion!

basedBaba:
@mhils Hello! Please have a look at this.

test/mitmproxy/addons/test_savehar.py
```diff
@@ -173,6 +173,18 @@ def test_savehar(log_file: Path, tmp_path: Path, monkeypatch):
     assert actual_har == expected_har
 
 
+def test_flow_entry():
+    s = SaveHar()
+    flow = tflow.twebsocketflow(
+        method="CONNECT", host="test.test", port=443, scheme="https"
```
mhils, line 179: A WebSocket flow shouldn"t have a CONNECT method, AFAIK that's typically GET with Upgrade headers? Or is this something you've seen in the wild?

mhils (commented):
Thank you two! Out of curiosity, could you explain to me how you even get CONNECT requests in mitmproxy? We typically don't surface them in the UI (or the event hooks), so I'm surprised to see them at all.

Otherwise this generally looks good, but I feel this could be a regular HTTP flow (and not a WebSocket one). In that case we don't need the changes to `tflow.py`. :)

test/mitmproxy/addons/test_savehar.py
DaniElectra: You are right, I think `ttcpflow` should be used instead? Correct me if I'm wrong

mhils: This should be an HTTP flow. :)

basedBaba: Due to my limited knowledge and understanding of the code-base, I looked at some other tests in `test_savehar.py` and concluded that `twebsocketflow` can be used for generating a sample flow without knowing it's actual use. My apologies for that.

As you said, no changes are required in `tflow.py`.

Instead, we can use `tflow` and pass a `CONNECT` request in order to generate a sample HTTP flow for testing.

test/mitmproxy/addons/test_savehar.py
```diff
@@ -173,6 +173,17 @@ def test_savehar(log_file: Path, tmp_path: Path, monkeypatch):
     assert actual_har == expected_har
 
 
+def test_flow_entry():
```
mhils, line 176:
```suggestion
def test_flow_entry():
    """https://github.com/mitmproxy/mitmproxy/issues/6579"""
```

mhils (commented):
Thanks! Let's add a ref for the original issue and fix the CHANGELOG conflict, then this is good to go. :)

I'd also still like to get an answer to my question in https://github.com/mitmproxy/mitmproxy/pull/6599#pullrequestreview-1830638895: How  do you get CONNECT requests visible within mitmproxy?

mhils:
> Based on how much I understand the working of `mitmproxy`, when we send a request to a non-existent domain with the `https` protocol, `mitmproxy` first sends a `CONNECT` request to the given non-existent domain, which itself fails, resulting in the connection request being terminated, and the failed `CONNECT` request appearing on the client side. I could be completely wrong here though. :/

Can you clarify what exactly you are doing / which proxy mode you are using?

mhils (approved):

basedBaba:
This is what I was trying to do.

image

image

mhils:
Thanks for the details, that's a great repro!

DaniElectra:
Thank you all! @mhils can you also fix the changelog and put this PR on `mitmproxy next`?

mhils:
Good catch, thanks! 🍰

PR #6653, opened 2024-02-09
Fix Multipart content view parsing

DaniElectra:
#### Description

On get_message_content_view, the content type wasn't including the boundary, and was only setting the MIME type. This made the multipart content view unusable, as the boundary was required on parsing. To fix the issue, we assign the full content type instead.

This wasn't triggered by any previous tests because they would test against the multipart parser directly, and not the generic parser.

#### Checklist

### DIFF
--- a/mitmproxy/contentviews/__init__.py
+++ b/mitmproxy/contentviews/__init__.py
@@ -130,8 +130,8 @@ def get_message_content_view(
     if isinstance(message, http.Message):
         http_message = message
         if ctype := message.headers.get("content-type"):
-            if ct := http.parse_content_type(ctype):
-                content_type = f"{ct[0]}/{ct[1]}"
+            if http.parse_content_type(ctype):
+                content_type = ctype
 
     tcp_message = None
     if isinstance(message, TCPMessage):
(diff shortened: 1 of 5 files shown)


mitmproxy/contentviews/__init__.py
```diff
@@ -130,8 +130,8 @@ def get_message_content_view(
     if isinstance(message, http.Message):
         http_message = message
         if ctype := message.headers.get("content-type"):
-            if ct := http.parse_content_type(ctype):
-                content_type = f"{ct[0]}/{ct[1]}"
+            if http.parse_content_type(ctype):
+                content_type = ctype
```
mhils, line 134: It looks like our test coverage is bad here because this is not caught, but this will break a few contentviews which currently do simple string comparisons. 

Can you pull the full content type header from `http_message.headers` in the multipart contentview instead? (http_message is passed as part of in `**metadata`). Thanks!

mhils (commented):
Thanks for catching this! :cake:

DaniElectra:
I'm not sure why the check is failing?

mhils:
https://github.com/mitmproxy/mitmproxy/actions/runs/7878070568/job/21495523975?pr=6653#step:5:74

Test is failing because `test_multipart` is not covering line 17. Only test_api does. We are (a bit too) strict here...

Cheap fix is to list http_message directly as an arg and then not have any branches: https://github.com/mitmproxy/mitmproxy/blob/1a02ebb89f6765d827f2fe0086dfe5960eb6e093/mitmproxy/contentviews/query.py#L8-L12

Thanks for the quick RTT! 🍰

DaniElectra:
This PR should be good now!

Issue #6656, opened 2024-02-12
Wireguard config has wrong endpoint when two active NICs


cybacoffee:
#### Problem Description
When running `mitmweb`in `wireguard` mode with more then one active network interface (two IPs/subnets), and running the `--listen-host` flag with fixing the IP from NIC2, the wireguard config (and QR) is generated with `endpoint <NIC1>`.
One has two manually modify the wireguard config after importing via e.g. QR code.
But the `listening` on NIC2 works as it should, only the config is generated wrong.

#### Steps to reproduce the behavior:
1. Run two network interfaces (e.g. primary NIC1=192.168.1.10/24, NIC2=192.168.2.10/24)
2. mitmweb --mode wireguard --listen-host 192.168.2.10/24
3. mitm wireguard config is generated with endpoint 192.168.1.10/32.

#### System Information
```
~ mitmweb --version
Mitmproxy: 10.2.2
Python:    3.12.1
OpenSSL:   OpenSSL 3.1.4 24 Oct 2023
Platform:  macOS-14.3.1-arm64-arm-64bit
```

mhils:
Thanks for the report! The relevant code is [here](https://github.com/mitmproxy/mitmproxy/blob/10.2.2/mitmproxy/proxy/mode_servers.py#L390), contributions welcome! :)

PR #6659, opened 2024-02-13
fix: generate wireguard config with correct endpoint when using two or more active nics

basedBaba:
#### Description

Fix issue #6656

This generates a wireguard config with the correct endpoint when using two or more active NICs.
 
#### Checklist


mhils (approved):
Thanks! 🍰

PR #6668, opened 2024-02-16
Fix bug scripts in Mitmweb

lups2000:
#### Description

This PR should fix issue: #6002 

#### Checklist

### DIFF
--- a/web/src/js/ducks/options.ts
+++ b/web/src/js/ducks/options.ts
@@ -33,15 +33,20 @@ export default function reducer(state = defaultState, action): OptionsState {
 }
 
 export async function pureSendUpdate(option: Option, value, dispatch) {
-    try {
-        const response = await fetchApi.put("/options", { [option]: value });
-        if (response.status === 200) {
-            dispatch(optionsEditorActions.updateSuccess(option));
-        } else {
-            throw await response.text();
+    //value is an array of strings, we need to check that the last element( path ) is not empty
+    if (value[value.length - 1] !== "") {
+        try {
+            const response = await fetchApi.put("/options", {
+                [option]: value,
+            });
+            if (response.status === 200) {
+                dispatch(optionsEditorActions.updateSuccess(option));
+            } else {
+                throw await response.text();
+            }
+        } catch (error) {
+            dispatch(optionsEditorActions.updateError(option, error));
         }
-    } catch (error) {
-        dispatch(optionsEditorActions.updateError(option, error));
     }
 }
 


mhils (commented):
Thanks for the PR! This looks like it fixes the immediate bug, but I'm not a big fan of the implementation. It is a bit unexpected that a method named `pureSendUpdate` silently modifies the `value` it's been passed. Can we 1) fix the UI edit component instead, and 2) add a matching test? :)

lups2000:
yes I agree, initially my idea was to put a check in the component `Option.jsx`, but for some implementation reasons it didn't work and I couldn't insert more than 1 path (basically I wasn't able to start a new line). Is that what you mean?

lups2000:
Hello @mhils , is this acceptable ?

lups2000:
Hello @mhils , I tried again to integrate the check in `Option.jsx` but it's not straightforward since here we simply propagate the value and call `updateOptions`. I found another solution: basically put the check in the function `update` in `options.ts`. Is it acceptable? 
Something like this: 
`export function update(name: Option, value: any): AppThunk {
    return (dispatch) => {
        console.log(value);
        dispatch(optionsEditorActions.startUpdate(name, value));
        if (value[value.length - 1] !== "") { //we send the value to the backend only if it is not empty
            console.log("send value to backend");
            sendUpdate(name, value, dispatch);
        }
    };
}`

lups2000:
sorry I misclicked the button , and I accidentally closed the pr

mhils:
> but it's not straightforward since here we simply propagate the value and call `updateOptions`

Yeah, I think that's what we need to change! How about we change `StringSequenceOption` so that it keeps some internal state but only propagates non-empty strings? That feels cleanest to me.

lups2000:
> > but it's not straightforward since here we simply propagate the value and call `updateOptions`
> 
> Yeah, I think that's what we need to change! How about we change `StringSequenceOption` so that it keeps some internal state but only propagates non-empty strings? That feels cleanest to me.

yes I totally agree with you, but I'm still having issues and I am not able to add more than one line(that's is due to the fact that we modify the value by removing the ""). So, basically it works if we don't filter for empty strings (but the initial bug is still there) or if remove `value={value.join("\n")}` (but of course in this case we don't have persistence).

it's quite challenging :) maybe my approach is wrong but I couldn't find a way to solve the problem in this component, at least without changing the logic behind it. 

 That's my change now: 

`function StringSequenceOption({ value, onChange, ...props }) {
    const height = Math.max(value.length, 1);

    return (
        <textarea
            rows={height}
            value={value.join("\n")}
            onChange={(e) =>
                onChange(
                    e.target.value
                        .split("\n")
                        .map((line) => line.trim())
                        .filter((line) => line !== "")
                )
            }
            {...props}
        />
    );
}`

lups2000:
I probably found a solution :) I'll push later

mhils (commented):
This looks awesome now, thanks! 😃

mhils (approved):
Thanks! 😃

PR #5996, opened 2023-03-18
Tree View for Flow items

RG7279805:
#### Description
closes #5627
I am starting this draft PR so you can start reviewing the code right now 
There are a lot of things to be done yet but a very basic prototype is done 

#### Checklist
-  [ ] Add Flow sorting for table
-  [ ] Add onClick Flow event
 - [ ] I have updated tests where applicable.
 - [ ] I have added an entry to the CHANGELOG.

I will be updating the code and features that are missing

Issue #5627, opened 2022-10-04
[New Feature Request] Tree View


reisraff:
#### Problem Description

Today we have the web interface, but we have a view of calls that is super hard do locate resources...

#### Proposal
The Idea is to add the feature maybe a feature toggle in where you can chose the view  you prefer, those being... the actual view that we have, and the requested feature in this issue that would be the **tree view**

reisraff:
I am not sure what framework mitmproxy uses for frontend, but I've done a little search and I found some stuff that may will help:

- https://mui.com/material-ui/react-tree-view/ (react component that resolves the tree view)
- https://stackoverflow.com/questions/65464890/how-to-convert-an-array-of-urls-into-a-tree-folder-structure-data (some js algorithm that resolves the "given a URL transform this in a folder structured data)

Prinzhorn:
Sounds like a great idea!

> I am not sure what framework mitmproxy uses for frontend

React. Code and instruction are here if you want to contribute https://github.com/mitmproxy/mitmproxy/tree/main/web

raychanks:
Hi there, I would like to give this a try.

raychanks:
@reisraff Oh I thought no one is working on this at the moment. If you're already on it, I can try to find other issues to contribute to 👍🏼. Don't want to waste your effort if there's already WIP, what do you say?

reisraff:
@raychanks go ahead.. I am not in this.

RG7279805:
@reisraff @raychanks if none of you are woking on this i would take this issue and give a PR

raychanks:
@RG7279805 I totally forgot about this. I haven't work on this yet, feel free to take it:+1:

lups2000:
Could I have a look at this?

lups2000:
> > Could I have a look at this?
> 
> There is already a PR #5996 Would be great if you can check it out locally and leave feedback on it.

Hi, @Prinzhorn . yes I saw it and I am taking inspiration from it. However, the logic to filter the flows is missing. My idea for that consists in reusing the component `FlowTable`. So basically we group the flows by host and then inject a `FlowTable` for each group. makes sense?

PR #6673, opened 2024-02-20
Feature/flow-tree-view

lups2000:
#### Description

This PR should implement the flows tree view feature requested here: #5627. For the initial structure I took inspiration from the work of @Rishabhg71 (#5996). Then, I tried to reuse the component `FlowTable` that handles a lot of stuff already such as the sorting, selection, highlighting etc. 
I am opening this draft PR so you can check what I am doing. 

https://github.com/mitmproxy/mitmproxy/assets/100372313/ea60af1a-1914-4112-8f6a-57d494ceb55c

#### Checklist

 - [ ] Adjust UI
 - [ ] Adjust sorting since now is "global"


lups2000:
Hello @mhils , I saw the sorting has been initially implemented in the state, so for the tree view now is "global" (as we can see in the recording). Basically, if I sort by time in the first table I sort by time in the other tables as well, which is not the best. What do you think?  Should we consider refining this behavior? In case of change, I must understand properly the logic behind it to avoid to break something :)

Prinzhorn:
Judging from the video this looks interesting and I can see this being helpful. But that's not a tree and you can achieve similar things by filtering for a given host. The way I understand the linked issue is that if you have hundreds of flows in deeply nested routes such as `/api/*/*/*/*` there is no high-level way to view this structure to understand it and look at sub-routes etc.

mhils:
I'd be inclined to say that this should be possible without a third-party library. 

But more importantly, the flow table is a really core UX element and we shouldn't rush to implement things here. For example, it could also make sense to have this tree in a dedicated sidebar with checkboxes for individual path parts. Have we looked at what other tools do here in terms of UX?

lups2000:
> I'd be inclined to say that this should be possible without a third-party library.
> 
> But more importantly, the flow table is a really core UX element and we shouldn't rush to implement things here. For example, it could also make sense to have this tree in a dedicated sidebar with checkboxes for individual path parts. Have we looked at what other tools do here in terms of UX?

yes regarding the third-party library it's fine! For the tree, you mean that we could have a tree on one side where we can select the paths we want to display in the table?
Ex. 

- Mitmproxy.com 
       - Mitmproxy.com/1
             -Mitmproxy.com/1/1  
       - Mitmproxy.com/2 
- example.com
- google.com

we can select the parent Mitmproxy and display just this subtree in the table? basically acting like a filter. correct ?

mhils:
Yes, basically that. I'm not saying it's a better UI/UX, just pointing out there are other options and we should look at existing work. :)

lups2000:
> Yes, basically that. I'm not saying it's a better UI/UX, just pointing out there are other options and we should look at existing work. :)

yes yes, you're right. I can define a prototype and then send you the result to get a feedback :)

mhils:
My suggestion would be to look at what others are doing and write that up. Before writing a single line of code. 😉

lups2000:
@mhils So far, I've only encountered examples where they provide the possibility to switch between the two modalities through a button. That's also the way I prefer the most! However, from what I've seen, it's quite challenging to reuse some of the table components because they are specifically designed for use with a table. So, if we want to proceed in this manner, we'll need to implement some new components and handle the filters in a different way(this is probably the hardest part). Otherwise, we can move forward with your idea, which seems a bit easier to implement and probably won't require too many changes (we just need to decide where to display this tree with the checkboxes).

Prinzhorn:
> So far, I've only encountered examples where they provide the possibility to switch between the two modalities through a button.

Could you please share screenshots and name the tools? Because all I've ever seen (e.g. Burp, ZAP, Charles) has this as a sidebar that allows you to filter the table. I can't really imagine how what you describe is supposed to work. The tree can have an arbitrary number of levels and on each level you could have a mix of sub-levels and then also the flows at the current level? How does the UX work? It would also be hard to virtualize this list mix I would assume.

Issue #4233, opened 2020-10-14
Crash when deleting temporary file


cweiske:
#### Problem Description
mitmproxy 5.1.1 crashed when I opened the request body in an external editor and then moved the temporary file to another location while the editor was open. When I exited the editor, mitmproxy probably wanted to clean up and delete the file and crashed.

I moved the file to another location because i wanted to use its contents for further analysis.

#### Steps to reproduce the behavior:
1. Run mitmproxy until a flow is visible
2. Show the flow details
3. Press "v" ("View flow body in an external viewer") and select "request".
4. Editor opens
5. Delete the temporary file
6. Close the editor
7. mitmproxy crashes

```
$ mitmproxy --mode transparent --showhost
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/mitmproxy/master.py", line 86, in run_loop
    loop()
  File "/usr/lib/python3/dist-packages/urwid/main_loop.py", line 287, in run
    self._run()
  File "/usr/lib/python3/dist-packages/urwid/main_loop.py", line 385, in _run
    self.event_loop.run()
  File "/usr/lib/python3/dist-packages/urwid/main_loop.py", line 1494, in run
    reraise(*exc_info)
  File "/usr/lib/python3/dist-packages/urwid/compat.py", line 58, in reraise
    raise value
  File "/usr/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/lib/python3/dist-packages/urwid/raw_display.py", line 416, in <lambda>
    wrapper = lambda: self.parse_input(
  File "/usr/lib/python3/dist-packages/urwid/raw_display.py", line 515, in parse_input
    callback(processed, processed_codes)
  File "/usr/lib/python3/dist-packages/urwid/main_loop.py", line 412, in _update
    self.process_input(keys)
  File "/usr/lib/python3/dist-packages/urwid/main_loop.py", line 513, in process_input
    k = self._topmost_widget.keypress(self.screen_size, k)
  File "/usr/lib/python3/dist-packages/mitmproxy/tools/console/window.py", line 310, in keypress
    k = super().keypress(size, k)
  File "/usr/lib/python3/dist-packages/urwid/container.py", line 1135, in keypress
    return self.body.keypress( (maxcol, remaining), key )
  File "/usr/lib/python3/dist-packages/mitmproxy/tools/console/window.py", line 44, in keypress
    ret = super().keypress(size, key)
[...]

PR #6677, opened 2024-02-20
Fix file deletion crash if file is already gone

androids-electric-sheep:
#### Description

In the event the file is no longer present, mitmproxy raises an exception in the deleting thread which is displayed in the TUI. Update to make the `os.unlink` call tolerant to file non-existence. Based on https://github.com/mitmproxy/mitmproxy/issues/4233

#### Checklist

 - [ :green_circle: ] I have updated tests where applicable.
 - [ :green_circle: ] I have added an entry to the CHANGELOG.

### DIFF
--- a/mitmproxy/tools/console/master.py
+++ b/mitmproxy/tools/console/master.py
@@ -33,6 +33,13 @@ from mitmproxy.tools.console import window
 T = TypeVar("T", str, bytes)
 
 
+def try_unlink(name: str) -> None:
+    try:
+        os.unlink(name)
+    except:
+        return
+
+
 class ConsoleMaster(master.Master):
     def __init__(self, opts: options.Options) -> None:
         super().__init__(opts)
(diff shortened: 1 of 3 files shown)


mitmproxy/tools/console/master.py
```diff
@@ -33,6 +33,13 @@
 T = TypeVar("T", str, bytes)
 
 
+def try_unlink(name: str) -> None:
+    try:
+        os.unlink(name)
+    except Exception:
```
androids-electric-sheep, line 39: Thought about whether this should only catch a `FileNotFoundError` so as not to hide other errors. Linting requires the except not be bare at minimum so put `Exception` for now

mhils: FileNotFoundError makes mose sense here I think.

test/mitmproxy/tools/console/test_master.py
```diff
@@ -0,0 +1,10 @@
+import os
+import tempfile
+
+from mitmproxy.tools.console import master
+
+
+def test_try_unlink_file_not_exists():
```
mhils, line 7: Can we use https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html please? Cleaning up after mkdtemp is a bit annoying.

mhils (commented):
Thanks for the PR and sorry for the slow turnaround time!

Issue #4448, opened 2021-02-15
View options menu doesn't display overflowed part


mikelei8291:
#### Problem Description
![Screenshot](https://user-images.githubusercontent.com/9620335/107893232-e4503e00-6f21-11eb-862c-ae6e8746d79e.png)

As shown in the screenshot above, the overflowed part of the view options menu doesn't get displayed if the flow detail section is too short.

Removing the `overflow` property on `.flow-detail` and the `overflow-y` property on `.flow-detail section` fixes the issue, but will conflict with the fix of issue #2083. Maybe find a different solution for that issue, or set the `overflow` property of `dropdown-menu` to `auto` and limit the height of it by set its `top` property to a fixed value, like `-300px`.

#### Steps to reproduce the behavior:
1. Open a flow detail that is short enough
2. Open the view options menu
3. The menu get displayed as shown in the screenshot above

#### System Information
Mitmproxy: 6.0.2
Python: 3.9.1
OpenSSL: OpenSSL 1.1.1i  8 Dec 2020
Platform: Windows-10-10.0.16299-SP0
Firefox: 86.0b9

Prinzhorn:
Thanks for reporting this bug! I think another option might be to make the dropdown `position: fixed` to avoid all that trouble. I don't know how that Bootstrap version works, recent ones use https://popper.js.org/ to intelligently position such elements.

PRs are welcome if you find a good solution.

PR #6697, opened 2024-02-29
Fix Bug view options menu

lups2000:
#### Description

This PR fixes bug described here : #4448 

I set a max-height property and a scroll in case of overflow on the y-axis.

#### Checklist


mhils (approved):
Thanks! 🍰

PR #6711, opened 2024-03-04
Allow --allow-hosts and --ignore-hosts to work together

dstd:
#### Description

The mutual exclusivity of the allow-hosts and ignore-hosts parameters looks like an unnecessary obstacle and does not make much sense.

It is very convenient to use a proxy only for the domain of your service, but at the same time ignore some subdomains, especially when they serve some kind of CDNs with a large amount of data.

Although this filtering could be implemented using regexp with negative lookahead, but it complicates configuration and is not as clear as conjuction of allow and deny filters.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 * Allow runtime modifications of HTTP flow filters for server replays
   ([#6695](https://github.com/mitmproxy/mitmproxy/pull/6695), @errorxyz)
 * Allow --allow-hosts and --ignore-hosts to work together
+  ([#6711](https://github.com/mitmproxy/mitmproxy/pull/6711), @dstd)
 
 
 ## 21 January 2024: mitmproxy 10.2.2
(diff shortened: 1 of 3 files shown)


mhils (commented):
Thanks for the PR! I'm happy with the change, but the current logic still feels more complex than it needs to be. Do you think the following would work?

```python
if ctx.options.allow_hosts:
    not_allowed = not any(
        re.search(rex, host, re.IGNORECASE)
        for host in hostnames
        for rex in ctx.options.allow_hosts
    )
    if not_allowed:
        return True

if ctx.options.ignore_hosts:
    ignored = any(
        re.search(rex, host, re.IGNORECASE)
        for host in hostnames
        for rex in ctx.options.ignore_hosts
    )
    if ignored:
        return True

return False
```

dstd:
> but the current logic still feels more complex than it needs to be. Do you think the following would work?

Yes, I think it will work. I was confused by the else branch with AssertionError. But according to the beginning of the function, it doesn't seem to make sense

mhils (approved):
Awesome, thanks! 😃 🍰

PR #6720, opened 2024-03-07
Add `--show-ignored-hosts`

### LINKED ISSUE #6421: Log ignored requests (`--ignore-hosts`) that are forwarded
#### Problem Description

Requests that go through `--ignore-hosts` are not logged as a request. This makes it inconvenient to know if any have happened even if you can't see their contents.

#### Proposal
Make it possible to see ignored requests in the request log of mitmproxy.

#### Additional context
It's sometimes necessary to use that with stuff that you can't make ignore SSL verification errors, but still want to go through the mitmproxy for any reason.

NicolaiSoeborg:
#### Description

Maybe a bit counterintuitive, but mitmproxy is very nice even without the MITM part.  When doing `--ignore-hosts '.*'` it is not possible to see SNI's, so add new flag to show the raw TCP/UDP streams.

Fixes #6421

Running `mitmproxy --ignore-hosts '.*' --show-ignored-hosts`:

image

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@
 
 ## Unreleased: mitmproxy next
 
-- Add parameter `--show-ignored-hosts` for showing SNI for ignored hosts.
+- Add `show_ignored_hosts` option to display ignored flows in the UI.
+  This option is implemented as a temporary workaround and will be removed in the future.
   ([#6720](https://github.com/mitmproxy/mitmproxy/pull/6720), @NicolaiSoeborg)
 - mitmproxy now supports transparent HTTP/3 proxying.
   ([#7202](https://github.com/mitmproxy/mitmproxy/pull/7202), @errorxyz, @meitinger, @mhils)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ## Unreleased: mitmproxy next
 
 - Add parameter `--show-ignored-hosts` for showing SNI for ignored hosts.
+  ([#6720](https://github.com/mitmproxy/mitmproxy/pull/6720), @NicolaiSoeborg)
 - mitmproxy now supports transparent HTTP/3 proxying.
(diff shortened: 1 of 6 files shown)


mhils:
User question: Do you ever have an interest in viewing the raw contents here?

I like how this gets us pretty close to what we want with very little effort. But ideally I'd really like to see a dedicated flow subclass that does not keep encrypted gibberish in memory and that allows for a better UI integration.

NicolaiSoeborg:
My use-case is just to easily see the SNI :)

True that keeping the (encrypted) request in mem might be a bit of a waste - I could look into `del`eting it if we want to do that?
A _dedicated flow type_ would be nice, but not sure I have the time :/

NicolaiSoeborg:
@mhils : Anything I can do to get this merged? :)

NicolaiSoeborg:
@mhils : I believe this PR actually don't keep unneeded (encrypted) raw traffic in memory.
I'll happily change the flag names of you have any suggestions - otherwise this PR should be ready for merging (and will add a new feature to mitmproxy)

borisspektor:
@NicolaiSoeborg will this feature be added to mitmweb as well? 🙏

mhils:
> @mhils : I believe this PR actually don't keep unneeded (encrypted) raw traffic in memory.

I am pretty sure it does. But I guess this is acceptable for a temporary workaround. I would still like us to move to dedicated flow subclasses, but for now may be good enough. Would you mind adding a changelog entry?

NicolaiSoeborg:
> @NicolaiSoeborg will this feature be added to mitmweb as well? 🙏

It appears to be working automagically:

image

> I am pretty sure it does. But I guess this is acceptable for a temporary workaround. I would still like us to move to dedicated flow subclasses, but for now may be good enough. Would you mind adding a changelog entry?

You are right, my bad! :)

Added a changelog line and rebased of master

borisspektor:
@NicolaiSoeborg can't wait for it to be released! :)
Is there an ETA on this?

mhils (approved):
Thanks again!

PR #6736, opened 2024-03-14
support for websockets in client playback

audi99:
early code and currently not capable of replaying msgs sequentially given timestamp info, but plan on it


mhils:
Thanks! Have you tested this?

PR #6766, opened 2024-03-30
Add primitive websocket interception and modification

errorxyz:
#### Description

This PR adds a feature(MVP) to allow for websocket interception and modification one at a time per flow. On the mitmweb side of things, the UI shows that websocket messages are being intercepted but there's no option to edit (yet?).
Websockets are intercepted by default if interception is on, to prevent websocket interception use `!~websocket`

#### Checklist

### DIFF
--- a/mitmproxy/tools/console/flowview.py
+++ b/mitmproxy/tools/console/flowview.py
@@ -119,7 +119,7 @@ class FlowDetails(tabs.Tabs):
         assert isinstance(flow, http.HTTPFlow)
 
         # prevent renaming when websocket messages are intercepted
-        if self.flow.intercepted and flow.response and not flow.websocket:
+        if self.flow.intercepted and flow.response and (not flow.websocket or len(flow.websocket.messages) == 0):
             return "Response intercepted"
         else:
             return "Response"
@@ -147,7 +147,14 @@ class FlowDetails(tabs.Tabs):
         return "UDP Stream"
 
     def tab_websocket_messages(self):
-        return "WebSocket Messages"
+        flow = self.flow
+        assert isinstance(flow, http.HTTPFlow)
+        assert flow.websocket
+
+        if self.flow.intercepted and len(flow.websocket.messages) != 0:
+            return "WebSocket Messages intercepted"
+        else:
+            return "WebSocket Messages"
 
     def tab_details(self):
         return "Detail"
--- a/mitmproxy/tools/console/flowview.py
+++ b/mitmproxy/tools/console/flowview.py
@@ -117,7 +117,9 @@ class FlowDetails(tabs.Tabs):
     def tab_http_response(self):
         flow = self.flow
         assert isinstance(flow, http.HTTPFlow)
-        if self.flow.intercepted and flow.response:
+
+        # prevent renaming when websocket messages are intercepted
+        if self.flow.intercepted and flow.response and not flow.websocket:
             return "Response intercepted"
         else:
             return "Response"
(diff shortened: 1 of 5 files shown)


test/mitmproxy/addons/test_intercept.py
```diff
@@ -89,3 +89,17 @@ async def test_udp():
         f = tflow.tudpflow()
         await tctx.cycle(r, f)
         assert not f.intercepted
+
+
+async def test_websocket_message():
+    r = intercept.Intercept()
+    with taddons.context(r) as tctx:
+        tctx.configure(r, intercept='~b "hello binary"')
```
errorxyz, line 97: Is there any better way to filter out websocket messages alone?
`~websocket` intercepts the response of the websocket upgrade request before intercepting websocket messages

mitmproxy/tools/console/flowview.py
```diff
@@ -117,7 +117,13 @@ def tab_http_request(self):
     def tab_http_response(self):
         flow = self.flow
         assert isinstance(flow, http.HTTPFlow)
-        if self.flow.intercepted and flow.response:
+
+        # prevent renaming when websocket messages are intercepted
+        if (
+            self.flow.intercepted
+            and flow.response
+            and (not flow.websocket or len(flow.websocket.messages) == 0)
```
errorxyz, line 125: Added this condition because the `Response` tab was renaming itself to `Response intercepted` whenever websocket messages were intercepted

errorxyz:
Do we need tests for changes to `mitmproxy/tools/console/consoleaddons.py`? Asking because we don't have tests for existing code in `mitmproxy/tools/console/consoleaddons.py` (`mitmproxy/tests/tools/console/test_consoleaddons.py` doesn't exist)
Same goes for changes to `flowview.py` - we don't have tests for `tab_http_request`, etc. Do we still want to add tests for `tab_websocket_messages`? Happy to add these if required

mitmproxy/tools/console/flowview.py
```diff
@@ -117,7 +117,13 @@ def tab_http_request(self):
...
+        ):
```
mhils, line 126:
```suggestion
        # there is not good way to detect what part of the flow is intercepted,
        # so we apply some heuristics to see if it's the HTTP response.
        response_is_intercepted = (
            self.flow.intercepted
            and flow.response
            and not (flow.websocket and flow.websocket.messages)
        )
        if response_is_intercepted:
```

 - the comment you added is useful in the context of your change, but if you look at this code without context it's a bit out of place
 - adding a variable for the boolean expression is a neat trick to convey intent of what the expression is supposed to check.

test/mitmproxy/addons/test_intercept.py
mhils: Not yet - I guess we should consider adding something like `~wq` and `~ws` that works analogous to `~q` and `~s` for WebSocket messages from the client and server.

For the test here this is fine. :)

mhils (commented):
Thanks, this looks very nice!

test/mitmproxy/addons/test_intercept.py
errorxyz: I've added these in my GSoC proposal ;)

errorxyz:
> Do we need tests for changes to `mitmproxy/tools/console/consoleaddons.py`? Asking because we don't have tests for existing code in `mitmproxy/tools/console/consoleaddons.py` (`mitmproxy/tests/tools/console/test_consoleaddons.py` doesn't exist) Same goes for changes to `flowview.py` - we don't have tests for `tab_http_request`, etc. Do we still want to add tests for `tab_websocket_messages`? Happy to add these if required
> 
> CI seems to be complaining because of these missing tests

@mhils added the suggested changes, how do we fix CI issues now?

mhils (approved):

mhils:
Thanks. Patch coverage is reasonable for what we are doing here, so this is good to go!

Issue #6721, opened 2024-03-07
 export websocket messages 


3052:
#### Problem Description

cannot export websocket message, can only view them

#### Steps to reproduce the behavior:
1. view websocket messages
2. try to export
3. can only export HTTP response body, which is something different

#### System Information

mitmproxy 10.2.4

thebigbone:
Can you give me an example? If I understand correctly, you are looking to export the websocket request body?

3052:
request and response

thebigbone:
> request and response

Isn't response body exportable by selecting the request and then using ```b``` filter?

thebigbone:
> > request and response
> 
> Isn't response body exportable by selecting the request and then using `b`?

3052:
@thebigbone the issue is not "is it possible to export websocket?" its not possible currently. stop this line of questioning now, this is your only warning.

also giving suggests without testing them is especially rude. if you had done so you would know that the result of your test is an empty file.

mhils:
@3052: Your point is correct, but let's be nice and assume good intent.

3052:
@mhils sorry man, but I am old enough that I have seen this type of person over and over again. they throw out half baked suggestions without testing them, putting the question asker on a wild goose chase wasting their time. its incredibly rude and I will not stand for it any more. thanks for the feedback.

txrp0x9:
Hi @3052 
I'd like to know if the below format works well for you as the saved file for websocket messages
```
[INCOMING] Hello client!
[OUTGOING] Hello server!
```
They will be in order as seen in the WebSocket Messages tab

3052:
yeah thats good

txrp0x9:
So I've investigated and found the root to be the cut.save command keybinding
```
    km.add(
        "b", "console.command cut.save @focus response.content ",
        ["flowlist", "flowview"],
        "Save response body to file"
    )
```
This keybind hardcodes the body as `response.content` and websockets store the body in `websocket.messages` leaving the response.content empty, which in turn leaves an empty file when ws connections are saved.

I've thought of two approaches:
1. Change the cut.save command, and include a simple if statement to check if the flow is a websocket and then if response.content is being saved, we instead save a formatted version from websocket.messages in the format of the previous message. This however seems hacky
[...]

PR #6767, opened 2024-04-01
Fix saving of websocket flows

txrp0x9:
#### Description
WebSocket flows when saved would return an empty file (cut.save) , and export raw command would not show the ws messages (#6721). This was because the cut.save command uses response.content and response.content is kept empty in websocket, with the actual content (messages) being saved in websocket.messages, while export raw has no support for ws messages.

~~This PR fixes this issue by storing formatted messages in response.content as they appear on websocket.messages
This also removes the "No content" default text on websocket response/request (If we don't do this, then response.content will be shown, but we already have WebSocket Messages tab so showing it again would be redundant)~~

This PR changes export addon and adds websocket message serialization support to the raw export format along with another "raw_messages" format that just saves the ws messages (can extend to tcp/udp messages).
This also adds a fix to cut.save keybind (b), by modifying the cut.save command to save websocket messages whenever request.content or response.content is being saved for a websocket flow.

#### Checklist

 - [✔️  ] I have updated tests where applicable.
 - [✔️  ] I have added an entry to the CHANGELOG.

### DIFF
--- a/mitmproxy/addons/cut.py
+++ b/mitmproxy/addons/cut.py
@@ -33,7 +33,7 @@ def extract(cut: str, f: flow.Flow) -> str | bytes:
     # Make "save body" keybind work for WebSocket flows.
     # Ideally the keybind would be smarter and this here can get removed.
     if (
-        isinstance(flow, http.HTTPFlow)
+        isinstance(f, http.HTTPFlow)
         and f.websocket
         and cut in ("request.content", "response.content")
     ):
--- a/mitmproxy/addons/cut.py
+++ b/mitmproxy/addons/cut.py
@@ -28,10 +28,12 @@ def is_addr(v):
 
 
 def extract(cut: str, f: flow.Flow) -> str | bytes:
+    # Hack for https://github.com/mitmproxy/mitmproxy/issues/6721:
+    # Make "save body" keybind work for WebSocket flows.
+    # Ideally the keybind would be smarter and this here can get removed.
     if (
-        hasattr(f, "websocket")
-        and f.websocket is not None
-        and (cut == "response.content" or cut == "request.content")
+        getattr(f, "websocket", None)
+        and cut == "response.content"
     ):
         formatted_messages = f.websocket.get_formatted_messages()
         return formatted_messages
--- a/mitmproxy/addons/cut.py
+++ b/mitmproxy/addons/cut.py
@@ -28,7 +28,11 @@ def is_addr(v):
 
 
 def extract(cut: str, f: flow.Flow) -> str | bytes:
-    if hasattr(f, "websocket") and f.websocket is not None and (cut == "response.content" or cut == "request.content"):
+    if (
+        hasattr(f, "websocket")
+        and f.websocket is not None
+        and (cut == "response.content" or cut == "request.content")
+    ):
         formatted_messages = f.websocket.get_formatted_messages()
         return formatted_messages
 
--- a/mitmproxy/addons/cut.py
+++ b/mitmproxy/addons/cut.py
@@ -28,6 +28,14 @@ def is_addr(v):
 
 
 def extract(cut: str, f: flow.Flow) -> str | bytes:
+    if hasattr(f, "websocket") and f.websocket is not None and (cut == "response.content" or cut == "request.content"):
+        formatted_messages = b""
+        for i, message in enumerate(f.websocket.messages):
+            formatted_messages += message.format_ws_message()
+            if i < len(f.websocket.messages) - 1:
+                formatted_messages += b"\n"
+        return formatted_messages
+    
     path = cut.split(".")
     current: Any = f
     for i, spec in enumerate(path):
--- a/mitmproxy/addons/cut.py
+++ b/mitmproxy/addons/cut.py
@@ -31,7 +31,7 @@ def extract(cut: str, f: flow.Flow) -> str | bytes:
     if hasattr(f, "websocket") and f.websocket is not None and (cut == "response.content" or cut == "request.content"):
         formatted_messages = f.websocket.get_formatted_messages()
         return formatted_messages
-    
+
     path = cut.split(".")
     current: Any = f
     for i, spec in enumerate(path):
--- a/mitmproxy/addons/cut.py
+++ b/mitmproxy/addons/cut.py
@@ -29,11 +29,7 @@ def is_addr(v):
 
 def extract(cut: str, f: flow.Flow) -> str | bytes:
     if hasattr(f, "websocket") and f.websocket is not None and (cut == "response.content" or cut == "request.content"):
-        formatted_messages = b""
-        for i, message in enumerate(f.websocket.messages):
-            formatted_messages += message.format_ws_message()
-            if i < len(f.websocket.messages) - 1:
-                formatted_messages += b"\n"
+        formatted_messages = f.websocket.get_formatted_messages()
         return formatted_messages
     
     path = cut.split(".")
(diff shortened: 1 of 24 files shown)


mhils (changes requested):
Thanks for the PR!

> This PR fixes this issue by storing formatted messages in response.content as they appear on websocket.messages

Architecturally that doesn't fly, sorry. WebSocket messages belong in `flow.websocket.messages`, we definitely do not want to duplicate them in `flow.response.content`. That's semantically incorrect and also doubles the size of saved flows.

I'm afraid this needs changes to the cut addon. I'm also open to UX improvements w.r.t. displaying "no content", but we can't just go and butcher response.content. :)

txrp0x9:
> I'm afraid this needs changes to the cut addon. I'm also open to UX improvements w.r.t. displaying "no content", but we can't just go and butcher response.content. :)

Yeah, I was honestly confused about what approach to follow but thats okay, I'll redo this and change cut addon instead. I've mentioned 2 approaches I thought of [here](https://github.com/mitmproxy/mitmproxy/issues/6721#issuecomment-2028641860), this PR followed approach 2. Can you take a look at approach 1 and tell me if its on the right path? It modifies the cut addon

txrp0x9:
I've added websocket support to the export command as you suggested @mhils , with test coverage
However for the cut.save keybinding (b), I've done a simple patch that saves websocket messages whenever cut.save is called on a websocket flow with response.content or request.content arguments. I do not seem to see another way unless we make keybindings dynamic (dependant on focused flow) to remove the hardcoded "response.content" in the binding.

txrp0x9:
@mhils I've added support for websocket message serialization in the raw format for the export addon. I've also added another new format called "websocket" which only saves the websocket messages (if the flow is a ws).
For the cut.save keybind (b), since response.content is hardcoded in the keybind itself, I've added code to the cut.save command that saves websocket messages whenever response.content or request.content is used with ws flows in cut.save arguments.

mitmproxy/addons/cut.py
```diff
@@ -28,6 +28,14 @@ def is_addr(v):
 
 
 def extract(cut: str, f: flow.Flow) -> str | bytes:
+    if (
+        hasattr(f, "websocket")
+        and f.websocket is not None
+        and (cut == "response.content" or cut == "request.content")
+    ):
```
mhils, line 35:
```suggestion
    # Hack for https://github.com/mitmproxy/mitmproxy/issues/6721:
    # Make "save body" keybind work for WebSocket flows.
    # Ideally the keybind would be smarter and this here can get removed.
    if (
        getattr(f, "websocket", None)
        and cut == "response.content"
    ):
```

mitmproxy/addons/export.py
```diff
@@ -130,20 +137,29 @@ def raw(f: flow.Flow, separator=b"\r\n\r\n") -> bytes:
...
+    if is_websocket_flow:
+        logging.info("Saving websocket messages")
+        serialized_flow = serialized_flow.join([separator, websocket_command(f)])
+    return serialized_flow
 
 
 formats: dict[str, Callable[[flow.Flow], str | bytes]] = dict(
     curl=curl_command,
     httpie=httpie_command,
+    websocket=websocket_command,
```
mhils, line 162: `raw_messages` maybe? We can then extend it to work for TCP and UDP flows as well.

Ideally we take this out here and handle it in a separate follow-up PR.

mitmproxy/websocket.py
```diff
@@ -92,6 +92,16 @@ def set_state(self, state: WebSocketMessageState) -> None:
         ) = state
         self.type = Opcode(typ)
 
+    def format_ws_message(self):
+        message = self.content
+        prefix = b""
+        if self.from_client:
+            prefix = b"[OUTGOING]"
+        else:
+            prefix = b"[INCOMING]"
```
mhils, line 101: Is this some known standard, or is this something we're making up here?

In the latter case I'd favor making this `_format_ws_message` (and same below), I don't want to commit here to a public API prematurely.

mitmproxy/addons/export.py
```diff
@@ -130,20 +137,29 @@ def raw(f: flow.Flow, separator=b"\r\n\r\n") -> bytes:
...
-        return raw_request(f)
+        serialized_flow = raw_request(f)
     elif response_present:
-        return raw_response(f)
+        serialized_flow = raw_response(f)
     else:
         raise exceptions.CommandError("Can't export flow with no request or response.")
+    if is_websocket_flow:
+        logging.info("Saving websocket messages")
+        serialized_flow = serialized_flow.join([separator, websocket_command(f)])
```
mhils, line 155: I think you misunderstand how `.join` works in Python. At least it's very unidiomatic.

mitmproxy/websocket.py
```diff
@@ -171,3 +181,11 @@ class WebSocketData(serializable.SerializableDataclass):
...
     def __repr__(self):
         return f"<WebSocketData ({len(self.messages)} messages)>"
+
+    def get_formatted_messages(self):
+        formatted_messages = b""
+        for i, message in enumerate(self.messages):
+            formatted_messages += message.format_ws_message()
+            if i < len(self.messages) - 1:
+                formatted_messages += b"\n"
+        return formatted_messages
```
mhils, line 191:
```suggestion
    def get_formatted_messages(self) -> bytes:
        return b"\n".join(
            m.format_ws_message()
            for m in self.messages
        )
```
```

File: mitmproxy/addons/cut.py [hunk 6]
```diff
@@ -28,6 +28,14 @@ def is_addr(v):
...
+        formatted_messages = f.websocket.get_formatted_messages()
+        return formatted_messages
```
mhils, line 37:
```suggestion
        return f.websocket.get_formatted_messages()
```

and this needs a test please. :)

mitmproxy/addons/cut.py
txrp0x9:
```
hasattr(f, "websocket")
and f.websocket is not None
```
Both above lines seems to be required, there is a test which fails otherwise. In the test scenario, websocket attribute exists but is `None`. Your change includes getattr(f, "websocket", None), but that just seems checks if the websocket attribute is present (otherwise returns None as default)

txrp0x9:
I've added complete test coverage and added changes you suggested @mhils

mitmproxy/addons/cut.py
mhils: > otherwise returns None as default

That is true, but `bool(None)` evaluates to false and the condition fails. :)


The intention behind just doing it for response.content was to limit the impact of this workaround, but I don't feel particularly strongly about this one. Let's keep it as-is if you think that's better. :)

mitmproxy/websocket.py
mhils: Is this some known standard, or is this something we're making up here?

mitmproxy/addons/export.py
```diff
@@ -130,15 +137,23 @@ def raw(f: flow.Flow, separator=b"\r\n\r\n") -> bytes:
...
         and f.response.raw_content is not None
     )
+    is_websocket_flow = hasattr(f, "websocket") and f.websocket is not None
 
+    serialized_flow = b""
     if request_present and response_present:
-        return b"".join([raw_request(f), separator, raw_response(f)])
+        serialized_flow = serialized_flow.join(
+            [raw_request(f), separator, raw_response(f)]
+        )
```
mhils:
```suggestion
        serialized_flow = raw_request(f) + separator + raw_response(f)
```

and then remove `serialized_flow = b""` above. The existing code here is unidiomatic, let's fix that while we're at it. Thanks!

mhils (approved):
Thanks! I simplified things a bit more, let me know if there's something you believe we should change. I removed `raw_messages` because that's a bit confusing when it appears in the list for plain HTTP. But the export is covered with `raw`.  🍰

PR #6768, opened 2024-04-01
Add hex editor for non text files

wnyyyy:
#### Description

Changed editor spawner so it checks if a file is mostly binary (using strutils), and opens the appropriate editor.

#### Checklist

### DIFF
--- a/mitmproxy/tools/console/master.py
+++ b/mitmproxy/tools/console/master.py
@@ -29,6 +29,7 @@ from mitmproxy.tools.console import keymap
 from mitmproxy.tools.console import palettes
 from mitmproxy.tools.console import signals
 from mitmproxy.tools.console import window
+from mitmproxy.utils import strutils
 
 T = TypeVar("T", str, bytes)
 
@@ -119,13 +120,24 @@ class ConsoleMaster(master.Master):
             return "notepad"
         else:
             return "vi"
+        
+    def get_hex_editor(self) -> str:
+        editors = ["ghex", "bless", "xxd", "imhex", "hexedit", "hexed", "hxd", "emacs"]
+        for editor in editors:
+            if shutil.which(editor):
+                return editor
+        return "hexdump"
 
     def spawn_editor(self, data: T) -> T:
-        text = not isinstance(data, bytes)
+        text = isinstance(data, str)
         fd, name = tempfile.mkstemp("", "mitmproxy", text=text)
+        with_hexeditor = not text and strutils.is_mostly_bin(data)
         with open(fd, "w" if text else "wb") as f:
             f.write(data)
-        c = self.get_editor()
+        if with_hexeditor:
+            c = self.get_hex_editor()            
+        else:
+            c = self.get_editor()
         cmd = shlex.split(c)
         cmd.append(name)
         with self.uistopped():
--- a/mitmproxy/tools/console/master.py
+++ b/mitmproxy/tools/console/master.py
@@ -131,7 +131,7 @@ class ConsoleMaster(master.Master):
     def spawn_editor(self, data: T) -> T:
         text = isinstance(data, str)
         fd, name = tempfile.mkstemp("", "mitmproxy", text=text)
-        with_hexeditor = not text and strutils.is_mostly_bin(data)
+        with_hexeditor = isinstance(data, bytes) and strutils.is_mostly_bin(data)
         with open(fd, "w" if text else "wb") as f:
             f.write(data)
         if with_hexeditor:
--- a/mitmproxy/tools/console/master.py
+++ b/mitmproxy/tools/console/master.py
@@ -122,21 +122,11 @@ class ConsoleMaster(master.Master):
             return "vi"
 
     def get_hex_editor(self) -> str:
-        editors = [
-            "ghex",
-            "bless",
-            "xxd",
-            "imhex",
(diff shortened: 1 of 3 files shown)


wnyyyy:
the data received by the editor spawner function seem to be always in binary format, even if the response headers have content-type text/utf-8, so the check for string might be redundant.

mhils:
> the data received by the editor spawner function seem to be always in binary format, even if the response headers have content-type text/utf-8, so the check for string might be redundant.

I don't think this holds for WebSocket messages, so let's keep it as-is.

mitmproxy/tools/console/master.py
```diff
@@ -120,12 +121,33 @@ def get_editor(self) -> str:
...
+            "hexedit",
+            "hexed",
+            "hxd",
+            "hexdump",
+            "emacs",
+        ]
+        for editor in editors:
+            if shutil.which(editor):
+                return editor
+        return "vi"
```
mhils, line 139: Can we return `self.get_editor()` instead? We don't want people to get stuck in vi. :)

mitmproxy/tools/console/master.py
```diff
@@ -120,12 +121,33 @@ def get_editor(self) -> str:
...
+            "imhex",
+            "hexedit",
+            "hexed",
+            "hxd",
+            "hexdump",
+            "emacs",
+        ]
+        for editor in editors:
+            if shutil.which(editor):
+                return editor
```
mhils, line 138: Stupid question, but have you tested that these editors behave sanely? We need them to 

1. Open the binary when called with `$editor $filename`.
2. Keep the spawned process alive until the editor is closed.

I know that for text editors there are some offenders that spawn a GUI, but then return immediately. That would be a subpar user experience. :)

(Also let's add a comment in case people add other editors in the future)

test/mitmproxy/tools/console/test_master.py
```diff
@@ -0,0 +1,30 @@
...
+    console.spawn_editor(text_data)
+    console.spawn_editor(binary_data)
+
+    console.get_editor.assert_called_once()
+    console.get_hex_editor.assert_called_once()
+
+
+def test_get_hex_editor(monkeypatch, console):
+    test_editor = "emacs"
+    monkeypatch.setattr("shutil.which", lambda x: True if x == test_editor else False)
```
mhils, line 28: nit:
```suggestion
    monkeypatch.setattr("shutil.which", lambda x: x == test_editor)
```

test/mitmproxy/tools/console/test_master.py
```diff
@@ -0,0 +1,30 @@
...
+
+def test_spawn_editor(monkeypatch, console):
+    text_data = "text"
+    binary_data = b"\x00\x01\x02\x03\x04\x05\x06\x07\x08\x09"
+
+    console.get_editor = Mock()
+    console.get_editor.return_value = "editor"
+    console.get_hex_editor = Mock()
+    console.get_hex_editor.return_value = "editor"
+    monkeypatch.setattr("subprocess.call", (lambda _: None))
```
mhils, line 12: I dislike that we need to do so much mocking here, but that's mostly a fault of the existing code. If you feel this could be improved I'd be open for a refactor, but also ok if we want to keep it that way.

test/mitmproxy/tools/console/test_master.py
```diff
@@ -0,0 +1,30 @@
...
+    console.loop = Mock()
+    console.loop.stop = Mock()
+    console.loop.start = Mock()
+    console.loop.draw_screen = Mock()
+
+    console.spawn_editor(text_data)
+    console.spawn_editor(binary_data)
+
+    console.get_editor.assert_called_once()
+    console.get_hex_editor.assert_called_once()
```
mhils, line 23:
```suggestion
    console.spawn_editor(text_data)
    console.get_editor.assert_called_once()

    console.spawn_editor(binary_data)
    console.get_hex_editor.assert_called_once()
```

mhils (commented):
Thanks for the nice PR! 🍰

mitmproxy/tools/console/master.py
wnyyyy: Don't know how I missed that! I'll update the PR

mitmproxy/tools/console/master.py
wnyyyy: I confess I only tested ghex and bless, which were the ones I had installed. Others I included out of my head. I've just finished testing all the editors, except for HxD as I'm on Linux. Emacs happened exactly what you described, so I removed it. hexdump isn't really what I thought it was, and hexed is a browser editor, not a common windows editor as I thought (woops). xxd I included as IIRC was the default in ubuntu, but also doesn't work. I added a couple others I found that worked, including a vi one, but that I couldn't test too well as my brain can't process vim. I'll test some windows ones tomorrow.

test/mitmproxy/tools/console/test_master.py
wnyyyy: I was having trouble trying to understand how the console worked, so I ended up just mocking it, to not leave to PR hanging. I can try to come up with a more appropriate solution for that. I actually wanted to create more tests, including some unrelated to this change, that I thought was missing.

mhils (approved):
Thanks for the PR! 🍰

PR #6799, opened 2024-04-14
Issue2936

### LINKED ISSUE #2936: header names are needlessly line-wrapped

gellge:
#### Description

My attempt at increasing header name line-wrap length according to unused console columns. Specifically:
* line-wrap length is never lower than 30 columns.
* it is increased to use as many unused columns as possible, as long as this does not cause the longest amongst the header values to be wrapped.
* column gaps and indentation whitespace are accounted for when calculating unused columns.

Fixes #2936 .

#### Checklist

### DIFF
--- a/mitmproxy/tools/console/common.py
+++ b/mitmproxy/tools/console/common.py
@@ -67,6 +67,20 @@ def format_keyvals(
     if indent > 2:
         indent -= 2  # We use dividechars=2 below, which already adds two empty spaces
 
+    cols,_ = urwid.raw_display.Screen().get_cols_rows()
+    available_cols = cols - (indent + 2*COL_GAP) # total console columns minus the indent whitespace and the column gaps
+    key_lengths = []
+    val_lengths = []
+    for k,v in entries:
+        if k is not None:
+            key_lengths.append(len(k))
+            if v is not None and isinstance(v, str):
+                val_lengths.append(len(v))
+    max_entry_key_len = max(key_lengths, default=0)
+    max_entry_val_len = max(val_lengths, default=0)
+    unused_cols = available_cols - (max_entry_key_len + max_entry_val_len)
+    max_key_len = max(KEY_MAX, available_cols - max_entry_val_len) if unused_cols < 0 else max_entry_key_len
+
     ret = []
     for k, v in entries:
         if v is None:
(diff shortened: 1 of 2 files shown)


mitmproxy/tools/console/common.py
```diff
@@ -67,6 +68,26 @@ def format_keyvals(
     if indent > 2:
         indent -= 2  # We use dividechars=2 below, which already adds two empty spaces
 
+    cols, _ = urwid.raw_display.Screen().get_cols_rows()
```
mhils, line 71: I haven't tested this, but this looks like it will fall apart when using layouts that don't span the entire width. Can we pass in the available space instead?

mhils (commented):
Thanks for the PR! 😃 🍰

mitmproxy/tools/console/common.py
gellge: Hi @mhils ,
thanks for the review and the awesome work you people do here!

I'm not sure I'm following you: my (admittedly empirical) tests proved the fix to 
work on layouts smaller than full screen (see attached screenshot - header values are not wrapped even if the terminal size is half that of the screen, instead the header name is wrapped at 30 chars). Are there different scenarios you're thinking of?

Furthermore, I investigated a bit more and saw that mitmproxy already uses the same way to retrieve screen columns elsewhere, for instance in [mitmproxy/tools/console/statusbar.py](https://github.com/mitmproxy/mitmproxy/blob/4d00ec03b4fd0a187ed346d438395af3b5e3c3d5/mitmproxy/tools/console/statusbar.py#L81) (note that `self.master.ui` inherits from `urwid.raw_display.Screen` as it can be seen in [mitmproxy/tools/console/window.py](https://github.com/mitmproxy/mitmproxy/blob/4d00ec03b4fd0a187ed346d438395af3b5e3c3d5/mitmproxy/tools/console/window.py#L306).

Anyhow I am by no means an expert on `urwid`, so I'll be glad to refine the PR if you still think this approach is broken.

Screenshot:
mitmproxy_pr

mhils: > Are there different scenarios you're thinking of?

\You can press `-` to toggle between views. If you enable the two column view, `urwid.raw_display.Screen().get_cols_rows()` will give you twice the space available:

image

The statusbar "gets away" with this by always being full width.

After looking into this for a bit, the correct fix is unfortunately more complex. Urwid widgets famously have no widths, so we'd probably need to implement a custom widget here or something like that. :/ I haven't looked super close, but it looks a bit ugly.

PR #6810, opened 2024-04-19
Prefix version tags with 'v'

mhils:
This got lost with mitmproxy 9 somehow. the plan is to recreate those tags with 'v', and then keep using 'v' going forward.


rugk:
Note the changelog is not true. Sever does not require the v prefix nor is it actually valid. It's common and fine, but only the string after v is a valid semantic version.

See https://semver.org/#is-v123-a-semantic-version

mhils:
Thanks. Not sure how I ended up with that, I removed the incorrect wording from the changelog.

rugk:
Maybe because of the branch name. Anyway great my silly nitpick here were incooperated! 😜

PR #6815, opened 2024-04-21
Issue 6739-Added custom DNS server for hostname resolution(Needs fixing)

smenon02:
- #### Description

Creates a DNSServer addon for hostname resolution. Python implementation includes an async getaddrinfo method that opens a UDP socket to send DNS queries to using hostname to resolve. Running into issues while testing as communication between UDP server and DNS client is timing out. Any assistance of this would be greatly appreciated :)
UDP server was created to test hostname resolution. DNS client is not properly talking to it, so once communication between the two is fixed, test cases can be generated. 
#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@
 
 * Release tags are now prefixed with `v` again to follow SemVer convention.
   ([#6810](https://github.com/mitmproxy/mitmproxy/pull/6810), @mhils)
-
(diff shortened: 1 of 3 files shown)


mhils (changes requested):
Thanks for the PR. The `async_getaddrinfo` method is neither async nor does it implement a getaddrinfo-equivalent. So this should be fixed before we get to any tests.

PR #6821, opened 2024-04-25
Update PyInstaller spec to set the u (unbuffered) flag

### LINKED ISSUE #6757: mitmdump stdout never reaches parent process
#### Problem Description

I'm spawning mitmdump from Node.js as a child process. Starting with mitmdump 8 stdout never reaches the parent. Only when I kill the mitmdump process will I finally get its output.

index.mjs

```js
import { spawn } from 'child_process';

const mitm = spawn(
  '/home/alex/Downloads/mitmproxy-10.2.4-linux-x86_64/mitmdump',
  // '/home/alex/Downloads/mitmproxy-9.0.1-linux/mitmdump',
  // '/home/alex/Downloads/mitmproxy-8.1.1-linux/mitmdump',
  // '/home/alex/Downloads/mitmproxy-7.0.4-linux/mitmdump',
  // '/home/alex/Downloads/mitmproxy-6.0.2-linux/mitmdump',
  // '/home/alex/forks/mitmproxy/venv/bin/mitmdump',
  {
    env: {
      // This has an effect on the venv but not the PyInstaller binary > 7
      PYTHONUNBUFFERED: '1',
    },
  },
);

mitm.stdout.on('data', (data) => {
  console.group(`mitm stdout`);
  console.log(`${data}`);
  console.groupEnd();
});

mitm.stderr.on('data', (data) => {
  console.group(`mitm stderr`);
  console.log(`${data}`);
  console.groupEnd();
});

mitm.on('exit', (code, signal) => {
  console.log(`mitm process exited with code ${code} from signal ${signal}`);
});
```

1. `node index.mjs`
2. _nothingness_
3. `killall mitmdump`
4. Now Node.js gets the output

Version 6 + 7: I get stdout
Version 8 + 9 + 10: I don't get stdout, only after I kill the mitmdump process

I assume it must be PyInstaller, because spawning directly from `venv/bin` works too (but _only_ with `PYTHONUNBUFFERED=1`). However, if I add the following addon I get all stdout:

```py
import sys


def running():
    print("running")
    sys.stdout.flush()
```

So I assume PyInstaller does not implement/respect `PYTHONUNBUFFERED` and mitmproxy does not explicitly flush.

Looking at the PyInstaller issue tracker I assume I must be the first human to spawn a PyInstaller process? Or maybe it's how we use PyInstaller and we can actually configure this on our end?

Prinzhorn:
#### Description

Fixes #6757

#### Checklist

 - [x] Only set this for mitmdump because I'm paranoid

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Fix a bug where mitmproxy would not exit when `-n` is passed.
   ([#6819](https://github.com/mitmproxy/mitmproxy/pull/6819), @mhils)
 * Set the `unbuffered` (stdout/stderr) flag for the `mitmdump` PyInstaller build.
+  ([#6821](https://github.com/mitmproxy/mitmproxy/pull/6821), @Prinzhorn)
 
 
 ## 17 April 2024: mitmproxy 10.3.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   ([#6810](https://github.com/mitmproxy/mitmproxy/pull/6810), @mhils)
 * Fix a bug where mitmproxy would not exit when `-n` is passed.
   ([#6819](https://github.com/mitmproxy/mitmproxy/pull/6819), @mhils)
+* Set the `unbuffered` (stdout/stderr) flag for the `mitmdump` PyInstaller build.
 
 
 ## 17 April 2024: mitmproxy 10.3.0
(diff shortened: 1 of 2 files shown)


Prinzhorn:
Can confirm, I downloaded the artifact and now I immediately get stdout.

Prinzhorn:
Will test the next artifact and then we're good if there are no objections

mhils (approved):

PR #6823, opened 2024-04-29
Temporarily prohibit pytest 8.2.0 in order to fix CI

alex:
See https://github.com/pytest-dev/pytest/issues/12263


mhils (approved):
Thanks! 🍰

Issue #6660, opened 2024-02-14
Add support to Sequence[int] similar to Sequence[str] for addon commands


m417z:
#### Problem Description

Addon commands can be sequences, but the only supported sequence is of type `str`. Add support for `Sequence[int]`.
https://github.com/mitmproxy/mitmproxy/blob/1f2e1a254a68e54e60360b66762ed6bdf2ff26be/docs/src/content/addons-commands.md?plain=1#L120
https://github.com/mitmproxy/mitmproxy/blob/1f2e1a254a68e54e60360b66762ed6bdf2ff26be/mitmproxy/optmanager.py#L368

#### Additional context
Example use case: passing multiple status codes.
https://github.com/realgam3/pymultitor/pull/38

mhils:
Happy to merge a PR for this! 😃

PR #6841, opened 2024-05-10
Add support to Sequence[int] for addon commands

Peacexoom:
#### Description

Addon commands can be sequences both Sequence[str] and Sequence[int] (#6660). 

#### Checklist


niderafeha64756:
1111

PR #6853, opened 2024-05-17
Update change_upstream_proxy.py

fosple:
Script was not working as ServerSpec accepts a tuple, which was not supplied. I fixed it so a tuple is supplied to ServerSpec.


mhils (approved):
Thanks! 🍰

PR #6871, opened 2024-05-23
add Copy as Python Requests

Nriver:
#### Description

ksnip_20240522-175559

#### Checklist

### DIFF
--- a/mitmproxy/addons/export.py
+++ b/mitmproxy/addons/export.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import shlex
 from collections.abc import Callable
@@ -106,6 +107,54 @@ def httpie_command(f: flow.Flow) -> str:
     return cmd
 
 
+def python_requests_command(f: flow.Flow) -> str:
+    # Copy as Python Requests
+    request = cleanup_request(f)
+    request = pop_headers(request)
+
+    def escape_quotes(s: str) -> str:
+        return s.replace('"', '\\"')
+
+    code = 'import requests\n\n'
+
+    code += f'url = "{request.pretty_url}"\n'
+    code += 'headers = {\n'
+    cookies = ''
+    for k, v in request.headers.items(multi=True):
+        if k.lower() == 'cookie':
+            cookies += f'{v};'
+        else:
+            code += f'    "{k}": "{escape_quotes(v)}",\n'
+    code += '}\n'
+
+    if cookies:
+        code += 'cookies = {\n'
+        for cookie in cookies.strip().split('; '):
+            key, value = cookie.split('=', 1)
+            code += f'    "{key}": "{escape_quotes(value)}",\n'
+        code += '}\n'
+    else:
+        code += 'cookies = {}\n\n'
+
+    if request.content:
+        # Try to decode request content as JSON if possible
+        try:
+            body = json.loads(request.content.decode('utf-8'))
+            body_str = json.dumps(body, indent=4)
+            code += f'body = """{escape_quotes(body_str)}"""\n'
+        except (json.JSONDecodeError, UnicodeDecodeError):
+            # Fall back to plain string representation
+            code += f'body = "{escape_quotes(request.content.decode("utf-8", "ignore"))}"\n'
+    else:
+        code += 'body = None\n'
+
+    code += (f'res = requests.request(method="{escape_quotes(request.method)}", '
+             f'url=url, headers=headers, cookies=cookies, data=body)\n')
+    code += 'print(res.text)\n'
+
+    return code
+
+
 def raw_request(f: flow.Flow) -> bytes:
     request = cleanup_request(f)
     if request.raw_content is None:
@@ -147,6 +196,7 @@ def raw(f: flow.Flow, separator=b"\r\n\r\n") -> bytes:
 formats: dict[str, Callable[[flow.Flow], str | bytes]] = dict(
     curl=curl_command,
     httpie=httpie_command,
+    python_requests=python_requests_command,
     raw=raw,
     raw_request=raw_request,
     raw_response=raw_response,
--- a/web/src/js/components/Header/FlowMenu.tsx
+++ b/web/src/js/components/Header/FlowMenu.tsx
@@ -206,6 +206,9 @@ function ExportButton({ flow }: { flow: Flow }) {
             <MenuItem onClick={() => copy(flow, "httpie")}>
                 Copy as HTTPie
             </MenuItem>
+            <MenuItem onClick={() => copy(flow, "python_requests")}>
+                Copy as Python Requests
+            </MenuItem>
         </Dropdown>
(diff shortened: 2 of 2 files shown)


mitmproxy/addons/export.py
```diff
@@ -106,6 +107,69 @@ def httpie_command(f: flow.Flow) -> str:
...
+    request = pop_headers(request)
+
+    def escape_quotes(s: str) -> str:
+        return s.replace('"', '\\"')
+
+    code = "import requests\n\n"
+
+    code += f'url = "{request.pretty_url}"\n'
+    code += "headers = {\n"
+    cookies = ""
```
mhils, line 122: Can you please use https://docs.mitmproxy.org/stable/api/mitmproxy/http.html#Request.cookies to construct the cookies attribute?

mitmproxy/addons/export.py
```diff
@@ -106,6 +107,69 @@ def httpie_command(f: flow.Flow) -> str:
...
+            key, value = cookie.split("=", 1)
+            code += f'    "{key}": "{escape_quotes(value)}",\n'
+        code += "}\n"
+    else:
+        code += "cookies = {}\n\n"
+
+    is_json = False
+    if request.content:
+        try:
+            decoded_content = request.content.decode("utf-8")
```
mhils, line 142: Let's streamline this:

1. If `json.loads(request.content)` works, go with that. json.loads accepts raw bytes, so we don't have to decode for that.
2. If https://docs.mitmproxy.org/stable/api/mitmproxy/http.html#Message.text does not raise, assign body_str to repr(text)
3. As a fallback, assign body_str to repr(request.content).

Technically steps 1-2 are optional, but definitely nice to have.

mitmproxy/addons/export.py
```diff
@@ -106,6 +107,69 @@ def httpie_command(f: flow.Flow) -> str:
...
+    code = "import requests\n\n"
+
+    code += f'url = "{request.pretty_url}"\n'
+    code += "headers = {\n"
+    cookies = ""
+    for k, v in request.headers.items(multi=True):
+        if k.lower() == "cookie":
+            cookies += f"{v};"
+        else:
+            code += f'    "{k}": "{escape_quotes(v)}",\n'
```
mhils, line 127: This will generate invalid code for repeating headers - I'd say it would be ok to go with multi=False here and then manually remove the cookie header. Maybe something like this works with some tweaking? (untested)
```python
request.headers.pop("cookie", None)
header_string = pprint.pformat(dict(request.headers.items()))
```

mhils (commented):
Thanks for the PR!

mitmproxy/addons/export.py
Nriver: Not sure how to do step 2.

mhils: I was thinking of something like this:

```python
from pprint import pformat

try:
    body_val = json.loads(request.content)
except ValueError:
    try:
        body_val = request.text
    except ValueError:
        body_val = request.content

body_str = f"body = {pformat(body_val, indent=4, sort_dicts=False)}"
```

mitmproxy/addons/export.py
```diff
@@ -106,6 +107,65 @@ def httpie_command(f: flow.Flow) -> str:
...
+
+    code = "import requests\n\n"
+
+    code += f'url = "{request.pretty_url}"\n'
+
+    # parse cookie
+    code += "cookies = {\n"
+    for k, v in request.cookies.items():
+        code += f'    "{k}": "{escape_quotes(v)}",\n'
+    code += "}\n"
```
mhils, line 126: Can we use pformat here instead of manually constructing stuff?

mitmproxy/addons/export.py
```diff
@@ -106,6 +107,65 @@ def httpie_command(f: flow.Flow) -> str:
...
+
+    # parse header
+    request.headers.pop("cookie", None)
+    code += "headers = {\n"
+    for k, v in request.headers.items():
+        code += f'    "{k}": "{escape_quotes(v)}",\n'
+    code += "}\n"
```
mhils, line 133: Can we use pformat here instead of manually constructing stuff?

mitmproxy/addons/export.py
```diff
@@ -106,6 +107,65 @@ def httpie_command(f: flow.Flow) -> str:
...
+                # Fall back to plain string representation
+                body_str = f'"{escape_quotes(repr(request.content)[2:-1])}"'
+
+        code += f"body = {body_str}\n"
+    else:
+        code += "body = None\n"
+
+    param = "json=body" if is_json else "data=body"
+    code += (
+        f'res = requests.request(method="{escape_quotes(request.method)}", '
```
mhils, line 161: Let's use `repr()` here, then we don't need any custom escaping functions

mitmproxy/addons/export.py
Nriver: Similar to the above, this will create
```
headers = Headers[(b'header', b'qvalue')]
```

mitmproxy/addons/export.py
Nriver: And this, will create weird code format like this
```
body = {   'string': 'Hello, world!',
    'number': 42,
    'float': 3.14,
    'boolean': True,
    'nullValue': None,
    'object': {'name': 'John', 'age': 30},
    'array': [1, 2, 3, 4]}

```

mitmproxy/addons/export.py
Nriver: And this, will create weird code format like this
```
body = {   'string': 'Hello, world!',
    'number': 42,
    'float': 3.14,
    'boolean': True,
    'nullValue': None,
    'object': {'name': 'John', 'age': 30},
    'array': [1, 2, 3, 4]}

```

sujaldev:
@Nriver are you still working on this?

Nriver:
Nope. I'm done here.

r3sting:
This seems to work. What needs to be done to merge this in?

sujaldev:
> This seems to work. What needs to be done to merge this in?

One issue I ran into trying to get this PR merged in was that the requests library provides no way to override dns resolution like curl's `--resolve` flag. I got away with it by patching urllib3 (which is what requests uses under the hood) in the generated output if the `export_preserve_original_ip` option is set to True, but patching urllib3 internals is not the nicest solution here.

r3sting:
> > This seems to work. What needs to be done to merge this in?
> 
> One issue I ran into trying to get this PR merged in was that the requests library provides no way to override dns resolution like curl's `--resolve` flag. I got away with it by patching urllib3 (which is what requests uses under the hood) in the generated output if the `export_preserve_original_ip` option is set to True, but patching urllib3 internals is not the nicest solution here.

I believe that's fine. If the user wants to add advance networking logic *after* they've generated the Python request code, then they can.

Issue #6873, opened 2024-05-24
Mitmproxy unusably slow with large responses in reverse mode


aib:
Thanks for everyone's work on this amazing piece of software!

#### Problem Description
Returning a large (tens of megabytes) response in reverse mode is _very_ slow, getting ~exponentially worse with the body size.

#### Steps to reproduce the behavior:
1. `git clone https://github.com/mitmproxy/mitmproxy.git && cd mitmproxy`
2. `python3 -m venv venv && source venv/bin/activate`
3. `pip install . && pip uninstall -y mitmproxy`
4.
```
cat <<EOF >/tmp/big.py

import mitmproxy

def request(flow):
    flow.response = mitmproxy.http.Response.make(200, "A" * (50 * 1048576))

EOF
```
5. `python -c 'from mitmproxy.tools.main import mitmproxy; mitmproxy(["--mode", "reverse:https://duckduckgo.com", "-s", "/tmp/big.py"])'`
7. `curl https://localhost:8080/big -o /dev/null`

#### System Information
Mitmproxy: 11.0.0.dev (+17, commit 552c320)
Python:    3.10.12
OpenSSL:   OpenSSL 3.2.1 30 Jan 2024
Platform:  Linux-5.15.0-107-generic-x86_64-with-glibc2.35
HEAD is 552c320fadb13a5068f45f34a4f0b30e7430a3a0

Possibly related to https://github.com/mitmproxy/mitmproxy/issues/6301 ?
I will investigate further and try to find the cause.

aib:
Ah! HTTPS/1.0 and 1.1 (as with curl's `--http1.0` and `--http1.1` parameters) respond instantly as well.
So does wget (HTTPS/1.1).

aib:
Profiler (cProfile) output:

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   1420/1    0.036    0.000   34.349   34.349 {built-in method builtins.exec}
        1    0.000    0.000   34.349   34.349 run.py:1(<module>)
        1    0.000    0.000   33.810   33.810 main.py:134(mitmproxy)
        1    0.000    0.000   33.718   33.718 main.py:45(run)
        1    0.000    0.000   33.718   33.718 runners.py:8(run)
        3    0.000    0.000   33.717   11.239 base_events.py:613(run_until_complete)
        3    0.000    0.000   33.717   11.239 base_events.py:589(run_forever)
[...]

PR #6875, opened 2024-05-25
Fix slowdown when sending large data over HTTP/2 (#6873)

aib:
#### Description

Do not mutate data when splitting it into outbound_frame_size-sized chunks, as discussed in #6873.

#### Checklist


mhils (approved):
Awesome, thank you so much! 🍰 😃 🚀

PR #6876, opened 2024-05-26
Add blockech addon

errorxyz:
#### Description

This PR adds a `blockech` addon that prevents encrypted SNI by removing HTTPS records from answers in an effort to prevent the `ech` key from being exchanged. The next step to this PR would be to parse HTTPS records and remove the `ech` key alone from those records.

#### Checklist

### DIFF
--- a/mitmproxy/addons/__init__.py
+++ b/mitmproxy/addons/__init__.py
@@ -1,6 +1,7 @@
 from mitmproxy.addons import anticache
 from mitmproxy.addons import anticomp
 from mitmproxy.addons import block
+from mitmproxy.addons import blockech
 from mitmproxy.addons import blocklist
 from mitmproxy.addons import browser
 from mitmproxy.addons import clientplayback
@@ -34,6 +35,7 @@ def default_addons():
         core.Core(),
         browser.Browser(),
         block.Block(),
+        blockech.BlockECH(),
         blocklist.BlockList(),
         anticache.AntiCache(),
         anticomp.AntiComp(),
--- a/mitmproxy/addons/__init__.py
+++ b/mitmproxy/addons/__init__.py
@@ -1,7 +1,7 @@
 from mitmproxy.addons import anticache
 from mitmproxy.addons import anticomp
 from mitmproxy.addons import block
-from mitmproxy.addons import blockech
+from mitmproxy.addons import block_ech
 from mitmproxy.addons import blocklist
 from mitmproxy.addons import browser
 from mitmproxy.addons import clientplayback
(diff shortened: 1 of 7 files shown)


mitmproxy/addons/__init__.py
```diff
@@ -1,6 +1,7 @@
 from mitmproxy.addons import anticache
 from mitmproxy.addons import anticomp
 from mitmproxy.addons import block
+from mitmproxy.addons import blockech
```
mhils, line 4: `blockech` does not read super well - maybe `block_ech` for the file and option name?

mitmproxy/addons/blockech.py
```diff
@@ -0,0 +1,21 @@
...
+            "Prevent encrypted SNI by blocking ECH key from passing through by removing HTTPS records from answers",
+        )
+
+    def dns_response(self, flow: dns.DNSFlow):
+        # TODO: parse HTTPS records and remove ech value alone. For now,
+        # if HTTPS record is part of response, remove that record.
+        if ctx.options.blockech and flow.response is not None:
+            for answer in flow.response.answers:
+                if answer.type == types.HTTPS:
+                    flow.response.answers.remove(answer)
```
mhils, line 21: Let's turn this into a list comprehension? I think the current code may crash if you try to modify the list while you iterate over it.

mitmproxy/addons/blockech.py
```diff
@@ -0,0 +1,21 @@
...
+from mitmproxy.net.dns import types
+
+
+class BlockECH:
+    def load(self, loader):
+        loader.add_option(
+            "blockech",
+            bool,
+            True,
+            "Prevent encrypted SNI by blocking ECH key from passing through by removing HTTPS records from answers",
```
mhils, line 12: Let's spell out acronyms here, not everyone will be familiar with these. How about

> Strip DNS HTTPS records to prevent clients from encrypting TLS ClientHello messages.

? Open to suggestions. :)

mitmproxy/addons/blockech.py
```diff
@@ -0,0 +1,21 @@
...
+        )
+
+    def dns_response(self, flow: dns.DNSFlow):
+        # TODO: parse HTTPS records and remove ech value alone. For now,
+        # if HTTPS record is part of response, remove that record.
+        if ctx.options.blockech and flow.response is not None:
```
mhils, line 18: Is `flow.response` ever None in `dns_response`?

mhils (commented):
Nice first step, some comments below! 😃

PR #6884, opened 2024-05-31
Allow parsing of HTTPS record from DNS RDATA

errorxyz:
#### Description

This PR allows us to unpack(parse) raw HTTPS record data from DNS RDATA and convert it into HTTPSRecord format which stores SVCParamValue as raw bytes corresponding to the SVCParamKey.
Exposes a `ResourceRecord.https_ech` API to access parsed ECH key alone from the SVCParamKeys 
It also allows us to pack this HTTPSRecord format back into raw bytes.
HTTPS record data format: [link](https://datatracker.ietf.org/doc/rfc9460/) (Section 2.2)

#### Checklist

#### TODO
- ~Test parameter types~
- ~Do we need a new class to represent HTTPS record data since there are too many parameters?~
- ~Finalize the print format for HTTPS record data~
- ~Error handling~
- ~Add tests~
- ~Add comments~

Note: You'll have to set `block_ech` to `false` to test this if your mitmproxy is updated till commit `de871df`


mitmproxy/net/dns/https_record.py
```diff
@@ -0,0 +1,91 @@
...
+            offset += 1
+            break
+        offset += 1
+        labels.append(data[offset : offset + length].decode("utf-8"))
+        offset += length
+    return ".".join(labels), offset
+
+
+def unpack(data: bytes) -> dict:
+    """Unpacks HTTPS RDATA from byte data."""
```
mhils, line 73: What should happen for malformed records? Do we raise an error, or is this method supposed to never fail?

I assume it's the former, but then we should document it in the docstring. :)

errorxyz:
Currently, the HTTPS record is printed as a dictionary as follows:
```
{'priority': 1, 'target_name': '', 'params': {'alpn': ['h2'], 'ipv4hint': ['162.159.135.79', '162.159.136.79'], 'echconfig': 'AEX+DQBBdwAgACCoDsczydyWEGIp/i/+ErPbmJkRt5VA6Lr0+W6arT4FHgAEAAEAAQASY2xvdWRmbGFyZS1lY2guY29tAAA=', 'ipv6hint': ['2606:4700:7::a29f:874f', '2606:4700:7::a29f:884f']}}
```
Other formats:
`dig`'s format:
 ```
1 . alpn="h2" ipv4hint=162.159.135.79,162.159.136.79 ech=AEX+DQBBdwAgACCoDsczydyWEGIp/i/+ErPbmJkRt5VA6Lr0+W6arT4FHgAEAAEAAQASY2xvdWRmbGFyZS1lY2guY29tAAA= ipv6hint=2606:4700:7::a29f:874f,2606:4700:7::a29f:884f
```
modified `dig`'s format:
```
priority=1 target_name="" alpn="h2" ipv4hint=162.159.135.79,162.159.136.79 ech=...
```

Any better format that we can think of?

**Review by mhils — COMMENTED** [review]
> Do we need a new class to represent HTTPS record data since there are too many parameters?

Up to you - I think it's on the edge and we can go with whatever you prefer.

> Finalize the print format for HTTPS record data

👍  We don't need to be overly fancy here, let's just go with what we need ourselves.

> Error handling

See comment

> Add tests

pytest.mark.parametrize is your friend. We may also want to use hypothesis for some fuzz testing to make sure that our parser is only ever returning data or raising a ValueError, never something else. https://github.com/mitmproxy/mitmproxy/blob/de871df53bfbec1c52c5a1b05bdbecee2efa1098/test/mitmproxy/io/test_io.py#L18-L28 is an example for this.

File: mitmproxy/net/dns/https_record.py [hunk 2]
```diff
@@ -0,0 +1,91 @@
+import base64
+import ipaddress
+import struct
+
+
+def unpack_params(data: bytes, offset: int) -> dict:
```
mhils, line 6: Nit: You can preface this method with `_` to mark it as private. We certainly don't do that consistently across the codebase, but we should do it for new code :)

mitmproxy/net/dns/https_record.py
errorxyz: Error handling is still a `TODO` as I've mentioned, I thought we can first make the API concrete then handle that part

mitmproxy/net/dns/https_record.py
```diff
@@ -0,0 +1,91 @@
...
+        offset += param_length
+
+        # Interpret parameters based on its type
+        if param_type == 0:  # Mandatory
+            mandatory_types = [
+                struct.unpack("!H", param_value[i : i + 2])[0]
+                for i in range(0, param_length, 2)
+            ]
+            params["mandatory"] = mandatory_types
+        elif param_type == 1:  # ALPN
```
mhils, line 24: Let's move the hardcoded constants either into numbers or into an enum?

mhils (commented):
This looks really nice now, the only thing I'm not a fan of is the implementation of `SVCParams`. More specifically, looking at the RFC, I think we shouldn't error for unknown keys, but instead keep a generic mapping of all keys. What do you think of the following:

1. Add an enum.Enum with the constants (IPV4HINT = 6, ...)
2. Make params simply `dict[int,bytes]`
3. Make HTTPSRecord.__repr__ use the enum to lookup key names.
4. Add `@property`s for the params that we need right now on HTTPSRecord to set/modify/delete. (I think right now it is just ECH?)
5. Ignore the rest for now. 🙃

errorxyz:
can we use the existing implementation to represent the constants instead of `enum`?
`enum`'s make the code look very quirky
```
...
    params[SVCParamKeys.ALPN.value] = alpn_protocols
elif param_type == SVCParamKeys.NO_DEFAULT_ALPN.value:
    params[SVCParamKeys.NO_DEFAULT_ALPN.value] = True
elif param_type == SVCParamKeys.PORT.value:
    port = struct.unpack("!H", param_value)[0]
    params[SVCParamKeys.PORT.value] = port
elif param_type == SVCParamKeys.IPV4HINT.value:
    params[SVCParamKeys.IPV4HINT.value] = [
...
```
Also, there's no API in `enum` to check if some value is in it (need it to check if we have a string representing the value or use the `keyNNNN` format) although there are workarounds to this

mhils:
> Also, there's no API in enum to check if some value is in it

You can do the following:

```python
try:
    name = Params(param_type).name
except ValueError:
    name = f"key{param_type}"
```

Not super elegant, but ok-ish.

> By this, do you mean that I shouldn't parse param_values while unpacking and store them as bytes?

Yes - just have the raw values in .params, and have a `@property` on HTTPSRecord that returns the parsed representation where we need it. Otherwise we are introducing a whole bunch of parsing code that is not really used for anything. :)

> can we use the existing implementation to represent the constants instead of `enum`? `enum`'s make the code look very quirky

Either is fine with me, no super strong opinions. :) Most of the enum usage goes away once `params` is a raw dict as mentioned above. For example, the code block you posted as an example will just be

```python
params[param_type] = param_value
```

But really both approaches work, go with what you prefer. :)

errorxyz:
> Yes - just have the raw values in .params, and have a `@property` on HTTPSRecord that returns the parsed representation where we need it. Otherwise we are introducing a whole bunch of parsing code that is not really used for anything. :)
> 
> >
If we are to do this, then `__repr__` would return the raw bytes as values while printing:
```
priority: 1 target_name: '' {'alpn': b'\x02h2', 'ipv4hint': b'\xa2\x9f\x87O\xa2\x9f\x88O', 'ech':
b"\x00E\xfe\r\x00AK\x00 \x00 nu,\x00<\xc4\xe4%\xdb\xb3\xa1\xbf\xafh\xdd\xd8{\xd7\x80\x15:\xa5\xe6\xef\x05U\xc0\x8d\x10'Ry\x00\x04\x00\x01\x00\x01\x00\
x12cloudflare-ech.com\x00\x00", 'ipv6hint':
b'&\x06G\x00\x00\x07\x00\x00\x00\x00\x00\x00\xa2\x9f\x87O&\x06G\x00\x00\x07\x00\x00\x00\x00\x00\x00\xa2\x9f\x88O'}
```
Is this what we want?

errorxyz:
Another API-related problem I encountered while trying to change `ech` value:
```
# changes ech value i.e reflected in ResourceRecord.data
record = answer.https_record
record.ech = None
answer.https_record = record

# changes not reflected in ResourceRecord.data
answer.https_record.ech = None
```
Or is this expected behaviour? Or maybe we can expose `ResourceRecord.ech` directly although the logical structure won't be ideal

errorxyz:
One solution I can think of is to make `ResourceRecord.data: RData` instead of bytes and `RData` class should implement `pack` and `unpack`. Each Record type(HTTPS, TXT, A, etc) should be subclasses to `RData` and implement their respective `pack` and `unpack` functions. Finally we can have `ResourceRecord.pack` pack the record back into bytes whenever needed. TLDR; store record data as python objects and have them converted to bytes whenever needed instead of storing as bytes and converting them to a more usable format whenever needed.

mhils:
> Is this what we want?

Not really, but it's probably good enough for a first version. I guess we could improve `__repr__` to use the properties for known keys, but that's a bit of side quest we can tackle whenever we are happy with the rest. :)

> Another API-related problem I encountered while trying to change `ech` value:
> 
> ```
> # changes ech value i.e reflected in ResourceRecord.data
> record = answer.https_record
> record.ech = None
> answer.https_record = record
> 
> # changes not reflected in ResourceRecord.data
> answer.https_record.ech = None
> ```
> 
> Or is this expected behaviour? Or maybe we can expose `ResourceRecord.ech` directly although the logical structure won't be ideal

Eh - annoying, but really good we caught this early on! 👏 
For context, we have had a similar problem with HTTP Request.cookies / Message.headers / .... Over there we have solved it with [MultiDictView](https://docs.mitmproxy.org/stable/api/mitmproxy/coretypes/multidict.html#MultiDictView) as a proxy object, see for instance [here](https://github.com/mitmproxy/mitmproxy/blob/bd39c7b96fd74f0ff83d69acf93898dbcdce9310/mitmproxy/http.py#L923-L956). Modifying the MultiDictView transparently updates the underlying datastructure. So I think we have four options:

1. Make `ResourceRecord.https_record` immutable. Users who want to modify would need to use `dataclasses.replace`, which makes it explicit that https_record needs to be reassigned. This works, but it's a rather annoying API.
2. Flatten the hierarchy and expose an `ech` (or `https_ech`) attribute on `ResourceRecord`. (I think this is what you suggested in https://github.com/mitmproxy/mitmproxy/pull/6884#issuecomment-2149581996)
3. Implement subclasses (I think this is what you suggested in https://github.com/mitmproxy/mitmproxy/pull/6884#issuecomment-2149691320). This can be either RData subclasses, or maybe even ResourceRecord subclasses. One downside with this approach is that serialization/deserialization becomes more tricky. Right now our serialization mechanism does not support type discriminators.
4. Implement a proxy object similar (in spirit) to MultiDictView.

What do you think? I think I'm leaning towards 2) or 4), but no super strong preferences on either.

/cc @Prinzhorn as someone with an API user perspective. :)

errorxyz:
Personally, I feel 4) might be a bit complicated and introducing this complexity for just this use case might be a bit overkill. Also, I think we can introduce this once this part of the DNS API matures a bit more and has a more concrete structure. My opinion is based on the limited knowledge that I have of the codebase, might be in wrong direction here too🤔 So that leaves us with 2) as a (hopefully temporary) solution

mhils:
Going with 2) for now sounds super reasonable to me 👍

mhils (commented):
This PR looks awesome now! The Record parsing got really nice and simple. 🍰 

Should we maybe remove the `https_record` property from the public API? This is still a bit misguiding because of the assignment-does-not-update-value issue you pointed out, and we can simply use unpack/pack in the `https_ech` property directly.

mhils (approved):
🚀🚀🚀

PR #6889, opened 2024-06-01
Change Windows Installer to delete old APR entries

### LINKED ISSUE #6888: Windows MSI installer does not clear old ARP entries on version upgrade
#### Problem Description
Upon upgrading to a newer version of mitmproxy with the MSI installer while an older one is already installed, the installer does not clear the old Windows ARP entries. This leads to mitmproxy being listed multiple twice under the installed apps.  

Additionally the winget package manager does not recognize that the newest version is already installed and therefore always displays an available update.

1. Download and install old mitmproxy msi installer (e.g. [mitmproxy-10.2.1](https://downloads.mitmproxy.org/10.2.1/mitmproxy-10.2.1-windows-x86_64-installer.exe)
2. Download and newer mitmproxy msi installer (e.g. [mitmproxy-10.3.0](https://downloads.mitmproxy.org/10.3.0/mitmproxy-10.3.0-windows-x86_64-installer.exe)
3. Take a look at Windows "Installed Apps"
grafik

#### System Information
Tested on Windows 10 and 11

a-mnich:
### Issue (#6888)
Currently upon upgrading to a newer version of mitmproxy with the MSI installer while an older mitmproxy version is already installed, the installer does not clear the old Windows ARP entries.

### PR Description
I've adapted the InstallBuilder XML configuration to check for registry entries within `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\` that match the `mitmproxy {Version}` naming scheme and delete the old versions when the installer detects that it is in the upgrade mode based on the installdir already existing.

Alternatively it would be possible to modify the configuration to only delete the entries for the currently installed version based on the `HKEY_LOCAL_MACHINE\Software\mitmproxy.org\mitmproxy` entry. But then duplicate entries that already exist as of today would never be deleted as this registry entry only holds the currently installed version and not past ones. Therefore I've decided for the regex approach within this PR.

### Important:
As a license key is required to run the InstallBuilder, I was not able to actually test this and only changed the configuration based on the documentation. 

#### InstallBuilder documentation:
- [Updates](https://releases.installbuilder.com/installbuilder/docs/installbuilder-userguide/_updates.html)
- [Actions](https://releases.installbuilder.com/installbuilder/docs/installbuilder-userguide/_actions.html)

resolves #6888

grafik

#### Checklist

 - [ ] ~~I have updated tests where applicable.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
   ([#6875](https://github.com/mitmproxy/mitmproxy/pull/6875), @aib)
 * Add an option to strip HTTPS records from DNS responses to block encrypted ClientHellos.
   ([#6876](https://github.com/mitmproxy/mitmproxy/pull/6876), @errorxyz)
+* Update InstallBuilder for Windows to delete old ARP entries on upgrade.
 
 
 ## 17 April 2024: mitmproxy 10.3.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * Add an option to strip HTTPS records from DNS responses to block encrypted ClientHellos.
   ([#6876](https://github.com/mitmproxy/mitmproxy/pull/6876), @errorxyz)
 * Update InstallBuilder for Windows to delete old ARP entries on upgrade.
+  ([#6889](https://github.com/mitmproxy/mitmproxy/pull/6889), @a-mnich)
 
 
 ## 17 April 2024: mitmproxy 10.3.0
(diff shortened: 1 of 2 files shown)


mhils:
Thanks for digging into this! 🍰 

To clarify, with this PR you couldn't really install multiple mitmproxy versions in parallel (to different directories)? The latest install would remove all previous versions from the software list?

mhils:
Alternatively, could we do this?

https://community.broadcom.com/vmware-cloud-foundation/discussion/uninstall-older-version-before-latest-installation#bmf191f1b0-1e9e-427f-8ae7-aa138a373eab

mhils:
Either option - (including doing both) is fine with me here. 😃 

Running the uninstaller silently should not pose any problem. Removing installs with the same InstallLocation is fine as well. 

It should be possible to test this with the evaluation version of InstallBuilder. If that doesn't work, let me know and I can do some testing or share the install key. :)

PR #6905, opened 2024-06-09
Update strip_ech addon to use the new DNS HTTPS records API

errorxyz:
#### Description

Updates the `strip_ech` addon to use the new DNS HTTPS records API that allows us to parse HTTPS records and remove `ech` value alone from the data instead of blocking HTTPS records entirely

#### Checklist


mhils (approved):
👍

PR #6906, opened 2024-06-09
Add `HttpConnectErrorHook` and `HttpConnectedHook`

### LINKED ISSUE #6632: CONNECT request triggers `response` hook
https://github.com/mitmproxy/mitmproxy/pull/6599#issuecomment-1912689476 has a repro. This line here makes us trigger the `response` hook for failed flows, which we don't want to do:

https://github.com/mitmproxy/mitmproxy/blob/c49c38e4c80fd4ad491986a2a7f8a18e75e16b2f/mitmproxy/proxy/layers/http/__init__.py#L760

The correct fix here is probably not to call `send_response` here but mirror what we do for the happy path. And then this needs a test.

errorxyz:
#### Description

Fixes #6632 

#### Checklist


mitmproxy/proxy/layers/http/__init__.py
```diff
@@ -756,6 +756,15 @@ def handle_connect_finish(self):
                 self.context.client,
             )
             yield SendHttp(ResponseEndOfMessage(self.stream_id), self.context.client)
+        elif self.flow.response.status_code == 502:
```
mhils, line 759: I think we should do this path for all non-2xx responses?

mitmproxy/proxy/layers/http/__init__.py
```diff
@@ -756,6 +756,15 @@ def handle_connect_finish(self):
...
+            self.flow.error = flow.Error("CONNECT failed")
+            yield HttpErrorHook(self.flow)
```
mhils, line 761: I realize the original issue wasn't that clear here: I think we neither want to call `response` nor `error` here - these are reserved for regular flows. We don't want flows to magically appear here without previously going through `requestheaders` first. Instead, we should add an `http_connect_error` hook (see comment in the test below).

test/mitmproxy/proxy/layers/http/test_http.py
```diff
@@ -674,23 +674,21 @@ def test_server_unreachable(tctx, connect):
...
-    )
-    if not connect:
-        playbook << CloseConnection(tctx.client)
+        playbook << SendData(
+            tctx.client, BytesMatching(b"502 Bad Gateway.+Connection failed")
+        )
+    else:
+        playbook << SendData(
+            tctx.client, BytesMatching(b"502 Bad Gateway.+CONNECT failed")
+        )
```
mhils, line 686:
```suggestion
    playbook << SendData(
        tctx.client, BytesMatching(b"502 Bad Gateway.+(CONNECT|Connection) failed")
    )
```

mitmproxy/proxy/layers/http/__init__.py
```diff
@@ -756,6 +757,15 @@ def handle_connect_finish(self):
                 self.context.client,
             )
             yield SendHttp(ResponseEndOfMessage(self.stream_id), self.context.client)
+        elif self.flow.response.status_code != 407:
```
errorxyz, line 760: Replacing this condition by an `else` fails this test:
https://github.com/mitmproxy/mitmproxy/blob/968a169077869182414b202520933408dd0a7414/test/mitmproxy/proxy/layers/http/test_http.py#L1636

It seems that we need to send a response to the client after a `connect error` and receive a request from the client in case the status_code is 407, how can we handle such a flow here?

mhils (commented):
Nice progress! 😃

mitmproxy/proxy/layers/http/__init__.py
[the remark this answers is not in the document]
errorxyz: check this https://github.com/mitmproxy/mitmproxy/pull/6906#discussion_r1634966758

mhils (commented):
LGTM % nits! 😃

mitmproxy/proxy/layers/http/_hooks.py
```diff
@@ -96,3 +96,28 @@ class HttpConnectUpstreamHook(commands.StartHook):
...
 
     flow: http.HTTPFlow
+
+
+@dataclass
+class HttpConnectedHook(commands.StartHook):
+    """
+    HTTP CONNECT was successful
+
+    warning::
```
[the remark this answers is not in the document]
errorxyz: Ah silly error, my bad

mhils (approved):
Awesome job! Please rebase onto main, check that the CHANGELOG entry is in the unreleased section, and then merge yourself! 🚀 😃

errorxyz:
I think I've messed up this PR with all the merge conflicts😅, I'll open another PR with these changes

PR #6910, opened 2024-06-11
Autofix generated JS files and do not patch them in tests

mhils:


web/gen/options_js.py
```diff
@@ -0,0 +1,64 @@
...
+                )
+            )
+        print("};")
+
+    await m.done()
+    return s.getvalue()
+
+
+if __name__ == "__main__":
+    filename.write_bytes(asyncio.run(make()).encode())
```
errorxyz, line 64: windows tests are failing because of [this](https://stackoverflow.com/questions/47384652/python-write-replaces-n-with-r-n-in-windows) issue, we should specify the default newline character as `\n`. One way to do this could be:
```suggestion
    filehandle = filename.open(mode='w', newline='\n')
    filehandle.write(asyncio.run(make()))
```

errorxyz (changes requested):
Tested and eveything looks good🚀, one minor issue on windows:

PR #6911, opened 2024-06-11
Export tab that works for Safari

canyesilyurt:
#### Description

For safari users export error fixed.


mhils:
Thanks for the PR! 🍰 

I was curious why this was so broken and fixed it properly in #6917. Turns out we just need to pass in a Promise so that Safari still registers transient activation.

PR #6912, opened 2024-06-11
Add support for DoT

errorxyz:
#### Description

This PR is a WIP to add DoT

#### Checklist


mitmproxy/proxy/mode_servers.py
```diff
@@ -307,18 +317,19 @@ async def listen(
                     )
                 except Exception:  # pragma: no cover
                     logger.debug("Failed to listen on '::', listening on IPv4 only.")
-                    return [ipv4]
+                    servers.append(ipv4)
```
mhils, line 320: Nit: We can do this unconditionally right after the ipv4 assignment now, and then only .append ipv6 in the `else` clause. And/or integrate the else clause into the try part. :)

test/mitmproxy/addons/test_proxyserver.py
```diff
@@ -290,7 +290,7 @@ async def test_dns(caplog_async) -> None:
         ps.running()
         await caplog_async.await_log("DNS server listening at")
         assert ps.servers
-        dns_addr = ps.servers["dns@127.0.0.1:0"].listen_addrs[0]
+        dns_addr = ps.servers["dns@127.0.0.1:0"].listen_addrs[1]
```
mhils, line 293: Why is this necessary? Are `listen_addrs[0]` and `listen_addrs[1]` not the same?

mhils (commented):
Looks great so far 😃 👍

test/mitmproxy/addons/test_proxyserver.py
errorxyz: `test_udp_*` tests fail with the earlier version of the code as there's no response returned from the server. I suspect `listen_addrs[0]` returns the address of the TCP server here and `listen_addrs[1]` is the UDP server's address

mhils: Thanks. I think the point here is that TCP and UDP should listen on the same port if possible. We probably need to extend this workaround here to also apply to UDP:

https://github.com/mitmproxy/mitmproxy/blob/40ae9ff05b3a9abe2f919f40c9ca4319ce8a62ca/mitmproxy/proxy/mode_servers.py#L275-L284

test/mitmproxy/proxy/test_mode_servers.py
```diff
@@ -278,12 +278,11 @@ async def test_udp_start_error():
...
         inst = ServerInstance.make("dns@127.0.0.1:0", manager)
         await inst.start()
-        port = inst.listen_addrs[0][1]
-        inst2 = ServerInstance.make(f"dns@127.0.0.1:{port}", manager)
+        port = inst.listen_addrs[1][1]
         with pytest.raises(
             Exception, match=f"Failed to bind UDP socket to 127.0.0.1:{port}"
         ):
-            await inst2.start()
+            await mitmproxy_rs.start_udp_server("127.0.0.1", port, MagicMock())
```
mhils, line 285: Changes here can be reverted too, right?

mhils (commented):
Nice work! Sneaky trick with reassigning `port` to 0, looks good!

test/mitmproxy/proxy/test_mode_servers.py
errorxyz: Using the DNS server here would raise a similar TCP exception first since we'd be starting the TCP server first in `listen()` but we're checking for UDP here

mhils: Ah, I see. Thanks. Let's adjust the test to use `reverse:udp://127.0.0.1:1234@0` instead of dns then, this is really supposed to test ServerInstance and not mitmproxy_rs. :)

errorxyz: I think you meant `reverse:udp@127.0.0.1:0`

errorxyz: My bad, realised that `reverse:udp@127.0.0.1:0` and `reverse:udp://127.0.0.1:0` perform different things. Can you please explain the difference?

mhils (approved):
This looks great and super clean now! 👍 

Let's get this merged right away and do the next DNS over TCP bits in a separate PR. 😃

Issue #6914, opened 2024-06-12
zstd decompression failure


zendai:
Dear Mitm team,

If the response content-encoding is zstd and the response has multiple frames (larger payloads), Response.content will only return the first frame up to 65536 bytes. This is really annoying as it silently "corrupts" the response. It doesn't raise an exception nor changing the max_output_size makes any difference.

The problem is that the python zstd package mitm is using (zstandard version 0.22, latest) has the decompress function which hasn't implemented yet handling multiple frames, which is [documented](https://python-zstandard.readthedocs.io/en/latest/decompressor.html#zstandard.ZstdDecompressor.decompress). They recommend to use stream_reader() instead (their link in the doc is wrong, it points to the Compressor's stream_reader()).

Currently, this is the [code](https://github.com/mitmproxy/mitmproxy/blob/460789b7beadebfdf3a99fd50c3c1e6004bc47be/mitmproxy/net/encoding.py#L178) we're using:

```python
def decode_zstd(content: bytes) -> bytes:
    if not content:
        return b""
    zstd_ctx = zstd.ZstdDecompressor()
    try:
        return zstd_ctx.decompress(content)
    except zstd.ZstdError:
        # If the zstd stream is streamed without a size header,
        # try decoding with a 10MiB output buffer
        return zstd_ctx.decompress(content, max_output_size=10 * 2**20)
```

Proposed solution is to use either pyzstd (works, tested):
```python
pyzstd.decompress(content)
```

or stick with zstandard and use stream_reader() like this, for example (also works, tested):
```python
dctx = zstd.ZstdDecompressor()
return dctx.stream_reader(io.BytesIO(content)).read()
```

Both of these will return the full payload.

I tested this with the following payload (can't attach the payload itself):

```bash
Frames  Skips  Compressed  Uncompressed  Ratio  Check  Filename
    16      0   188.87 KB     971.52 KB  5.144   None  /tmp/1.zstd
```

sendai

mhils:
> or stick with zstandard and use stream_reader() like this, for example (also works, tested):
> 
> ```python
> dctx = zstd.ZstdDecompressor()
> return dctx.stream_reader(io.BytesIO(content)).read()
> ```

This sounds great to me, thanks for looking into things! A PR would be most welcome. Ideally we could have some sort of test payload to make sure we don't regress in the future. :)

zendai:
Thank you @mhils , will raise a PR soon.

zendai:
[...]

PR #6921, opened 2024-06-12
Fix zstd decomporession

zendai:
#### Description

This is a fix to issue #6914 .

#### Checklist


test/mitmproxy/net/test_zstd.py
```diff
@@ -0,0 +1,21 @@
+from mitmproxy.net.encoding import decode_zstd
```
mhils, line 1: Please add this to `test_encoding.py`. We want to keep parity between source file and test file name. :)

mitmproxy/net/encoding.py
```diff
@@ -174,12 +174,7 @@ def decode_zstd(content: bytes) -> bytes:
     if not content:
         return b""
     zstd_ctx = zstd.ZstdDecompressor()
-    try:
-        return zstd_ctx.decompress(content)
-    except zstd.ZstdError:
-        # If the zstd stream is streamed without a size header,
-        # try decoding with a 10MiB output buffer
-        return zstd_ctx.decompress(content, max_output_size=10 * 2**20)
+    return zstd_ctx.stream_reader(BytesIO(content)).read()
```
mhils, line 177: Why does this work without passing `read_across_frames=True`?

According to https://python-zstandard.readthedocs.io/en/stable/decompressor.html#zstandard.ZstdDecompressor.stream_reader, it's false by default. Is this a bug in the underlying library, or why are things working here?

mhils (commented):
Thanks! 🍰

test/mitmproxy/net/test_zstd.py
zendai: Sure! Thanks for the guidance.

mitmproxy/net/encoding.py
zendai: Fair question, I don't know. I can add explicitly `read_across_frames=True` and test to ensure it works. Fair point, I'd sleep better if the code would align with the documentation. Let me play with this a little, will come back.

test/mitmproxy/net/test_zstd.py
zendai: Moved the test to test_encoding.py.

mitmproxy/net/encoding.py
zendai: So, I tested it and it works both with explicit False and True as well. I tracked down where this flags comes into play during Decompress' stream_reader().[read()](https://github.com/indygreg/python-zstandard/blob/518ade454f41ffbad8b9b542f6938cd308653ba2/zstandard/backend_cffi.py#L3288 ), it's [here ](https://github.com/indygreg/python-zstandard/blob/518ade454f41ffbad8b9b542f6938cd308653ba2/zstandard/backend_cffi.py#L3282) and it's part of an OR condition:

```python
        # Emit data if there is data AND either:
        # a) output buffer is full (read amount is satisfied)
        # b) we're at end of a frame and not in frame spanning mode
        return out_buffer.pos and (
            out_buffer.pos == out_buffer.size
            or zresult == 0
            and not self._read_across_frames
        )
```

I'm not going to pretend that I fully understand this part, I have some assumptions but I'm not familiar with zstd to this depth to say anything here with certainty. Still, I added the `read_across_frames=True` to the PR as based on my tests it works as well and based on the documentation it makes sense to have it.

mhils (approved):

PR #6923, opened 2024-06-13
Feature/implement mvp new tab page

lups2000:
#### Description

In this PR, I've begun implementing the initial MVP for the new **Capture** tab, following the mockup proposed [here](https://github.com/mitmproxy/mitmproxy/discussions/6868). I started implementing the **Regular mode** and a simplified version of the **Local mode**.

TODOs:
 - [x] Improve design mode toggle
 - [x] Remove inline style
 - [x] Make Regular mode functional
 - [x] Make Local mode functional
 - [x] Add new tests

DEMO:

https://github.com/mitmproxy/mitmproxy/assets/100372313/859c5796-21a7-4e65-ab7a-f47e4a2b0b45

#### Checklist


web/src/js/components/Modes/HTTPS.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from "react";
+import Mode, { ModeType } from "./Mode";
+
+export default function HTTPS() {
```
mhils, line 4: Nit: HTTPS -> Regular

web/src/js/components/Modes.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from "react";
+import HTTPS from "./Modes/HTTPS";
+import Local from "./Modes/Local";
+import Wireguard from "./Modes/Wireguard";
+import Reverse from "./Modes/Reverse";
+
+export default function Modes() {
+    return (
+        <div style={{ padding: "1em 2em" }}>
```
mhils, line 9: Nit: I'd just go with a className for the mode tab and then use `.modes h3` instead of `.modes-subheader` below.

web/src/js/components/Modes/HTTPS.tsx
```diff
@@ -0,0 +1,14 @@
...
+    return (
+        <Mode
+            title="Explicit HTTP(S) Proxy"
+            description="You manually configure your client application or device to use an HTTP(S) proxy."
+            type={ModeType.REGULAR}
+        >
+            <p>Run HTTP/S Proxy</p>
+        </Mode>
```
mhils, line 12: My hot take here is that we're erring too much on the side of reusing code here, which causes unnecessary complexity by adding too many layers of abstraction. There is a tradeoff here, and we shouldn't be afraid to have minor amounts of duplication.

This here could also just be like that:

```jsx
export default function Regular() {
    const dispatch = useAppDispatch(),
          active = useAppSelector((state) => state.modes.regular.active);
    return <div>
        <h4>Explicit HTTP(S) Proxy</h4>
        <p class="description">You manually configure your client application or device to use an HTTP(S) proxy.</p>
        <div>
            <label>
                <input type="checkbox" checked={active} onChange={() => dispatch(modeActions.toggleRegular())} />
                Run HTTP/S Proxy
            </label>
        </div>
    </div>
}
```

Yes this adds some duplication, but we now don't have to look at Mode, ModeType, ModeToggle, or MenuToggle. Things become much easier to reason about and test.

mhils (changes requested):
Thanks! Some nits and one major comment. I'd like us to make sure we keep things simple, and also move the business logic here into [ducks/ui](https://github.com/mitmproxy/mitmproxy/tree/main/web/src/js/ducks). 

Eventually we will need to persist the configuration for each mode outside of `.options` anyhow. Consider a user that does the following: 1. Use the cog next to regular mode to set the proxy to port 8085. 2. Uncheck regular mode. 3. Reenable regular mode. If we only update .options, we will remove `regular@8085` in step 2 and then loose the information about the custom port.
(Let me know if that doesn't make sense, happy to expand)

web/src/js/components/Modes/HTTPS.tsx
lups2000: I see, that makes sense! 👍 I thought it was generally best practice to reuse code and avoid duplication, which is why I tried to code it abstractly. I agree, however, that it turned out to be difficult to understand. So, ideally, we'll have a file (similar to the one above) for each mode, regardless of code duplication, right? Regarding the ducks, I still need to understand how they work because they still seem like black magic to me 😅

lups2000:
Hi @mhils, 
I created a first prototype for the regular mode! Before proceeding with the rest, I wanted to know if the structure makes sense. What would you change? (Error handling hasn't been implemented yet). Let me know :)
Thanks!

web/src/js/ducks/modes.ts
```diff
@@ -0,0 +1,32 @@
...
+import regularReducer from "./modes/regularDuck";
+import { fetchApi } from "../utils";
+
+export interface ModeState {
+    active: boolean;
+    listen_port?: number;
+    listen_host?: string;
+}
+
+export const updateMode = async (mode, active) => {
```
mhils, line 11: When calling `await fetchApi.put("/options", { mode: ... })` we always need to pass all currently active modes. So I would propose that updateMode does not take any arguments, but goes and interrogates the current state to assemble the list of modes and the call `fetchApi`. 

`toggleRegular` would just dispatch an action to update `active` and then call `updateMode` (without arguments).

web/src/js/components/Modes/Regular.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from "react";
+import { ModeToggle } from "./ModeToggle";
+import { useAppDispatch, useAppSelector } from "../../ducks";
+import { toggleRegular } from "../../ducks/modes/regularDuck";
```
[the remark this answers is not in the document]
lups2000: 🦆

web/src/js/ducks/modes.ts
lups2000: okay:)

web/src/js/components/Modes/Regular.tsx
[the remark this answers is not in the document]
lups2000: Does it make sense to you to have a duck for each mode?

lups2000:
this PR is becoming huge 🤯 😮‍💨

web/src/js/ducks/modes/local.ts
```diff
@@ -0,0 +1,111 @@
...
+            return {
+                ...state,
+                applications: action.applications,
+                error: undefined,
+            };
+        case UPDATE_OPTIONS:
+        case RECEIVE_OPTIONS:
+            if (action.data && action.data.mode) {
+                const modes = action.data.mode.value;
+                const isActive = modes.some((mode) => mode.includes("local"));
```
mhils, line 83: this triggers for `reverse:http://localhost/` :)

Probably useful to have a `getModesOfType` method like that mirrors what ProxyMode.parse does:

```javascript
> parseMode("foo:x@0")
{
    type: "foo",
    data: "x",
    custom_listen_host: null,
    custom_listen_port: 0,
}
```

and then work with that everywhere.

web/src/js/ducks/modes/local.ts
```diff
@@ -0,0 +1,111 @@
...
+                modes.forEach((mode) => {
+                    if (mode.startsWith("local:")) {
+                        extractedApplications = mode.substring("local:".length);
+                    }
+                });
+                return {
+                    ...state,
+                    active: isActive,
+                    applications:
+                        extractedApplications !== ""
```
mhils, line 94: I think this should be contingent on whether isActive is true or not, not on whether we have a list of applications as a filter.

web/src/js/ducks/modes/reverse.ts
```diff
@@ -0,0 +1,139 @@
...
+    protocols: [
+        { name: "http", isSelected: false },
+        { name: "https", isSelected: false },
+        { name: "dns", isSelected: false },
+        { name: "http3", isSelected: false },
+        { name: "quic", isSelected: false },
+        { name: "tcp", isSelected: false },
+        { name: "tls", isSelected: false },
+        { name: "udp", isSelected: false },
+        { name: "dtls", isSelected: false },
```
mhils, line 31: Is there a reason why this is not
```javascript
protocol: "http" | "https" | "dns" | ...
```
? 😅

web/src/js/ducks/modes/regular.ts
```diff
@@ -0,0 +1,71 @@
...
+    let mode = "regular";
+    if (regularMode.active) {
+        if (regularMode.listen_host) {
+            mode += `@${regularMode.listen_host}`;
+        }
+        if (regularMode.listen_port) {
+            mode += `:${regularMode.listen_port}`;
+        }
+        return mode;
+    }
```
mhils, line 27: This pattern repeats in almost all modes, so maybe that's a candidate for a `addListenAddr(mode)` helper? 😃

mhils (commented):
Excellent progress! 😃 

My proposal for this PR would be to remove the following things from scope:

1. Reverse and WireGuard mode (just remove them from the UI for now)
2. Error handling

and we focus on the comments below to get this merged. Then we focus on those points in a new PR. 😄

web/src/js/components/ProxyApp.tsx
```diff
@@ -77,13 +77,28 @@ class ProxyAppMain extends Component<ProxyAppMainProps, ProxyAppMainState> {
...
     componentDidMount() {
-        window.addEventListener("keydown", this.props.onKeyDown);
+        //window.addEventListener("keydown", this.props.onKeyDown);
+        this.updateKeyDownListener(this.state.ActiveMenu);
+    }
+
+    componentDidUpdate(_, prevState: ProxyAppMainState) {
+        if (prevState.ActiveMenu !== this.state.ActiveMenu) {
+            this.updateKeyDownListener(this.state.ActiveMenu);
+        }
```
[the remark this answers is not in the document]
lups2000: I'm not sure if I understood it correctly. Do we have a keyboard duck? I only see a `Keyboard.tsx`. How can I access the current screen?
P.S. I added this because I couldn't type anything in the input field for the local applications. This was the only solution I found, even though the input will probably disappear in the future.

web/src/js/ducks/modes.ts
```diff
@@ -0,0 +1,56 @@
...
+    return async (_, getState) => {
+        try {
+            const modes = getState().modes;
+
+            const activeModes: string[] = [
+                getRegularModeConfig(modes),
+                getLocalModeConfig(modes),
+                getWireguardModeConfig(modes),
+                getReverseModeConfig(modes),
+            ].filter((mode) => mode !== "");
```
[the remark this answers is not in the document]
lups2000: But in that case, we'll have an array of arrays of strings, right? The **options endpoint** takes an array of strings. Ideally, the `getReverseModeConfig(modes)` function should already return multiple values separated by commas. What do you think?

web/src/js/components/ProxyApp.tsx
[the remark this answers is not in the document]
mhils: It is a bit of a hack, but we have https://github.com/mitmproxy/mitmproxy/blob/main/web/src/js/ducks/ui/keyboard.tsx - I guess it's not a proper duck. 😛 If this is catching events, we can maybe call `e.stopPropagation` similar to how we do it [here](https://github.com/mitmproxy/mitmproxy/blob/95bf441bc3a1dfcac2cb313cac8e72d3716a4bed/web/src/js/components/editors/ValueEditor.tsx#L159). Does that make more sense?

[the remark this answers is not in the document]
lups2000: I see thank you, I'll figure that out :)

web/src/js/ducks/modes.ts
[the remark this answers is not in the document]
mhils: Not sure if I am following.  `getReverseModeConfig(modes)` should return for instance
```
[
  "reverse:http://example.com",
  "reverse:http://example.org@4000"
]
```

Here we can then do
```javascript
const activeModes: string[] = [
    ...getReverseModeConfig(modes),
    ...
]
```

Using the spread syntax (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax), we will end up with one array of strings.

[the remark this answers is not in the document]
lups2000: okay cool, didn't know that :) thank you! Regarding the other modalities, is it fine to return just a string?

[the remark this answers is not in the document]
lups2000: okay :)

web/src/js/ducks/modes/reverse.ts
lups2000: yes I don't know why I opted for this solution, it's probably not necessary to have an `isSelected` flag 😅

lups2000:
Hi @mhils,
I've applied your feedback and completed all the tasks :) Should I add more tests to improve coverage?

web/src/js/ducks/modes/local.ts
```diff
@@ -0,0 +1,103 @@
+import { ModeState, updateMode } from "../modes";
+import {
+    RECEIVE as RECEIVE_OPTIONS,
+    UPDATE as UPDATE_OPTIONS,
+} from "../options";
+import { getModesOfType } from "./utils";
+
+export const TOGGLE_LOCAL = "TOGGLE_LOCAL";
+export const SET_APPLICATIONS = "SET_APPLICATIONS";
```
mhils, line 9: We have two options here: 

1. Prefix events with something like `MODE_LOCAL_` so that we don't get name collisions.
2. Use Redux's createSlice API that avoids this footgun. 😛

web/src/js/__tests__/ducks/tutils.ts
```diff
@@ -128,6 +128,35 @@ export const testState: RootState = {
...
+            protocol:
+                "http" ||
+                "https" ||
+                "dns" ||
+                "http3" ||
+                "quic" ||
+                "tcp" ||
+                "tls" ||
+                "udp" ||
+                "dtls",
```
mhils, line 157: This is a bit of a lazy copy paste here, let's fix it.

mhils (commented):
This looks good so far, but I'm a bit worried that we're getting too much work-in-progress code sneak through in the reverse and wireguard modes. Can you remove those files in a single commit please? Once this is through, you can then revert just this commit and everything should be back.

web/src/js/components/Modes/Local.tsx
```diff
@@ -0,0 +1,46 @@
...
+    );
+
+    React.useEffect(() => {
+        setListApplications(applications || "");
+    }, [applications]);
+
+    const handleListApplicationsChange = (applications: string) => {
+        setListApplications(applications);
+        dispatch(setApplications(applications));
+    };
```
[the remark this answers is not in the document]
lups2000: At the beginning I was handling errors and I noticed a weird behaviour using directly the redux state. Let's delete the additional state and I'll come back to that when I'll manage error handling :)

lups2000:
I removed completely the reverse and the wireguard modes and adjust the last things! It should be fine now 🤞 😃

mhils (commented):
I've integrated the latest capture-tab changes and did a few commits with cleanups. Please take a look at them and extrapolate to the rest. 😄

lups2000:
thank you @mhils 🙏 , I updated the last tests (FIXMEs) according to what you have done before :)

web/src/js/__tests__/ducks/modes/localSpec.tsx
```diff
@@ -0,0 +1,147 @@
...
+        expect(state).toEqual(initialState);
+    });
+
+    it("should handle MODE_LOCAL_SET_APPLICATIONS action", () => {
+        const applications = "app1, app2";
+        const action = { type: MODE_LOCAL_SET_APPLICATIONS, applications };
+        const newState = localReducer(initialState, action);
+        expect(newState.applications).toBe(applications);
+        expect(newState.error).toBeUndefined();
+    });
```
mhils, line 26: We don't want to manually construct actions in tests, that's a maintenance nightmare. The test two below covers this comprehensively, this one can go entirely. :)

The only exception to this is actions for which we don't have action creators (OPTIONS_RECEIVE, ...).

web/src/js/__tests__/ducks/modes/localSpec.tsx
```diff
@@ -0,0 +1,147 @@
...
+    it("should return local mode with multiple applications when active and multiple applications are present", () => {
+        const modes = {
+            local: {
+                active: true,
+                applications: "curl, wget",
+            },
+        };
+        const result = getMode(modes);
+        expect(result).toEqual(["local:curl, wget"]);
+    });
```
mhils, line 120: I do not mind these two tests here, but if you are comfortable with this you can skip such tests if you want. We already have 100% line (and branch) coverage with the tests above, which is more than good enough in my book. 😉

web/src/js/__tests__/ducks/modes/localSpec.tsx
```diff
@@ -0,0 +1,53 @@
...
+                mode: {
+                    value: ["othermode"],
+                },
+            },
+        };
+        const newState = localReducer(initialState, action);
+        expect(newState.active).toBe(false);
+        expect(newState.applications).toBe("");
+        expect(newState.error).toBeUndefined();
+    });
```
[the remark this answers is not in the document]
mhils: ^ this test is still missing. I've rephrased it in case it wasn't clear. :)

web/src/js/__tests__/ducks/modes/regularSpec.tsx
```diff
@@ -0,0 +1,123 @@
...
+        expect(store.getState().modes.regular.active).toBe(false);
+        expect(fetchMock).toHaveBeenCalled();
+    });
+
+    it('should handle RECEIVE_OPTIONS action with data.mode containing "regular", an host and a port', () => {
+        const action = {
+            type: options.RECEIVE,
+            data: {
+                mode: {
+                    value: ["regular@http:8081"],
```
mhils, line 33: Let's use realistic values - `http` does not conceptually make sense as listen_host. We can just go with `127.0.0.1`. It doesn't matter for the test, but inevitably some new contributor will grep through the codebase and be confused by this.

web/src/js/__tests__/ducks/modes/regularSpec.tsx
```diff
@@ -0,0 +1,123 @@
...
+                mode: {
+                    value: ["regular"],
+                },
+            },
+        };
+        const newState = regularReducer(initialState, action);
+        expect(newState.active).toBe(true);
+        expect(newState.listen_host).toBe(undefined);
+        expect(newState.listen_port).toBe(undefined);
+    });
```
mhils, line 70: Same pattern here: These tests aren't too interesting because they hit exactly the same code path as the one above, but we're missing out on the interesting test that would uncover a bug (see below).

web/src/js/ducks/modes/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+            }
+        } catch (error) {
+            //TODO: handle error
+        }
+    };
+};
+
+export const addListenAddr = (mode: ModeState) => {
+    let stringMode = mode.name;
+    if (mode.active) {
```
mhils, line 40: Should `addListenAddr` really peek at `.active`? I wouldn't expect that at all from the name. We should move that out of here and do it manually in each relevant `getMode`.

web/src/js/__tests__/ducks/modes/regularSpec.tsx
```diff
@@ -0,0 +1,123 @@
...
+        const modes = {
+            regular: {
+                active: true,
+                name: "regular",
+                listen_port: 8080,
+            },
+        };
+        const mode = getMode(modes);
+        expect(JSON.stringify(mode)).toBe(JSON.stringify(["regular@8080"]));
+    });
```
mhils, line 122: Two tests - one for active, one for not active is good enough here as long as we test addListenAddr separately at one central place. :)

web/src/js/__tests__/ducks/modesSpec.tsx
```diff
@@ -0,0 +1,37 @@
...
+                name: "regular",
+            },
+            local: {
+                active: false,
+                name: "local",
+                applications: "",
+            },
+        });
+    });
+});
```
mhils, line 37: Move to utils.ts?
(We can also consider using jest fetch mock here)

web/src/js/ducks/modes/regular.ts
```diff
@@ -0,0 +1,64 @@
...
+                const currentModeConfig = getModesOfType(
+                    "regular",
+                    action.data.mode.value,
+                )[0];
+                const isActive = currentModeConfig !== undefined;
+                return {
+                    ...state,
+                    active: isActive,
+                    listen_host:
+                        currentModeConfig?.listen_host || state.listen_host,
```
mhils, line 53: This here is still bugged (and uncaught by tests): If we configure an alternative listen port and then receive an update with just `regular`, that change will not be reflected in the UI. (Same type of issue as we had with local mode).

web/src/js/ducks/modes/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+            stringMode += `@${mode.listen_host}:${mode.listen_port}`;
+        } else if (mode.listen_port) {
+            stringMode += `@${mode.listen_port}`;
+        }
+        return [stringMode];
+    }
+    return [];
+};
+
+const parseMode = (modeConfig) => {
```
mhils, line 51: parseMode needs dedicated tests and probably needs to follow the logic in https://github.com/mitmproxy/mitmproxy/blob/v10.3.1/mitmproxy/proxy/mode_specs.py#L102-L123 more closely.

See https://github.com/mitmproxy/mitmproxy/blob/main/test/mitmproxy/proxy/test_mode_specs.py for inspiration. :) It's a bit easier for us because we can expect well-formed data. :)

web/src/js/__tests__/ducks/modes/regularSpec.tsx
lups2000: yes typo, I use `localhost` which is used below as well :)

web/src/js/ducks/modes/utils.ts
lups2000: mhh okay, but what's the point to move it outside and write it in each `getMode`? The check is always the same and we would repeat the code

web/src/js/ducks/modes/regular.ts
lups2000: Could you please explain it better please?

lups2000: Could you please explain it better please? (btw the logic to change port and host hasn't been implemented yet since we said to start with a single toggle for the regular)

lups2000: Also related to this and to the local mode:
Let's say we write `curl` in the applications field in the **Capture tab**, without activating the local mode. Then we go in the **options modal** and we write `local`. Should we activate `local:curl` or just `local`? At the moment it's the second case. Same thing for the regular mode with the `listen_host` and the `listen_port`. (For the future, does having a mode field in the option modal still make sense once we'll have a dedicated page? 🤔 )

web/src/js/__tests__/ducks/modes/localSpec.tsx
```diff
@@ -0,0 +1,148 @@
...
+            type: options.RECEIVE,
+            data: {
+                mode: {
+                    value: ["local"],
+                },
+            },
+        };
+        const newState = localReducer(initialState, action);
+        expect(newState.active).toBe(true);
+        expect(newState.applications).toBe("curl");
```
mhils, line 84: Great we have this in a test now! But this here looks wrong? UI would show "curl" whereas backend ingests everything without filter.

web/src/js/__tests__/ducks/modes/regularSpec.tsx
```diff
@@ -0,0 +1,87 @@
...
+            data: {
+                mode: {
+                    value: ["regular"],
+                },
+            },
+        };
+        const newState = regularReducer(initialState, action);
+        expect(newState.active).toBe(true);
+        expect(newState.listen_host).toBe(initialState.listen_host);
+        expect(newState.listen_port).toBe(initialState.listen_port);
```
mhils, line 60: Same here - these should be empty after the update I think.

web/src/js/__tests__/ducks/modes/utilsSpec.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { addListenAddr, updateMode } from "../../../ducks/modes/utils";
+import { parseMode } from "../../../ducks/modes/utils";
+
+enableFetchMocks();
+
+jest.mock("../../../utils", () => ({
+    fetchApi: {
+        put: jest.fn(),
+    },
+}));
```
mhils, line 13: Is this still necessary with enableFetchMocks?

web/src/js/__tests__/ducks/modes/utilsSpec.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+describe("updateMode action creator", () => {
+    beforeEach(() => {
+        fetchMock.resetMocks();
+    });
+
+    it("should call updateMode method with a successful api call", async () => {
+        const store = TStore();
+        (fetchApi.put as jest.Mock).mockResolvedValueOnce({ status: 200 });
```
mhils, line 22: Can't we use fetchMock from jest-fetch-mock?

web/src/js/ducks/modes/utils.ts
mhils: We can either go with a separate `skipInactive()` helper func, or we come up with a good function name that conveys this method does both. Maybe something like `includeModeState`? We can then do this:

```javascript
    return includeModeState("regular", modes.regular);
```

web/src/js/ducks/modes/utils.ts
```diff
@@ -0,0 +1,94 @@
+import { getMode as getRegularModeConfig } from "./regular";
+import { getMode as getLocalModeConfig } from "./local";
+import { fetchApi } from "../../utils";
+
+export interface ModeState {
+    active: boolean;
+    name: string;
```
mhils, line 7: I missed this in the initial review: name is not state, so it shouldn't be in state. :) Let's remove that here and make it explicit in the getMode implementation, we need to do that anyhow when we get to modes with arguments.

web/src/js/ducks/modes/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+    } else {
+        mode = head;
+    }
+
+    let host = "";
+    let port: string | number = "";
+
+    if (listenAt) {
+        if (listenAt.includes(":")) {
+            [host, port] = listenAt.split(":");
```
mhils, line 70: If you add tests for IPv6 listen_addrs, this will fail. 😉 

We can put this into web/src/js/utils.ts:
```javascript
export function rpartition(str: string, sep: string): [string, string] {
  const lastIndex = str.lastIndexOf(sep);
  if (lastIndex === -1) {
    return ['', str];
  }
  const before = str.slice(0, lastIndex);
  const after = str.slice(lastIndex + sep.length);
  return [before, after];
}
```

mhils (commented):
This looks much more solid now - thanks!

web/src/js/__tests__/ducks/modes/localSpec.tsx
lups2000: This was a question I added above:

Also related to this and to the local mode:
Let's say we write curl in the applications field in the Capture tab, without activating the local mode. Then we go in the options modal and we write local. Should we activate local:curl or just local? At the moment it's the second case. Same thing for the regular mode with the listen_host and the listen_port.

web/src/js/__tests__/ducks/modes/regularSpec.tsx
lups2000: ^^

web/src/js/ducks/modes/regular.ts
mhils: > Then we go in the **options modal** and we write `local`. Should we activate `local:curl` or just `local`?

Just local. If we get an update from the server with mode X activated, we overwrite our UI state for mode X. If we get a mode update from the server where mode X is not active, we can keep our UI state as it's deactivated anyhow. Does that make sense?

web/src/js/__tests__/ducks/modes/localSpec.tsx
mhils: Answered above, sorry for missing that thread.

web/src/js/ducks/modes/regular.ts
lups2000: okay makes sense :)

web/src/js/ducks/modes/utils.ts
lups2000: example?

lups2000: what do you think with this new version?

lups2000:
just a small "issue" now: I don't know why when toggling the local mode we're getting : `VM105:6 ImmutableStateInvariantMiddleware took 43ms, which is more than the warning threshold of 32ms. 
If your state or actions are very large, you may want to disable the middleware as it might cause too much of a slowdown in development mode. See https://redux-toolkit.js.org/api/getDefaultMiddleware for instructions.
It is disabled in production builds, so you don't need to worry about that.` in the console 🤔

mhils (approved):
🚀🚀🚀

PR #6935, opened 2024-06-17
Parse DNS over TCP messages

errorxyz:
#### Description

Changes made in this PR allow parsing of DNS over TCP messages. This is done by extracting the length label from the raw data and waiting until we receive data of that size.
Also handles query pipelining

Continuation of https://github.com/mitmproxy/mitmproxy/pull/6912

#### Checklist


mitmproxy/addons/next_layer.py
```diff
@@ -167,14 +167,15 @@ def s(*layers):
...
-            else:
-                return layers.DNSLayer(context)
+        is_dns = (
+            udp_based
+            and starts_like_dns_record(data_client)
+            or tcp_based
+            and starts_like_dns_over_tcp_record(data_client)
+        )
+        if is_dns:
+            return layers.DNSLayer(context)
```
mhils, line 177: I think we should just set is_dns based on the destination port (53 or 5353) for now and remove the heuristic methods. `starts_like_dns_over_tcp_record` will currently return False if the first packet is not complete yet, which can happen for DNS over TCP. 53/5353 should be good enough really since you cannot configure the DNS port typically.

mitmproxy/proxy/layers/dns.py
```diff
@@ -91,24 +113,44 @@ def state_query(self, event: events.Event) -> layer.CommandGenerator[None]:
...
         if isinstance(event, events.DataReceived):
+            msg = None
             try:
-                msg = dns.Message.unpack(event.data)
+                if self.context.client.transport_protocol == "udp":
+                    msg = dns.Message.unpack(event.data)
+                elif self.context.client.transport_protocol == "tcp":
+                    if self.expected_size is None:
+                        assert self.buf is None
+                        (self.expected_size,) = _LENGTH_LABEL.unpack_from(event.data, 0)
```
mhils, line 123: Nit: this can fail if the client only sends one byte. More realistically, this will fail if the client does query pipelining and the last send frame had only one byte of the next query.

mitmproxy/proxy/layers/dns.py
```diff
@@ -91,24 +113,44 @@ def state_query(self, event: events.Event) -> layer.CommandGenerator[None]:
...
+                        self.buf = event.data[_LENGTH_LABEL.size :]
+                    else:
+                        assert self.buf is not None
+                        self.buf += event.data
+
+                    if self.expected_size == len(self.buf):
```
mhils, line 129: `>=` - we can have query pipelining.

mhils (commented):
This looks great so far, but definitely needs tests and fuzz tests! :)

mitmproxy/addons/next_layer.py
errorxyz: the following test would fail if we were to use this heuristic:
https://github.com/mitmproxy/mitmproxy/blob/fc24f85929db05b615f5fb6f028a10388dd54de9/test/mitmproxy/addons/test_next_layer.py#L673-L682
how would we want to handle this?

mhils: Let's just change the port for the test. If we find that raw traffic on 53 is a problem (or DNS on other ports), we can come back to this and tune it later. My gut feeling is that we should be fine.

mitmproxy/proxy/layers/dns.py
```diff
@@ -91,24 +148,32 @@ def state_query(self, event: events.Event) -> layer.CommandGenerator[None]:
...
         if isinstance(event, events.DataReceived):
+            msgs: List[dns.Message] = []
             try:
-                msg = dns.Message.unpack(event.data)
+                if self.context.client.transport_protocol == "udp":
+                    assert self.context.server.transport_protocol == "udp"
+                    msgs.append(dns.Message.unpack(event.data))
+                elif self.context.client.transport_protocol == "tcp":
+                    assert self.context.server.transport_protocol == "tcp"
+                    self.buf.extend(event.data)
```
errorxyz, line 158: @mhils can we expect to have race conditions corrupting `DNSLayer.buf` here?
for eg: `DNSLayer.buf` might have incomplete data from a previously pipelined request that might be corrupted when we receive a response from the server before we receive the rest of the query.
One solution I can think of is to have a `DNSLayer.req_buf` and `DNSLayer.resp_buf`

mhils: @errorxyz: yes, absolutely. I didn't realize we're sharing a buffer here, we definitely need to split those. 😅

errorxyz: alright👍

test/mitmproxy/proxy/test_mode_servers.py
```diff
@@ -318,6 +318,33 @@ async def test_udp_dual_stack(caplog_async):
         assert await caplog_async.await_log("stopped")
 
 
+@pytest.mark.parametrize("transport_protocol", ["udp", "tcp"])
+async def test_dns_start_stop(caplog_async, transport_protocol):
```
mhils, line 322: Is there a specific motivation for this test?

(I'm mostly asking because DnsInstance doesn't do anything special, and tests that do socket I/O are expensive in terms of runtime)

mhils (commented):
This looks super great % nits!

mitmproxy/proxy/layers/dns.py
```diff
@@ -91,24 +142,31 @@ def state_query(self, event: events.Event) -> layer.CommandGenerator[None]:
         from_client = event.connection is self.context.client
 
         if isinstance(event, events.DataReceived):
+            msgs: List[dns.Message] = []
             try:
-                msg = dns.Message.unpack(event.data)
+                buf = self.resp_buf
+                if from_client:
+                    buf = self.req_buf
```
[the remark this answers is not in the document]
errorxyz: > Style nit: Let's not pretend one of these two is special/different. We can either do

Makes sense, going forward with the first option

test/mitmproxy/proxy/test_mode_servers.py
errorxyz: I added this test to check if DNS is listening on both TCP and UDP ports

mhils: Fair enough, thanks!

errorxyz:
A couple doubts:
1. Can `expected_size` be zero i.e would `\x00` be a valid message?
2. Do we need to update these tests to use TCP as well?
https://github.com/mitmproxy/mitmproxy/blob/b8fd7eb55de84e385b7eb396c6543867ec2962a9/test/mitmproxy/addons/test_proxyserver.py#L281
https://github.com/mitmproxy/mitmproxy/blob/b8fd7eb55de84e385b7eb396c6543867ec2962a9/test/mitmproxy/proxy/layers/test_modes.py#L176

errorxyz:
How do we fix the failing test?

mhils:
> How do we fix the failing test?

We specify dependency version ranges where possible so that downstream packagers have an easier time packaging mitmproxy. If we realize we're incompatible with an older version, we just bump the minimum in pyproject.toml! I try to make a bit of an effort to keep the range as wide as possible, but if it's not trivial to figure out I just raise it to the latest release.

mitmproxy/proxy/layers/dns.py
```diff
@@ -37,16 +41,30 @@ class DnsErrorHook(commands.StartHook):
...
+
+
 class DNSLayer(layer.Layer):
     """
     Layer that handles resolving DNS queries.
     """
 
     flows: dict[int, dns.DNSFlow]
+    req_buf: bytearray
+    resp_buf: bytearray
```
mhils, line 61: FWIW, bytearray is great here - things are small anyhow, and mutability is useful for this.

mhils (approved):

PR #6941, opened 2024-06-21
test: fix test on host with ipv6 disabled

godsarmy:
#### Description

Fix test failure on a host with ipv6 disabled

#### Checklist

### DIFF
--- a/test/mitmproxy/net/test_local_ip.py
+++ b/test/mitmproxy/net/test_local_ip.py
@@ -1,3 +1,4 @@
+from ...conftest import skip_no_ipv6
 from mitmproxy.net import local_ip
 
 
@@ -9,6 +10,7 @@ def test_get_local_ip():
     local_ip.get_local_ip("invalid!")
 
 
+@skip_no_ipv6
 def test_get_local_ip6():
     # should never error, but may return None depending on the host OS configuration.
     local_ip.get_local_ip6()


test/mitmproxy/net/test_local_ip.py
```diff
@@ -9,6 +10,7 @@ def test_get_local_ip():
     local_ip.get_local_ip("invalid!")
 
 
+@skip_no_ipv6
 def test_get_local_ip6():
     # should never error, but may return None depending on the host OS configuration.
```
mhils, line 15: The test is really here to ensure that we don't error, even in environments without IPv6. What error are you getting?

godsarmy: On a ipv6 diabled host, the error message is like

"OSError: [Errno 97] Address family not supporterd by protocol"

I think skip_no_ipv6 is used to skip the test when detecting ipv6 is diabled on the host

mhils: The test is correct, the underlying method needs to return None instead of raising.

godsarmy: so do you mean the code itself is wrong? If so, i can fix the code.

mhils: Yes, correct. Thanks!

PR #6942, opened 2024-06-21
fix: OSError raised when ipv6 is disabled

godsarmy:
#### Description

When ipv6 is disabled completed (e.g, in grub), _socket.socket_ throws OSError exception.

The _get_local_ipv6_ function should return None in this case

#### Checklist


mhils (approved):
Thanks! 🍰

Issue #6944, opened 2024-06-23
Non-linear growth in processing time with mitmproxy regarding packet size


jackfromeast:
#### Problem Description
Recently, I've observed an additional overhead that impacts performance which is primarily caused by the `mitmproxy` proxy server. Through my analysis, I identified two critical issues:

1. Non-linear Growth in Processing Time: The processing time for network requests increases non-linearly with the size of the packets. This means that as the packet size grows, the time required to process these requests increases disproportionately.

2. Significant Slowdown with Concurrent Requests: The processing time for network requests slows down dramatically when handling multiple requests simultaneously.

The worst-case scenario occurs when there is a high volume of concurrent requests, especially when some of these requests are significantly large. In such cases, the performance degradation is severe, leading to substantial delays and inefficiencies in processing.

I demonstrated these findings through two experimental settings: 1/ **Sequential Requests**: The first HTML page loads the JavaScript files one by one. Each request is triggered by the `onload` event of the previously requested JavaScript file. 2/ **Concurrent Requests**: The second HTML request loads all the JavaScript files simultaneously. For these experiments, I created 20 JavaScript files ranging from 1MB to 100MB, increasing by 5MB each. I host the proxy and test server on the `localhost` with different ports.

#### Sequential Requests: Without mitmproxy vs. With mitmproxy

![mitmproxy_performance_impact-1](https://github.com/mitmproxy/mitmproxy/assets/51242003/41ae3a87-a3e7-47b8-906c-99a090a41d92)

| Name                | Size     | Loading Time (Without mitmproxy) | Loading Time (With mitmproxy) |
|---------------------|----------|----------------------------------|-------------------------------|
| large_script_1.js   | 1.0 MB   | 4 ms                             | 71 ms                         |
| large_script_2.js   | 6.3 MB   | 35 ms                            | 62 ms                         |
| large_script_3.js   | 11.5 MB  | 83 ms                            | 195 ms                        |
| large_script_4.js   | 16.8 MB  | 142 ms                           | 272 ms                        |
| large_script_5.js   | 22.0 MB  | 201 ms                           | 416 ms                        |
| large_script_6.js   | 27.3 MB  | 254 ms                           | 602 ms                        |
| large_script_7.js   | 32.5 MB  | 294 ms                           | 818 ms                        |
| large_script_8.js   | 37.7 MB  | 357 ms                           | 1.78 s                        |
[...]

PR #6952, opened 2024-06-24
Fix issue #6944: Optimize response/request_body_buf by using list of byte chunks to avoid concatenation overhead

jackfromeast:
#### Description

Fix issue #6944: Optimize response/request_body_buf by using a list of byte chunks to avoid concatenation overhead. 

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Add `HttpConnectedHook` and `HttpConnectErrorHook`
   ([#6930](https://github.com/mitmproxy/mitmproxy/pull/6930), @errorxyz)
 * Fix the issue of non-linear growth in processing time in mitmproxy related to packet size.
+  ([#6952](https://github.com/mitmproxy/mitmproxy/pull/6952), @jackfromeast)
 
 ## 12 June 2024: mitmproxy 10.3.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@
 * Add `HttpConnectedHook` and `HttpConnectErrorHook`
   ([#6930](https://github.com/mitmproxy/mitmproxy/pull/6930), @errorxyz)
 * Fix the issue of non-linear growth in processing time in mitmproxy related to packet size.
-  ([#6944](https://github.com/mitmproxy/mitmproxy/issues/6944), @jackfromeast)
 
 ## 12 June 2024: mitmproxy 10.3.1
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
   ([#6921](https://github.com/mitmproxy/mitmproxy/pull/6921), @zendai)
 * Add `HttpConnectedHook` and `HttpConnectErrorHook`
   ([#6930](https://github.com/mitmproxy/mitmproxy/pull/6930), @errorxyz)
+* Fix the issue of non-linear growth in processing time in mitmproxy related to packet size.
+  ([#6944](https://github.com/mitmproxy/mitmproxy/issues/6944), @jackfromeast)
```
errorxyz, line 15: This line should be added automatically by the CI, you should remove this line

mitmproxy/proxy/layers/http/__init__.py
```diff
@@ -145,8 +145,10 @@ class DropStream(HttpCommand):
 
 
 class HttpStream(layer.Layer):
-    request_body_buf: bytes
-    response_body_buf: bytes
+    request_body_buf: list
+    response_body_buf: list
```
errorxyz, line 149: I think a `bytearray` type might be more appropriate for this purpose

mhils: I _think_ `list[bytes]` might be better here: Appending is amortized O(1), so N slices can be appended in O(n). Joining is O(n) as well: https://github.com/python/cpython/blob/65a12c559cbc13c2c5a4aa65c76310bd8d2051a7/Objects/stringlib/join.h#L57-L130. 

For bytearray I am not so sure. Is there anything in the Python docs that talks about how it does memory allocation?

mhils:
```
class ReceiveBuffer:
    _chunks: list[bytes]
    _len: int

    def __init__(self):
        self._chunks = []
        self._len = 0

    def __iadd__(self, other: bytes):
        assert isinstance(other, bytes)
        self._chunks.append(other)
        self._len += len(other)
        return self

    def __len__(self):
        return self._len

    def __bytes__(self):
        return b"".join(self._chunks)

class ReceiveBufferWithByteArray:
    _buf: bytearray

    def __init__(self):
        self._buf = bytearray()

    def __iadd__(self, other: bytes):
        self._buf.extend(other)
        return self

    def __len__(self):
        return len(self._buf)

    def __bytes__(self):
        return bytes(self._buf)


if __name__ == "__main__":
    import time
    chunk_size = 65_000  # ~ one TCP packet
    num_chunks = 1024 * 1024 * 100 // chunk_size  # 100mb

    data_chunk = b'a' * chunk_size

    def bench(buf: ReceiveBuffer | ReceiveBufferWithByteArray):
        start = time.time()

        for _ in range(num_chunks):
            buf += data_chunk
        ret = bytes(buf)

        print(buf.__class__.__name__, time.time() - start)
        return ret

    _ = bench(ReceiveBuffer())
    _ = bench(ReceiveBufferWithByteArray())
```

```
ReceiveBuffer 0.021017074584960938
ReceiveBufferWithByteArray 0.2989828586578369
```

:-)

Issue #6951, opened 2024-06-24
Show More Not working


Botytec:
I'm trying to look at this WebSocket data in mitmweb, after I click Show More 2 times and the button no longer works. How do I fix this?

lups2000:
Hi, thank you! Could you please be more precise, perhaps with a screenshot? Did you get any errors in the console?

Botytec:
https://streamable.com/yv3f14
here's a video of it happening

Prinzhorn:
Can confirm, second click on button does nothing. No error in JS console.

Prinzhorn:
For some reason it is hardcoded to double the lines once and then that's it:

https://github.com/mitmproxy/mitmproxy/blob/7f7cb28d32cac2752bf00cfdb8343156855425e6/web/src/js/components/FlowView/Messages.tsx#L26-L29

PRs welcome

mhils:
This Math.max should probably have been a Math.min. 🫣

lups2000:
> This Math.max should probably have been a Math.min. 🫣

I quickly tried to change that, and apparently, it's not the only problem 😅 It's weird since, in other cases, the button seems to work fine, while with web socket messages, it doesn't 🤔

Prinzhorn:
> This Math.max should probably have been a Math.min. 🫣

I don't think that's it, with `Math.min()` it would never grow over 1024. I think the `Math.max()` was meant for cases where someone defined a very small `content_view_lines_cutoff` and clicking the button would then only give you a few lines. This makes sure you get _at least_ 1024 when clicking the button.

Why is this line mutating state directly?

https://github.com/mitmproxy/mitmproxy/blob/7f7cb28d32cac2752bf00cfdb8343156855425e6/web/src/js/components/FlowView/Messages.tsx#L81

Prinzhorn:
I've fixed it locally

PR #6956, opened 2024-06-25
Added eslint

Prinzhorn:
Kind of wild that we don't have eslint set up. It would have _somewhat_ caught #6951

# Do not merge yet

This is mostly for discussion. Right now we're at::

> ✖ 414 problems (414 errors, 0 warnings)
>  183 errors and 0 warnings potentially fixable with the `--fix` option.

So if anyone is bored, feel free to set up eslint properly and fix all the issues :smile:

### DIFF
--- a/web/.eslintrc.js
+++ b/web/.eslintrc.js
@@ -3,26 +3,30 @@ module.exports = {
         browser: true,
         es2021: true,
     },
-    extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended', 'plugin:react/recommended'],
+    extends: [
+        "eslint:recommended",
+        "plugin:@typescript-eslint/recommended",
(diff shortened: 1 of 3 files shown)


mhils:
We used to have eslint, but apparently I removed it in https://github.com/mitmproxy/mitmproxy/commit/0e9206ceb11b1c4a3ea298e6b5bda54b9baf08ee as part of the transition to esbuild. IIRC it wasn't that easy to set up with TypeScript back then. Happy to bring it back.

Strategically, I'd like to wait until @lups2000's branch is in to avoid any merge conflicts. Let's prioritize getting that into main to unblock efforts like this one. :)

mhils:
When adopting eslint, we should also run it with `--fix` as part of our autofixer. :)

Issue #6955, opened 2024-06-25
Tracking Issue: MVP Capture Tab


lups2000:
### Overview

We're adding a new tab (**Capture**) to improve the user experience for mitmproxy's users. We're following the mockup we discussed [here](https://github.com/mitmproxy/mitmproxy/discussions/6868).

#### Current Status

- [x] Regular mode (just a toggle)
- [x] Local mode (simple input field)
- [x] Wireguard mode (just a toggle)
- [x] Reverse mode (just one)
- [x] Error handling
- [x] Change `listen_port` in the Regular mode

PR #6972, opened 2024-06-28
Feature/error handling regular mode

lups2000:
#### Description

In this PR, I started implementing error handling for the **regular mode**. 
Reference: #6955 

#### Checklist

### DIFF
--- a/mitmproxy/addons/proxyserver.py
+++ b/mitmproxy/addons/proxyserver.py
@@ -77,7 +77,7 @@ class Servers:
             ]
 
             if not start_tasks and not stop_tasks:
-                return True # nothing to do
+                return True  # nothing to do
 
             self._instances = new_instances
             # Notify listeners about the new not-yet-started servers.
--- a/mitmproxy/addons/proxyserver.py
+++ b/mitmproxy/addons/proxyserver.py
@@ -76,6 +76,9 @@ class Servers:
                 if spec not in new_instances
             ]
 
+            if not start_tasks and not stop_tasks:
+                return True # nothing to do
+
             self._instances = new_instances
             # Notify listeners about the new not-yet-started servers.
             await self.changed.send()
--- a/mitmproxy/addons/proxyserver.py
+++ b/mitmproxy/addons/proxyserver.py
@@ -77,7 +77,9 @@ class Servers:
             ]
 
             if not start_tasks and not stop_tasks:
-                return True  # nothing to do, so we don't need to trigger `self.changed`.
+                return (
+                    True  # nothing to do, so we don't need to trigger `self.changed`.
+                )
 
             self._instances = new_instances
             # Notify listeners about the new not-yet-started servers.
--- a/mitmproxy/addons/proxyserver.py
+++ b/mitmproxy/addons/proxyserver.py
@@ -77,7 +77,7 @@ class Servers:
             ]
 
             if not start_tasks and not stop_tasks:
-                return True  # nothing to do
+                return True  # nothing to do, so we don't need to trigger `self.changed`.
 
             self._instances = new_instances
             # Notify listeners about the new not-yet-started servers.
--- a/web/src/js/components/Modes/Regular.tsx
+++ b/web/src/js/components/Modes/Regular.tsx
@@ -7,11 +7,9 @@ import ValueEditor from "../editors/ValueEditor";
 export default function Regular() {
     const dispatch = useAppDispatch();
 
-    const { active, error, listen_port } = useAppSelector(
-        (state) => state.modes.regular,
-    );
+    const { active, error: ui_error, listen_port } = useAppSelector((state) => state.modes.regular);
 
-    const error_state = useAppSelector((state) => {
+    const backend_error = useAppSelector((state) => {
         if (state.backendState.servers[0]) {
             return state.backendState.servers[0].last_exception;
         }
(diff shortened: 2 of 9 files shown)


lups2000:
I'm opening this PR now but I'm facing a problem with this test `should handle error when toggling regular`, so the pipeline is failing 😅

web/src/js/components/Modes/Regular.tsx
```diff
@@ -33,7 +39,11 @@ export default function Regular() {
                     onEditDone={(port) => handlePortChange(port)}
                 />
             </ModeToggle>
-            {error && <div className="mode-error text-danger">{error}</div>}
+            {(error || error_state) && (
```
mhils, line 42: Let's rename `error` to `ui_error` and `error_state` to `backend_error` to make more clear what's going on here?

web/src/js/components/Modes/Regular.tsx
```diff
@@ -11,6 +11,12 @@ export default function Regular() {
         (state) => state.modes.regular,
     );
 
+    const error_state = useAppSelector((state) => {
+        if (state.backendState.servers[0]) {
```
mhils, line 15: Right now servers is an array with an undefined order, we can't expect `[0]` to be regular mode. :)

web/src/js/ducks/modes/regular.ts
```diff
@@ -51,11 +52,13 @@ const regularReducer = (state = initialState, action): RegularState => {
...
+                error: undefined,
             };
         case MODE_REGULAR_SET_PORT:
             return {
                 ...state,
                 listen_port: action.port as number,
+                error: undefined,
             };
         case UPDATE_OPTIONS:
         case RECEIVE_OPTIONS:
```
mhils, line 64: We also need to potentially reset the error here, don't we?

web/src/js/components/Modes/Regular.tsx
lups2000: make sense sure :)

web/src/js/components/Modes/Regular.tsx
lups2000: Ahhh right! Could we change that? something like `state.backendState.servers[regular]...` ?

web/src/js/ducks/modes/regular.ts
lups2000: What do you mean? We're resetting it now. If there is an error from port X and we change the port, the error will reset. If everything is fine, no new error will be displayed. Otherwise, a new error will appear

web/src/js/components/Modes/Regular.tsx
lups2000: Which solution is better in your opinion?
```
export interface ServerInfo {
    description: string;
    full_spec: string;
    is_running: boolean;
    last_exception: string | null;
    listen_addrs: [string, number][];
    //type: string;
    wireguard_conf?: string;
}

export interface BackendState {
    available: boolean;
    version: string;
    contentViews: string[];
    servers: { [type: string] : ServerInfo};
}
```
or
```
export interface ServerInfo {
    description: string;
    full_spec: string;
    is_running: boolean;
    last_exception: string | null;
    listen_addrs: [string, number][];
    type: string;
    wireguard_conf?: string;
}

export interface BackendState {
    available: boolean;
    version: string;
    contentViews: string[];
    servers: {
        regular: ServerInfo | undefined;
        local: ServerInfo | undefined;
        wireguard: ServerInfo | undefined;
        reverse: ServerInfo | undefined;
    };
}
```
or we can keep the list and check the type 😅

mhils: I think 1) is best. It probably should still have a `type` field because the spec can be something like `regular@8085` and we don't want to parse it to figure out it's regular mode. :)

2) falls short for multiple reverse proxies for instance.

lups2000: alright thanks! Let's go for 1! :)

web/src/js/ducks/modes/regular.ts
mhils: 1. User enters an occupied port. -> error is set.
2. A different tab updates the currently active modes to just "regular" (8080 is not taken yet)
3. UPDATE_OPTIONS is triggered.
4. User will see updated config, but UI is still in the error state

I think we need to reset `error` as well if isActive is true.

web/src/js/components/Modes/Regular.tsx
lups2000: I realize now that I need to change some things on the backend as well. For this reason, does it make sense to make these changes in this PR and include them in our MVP? Is it worth it?

lups2000: we could do something like this now for each mode: 
```
const backend_error = useAppSelector((state) => {
        if (state.backendState.servers) {
            for (const server of state.backendState.servers) {
                if (server.type === "regular") {
                    console.log(server.last_exception);
                    return server.last_exception;
                }
            }
        }
        return "";
    });
```

mitmproxy/addons/proxyserver.py
```diff
@@ -76,6 +76,9 @@ async def update(self, modes: Iterable[mode_specs.ProxyMode]) -> bool:
                 if spec not in new_instances
             ]
 
+            if not start_tasks and not stop_tasks:
+                return True  # nothing to do
```
mhils, line 80:
```suggestion
                return True  # nothing to do, so we don't need to trigger `self.changed`.
```

web/src/js/components/Modes/Regular.tsx
mhils: Let's get this one here flying with the current crappy solution, and once we're in main again we can follow up with a quick PR that changes it to 1).

mhils (approved):
LGTM % nits!

mhils:
Please merge yourself once test failures are addressed. :)

lups2000:
okay thanks :) I'm fixing the tests right now

lups2000:
it should be fine now, I'll merge it then :)

Issue #6739, opened 2024-03-15
Add `dns_server` option.


mhils:
#### Problem Description

This has come up a few times now: mitmproxy currently just opens asyncio sockets with the unresolved hostnames. Python's stdlib will then use libc's getaddrinfo, which always includes /etc/hosts. This can be very annoying, see e.g. https://github.com/mitmproxy/mitmproxy/issues/6737. We want a way to teach mitmproxy to use a custom DNS server.

#### Proposal

1. Implement an async `getaddrinfo` method that resolves a hostname using a custom DNS server. This could be done in either Python or Rust in mitmproxy-rs. My first best guess would be to go with https://docs.rs/hickory-resolver/latest/hickory_resolver/ in Rust.
2. Write an addon that defines a `dns_server` option and monkeypatches/replaces the current event loop's `getaddrinfo` to use our custom implementation if `dns_server` is set.

Ideally we start with a PR that does just the first part + tests before getting to part two.

oge1ata:
Hi @mhils Could I have a bit of explanation on this?

smenon02:
I can help with this! If anyone is currently working on this

smenon02:
anyone have any advice for creating test cases or testing the getaddrinfo method?

MiirzaBaig:
Can anyone elaborate more on this!
Has the issue resolved ?

PR #6975, opened 2024-06-29
Support all query types in DNS mode

errorxyz:
#### Description

Uses `mitmproxy_rs.DnsResolver` to resolve domain names that need to be checked in the hosts file first and forwards any other type of query to the server specified in the options
Fixes https://github.com/mitmproxy/mitmproxy/issues/6739

#### Checklist

~TODO: check and fix tests after we release latest version of `mitmproxy_rs`~

### DIFF
--- a/mitmproxy/addons/dns_resolver.py
+++ b/mitmproxy/addons/dns_resolver.py
@@ -5,19 +5,16 @@ from collections.abc import Sequence
 from functools import cached_property
 
 import mitmproxy_rs
-
 from mitmproxy import ctx
 from mitmproxy import dns
 from mitmproxy.proxy import mode_specs
-
-IP4_PTR_SUFFIX = ".in-addr.arpa"
-IP6_PTR_SUFFIX = ".ip6.arpa"
 
 
 class ResolveError(Exception):
     """Exception thrown by different resolve methods."""
 
     def __init__(self, response_code: int) -> None:
+        assert response_code != dns.response_codes.NOERROR
         self.response_code = response_code
 
 
@@ -83,7 +80,7 @@ class DnsResolver:
             if all_ip_lookups and ctx.options.use_hosts_file:
                 # TODO: We need to handle overly long responses here.
                 flow.response = await self.resolve_message(flow.request)
-            else:
+            elif not flow.server_conn.address:
                 flow.server_conn.address = (self.name_servers[0], 53)
 
     async def resolve_message(self, message: dns.Message) -> dns.Message:
@@ -109,9 +106,7 @@ class DnsResolver:
         except socket.gaierror as e:
             if e.args[0] == "NXDOMAIN":
                 raise ResolveError(dns.response_codes.NXDOMAIN)
-            elif e.args[0] == "NOERROR":
-                raise ResolveError(dns.response_codes.NOERROR)
-            else:
+            else: # pragma: no cover
                 raise ResolveError(dns.response_codes.SERVFAIL)
 
         return map(
--- a/mitmproxy/addons/dns_resolver.py
+++ b/mitmproxy/addons/dns_resolver.py
@@ -5,6 +5,7 @@ from collections.abc import Sequence
 from functools import cached_property
 
 import mitmproxy_rs
+
 from mitmproxy import ctx
 from mitmproxy import dns
 from mitmproxy.proxy import mode_specs
@@ -106,7 +107,7 @@ class DnsResolver:
         except socket.gaierror as e:
             if e.args[0] == "NXDOMAIN":
                 raise ResolveError(dns.response_codes.NXDOMAIN)
-            else: # pragma: no cover
+            else:  # pragma: no cover
                 raise ResolveError(dns.response_codes.SERVFAIL)
 
         return map(
--- a/mitmproxy/addons/dns_resolver.py
+++ b/mitmproxy/addons/dns_resolver.py
@@ -22,35 +22,34 @@ class ResolveError(Exception):
 class DnsResolver:
     def load(self, loader):
         loader.add_option(
-            "use_hosts_file",
+            "dns_use_hosts_file",
             bool,
             True,
             "Use the hosts file for DNS lookups in regular DNS mode/wireguard mode.",
         )
 
         loader.add_option(
-            "name_servers",
+            "dns_name_servers",
             Sequence[str],
             [],
             "Name servers to use for lookups. Default: operating system's name servers",
         )
 
     def configure(self, updated):
-        if "use_hosts_file" in updated or "name_servers" in updated:
-            del self.resolver
-            del self.name_servers
+        if "dns_use_hosts_file" in updated or "dns_name_servers" in updated:
+            self.__dict__.pop('resolver', None)
+            self.__dict__.pop('name_servers', None)
 
     @cached_property
     def resolver(self) -> mitmproxy_rs.DnsResolver:
         return mitmproxy_rs.DnsResolver(
-            name_servers=ctx.options.name_servers
-            or mitmproxy_rs.get_system_dns_servers(),
-            use_hosts_file=ctx.options.use_hosts_file,
+            name_servers=self.name_servers,
+            use_hosts_file=ctx.options.dns_use_hosts_file,
         )
 
     @cached_property
     def name_servers(self) -> list[str]:
-        return ctx.options.name_servers or mitmproxy_rs.get_system_dns_servers()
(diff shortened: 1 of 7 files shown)


errorxyz:
How do I tell the tests to use the updated `mitmproxy_rs`?

mhils:
> How do I tell the tests to use the updated `mitmproxy_rs`?

This currently requires us to ship a new release of mitmproxy_rs. There is an upper bound in https://github.com/mitmproxy/mitmproxy/blob/main/pyproject.toml#L45 though, so we can safely do that without affecting our users.

mhils:
Regarding this PR: Could we just expose a `get_system_dns_server()` from Rust (that uses https://docs.rs/hickory-resolver/latest/hickory_resolver/system_conf/fn.read_system_conf.html internally) instead? We can then treat DNS mode just like we treat reverse DNS mode.

(We still need our getaddrinfo for mitmproxy.proxy.server - but the addon here we can maybe simplify dramatically 😃)

errorxyz:
> that uses https://docs.rs/hickory-resolver/latest/hickory_resolver/system_conf/fn.read_system_conf.html internally

This is available for UNIX only

mhils:
> > that uses https://docs.rs/hickory-resolver/latest/hickory_resolver/system_conf/fn.read_system_conf.html internally
> 
> This is available for UNIX only

The docs are misleading here - it's available for Windows as well if you look at the sources. :)

mitmproxy/addons/dns_resolver.py
```diff
@@ -155,7 +38,85 @@ async def dns_request(self, flow: dns.DNSFlow) -> None:
...
+            "use_hosts_file",
+            bool,
+            True,
+            "Use the hosts file for DNS lookups in regular DNS mode/wireguard mode.",
+        )
+
+        loader.add_option(
+            "name_servers",
+            Sequence[str],
+            mitmproxy_rs.get_system_dns_servers(),
```
mhils, line 108: Let's default this to empty list and then do `ctx.options.name_servers or mitmproxy_rs.get_system_dns_servers()` below. Two reasons:

1. get_system_dns_servers() may potentially raise.
2. The config in `_options_gen.ts` shouldn't be platform specific.

mhils (commented):
Looks great, thanks! I'm curious whether we will run into any issues with overly long records or so, but for now this is a great way forward. 👍

mitmproxy/addons/dns_resolver.py
errorxyz: Do you think we should implement a `Sequence[str] | None` option type for this case?

errorxyz:
> I'm curious whether we will run into any issues with overly long records or so, but for now this is a great way forward

I don't think we'll run into this issue practically since we're only handling A/AAAA records(they typically won't cross 512 bytes). That TODO is a remnant of the old code and theoretically speaking, we *may* have a response size greater than 512 bytes that would need to be handled.

mitmproxy/addons/dns_resolver.py
```diff
@@ -155,7 +67,56 @@ async def dns_request(self, flow: dns.DNSFlow) -> None:
...
+        try:
+            if question.type == dns.types.A:
+                addrinfos = await self.resolver.lookup_ipv4(question.name)
+            elif question.type == dns.types.AAAA:
+                addrinfos = await self.resolver.lookup_ipv6(question.name)
+        except socket.gaierror as e:
+            if e.args[0] == "NXDOMAIN":
+                raise ResolveError(dns.response_codes.NXDOMAIN)
+            else:  # pragma: no cover
+                raise ResolveError(dns.response_codes.SERVFAIL)
```
errorxyz, line 111: How do we want to handle `NOERROR` codes here i.e `if e.args[0] == "NOERROR"`? 
(we can get a `NOERROR` here if there are no records returned but the name is valid)
https://datatracker.ietf.org/doc/html/rfc2308#section-2.2

mitmproxy/addons/dns_resolver.py
```diff
@@ -19,129 +19,41 @@ def __init__(self, response_code: int) -> None:
...
-        dns.ResourceRecord(
-            name=question.name,
-            type=question.type,
-            class_=question.class_,
-            ttl=dns.ResourceRecord.DEFAULT_TTL,
-            data=dns.domain_names.pack(name),
+class DnsResolver:
+    def load(self, loader):
+        loader.add_option(
+            "use_hosts_file",
```
mhils, line 25: We generally suck at this, but maybe we should prefix the new options with `dns_`?

mitmproxy/addons/dns_resolver.py
mhils: Then we should also send a NOERROR to the client, no?

test/mitmproxy/addons/test_dns_resolver.py
```diff
@@ -32,107 +25,86 @@ async def test_simple(monkeypatch):
...
-        lambda rr: rr.domain_name == "dns.google",
+        res_old = dr.resolver
+        tctx.options.use_hosts_file = False
+        assert dr.resolver != res_old
+
+
+async def lookup_ipv4(name: str):
+    if name == "not.exists":
+        raise socket.gaierror("NXDOMAIN")
+    return await asyncio.sleep(0, ["8.8.8.8"])
```
mhils, line 44: Not sure if I want to know - what is this sleep here about? 😄

test/mitmproxy/addons/test_dns_resolver.py
```diff
@@ -1,25 +1,18 @@
...
 from mitmproxy.test import tflow
 from mitmproxy.test import tutils
 
 
-async def test_simple(monkeypatch):
-    monkeypatch.setattr(
-        dns_resolver, "resolve_message", lambda _, __: asyncio.sleep(0, "resp")
-    )
-
+async def test_simple():
```
mhils, line 15:
```suggestion
async def test_ignores_reverse_mode():
```

mhils (commented):
LGTM % nits! 😃

mitmproxy/addons/dns_resolver.py
errorxyz: What about the authority section? Can we ignore that?

test/mitmproxy/addons/test_dns_resolver.py
errorxyz: oops, my bad

mitmproxy/addons/dns_resolver.py
```diff
@@ -19,129 +19,41 @@ def __init__(self, response_code: int) -> None:
...
+    def configure(self, updated):
+        if "use_hosts_file" in updated or "name_servers" in updated:
+            del self.resolver
+            del self.name_servers
+
+    @cached_property
+    def resolver(self) -> mitmproxy_rs.DnsResolver:
+        return mitmproxy_rs.DnsResolver(
+            name_servers=ctx.options.name_servers
+            or mitmproxy_rs.get_system_dns_servers(),
```
errorxyz, line 47:
```suggestion
            name_servers=self.name_servers
```
Adding this change along with the suggested ones

mitmproxy/addons/dns_resolver.py
```diff
@@ -19,129 +19,41 @@ def __init__(self, response_code: int) -> None:
...
+            "Name servers to use for lookups. Default: operating system's name servers",
         )
-    elif question.type == dns.types.AAAA:
-        return await resolve_question_by_name(
-            question, loop, socket.AddressFamily.AF_INET6, ipaddress.IPv6Address
+
+    def configure(self, updated):
+        if "use_hosts_file" in updated or "name_servers" in updated:
+            del self.resolver
+            del self.name_servers
```
errorxyz, line 41: Changing both options at the same time results in an `AttributeError` perhaps we want to use:
```suggestion
    def configure(self, updated):
        if "use_hosts_file" in updated or "name_servers" in updated:
            self.__dict__.pop('resolver', None)
            self.__dict__.pop('name_servers', None)
```

mitmproxy/addons/dns_resolver.py
mhils: Yes, I'd say let's just ignore that for now. :)

mhils:
https://github.com/mitmproxy/mitmproxy/pull/6975#discussion_r1676522249, otherwise LGTM :)

mhils (approved):

errorxyz:
Sorry, messed up the commit history a bit while hopping between windows and linux. Should be fixed now

mhils:
No worries, this is why we squash! :)

PR #6992, opened 2024-07-02
Fix QUIC version negotiation test

jlaine:
To exercise incompatible version negotiation, we must not use version 0, as this would cause the client's INITIAL packet to be erroneously interpreted as a VERSION_NEGOTIATION packet - but with a corrupt format.

So far this has not caused any errors, because aioquic's `pull_quic_header` did not fully parse VERSION_NEGOTIATION packets, but this will change with the next aioquic version.

Use one of the reserved version numbers of the form 0x?a?a?a?a, as defined in https://datatracker.ietf.org/doc/html/rfc9000#section-15


mhils (approved):
Thanks! 🍰🍰🍰

PR #6993, opened 2024-07-02
QUIC: Fix Supported Versions

mhils:
As suggested by @jlaine via email. :)


meitinger:
Definitely more stable :)

PR #6995, opened 2024-07-02
Feature/error handling wireguard local

lups2000:
#### Description

In this PR, I've added error handling for both the **wireguard** and **local** modes. 

#### Checklist

### DIFF
--- a/web/src/js/ducks/modes/regular.ts
+++ b/web/src/js/ducks/modes/regular.ts
@@ -30,7 +30,9 @@ export const toggleRegular =
         const result = await dispatch(updateModeFunc());
 
         if (!result.success) {
-            dispatch({ type: MODE_REGULAR_ERROR, error: result.error });
+            if (result.error.includes("regular")) {
+                dispatch({ type: MODE_REGULAR_ERROR, error: result.error });
+            }
         }
     };
 
@@ -42,7 +44,9 @@ export const setPort =
         const result = await dispatch(updateModeFunc());
 
         if (!result.success) {
-            dispatch({ type: MODE_REGULAR_ERROR, error: result.error });
+            if (result.error.includes("regular")) {
+                dispatch({ type: MODE_REGULAR_ERROR, error: result.error });
+            }
         }
     };
 
--- a/web/src/js/ducks/modes/wireguard.ts
+++ b/web/src/js/ducks/modes/wireguard.ts
@@ -26,17 +26,17 @@ export const getMode = (modes) => {
     return includeModeState("wireguard", wireguardMode);
 };
 
-export const toggleWireguard = () => {
-    return async (dispatch) => {
+export const toggleWireguard =
+    (updateModeFunc = updateMode) =>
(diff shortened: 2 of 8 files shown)


web/src/js/ducks/modes/regular.ts
```diff
@@ -18,8 +18,11 @@ export const initialState: RegularState = {
 };
 
 export const getMode = (modes) => {
-    const regularMode = modes.regular;
-    return includeModeState("regular", regularMode);
+    const regularMode: RegularState = modes.regular;
+    if (!regularMode.error) {
```
mhils, line 22: Can't we have `includeModeState` handle this? :)

mhils:
LGTM, but let's move the error check to includeModeState :)

web/src/js/ducks/modes/regular.ts
lups2000: sure :) let me change it

lups2000:
should be okay now :)

web/src/js/ducks/modes/wireguard.ts
```diff
@@ -21,21 +22,23 @@ export const initialState: WireguardState = {
...
+    (updateModeFunc = updateMode) =>
+    async (dispatch) => {
         dispatch({ type: MODE_WIREGUARD_TOGGLE });
 
-        const result = await dispatch(updateMode());
+        const result = await dispatch(updateModeFunc());
 
         if (!result.success) {
-            //TODO: handle error
+            if (result.error.includes("wireguard")) {
```
mhils, line 37: Is this condition still necessary?

web/src/js/ducks/modes/wireguard.ts
```diff
@@ -21,21 +22,23 @@ export const initialState: WireguardState = {
...
 export const getMode = (modes) => {
-    const wireguardMode = modes.wireguard;
+    const wireguardMode: WireguardState = modes.wireguard;
     return includeModeState("wireguard", wireguardMode);
 };
 
-export const toggleWireguard = () => {
-    return async (dispatch) => {
+export const toggleWireguard =
+    (updateModeFunc = updateMode) =>
```
mhils, line 30: Is there a reason why we don't mock the fetch itself? :)

lups2000: what do you mean?

web/src/js/ducks/modes/wireguard.ts
lups2000: nope, right! 👍

web/src/js/ducks/modes/wireguard.ts
mhils: Make toggleWireguard not accept any arguments, and in tests we mock the fetch instead of updateMode.

lups2000: it's so annoying that with `fetchMock` I cannot simulate properly the error and have the `state.error`updated

PR #6999, opened 2024-07-05
Mitmweb: Add Capture Tab

mhils:
#### Checklist

 - [ ] Changelog entry (@lups2000)
 - [ ] Take a look at urlState to figure out if we need to address things before merging into main. (#7000)


web/src/js/__tests__/components/Header/__snapshots__/CaptureMenuSpec.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`CaptureMenu Component should render correctly 1`] = `null`;
```
mhils, line 3: Surprised there is nothing rendered here?

lups2000: @mhils probably that's why: 
```
export default function CaptureMenu() {
    //empty component since we don't have an header menu for capture, but we just display the title
    return <></>;
}
```

mhils: Addressed in #7000. :)

PR #7000, opened 2024-07-05
Web: Switch to flow list upon arrival of the first flow.

mhils:
Once we have a flow, we should display that instead of configuration! In particular, `mitmweb -r file` should spawn with the flow list.

### DIFF
--- a/web/src/js/components/Header.tsx
+++ b/web/src/js/components/Header.tsx
@@ -28,12 +28,22 @@ export default function Header() {
             shallowEqual,
         ),
         [wasFlowSelected, setWasFlowSelected] = useState(false);
+    const hasFlows = useAppSelector(state => state.flows.list.length > 0),
+        isInitialTab = useAppSelector(state => state.ui.tabs.isInitial);
 
     let entries: Tab[] = [Tab.Capture, Tab.FlowList, Tab.Options];
     if (selectedFlows.length > 0) {
         entries.push(Tab.Flow);
     }
 
+    // Switch to "Flow List" when the first flow appears.
+    useEffect(() => {
+        if(hasFlows && isInitialTab) {
+            dispatch(setCurrent(Tab.FlowList));
+        }
+    }, [hasFlows]);
+
+    // Switch to "Flow" tab if we just selected a new flow.
     useEffect(() => {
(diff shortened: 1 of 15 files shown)


web/src/js/ducks/ui/tabs.ts
```diff
@@ -0,0 +1,26 @@
+import { createSlice } from "@reduxjs/toolkit";
+
+export enum Tab {
+    Capture,
+    FlowList,
+    Options,
+    Flow,
+}
+
+const tabsSlice = createSlice({
```
mhils, line 10: This here is quite nice really! We should consider using it for more redux stuff going forward.

lups2000: oh that's really nice, so we shouldn't specify anymore stuff like this:
`const TOGGLE_MODE = "TOGGLE_MODE"`, right?

web/src/js/components/Header.tsx
```diff
@@ -4,73 +4,87 @@ import FileMenu from "./Header/FileMenu";
...
                 ))}
                 <HideInStatic>
                     <ConnectionIndicator />
                 </HideInStatic>
             </nav>
-            <div
-                className={classnames({
-                    "empty-header": ActiveMenu === CaptureMenu,
-                })}
-            >
```
lups2000, line 73: you're missing the css class in case of `CaptureMenu` which is different from the other tabs!

lups2000 (changes requested):
The code looks good to me! There is just a small UI problem with `CaptureMenu`
We shouldn't have that space :)
Here the screenshot:

Screenshot 2024-07-05 at 20 44 11

web/src/js/ducks/ui/tabs.ts
mhils: yup, much nicer. The old style has benefits (it is more obvious what is actually happening), but I like it nonetheless!

web/src/js/components/Header.tsx
mhils: Missed the updated CSS when staging. Fixed now. :)

lups2000 (approved):
🍰

PR #7001, opened 2024-07-05
Allow `typing.Sequence[str]` to be an editable option

errorxyz:
#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 * Add first MVP new Capture Tab in mitmweb
   ([#6999](https://github.com/mitmproxy/mitmproxy/pull/6999), @lups2000)
 * Allow `typing.Sequence[str]` to be an editable option
+  ([#7001](https://github.com/mitmproxy/mitmproxy/pull/7001), @errorxyz)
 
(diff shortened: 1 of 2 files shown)


mhils:
I thought you were running into issues, but from the diff it reads like typing.Sequence would work? Any idea why it doesn't?

errorxyz:
> I thought you were running into issues, but from the diff it reads like typing.Sequence would work? Any idea why it doesn't?

It does accept `typing.Sequence[str]` as a typespec but if we want to edit the option we get a NotImplementedError because of this line:
https://github.com/mitmproxy/mitmproxy/blob/5b2659acb927b59bd010c32823a82cbe70dca448/mitmproxy/tools/console/options.py#L210
i.e it only accept `collections.abc.Sequence[str]` types for editing. We could allow `typing.Sequence[str]` to work as well if that's what we want.
```
elif foc.opt.typespec in (Sequence[str], typing.Sequence[str])
```

mhils:
If that's all that's necessary, I'd say let's allow typing.Sequence as well. There's a bunch of existing code with that. :)

mhils (approved):

PR #7008, opened 2024-07-07
Adopt ESLint

mhils:
eslint is happy now!


.github/workflows/autofix.yml
```diff
@@ -27,10 +27,11 @@ jobs:
...
           node-version-file: .github/node-version.txt
-      - name: Run prettier
-        run: |
-          npm ci
-          npm run prettier
+      - run: npm ci
+        working-directory: web
+      - run: npx eslint --fix .
+        working-directory: web
+      - run: npx prettier --write .
```
Prinzhorn, line 34: I personally would use `npm run prettier` (and add a command for eslint as well) so that this config does not need to know any details (arguments etc.).

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@
   ([#6930](https://github.com/mitmproxy/mitmproxy/pull/6930), @errorxyz)
 * Parse compressed domain names in ResourceRecord data
   ([#6954](https://github.com/mitmproxy/mitmproxy/pull/6954), @errorxyz)
+* Fix a bug where mitmweb's flow list would not stay at the bottom.
+  ([#7008](https://github.com/mitmproxy/mitmproxy/pull/7008), @mhils)
```
Prinzhorn, line 23: #7008 is this PR

harold

web/eslint.config.mjs
```diff
@@ -0,0 +1,35 @@
+import globals from "globals";
+import pluginJs from "@eslint/js";
+import tseslint from "typescript-eslint";
+import pluginReactConfig from "eslint-plugin-react/configs/recommended.js";
+
+export default [
+    { files: ["**/*.{js,mjs,cjs,ts,jsx,tsx}"] },
```
Prinzhorn, line 7: Weirdly enough this does not stop `.snap` from being touched? I don't think we want to lint auto-generated snapshots, or do we?

web/src/js/__tests__/components/FlowTableSpec.tsx
```diff
@@ -1,57 +1,27 @@
...
-import { Provider } from "react-redux";
+import FlowTable, { PureFlowTable } from "../../components/FlowTable";
+
+import { render } from "../test-utils";
+import { select } from "../../ducks/flows";
 
 window.addEventListener = jest.fn();
 
 describe("FlowTable Component", () => {
-    let selectFn = jest.fn(),
```
Prinzhorn, line 11: What exactly happened to those tests? They look completely different now?

web/src/js/__tests__/components/Header/MenuToggleSpec.tsx
```diff
@@ -5,19 +5,15 @@ import {
     MenuToggle,
     OptionsToggle,
 } from "../../../components/Header/MenuToggle";
-import { Provider } from "react-redux";
-import { TStore } from "../../ducks/tutils";
-import * as optionsEditorActions from "../../../ducks/ui/optionsEditor";
 import { fireEvent, render, screen, waitFor } from "../../test-utils";
-import { fetchApi, runCommand } from "../../../utils";
```
Prinzhorn, line 12: Getting rid of unused imports

so_good

web/src/js/__tests__/ducks/utils/storeSpec.tsx
```diff
@@ -13,9 +13,9 @@ describe("store reducer", () => {
     });
 
     it("should handle add action", () => {
-        let a = { id: "1" },
-            b = { id: "9" },
-            state = reduce(undefined, {});
+        const a = { id: "1" },
+            b = { id: "9" };
```
Prinzhorn, line 17: Ew, what happened to the indentation? Is that really what prettier does? I personally forbid multi-declarations and would just have two `const` lines.

Prinzhorn (changes requested):
image

web/eslint.config.mjs
```diff
@@ -0,0 +1,35 @@
...
+    { languageOptions: { parserOptions: { ecmaFeatures: { jsx: true } } } },
+    { languageOptions: { globals: globals.browser } },
+    pluginJs.configs.recommended,
+    ...tseslint.configs.recommended,
+    pluginReactConfig,
+    {
+        rules: {
+            "@typescript-eslint/no-explicit-any": "off",
+            "@typescript-eslint/no-unused-vars": "off",
```
Prinzhorn, line 16: What's the problem with no-unused-vars?

web/src/js/__tests__/components/Header/MenuToggleSpec.tsx
lups2000: my bad 🙈

web/src/js/__tests__/ducks/utils/storeSpec.tsx
lups2000: I think so yes

CHANGELOG.md
mhils: Yes, this was fixed in this PR. 😛

web/src/js/__tests__/components/FlowTableSpec.tsx
mhils: They were utterly wrong, now they are utterly right. 

image

web/eslint.config.mjs
mhils: Huh - snap files are not linted for me?

web/eslint.config.mjs
mhils: This produces a lot of false-positives for arguments by default. I've found a middle ground now,

web/eslint.config.mjs
Prinzhorn: Oh, never mind then. I though the changes in the snap files in this PR were caused by eslint. But I assume some other changed caused that?

mhils: Yes, those are test refactors / upgrades.

lups2000 (approved):

lups2000:
@Prinzhorn all good now?

Prinzhorn:
Please explain how `no-unused-vars` causes false positives? They are unused and can be removed?

https://github.com/Prinzhorn/mitmproxy/commit/376a70a23b89081238ba9d6b2a36d606e8333678

This makes `npm run test` green

Prinzhorn:
`no-case-declarations` -> you make a block for proper scoping

image

mhils:
> Please explain how `no-unused-vars` causes false positives? They are unused and can be removed?

https://github.com/Prinzhorn/mitmproxy/commit/376a70a23b89081238ba9d6b2a36d606e8333678#diff-6c1ddec6d8e99e96e4006721adc06df292d9aba4fcf0642a9a0afc17a2cc4109L37 for instance is a legitimate type assertion that makes sure the switch statement is exhaustive. https://github.com/Prinzhorn/mitmproxy/commit/376a70a23b89081238ba9d6b2a36d606e8333678#diff-74a34e766de69102a7ce29252a4b1277eaeef1e27ffe7e68e8a8c00efa86cb6bR281 is another example: it's useful to me when the default implementation documents the "normal" expected arguments.

> no-case-declarations -> you make a block for proper scoping

Thanks, this looks sane. My editor proposed some other nonesense. Will update. 👍

Prinzhorn:
> for instance is a legitimate type assertion that makes sure the switch statement is exhaustive.

Since `tsc` passes I figured this line doesn't do anything (I don't really understand what it's supposed to do, I don't use TypeScript, the line looks like gibberish to me).

> I think the current approach (_-prefixed is allowed) is a reasonable compromise.

Fair :+1:

Prinzhorn (approved):

mhils:
> the line looks like gibberish to me

```typescript
_exhaustiveCheck: never = connState;
```

If TypeScript would think that `connState` could be reachable, this line would error. Put differently, based on the previous case statements, the TypeScript compiler infers that the type of `connState` must be `never` here as all cases are covered. So it is fine with assigning.

This asserts that at this point the typescript compiler believes that `connState` cannot be reached.

Prinzhorn:
```diff
-_exhaustiveCheck: never = connState;
+connState satisfies never;
```

Prinzhorn:
This also works (return type + no default)

```tsx
import * as React from 'react';
import { ConnectionState } from '../../ducks/connection';
import { useAppSelector } from '../../ducks';
import type { ReactElement } from 'react';

export default React.memo(function ConnectionIndicator(): ReactElement {
    const connState = useAppSelector((state) => state.connection.state);
    const message = useAppSelector((state) => state.connection.message);

    switch (connState) {
        case ConnectionState.INIT:
            return <span className="connection-indicator init">connecting…</span>;
        case ConnectionState.FETCHING:
            return <span className="connection-indicator fetching">fetching data…</span>;
        case ConnectionState.ESTABLISHED:
            return <span className="connection-indicator established">connected</span>;
        case ConnectionState.ERROR:
            return (
                <span className="connection-indicator error" title={message}>
                    connection lost
                </span>
            );
        case ConnectionState.OFFLINE:
            return <span className="connection-indicator offline">offline</span>;
    }
});
```

If a case is missing this will cause

> Function lacks ending return statement and return type does not include 'undefined'.

mhils:
> ```diff
> -_exhaustiveCheck: never = connState;
> +connState satisfies never;
> ```
> 
> (requires a recent TS I think)

This failed when running tests somehow, but https://github.com/mitmproxy/mitmproxy/pull/7008#issuecomment-2217013030 is very nice. 👍

Prinzhorn:
Feel free to chose whatever version, I'm not the one maintaining mitmweb :smile:

Prinzhorn:
> This failed when running tests somehow, but [#7008 (comment)](https://github.com/mitmproxy/mitmproxy/pull/7008#issuecomment-2217013030) is very nice. 👍

Probably because we're on TypeScript `4.3.5` (Jul 1, 2021 :older_adult: )

PR #7011, opened 2024-07-09
Update Web Dependencies

mhils:
JS churn, so exciting!


Prinzhorn:
Maybe we want to enable dependabot on web as well? Or too much noise?

mhils:
I'm afraid that there could indeed be quite some noise, but I'd be open to trying it out. :)

lups2000 (commented):
I saw we're using this new `act()` function for dispatch events. Is there a reason why for example here (**ducks/localSpec.tsx**) we're not using it ? 
```
it("should handle error when toggling local", async () => {
        fetchMock.mockReject(new Error("invalid spec"));
        const store = TStore();

        await store.dispatch(toggleLocal());

        expect(fetchMock).toHaveBeenCalled();
        expect(store.getState().modes.local.error).toBe("invalid spec");
    });
```

mhils:
I added `act` at all the locations where jest was giving me a warning. For the example you posted I don't think we render any React components, so it's not necessary here. :)

PR #7012, opened 2024-07-09
Feature/single reverse mode

lups2000:
#### Description

In this MR, I'm adding the reverse mode to the new **Capture** page. So far, just one reverse server is supported.

Screenshot 2024-07-09 at 20 26 52
Screenshot 2024-07-09 at 20 27 29

#### Checklist


web/src/js/components/Modes/Reverse.tsx
```diff
@@ -0,0 +1,105 @@
...
+        "http",
+        "https",
+        "dns",
+        "http3",
+        "quic",
+        "tcp",
+        "tls",
+        "udp",
+        "dtls",
+    ];
```
mhils, line 29: It's super easy for this to go out of sync with the backend: https://github.com/mitmproxy/mitmproxy/blob/59fbbdc4b293cace2195c0250fcf4e7c0a808359/mitmproxy/proxy/mode_specs.py#L227.

My proposal would be to do the following:

1. Start a new PR that removes the mitmproxy version from `/state`: https://github.com/mitmproxy/mitmproxy/blob/59fbbdc4b293cace2195c0250fcf4e7c0a808359/mitmproxy/tools/web/app.py#L644
2. And then adds a backend_consts.py (open to better naming suggestions) script to https://github.com/mitmproxy/mitmproxy/tree/main/web/gen that writes the version into a TypeScript file. This "hardcodes" the mitmproxy version into the frontend so that it's available from the start.

Once this is in, we can (in this PR) extend backend_consts.py to also provide the list of supported protocols to the front end. Does that make sense?

web/src/js/ducks/modes/reverse.ts
```diff
@@ -0,0 +1,140 @@
...
+                let protocol,
+                    host = "";
+                if (isActive) {
+                    [protocol, host] = currentModeConfig.data.split("://");
+                }
+                return {
+                    ...state,
+                    active: isActive,
+                    protocol: isActive ? protocol : state.protocol,
+                    host: isActive ? host : state.host,
```
mhils, line 118: Nit: Let's call this `destination` to make it clear that it's not only the hostname, but also potentially port information.

web/src/js/ducks/modes/utils.ts
```diff
@@ -61,9 +63,10 @@ export const parseMode = (spec: string) => {
         listenAt = "";
     }
 
-    const [mode, data] = head.split(":");
-    let host = "",
-        port: string | number = "";
+    let [mode, ...rest] = head.split(":"); // trick to split on the first occurence of ":"
+    let data = rest.join(":")
```
mhils: Time for a `partition` helper next to `rpartition`? 😉

web/src/js/components/Modes/Reverse.tsx
lups2000: okay I've understood what I should do! Now, I still to figure out how since I'm not so experienced with python. But, I'll find a way and, in case, I'll ask you :)

lups2000:
I will leave this PR in stand-by then

web/src/js/components/Modes/Reverse.tsx
mhils: If you take https://github.com/mitmproxy/mitmproxy/blob/main/web/gen/state_js.py as a template, this should be super straightforward. Otherwise you can also ask @errorxyz when he's back. :)

lups2000: @mhils to generate the Typescript file and test the behaviour (for instance this: https://github.com/mitmproxy/mitmproxy/blob/main/web/gen/state_js.py), should i run the script manually? or is it somehow automated?

mhils: The autofix CI job should pick it up.

lups2000: mhh okay, right now the tests are failing and I'm getting this: `FAILED test/mitmproxy/tools/web/test_app.py::test_generated_files[filename2] - FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/mitmproxy/mitmproxy/web/gen/../src/js/backends/consts.ts'` when running `tox -e py` since of course the file doesn't exists. If I open a PR, the autofix will run and we should solve those errors automatically, right?

lups2000: move discussions here: #7016

lups2000: okay @mhils , I'm now using the protocols that I'm generating from the backend!

web/src/js/ducks/modes/reverse.ts
```diff
@@ -0,0 +1,140 @@
...
+import type { ModesState } from "../modes";
+
+export const MODE_REVERSE_TOGGLE = "MODE_REVERSE_TOGGLE";
+export const MODE_REVERSE_SET_LISTEN_CONFIG = "MODE_REVERSE_SET_LISTEN_CONFIG";
+export const MODE_REVERSE_SET_DESTINATION = "MODE_REVERSE_SET_DESTINATION";
+export const MODE_REVERSE_SET_PROTOCOL = "MODE_REVERSE_SET_PROTOCOL";
+export const MODE_REVERSE_ERROR = "MODE_REVERSE_ERROR";
+
+export interface ReverseState extends ModeState {
+    protocol?: string;
```
mhils, line 20: This can be typed more strongly. :)

web/src/js/ducks/modes/reverse.ts
```diff
@@ -0,0 +1,140 @@
...
+    destination?: string;
+}
+
+export const initialState: ReverseState = {
+    active: false,
+    protocol: "",
```
mhils, line 26: default to ReverseProxyMode.HTTPS? :)

web/src/js/utils.ts
```diff
@@ -191,3 +191,13 @@ export function rpartition(str: string, sep: string): [string, string] {
     const after = str.slice(lastIndex + sep.length);
     return [before, after];
 }
+
+export function lpartitionFirstOcc(str: string, sep: string): [string, string] {
```
mhils, line 195:
```suggestion
/** A JS equivalent of Python's https://docs.python.org/3/library/stdtypes.html#str.partition */
export function partition(str: string, sep: string): [string, string] {
```

mhils (commented):
Looks all good now, only some small comments + tests! :)

web/src/js/ducks/modes/reverse.ts
lups2000: you mean that the protocol should be one of these: `"https","http", "dns" etc...`

web/src/js/backends/consts.ts
```diff
@@ -1 +1,14 @@
 /** Auto-generated by web/gen/backend_consts.py */
+export const BACKEND_CONSTS = {
+    protocols: [
```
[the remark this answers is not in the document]
lups2000: okay alright! :)

mhils (approved):

PR #7016, opened 2024-07-10
Sync consts backend-frontend

lups2000:
#### Description

In this PR, I'm deleting the version from the state on the backend side and I created a new script to generate the TS file on the frontend. Still WIP

#### Checklist


web/gen/backend_consts.py
```diff
@@ -0,0 +1,31 @@
...
+
+async def make() -> str:
+    data = {
+        "version": version.VERSION,
+    }
+
+    # language=TypeScript
+    content = (
+        "/** Auto-generated by web/gen/backend_consts.py */\n"
+        "export const BACKEND_CONSTS = %s\n"
```
mhils, line 23: Why not just `export const VERSION = ...`? :)

lups2000: because we'll have also other constants. So, I was thinking we could access them like this: 
`BACKEND_CONSTS.version` or `BACKEND_CONSTS.protocols`

test/mitmproxy/test_version.py
```diff
@@ -17,20 +17,18 @@ def test_version(capsys):
 
 
 def test_get_version():
-    version.VERSION = "3.0.0rc2"
-
```
lups2000, line 21: I removed this hard coded version since when testing my generated TS file for the BACKEND_CONSTS there was a mismatch between `"3.0.0rc2"` and `"11.0.0 dev"`

mhils (approved):

lups2000:
thanks for having handled that :)

PR #7029, opened 2024-07-18
Feature/multiple reverse servers

lups2000:
#### Description

In this PR, I'm adding a functionality to handle multiple reverse servers in mitmweb. I'm opening this draft so you can see the current progress :) Now, I'll proceed with the approach we've discussed in the meeting together!

- [x] Toggling single reverse servers (using index) --> properly set in the backend
- [x] Update servers based on the state of the backend

Here some screenshots:

Screenshot 2024-07-18 at 13 38 22
Screenshot 2024-07-18 at 13 38 13

#### Checklist


lups2000:
Hi @mhils, I'm finally back :) 
The multiple reverse modes are "almost" working fine, except for one "small" bug. This is the current status:

https://github.com/user-attachments/assets/c7d994d8-cb27-4dbb-b99e-f1d19f911df2

As you can see, it works correctly when we set the modes through the new Capture Tab page. However, if we change the mode using the modal, it introduces intermediate results that are feasible but not necessarily the intended one.

For example, when changing from `reverse:https://google.com@localhost:8082` to `reverse:https://google.com@localhost:8085`, it temporarily passes through `reverse:https://google.com@localhost:808`, which results in this configuration being added.

lups2000:
what do you think about this new approach? I removed the map. The problem with the intermediate values is still persistent but at least the config are inactive

mhils:
Yes, this looks very nice now! 🚀 

I've slightly simplified things and unified naming between frontend and backend. Todo here:

- [ ] We need an unintrusive way to remove a row as well.
- [ ] Fix tests

There are still some UX nits we really should fix (e.g. select box should just be a regular HTML element that doesn't resize on selection), but let's keep that out of this PR to keep things simple. I've already sneaked in the renaming, which is bad. :)

mhils (approved):

PR #7037, opened 2024-07-24
Add experimental option for HTTP/3 in transparent mode

mhils:
With these changes I can get (static-curl, macOS, local mode) to work. 🎉 

Does that help with any of the other entries on our matrix?


test/mitmproxy/addons/test_next_layer.py
```diff
@@ -96,6 +99,8 @@
 
 http_connect = b"CONNECT example.com:443 HTTP/1.1\r\nHost: example.com:443\r\n\r\n"
 
+http2_preface = b"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n"
+
```
errorxyz, line 103: Are we using this anywhere?

errorxyz (approved):
Works on windows local mode as well🎉

errorxyz:
refs #7025

PR #7039, opened 2024-07-24
web: add index column

mhils:
#### Description

See #6976 for context.

#### Checklist

### DIFF
--- a/web/src/js/flow/utils.ts
+++ b/web/src/js/flow/utils.ts
@@ -287,7 +287,7 @@ export const getVersion = (flow: Flow): string => {
 export const sortFunctions = {
     tls: (flow: Flow) => flow.type === "http" && flow.request.scheme,
     icon: getIcon,
-    index: () => 0,  // this is broken right now - ideally we switch to uuid7s on the backend and use that.
+    index: () => 0, // this is broken right now - ideally we switch to uuid7s on the backend and use that.
     path: mainPath,
     method: getMethod,
     version: getVersion,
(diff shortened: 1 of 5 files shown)


web/src/js/flow/utils.ts
```diff
@@ -287,6 +287,7 @@ export const getVersion = (flow: Flow): string => {
 export const sortFunctions = {
     tls: (flow: Flow) => flow.type === "http" && flow.request.scheme,
     icon: getIcon,
+    index: () => 0, // this is broken right now - ideally we switch to uuid7s on the backend and use that.
```
lups2000, line 290: what do you mean? why?

lups2000, line 290: what do you mean? why?

lups2000 (commented):
I tested it and it worked as expected :) Just a clarification

web/src/js/flow/utils.ts
mhils: Currently you can't sort by the index column, it's just a no-op. The problem is that we only have the flow as input here and not the index. Fixable, but not worth to complicate the API here for an off-by-default column. An easy fix would be to use uuid7s instead of uuid4s for `flow.id` and then sort by that, but that's future work. :)

lups2000: ahh I see, I forgot to try that 😅 okay! All good then :)

lups2000 (approved):
🍰

PR #7067, opened 2024-08-01
Handle fragmented QUIC client hello

errorxyz:
#### Description
Handles fragmented QUIC client hello
refs #7025 

#### Checklist

### DIFF
--- a/mitmproxy/proxy/layers/quic.py
+++ b/mitmproxy/proxy/layers/quic.py
@@ -1239,7 +1239,7 @@ class ClientQuicLayer(QuicLayer):
             parent_layer._handle_event = replacement_layer._handle_event  # type: ignore
             yield from parent_layer.handle_event(events.Start())
             yield from parent_layer.handle_event(
-                events.DataReceived(self.context.client, self.recv_buffer)
+                events.DataReceived(self.context.client, bytes(self.recv_buffer))
             )
             self.recv_buffer.clear()
             return True, None
--- a/mitmproxy/proxy/layers/quic.py
+++ b/mitmproxy/proxy/layers/quic.py
@@ -381,7 +381,13 @@ def quic_parse_client_hello(data: bytes) -> Optional[ClientHello]:
             raise ValueError("Invalid ClientHello data.") from e
     except QuicConnectionError as e:
         raise ValueError(e.reason_phrase) from e
-    return None
+
+    # Fragmented Clienthellos seem to produce valid replies whereas invalid Clienthellos don't
+    # We delay sending these replies until we receive the entire client hello
+    if quic.datagrams_to_send(now=0.1):
+        return None
+
+    raise ValueError("No ClientHello returned.")
 
 
 class QuicStreamNextLayer(layer.NextLayer):
--- a/mitmproxy/proxy/layers/quic.py
+++ b/mitmproxy/proxy/layers/quic.py
@@ -1269,7 +1269,6 @@ class ClientQuicLayer(QuicLayer):
         self.recv_buffer.clear()
         return ret
 
-
     def start_server_tls(self) -> layer.CommandGenerator[str | None]:
         if not self.server_tls_available:
             return f"No server QUIC available."
--- a/mitmproxy/proxy/layers/quic.py
+++ b/mitmproxy/proxy/layers/quic.py
@@ -20,10 +20,10 @@ from aioquic.quic.connection import QuicConnectionState
 from aioquic.quic.connection import QuicErrorCode
 from aioquic.quic.connection import stream_is_client_initiated
 from aioquic.quic.connection import stream_is_unidirectional
+from aioquic.quic.logger import QuicLogger
 from aioquic.quic.packet import encode_quic_version_negotiation
 from aioquic.quic.packet import PACKET_TYPE_INITIAL
 from aioquic.quic.packet import pull_quic_header
-from aioquic.quic.logger import QuicLogger
 from aioquic.tls import CipherSuite
 from aioquic.tls import HandshakeType
 from cryptography import x509
@@ -349,7 +349,7 @@ def quic_parse_client_hello(msgs: list[bytes]) -> Optional[ClientHello]:
         - A ValueError, if the passed ClientHello is invalid
     """
 
-    buffer = QuicBuffer(data=b''.join(msgs))
+    buffer = QuicBuffer(data=b"".join(msgs))
     header = pull_quic_header(buffer, 8)
     if header.packet_type != PACKET_TYPE_INITIAL:
         raise ValueError("Packet is not initial one.")
@@ -399,7 +399,7 @@ def quic_parse_client_hello(msgs: list[bytes]) -> Optional[ClientHello]:
 
     assert quic._quic_logger
     for event in quic._quic_logger._events:
-        if event['name'] == 'transport:packet_dropped':
+        if event["name"] == "transport:packet_dropped":
             raise ValueError(f"Invalid ClientHello packet: {event['data']['trigger']}")
 
     return None
@@ -1227,7 +1227,7 @@ class ClientQuicLayer(QuicLayer):
         try:
             client_hello = quic_parse_client_hello(self.msgs)
         except ValueError as e:
-            msgs = b'\n'.join(self.msgs)
+            msgs = b"\n".join(self.msgs)
             dbg = f"Cannot parse ClientHello: {str(e)} ({msgs.hex()})"
             self.msgs.clear()
             return False, dbg
@@ -1263,7 +1263,9 @@ class ClientQuicLayer(QuicLayer):
             parent_layer._handle_event = replacement_layer._handle_event  # type: ignore
             yield from parent_layer.handle_event(events.Start())
             for msg in self.msgs:
-                yield from parent_layer.handle_event(events.DataReceived(self.context.client, msg))
+                yield from parent_layer.handle_event(
+                    events.DataReceived(self.context.client, msg)
+                )
             self.msgs.clear()
             return True, None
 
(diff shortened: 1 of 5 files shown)


errorxyz:
Chrome doesn't seem to trust the certificate that we provide during the QUIC handshake:
`Client QUIC handshake failed. 199:TLS handshake failure (ENCRYPTION_HANDSHAKE) 46: certificate unknown`

Workaround:
`.\chrome.exe --ignore-certificate-errors-spki-list="<cert fingerprint>" --origin-to-force-quic-on=localhost:443`
We can generate cert fingerprint using:
`openssl x509 -pubkey -noout -in <cert path> | openssl rsa -pubin -outform der | openssl dgst -sha256 -binary | base64`

mhils:
@errorxyz: this is for reverse proxy mode _and_ the cert is installed in the system?

errorxyz:
> @errorxyz: this is for reverse proxy mode _and_ the cert is installed in the system?

Yes. Searching for the error a bit online, it seems it is an issue with quiche?

errorxyz:
Certificate issue fixed in https://github.com/mitmproxy/mitmproxy/pull/7073

errorxyz:
@mhils chrome still does not trust the self-signed certificate i.e accessing our http3 reverse proxy on localhost using chrome doesn't work
error: `Client QUIC handshake failed. 199:TLS handshake failure (ENCRYPTION_HANDSHAKE) 46: certificate unknown`
However, curl works with the mitmproxy cert specified using `--cacert`

check this out: https://github.com/GoogleChrome/samples/blob/8223e1b7f619bde037cf010dd9a2d899556a2d89/quictransport/quic_transport_server.py#L44-L75
Following these instructions fixes the issue

Using a trusted cert for eg. from letsencrypt on our test server works as intended

mitmproxy/proxy/layers/quic.py
```diff
@@ -380,6 +381,12 @@ def initialize_replacement(peer_cid: bytes) -> None:
             raise ValueError("Invalid ClientHello data.") from e
     except QuicConnectionError as e:
         raise ValueError(e.reason_phrase) from e
+
+    # Fragmented Clienthellos seem to produce valid replies whereas invalid Clienthellos don't
+    # We delay sending these replies until we receive the entire client hello
+    if quic.datagrams_to_send(now=0.1):
+        return None
```
errorxyz, line 388: One case that we miss here is that if the second fragment happens to be invalid i.e. we receive a valid first fragment and a corrupt second fragment (and the entire Clienthello is sent across two fragments). Since the entire clienthello is invalid, but the first fragment is valid, we don't get valid Clienthello but we receive a non empty value from `datagrams_to_send` and we end up waiting for more fragments instead of raising an error.

One solution to this problem is that we could make `recv_buffer` a `list[bytes]` and add new fragments as a whole to the start of the buffer (and then convert to `bytes` while passing to `quic_parse_client_hello`). Doing so, we raise an error as the first packet being read from the buffer is invalid (for the above case) (If it were valid we will return a Clienthello as we can also parse out-of-order fragments). But would this added complexity be a good tradeoff for a seemingly rare edgecase?

mitmproxy/proxy/layers/quic.py
```diff
@@ -1121,6 +1123,7 @@ class ClientQuicLayer(QuicLayer):
 
     server_tls_available: bool
     """Indicates whether the parent layer is a ServerQuicLayer."""
+    recv_buffer: bytearray
```
mhils, line 1126: Shouldn't this much rather be `list[bytes]` and we invoke `quic.receive_datagram` repeatedly? For UDP I think it's important that we preserve message boundaries. aioquic happens to tolerate this, but that's not necessarily true for other QUIC implementations.

(fundamentally this also needs a fix in the NextLayer logic - but maybe not in this PR)

mitmproxy/proxy/layers/quic.py
```diff
@@ -331,7 +332,7 @@ class QuicClientHello(Exception):
     data: bytes
 
 
-def quic_parse_client_hello(data: bytes) -> ClientHello:
+def quic_parse_client_hello(data: bytes) -> Optional[ClientHello]:
```
mhils, line 335: This needs a docstring like `tls.parse_client_hello` now I think.

test/mitmproxy/proxy/layers/test_quic.py
```diff
@@ -1241,20 +1326,6 @@ def test_non_init_clienthello(self, tctx: context.Context):
         )
         assert client_layer.tunnel_state == tls.tunnel.TunnelState.ESTABLISHING
 
-    def test_invalid_clienthello(self, tctx: context.Context):
```
mhils, line 1244: Can we make up a new tests with a ClientHello that's actually invalid somehow?

mhils (commented):
Very, very cool debugging work! 🚀

mitmproxy/proxy/layers/quic.py
```diff
@@ -1257,7 +1290,13 @@ def receive_handshake_data(
             return False, "connection closed early"
 
         # send the client hello to aioquic
-        return (yield from super().receive_handshake_data(data))
+        ret: tuple[bool, str | None]
+        for msg in self.msgs:
+            ret = yield from super().receive_handshake_data(msg)
+            if not ret[0]:
+                break
```
mhils, line 1297: Why `break` here? Do we not want to send the remaining packets to aioquic?

(if there is a good reason for this: add an explanatory comment)

mitmproxy/proxy/layers/quic.py
```diff
@@ -1121,6 +1143,7 @@ class ClientQuicLayer(QuicLayer):
 
     server_tls_available: bool
     """Indicates whether the parent layer is a ServerQuicLayer."""
+    msgs: list[bytes]
```
mhils, line 1146:
```suggestion
    handshake_datagram_buf: list[bytes]
```

(or anything else that's more descriptive than `msgs`. :)

mhils (commented):
🚀

mitmproxy/proxy/layers/quic.py
errorxyz: > Do we not want to send the remaining packets to aioquic?

Yes, we want to close the connection if for any packet we receive `done=False` from `super().receive_handshake_data`. There's no QUIC-specific reason behind this but I don't see any other way to handle the case where say the second packet fails and we still have a third packet to be sent.

mitmproxy/proxy/layers/quic.py
```diff
@@ -1257,7 +1290,13 @@ def receive_handshake_data(
             return False, "connection closed early"
 
         # send the client hello to aioquic
-        return (yield from super().receive_handshake_data(data))
+        ret: tuple[bool, str | None]
+        for msg in self.msgs:
+            ret = yield from super().receive_handshake_data(msg)
```
mhils, line 1295:
```suggestion
            (done, err_) = yield from super().receive_handshake_data(msg)
```

mitmproxy/proxy/layers/quic.py
mhils: Now I'm more confused than before - if we receive out-of-order packets, isn't `done=False` expected after the first packet?

errorxyz: Oh yes makes sense, I think I've interpreted things incorrectly. But what's weirder is that HTTP3 (reverse proxy) works irrespective of the `break` statement. Digging deeper, it seems we always get `done=False` here because of the `quic_events.ProtocolNegotiated` event raised in `QuicLayer.receive_handshake_data` and we can't return `True, None` for this event yet since the handshake isn't completed yet.

mhils: > But what's weird is that HTTP3 (reverse proxy) works irrespective of the break statement.

(this could also be due to retransmissions)

> it seems we always get `done=False` here because we raise `quic_events.ProtocolNegotiated` event in `QuicLayer.receive_handshake_data`

From a very cursory read, we swallow that immediately and continue with the. `while event := self.quic.next_event():` loop:
https://github.com/mitmproxy/mitmproxy/blob/0f42f00baa2925964213ed5c28aac1655abbe0db/mitmproxy/proxy/layers/quic.py#L954-L963

Without being fully in the weeds here: Intuitively we can maybe just do something like this?

```python
for msg in self.msgs:
    self.quic.receive_datagram(data, self.conn.peername, now=self._time())

return (yield from super().receive_handshake_data(b""))
```

errorxyz: > Without being fully in the weeds here: Intuitively we can maybe just do something like this?

Yup, this works as well.

errorxyz:
We still get an error related to not handling `StopSendingReceived` event and a stream related error, but I'm confident that these changes are unrelated to those bugs.

mhils (approved):

mhils:
Changelog entry and this is good to go. 

(Focus on user-facing changes 😉 )

mitmproxy/proxy/layers/quic.py
```diff
@@ -372,15 +390,29 @@ def initialize_replacement(peer_cid: bytes) -> None:
...
+    assert isinstance(quic_logger, QuicLogger)
+    traces = quic_logger.to_dict().get("traces")
+    assert isinstance(traces, list)
+    for trace in traces:
+        quic_events = trace.get("events")
+        for event in quic_events:
+            if event["name"] == "transport:packet_dropped":
+                raise ValueError(
+                    f"Invalid ClientHello packet: {event['data']['trigger']}"
+                )
```
errorxyz, line 413: Really don't like this part, but I don't think we have any other way around this.

mhils: `quic_parse_client_hello_from_datagrams` is a strong contender for the worst-function-in-mitmdump award, I'm glad we're strengthening its position!

PR #7073, opened 2024-08-02
Improve UX when users specify invalid certs

mhils:
If we don't do this ourselves, OpenSSL will greet users with the ever-fascinating 'no shared ciphers' error during the first handshake.

#### Checklist


mitmproxy/certs.py
```diff
@@ -502,11 +509,28 @@ def add_cert_file(
...
+            private_key = load_pem_private_key(raw, password=passphrase)
+        except ValueError as e:
+            private_key = self.default_privatekey
+            if cert.public_key() != private_key.public_key():
+                raise ValueError(
+                    f'Unable to find private key in "{path.absolute()}": {e}'
+                ) from e
+        else:
+            if cert.public_key() != private_key.public_key():
+                raise ValueError(
```
errorxyz, line 521: Do we need to handle this line in a similar fashion? https://github.com/mitmproxy/mitmproxy/blob/0ff08f975e52700c02ab590a9af8e0bd10729902/mitmproxy/certs.py#L419

test/mitmproxy/net/data/verificationcerts/generate.py
```diff
@@ -92,3 +92,22 @@ def mkcert(cert, subject, ip: bool):
         pem.write(key.read())
 
 shutil.copyfile("trusted-leaf.pem", "example.mitmproxy.org.pem")
+with (
```
errorxyz, line 95: Should we also test if the file has an intermediate cert but no leaf cert?

errorxyz (commented):
A few questions regarding the changes:

test/mitmproxy/net/data/verificationcerts/generate.py
mhils: Good call, added!

errorxyz (approved):
🎉

PR #7087, opened 2024-08-06
support full alpha tcp connection

SafaSafari:
#### Description

add a filter to non-tcp connections for full alphabetics tcp connection
this depends of first word in tcp payload
after split by whitespace and checking length of that

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 ## Unreleased: mitmproxy next
 
+- Support full alpha tcp connection.
+  ([#7087](https://github.com/mitmproxy/mitmproxy/pull/7087))
 - Improve the error message when users specify the `certs` option without a matching private key.
   ([#7073](https://github.com/mitmproxy/mitmproxy/pull/7073), @mhils)
 - Fix a bug where intermediate certificates would not be transmitted when using QUIC.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 ## Unreleased: mitmproxy next
 
-- Support full alpha tcp connection.
+- Tighten HTTP detection heuristic to better support custom TCP-based protocols.
   ([#7087](https://github.com/mitmproxy/mitmproxy/pull/7087))
 - Improve the error message when users specify the `certs` option without a matching private key.
   ([#7073](https://github.com/mitmproxy/mitmproxy/pull/7073), @mhils)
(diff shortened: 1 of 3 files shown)


SafaSafari:
In fact, this also applies to the SSH protocol.
However, the case I encountered was in reverse engineering an Android app that sent the information after Base64 encoding =)))

mitmproxy/addons/next_layer.py
```diff
@@ -184,6 +184,8 @@ def s(*layers):
         probably_no_http = (
             # the first three bytes should be the HTTP verb, so A-Za-z is expected.
             len(data_client) < 3
+            # limit first word of http request len ex. OPTIONS, CONNECT (full alpha tcp connection)
+            or not len(data_client.split(b" ")[0]) <= 7
```
mhils, line 188:
```suggestion
            # HTTP would require whitespace before the first newline
            # if we have neither whitespace nor a newline, it's also unlikely to be HTTP.
            or (data_client.find(b" ") >= data_client.find(b"\n"))
```

test/mitmproxy/addons/test_next_layer.py
```diff
@@ -94,6 +94,8 @@
 
 dns_query = bytes.fromhex("002a01000001000000000000076578616d706c6503636f6d0000010001")
 
+tcp_fullalpha = b"AAAAAAAAAAAAAAAAAAAAAA=="
```
mhils, line 97:
```suggestion
# Custom protocol with just base64-encoded messages
# https://github.com/mitmproxy/mitmproxy/pull/7087
custom_base64_proto = b"AAAAAAAAAAAAAAAAAAAAAA=="
```

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 
 ## Unreleased: mitmproxy next
 
+- Support full alpha tcp connection.
```
mhils, line 10:
```suggestion
- Tighten HTTP detection heuristic to better support custom TCP-based protocols.
```

mhils (commented):
Thanks! SSH we already special-cased lately. But I do agree that we can improve here. :) 
len <= 7 isn't a great heuristic because HTTP methods can be longer (https://www.iana.org/assignments/http-methods/http-methods.xhtml). Please take a look at the suggestions below and let me know if that's a viable alternative!

SafaSafari:
Yes, you're right. Decision-making based on fixed characteristics in HTTP requests is better.

mhils (approved):
Thanks! 🍰

PR #7093, opened 2024-08-09
Addon to update the ports in alt-svc header in reverse mode

errorxyz:
#### Description

Alt-Svc headers in HTTP responses advertise the host to use while migrating from HTTP1/2 to HTTP3 (amongst other information). This addon modifies the hosts specified in this header to our reverse proxy target.
refs #7025 

#### Checklist

### DIFF
--- a/mitmproxy/addons/update_alt_svc.py
+++ b/mitmproxy/addons/update_alt_svc.py
@@ -1,22 +1,30 @@
-from mitmproxy.http import HTTPFlow
 import re
+
 from mitmproxy import ctx
+from mitmproxy.http import HTTPFlow
 from mitmproxy.proxy import mode_specs
 
-ALT_SVC = 'alt-svc'
+ALT_SVC = "alt-svc"
 PORT_PATTERN = r'(?<=:)\d{1,5}(?=")'
 
 
 def update_alt_svc_port(data: str, port: int) -> str:
-    return re.sub(PORT_PATTERN, f'{port}', data)
+    return re.sub(PORT_PATTERN, f"{port}", data)
 
 
 class UpdateAltSvc:
     def load(self, loader):
-        loader.add_option("update_alt_svc", bool, True, "Update the ports in alt-svc header to the port that we are listening on in reverse mode")
+        loader.add_option(
+            "update_alt_svc",
+            bool,
+            True,
+            "Update the ports in alt-svc header to the port that we are listening on in reverse mode",
+        )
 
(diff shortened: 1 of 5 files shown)


errorxyz:
Notes for self:
Connecting to a remote reverse proxy works[^1] via chrome but not using firefox
Connecting to a locally hosted reverse proxy does not work in chrome and firefox
[^1]: works as in browsers attempt to connect to the updated port in the alt-svc header

mitmproxy/addons/update_alt_svc.py
```diff
@@ -0,0 +1,31 @@
...
+from mitmproxy import ctx
+from mitmproxy.http import HTTPFlow
+from mitmproxy.proxy import mode_specs
+
+ALT_SVC = "alt-svc"
+PORT_PATTERN = r'(?<=:)\d{1,5}(?=")'
+
+
+def update_alt_svc_port(data: str, port: int) -> str:
+    return re.sub(PORT_PATTERN, f"{port}", data)
```
mhils, line 12: I think we need to override the optional host override as well, no?

mitmproxy/addons/update_alt_svc.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class UpdateAltSvc:
+    def load(self, loader):
+        loader.add_option(
+            "update_alt_svc",
```
mhils, line 18: `keep_alt_svc_header` to match `keep_host_header`?

mhils (commented):
Looks good so far. 👍

errorxyz:
Do we want an option for the user to specify the host value that will be used to modify the `Alt-Svc` header?

errorxyz:
> Notes for self: Connecting to a remote reverse proxy works[1](#user-content-fn-1-9cee89b6c97433b5ee61df40af6e38df) via chrome but not using firefox[2](#user-content-fn-2-9cee89b6c97433b5ee61df40af6e38df) Connecting to a locally hosted reverse proxy does not work in chrome and firefox

I can swear that firefox randomly started working for a locally hosted reverse proxy after hours of unsuccessful debugging.

mhils:
> Do we want an option for the user to specify the host value that will be used to modify the `Alt-Svc` header?

Do we have a good use case for that? For esoteric use cases they can always turn the option on and write their own addon (or use modify_headers).

mhils (approved):

PR #7094, opened 2024-08-10
hide local mode on linux in mitmweb

lups2000:
#### Description

In this PR, we hide the local mode when mitmweb is running on Linux. I chose to hide the mode entirely rather than grey it out, as the message "Remaining modes are coming soon" is already present.

#### Checklist

### DIFF
--- a/mitmproxy/tools/web/app.py
+++ b/mitmproxy/tools/web/app.py
@@ -6,6 +6,7 @@ import json
 import logging
 import os.path
 import re
+import sys
 from collections.abc import Callable
 from collections.abc import Sequence
 from io import BytesIO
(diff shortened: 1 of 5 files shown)


mhils (commented):
This is confusing the OS of the browser with the OS of where mitmproxy runs on. I think we need to expose additional backend state here :)

lups2000:
> This is confusing the OS of the browser with the OS of where mitmproxy runs on. I think we need to expose additional backend state here :)

ahh I see, I thought it was much simpler 😄 we can discuss it then

lups2000:
@mhils better now?

mhils (approved):
Yes, great. Thanks! 🚀

Issue #7063, opened 2024-07-30
Tracking Issue: Capture Tab


mhils:
Mitmweb is getting a new capture tab. See https://github.com/mitmproxy/mitmproxy/discussions/6868 for the original discussion and https://github.com/mitmproxy/mitmproxy/issues/6955 for the MVP. This issue collects everything for the first version to be called feature complete.

- [x] Display current status for each mode. We want to integrate what's currently in [web/src/js/components/CaptureSetup.tsx](https://github.com/mitmproxy/mitmproxy/blob/main/web/src/js/components/CaptureSetup.tsx) and remove that page entirely. The CaptureSetup page can just refer to the new tab and not display anything itself.
- [x] Add ⚙️ button for an "advanced configuration" popover.
- [x] Move listen_host / listen_port into popover.
- [x] Make sure local mode is hidden/greyed out/... on unsupported platforms.
- [x] Support for multiple reverse proxies.
- [x] Application selector with process names and icons for local redirect mode.
- [x] Add remaining proxy modes in "Advanced" category
  - [x] SOCKS proxy mode
  - [x] DNS server mode
  - [x] Upstream proxy mode
  - [x] Transparent mode

Minor Stuff:
- [x] Improve dropdown when selecting protocols.
- [x] Clicking on the label should activate the mode.
- [x] Reverse Proxy Mode needs a popover.
- [x] Highlight recommended modes visually, e.g. like [this](https://github.com/user-attachments/assets/9f2d2416-f79d-44e3-9a7b-9f2824aca6c3).
- [x] On unsupported platforms, show a greyed out "Local Redirect Mode" header with "This mode is only supported on Windows and Linux" in the list. I was super confused where local redirect mode was when starting mitmweb from WSL, which made me realize that just hiding it is bad UX.
- [x] Make it possible to add arbitrary strings in local mode, not just executables from the list.
- [x] Add cache headers for the executable icons endpoint.
- [x] Enabling both regular and local redirect mode on Windows (on MacOS as well) currently gives an incorrect "Cannot spawn multiple servers on the same address: *:8080" error.

lups2000:
@mhils 🥳 🎉

PR #7100, opened 2024-08-14
Feature/socks and transparent modes

lups2000:
#### Description

In this PR, I've added the **Transparent** and **SOCKS** modes to the Advanced category. I was also considering whether it would be beneficial to implement a dropdown menu for the Advanced modes. This way, the Advanced options would be collapsed by default, and users could expand them by clicking a "+" icon. This could help save space for those who prefer to stick with the Recommended modes. What do you think?

Ref: #7063

https://github.com/user-attachments/assets/b71c8403-18d0-492e-93ca-4ecd83922c85

#### Checklist


web/src/js/components/Modes/Transparent.tsx
```diff
@@ -0,0 +1,61 @@
...
+            />
+        );
+    });
+
+    return (
+        <div>
+            <h4 className="mode-title">Transparent Proxy</h4>
+            <p className="mode-description">
+                A transparent proxy routes traffic at the network layer without
+                client configuration.
```
mhils, line 28: Replace with something like this?
> You [configure your routing table](https://docs.mitmproxy.org/stable/howto-transparent/) to send traffic through mitmproxy.

web/src/js/components/Modes/Socks.tsx
```diff
@@ -0,0 +1,61 @@
...
+            />
+        );
+    });
+
+    return (
+        <div>
+            <h4 className="mode-title">SOCKS Proxy</h4>
+            <p className="mode-description">
+                Similar the regular proxy mode, but using SOCKS5 instead of HTTP
+                for connection establishment with the proxy.
```
mhils, line 28: Or just:

> You manually configure your client application or device to use a SOCKSv5 proxy.

To match what we have for the regular proxy. What do you think? :)

web/src/js/components/Modes.tsx
```diff
@@ -18,6 +20,11 @@ export default function Modes() {
                 {!platform.startsWith("linux") ? <Local /> : undefined}
                 <Wireguard />
                 <Reverse />
+            </div>
+            <h3>Advanced</h3>
+            <div className="modes-container">
+                <Transparent />
+                <Socks />
```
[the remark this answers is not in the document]
lups2000: ahaha okay 😅

web/src/js/components/Modes/Socks.tsx
lups2000: yes yes it's fine for me, I just copied some text from the docs but I wanted to know your opinion about that :)

mhils (commented):
We're missing a bit of coverage here because our test backendState is hardcoded to Windows (which now has no transparent mode).

Let's update https://github.com/mitmproxy/mitmproxy/blob/0f0c5ee250369476a689eea83e82b042c223ead0/web/gen/state_js.py#L48
to macOS (darwin) instead?

web/src/js/components/Modes/Transparent.tsx
```diff
@@ -0,0 +1,62 @@
...
+                backendState={backendState[getSpec(server)]}
+            />
+        );
+    });
+
+    return (
+        <div>
+            <h4 className="mode-title">Transparent Proxy</h4>
+            <p className="mode-description">
+                You configure your routing table to send traffic through
```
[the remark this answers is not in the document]
lups2000: yes sure, forgot it

PR #6412, opened 2023-10-19
Fix path() docstring per #6345

tddschn:
#### Description

Fix path() docstring per #6345

Before:

https://docs.mitmproxy.org/stable/api/mitmproxy/http.html#Request.path

CleanShot-2023-10-19T09-41-04@2x

After:

CleanShot-2023-10-19T09-39-07@2x

#### Checklist

 - [ ] I have updated tests where applicable.
 - [x] I have added an entry to the CHANGELOG.

PR #7110, opened 2024-08-16
Feature/dns and upstream

lups2000:
#### Description

In this PR, I'm adding the last modes: **upstream** and **dns** #6412 :)
The descriptions are extracted from the docs, so let me know if you would change something!
Ref: #7063 

PS: I've also cleaned the structure by moving the `ReverseRow.tsx` component into the parent ,` Reverse.tsx `

https://github.com/user-attachments/assets/682a7013-55ad-4250-95d9-7f4a72398c48

#### Checklist


lups2000:
@mhils maybe you've missed it :) let me know if I should add/fix something! Thx 😃

web/src/js/components/Modes/Dns.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    return (
+        <div>
+            <ModeToggle
+                value={server.active}
+                onChange={() =>
+                    dispatch(setActive({ server, value: !server.active }))
+                }
+            >
+                Run DNS Proxy
```
mhils, line 57:
```suggestion
                Run DNS Server
```
It's technically a recursive DNS resolver, but I feel "DNS Server" should be good enough.

web/src/js/__tests__/ducks/modes/upstreamSpec.tsx
```diff
@@ -0,0 +1,165 @@
...
+} from "../../../ducks/modes/upstream";
+import {
+    RECEIVE as STATE_RECEIVE,
+    BackendState,
+} from "../../../ducks/backendState";
+import { TStore } from "../tutils";
+import { PayloadAction } from "@reduxjs/toolkit";
+import { UpstreamProxyProtocols } from "../../../modes/upstream";
+
+describe("upstreanSlice", () => {
```
mhils, line 18:
```suggestion
describe("upstreamSlice", () => {
```

web/src/js/components/Modes/Upstream.tsx
```diff
@@ -55,7 +72,7 @@ export default function ReverseToggleRow({
                     dispatch(setActive({ server, value: !server.active }));
                 }}
             >
-                Forward
+                Transfer
```
mhils, line 75: Diff is borked, but I'll comment here anyhow: I think we should move the wording closer to what we have for the explicit HTTP(S) Proxy. How about something like this?


--------------

### Explicit HTTP(S) Proxy (With Upstream Proxy)
All requests are forwarded to a second HTTP(S) proxy server.

- [x] Run HTTP/S Proxy and forward requests to [text input with placeholder: http://example.com:8080].

---------------------------------

I'd be fine with both a single text input for the destination, as well as dropdown + hostname/port inputs. I think for this one it's way more intuitive than for reverse proxy mode, where we "abuse" the scheme a bit with non-standard values such as `udp://`.

lups2000: okay, let's keep the things simple and use just a single input field for the destination :)

mhils (approved):
🚀 🚀 🚀

Issue #7112, opened 2024-08-16
QUIC layer does not handle `StopSendingReceived` event


errorxyz:
#### Problem Description
Our `QuicLayer` implementation does not handle the `StopSendingReceived` event raised by `QuicLayer.quic`
https://github.com/mitmproxy/mitmproxy/blob/6bb536e0ff9e4abe940cdd0f17ee76851e38bb2f/mitmproxy/proxy/layers/quic.py#L1023
refs https://github.com/mitmproxy/mitmproxy/issues/7025

#### Steps to reproduce the behavior:
1. Start HTTP3 reverse proxy: `mitmdump --mode reverse:https://http3.is@443 --mode reverse:http3://http3.is@443`
2. Ensure HTTP3 works by visiting the reverse proxy address
3. Repeatedly reload the website

mitmproxy crashes with the following error: `AssertionError: Unexpected event: StopSendingReceived(error_code=268, stream_id=20)`

Note: This crash might arise while accessing the website for the first time via firefox as well.
Also, despite the crash, the website seems to be loading up normally on the browser.

#### System Information
Mitmproxy: 11.0.0.dev (+17, commit be4ba11)
Python:    3.11.3
OpenSSL:   OpenSSL 3.2.2 4 Jun 2024
Platform:  Windows-10-10.0.22631-SP0

PR #7114, opened 2024-08-17
Merge https and http3 reverse-proxy modes

errorxyz:
#### Description

Merge http3 reverse proxy mode into https mode so that we listen for http3 connections as well while in https mode

#### Checklist

### DIFF
--- a/mitmproxy/addons/next_layer.py
+++ b/mitmproxy/addons/next_layer.py
@@ -350,10 +350,15 @@ class NextLayer:
                     stack /= ClientTLSLayer(context)
                 stack /= HttpLayer(context, HTTPMode.transparent)
             case "https":
-                stack /= ServerTLSLayer(context)
-                if starts_like_tls_record(data_client):
-                    stack /= ClientTLSLayer(context)
-                stack /= HttpLayer(context, HTTPMode.transparent)
+                if _starts_like_quic(data_client):
+                    stack /= ServerQuicLayer(context)
+                    stack /= ClientQuicLayer(context)
+                    stack /= HttpLayer(context, HTTPMode.transparent)
+                else:
+                    stack /= ServerTLSLayer(context)
+                    if starts_like_tls_record(data_client):
+                        stack /= ClientTLSLayer(context)
+                    stack /= HttpLayer(context, HTTPMode.transparent)
 
             case "tcp":
                 if starts_like_tls_record(data_client):
@@ -386,10 +391,6 @@ class NextLayer:
                 #     stack /= ClientTLSLayer(context)
                 stack /= DNSLayer(context)
 
-            case "http3":
-                stack /= ServerQuicLayer(context)
-                stack /= ClientQuicLayer(context)
-                stack /= HttpLayer(context, HTTPMode.transparent)
             case "quic":
                 stack /= ServerQuicLayer(context)
                 stack /= ClientQuicLayer(context)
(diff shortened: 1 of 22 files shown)


mitmproxy/addons/next_layer.py
```diff
@@ -350,10 +350,15 @@ def _setup_reverse_proxy(context: Context, data_client: bytes) -> Layer:
                     stack /= ClientTLSLayer(context)
                 stack /= HttpLayer(context, HTTPMode.transparent)
             case "https":
-                stack /= ServerTLSLayer(context)
-                if starts_like_tls_record(data_client):
-                    stack /= ClientTLSLayer(context)
-                stack /= HttpLayer(context, HTTPMode.transparent)
+                if _starts_like_quic(data_client):
```
mhils, line 353: Can we inspect client_conn.transport_protocol instead? That feels way more robust.

mitmproxy/addons/proxyserver.py
```diff
@@ -259,14 +259,21 @@ def configure(self, updated) -> None:
...
-                )
-                for m in modes
-            ]
+            listen_addrs = []
+            for m in modes:
+                protocols = []
+                if m.transport_protocol == "both":
+                    protocols.extend(["tcp", "udp"])
+                else:
+                    protocols.append(m.transport_protocol)
```
mhils, line 268:
```suggestion
                if m.transport_protocol == "both":
                    protocols = ["tcp", "udp"]
                else:
                    protocols = [m.transport_protocol]
```

mitmproxy/proxy/mode_specs.py
```diff
@@ -223,17 +223,15 @@ class ReverseMode(ProxyMode):
 
     description = "reverse proxy"
     transport_protocol = TCP
-    scheme: Literal[
-        "http", "https", "http3", "tls", "dtls", "tcp", "udp", "dns", "quic"
-    ]
+    scheme: Literal["http", "https", "tls", "dtls", "tcp", "udp", "dns", "quic"]
```
mhils, line 226: For reverse proxy mode, does it maybe make sense to keep `http3`? This way a user can explicitly specify HTTP/3 only. `quic://` forces raw QUIC instead.

(I wouldn't mind adding `http1://` and `http2://` eventually)

mhils (commented):
So nice how small and easy this change is now. 🤩

mitmproxy/proxy/mode_specs.py
errorxyz: Sounds good.

mitmproxy/addons/proxyserver.py
errorxyz: We get a mypy error with this if-else pattern
`error: List item 0 has incompatible type "Literal['tcp', 'udp'] | None"; expected "str"  [list-item]`
so I've reversed the conditions
We could instead set the variable type using `protocols: list[Literal['tcp', 'udp'] | None]`, lmk if that's better

errorxyz:
shall we block this untill https://github.com/mitmproxy/mitmproxy/issues/7112 is fixed?

mitmproxy/addons/proxyserver.py
mhils: Ah, thanks. The mypy complaint isn't completely unjustified, `m.transport_protocol` can be None by it's type annotation. It's only set to None for local mode, and that comes from a time where "both" was not a valid option yet.  Let's remove the `| None` part for transport_protocol, make it `both` for Local/OsProxy mode, and this error should go away? 
(If it doesn't, a `protocols: list[TransportProtocol]` line before the if/else should fix it.)

mhils (approved):

PR #7118, opened 2024-08-18
Miscellaneous QUIC Improvements

mhils:
#### Description

This PR fixes a bunch of smaller QUIC and HTTP/3 things. Most notably, it removes all support for PUSH_PROMISE frames. mitmproxy will not advertise PUSH_PROMISE support going forward, matching our current HTTP/2 implementation. We haven't received a single complaint or feature request after removing HTTP/2 push promise support, so we shouldn't burden ourselves with a half-completed HTTP/3 variant.

The easiest way to remove this is probably to go commit-by-commit.


errorxyz (approved):
LGTM🚀

PR #7121, opened 2024-08-20
Optimize tnet parsing

### LINKED ISSUE #7120: Endless loop when processing very large tnetstring
#### Problem Description

The loading of flows stored in mitmdump capture files can get stuck endlessly (manually terminated after 10+ hours). This happens for capture files that contain very large `tnetstring` for dict and/or list objects. The issue originates from the recursive slicing performed in the `pop` function while parsing (`parse` function) a tnetstring using the [tnetstring](https://github.com/mitmproxy/mitmproxy/blob/6bb536e0ff9e4abe940cdd0f17ee76851e38bb2f/mitmproxy/io/tnetstring.py#L229) module. 

I solved this issue by using `memoryview` to represent the data to be parsed in the `tnetstring` module. In the following archive
[tnet.zip](https://github.com/user-attachments/files/16661981/tnet.zip), you can find the modified tnetstring module and code  + data to test it.

Env:
- python 3.10.13
- mitmproxy 10.2.4

1. Unzip [tnet.zip](https://github.com/user-attachments/files/16661981/tnet.zip)
2. run `python test_tnetstring.py`

The test script will try to parse a large tnetstring (100+ MB), stored in the file part of the .zip, using both the original and the modified tnetstring module. On my machine the modified code ends in ~ 7 seconds while the original runs for hours (I have never seen the end of it). To avoid waiting forever, the script kills the execution of the original code after a timeout 120 seconds.
 
In the `test_tnetstring.py` script there is also code to check that the result produced by the original code and the modified one are the same. The call to this test function is commented out; you can specify the path of a not-problematic mitmdump capture and run it.

#### System Information
- python 3.10.13
- mitmproxy 10.2.4

mik1904:
#### Description

Fix #7120

**Note to reviewers:** I had to add commit 5a67118a37e811c3fe8c402f835547aef1a381a3 to accomodate the existing pytest. The reason is that in `test/mitmproxy/io/test_tnetstring.py` the function `tnetstring.pop` is called directly and bytes are passed as `data` argument. However, as far as I can tell, `pop` is never called directly in mitmproxy main code but only indirectly through the `tnetstring.parse` and `tnetstring.loads` functions. I let you decided whether it is better to modify calls to `pop` in `test_tnetstring.py` and discard commit 5a67118a37e811c3fe8c402f835547aef1a381a3 or keep it as it is.

#### Checklist

### DIFF
--- a/mitmproxy/io/tnetstring.py
+++ b/mitmproxy/io/tnetstring.py
@@ -229,7 +229,8 @@ def parse(data_type: int, data: memoryview) -> TSerializable:
 def split(data: memoryview, sep: bytes) -> tuple[int, memoryview]:
     i = 0
     try:
-        while data[i] != ord(sep):
+        ord_sep = ord(sep)
+        while data[i] != ord_sep:
             i += 1
         # here i is the position of b":" in the memoryview
         return int(data[:i]), data[i + 1 :]
--- a/mitmproxy/io/tnetstring.py
+++ b/mitmproxy/io/tnetstring.py
@@ -237,12 +237,14 @@ def split(data: memoryview, sep: bytes) -> tuple[int, memoryview]:
 
 
 
-def pop(data: memoryview) -> tuple[TSerializable, memoryview]:
+def pop(data: memoryview | bytes) -> tuple[TSerializable, memoryview]:
     """
     This function parses a tnetstring into a python object.
     It returns a tuple giving the parsed object and a string
     containing any unparsed data from the end of the string.
     """
+    if isinstance(data, bytes):
+        data = memoryview(data)
     #  Parse out data length, type and remaining string.
     length, data = split(data, b":")
     try:
--- a/mitmproxy/io/tnetstring.py
+++ b/mitmproxy/io/tnetstring.py
@@ -202,7 +202,7 @@ def parse(data_type: int, data: memoryview) -> TSerializable:
     if data_type == ord(b"!"):
         if data == b"true":
             return True
-        if data == b"false":
+        elif data == b"false":
             return False
         else:
             raise ValueError(f"not a tnetstring: invalid boolean literal: {data!r}")
--- a/mitmproxy/io/tnetstring.py
+++ b/mitmproxy/io/tnetstring.py
@@ -154,7 +154,7 @@ def loads(string: bytes) -> TSerializable:
     """
     This function parses a tnetstring into a python object.
     """
-    return pop(string)[0]
+    return pop(memoryview(string))[0]
 
 
 def load(file_handle: BinaryIO) -> TSerializable:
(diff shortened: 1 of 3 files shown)


mitmproxy/io/tnetstring.py
```diff
@@ -178,17 +178,17 @@ def load(file_handle: BinaryIO) -> TSerializable:
...
 
 
-def parse(data_type: int, data: bytes) -> TSerializable:
+def parse(data_type: int, data: memoryview) -> TSerializable:
     if data_type == ord(b","):
-        return data
+        return data.tobytes()
     if data_type == ord(b";"):
-        return data.decode("utf8")
+        return data.tobytes().decode("utf8")
```
mhils, line 191:
```suggestion
        return str(data, "utf8")
```
(I think this should avoid one copy operation)

mitmproxy/io/tnetstring.py
```diff
@@ -202,7 +202,7 @@ def parse(data_type: int, data: bytes) -> TSerializable:
     if data_type == ord(b"!"):
         if data == b"true":
             return True
-        elif data == b"false":
+        if data == b"false":
```
mhils, line 205:
```suggestion
        elif data == b"false":
```

(both should be fine and I'd like to keep the diff minimal)

mitmproxy/io/tnetstring.py
```diff
@@ -226,20 +226,29 @@ def parse(data_type: int, data: bytes) -> TSerializable:
     raise ValueError(f"unknown type tag: {data_type}")
 
 
-def pop(data: bytes) -> tuple[TSerializable, bytes]:
+def split(data: memoryview, sep: bytes) -> tuple[int, memoryview]:
+    i = 0
+    try:
+        while data[i] != ord(sep):
```
mhils, line 232: If I'm not mistaken this will evaluate `ord(sep)` for loop iteration. Do we want to cache it? I'd also be fine with inlining this method below and then hardcoding the separator. :)

mhils (commented):
Thanks, this is great! Adjusting the tests in `test_tnetstring.py` would be great IMO. :)

mitmproxy/io/tnetstring.py
mik1904: Personally, I would keep it more modular and therefore keep the split function. You make a good point about "caching" `ord(sep)` !

mik1904:
> Thanks, this is great! Adjusting the tests in `test_tnetstring.py` would be great IMO. :)

Took care of this at 6712916f042837e91983ee190b8a17ee8a1eb847

mhils (approved):
Thanks a bunch for contributing this! 🍰

PR #7136, opened 2024-08-29
add 2 new endpoints for processes extraction and process image

lups2000:
#### Description

In this PR, I'm adding two new endpoints to extract the current processes and to extract the executable image given a path. 

#### Checklist


lups2000:
Hi @mhils,
This seems to be working well, but I need some guidance, especially on structuring the code properly. Let me know what I should change!
Thank you! :)

mitmproxy/tools/web/app.py
```diff
@@ -654,6 +656,43 @@ def get(self):
...
+            "executable": process.executable,
+            "is_system": process.is_system,
+            "display_name": process.display_name,
+        }
+
+    @staticmethod
+    def get_json():
+        processes = mitmproxy_rs.active_executables()
+        processes_json = [ProcessList.process_to_dict(process) for process in processes]
+        return {"processes": processes_json}
```
mhils, line 673: Let's keep this simpler:
```suggestion
    @staticmethod
    def get_json():
        processes = mitmproxy_rs.active_executables()
        return [
            {
                "is_visible": process.is_visible,
                "executable": process.executable,
                "is_system": process.is_system,
                "display_name": process.display_name,
            }
            for process in processes
        ]
```

mitmproxy/tools/web/app.py
```diff
@@ -654,6 +656,43 @@ def get(self):
...
+    def get(self):
+        path = self.get_query_argument("path", None)
+
+        if not path:
+            self.set_status(400)
+            self.write({"error": "Missing 'path' parameter."})
+            return
+
+        try:
+            icon_bytes = mitmproxy_rs.executable_icon(pathlib.Path(path))
```
mhils, line 689: Passing `path` directly does not work?
(I'm mostly asking because the mitmproxy_rs API docs indicate it should, and if it doesn't we should fix the docs)

mitmproxy/tools/web/app.py
```diff
@@ -654,6 +656,43 @@ def get(self):
...
+
+
+class ProcessImage(RequestHandler):
+    def get(self):
+        path = self.get_query_argument("path", None)
+
+        if not path:
+            self.set_status(400)
+            self.write({"error": "Missing 'path' parameter."})
+            return
```
mhils, line 686:
```suggestion
            raise APIError(400, "Missing 'path' parameter.")
```

mhils (commented):
Looks good! Can we add tests, please? 😃

mitmproxy/tools/web/app.py
lups2000: it works! You're right!

lups2000:
the tests are now fixed but I'm still unsure about them 😅

test/mitmproxy/tools/web/test_app.py
```diff
@@ -403,3 +404,72 @@ def test_websocket(self):
         # trigger on_close by opening a second connection.
         ws_client2 = yield websocket.websocket_connect(ws_url)
         ws_client2.close()
+
+
+class TestProcessList(tornado.testing.AsyncHTTPTestCase):
```
mhils, line 409: Looks good, but do we need to mock here? We can easily avoid that here, we don't have to assert anything specific about the processes. I was thinking of just having a test like this:

https://github.com/mitmproxy/mitmproxy/blob/b3a3f43bd10b23f3911d015eb4633e5cb9580d5a/test/mitmproxy/tools/web/test_app.py#L344-L347

with just `assert get_json(resp)` in the last line.

test/mitmproxy/tools/web/test_app.py
```diff
@@ -403,3 +404,72 @@ def test_websocket(self):
...
+                    "is_visible": False,
+                    "executable": "process2",
+                    "is_system": True,
+                    "display_name": "Process 2",
+                },
+            ]
+            assert json.loads(response.body) == expected_result
+
+
+class TestProcessImage(tornado.testing.AsyncHTTPTestCase):
```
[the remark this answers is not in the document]
lups2000: let me know what do you think now with this solution :)

PR #7139, opened 2024-08-30
Warn if TLS version is unsupported by OpenSSL

### LINKED ISSUE #7138: Warn if `tls_version_client_min` is not supported by the current OpenSSL build
From https://github.com/mitmproxy/mitmproxy/discussions/7082: When the user configures tls_version_(client|server)_min, we should find a way to check if that is actually supported and warn if it's not.

mhils:
#### Description

fix #7138

#### Checklist


errorxyz (approved):
LGTM

PR #7141, opened 2024-09-01
Feature/local applications selector

lups2000:
#### Description

In this PR, I'm adding the local applications selector to the new capture page. The PR is still in draft since we're still missing the endpoints (#7136) and I'm using mocked data now. Any feedback is more than welcome :)
Demo: 

https://github.com/user-attachments/assets/19f02071-edb2-45ca-87a4-084ac99243d2

#### Checklist


mhils:
Thanks! This looks like a good first step. I like the idea with using badges for active applications. Can we turn the dropdown into an input field with an autocomplete-style popover?

Something in the style of https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist would be nice. But we probably can't use datalist itself, because that doesn't allow for icons or text that's different from the executable name. But just a popover that lists processes + adds them on click would maybe be good enough?

lups2000:
> Thanks! This looks like a good first step. I like the idea with using badges for active applications. Can we turn the dropdown into an input field with an autocomplete-style popover?
> 
> Something in the style of https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist would be nice. But we probably can't use datalist itself, because that doesn't allow for icons or text that's different from the executable name. But just a popover that lists processes + adds them on click would maybe be good enough?

Sure! :) For the badges, would you prefer them to be displayed _outside_ the input field (like in the DEMO) or _inside_ it? I chose the first option because the second could become quite confusing if there are more than 10 selected applications.

mhils:
> For the badges, would you prefer them to be displayed _outside_ the input field (like in the DEMO) or _inside_ it? 

Outside will likely be much easier to implement, so let's go with this. I don't want us to pick up an additional dependency for this. That being said, I did like the layout in the demo GIF at https://www.npmjs.com/package/react-tag-input where the tags are to the left of the input (we don't want that drag and drop though please, too much complexity!). Maybe we can copy that?

> I chose the first option because the second could become quite confusing if there are more than 10 selected applications.

FWIW, I think we should optimize for 1-2 selected applications. Everything else is really a weird use case. :)



One detail we didn't touch upon so far: The default placeholder with no badges needs to be something like "all applications".

lups2000:
@mhils current status: 

https://github.com/user-attachments/assets/de26d699-ca6a-442f-838e-bcd62268fc86

mhils:
Very cool progress! 🤩 

1. I think one part of the UX that's a bit weird at the moment is removing entries. It would probably be simpler to have small x buttons on each badge?
2. When I was talking about the default placeholder in my previous comment, I meant the text that current reads "Search Applications". I would replace that with "(all applications)" and not have an extra badge.
3. Popover should probably move below the input in the style of a dropdown. Do we need the checkmarks?
4. Nit: Availability of an icon should not change text position.

lups2000:
> Very cool progress! 🤩
> 
> 1. I think one part of the UX that's a bit weird at the moment is removing entries. It would probably be simpler to have small x buttons on each badge?
> 2. When I was talking about the default placeholder in my previous comment, I meant the text that current reads "Search Applications". I would replace that with "(all applications)" and not have an extra badge.
> 3. Popover should probably move below the input in the style of a dropdown. Do we need the checkmarks?
> 4. Nit: Availability of an icon should not change text position.

Thanks @mhils :)  I've almost fixed everything now, but I'm still having trouble placing the popover below 😡 I'll try to fix it tomorrow😅
Screenshot 2024-09-02 at 21 16 32
Screenshot 2024-09-02 at 21 16 42

lups2000:
okay now the popover position is finally fixed 😄 

Screenshot 2024-09-03 at 08 26 02

lups2000:
> okay now the popover position is finally fixed 😄
> 
> Screenshot 2024-09-03 at 08 26 02
> But, it super weird that the webCI is failing due to this test (cannot find "my data"):
> 
> ```
> test("ProxyApp", async () => {
>     const cv: ContentViewData = {
>         lines: [[["text", "my data"]]],
>         description: "",
>     };
>     fetchMock.doMockOnceIf(
>         "./flows/flow2/request/content/Auto.json?lines=513",
>         JSON.stringify(cv),
>     );
>     render(<ProxyApp />);
>     expect(screen.getByTitle("Mitmproxy Version")).toBeDefined();
>     await waitFor(() => screen.getByText("my data"));
> });
> ```
> 
> I haven't touched anything related to `ProxyApp`

@mhils do you have any clue for this?

mhils:
> @mhils do you have any clue for this? This PR should be ready "content-wise". I should fix the CI and probably add some tests (?)

A good opportunity to improve your debugging skills. :) Here's how I went on this:

1. This PR has 23 commits. We can use `git bisect` to figure out that https://github.com/mitmproxy/mitmproxy/pull/7141/commits/d94645be76afe3f1301a3ed7e0bc34862e948d5f is the first bad one.
2. Based on that we can make an easy educated guess that the `fetchApi` call is the culprit here. 😛 
3. Uncommenting the fetchApi call does indeed fix the test.
4. To validate, adding `fetchMock.doMockOnceIf("/processes", "[]");` to the failing test also fixes it. But brings up the question why we need to do this here.

Looking closer at this, I think the best fix is to move the process list into redux state. https://redux.js.org/usage/writing-logic-thunks#using-createasyncthunk should be fairly straightforward to adapt. The useEffect in LocalDropdown turns into something like this:

```typescript
    React.useEffect(() => {
        if(processes.length === 0) dispatch(fetchProcesses());
    }, []);
```

 And `testState` can get some default processes = the fetch call disappears. We also get rid of the atrocious `isRefreshing` by directly dispatching from the refresh button. :)

lups2000:
Thanks a lot, @mhils, for your debugging approach! It's really helpful! :)
What confused me was that test, as I didn't really understand what it was testing. I didn’t consider `fetchApi` since this function was used in other parts of the code without causing any issues.
I'll adapt the code asap :)

web/src/js/ducks/modes/local.ts
```diff
@@ -3,20 +3,35 @@ import {
...
+    },
 );
 
 export const initialState: LocalState[] = [
     {
         active: false,
-        applications: "",
+        isLoading: false,
+        currentProcesses: [],
+        selectedProcesses: "",
```
mhils, line 34: What do you think about moving the process list state into its own slice at the top level? Technically speaking it's completely separate from the mode spec. :) And then I'm not confused on why we have both `currentProcesses` and `selectedProcesses`. 😅

mhils (commented):
This is so amazing now. 😍 

There is one final blocker left here that we need to address before merging: We can't use display name ("Google Chrome") for the spec, we need to use the process name ("chrome.exe"). For the process list I'd be happy with either "Google Chrome" (as it is right now), "chrome.exe", or "Google Chrome (chrome.exe)". Whatever you think works best! But for the spec it needs to be chrome.exe.

image


I've also added some new nits to https://github.com/mitmproxy/mitmproxy/issues/7063#issue-2438153247, but they are not blocking this PR at all. :)

lups2000:
> This is so amazing now. 😍
> 
> There is one final blocker left here that we need to address before merging: We can't use display name ("Google Chrome") for the spec, we need to use the process name ("chrome.exe"). For the process list I'd be happy with either "Google Chrome" (as it is right now), "chrome.exe", or "Google Chrome (chrome.exe)". Whatever you think works best! But for the spec it needs to be chrome.exe.
> 
> image
> 
> I've also added some new nits to [#7063 (comment)](https://github.com/mitmproxy/mitmproxy/issues/7063#issue-2438153247), but they are not blocking this PR at all. :)

Thanks for the feedback :) just one question, sometimes the `display_name` seems working and returning a “.exe”, sometimes not (like for Google Chrome). Is this an issue we need to address on the backend? Or can we extract simply the `process name` from the `executable` field of a `Process` ?

mhils:
> Or can we extract simply the `process name` from the `executable` field of a `Process` ?

Let's do that. Display name is supposed to be nicer for display, but lots of executables don't expose it. 🙃

lups2000:
@mhils let me know what you think with this solution :) I should have applied all the requested changes

mhils:
This looks great, only thing missing is tests for the UI component! You can take https://github.com/mitmproxy/mitmproxy/blob/main/web/src/js/__tests__/components/CommandBarSpec.tsx as inspiration - we don't need 100% coverage, but at least a simple "render, add text, observe list is filtered, click on curl.exe" kind of test would be good. :)

PR #7153, opened 2024-09-03
Click on mode's label to activate it

lups2000:
#### Description

In this PR, I'm adding the ability to activate or deactivate a mode by clicking on its label

Ref: minor stuff of #7063 

#### Checklist


lups2000:
> I think we can solve this much cleaner by using a label, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label :)

Okay I’ll adapt the code to that!

mhils (approved):

PR #7161, opened 2024-09-04
add popover to reverse mode

lups2000:
#### Description

In this PR, I'm adding a popover to the reverse mode. So now the `listen_host` and the `listen_port` are set in the popover. Ref: #7063 

Screenshot 2024-09-04 at 14 55 04

#### Checklist


mhils:
LGTM % merge conflict. I think we don't need to make "traffic to" clickable, the important part is the text close to the checkbox :)

mhils (approved):

Issue #7156, opened 2024-09-03
mitmproxy user in Docker container assigned UID/GID 0 by default


dr-yd:
#### Problem Description
By default, the `mitmproxy` process and user run as UID/GID 0. This seems unintentional and like an implementation artifact.

#### Steps to reproduce the behavior:
1. `docker run --rm -it --entrypoint /bin/bash mitmproxy/mitmproxy -c "ls -la /home/mitmproxy"`
3. Observe that `.mitmproxy` is owned by `root:root`
4. Observe that the entrypoint changes the UID/GID of `mitmproxy` to the UID/GID of this directory

#### Note
This is caused by the `VOLUME` declaration in the Dockerfile - it creates the directory with the current `USER` which is `root`. Subsequently chowning it to `mitmproxy` would fix this.

#### System Information
Mitmproxy: 10.4.2
Python:    3.11.9
OpenSSL:   OpenSSL 3.3.1 4 Jun 2024
Platform:  Linux-6.10.0-linuxkit-aarch64-with-glibc2.31

mhils:
Thanks for pointing this out! Would you mind sending a PR with a fix? 😃

dr-yd:
Oh, sorry for the lack of feedback, looks like the previous notification passed by me. Thanks @jasonstultz for making the change!

PR #7168, opened 2024-09-08
added chown for .mitmproxy volume

jasonstultz:
#### Description

A one-liner, resolving https://github.com/mitmproxy/mitmproxy/issues/7156

#### Checklist

### DIFF
--- a/release/docker/Dockerfile
+++ b/release/docker/Dockerfile
@@ -15,7 +15,7 @@ RUN pip install --no-index --find-links=/wheels mitmproxy
 RUN rm -rf /wheels
 
 VOLUME /home/mitmproxy/.mitmproxy
-RUN chown mitmproxy /home/mitmproxy/.mitmproxy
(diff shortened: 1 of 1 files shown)


mhils:
Thanks for the PR! 🍰 In #7156 you mentioned "Subsequently chowning it to mitmproxy would fix this." and now we chown first and then call VOLUME. I assume you've tested this and it works as expected now? :)

jasonstultz:
I'm not the original poster who created the issue; I just saw that there was a fairly straightforward issue to resolve.  

That said, I've confirmed that the mount has the expected permissions when the image is built.  I have not tested exhaustively for downstream bugs that may arise as a result of the permissions change.

mhils:
> I'm not the original poster who created the issue; I just saw that there was a fairly straightforward issue to resolve.

Ah, I missed that. Thanks! :)

PR #7175, opened 2024-09-09
options: add request_client_cert to enable mutual TLS

Kriechi:
This capability was already built-in but hard-coded to be disabled. Making it configurable as option (defaulting to off) enables mTLS connections between clients and mitmproxy. If true, mitmproxy will send a TLS `CertificateRequest` message to the client during the TLS handshake, upon which a client needs to present a client certificate to mitmproxy to successfully establish an mTLS connection.

This option can be used together with the `client_certs` option to also establish an mTLS connection between mitmproxy and the upstream server. In this case, mitmproxy needs to have a full client cert, including matching private key, that is trusted and accepted by the upstream server. This is a common scenario with MQTT or IoT connections.

Example usage:
$ mitmproxy --set request_client_cert=True --set client_certs=some_directory/

With `some_directory/` containing a `mqtt.example.com.pem` x509 certificate file (including private key).

This allows a client connecting using mTLS, to be intercepted by mitmproxy, which is itself establishing an mTLS connection to the `mqtt.example.com` upstream server. Restricting the client_certs using a directory and PEM files named after the upstream domain, narrows down the mTLS requirement to this single domain, while leaving all other traffic through mitmproxy untouched (normal TLS without client certs).

#### Checklist

### DIFF
--- a/docs/src/content/concepts-certificates.md
+++ b/docs/src/content/concepts-certificates.md
@@ -1,8 +1,8 @@
 ---
 title: "Certificates"
 menu:
-    concepts:
-        weight: 3
+  concepts:
+    weight: 3
 ---
 
 # About Certificates
@@ -47,7 +47,7 @@ that other mitmproxy users cannot intercept your traffic.
 
 ### Installing the mitmproxy CA certificate manually
 
-Sometimes using the [quick install app](#quick-setup) is not an option and you need to install the CA manually. 
+Sometimes using the [quick install app](#quick-setup) is not an option and you need to install the CA manually.
 Below is a list of pointers to manual certificate installation
 documentation for some common platforms. The mitmproxy CA cert is located in
 `~/.mitmproxy` after it has been generated at the first start of mitmproxy.
@@ -81,7 +81,7 @@ documentation for some common platforms. The mitmproxy CA cert is located in
 
 When mitmproxy receives a request to establish TLS (in the form of a ClientHello message), it puts the client on hold
 and first makes a connection to the upstream server to "sniff" the contents of its TLS certificate.
-The information gained -- Common Name, Organization, Subject Alternative Names -- is then used to generate a new 
+The information gained -- Common Name, Organization, Subject Alternative Names -- is then used to generate a new
 interception certificate on-the-fly, signed by the mitmproxy CA. Mitmproxy then returns to the client and continues
 the handshake with the newly-forged certificate.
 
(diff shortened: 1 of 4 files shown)


docs/src/content/concepts-certificates.md
```diff
@@ -198,16 +195,59 @@ openssl req -x509 -new -nodes -key ca.key -sha256 -out ca.crt -addext keyUsage=c
...
+generating a signature.
+
+Client certificates, just as the more commonly known server certificates, are
+full x509 certs with a public and private key, commonly stored as PEM
+certificate file.
+
+Some internet protocols use client and server certificates as a form of
+authentication and authorization, e.g., the MQTT protocol used in
+Internet-of-Things networks. To intercept such protocols using mitmproxy, both
+the client needs to asked to establish mTLS, and the upstream server needs to
```
mhils, line 216:
```suggestion
the client needs to be asked to establish mTLS, and the upstream server needs to
```

docs/src/content/concepts-certificates.md
```diff
@@ -198,16 +195,59 @@ openssl req -x509 -new -nodes -key ca.key -sha256 -out ca.crt -addext keyUsage=c
...
+believe that mitmproxy is a mTLS-compliant device. Such protocols require both
+options outlined in the next sections for mitmproxy to work correctly.
```
mhils, line 218: Why are both options required? I guess there may be some applications that do check this, but generally speaking (especially when using OpenSSL APIs) most clients will work just fine without `request_client_cert`. The important bit is that mitmproxy does have the client cert to authenticate upstream, but `request_client_cert` is typically optional. Do you have clients where this is a hard requirement?

docs/src/content/concepts-certificates.md
```diff
@@ -198,16 +195,59 @@ openssl req -x509 -new -nodes -key ca.key -sha256 -out ca.crt -addext keyUsage=c
...
+
+### mTLS between client and mitmproxy
+
+To instruct mitmproxy to request a client certificate, you can pass the `--set
+request_client_cert=True` option to mitmproxy. This will generate the
+`CertificateRequest` TLS handshake message and (if successful) establish an mTLS
+connection.
+
+This option can be used together with the `client_certs` option to also
+establish an mTLS connection between mitmproxy and the upstream server. In this
```
mhils, line 238: This is a bit misleading in the sense that we don't do any validation. So it's just TLS where we also ask the client nicely to maybe send a certificate. There is no authentication going on. 😛

mhils (commented):
Thanks! My high-level feedback is that I think we should rephrase this slightly towards two main messages:

1. `client_certs` can be used to establish mTLS between mitmproxy and server.
2. Some clients expect the server to request the client certificate (assuming you have one where this is the case?). To make those clients work, specify `request_client_cert`.

docs/src/content/concepts-certificates.md
Kriechi: Yes, this is a (at least in my visibility) common requirement of IoT devices and related firmware connectivity libraries. I'll rephrase this docs section to point out that this is not a hard requirement for every protocol and device.

Kriechi:
reworded as suggested!

Kriechi:
Happy to merge!

mhils (approved):

PR #7177, opened 2024-09-11
proxy server: skip modes without listen_port when calculating conflicts

mhils:


mitmproxy/proxy/mode_specs.py
```diff
@@ -147,7 +147,7 @@ def listen_host(self, default: str | None = None) -> str:
         else:
             return ""
 
-    def listen_port(self, default: int | None = None) -> int:
+    def listen_port(self, default: int | None = None) -> int | None:
```
errorxyz, line 150: nit: do we want to update docstring for the new function signature?

errorxyz (approved):

PR #7179, opened 2024-09-11
Fix of duration in import from HAR

dstd:
#### Description

Duration in `log.entries[].time` is specified in milliseconds, but was interpreted as seconds in `har.py: request_to_flow`. Which led to scary `123s` instead of real `123ms`.

#### Checklist


mhils (approved):
Thanks, super appreciated! 😃 🍰

Issue #7152, opened 2024-09-03
Document possible options for `web_columns`


nikhilweee:
#### Problem Description

The current [documentation](https://github.com/mitmproxy/mitmproxy/blob/a5b5963d8ca06ba70462082eed5ab595f14d8d56/mitmproxy/tools/web/webaddons.py#L17) about `web_columns` only lists the defaults. 

![image](https://github.com/user-attachments/assets/31f95d0f-cf73-4889-963a-48aa9cff5f49)

I recently discovered that there are [other columns](https://github.com/mitmproxy/mitmproxy/blob/a5b5963d8ca06ba70462082eed5ab595f14d8d56/web/src/js/components/FlowTable/FlowColumns.tsx#L196-L209) as well, and it would be great to include them as "possible options"

#### Proposal
Add a list of [possible columns](https://github.com/mitmproxy/mitmproxy/blob/a5b5963d8ca06ba70462082eed5ab595f14d8d56/web/src/js/components/FlowTable/FlowColumns.tsx#L196-L209).

#### Alternatives
Don't document additional columns.

mhils:
Yes, 💯 . We should add a script in https://github.com/mitmproxy/mitmproxy/tree/main/web/gen that generates a Python file with all column names. We can then use that for the option description. Contributions welcome!

PR #7182, opened 2024-09-14
web_columns: added a script in /mitmproxy/web/gen that generates a Python file with all column names. 

Abhishek-Bohora:
#### Description

This PR addresses issue #7152 
1. Created a new script generate_columns.py in web/gen/ to automatically generate a list of available web columns from the TypeScript source.  generated file web_columns.py contains AVAILABLE_WEB_COLUMNS and DEFAULT_COLUMNS.
2. Updated option_js.py to include the full list of available columns in the TypeScript output _options_gen.ts file.
3. Modified webaddons.py to use the new AVAILABLE_WEB_COLUMNS
#### Checklist

-  [x] I have updated tests where applicable.
 - [] I have added an entry to the CHANGELOG.

#### Note for reviewers 
any suggestions, improvements, or feedback on anything I may have missed or misunderstood. I'm happy to follow up on your comments.


web/src/js/ducks/_options_gen.ts
```diff
@@ -192,11 +192,14 @@ export const defaultState: OptionsState = {
     view_filter: undefined,
     view_order: "time",
     view_order_reversed: false,
-    web_columns: ["tls", "icon", "path", "method", "status", "size", "time"],
+    web_columns: ["tls", "icon", "index", "path", "method", "version", "status", "size", "time", "timestamp", "quickactions", "comment"],
```
lups2000, line 195: since this is the default state, I would list just the `DEFAULT_WEB_COLUMNS`

web/src/js/ducks/_options_gen.ts
```diff
@@ -192,11 +192,14 @@ export const defaultState: OptionsState = {
...
     web_debug: false,
     web_host: "127.0.0.1",
     web_open_browser: true,
     web_port: 8081,
     web_static_viewer: "",
     websocket: true,
 };
+
+export const AVAILABLE_WEB_COLUMNS = ["tls", "icon", "index", "path", "method", "version", "status", "size", "time", "timestamp", "quickactions", "comment"];
+export const DEFAULT_WEB_COLUMNS = ["tls", "icon", "path", "method", "status", "size", "time"];
```
lups2000, line 205: This is hardcoded and may easily fall out of sync with the backend

web/gen/generate_columns.py
```diff
@@ -0,0 +1,37 @@
...
+    with open(utils_path, "r") as file:
+        content = file.read()
+
+    match = re.search(r"export const sortFunctions = {([\s\S]*?)};", content)
+    if not match:
+        raise ValueError("Could not find sortFunctions in utils.ts")
+
+    sort_functions_text = match.group(1)
+
+    columns = re.findall(r"^\s*(\w+):", sort_functions_text, re.MULTILINE)
```
lups2000, line 20: Personally, I'm not a big fan of this approach. Here's what I would have done:

1. Hardcode the two string arrays here and then define them in `web_columns.py` (backend) and, for instance, in `backends/const.ts` (frontend).
2. This way, we ensure the two arrays are generated and available on both the frontend and backend.
3. We can then refactor a bit `sortFunctions` by improving the typing and utilizing the newly created variable `AVAILABLE_COLUMNS`.

what do you think? :)

lups2000 (commented):
Thanks for the PR! 😃 I've added a few comments for you to review. Maybe @mhils will have some feedback to share as well!

Abhishek-Bohora:
@lups2000 Thank you for your review I am working on it to resolve it.

Abhishek-Bohora:
Hello, @lups2000 I have tried to do it as per the comments can you please review it? Please let me know if anything I can do to improve or change. Thank you.

lups2000:
Hey @Abhishek-Bohora , thanks. I’ll have a look at it asap. Meanwhile could you please remove the generated files? These files (`app.css, app.js, vendor.css`) shouldn’t be pushed 😄

mitmproxy/tools/web/webaddons.py
```diff
@@ -14,8 +15,8 @@ def load(self, loader):
         loader.add_option(
             "web_columns",
             Sequence[str],
-            ["tls", "icon", "path", "method", "status", "size", "time"],
-            "Columns to show in the flow list",
+            AVAILABLE_WEB_COLUMNS,
+            "all available columns to show in the flow list",
```
mhils, line 19:
```suggestion
            f"Columns to show in the flow list. Can be one of the following: {', '.join(AVAILABLE_WEB_COLUMNS)}",
```

(or something along those lines)

mhils (commented):
Thanks! 🍰 

I would like to simplify this. We already have the list of all columns in TypeScript, e.g. [here](https://github.com/mitmproxy/mitmproxy/blob/65db82f4cf0c294a3e9de103c4789ba4b9b19a41/web/src/js/components/FlowTable/FlowColumns.tsx#L196-L209). We can use this as the source of truth so that we don't need to duplicate it in the generation script.

We can add a marker here:
```typescript
const FlowColumns: { [key in keyof typeof sortFunctions]: FlowColumn } = {
    // parsed by web/gen/web_columns
    icon,
    [...]
    comment,
};
```

and then use a simple regex (`// parsed by web/gen/web_columns([\s\w,]+)`) to extract all names and generate `mitmproxy/tools/web/web_columns.py`. We don't need to do anything with the default columns. Does that make sense?

Abhishek-Bohora:
Hello, @mhils Thanks for your clarifying comments, quick question here.

In the `_option_gen.ts` file, which is auto-generated by `web/gen/options_js.py`, in `web_columns: []`, do we need to have `AVAILABLE_WEB_COLUMNS`(all the columns) or (only some columns) which I was referring earlier as `DEFAULT_COLUMNS` ?

```ts
export const defaultState: OptionsState = {
    web_columns: ["tls", "icon", "index", "path", "method", "version", "status", 
                  "size", "time", "timestamp", "quickactions", "comment"],

    // or

    web_columns: ["tls", "icon", "path", "method", "status", "size", "time"],
};
```
If we need only some columns, then while generating the `mitmproxy/tools/web/web_columns.py`, do we also have to create `DEFAULT_WEB_COLUMNS` along with `AVAILABLE_WEB_COLUMNS` so we can use it here ? otherwise while running `web/gen/options_js.py` this `web_columns: []` will be filled with all the `AVAILABLE_WEB_COLUMNS` ? any idea or suggestion ?

mhils:
> _option_gen.ts

This file contains the default options. It should just have what the default value for `web_columns` is. The default columns shouldn't change and can remain hardcoded in webaddons.py as-is.

Abhishek-Bohora:
Main Summary of what I have done to simplify according to the constructive comments from @mhils 

in `mitmproxy\web\gen\web_columns.py`
Instead of hardcoding the `AVAILABLE_WEB_COLUMNS` I have extracted all the available columns 
from [here](https://github.com/mitmproxy/mitmproxy/blob/65db82f4cf0c294a3e9de103c4789ba4b9b19a41/web/src/js/components/FlowTable/FlowColumns.tsx#L196-L209) using regex which acts as a single source of truth for all the available columns we have and generated the `mitmproxy\tools\web\web_columns.py` file which contains `AVAILABLE_WEB_COLUMNS` and `DEFAULT_WEB_COLUMNS`.
why `DEFAULT_WEB_COLUMNS` hardcoded in the generation script it will always be consistent and makes sure that it will be generated the same when running generation script.  and we and we can use that variable here in `mitmproxy\web\gen\options_js.py` while creating defaultState 
```python
print("export const defaultState: OptionsState = {")
for _, opt in sorted(m.options.items()):
    if opt.name == "web_columns":
        print(f" {opt.name}: {json.dumps(DEFAULT_WEB_COLUMNS)},")
    else:
        print(
            f" {opt.name}: {json.dumps(opt.default)},".replace(
                ": null", ": undefined"
            )
        )
print("};")
```
Thank You

mitmproxy/tools/web/webaddons.py
```diff
@@ -14,8 +15,8 @@ def load(self, loader):
         loader.add_option(
             "web_columns",
             Sequence[str],
-            ["tls", "icon", "path", "method", "status", "size", "time"],
-            "Columns to show in the flow list",
+            AVAILABLE_WEB_COLUMNS,
```
mhils, line 18:
```suggestion
            ["tls", "icon", "path", "method", "status", "size", "time"],
```

mitmproxy/tools/web/web_columns.py
```diff
@@ -0,0 +1,16 @@
...
+    "path",
+    "quickactions",
+    "size",
+    "status",
+    "time",
+    "timestamp",
+    "tls",
+    "comment",
+]
+DEFAULT_WEB_COLUMNS = ["tls", "icon", "path", "method", "status", "size", "time"]
```
mhils, line 16: This is already hardcoded in mitmproxy/tools/web/webaddons.py, we don't need this to be generated somehow.

web/gen/backend_consts.py
```diff
@@ -25,7 +27,10 @@ async def make() -> str:
         "/** Auto-generated by web/gen/backend_consts.py */\n"
         "export enum ReverseProxyProtocols {\n"
         f"    {protocols},\n"
-        "}\n"
+        "}\n\n"
+        "export const AVAILABLE_WEB_COLUMNS = "
+        + json.dumps(AVAILABLE_WEB_COLUMNS, indent=4)
+        + ";\n\n"
```
mhils, line 33: This needs to go - we don't need to add anything to `consts.ts`.

web/gen/options_js.py
```diff
@@ -49,11 +50,14 @@ async def make() -> str:
...
-            print(
-                f"    {opt.name}: {json.dumps(opt.default)},".replace(
-                    ": null", ": undefined"
+            if opt.name == "web_columns":
+                print(f"    {opt.name}: {json.dumps(DEFAULT_WEB_COLUMNS)},")
+            else:
+                print(
+                    f"    {opt.name}: {json.dumps(opt.default)},".replace(
+                        ": null", ": undefined"
+                    )
```
mhils, line 59: This needs to go unless there's justification for it. You didn't address my previous comment.

web/gen/web_columns.py
```diff
@@ -0,0 +1,48 @@
...
+from pathlib import Path
+
+
+def extract_flow_columns(filename):
+    try:
+        with open(filename, "r") as file:
+            content = file.read()
+    except FileNotFoundError:
+        print(f"Error: File '{filename}' not found.")
+        return []
```
mhils, line 14: This here is actively harmful error handling. First this case should never happen; and second, if it happens, it should not silently swallow the error.

web/src/js/flow/utils.ts
```diff
@@ -284,7 +285,9 @@ export const getVersion = (flow: Flow): string => {
     }
 };
 
-export const sortFunctions = {
+type SortFunction = (flow:Flow) => string | number | boolean | undefined
+
+export const sortFunctions:Record<typeof AVAILABLE_WEB_COLUMNS[number] , SortFunction> = {
```
mhils, line 290: This file does not need any changes.

web/gen/web_columns.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
+
+import json
+import re
+from pathlib import Path
+
+
+def extract_flow_columns(filename):
```
mhils, line 8: There is no need to parametrize this, it's always the same hardcoded filename.

mhils (changes requested):
Please also fix CI failures and remove the files in mitmproxy/tools/web/static, thanks.

web/gen/options_js.py
Abhishek-Bohora: I didn't know the default values were hardcoded in the `webaddons.py` so I added If else to check to fill the web_columns with the default value. Now I will remove this 😊

mhils:
> so I am thinking of writing this way

I don't quite understand what your question is here, but overall this looks ok. You can inline `generate_web_columns` into `make`. 

>And another quick question do I have to write individual test for this file ?

Yes. This can be as simple as

```python
def test_simple():
    assert isinstance(AVAILABLE_WEB_COLUMNS, list)
```

Abhishek-Bohora:
Hello, @mhils I have made all the changes according to your guidance. But this one CI error I couldn't figure out why it is happening. Some permissions denied type of error in CI `autofix.ci`. 
pm

Abhishek-Bohora:
Hello @mhils I am running Windows 11 so I tried to make the scripts executable via gitbash like this
gitbash
but it didn't work somehow it failed again. so I decided to add `- run: chmod +x web/gen/*.py` this one line in `autofix.yml` so maybe in the future anyone running Windows wouldn't face the problem. If this breaks something or is bad please let me know. thank you.

mhils:
Please clean up your PR. We don't need to change in `autofix.yml` (check out `git chmod`), and there are still unrelated changes.

mhils (changes requested):
You still need to clean up your PR. There are still tons of unrelated changes: https://github.com/mitmproxy/mitmproxy/pull/7182/files

Please put in more effort at addressing my comments before requesting a review again. I've already remarked this in https://github.com/mitmproxy/mitmproxy/pull/7182#pullrequestreview-2310267845 and https://github.com/mitmproxy/mitmproxy/pull/7182#issuecomment-2363493014.

PR #7183, opened 2024-09-15
Fix error while changing string-based options

errorxyz:
#### Description

Trying to change string based options such as `mode` or `dns_name_servers` results in a `AttributeError: 'AttrMap' object has no attribute 'get_text'`.
cause: https://github.com/mitmproxy/mitmproxy/pull/7098

#### Checklist


mhils (approved):
Thanks for catching this! 🍰

PR #7185, opened 2024-09-17
strip HTTPS records ALPN if http3 is disabled

mhils:
#### Description

This PR makes us strip `h3` from HTTPS records if http3 is disabled. :)

#### Checklist


mitmproxy/addons/strip_dns_https_records.py
```diff
@@ -0,0 +1,29 @@
...
+        if ctx.options.strip_ech:
+            for answer in flow.response.answers:
+                if answer.type == types.HTTPS:
+                    answer.https_ech = None
+        if not ctx.options.http3:
+            for answer in flow.response.answers:
+                if (
+                    answer.type == types.HTTPS
+                    and answer.https_alpn is not None
+                    and b"h3" in answer.https_alpn
```
errorxyz, line 26: alpn values can also be something like `h3-29` or `h3-28` that'd represent http3 as well

mitmproxy/addons/strip_dns_https_records.py
```diff
@@ -0,0 +1,35 @@
...
+            for answer in flow.response.answers:
+                if answer.type == types.HTTPS:
+                    answer.https_ech = None
+        if not ctx.options.http3:
+            for answer in flow.response.answers:
+                if (
+                    answer.type == types.HTTPS
+                    and answer.https_alpn is not None
+                    and any(
+                        a == b"h3" or a.startswith(b"h3-") for a in answer.https_alpn
```
errorxyz, line 27:
```suggestion
                        a.startswith(b"h3") for a in answer.https_alpn
```
can we do this? (and similar change few lines below)

errorxyz (commented):
LGTM % nit:

mitmproxy/addons/strip_dns_https_records.py
mhils: We surely can, but that makes the code less obvious I think. :) Right now it's more obvious that we want `h3` or `h3-` draft variants. With just `.startswith(b"h3")`, you may quickly be wondering why it's not just `==`.

(the best solution here is a comment, which I will add now)

PR #7192, opened 2024-09-19
Support empty string as marker in flow.mark

bburky:
#### Description

"Un-set all marks" is implemented as `flow.mark @all false`

`_MarkerType()` coerces `False` into `""`, but this was previously unsupported by `flow.mark`. This previously caused `flow.mark` to error with "invalid marker value".
https://github.com/mitmproxy/mitmproxy/blob/2c1802692ddb99b8d02a353d8382772029168616/mitmproxy/types.py#L457-L464

Add a check for the empty string to `flow.mark` to fix this.

I assume this previously worked, but the error checking got too strict at some point.

#### Checklist


mhils (approved):
Makes sense, thanks for the PR! 🍰

PR #7196, opened 2024-09-19
Propogate end_stream from downstream http/2 requests

rubu:
#### Description

I encountered an issue where mitmdump was failing with a request that was working in the browser. It was a GET request with no body but the server was throwing an error about GET request having a body.

After looking at packet captures the difference between mitmdump and browser was that browser was sending `END STREAM` on the `HEADERS` frame whereas mitmdump was sending an empty `DATA` frame with `END STREAM`.

I patched mitmdump to propogate `END STREAM` if it is present on the downstream `HEADERS` frame.

Tests pass and my test case works, could anyone elaborate if this can cause other issues? If so I can adjust this to get to a state where this can be merged - since this is breaking for me ill roll with a local patch now but it would be great to use an official release.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@
   ([#7114](https://github.com/mitmproxy/mitmproxy/pull/7114), @errorxyz)
 - Addon to update the hosts in alt-svc header in reverse mode
   ([#7093](https://github.com/mitmproxy/mitmproxy/pull/7093), @errorxyz)
+- Do not send unnecessary empty data frames when streaming HTTP/2.
 - Fix a bug where mitmproxy would ignore Ctrl+C/SIGTERM on OpenBSD.
   ([#7130](https://github.com/mitmproxy/mitmproxy/pull/7130), @catap)
 - Fix of measurement unit in HAR import, duration is in milliseconds
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - Addon to update the hosts in alt-svc header in reverse mode
   ([#7093](https://github.com/mitmproxy/mitmproxy/pull/7093), @errorxyz)
 - Do not send unnecessary empty data frames when streaming HTTP/2.
+  ([#7196](https://github.com/mitmproxy/mitmproxy/pull/7196), @rubu)
 - Fix a bug where mitmproxy would ignore Ctrl+C/SIGTERM on OpenBSD.
   ([#7130](https://github.com/mitmproxy/mitmproxy/pull/7130), @catap)
 - Fix of measurement unit in HAR import, duration is in milliseconds
--- a/mitmproxy/proxy/layers/http/__init__.py
+++ b/mitmproxy/proxy/layers/http/__init__.py
@@ -278,13 +278,13 @@ class HttpStream(layer.Layer):
             )
             self.flow.request.headers.pop("expect")
 
-        if self.flow.request.stream:
-            yield from self.start_request_stream(event)
+        if self.flow.request.stream and not event.end_stream: 
+            yield from self.start_request_stream()
         else:
             self.client_state = self.state_consume_request_body
         self.server_state = self.state_wait_for_response_headers
 
-    def start_request_stream(self, event=None) -> layer.CommandGenerator[None]:
+    def start_request_stream(self) -> layer.CommandGenerator[None]:
         if self.flow.response:
             raise NotImplementedError(
                 "Can't set a response and enable streaming at the same time."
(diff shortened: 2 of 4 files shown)


mhils:
Thanks! Generally strong +1 on this, great find. :) 

This needs a test in `test/mitmproxy/proxy/layers/http/test_http2.py` and ideally the same logic applied to responses. You can probably add a new test based on the following two without too much hassle:
https://github.com/mitmproxy/mitmproxy/blob/v10.4.2/test/mitmproxy/proxy/layers/http/test_http2.py#L95-L144
https://github.com/mitmproxy/mitmproxy/blob/v10.4.2/test/mitmproxy/proxy/layers/http/test_http2.py#L223-L243
Let me know if something is unclear, happy to answer questions. 

I'm also wondering if we can just change
https://github.com/mitmproxy/mitmproxy/blob/8964deda8a4476f220737296cba776b7a7b802fc/mitmproxy/proxy/layers/http/__init__.py#L281
to 
```python3
if self.flow.request.stream and not event.end_stream:
```

as a simpler patch, but that is easy to test out once we have a test. Thanks!

rubu:
Thank you for the comment - the suggested change does make sense and could be more elegant, ill try to make a test and then verify if that change works.

What did you mean by "and ideally the same logic applied to responses" - this was regarding tests or something else?

mhils:
> What did you mean by "and ideally the same logic applied to responses" - this was regarding tests or something else?

The PR currently removes empty data frames for requests, but it doesn't do so for responses. We basically have the same again here for responses:
https://github.com/mitmproxy/mitmproxy/blob/db61b4bca632827175ef31fc0d1fb398abab3043/mitmproxy/proxy/layers/http/__init__.py#L412-L415

This should be fixable in exactly the same way, so I suggest we do both in one go. :)

mhils:
Hi @rubu!

> are there some docs on the playbook apis to understand all the hooks?

Not really, but [this here](https://github.com/mitmproxy/mitmproxy/blob/e1ea491ad7508d4f9ce4630d3f6aba5af99cbfef/mitmproxy/proxy/__init__.py#L13-L17) is relevant.

```
        << OpenConnection(server)
        << DataReceived(tctx.server, frames_from_mitmproxy := tctx.server.data) # there is no data in server, this is what Chat GPT gave me
        << reply()
```

My recommendation would be to skip ChatGPT for this. `<< DataReceived` doesn't make sense because DataReceived is an _event_ received from the outside. So `>> DataReceived` to feed it into your layers. Outgoing data is sent as a command (`<< SendData(...)`). Same with `<< reply()`, that can only come from the outside.

For OpenConnection you need a `reply(None)` (where `None` is the error), or you do `tctx.server = open_h2_server_conn` before the playbook and avoid the OpenConnection altogether (see tests I linked). I'll make a note to document that a reply is expected in the `OpenConnection` docstring, that's currently missing.

Try making a copy of `test_simple` and turn
```python
        << http.HttpRequestHeadersHook(flow)
        >> reply()
```
into
```python
        << http.HttpRequestHeadersHook(flow)
        >> reply(side_effect=enable_streaming)
```
That should get you onto the right track. If not let me know and I'm happy to lend a hand. :)

rubu:
Hi @mhils ,

ok i figured how to make the test to cover both directions (request and resposne), but the response fix was a bit more complex since applying the same logic broke a websocket test - the bytes on the wire were not deframed properly - which seemed logical since your proposed strategy of avoiding the call to `start_response_stream()` vs mine that propagates the end stream does not really work for websockets since i assume the hooks need to be called as early as possible (i assume for web socket frame handlers to be set up?). So i moved the `is_websocket` logical condition to a member method so that it can be reused in the original place and now in the new logic.

Do this PR not seem good or is there something that should be adjusted - I am no a python person so I'm not sure if the `bool | None` thing is proper style or no.

rubu:
Ok, now i actually see the code coverage thing failing due to the extra if:

```
def is_websocket(self) -> bool | None:
        if self.flow.response is None:
            return None
        return (
            self.flow.response.status_code == 101
            and self.flow.response.headers.get("upgrade", "").lower() == "websocket"
            and self.flow.request.headers.get("Sec-WebSocket-Version", "").encode()
            == wsproto.handshake.WEBSOCKET_VERSION
            and self.context.options.websocket
        )
```

What is the proper approach here? Move the self.flow.reponse inside the whole check and just leave the method as bool ? I wanted the `bool | None` return to signify is the result is known.

mhils:
@rubu: Thanks! This looks great so far, thank you for doing the majority of the grunt work with tests etc. I can't give you an immediate answer without digging deeper, let me play around with this in the coming days. :)

mhils (approved):
Thanks again! I ended up adapting the WebSocket test, now the actual code diff is super simple. 🎉

PR #7204, opened 2024-09-22
allow custom content view to refer to the current websocket message

kennytm:
#### Description

Many protocols have different encodings between requests and responses, to properly deserialize the message we need to know whether the message is `from_client`. The current implementation of `get_message_content_view` only forwarded the message to `View.__call__` only if the message is HTTP/TCP/UDP, but the parameter is dropped if it is of any other classes. This PR handled the case when the `message` is a `WebSocketMessage`, so a custom View can now inspect the direction through the `websocket_message` kwarg.

#### Checklist


mhils (approved):
Thanks! 🍰

PR #7228, opened 2024-10-04
Improve HTTP detection heuristic

fatanugraha:
#### Description

Improve the HTTP detection heuristic introduced in this PR: https://github.com/mitmproxy/mitmproxy/pull/7087

In my mitmproxy `transparent` mode setup, [testcontainer's](https://testcontainers.com/getting-started/) ryuk breaks because it sends the string below in raw TCP
```
label=org.testcontainers.version=0.30.0&label=org.testcontainers.sessionId=eb48d24ade2174c56e8951a306241ad68eccdd8fcb7686b62ed03e638ec73870&label=org.testcontainers=true&label=org.testcontainers.lang=go\n
```

Upon checking the proxy log, I found out that mitmproxy thinks that it's an HTTP protocol and try to parse it as such, causing the connection to appear as "stuck"

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## Unreleased: mitmproxy next
 
+- Tighten HTTP detection heuristic to better support custom TCP-based protocols.
 
 ## 02 October 2024: mitmproxy 11.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ## Unreleased: mitmproxy next
 
 - Tighten HTTP detection heuristic to better support custom TCP-based protocols.
+  ([#7228](https://github.com/mitmproxy/mitmproxy/pull/7228), @fatanugraha)
 
 ## 02 October 2024: mitmproxy 11.0.0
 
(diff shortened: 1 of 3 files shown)


mitmproxy/addons/next_layer.py
```diff
@@ -184,6 +184,8 @@ def s(*layers):
             len(data_client) < 3
             # HTTP would require whitespace before the first newline
             # if we have neither whitespace nor a newline, it's also unlikely to be HTTP.
+            or (data_client.find(b" ") == -1)
+            or (data_client.find(b"\n") == -1)
```
mhils, line 188: We can have incomplete request lines in the first packet, so I would rather go for something like this:

```suggestion
            or ("\n" in data_client and " " not in data_client)
```

What do you think?

CHANGELOG.md
```diff
@@ -23,7 +25,7 @@
   ([#6720](https://github.com/mitmproxy/mitmproxy/pull/6720), @NicolaiSoeborg)
 - Fix slow tnetstring parsing in case of very large tnetstring.
   ([#7121](https://github.com/mitmproxy/mitmproxy/pull/7121), @mik1904)
-- Add `getaddrinfo`-based fallback for DNS resolution if we are unable to 
+- Add `getaddrinfo`-based fallback for DNS resolution if we are unable to
```
mhils, line 28: Please don't add unrelated changes here.

mhils (commented):
Thanks!

mitmproxy/addons/next_layer.py
fatanugraha: That makes sense, thanks!

mhils (approved):
Thanks! 🍰

PR #7234, opened 2024-10-10
Remove Apple Silicon note

thonkinator:
#### Description

mitmproxy now supplies native Apple Silicon binaries, even on honebrew. this shouldn't be needed anymore


mhils (approved):
Thanks! 🍰

PR #7239, opened 2024-10-13
Add support for starting a preconfigured Firefox instance via `browser.start` command.

sujaldev:
#### Description

`browser.start` command will now allow for a `browser` parameter defaulting to "chrome". (Refer #5247)

#### Checklist

### DIFF
--- a/mitmproxy/addons/browser.py
+++ b/mitmproxy/addons/browser.py
@@ -2,14 +2,13 @@ import logging
 import shutil
 import subprocess
 import tempfile
-from typing import List
 
 from mitmproxy import command
 from mitmproxy import ctx
 from mitmproxy.log import ALERT
 
 
-def find_executable_cmd(*search_paths) -> List[str] | None:
+def find_executable_cmd(*search_paths) -> list[str] | None:
     for browser in search_paths:
         if shutil.which(browser):
             return [browser]
@@ -17,7 +16,7 @@ def find_executable_cmd(*search_paths) -> List[str] | None:
     return None
 
 
-def find_flatpak_cmd(*search_paths) -> List[str] | None:
+def find_flatpak_cmd(*search_paths) -> list[str] | None:
     if shutil.which("flatpak"):
         for browser in search_paths:
             if (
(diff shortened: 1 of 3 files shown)


mitmproxy/addons/browser.py
```diff
@@ -100,6 +100,59 @@ def start(self) -> None:
...
+            return
+
+        host = ctx.options.listen_host or "127.0.0.1"
+        port = ctx.options.listen_port or 8080
+        prefs = [
+            'user_pref("datareporting.policy.firstRunURL", "");',
+            'user_pref("network.proxy.type", 1);',
+            'user_pref("network.proxy.share_proxy_settings", true);',
+        ]
+        for service in ("http", "ssl", "socks", "ftp"):
```
mhils, line 129: `socks` and `ftp` seem wrong here, no?

mhils (commented):
Thanks for the PR! 🎉 

1. I get this popup here when running `browser.start firefox`:
    image
    Can we skip that somehow?

2. Are there additional Firefox prefs we can configure to disable telemetry spam / update checks / ...?

mitmproxy/addons/browser.py
sujaldev: sure, I'll remove this.

sujaldev:
> I get this popup here when running browser.start firefox Can we skip that somehow?

This is caused by the `--safe-mode` flag, I'll remove it.

> Are there additional Firefox prefs we can configure to disable telemetry spam / update checks / ...?

I have tried experimenting with a lot of preferences that seem to reduce a tiny amount of telemetry, but requests to `*.services.mozilla.com` and a few other domains just can't seem to be turned off, perhaps we could drop requests to matching domains in mitmproxy?

mhils (approved):
Awesome, thanks! 🍰 

Good to go from my end - are you still planning any further tweaks, or shall we merge? ☺️

sujaldev:
Thanks for the review! I don't have any more changes planned, so we're good to merge :)

About filtering the remaining telemetry traffic, I think it would just add complexity without much gain, and most users might actually want to see those requests.

PR #7257, opened 2024-10-17
Arbitrary strings in local mode

lups2000:
#### Description

In this PR, I've added the ability to include executable names that are not part of the current list. The addition occurs when pressing the **Enter** key. This change addresses and closes issue #7063.

https://github.com/user-attachments/assets/1151ab92-f81a-4334-b436-bdd8573377f8

#### Checklist


lups2000:
@mhils This should finally close #7063 :) I noticed we haven't added a changelog entry yet—should we include one?

mhils:
Nice! 🚀 

Instead of "No results", can we have a message that's closer to "Press Enter to capture traffic for programs matching $CURRENT_INPUT"?

lups2000:
> Nice! 🚀
> 
> Instead of "No results", can we have a message that's closer to "Press Enter to capture traffic for programs matching $CURRENT_INPUT"?

yes sure 👍

mhils (approved):
🎉🎉🎉

PR #7286, opened 2024-10-31
Fix interaction of modifyheaders addon with stream_large_bodies option

### LINKED ISSUE #7030: Cannot modify headers on requests with bodies (POST etc) when stream_large_bodies is enabled
#### Problem Description
When `stream_large_bodies` is set to any value above 0, any flow that modifies headers on requests that have bodies will stop working. This is unexpected because while the documentation and code [warns about this](https://github.com/mitmproxy/mitmproxy/blob/460789b7beadebfdf3a99fd50c3c1e6004bc47be/mitmproxy/addons/modifybody.py#L51-L52) when it comes to modifying **bodies** (as expected), it does not warn that modifying **headers** won't work.

In my case, I am adding a header for basic auth and had `stream_large_bodies=0` because I don't need to modify any request or response bodies and this is much better for performance on any page with video, etc. However, I noticed that all POST requests (with a body) stopped working because they were missing the auth header, even though GET requests (without a body) still had it.

1. Set `stream_large_bodies=0`
2. Add a flow like `modify_headers=/~d ^example.com$/authorization/Basic credsgohere`
3. Send both a GET and a POST (with a body) request through the proxy
4. See that the GET request has the header added, the POST request does not

#### System Information
```
Mitmproxy: 10.3.0 binary
Python:    3.12.3
OpenSSL:   OpenSSL 3.2.1 30 Jan 2024
Platform:  Linux-6.1.0-21-amd64-x86_64-with-glibc2.36
```

lukant:
#### Description

`modifyheaders` addon uses `request` and `response` hooks which, when `stream_large_bodies` is enabled, are called after headers have already been sent and can no longer be modified.

This commit changes `modifyheaders` addon to use `requestheaders` and `responseheaders` hooks.

This change aims to fix #7030 

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - `browser.start` command now supports Firefox.
   ([#7239](https://github.com/mitmproxy/mitmproxy/pull/7239), @sujaldev)
 - Fix interaction of modifyheaders addon with stream_large_bodies option
+  ([#7286](https://github.com/mitmproxy/mitmproxy/pull/7286), @lukant)
 
 ## 02 October 2024: mitmproxy 11.0.0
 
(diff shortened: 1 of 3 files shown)


mhils (approved):
Awesome, great fix. Thank you! 🍰

PR #7311, opened 2024-11-08
Update flow.ts and modal.ts ducks to RTK

lups2000:
#### Description

In this PR, I've updated two files, [flow.ts](https://github.com/mitmproxy/mitmproxy/blob/main/web/src/js/ducks/ui/flow.ts) and [modal.ts](https://github.com/mitmproxy/mitmproxy/blob/main/web/src/js/ducks/ui/modal.ts), to use Redux Toolkit.

#### Checklist

### DIFF
--- a/web/src/js/__tests__/ducks/ui/flowSpec.tsx
+++ b/web/src/js/__tests__/ducks/ui/flowSpec.tsx
@@ -1,22 +1,26 @@
-import reduceFlow, * as FlowActions from "../../../ducks/ui/flow";
+import flowReducer, {
+    selectTab,
+    setContentViewFor,
+    defaultState,
+} from "../../../ducks/ui/flow";
 
 describe("option reducer", () => {
     it("should return initial state", () => {
-        expect(reduceFlow(undefined, { type: "other" })).toEqual(
-            FlowActions.defaultState,
-        );
+        expect(flowReducer(undefined, { type: "other" })).toEqual(defaultState);
     });
 
     it("should handle set tab", () => {
         expect(
-            reduceFlow(undefined, FlowActions.selectTab("response")).tab,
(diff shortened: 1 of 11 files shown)


lups2000:
I also have a question about something not directly related to this PR. In [option.ts](https://github.com/mitmproxy/mitmproxy/blob/main/web/src/js/ducks/ui/optionsEditor.ts), if we create a slice and specify reducers, such as `startUpdate(state, action)`, I wanted to specify the type of action, so I added the following:
```
interface OptionUpdate<T> {
    isUpdating: boolean;
    value: T;
    error: string | false;
}

type OptionsEditorState = Partial<{
    [name in Option]: OptionUpdate<OptionsState[name]>;
}>;

const defaultState: OptionsEditorState = {};

const optionsEditorSlice = createSlice({
    name: "ui/optionsEditor",
    initialState: defaultState,
    reducers: {
        startUpdate(state, action: PayloadAction<{ option: Option; value: OptionUpdate<Option> }>) {
            state[action.payload.option] = {
                isUpdating: true,
                value: action.payload.value,
                error: false,
            };
        },
        ...
    },
});
```
However, I encountered this error: `Type '{ isUpdating: boolean; value: any; error: boolean; }' is not assignable to type 'undefined'.ts(2322)` because of the `Partial` type. To address this, I had to change method signatures in `options.ts`, as shown here:

```
export function update(name: Option, value: any): AppThunk {
    return (dispatch) => {
        dispatch(optionsEditorActions.startUpdate(name, value));
        sendUpdate(name, value, dispatch);
    };
}
```

Since these changes were quite a lot, I didn’t include them in this PR. Do you think enforcing TypeScript here makes sense, or could it be too complex to justify? I also noticed that some values in `options.ts` are typed as `any`.

PR #7319, opened 2024-11-12
Multiple flows selection in mitmweb 

lups2000:
#### Description

In this PR, I'm introducing the ability to select multiple flow rows by pressing `Ctrl` (or `Command` on macOS) and clicking on a flow, or by using `Shift + Click`. So far, multi-flow selection is supported for **removing**, **duplicating**, **reverting**, **resuming**, and **aborting** flows. Replay functionality is still missing.

The idea comes from here: #3313 

https://github.com/user-attachments/assets/dcd5bafa-d6ba-4149-8f7b-f285cc730d36

#### Checklist


web/src/js/components/FlowTable.tsx
```diff
@@ -12,7 +12,7 @@ type FlowTableProps = {
     flows: Flow[];
     rowHeight: number;
     highlight: string;
-    selected: Flow;
+    selectedFlows: string[];
```
Prinzhorn, line 15: `selectedFlowIds`?

web/src/js/ducks/flows.ts
```diff
@@ -23,13 +24,15 @@ export interface FlowsState extends store.State<Flow> {
     filter?: string;
     sort: { column?: keyof typeof sortFunctions; desc: boolean };
     selected: string[];
```
Prinzhorn, line 26: I'm confused why here `selected` has always been a `string[]` but in a different part `select` was `Flow` and now became `selectedFlows` as `string[]`?

web/src/js/components/MainView.tsx
```diff
@@ -11,6 +11,9 @@ export default function MainView() {
     const hasSelection = useAppSelector(
         (state) => !!state.flows.byId[state.flows.selected[0]],
     );
+    const isMultipleFlowsSelection = useAppSelector(
```
Prinzhorn, line 14: I don't quite understand why `isMultipleFlowsSelection` exists? In most parts of the UI you use `selectedFlows.length > 1` but then in this one part you use this flag instead. Either consistently track `isMultipleFlowsSelection` or consistently imply it via `selectedFlows.length > 1`. Or am I misunderstanding what's happening?

web/src/js/ducks/flows.ts
```diff
@@ -110,8 +113,15 @@ export default function reducer(
             return {
                 ...state,
                 selected: action.flowIds,
```
Prinzhorn, line 115: Shouldn't this be `selected: [action.flowIds],`?

Prinzhorn (changes requested):

web/src/js/ducks/flows.ts
lups2000: `action.flowIds` is already an array of strings. Using `[action.flowIds]` would create an array containing an array of strings

web/src/js/components/MainView.tsx
lups2000: you're right, having both sounds confusing! I've removed `isMultipleFlowsSelection`

web/src/js/ducks/flows.ts
```diff
@@ -198,6 +204,15 @@ export function remove(flow: Flow) {
     return () => fetchApi(`/flows/${flow.id}`, { method: "DELETE" });
 }
 
+export function removeMultiple(flowIds: string[]) {
```
mhils, line 207: I think the general pattern of adding separate "multi" actions isn't great. Why don't we change the existing `remove` action to always accept a list of flow ids? It's not used in a million places, we can easily adapt existing callsites. :)

web/src/js/ducks/flows.ts
```diff
@@ -241,3 +256,10 @@ export function select(id?: string) {
         flowIds: id ? [id] : [],
     };
 }
+
+export function multiSelect(id: string) {
```
mhils, line 260: Same here.

web/src/js/components/MainView.tsx
```diff
@@ -11,6 +11,9 @@ export default function MainView() {
...
+        (state) => state.flows.selected.length > 1,
+    );
```
mhils, line 16:
```suggestion
    const hasFlowSelected = useAppSelector(
        (state) => state.flows.selected.length === 1 && !!state.flows.byId[state.flows.selected[0]],
    );
```

The primary benefit of this change is that you will need to address the other usage of `hasSelection` and fix it in the process. No need for a splitter if there's no flow view 😉

web/src/js/components/Header/FlowMenu.tsx
```diff
@@ -222,6 +255,7 @@ const markers = {
 
 function MarkButton({ flow }: { flow: Flow }) {
     const dispatch = useAppDispatch();
+    const selectedFlows = useAppSelector((state) => state.flows.selected);
```
mhils, line 258: Add a TODO comment here for multi-selection to indicate intent (it can/should be done, it just hasn't been done yet).

web/src/js/components/Header/FlowMenu.tsx
```diff
@@ -24,6 +26,8 @@ export default function FlowMenu(): JSX.Element {
         (state) => state.flows.byId[state.flows.selected[0]],
     );
 
+    const selectedFlows = useAppSelector((state) => state.flows.selected);
```
mhils, line 29: `useAppSelector` is instrumental for caching, see https://redux.js.org/tutorials/fundamentals/part-5-ui-react#reading-state-from-the-store-with-useselector. If we only use `selectedFlows` for `selectedFlows.length > 1` checks, we should only return the boolean from the selector.

Considering this is used repeatedly, I wouldn't mind a `selectHasSingleFlowSelected` helper, but both with and without helper is fine. :)

mhils (commented):
I think this is super valuable, and it's a good idea to start small here and only support deletions. Left a few comments below, but please also add at least a test to select/deselect a second flow with Ctrl+Click. :)

web/src/js/components/FlowTable/FlowRow.tsx
```diff
@@ -7,22 +7,23 @@ import * as columns from "./FlowColumns";
...
-    selected,
-    highlighted,
+    isSelected,
+    isHighlighted,
 }: FlowRowProps) {
     const dispatch = useAppDispatch();
     const displayColumnNames = useAppSelector(
         (state) => state.options.web_columns,
     );
+    const selectedFlowsIds = useAppSelector((state) => state.flows.selected); // used for multiple flows selection
```
mhils, line 23: Nit:
```suggestion
    const selectedFlowIds = useAppSelector((state) => state.flows.selected); // used for multiple flows selection
```

web/src/js/components/Header/FlowMenu.tsx
```diff
@@ -24,6 +25,13 @@ export default function FlowMenu(): JSX.Element {
...
+
+    const hasSingleFlowSelected = React.useMemo(
+        () => selectedFlows.length === 1,
+        [selectedFlows],
+    );
```
mhils, line 33: I don't think memoization here is helping. We already get the forced rerender from `const selectedFlows = useAppSelector((state) => state.flows.selected);`. Memoizing here isn't going to change this. Two possible ways forward:

1. We decide that performance is good enough here. Then we can just remove the useMemo call and be done with it.
2. We decide that performance is important here. Then we need to get rid of the `useAppSelector((state) => state.flows.selected)` call. In this case we cannot call `removeFlow(selectedFlows)`. Alternatively, we could introduce an `removeSelectedFlows()` action. In this action we would then access the state to get the selected flows and work with that.

Both is fine with me, up to you what makes most sense. :)

web/src/js/components/Header/FlowMenu.tsx
```diff
@@ -108,6 +135,12 @@ const openInNewTab = (url) => {
 };
 
 function DownloadButton({ flow }: { flow: Flow }) {
+    const hasSingleFlowSelected = useAppSelector(
+        (state) => state.flows.selected.length === 1,
+    );
+
+    if (!hasSingleFlowSelected) return null;
```
mhils, line 142: Is there a reason why this button is treated differently (removed and not disabled)?

mhils (commented):
Nice progress! 🚀

web/src/js/components/Header/FlowMenu.tsx
lups2000: no my bad, thanks for spotting this :)

web/src/js/components/Header/FlowMenu.tsx
lups2000: I would keep it simple and remove the **useMemo**. I don't think we will encounter any significant performance issues here. In most cases, users will select a relatively limited number of flows since they have to do it manually. I tested it with ~50 flows, and it runs smoothly. Does that make sense to you?

mhils: SGTM

lups2000: perfect :)

web/src/js/components/FlowTable.tsx
```diff
@@ -60,15 +60,19 @@ export class PureFlowTable extends React.Component<
...
             const { rowHeight, flows, selected } = this.props;
             const viewport = this.viewport.current!;
             const head = this.head.current;
 
             const headHeight = head ? head.offsetHeight : 0;
 
-            const rowTop = flows.indexOf(selected) * rowHeight + headHeight;
+            const selectedFlowIndex = flows.findIndex(
+                (flow) => flow.id === selected[0],
+            );
```
mhils, line 74: Perf: The store maintains a flowId -> index mapping:
https://github.com/mitmproxy/mitmproxy/blob/6269898b434e1040c4ea46a425526c5d266c554b/web/src/js/ducks/utils/store.ts#L25

mhils (approved):
This looks fantastic now, thanks! One small perf optimization, feel free to merge yourself after. 😃

web/src/js/components/Header/FlowMenu.tsx
```diff
@@ -56,10 +68,14 @@ export default function FlowMenu(): JSX.Element {
                         <Button
                             title="[d]elete flow"
                             icon="fa-trash text-danger"
-                            onClick={() => dispatch(removeFlow(flow))}
+                            onClick={() => {
+                                dispatch(removeFlow(selectedFlows));
+                                dispatch(selectFlow([])); // clear flow selection
```
lups2000, line 73: I noticed a small bug: previously, when a user removed a flow, the next flow was automatically selected. Now, this no longer happens. It would be nice to keep that behaviour, what do you think?
This line seems to be the issue. I remember adding it because, without it, multiple deletions caused this API call:

`http://127.0.0.1:8081/flows/e394b6f1-7425-40ec-9387-c72d949feee6/request/content/Auto.json?lines=513
`

to fail with a **404 Not Found**. This happens because the flow ID is selected as the next flow even though it has already been deleted. After this failed call, we correctly call the endpoint again with the actual next flow, and everything works fine.
Now, I need to figure out how to solve that 😄

mhils: Great catch! This one is important. Folks coming from the TUI will repeatedly press `d` to get rid of flows. 😄

lups2000: okay this is solved now

web/src/js/components/FlowTable.tsx
lups2000: Regarding this, I replaced that line with **listIndex**, and it worked fine. However, during testing, I encountered the following error:

"_Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops_"

I had to add this statement 
```
if (!shallowEqual(this.props.flowIndexes, prevProps.flowIndexes)) {
      this.onViewportUpdate();
}
```
which solved the issue but it introduces some inefficiency again. Do you have any idea?

lups2000:
Looks good to me @mhils . Maybe we can update the changelog entry to be more accurate.

lups2000:
Screenshot 2025-04-25 at 09 06 46

when selecting a flow

lups2000:
@mhils I have fixed a couple of issues:
1. A crash that occurred when selecting a flow (see image above).
2. An error when removing multiple flows; same issue as described here: https://github.com/mitmproxy/mitmproxy/pull/7319#discussion_r2001960571
3. Previously, it was sometimes impossible to clear all flows, because removing the last one would throw: `Cannot read property 'id' of undefined`

 would you mind taking a final look ? thank you 😄

web/src/js/ducks/flows.ts
```diff
@@ -65,15 +65,18 @@ function updateSelected(
                 break;
             }
             if (selected.length > 1) {
-                selected = selected.filter((f) => f.id !== action.data);
+                // clear selection when multiple flows are selected and removed
+                selected = [];
```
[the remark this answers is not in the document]
lups2000: If we enter this branch, it means we want to remove more than one element, and after removing the flows, the selection is cleared. How is it possible that only one of the selected flows gets removed and not everything?
Without this change, if you try to remove multiple flows we get a 404 Not found exception.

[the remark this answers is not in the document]
mhils: > If we enter this branch, it means we want to remove more than one element

This is not true AFAIK. `REMOVE` is triggered individually for every flow in the backend.

> Without this change, if you try to remove multiple flows we get a 404 Not found exception.

Only in the devtools, not user visible, correct?

This is a race: If we have two selected flows (A and B) and click remove, that's currently two requests. Both will reach the backend at about the same time, then for both we get a REMOVE event. After the first REMOVE (of A) the selection switches to B only. The frontend then tries to get the view for B, but B has already been removed in the backend (the frontend hasn't learned about this yet).

One way to fix this race is to batch the `POST /remove` requests and also the resulting event. But honestly I'm fine with how things are at the moment. Fundamentally, user A can always select flow X which has just been deleted in the backend (which will trigger a 404). Those are harmless IMO. :)

lups2000:
I've implemented marking for multiple flows as well. The only thing missing now is replay, but we can leave that for another PR; this one’s already big enough 😅. All good?

web/src/js/ducks/flows.ts
```diff
@@ -22,17 +24,67 @@ export interface FlowsState extends store.State<Flow> {
...
     selected: [],
+    selectedIndex: {},
     ...store.defaultState,
 };
 
+function updateSelected(
+    state: FlowsState = defaultState,
+    newStoreState: State<Flow>,
+    action,
+): Pick<FlowsState, "selected" | "selectedIndex"> {
```
mhils, line 44: We need to fix the lack of test coverage here, otherwise this PR is good to go. 👍

lups2000: I added some tests to cover those lines. Let me know :)

mhils (approved):
🚀🚀🚀

jeromesegura:
> I've implemented marking for multiple flows as well. The only thing missing now is replay, but we can leave that for another PR; this one’s already big enough 😅. All good?

Love the new feature! mitmweb has a lot of potential as a full client for those that prefer this kind of user experience. Right click support on individual flows would be amazing ;-)

mhils:
What would you like to see in a right-click menu? 

FWIW, one of the next things on the mitmweb front I would like to tackle is https://github.com/mitmproxy/mitmproxy/issues/7694.

(Needless to say: contributions are welcome! ;-))

jeromesegura:
> What would you like to see in a right-click menu?
> 
> FWIW, one of the next things on the mitmweb front I would like to tackle is #7694.
> 
> (Needless to say: contributions are welcome! ;-))

I will reply in that thread with more details, thanks!

Issue #6881, opened 2024-05-30
mac m1 | protobuf<3.19,>=3.14| only 3 can be used at least 20


zsq2010:
protobuf<3.19,>=3.14

but as follows:

---------------------------------------
==> Updating Homebrew...
Already up-to-date.
(base) ➜  mitm brew search protobuf

==> Formulae
protobuf        protobuf-c      protobuf@21     protobuf@3 ✔    swift-protobuf

==> Casks

/////////////////////////Can only get 3.20 ,which is protobuf@3 ✔   ////////////////////////////////////////////////////////////////
(base) ➜  mitm brew install protobuf@3.18                          

==> Downloading https://formulae.brew.sh/api/formula.jws.json
##O#- #                                                                        
==> Downloading https://formulae.brew.sh/api/cask.jws.json
##O#- #                                                                        
Warning: No available formula with the name "protobuf@3.18". Did you mean protobuf@3 or protobuf@21?
==> Searching for similarly named formulae and casks...
==> Formulae
protobuf@3                               protobuf@21

To install protobuf@3, run:
  brew install protobuf@3
(base) ➜  mitm brew install protobuf@3

==> Downloading https://ghcr.io/v2/homebrew/core/protobuf/3/manifests/3.20.3-4
######################################################################### 100.0%
==> Fetching protobuf@3
==> Downloading https://ghcr.io/v2/homebrew/core/protobuf/3/blobs/sha256:38970c2
######################################################################### 100.0%
==> Pouring protobuf@3--3.20.3.arm64_sonoma.bottle.4.tar.gz
==> Caveats
protobuf@3 is keg-only, which means it was not symlinked into /opt/homebrew,
because this is an alternate version of another formula.

If you need to have protobuf@3 first in your PATH, run:
  echo 'export PATH="/opt/homebrew/opt/protobuf@3/bin:$PATH"' >> ~/.zshrc

mhils:
Can you please rephrase what exactly the issue is? You cannot install mitmproxy?

zsq2010:
when running a script,for example,

> Can you please rephrase what exactly the issue is? You cannot install mitmproxy?

<img width="759" alt="图片" src="https://github.com/mitmproxy/mitmproxy/assets/4374659/f8775721-d592-480f-a028-15194eb281f9">

zsq2010:
brew on mac,can only find protobuf3 (which is 3.20 after install) 
there is no protobuf3x which version is lower than 3.20

[...]

PR #4349, opened 2020-12-13
nuke deprecated components

Kriechi:
#### Description

closes #4273 

* removes pathod and pathoc
* removes legacy protocol stack

ToDo: fix tests that are still relying on pathod and not to-be-removed with the legacy stack.

#### Checklist

 - [x] I have updated tests where applicable.
 - [X] I have added an entry to the CHANGELOG.

PR #7327, opened 2024-11-17
Remove dependency on protobuf library

matthew16550:
#### Description

Hello, this is my first contribution to mitmproxy.

The `protobuf` dependency seems unused now, I think the need went away in #4349.

All the tests pass on my Mac without protobuf and hopefully CI will find if there is a need I overlooked.

If the dependency can be removed it will simplify my attempt at updating MacPorts which currently installs v7.0.4.

Also it would help some older issues - #5160, #6881

#### Checklist

 - [n/a] I have updated tests where applicable.


mhils (approved):
Oh wow, thanks! 🍰

PR #6033, opened 2023-03-29
Button to copy formatted `ContentView`

Rishabhg71:
#### Description
Added Copy button in the HTTP content view that will copy formatted data to the clipboard
image

closes #5730 
#### Checklist

 - [x] I have updated tests where applicable.
 - [x] I have added an entry to the CHANGELOG.

Issue #5730, opened 2022-11-13
copy raw request data button mitmweb


Mattwmaster58:
#### Problem Description
Often, I find myself copying and pasting request data. My current mitmweb workflow is `View -> raw`, then selecting all, then copying

#### Proposal
I would like a button that simply copies the raw response. I think that raw response is the most useful here and also the easiest to implement. Example of what this could look like:
![image](https://user-images.githubusercontent.com/26337069/201510773-caec4612-04f3-4997-bcd6-374f642cd6bd.png)

#### Alternatives
2 copy buttons, one for raw, and one for formatted. This would be more complex, ~~and more importantly my workflow doesn't depend on this~~ (eg, "sending" images to the clipboard) and not as useful as a raw copy option which makes this not worth it in my opinion.

I don't think it would be that hard for me to write this PR, but I wanted to post here for feedback first.

RG7279805:
i will take this issue

RG7279805:
@Mattwmaster58 do you mind giving your opinion on the PR

lups2000:
Hi,
this has been addressed in #7352 😃

PR #7352, opened 2024-11-26
Copy request/response data to clipboard

lups2000:
#### Description

In this PR, I've added the ability to copy request or response data to the clipboard. There was a previous PR (#6033), but it didn't account for the **Show More** button. If the content was long and copied, only part of it would be copied.

Solves: #5730.

Demo:

https://github.com/user-attachments/assets/5d8dcf9f-d524-4062-b511-e9a908fe6082

#### Checklist


lups2000:
I will add the tests asap :)

web/src/js/components/contentviews/HttpMessage.tsx
```diff
@@ -60,6 +63,28 @@ export default function HttpMessage({ flow, message }: HttpMessageProps) {
...
+                    throw new Error(
+                        `${response.status} ${response.statusText}`.trim(),
+                    );
+                }
+                return response.json();
+            })
+            .then((data: ContentViewData) => {
+                setIsCopied(true);
+                copyViewContentDataToClipboard(data);
+                setTimeout(() => setIsCopied(false), 10000);
```
Prinzhorn, line 81: 10 seconds feels a little excessive for a short user feedback. Does it reset when switching to a different flow? Maybe it's enough if it flashes `Copied!` for a short time?

Depending on if the component is re-used for multiple flows then `setTimeout` also needs to be cleared otherwise if I copied multiple flows in sequence you have multiple `setTimeout` racing and `isCopied` loses any meaning.

I'd remove the `disabled={isCopied}` part. There's no reason to limit the user in this way. If I want to copy the content again, let me.

Prinzhorn: What you could do is `disabled={isFetching}`, so that when a big response is fetched the user can not click the button multiple times during the fetch (because there might not be immediate user feedback that the click did something).

> I didn't want the user to abuse it and make too many unnecessary calls

I just checked and mitmproxy serves the requests with ETag. If the user click the button again after the first request was made, the content should come from the browser cache. Can you verify this with the copy button as well?

lups2000: Yeah, there is an ETag, but it doesn't look like the response is cached. It seems that a new request is being made. The request header includes: `cache-control: no-cache`

Prinzhorn: You probably have cache disabled in dev tools network tab?

lups2000: yes @Prinzhorn, you are right! I forgot to enable caching. Everything works as expected then :)

lups2000:
It should be fine now! I removed the disabling of the copy button. There’s just a brief (1 sec) animation to provide feedback to the user after clicking the button. Also, @mhils, if you have any suggestions or thoughts, feel free to share!

Prinzhorn (changes requested):
I think adding a loading state via `isFetching` (`disabled` + `cursor: wait`) would be a nice addition. Especially if you host mitmweb on a remote host there will be latency and no immediate user feedback.

Try slowing down your network e.g.:

image

and click the button.

lups2000:
> I think adding a loading state via `isFetching` (`disabled` + `cursor: wait`) would be a nice addition. Especially if you host mitmweb on a remote host there will be latency and no immediate user feedback.
> 
> Try slowing down your network e.g.:
> 
> image
> 
> and click the button.

thanks for the feedback! Yes makes sense! I will add it :)

web/src/js/components/contentviews/HttpMessage.tsx
```diff
@@ -60,6 +65,29 @@ export default function HttpMessage({ flow, message }: HttpMessageProps) {
         }
     }, [content]);
 
+    const handleClickCopyButton = () => {
+        const url = MessageUtils.getContentURL(flow, message, contentView); //without the 'maxLines' parameter, so we can get the full content of the content view
+        setIsFetchingFullContent(true);
+        setIsCopied(true);
```
Prinzhorn, line 71: That line should probably come right before the setTimeout, when you've actually copied the data to the clipboard

lups2000: I refactored it a bit and also hid the button in case there is no content

web/src/js/components/contentviews/HttpMessage.tsx
```diff
@@ -32,7 +33,11 @@ export default function HttpMessage({ flow, message }: HttpMessageProps) {
         [maxLines],
     );
     const [edit, setEdit] = useState<boolean>(false);
-    let url;
+    const [isCopied, setIsCopied] = useState<boolean>(false);
```
Prinzhorn, line 36: I don't mean to be super nit-picky. Just wondering if there's any naming convention in the rest of the project. `edit` vs `isEditing` and `copied` vs `isCopied`.

lups2000: Good point! Normally, if it's a boolean value, we use the prefix `is`. For this reason, I've adjusted the `edit` value to `isEdited`

lups2000:
@Prinzhorn all good now?

web/src/js/components/contentviews/HttpMessage.tsx
```diff
@@ -60,11 +65,38 @@ export default function HttpMessage({ flow, message }: HttpMessageProps) {
...
+                }
+                return response.json();
+            })
+            .then((data: ContentViewData) => {
+                copyViewContentDataToClipboard(data);
+                setIsCopied(true);
+                setTimeout(() => setIsCopied(false), 2000);
+            })
+            .catch((e) => {
+                console.error(e);
```
Prinzhorn, line 88: This PR seems like a great place for me to learn about mitmweb :smile: . Sorry for delaying this PR.

I wonder if there is currently any error handling in similar places? In this case something like an error toast would be great. Or are there other places were we also simply `console.error`? The user should probably get some sort of feedback right?

lups2000: It depends, actually. In some cases, we display error messages as red text (e.g., on the capture page and in the options modal). However, in other cases like this one, displaying an error message is tricky because, to be honest, I’m not sure where it would fit best for an optimal UX. Using toasts could be a good alternative, and I think we should discuss whether we want to adopt toasts consistently across mitmweb. For now, to keep things simple, I’m just logging the error to the console.

lups2000:
@Prinzhorn this has been open for a while now. Could we merge it? I think we've addressed everything!

Prinzhorn (approved):

lups2000:
@mhils all good also by your side?

lups2000:
@Prinzhorn I cannot merge it, can you?

Prinzhorn:
@lups2000 I don't think so. I have only this option:

image

and I would assume it won't auto-marge as long as there are failing checks.

lups2000:
For some strange reason, these lines aren't covered by the clicking event 🤯 

`await copyViewContentDataToClipboard(data);`
`setIsCopied(true);`
`setTimeout(() => setIsCopied(false), 2000);`

lups2000:
Okay now it should be ready @mhils, I've improved test coverage and I don't think I can do better than this 😅

mhils (approved):
🚀🚀🚀

PR #7372, opened 2024-12-03
Fix Hugo deprecation error

aphedges:
I encountered this problem because the version of Hugo installed with Homebrew is much newer than the one in CI. I fixed the error and updated Hugo in CI so the build keeps working.

#### Checklist

	- Does a documentation build fix like this require a CHANGELOG entry?


mhils (approved):
Thanks! 🍰

PR #7385, opened 2024-12-09
allow single click to edit URL request

lups2000:
#### Description

In this PR, I’m introducing the ability to directly start editing an input field with a single click, without resetting the input content. The default behavior is:

1. Clicking once selects the entire text.
2. Typing resets the text.

This behavior was particularly frustrating for URLs. With this change, the user can click once and immediately start editing from the selected position.

https://github.com/user-attachments/assets/4e1baa80-bab5-4ab2-8687-9b008906e1b8

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 
 - Add cache-busting for mitmweb's front end code.
   ([#7386](https://github.com/mitmproxy/mitmproxy/pull/7386), @mhils)
-- Allow single-click editing of URL requests in mitmweb.
+- Clicking the URL in mitmweb now places the cursor at the current position instead of selecting the entire URL.
   ([#7385](https://github.com/mitmproxy/mitmproxy/pull/7385), @lups2000)
 
 ## 05 December 2024: mitmproxy 11.0.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@
 - Allow single-click editing of URL requests in mitmweb.
   ([#7385](https://github.com/mitmproxy/mitmproxy/pull/7385), @lups2000)
 
-
 ## 05 December 2024: mitmproxy 11.0.2
 
 - Stop sorting keys in JSON contentview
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -6,7 +6,8 @@
 -->
 
 ## Unreleased: mitmproxy next
-
+- Allow single-click editing of URL requests in mitmweb.
```
mhils, line 9:
```suggestion
- Clicking the URL in mitmweb now places the cursor at the current position instead of selecting the entire URL.
```

web/src/js/components/editors/ValueEditor.tsx
```diff
@@ -9,6 +9,7 @@ export interface ValueEditorProps {
     onInput?: (newVal: string) => void;
     onKeyDown?: (e: React.KeyboardEvent<HTMLSpanElement>) => void;
     placeholder?: string;
+    setCursorOnClick?: boolean;
```
mhils, line 12: I'm wondering if this would be easier to understand if we'd invert it, e.g. calling it `selectAllOnClick`. WDYT?

(changing the few places where we use it shouldn't be a big deal)

mhils (commented):
Awesome! This is a great UX improvement. 🎉

web/src/js/components/editors/ValueEditor.tsx
lups2000: Yes, we can invert that if you want. I noticed that the `ValueEditor.tsx` component is used in many files, and its current behavior across all of them is to select the entire text on click. Where do we actually need this **"entire text selection"** feature? If it's only required in a few places, we can definitely invert the default behavior of this component.

lups2000: I removed the `selectAllOnClick` feature from the capture tab because it also includes hosts, URLs, and paths, ad it was a bit weird. What do you think?

mhils (approved):
🎉

PR #7386, opened 2024-12-09
Implement Cache-Busting for mitmweb

### LINKED ISSUE #7381: Implement Cache-Busting for mitmweb
#### Problem Description
mitmweb's web interface is currently served from a static path, and browsers like to cache that. We should add some cache busting [here](https://github.com/mitmproxy/mitmproxy/blob/e0c40b1beb8f9fc2460962aa89c231cec50b10d3/mitmproxy/tools/web/templates/index.html#L10). A simple implementation would be to add `?hash` to the URL, and calculate the actual file hash once on demand.

1. Use old mitmweb version.
2. Update mitmproxy.
3. Open web interface again: Version number in the status bar has updated (comes from the backend), but functionality has not.

#### System Information

```
Mitmproxy: 12.0.0.dev (+20, commit d51df5f)
Python:    3.12.8
OpenSSL:   OpenSSL 3.3.2 3 Sep 2024
Platform:  macOS-15.1.1-arm64-arm-64bit
```

mhils:
#### Description

Turns out this is a super easy fix because tornado has builtin functionality. 🎉 

fixes #7381

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## Unreleased: mitmproxy next
 
+- Add cache-busting for mitmweb's front end code.
 
 ## 05 December 2024: mitmproxy 11.0.2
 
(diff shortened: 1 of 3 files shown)


Prinzhorn (approved):
That was easy :smile:, nice :broccoli: !

PR #7390, opened 2024-12-11
Exposing 'localModeUnavailable' to the frontend

lups2000:
#### Description

In this PR, I'm exposing `localModeUnavailable` to the frontend. This allows us to display a string explaining why local mode is unavailable on unsupported platforms.

#### Checklist

### DIFF
--- a/web/src/js/components/Modes.tsx
+++ b/web/src/js/components/Modes.tsx
@@ -30,7 +30,10 @@ export default function Modes() {
                     ) : (
                         <MissingMode
                             title="Local Redirect Mode"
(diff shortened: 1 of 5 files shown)


web/src/js/components/Modes.tsx
```diff
@@ -27,7 +30,10 @@ export default function Modes() {
                     ) : (
```
mhils, line 30: Remove the check above and check whether localModeUnavailable is set instead?

lups2000: Adjusted! The `platform` doesn't disappear, though, since we need it for **transparent mode**.

mhils (approved):

PR #7397, opened 2024-12-16
Add script to enable use of pac server

angelbot:
#### Description

I created this script for use at work and thought I'd contribute it to the examples in case anyone else would find it useful. The script uses pypac to evaluate requests to determine if they should go to a configured upstream proxy or not. In the case of a pac evaluation of direct, it also allows you to configure a direct proxy allowing you to use 2 proxies if needed.

#### Checklist


mhils (approved):
Thanks! 🍰

PR #7421, opened 2024-12-29
Add 'updateFilter' method to the current websocket connection

lups2000:
#### Description

This is the first PR for our new feature.
In this PR, I’ve added a new method `updateFilter` to the current WebSocket connection. This allows the client to update the filter and notify the server.

DEMO:

https://github.com/user-attachments/assets/e9c06fc0-69a5-4baf-9e6f-4b87ca3b3f08

#### Checklist


web/src/js/app.tsx
```diff
@@ -5,19 +5,12 @@ import { Provider } from "react-redux";
...
 
 useUrlState(store);
-// @ts-expect-error custom property on window
-if (window.MITMWEB_STATIC) {
-    // @ts-expect-error new property on window for debugging
-    window.backend = new StaticBackend(store);
-} else {
-    // @ts-expect-error new property on window for debugging
-    window.backend = new WebSocketBackend(store);
-}
```
mhils, line 20: This already was a global static instance / singleton, no? I don't quite understand what problem we are solving here.

We can turn this into `export const backend = ...` followed by `window.backend = backen`. Then you can just import `backend`. No need for Java-isms. :)

web/src/js/backends/websocket.tsx
```diff
@@ -94,6 +95,19 @@ export default class WebsocketBackend {
         }
     }
 
+    sendMessage(resource, data) {
```
mhils, line 98: Maybe we should add a dedicated `updateFilter(name, expression)` here? This way the actual WebSocket message format is contained in the WebSocket backend and does not leak into the rest of the codebase.

web/src/js/components/Header/FilterInput.tsx
```diff
@@ -2,14 +2,16 @@ import React, { Component } from "react";
...
+interface FilterInputProps {
     type: string;
-    color: any;
+    color: string;
     placeholder: string;
     value: string;
-    onChange: (value) => void;
-};
+    onChange: (value: string) => void;
+}
```
mhils, line 14: Thank you for improving type hints! 🍰 

Feel free to submit them as a separate micro PR so that we get them out of the critical path.

web/src/js/app.tsx
lups2000: Yes, the functionality remains unchanged; it's still a singleton. The idea was to remove the dependency on the `window` object since I didn't like doing something like `window.backend.sendMessage(...)`. I hadn't thought about exporting it! thanks for the tip :)

web/src/js/backends/websocket.tsx
lups2000: okay yes makes sense! Let me know what do you think about this solution!

web/src/js/components/Header/FilterInput.tsx
lups2000: sure! Done :)

web/src/js/backends/websocket.tsx
```diff
@@ -94,6 +94,27 @@ export default class WebsocketBackend {
...
+    updateFilter(name: string, expr: string) {
+        this.sendMessage("flows", {
+            cmd: "updateFilter",
+            name,
+            expr,
+        });
+    }
+
+    private sendMessage(resource, data) {
+        if (this.socket && this.socket.readyState === WebSocket.OPEN) {
```
mhils, line 106: What about `WebSocket.CONNECTING`? We probably should collect messages during connection establishment and then send them in onOpen. The filter state from the URL will come very early, likely before a connection has been established.

mhils (commented):
Thanks for splitting out the typing changes! The general design in `backends/websocket.tsx` look good now, but I don't think we should tie the backend update to the UI component. This should much rather happen in https://github.com/mitmproxy/mitmproxy/blob/18a29df7d07117d11b5d7019f20f169733e6abd3/web/src/js/ducks/flows.ts#L153C17-L159. One example why this is useful is that we would otherwise ignore filter/highlight state from the URL.

web/src/js/backends/websocket.tsx
lups2000: can't we do something like this?
`if (
            this.socket &&
            (this.socket.readyState === WebSocket.CONNECTING ||
                this.socket.readyState === WebSocket.OPEN)
        ) {...}`

lups2000:
> Thanks for splitting out the typing changes! The general design in `backends/websocket.tsx` look good now, but I don't think we should tie the backend update to the UI component. This should much rather happen in https://github.com/mitmproxy/mitmproxy/blob/18a29df7d07117d11b5d7019f20f169733e6abd3/web/src/js/ducks/flows.ts#L153C17-L159. One example why this is useful is that we would otherwise ignore filter/highlight state from the URL.

Oh true! yes definitely

web/src/js/backends/websocket.tsx
mhils: Can you already send on a socket while it is still in CONNECTING state? If that works that's great, but I don't know. :)

lups2000: If I do something like this: 
`if (this.socket && this.socket.readyState === WebSocket.OPEN) {...}` and then refresh or change the url, I get an error saying that I cannot send messages because the socket is not open; while if I do something like this:
`if ( this.socket && (this.socket.readyState === WebSocket.CONNECTING || this.socket.readyState === WebSocket.OPEN) ) {...}`, it works fine 😅

lups2000:
I managed to fix the circular dependency, which was caused by `import { backend } from "./app"` in `ducks/flows.ts`. Initially, I wanted to avoid directly accessing `window.backend`, but I couldn’t find a way to resolve the issue while using the `backend` import. Maybe you have some ideas. That said, `window.backend `doesn’t seem so bad now 😅, especially since I’ve explicitly defined the following:
`declare global {
    interface Window {
        MITMWEB_STATIC?: boolean;
        backend: WebSocketBackend | StaticBackend;
    }
}`

web/src/js/ducks/flows.ts
```diff
@@ -151,10 +152,16 @@ export function makeSort({
 }
 
 export function setFilter(filter: string) {
+    if (window.backend instanceof WebsocketBackend) {
```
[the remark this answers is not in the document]
lups2000: okay! So, in that case we'll have the updateFilter function with simply an empty body?

web/src/js/components/Header/FilterInput.tsx
```diff
@@ -3,8 +3,14 @@ import classnames from "classnames";
 import Filt from "../../filt/filt";
 import FilterDocs from "./FilterDocs";
 
+export enum FILTER_TYPE {
```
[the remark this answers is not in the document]
lups2000: may I change the attribute name from `type` to `icon` ? 😅

web/src/js/components/Header/FilterInput.tsx
```diff
@@ -102,7 +108,7 @@ export default class FilterInput extends Component<
             // If closed using ESC/ENTER, hide the tooltip.
             this.setState({ mousefocus: false });
         }
-        e.stopPropagation();
+        e.stopPropagation?.();
```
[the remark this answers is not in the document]
lups2000: this is due to: `Partial<React.KeyboardEvent<HTMLInputElement>>`. I remove it now, since it creates confusion!

web/src/js/ducks/flows.ts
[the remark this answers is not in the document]
lups2000: by doing so:
`export function setFilter(filter: string) {
    window.backend.updateFilter("search", filter);
    return { type: SET_FILTER, filter };
}`, it works fine; but a lot of tests fail saying: ` TypeError: Cannot read properties of undefined (reading 'updateFilter')`. I managed to solve the tests by doing `window.backend?.updateFilter("search", filter)` but this doesn't make much sense to me. Why would we need that ? 🤔

web/src/js/components/Header/FilterInput.tsx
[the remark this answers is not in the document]
mhils: Yes, absolutely! 👍

web/src/js/ducks/flows.ts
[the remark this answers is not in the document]
mhils: I pushed a commit that shows how to resolve the cyclic dependency :)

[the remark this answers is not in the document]
lups2000: it looks like the cyclic dependency is still there 🤯 because of this `import { backend } from "../backends"` in `ducks/flows.ts`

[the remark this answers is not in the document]
lups2000: I've really no idea why with the **backend** variable, which is cleaner, it doesn't work...yeah there is the import but I cannot find a way to solve it.
this works:

`
export function setFilter(filter: string) { 
    if (window.backend) {
        window.backend.updateFilter("search", filter);
    }
    return { type: SET_FILTER, filter };
}
`
but in this case we'd remove completely the backend variable.

lups2000:
this should work fine now :)

mhils (approved):

PR #7440, opened 2025-01-03
Require Python 3.12+

mhils:
#### Description

All recommended installation methods (https://docs.mitmproxy.org/stable/overview-installation/) ship with Python 3.13 already, we shouldn't drag us down with supporting ancient configurations.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@
 ## Unreleased: mitmproxy next
 
 - **Local Capture Mode** is now available on Linux as well.
+  ([#7440](https://github.com/mitmproxy/mitmproxy/pull/7440), @mhils)
 - mitmproxy now requires Python 3.10 or above.
+  ([#7440](https://github.com/mitmproxy/mitmproxy/pull/7440), @mhils)
 - Add cache-busting for mitmweb's front end code.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 
 ## Unreleased: mitmproxy next
 
+- **Local Capture Mode** is now available on Linux as well.
+  ([#7440](https://github.com/mitmproxy/mitmproxy/pull/7440), @mhils)
+- mitmproxy now requires Python 3.10 or above.
```
jllopezpino, line 12: @mhils shouldn't it be "Python 3.12 or above" ?

mhils: Oops, great catch. Thanks! 🍰

grickit:
Python 3.11 and 3.10 are still in official security support.

mitmproxy 11.0.2 requires a version of cryptography with a known vulnerability.

And now users cannot patch that if they use your library without first upgrading one or two major python releases. This was premature IMO.

mhils:
I'm sorry this is causing you trouble, but as a project owner one of my main objectives is to reduce maintenance costs for the project itself. All recommended installation methods (https://docs.mitmproxy.org/stable/overview/installation/) are already on Python 3.13.

PR #7445, opened 2025-01-06
Prompt user before clearing screen

### LINKED ISSUE #7413: Ask confirmation before clearing flow list (key z)
#### Problem Description
The `z` keybinding, used to clear flow list, doesn't ask for confirmation, immediately and irreversibly deleting all flows.
It's not impossible to accidentally hit the z key and lose all the flows after a long debugging session.

#### Proposal
Similar to the exit key (`q` or `esc`), it would be nice to add an action confirmation for `z` keybinding to prevent accidental clearing of the flow list.

errorxyz:
#### Description
Fixes #7413 
Refs #7282

~- Added prompts for destructive shortcuts~
~- Mapped the no-prompts version to `ctrl+[key]`~

~Yet to add tests and update [KeyMapConfig.load_path](https://github.com/mitmproxy/mitmproxy/blob/6e4cb235fdc7d0f55b8edcda37e9f7cb0c8fda22/mitmproxy/tools/console/keymap.py#L209), wanted to get feedback before proceeding further.~

Prompts user before clearing screen

#### Checklist

### DIFF
--- a/mitmproxy/tools/console/consoleaddons.py
+++ b/mitmproxy/tools/console/consoleaddons.py
@@ -301,7 +301,7 @@ class ConsoleAddon:
             try:
                 self.master.commands.call_strings(cmd, args)
             except exceptions.CommandError as e:
-                logger.error(str(e))
+                logger.exception(str(e))
 
         self.master.prompt_for_user_choice(prompt, callback)
 
--- a/mitmproxy/tools/console/consoleaddons.py
+++ b/mitmproxy/tools/console/consoleaddons.py
@@ -294,6 +294,7 @@ class ConsoleAddon:
         """
         Prompt the user before running the specified command.
         """
+
         def callback(opt):
             if opt == "n":
                 return
--- a/mitmproxy/tools/console/consoleaddons.py
+++ b/mitmproxy/tools/console/consoleaddons.py
@@ -284,6 +284,26 @@ class ConsoleAddon:
             quoted += " "
         signals.status_prompt_command.send(partial=quoted)
 
+    @command.command("console.command.confirm")
+    def console_command_confirm(
+        self,
+        prompt: str,
+        cmd: mitmproxy.types.Cmd,
+        *args: mitmproxy.types.CmdArgs,
+    ) -> None:
+        """
+        Prompt the user before running the specified command.
+        """
+        def callback(opt):
+            if opt == "n":
+                return
+            try:
+                self.master.commands.call_strings(cmd, args)
+            except exceptions.CommandError as e:
+                logger.error(str(e))
+
+        self.master.prompt_for_user_choice(prompt, callback)
+
     @command.command("console.command.set")
     def console_command_set(self, option_name: str) -> None:
         """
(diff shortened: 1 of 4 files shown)


errorxyz:
I've taken a different approach at: https://github.com/mitmproxy/mitmproxy/compare/main...errorxyz:mitmproxy:prompt, where I've added a command to prompt the user. Let me know your thoughts on which do you feel is better.

mhils:
https://github.com/mitmproxy/mitmproxy/compare/main...errorxyz:mitmproxy:prompt looks much more reasonable to me. IMO we should only add this for `z`/`Z` for now (which clears everything), for deleting individual rows it can get a bit too much in the way. But open to other opinions here.

errorxyz:
> [main...errorxyz:mitmproxy:prompt](https://github.com/mitmproxy/mitmproxy/compare/main...errorxyz:mitmproxy:prompt) looks much more reasonable to me. IMO we should only add this for `z`/`Z` for now (which clears everything), for deleting individual rows it can get a bit too much in the way. But open to other opinions here.

I don't like that approach since it 
1) changes the original command if `prompt` is set to `True` 
2) `console.command.confirm` has no use by itself: users will not be typing this command in the UI and it would only be used for keymaps
3) users might be confused between setting `prompt=True` and using `console.command.confirm`

If you still feel https://github.com/mitmproxy/mitmproxy/compare/main...errorxyz:mitmproxy:prompt is better, I'll open another PR with that branch :)

mhils:
> 1. changes the original command if `prompt` is set to `True`

Yes, let's remove the `prompt` argument for `add()`.

```python
    km.add("z", "view.flows.remove @all", ["flowlist"], "Clear flow list", True)
```
will become
```python
    km.add("z", "console.command.confirm \"Delete all flows?\" \
        view.flows.remove @all", ["flowlist"], "Clear flow list")
```
but that's perfectly fine.

> 2. `console.command.confirm` has no use by itself: users will not be typing this command in the UI and it would only be used for keymaps

I think that's fine - there are more commands like this (`view.properties.inbounds`, ...).

> 3. users might be confused between setting `prompt=True` and using `console.command.confirm`

See first point. :)

Alternatively, we could also turn `view.flows.remove` into two commands: `view.flows.remove` and `view.flows.remove.noconfirm`. But I like that less to be honest. Honestly I think the approach you took there is quite nice to leverage our scripting language. :)

mitmproxy/tools/console/consoleaddons.py
```diff
@@ -284,6 +284,27 @@ def console_command(self, *command_str: str) -> None:
...
+        Prompt the user before running the specified command.
+        """
+
+        def callback(opt):
+            if opt == "n":
+                return
+            try:
+                self.master.commands.call_strings(cmd, args)
+            except exceptions.CommandError as e:
+                logger.error(str(e))
```
mhils, line 304:
```suggestion
                logger.exception(str(e))
```

mhils (commented):
Super nice! LGTM % 1 nit

PR #7458, opened 2025-01-07
Make all filter expressions case-insensitive by default

### LINKED ISSUE #7457: Streamline filter expression case-sensitivity
## Problem description

This is an idea proposed by @mhils in #7450 to fix filter expression inconsistencies when it comes to `re`'s case-sensitivity. At the time of writing `~u`, `~m` and `~d` use `re.IGNORECASE`  flag (see [flowfilter.py](https://github.com/mitmproxy/mitmproxy/blob/v11.0.2/mitmproxy/flowfilter.py)). Other filters like `~hq` or `~bq` don't use that which effectively translates to the following:

- both `~u example.com` and `~u EXAMPLE.COM` work exactly the same (the regex is treated as case-**insensitive**)
- `~hq Host:` and `~hq host:` are case-**sensitive** - a request with `Host: example.org` header matches to the first filter, but not to the second one

It's my gut feeling rather than data-backed fact, but in vast majority of cases case-insensitive matching is expected.

## Proposal

All relevant filter operators should use case-insensitive matching by default to make things consistent.

1. Identify for which filter operators default `re.IGNORECASE` flag makes sense and apply the change accordingly
2. Implement a global option (e.g. an env variable or config parameter) that disables the `re.IGNORECASE` flag, so case-sensitive filter expressions can be used (e.g. _"I'd like to block access to `example.com/FOO`, but leave `example.com/foo` unaffected"_)
3. Add relevant test cases to cover new behaviour

## Alternatives

Instead of global "disable case-insensitive matching" flag it'd be great to control case-sensitivity per each instance filter expression. Most likely, that'd imply a change of filter expression format which sounds like a breaking change to me.

## Additional context

- original discussion: #7450
- according to [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html) header names are case-insensitive and should be treated as such. Nevertheless, client can send them in upper- or mixed-case
- as stated in [RFC 7540](https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2) (which describes HTTP/2): _"(...) header field names MUST be converted to lowercase prior to their encoding in HTTP/2. A request or response containing uppercase header field names MUST be treated as malformed"_. Header lowercasing is not mandatory for HTTP/1.1 clients though.

mhils:
#### Description

fixes #7457, tests not adjusted yet.

#### Checklist

### DIFF
--- a/mitmproxy/flowfilter.py
+++ b/mitmproxy/flowfilter.py
@@ -34,6 +34,7 @@ Headers:
 """
 
 import functools
+import os
 import re
 import sys
 from collections.abc import Sequence
@@ -47,6 +48,10 @@ from mitmproxy import flow
 from mitmproxy import http
 from mitmproxy import tcp
 from mitmproxy import udp
+
+maybe_ignore_case = (
+    re.IGNORECASE if os.environ.get("MITMPROXY_CASE_SENSITIVE_FILTERS") != "1" else 0
+)
 
 
 def only(*types):
@@ -180,7 +185,7 @@ class _Rex(_Action):
         if self.is_binary:
             expr = expr.encode()
         try:
-            self.re = re.compile(expr, self.flags)
+            self.re = re.compile(expr, self.flags | maybe_ignore_case)
         except Exception:
             raise ValueError("Cannot compile expression.")
 
@@ -379,7 +384,6 @@ class FBodResponse(_Rex):
 class FMethod(_Rex):
     code = "m"
     help = "Method"
-    flags = re.IGNORECASE
 
     @only(http.HTTPFlow)
     def __call__(self, f):
@@ -389,7 +393,6 @@ class FMethod(_Rex):
 class FDomain(_Rex):
     code = "d"
     help = "Domain"
-    flags = re.IGNORECASE
     is_binary = False
 
     @only(http.HTTPFlow)
@@ -403,7 +406,6 @@ class FUrl(_Rex):
     code = "u"
     help = "URL"
     is_binary = False
-    flags = re.IGNORECASE
 
     # FUrl is special, because it can be "naked".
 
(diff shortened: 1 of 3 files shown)


mitmproxy/flowfilter.py
```diff
@@ -432,7 +434,7 @@ def __call__(self, f):
         if not f.client_conn or not f.client_conn.peername:
             return False
         r = f"{f.client_conn.peername[0]}:{f.client_conn.peername[1]}"
-        return f.client_conn.peername and self.re.search(r)
```
lups2000, line 435: is there a reason why you removed this? Just wondering

lups2000 (commented):
Looks good to me! Ping me when you'll add the tests :)

mitmproxy/flowfilter.py
mhils: That check is already being made three lines up. :)

lups2000: oh true 😅

AdityaPatadiya:
Hello @mhils , how it will check for the case-sensitivity for code that you've changed?

mhils:
@AdityaPatadiya: I don't understand your question, please rephrase. :)

AdityaPatadiya:
Field was there `re.IGNORECASE` before but now after removing, how it will know that it needs to check whether the requests are case-sensitive or not even after setting MITMPROXY_CASE_SENSITIVE_FILTERS=1.

AdityaPatadiya:
actually I don't understand that how you keep `re.IGNORECASE` as global?
and how it will work?
like at every request, it will gonna check and apply `re.IGNORECASE` but now how it will point out to that?
I was trying to modified or add tests according to the updated code.

mhils:
I'm sorry, but we don't have the capacity here to help you with programming questions. Let's please keep this PR focused on getting the actual change in eventually.

lups2000:
I think this can be closed, right?

emanuele-em (approved):

Issue #7452, opened 2025-01-07
Failing to parse compressed (gzipped) responses


s9807:
`Addon error: EOFError when decoding b'\x1f\x8b with 'gzip': EOFError('Compressed file ended before the end-of-stream marker was reached')
Traceback (most recent call last):
  File "/home/subodh/API-FUZZER/local/rapifuzz-backend/venv/lib/python3.11/site-packages/mitmproxy/net/encoding.py", line 62, in decode
    decoded = custom_decode[encoding](encoded)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/subodh/API-FUZZER/local/rapifuzz-backend/venv/lib/python3.11/site-packages/mitmproxy/net/encoding.py", line 151, in decode_gzip
    return f.read()
           ^^^^^^^^
  File "/usr/lib/python3.11/gzip.py", line 301, in read
    return self._buffer.read(size)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/_compression.py", line 118, in readall
    while data := self.read(sys.maxsize):
                  ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/gzip.py", line 518, in read
    raise EOFError("Compressed file ended before the "
EOFError: Compressed file ended before the end-of-stream marker was reached

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/subodh/API-FUZZER/local/rapifuzz-backend/venv/lib/python3.11/site-packages/mitmproxy/addons/savehar.py", line 128, in done
    self.export_har(self.flows, ctx.options.hardump)
  File "/home/subodh/API-FUZZER/local/rapifuzz-backend/venv/lib/python3.11/site-packages/mitmproxy/command.py", line 322, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/subodh/API-FUZZER/local/rapifuzz-backend/venv/lib/python3.11/site-packages/mitmproxy/addons/savehar.py", line 39, in export_har
    har = json.dumps(self.make_har(flows), indent=4).encode()
                     ^^^^^^^^^^^^^^^^^^^^
  File "/home/subodh/API-FUZZER/local/rapifuzz-backend/venv/lib/python3.11/site-packages/mitmproxy/addons/savehar.py", line 58, in make_har
[...]

PR #7460, opened 2025-01-08
Failing to parse compressed (gzipped) responses

s9807:
#### Checklist


mhils (changes requested):
I don't agree with the changes here, `decode_gzip` should continue to raise for invalid input and not silently swallow errors.

Assuming this is meant to fix https://github.com/mitmproxy/mitmproxy/issues/7452: Raising `.content` needs to be manually handled in the har addon, and even then we shouldn't throw the contents away.

PR #7479, opened 2025-01-15
Fix reading har with empty response content

dmm:
#### Description

I exported a mitmproxy flow file to a har file but discovered that mitmproxy couldn't read the resulting har. The problem appears to be a response with an empty content. The empty byte string gets passed to http.encoding.encode but it can't encode it, resulting in an unhandled exception.

I added a har test file that reproduces the problem and a simple fix.

#### Checklist

### DIFF
--- a/mitmproxy/io/har.py
+++ b/mitmproxy/io/har.py
@@ -8,6 +8,7 @@ from datetime import datetime
 from mitmproxy import connection
 from mitmproxy import exceptions
 from mitmproxy import http
+from mitmproxy.net import encoding
 from mitmproxy.net.http.headers import infer_content_encoding
 
 logger = logging.getLogger(__name__)
@@ -35,6 +36,15 @@ def fix_headers(
         flow_headers.append((key.encode(), value.encode()))
 
     return http.Headers(flow_headers)
+
+def fix_content_encoding(content_encoding: str | None) -> str:
+    """
+    Converts invalid content-encoding values to "indentity" while leaving valid values unchanged.
+    """
+    if not content_encoding or content_encoding not in encoding.custom_encode:
+        return "identity"
+
+    return content_encoding
 
 
 def request_to_flow(request_json: dict) -> http.HTTPFlow:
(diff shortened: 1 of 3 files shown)


mhils (changes requested):
Thanks for the PR! The problem here isn't `b""`, the problem here is an invalid `Content-Encoding` header. Compare https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding with what you have in the example.

Long story short, we should fall back gracefully for all invalid content-encodings, not just for empty bodies.

dmm:
Thanks for the feedback! I've renamed the test files to and updated the fix to instead check the provided "content-encoding" value to those supported by "mitmproxy.net.encoding", otherwise falling back to "identity". 

I'm investigating the failing tests (test_har_to_flow[safari] and test_har_to_flow[firefox]) but they seem to be failing for me on main, as well?

mitmproxy/io/har.py
```diff
@@ -109,7 +120,7 @@ def request_to_flow(request_json: dict) -> http.HTTPFlow:
 
     # Then encode the content, as in `Response.set_content`
     response_content = http.encoding.encode(
-        response_content, response_headers.get("content-encoding") or "identity"
+        response_content, fix_content_encoding(response_headers.get("content_encoding"))
     )
```
mhils, line 124: I'd like us to be a bit more Pythonic here and ask for forgiveness rather than for permission. Can we just do something like this please?

```python3
    try:
        # Then encode the content, as in `Response.set_content`
        response_content = http.encoding.encode(
            response_content, response_headers.get("content-encoding") or "identity"
        )
    except ValueError:
        logger.warn(f"HAR import: Ignoring invalid content-encoding in request to {request_url!r}")
```

mhils (commented):
> I'm investigating the failing tests (test_har_to_flow[safari] and test_har_to_flow[firefox]) but they seem to be failing for me on main, as well?

CI is passing, so this needs more details. My only suspicion would be different line endings on Windows vs Linux, but that's just a guess.

Issue #5077, opened 2022-01-18
Small bug in handling of streamed data from callable flow.*.stream


EndUser509:
I have finished writing an addon to save streamed requests and responses to files. While getting it to work I stumbled over the problem that some sites fail to load with: 

```
HTTP/2 connection closed: <ConnectionTerminated error_code:ErrorCodes.FRAME_SIZE_ERROR, last_stream_id:1, additional_data:None>
```

The issue is that the callable `flow.request.stream` and `flow.response.stream` at the end of the stream is supposed to return the zero byte string it receives which is then sent as an HTTP/2 chunk. For responses this may be tolerated by the client but for requests it breaks protocol.

I fixed this issue with this patch which I _think_ is probably the best way to fix it given the contract with the callable:

```diff
diff --git a/mitmproxy/proxy/layers/http/__init__.py b/mitmproxy/proxy/layers/http/__init__.py
index 0b134f7bb..9068e9020 100644
--- a/mitmproxy/proxy/layers/http/__init__.py
+++ b/mitmproxy/proxy/layers/http/__init__.py
@@ -244,10 +244,14 @@ class HttpStream(layer.Layer):
         elif isinstance(event, RequestEndOfMessage):
             if callable(self.flow.request.stream):
                 chunks = self.flow.request.stream(b"")
+                send = True
                 if isinstance(chunks, bytes):
+                    if len(chunks) == 0:
+                        send = False
                     chunks = [chunks]
-                for chunk in chunks:
-                    yield SendHttp(RequestData(self.stream_id, chunk), self.context.server)
+                if send:
+                    for chunk in chunks:
+                        yield SendHttp(RequestData(self.stream_id, chunk), self.context.server)
 
             self.flow.request.timestamp_end = time.time()
             yield HttpRequestHook(self.flow)
@@ -336,10 +340,14 @@ class HttpStream(layer.Layer):
         elif isinstance(event, ResponseEndOfMessage):
             if callable(self.flow.response.stream):
                 chunks = self.flow.response.stream(b"")
+                send = True
                 if isinstance(chunks, bytes):
+                    if len(chunks) == 0:
+                        send = False
                     chunks = [chunks]
-                for chunk in chunks:
-                    yield SendHttp(ResponseData(self.stream_id, chunk), self.context.client)
+                if send:
+                    for chunk in chunks:
+                        yield SendHttp(ResponseData(self.stream_id, chunk), self.context.client)
[...]

PR #7480, opened 2025-01-16
don't send empty data frame without end-of-stream flag

Dieken:
Due to CVE-2019-9518[1], HTTP/2 server and client may reject empty data frame without end-of-stream
flag with GOAWAY frame and FRAME_SIZE_ERROR.
    
This patch completes the fix at [2] for events RequestData and ResponseData too.
    
[1]. https://nvd.nist.gov/vuln/detail/CVE-2019-9518
[2]. https://github.com/mitmproxy/mitmproxy/issues/5077

### DIFF
--- a/mitmproxy/proxy/layers/http/__init__.py
+++ b/mitmproxy/proxy/layers/http/__init__.py
@@ -321,21 +321,23 @@ class HttpStream(layer.Layer):
             else:
                 chunks = [event.data]
             for chunk in chunks:
-                yield SendHttp(RequestData(self.stream_id, chunk), self.context.server)
+                if chunk:  # empty data frame without End Stream flag may be rejected due to CVE-2019-9518
+                    yield SendHttp(
+                        RequestData(self.stream_id, chunk), self.context.server
+                    )
         elif isinstance(event, RequestTrailers):
             # we don't do anything further here, we wait for RequestEndOfMessage first to trigger the request hook.
             self.flow.request.trailers = event.trailers
         elif isinstance(event, RequestEndOfMessage):
             if callable(self.flow.request.stream):
                 chunks = self.flow.request.stream(b"")
-                if chunks == b"":
-                    chunks = []
-                elif isinstance(chunks, bytes):
+                if isinstance(chunks, bytes):
                     chunks = [chunks]
                 for chunk in chunks:
-                    yield SendHttp(
-                        RequestData(self.stream_id, chunk), self.context.server
-                    )
+                    if chunk:  # empty data frame without End Stream flag may be rejected due to CVE-2019-9518
+                        yield SendHttp(
+                            RequestData(self.stream_id, chunk), self.context.server
+                        )
 
             self.flow.request.timestamp_end = time.time()
             yield HttpRequestHook(self.flow)
@@ -444,21 +446,23 @@ class HttpStream(layer.Layer):
             else:
                 chunks = [event.data]
             for chunk in chunks:
-                yield SendHttp(ResponseData(self.stream_id, chunk), self.context.client)
+                if chunk:  # empty data frame without End Stream flag may be rejected due to CVE-2019-9518
+                    yield SendHttp(
+                        ResponseData(self.stream_id, chunk), self.context.client
+                    )
         elif isinstance(event, ResponseTrailers):
             self.flow.response.trailers = event.trailers
             # will be sent in send_response() after the response hook.
         elif isinstance(event, ResponseEndOfMessage):
             if callable(self.flow.response.stream):
                 chunks = self.flow.response.stream(b"")
-                if chunks == b"":
-                    chunks = []
-                elif isinstance(chunks, bytes):
+                if isinstance(chunks, bytes):
                     chunks = [chunks]
                 for chunk in chunks:
-                    yield SendHttp(
-                        ResponseData(self.stream_id, chunk), self.context.client
-                    )
+                    if chunk:  # empty data frame without End Stream flag may be rejected due to CVE-2019-9518
+                        yield SendHttp(
+                            ResponseData(self.stream_id, chunk), self.context.client
+                        )
             yield from self.send_response(already_streamed=True)
 
     @expect(ResponseData, ResponseTrailers, ResponseEndOfMessage)


mhils:
Thank you for your PR! Much appreciated. 😃 

> The original code breaks HTTP/2 protocol, the HTTP/2 server will return GOAWAY: FRAME_SIZE_ERROR.

Generally speaking, sending empty DATA frames is valid HTTP/2. This is a server bug.

That being said: If the original client is not sending an empty data frame, and mitmproxy adds an empty data frame, I'd consider that to be a bug. On the other hand, if a client _does_ send empty data frames, we shouldn't be eating them if we can avoid it. This creates a really annoying "client fails without mitmproxy, but when I add mitmproxy as a debugger, it suddenly works" situation.

Long story short, can we please add one simple test that shows incorrect behavior without your patch applied, but correct behavior with your patch? You can probably do something similar to [this one](https://github.com/mitmproxy/mitmproxy/blob/main/test/mitmproxy/proxy/layers/http/test_http2.py#L95-L121). Thanks! 🍰

Dieken:
I'm unfamiliar with the test framework, will check it later,  let me show the packets captured with WireShark,   it's mitmproxy that adds an extra empty data frame.

I use mitmproxy to audit traffic to cursor.sh, the API backend of famous AI assisted code editor Cursor,  cursor.sh always reports error on empty data frame,  I think they must use mature HTTP/2 infrastructure, so I think empty data frame for HTTP/2 is invalid, as described similarly at #5077.

image

image

image

image

The network topology: 

Cursor editor  -----gateway(192.168.180.1)----> mitmproxy(192.168.180.169) ---> api<N>.cursor.sh(35.71.162.24)

Packet flow:

(1) Cursor editor <-> mitmproxy:

44.  PRI * HTTP/2.0,    empty settings frame,  non-empty headers frame
46.  empty data frame with End Stream flag
50.  non-empty settings frame,  non-empty window_update frame
52.  empty settings frame with ACK flag
53.  empty settings frame with ACK flag

(2) mitmproxy -> cursor.sh:

68.  PRI * HTTP/2.0, non-empty settings frame, non-empty window_update frame
70.  non-empty headers frame
72.  empty data frame without flag     <----  bug here
74.  empty data frame with End Stream fag

(3) mitmproxy <-> cursor.sh

76.  non-empty settings frame, non-empty window_update frame
77.  empty settings frame with ACK flag
79.  empty settings frame with ACK flag
80.  GOAWAY frame with Error: FRAME_SIZE_ERROR (6)

Dieken:
Although the caller side of `state_stream_request_body()` and `state_stream_response_body()` can remove the additional empty data frame without flag, it's still good to filter out empty data for the stream callback function as defensive programming, people probably forgets empty data frame without flag will break HTTP/2.0 server implementation.

mhils:
> I'm unfamiliar with the test framework, will check it later, let me show the packets captured with WireShark, it's mitmproxy that adds an extra empty data frame.

Ok, that's clearly a bug then. If the test framework doesn't make sense to you, feel free to send me a PCAP (email is on my profile) and I will take a look.

Dieken:
The original packets contains sensitive HTTP headers, so I write a nodejs script to recur, and also attach the pcapng file.

a.js
```javascript
//const https = require("https");
const http2 = require("http2");

(() => {
  try {
    const tls = require("tls");
    const fs = require("fs");
    const os = require("os");
    const content = fs.readFileSync(os.homedir() + "/Library/Application Support/Cursor/cacerts.pem.1", "utf8");
    const extraCerts = content.split(/(?=-----BEGIN CERTIFICATE-----)/g).filter(pem => !!pem.length);

    const createSecureContext = tls.createSecureContext;
    tls.createSecureContext = function (...args) {
      const options = args[0] || {};

      if (options.ca) {
        if (Array.isArray(options.ca)) {
          options.ca = [...options.ca, ...extraCerts];
        } else {
          options.ca = [options.ca, ...extraCerts];
        }
      } else {
        options.ca = [...tls.rootCertificates, ...extraCerts];
      }

      return createSecureContext.call(this, options);
    };
  } catch (err) {
    console.log("Failed to inject tls.createSecureContext() with ~/Library/Application Support/Cursor/cacerts.pem: ", err);
  }
})();

/*
https.get('https://api3.cursor.sh/', (res) => {
  console.log('statusCode:', res.statusCode);
  console.log('headers:', res.headers);

  res.on('data', (d) => {
    process.stdout.write(d);
  });

}).on('error', (e) => {
  console.error(e);
});
*/

const client = http2.connect('https://api3.cursor.sh');

client.on('error', (err) => console.error(err));

const req = client.request({ ':path': '/' }, {
  endStream: false,
});

req.on('response', (headers, flags) => {
  console.log(headers);
  console.log(flags);
});

let data = '';
req.on('data', (chunk) => { data += chunk; });
req.on('end', () => {
  console.log(data);
  client.close();
});

req.setEncoding('utf8');
req.write('');
req.end();         // explicitly create an empty data frame with EndStream flag
```

The script can request api3.cursor.sh successfully without mitmproxy, but with mitmproxy it will get:
```
<html>
<head>
    <title>502 Bad Gateway</title>
</head>
<body>
    <h1>502 Bad Gateway</h1>
    <p>HTTP/2 connection closed: &lt;ConnectionTerminated error_code:6, last_stream_id:1, additional_data:None&gt;</p>
</body>
</html>
```

[example.tar.gz](https://github.com/user-attachments/files/18463942/example.tar.gz)

mhils:
Your script isn't easily reproducible for me (I don't have `"/Library/Application Support/Cursor`). Can you send me a PCAP with a failing attempt?

Dieken:
> Your script isn't easily reproducible for me (I don't have `"/Library/Application Support/Cursor`). Can you send me a PCAP with a failing attempt?

I attached the pcap file at the end of the comment above, the tar.gz file contains sslkeylogfile.txt as described at https://docs.mitmproxy.org/stable/howto-wireshark-tls/

You may adjust the path to `os.homedir() + ".mitmproxy/ca.crt"`,  it's the CA certificate of mitmproxy.

Dieken:
Hi @mhils , any progress?

Dieken:
@mhils , ping,  any more info needed?

mhils:
@Dieken: No more info needed I think, I just didn't find time to look into this yet. :)

Issue #7392, opened 2024-12-12
Don't put "run.ps1" on PATH


laundmo:
#### Problem Description

Currently, the Windows installer puts "run.ps1" in `Program Files/bin` which is a directory added to the PATH environment variable.
This leads to 2 issues:
- It will take precedence over all other executables or scripts named "run" defined later in PATH.
- It will take precedence over all functions/cmdlets named "run" defined in unloaded PowerShell modules, preventing them from being automatically loaded. Specifically, this causes inconsistent behaviour when a different command from the same module causes the module to be automatically loaded, in which case the function/cmdlet takes precedence afterwards.

For these reasons, it's best to clutter PATH.

Probably caused by this:
https://github.com/mitmproxy/mitmproxy/blob/e0434726f38394bcb336baf57672f275d4feec15/release/installbuilder/mitmproxy.xml#L27-L37

#### Proposal
Move `run.ps1` to a non-PATH directory.

#### Alternatives
Rely on some other way of solving the problem, `run.ps1` solves without needing a script.
Rename `run.ps1` to another name unlikely to cause conflicts like `run_helper_mitmproxy.ps1`

#### Additional context
Regarding the automatic loading precedence:
Assume I have a module "mytestmodule" which contains the commands: `run` and `other`
```ps1
run # calls run.ps1 from mitmproxy
other # auto-loads mytestmodule calls mytestmodule\other
run # calls mytestmodule\run
```

mhils:
Agreed that this is not ideal and renaming the file is the way to go (`run-mitmproxy` maybe?). Contributions welcome!

AdityaPatadiya:
Hello, @mhils , @laundmo can you give more clarification on this?

laundmo:
> Hello, @mhils , @laundmo can you give more clarification on this?

Please ask for concrete information, i do not know what part you want clarification on and i tried to be quite clear in the issue.

AdityaPatadiya:
we can change the file name
but what else we should do?

laundmo:
that's probably the most plausible way forward. It would probably be nicer if it wasn't needed at all but I can't judge that.

I already outlined the 3 solutions I could think of in my issue:
- rename
- move
- remove

PR #7481, opened 2025-01-16
fix run.ps1 being in %PATH%

meitinger:
#### Description

Proposed fix for #7392. Moves `run.ps1` out of `bin` folder to not clutter `%PATH%`.

#### Checklist


mhils (approved):
Thanks! 🎉

PR #7483, opened 2025-01-16
fixes #7457: streamline filter expression case-sensitivity

### LINKED ISSUE #7457: Streamline filter expression case-sensitivity

AdityaPatadiya:
#### Description

added tests for issue #7457 

#### Checklist


AdityaPatadiya:
Hello @mhils , please review this pr and let me know if there are any changes to make.

AdityaPatadiya:
Hello @mhils, I just need to know that, is there anything missing in this PR?
or you're busy to review my PR.
need your confirmation.

mhils (changes requested):
Please stop pinging me repeatedly. I appreciate a friendly reminder if it's a week or so, but not more than one.

Three problems with your PR:

1. Please base it on my commit.
2. Please use https://docs.pytest.org/en/stable/reference/reference.html#pytest.MonkeyPatch.setenv, we don't want tests to have side effects.
3. A single test for `MITMPROXY_CASE_SENSITIVE_FILTERS` would be more efficient than testing it for every filter.

AdityaPatadiya:
I've done with the changes.

mhils:
Can you please base your changes onto my commit? Thanks!

mhils (commented):
Thanks! This looks great now. Can we please move the case sensitivity checks into a dedicated `test_case_sensitive` test and leave test_method as-is?

Also let's remove the comment on line 62 in `test/mitmproxy/addons/test_blocklist.py`. The previous behavior was stupid and deserved a callout, this is reasonable now.

AdityaPatadiya:
Changes are done now!
Let me know if anything is missing.
Please give feedback for future purpose regarding to all changes and processes that I've done.
Thanks!

AdityaPatadiya:
Something is happen but I don't get it.
Because there is something error while committing the changes.

AdityaPatadiya:
Changes are done!

mhils (approved):
Awesome, thanks! 🍰 😃

PR #7491, opened 2025-01-17
Web: harden `xsrf_token` usage

mhils:
#### Description

This PR adds more defense-in-depth mechanisms for `xsrf_token`. Namely, we make the cookie HttpOnly and expose it to the web app as part of the HTML template. This should make it impossible for other pages on the same origin to obtain the cookie. Finally, we guard the web app HTML against access from `fetch`/`XMLHttpRequest` by checking the `Sec-Fetch-Mode` header.

Updating `mitmproxy/tools/web/static/app.js` as well to keep the web app working at all on main.

#### Checklist


mitmproxy/tools/web/app.py
```diff
@@ -718,7 +725,8 @@ def __init__(
             template_path=os.path.join(os.path.dirname(__file__), "templates"),
             static_path=os.path.join(os.path.dirname(__file__), "static"),
             xsrf_cookies=True,
-            cookie_secret=os.urandom(256),
+            xsrf_cookie_kwargs=dict(httponly=True, samesite="Strict"),
```
errorxyz, line 728: If the xsrf_token is sent as part of the HTML template, why are we setting their cookie preferences?

mhils: This is how Tornado's XSRF protection works: Requests need to have matching `Cookie` header and `X-XSRFToken` headers. https://www.tornadoweb.org/en/stable/guide/security.html#cross-site-request-forgery-protection has details.

errorxyz (approved):

PR #7506, opened 2025-01-26
Prototype FiltersManager server side 

lups2000:
#### Description

I'm opening this draft PR to share the current progress on the server logic for handling filters in mitmweb. Since I'm not very experienced with the server codebase or Python, I'd appreciate any tips or feedback!  

So far, I've implemented the data structure to store the filters, as well as the function to listen for messages from the client, update the filters, and send back the flow IDs that match the current filters. This has been briefly tested and it works!

#### Checklist


mitmproxy/tools/web/master.py
```diff
@@ -51,6 +52,7 @@ def __init__(self, opts: options.Options, with_termlog: bool = True):
         self.app = app.Application(self, self.options.web_debug)
         self.proxyserver: Proxyserver = self.addons.get("proxyserver")
         self.proxyserver.servers.changed.connect(self._sig_servers_changed)
+        self.filters_manager = FiltersManager(self.view)
```
lups2000, line 55: this is useless now, but ideally it will be helpful when we evaluate the filters on a single flow (add or update)

mitmproxy/tools/web/app.py
```diff
@@ -290,13 +292,44 @@ def get(self):
...
     connections: ClassVar[set[WebSocketEventBroadcaster]]
+    application: Application  # reference to the application, in order to be able to update the filters
 
     def open(self, *args, **kwargs):
         self.connections.add(self)
 
     def on_close(self):
         self.connections.discard(self)
 
+    async def on_message(self, message: str):
```
mhils, line 303: Let's move this onto `ClientConnection` class, `WebSocketEventBroadcaster` is just meant to be a generic base class to fan out events to multiple connected clients.

mitmproxy/tools/web/app.py
```diff
@@ -290,13 +292,44 @@ def get(self):
...
+        try:
+            data = json.loads(message)
+
+            # Validate the required fields
+            if not all(key in data for key in ["resource", "cmd", "name", "expr"]):
+                print(f"Invalid message format: Missing required fields.")
+                return
```
mhils, line 310: Nit: https://docs.python.org/3.10/glossary.html#term-EAFP
```suggestion
```

mitmproxy/tools/web/app.py
```diff
@@ -290,13 +292,44 @@ def get(self):
...
+
+            resource = data["resource"]
+            command = data["cmd"]
+            name = data["name"]
+            expression = data["expr"]
+
+            if resource == "flows" and command == "updateFilter":
+                filters_manager = self.application.master.filters_manager
```
mhils, line 318: Sharing one filter manager across all client connections is a bad idea. For example, two different clients may use two different "search" filters at the same time. These would collide right now.


I think it makes most sense to track filter state as part of the `ClientConnection` instance. ClientConnection needs dedicated methods like `broadcast_flow_update(f: Flow)` so that it can just apply the filter.

mitmproxy/tools/web/app.py
```diff
@@ -290,13 +292,44 @@ def get(self):
...
+
+                filters_manager.update_filter(name, expression)
+                print(f"Updated filters: {filters_manager.get_all_filters()}")
+                self.broadcast(
+                    resource="flows",
+                    cmd="filtersUpdated",
+                    data=filters_manager.get_matching_flow_ids(),
```
mhils, line 325: Just to be on the same page: `filtersUpdated` needs to contain the following:

1. The name of the filter that was just updated.
2. The current value of that filter.
3.  The list of flows that match that particular filter expression at the moment.

It does not need to send over the filter state of other existing filters.

mitmproxy/tools/web/filters.py
```diff
@@ -0,0 +1,37 @@
+from typing import List
+
+from mitmproxy import flowfilter
+
+
+class FiltersManager:
+    def __init__(self, view):
+        self.filters: dict[str, str] = {
+            "search": "",
+            "highlight": "",
```
mhils, line 10: Coding style tips:

1. Don't treat "search" and "highlight" special on the back end side. Work with arbitrary filter names.
2. If filters are inactive/empty, drop them from the list. Only keep active filters in the list.

mitmproxy/tools/web/app.py
lups2000: Good point! I also removed the broadcasting function when sending the matching flow IDs; otherwise, every client connection would have received updates meant for other clients, making it impossible to maintain different search filters for each client

mitmproxy/tools/web/app.py
```diff
@@ -324,6 +327,59 @@ def broadcast(cls, **kwargs):
...
+
+        message = json.dumps(
+            {
+                "resource": "flows",
+                "cmd": "filtersUpdated",
+                "name": name,
+                "expr": expr,
+                "data": matching_flow_ids,
+            },
+            ensure_ascii=False,
```
mhils, line 357:
```suggestion
```

mitmproxy/tools/web/app.py
```diff
@@ -324,6 +327,59 @@ def broadcast(cls, **kwargs):
 class ClientConnection(WebSocketEventBroadcaster):
     connections: ClassVar[set] = set()
 
+    def __init__(self, application: Application, request, **kwargs):
+        super().__init__(application, request, **kwargs)
+        self.application = application
+        self.filters: dict[str, str] = {}  # Instance-level filters
```
mhils, line 333:
```suggestion
        self.filters: dict[str, TFilter] = {}  # Instance-level filters
```

(this way we don't have to re-parse all the time, and we indirectly ensure that we only get valid filter exprs in there)

mhils (commented):
Architecturally this looks great now 👍👍👍

lups2000:
This seems to work, but I don't like the implementation. I couldn't find a way to implement `broadcast_flow(cmd, flow)` so that it can be called in `master.py` as `app.Connection.broadcast_flow(cmd, flow)` while still allowing access to the filters that refer to an instance of `ClientConnection`. Is there another way to achieve this? Right now, `master.py` feels confusing.

mitmproxy/tools/web/app.py
```diff
@@ -324,6 +327,72 @@ def broadcast(cls, **kwargs):
 class ClientConnection(WebSocketEventBroadcaster):
     connections: ClassVar[set] = set()
 
+    def __init__(self, application: Application, request, **kwargs):
+        super().__init__(application, request, **kwargs)
+        self.application: Application = application
```
mhils, line 332: Why are we assigning this again here? Can we just do a type annotation only?

mitmproxy/tools/web/app.py
```diff
@@ -324,6 +327,72 @@ def broadcast(cls, **kwargs):
...
+            name: str = data["name"]
+            expr: str = data["expr"]
+
+            if resource == "flows" and command == "updateFilter":
+                if expr:
+                    self.filters[name] = flowfilter.parse(expr)
+                else:
+                    del self.filters[name]
+
+                self.send_matching_flow_ids(name, expr)
```
mhils, line 390: If you move the first four lines into `send_matching_flow_ids` (we can then rename it to `update_filter`), and then also inline `get_matching_flow_ids`, you can reduce the number of conditionals and jumps:

```python3
def update_filter(self, name: str, expr: str) -> None:
    if expr:
        filt = flowfilter.parse(expr)
        self.filters[name] = filt
        matching_flow_ids = [f.id for f in self.application.master.view if filt(f)]
    else:
        del self.filters[name]
        matching_flow_ids = []
    
    message = json.dumps( 
    # ...
```

mitmproxy/tools/web/app.py
```diff
@@ -324,6 +327,72 @@ def broadcast(cls, **kwargs):
...
+
+        message = json.dumps(
+            {
+                "resource": "flows",
+                "cmd": "filtersUpdated",
+                "name": name,
+                "expr": expr,
+                "data": matching_flow_ids,
+            },
+        ).encode("utf8", "surrogateescape")
```
mhils, line 369:
```suggestion
        ).encode()
```

We should never have surrogate escapes here, so let's keep it simple.

mitmproxy/tools/web/app.py
```diff
@@ -324,6 +327,72 @@ def broadcast(cls, **kwargs):
...
+    def __init__(self, application: Application, request, **kwargs):
+        super().__init__(application, request, **kwargs)
+        self.application: Application = application
+        self.filters: dict[str, flowfilter.TFilter] = {}  # Instance-level filters
+
+    def send_flow(self, cmd: str, f: mitmproxy.flow.Flow):
+        matches = [
+            {expr.pattern: bool(flowfilter.match(expr, f))}
+            for expr in self.filters.values()
+        ]
```
mhils, line 339:
```suggestion
        matches = {
            expr.pattern: expr(f)
            for expr in self.filters.values()
        }
```

If `bool()` is necessary, then the type annotation for TFilter is incorrect and we should fix that.

mitmproxy/tools/web/app.py
```diff
@@ -324,6 +327,72 @@ def broadcast(cls, **kwargs):
 class ClientConnection(WebSocketEventBroadcaster):
     connections: ClassVar[set] = set()
 
+    def __init__(self, application: Application, request, **kwargs):
+        super().__init__(application, request, **kwargs)
+        self.application: Application = application
+        self.filters: dict[str, flowfilter.TFilter] = {}  # Instance-level filters
+
+    def send_flow(self, cmd: str, f: mitmproxy.flow.Flow):
```
mhils, line 335: I think we can make this work as follows:

```python3
    @classmethod
    def broadcast_flow(cls, cmd: str, f: mitmproxy.flow.Flow) -> None:
        flow_json = flow_to_json(f)
        for conn in cls.connections:
            conn._broadcast_flow(cmd, flow_json)

    def _broadcast_flow(self, cmd: str, f: Any) -> None:
        # <current contents of send_flow>
```

mitmproxy/tools/web/app.py
lups2000: what do you mean precisely? somehow I need to access `self.application.master.view` below. 
I've specified again the type here: `self.application: Application = application`, because mypy was complaining.

mitmproxy/tools/web/app.py
lups2000: without an explicit cast to bool, this doesn't work for some reasons

mitmproxy/tools/web/app.py
mhils: `super()` already assigns self.application. A type annotations [like this](https://github.com/mitmproxy/mitmproxy/blob/16660ca6e322df47c372afca7c5204fe91cd0005/mitmproxy/tools/web/app.py#L210) is enough.

lups2000: it works :) thanks for the advice!

mitmproxy/tools/web/app.py
mhils: Addressed in [3ff2680](https://github.com/mitmproxy/mitmproxy/pull/7506/commits/3ff26802302666afef497cc26fad46640af55bf5).

lups2000: thanks :)

mitmproxy/tools/web/app.py
lups2000: yes but with a small change:
```
def _broadcast_flow(self, cmd: str, f: mitmproxy.flow.Flow) -> None:
        matches = {
            expr.pattern: bool(expr(f))
            for expr in self.filters.values()
        }
```
so we can use `expr(f)` because f is a `Flow`, otherwise it would be a `dict` and it doesn't work!

mhils: Ah, great catch. But let's pass both the dict and the flow as arguments then. `flow_to_json` is doing a bunch of work which would be nice to reuse across clients. :)

lups2000: makes sense! done :)

lups2000:
this should be good now, does it make sense to add tests? It would be nice to test just that the filters are set properly but inside `update_filter` for example there is a `self.send(...)` that complicates things

mhils:
>  does it make sense to add tests?

If you want it to be merged, then yes. 😛  I don't see a reason why this should be <100% coverage.

We have one existing WebSocket test you can use for inspiration:
https://github.com/mitmproxy/mitmproxy/blob/main/test/mitmproxy/tools/web/test_app.py#L397

lups2000:
I've added tests that cover everything, but I'm unsure if it's fine that I mocked the application separately. Initially, I wrote tests inside the `TestApp` class and tried to access the application using `self.get_app()`. So, when creating a client connection, I was doing something like this: `client = app.ClientConnection(self.get_app(), mock.Mock())`, which caused errors such as: `FAILED test_app.py::TestApp::test_update_filters - RuntimeError: This event loop is already running`.
In the end, I mocked the application separately, and this allowed me to cover everything. Please let me know if I should make any changes. Thank you! :)

mhils:
What happend to the idea of writing an e2e test using a real WebSocket connection as in https://github.com/mitmproxy/mitmproxy/blob/main/test/mitmproxy/tools/web/test_app.py#L397?

lups2000:
> What happend to the idea of writing an e2e test using a real WebSocket connection as in https://github.com/mitmproxy/mitmproxy/blob/main/test/mitmproxy/tools/web/test_app.py#L397?

how? I tried but I couldn't invoke `update_filter` on `ws_client`

lups2000:
@mhils is something like this what you meant?
```
@tornado.testing.gen_test
    def test_websocket_filter_application(self):
        ws_url = f"ws://localhost:{self.get_http_port()}/updates"
        ws_client = yield tornado.websocket.websocket_connect(ws_url)

        filter_expr = "~bq foo"
        filter_name = "search"
        message = json.dumps({
            "resource": "flows",
            "cmd": "updateFilter",
            "name": filter_name,
            "expr": filter_expr,
        }).encode()
        yield ws_client.write_message(message)

        response = yield ws_client.read_message()
        response = json.loads(response)

        expected_message = {
            "resource": "flows",
            "cmd": "filtersUpdated",
            "name": filter_name,
            "expr": filter_expr,
            "data": ["42"],
        }

        assert response == expected_message
        ws_client.close()
```

mhils:
Yes, this looks great!

Nits:
 - https://abseil.io/resources/swe-book/html/ch12.html#donapostrophet_put_logic_in_tests for `filter_expr` and `filter_name` (the entire chapter is a great read on how to do tests)
 - inline expected_message, i.e. `assert response == { ...`
 - Let's add one flow update to cover that part as well. :)

mhils (approved):

Issue #7517, opened 2025-01-27
Exporting a curl command with escaped characters results in a different request


proteusvacuum:
#### Problem Description
When copying a command to curl that contains newline or other characters, the resulting curl request is different to the original.

#### Steps to reproduce the behavior:

1. Make a request that includes escaped characters:
`curl -H 'content-type: application/octet-stream' -d $'foo\nbar' -x localhost:8080 http://example.com`

    In mitmproxy, you will see that the request body is

    ```
    foo
    bar
    ```

2. Copy this request as curl from mitmproxy (export -> curl)
    You get:
    ```
    curl -H 'User-Agent: curl/8.5.0' -H 'Accept: */*' -H 'Proxy-Connection: Keep-Alive' -H 'content-type: application/octet-stream' -X POST http://example.com/ -d 'foo\x0abar'
    ```

3. Run the resulting command (appending the proxy again)
    ```
    curl -H 'User-Agent: curl/8.5.0' -H 'Accept: */*' -H 'Proxy-Connection: Keep-Alive' -H 'content-type: application/octet-stream' -X POST http://example.com/ -d 'foo\x0abar' -x localhost:8080
    ```

4. You see that the request body is not the same as the original:
    ```
    foo\x0abar
    ```
The escape characters that are added [here](https://github.com/mitmproxy/mitmproxy/blob/c5402a490a20cc014e350f2b353dca6eec4bf29b/mitmproxy/addons/export.py#L47-L56) are not being unquoted [when being sent by curl](https://curl.se/docs/manpage.html#-d), and so are just "characters". 
    
I think the solution would be to let bash replace those characters using `$` in front of the string: `... -d $'foo\x0abar` or using [`--data-binary` ](https://curl.se/docs/manpage.html#--data-binary) with a temporary file.

#### System Information
```
Mitmproxy: 11.1.0
Python:    3.12.2
OpenSSL:   OpenSSL 3.4.0 22 Oct 2024
Platform:  Linux-6.5.0-1021-oem-x86_64-with-glibc2.39
```

proteusvacuum:
Thanks @mhils , I'll put together a PR.

PR #7520, opened 2025-01-28
Unescape escaped characters in curl and httpie args

proteusvacuum:
#### Description

This fixes a bug where exporting a curl or httpie command with escaped characters would lead to different data being sent. 

Fixes https://github.com/mitmproxy/mitmproxy/issues/7517

#### Checklist


mhils (approved):
Thanks! 🍰

PR #7535, opened 2025-02-04
Update bug report template to use form

mhils:
#### Description

This is what it looks like:

image


lups2000 (approved):
Thanks! This looks much clearer now. Regarding the **Create More** toggle, what does it do?

PR #7570, opened 2025-02-21
Content view for socket.io over websocket transport

DenizenB:
#### Description

This PR implements a socket.io content view and auto-selects it in the "WebSocket Messages" tab if the view mode is set to "Auto" and the request path contains `/socket.io/?`.

This content view hides engine.io ping/pong packets as well as socket.io acknowledgements, so that it's easier to analyze the messages being exchanged. It also prefixes the data with the name of the engine.io/socket.io packet type in a muted color.

#### Checklist

### DIFF
--- a/mitmproxy/tools/console/flowview.py
+++ b/mitmproxy/tools/console/flowview.py
@@ -217,10 +217,6 @@ class FlowDetails(tabs.Tabs):
 
         viewmode = self.master.commands.call("console.flowview.mode")
 
-        if viewmode == "auto" and "/socket.io/?" in flow.request.path:
-            self.master.commands.call("console.flowview.mode.set", "socket.io")
-            viewmode = self.master.commands.call("console.flowview.mode")
-
         widget_lines = []
         for m in flow.websocket.messages:
             _, lines, _ = contentviews.get_message_content_view(viewmode, m, flow)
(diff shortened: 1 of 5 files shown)


mitmproxy/tools/console/flowview.py
```diff
@@ -217,6 +217,10 @@ def view_websocket_messages(self):
 
         viewmode = self.master.commands.call("console.flowview.mode")
 
+        if viewmode == "auto" and "/socket.io/?" in flow.request.path:
+            self.master.commands.call("console.flowview.mode.set", "socket.io")
+            viewmode = self.master.commands.call("console.flowview.mode")
```
mhils, line 222: Let's make `render_priority` return a sufficiently large value to make `ViewSocketIO` the default view in this case, and we don't need this hack here. :)

mhils (commented):
Thank you! This looks good in general, but needs some tests. :)

mitmproxy/tools/console/flowview.py
DenizenB: Ah sweet, I've fixed this in the latest commit

DenizenB:
Thanks for the feedback, I've added tests in the latest commit. Let's see if the checks turn green

mhils (approved):
Thank you, this looks great now! 🍰 😃 🚀

PR #7573, opened 2025-02-24
Whitelist proxy using a valkey database

FalafelGood:
#### Description

(Related to #7539) 
I've written a plugin for mitmproxy that enforces a whitelist with a [valkey](https://github.com/valkey-io/valkey-py) (redis) database.

On `configure`, this plugin:
+ Checks if the valkey server at the specified address is live.
+ Reads a list of allowed domain names from "whitelist.txt"
+ Uses a sha256 checksum to see if any changes were made to "whitelist.txt" since the last read
+ Pipes the elements into a valkey set called "whitelist"

On `request`, this plugin:
+ extracts the domain name from the http request
+ checks if the domain exists in the valkey database
+ If no, returns a custom 403 page.

#### Checklist

 **Note**: Because this is "just a plugin", I'm not sure if it's applicable for me to update tests and modify the changelog. Happy to do this if need be :-)
 
 ### Alternatives to Valkey
 
 Probably, you don't want to add a `valkey` dependency for `mitmproxy` just to support my random plugin. That's very understandable! A more knowledgeable friend of mine suggested I could work around this issue by writing a [protocol](https://typing.readthedocs.io/en/latest/spec/protocol.html#protocols) for an abstract database object. I'm new to the concept of protocols, but I think I get the gist!
 
Let me know if this is the preferred approach, or if there's some other way I can improve this.
Open to feedback! Thanks for your time


SamuelMarks:
@mhils We were also thinking about having an [abstractclass](https://docs.python.org/3/library/abc.html) or [`Protocol`](https://docs.python.org/3/library/typing.html#typing.Protocol) for the whitelist. Perhaps with two implementations:

  0. Read from filesystem, store and query-from current-thread memory (builtin data-structure; maybe a [`frozenset`](https://docs.python.org/3/library/stdtypes.html#frozenset)?)
  1. Read from filesystem, store and query-from [Valkey](https://valkey.io) (the open-source [redis](https://redis.io) fork)

Would you accept a PR for the [abstractclass](https://docs.python.org/3/library/abc.html)|[`Protocol`](https://docs.python.org/3/library/typing.html#typing.Protocol) & in-memory dependency-free implementation?

Then we can put the [Valkey](https://valkey.io) one in its own `pip install`able python package in a separate repository; and if you're API changes we'll get a nice error in our codebase and don't need to have a separate fork to constantly keep in sync.

In our humble opinion: comprehensive whitelist support will greatly benefit the userbase—and encourage new users of—mitmproxy =)

SamuelMarks:
@mhils We were also thinking about having an [abstractclass](https://docs.python.org/3/library/abc.html) or [`Protocol`](https://docs.python.org/3/library/typing.html#typing.Protocol) for the whitelist. Perhaps with two implementations:

  0. Read from filesystem, store and query-from current-thread memory (builtin data-structure; maybe a [`frozenset`](https://docs.python.org/3/library/stdtypes.html#frozenset)?)
  1. Read from filesystem, store and query-from [Valkey](https://valkey.io) (the open-source [redis](https://redis.io) fork)

Would you accept a PR for the [abstractclass](https://docs.python.org/3/library/abc.html)|[`Protocol`](https://docs.python.org/3/library/typing.html#typing.Protocol) & in-memory dependency-free implementation?

Then we can put the [Valkey](https://valkey.io) one in its own `pip install`able python package in a separate repository; and if your API changes we'll get a nice error in our codebase and don't need to have a separate fork to constantly keep in sync.

In our humble opinion: comprehensive whitelist support will greatly benefit the userbase—and encourage new users of—mitmproxy =)

mhils:
I think this is really cool work and quite useful. But it's relatively specialized and I don't want to take over any maintenance burden. So putting it into examples/contrib is best for the time being.

SamuelMarks:
@mhils How about a concurrently more extreme and more general solution, expanding your `mitmproxy.addons.script` to be able to refer to installed modules, that way we can maintain valkey & other whitelist support in an external Python package and use the script addon to load it in; something like:
```patch
diff --git a/mitmproxy/addons/script.py b/mitmproxy/addons/script.py
index bc5bc0e..c32aa54 100644
--- a/mitmproxy/addons/script.py
+++ b/mitmproxy/addons/script.py
@@ -1,6 +1,7 @@
 import asyncio
 import importlib.machinery
 import importlib.util
+import importlib
 import logging
 import os
 import sys
@@ -21,6 +22,11 @@ logger = logging.getLogger(__name__)
 
 
 def load_script(path: str) -> types.ModuleType | None:
+    if not os.path.isfile(path):
+        mod = importlib.import_module(path)
+        sys.modules[path] = mod
+        return mod
+
     fullname = "__mitmproxy_script__.{}".format(
         os.path.splitext(os.path.basename(path))[0]
     )
```

FalafelGood:
Hey @mhils, how does this look? I did everything [you recommended](https://github.com/mitmproxy/mitmproxy/pull/7573#issuecomment-2681476814). Seems like some checks are failing though... I don't know how to fix that, but if you point me in the right direction I'll give it my best shot!

Any thoughts on what @SamuelMarks suggested about [supporting external plugins](https://github.com/mitmproxy/mitmproxy/pull/7573#issuecomment-2706916884)? 

Thanks again for your time!

mhils:
Thanks! Please remove the old file from `mitmproxy/` and CI should turn green :)

PR #7574, opened 2025-02-25
Do not send empty HTTP/2 data frames without EoS bit unless the client sent some

mhils:
#### Description

Closes #7480

/cc @Dieken

#### Checklist

### DIFF
--- a/mitmproxy/proxy/layers/http/_http2.py
+++ b/mitmproxy/proxy/layers/http/_http2.py
@@ -204,10 +204,7 @@ class Http2Connection(HttpConnection):
         if isinstance(event, h2.events.DataReceived):
             state = self.streams.get(event.stream_id, None)
             if state is StreamState.HEADERS_RECEIVED:
-                is_empty_eos_data_frame = (
-                    event.stream_ended and
-                    not event.data
-                )
+                is_empty_eos_data_frame = event.stream_ended and not event.data
                 if not is_empty_eos_data_frame:
                     yield ReceiveHttp(self.ReceiveData(event.stream_id, event.data))
             elif state is StreamState.EXPECTING_HEADERS:
--- a/mitmproxy/proxy/layers/http/_http2.py
+++ b/mitmproxy/proxy/layers/http/_http2.py
@@ -204,7 +204,12 @@ class Http2Connection(HttpConnection):
         if isinstance(event, h2.events.DataReceived):
             state = self.streams.get(event.stream_id, None)
             if state is StreamState.HEADERS_RECEIVED:
-                yield ReceiveHttp(self.ReceiveData(event.stream_id, event.data))
+                is_empty_eos_data_frame = (
+                    event.stream_ended and
+                    not event.data
+                )
+                if not is_empty_eos_data_frame:
+                    yield ReceiveHttp(self.ReceiveData(event.stream_id, event.data))
             elif state is StreamState.EXPECTING_HEADERS:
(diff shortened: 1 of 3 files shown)


mitmproxy/proxy/layers/http/_http2.py
```diff
@@ -204,7 +204,9 @@ def handle_h2_event(self, event: h2.events.Event) -> CommandGenerator[bool]:
         if isinstance(event, h2.events.DataReceived):
             state = self.streams.get(event.stream_id, None)
             if state is StreamState.HEADERS_RECEIVED:
-                yield ReceiveHttp(self.ReceiveData(event.stream_id, event.data))
+                is_empty_eos_data_frame = event.stream_ended and not event.data
+                if not is_empty_eos_data_frame:
+                    yield ReceiveHttp(self.ReceiveData(event.stream_id, event.data))
```
errorxyz, line 209: > On the other hand, if a client does send empty data frames, we shouldn't be eating them if we can avoid it. This creates a really annoying "client fails without mitmproxy, but when I add mitmproxy as a debugger, it suddenly works" situation.



Aren't we blocking all empty EoS data frames here? What if they're client initiated?

mitmproxy/proxy/layers/http/_http2.py
```diff
@@ -204,7 +204,9 @@ def handle_h2_event(self, event: h2.events.Event) -> CommandGenerator[bool]:
         if isinstance(event, h2.events.DataReceived):
             state = self.streams.get(event.stream_id, None)
             if state is StreamState.HEADERS_RECEIVED:
-                yield ReceiveHttp(self.ReceiveData(event.stream_id, event.data))
+                is_empty_eos_data_frame = event.stream_ended and not event.data
```
errorxyz, line 207: Can we have a non-empty data frame with `event.stream_ended` set? If so, will we be sending the event data and the EoS flag in separate frames?

test/mitmproxy/proxy/layers/http/test_http2.py
```diff
@@ -1281,3 +1281,61 @@ def test_alt_svc(tctx):
...
+    initial_frames = decode_frames(initial())
+    assert [type(x) for x in initial_frames] == [
+        hyperframe.frame.SettingsFrame,
+        hyperframe.frame.WindowUpdateFrame,
+        hyperframe.frame.HeadersFrame,
+    ]
+
+    assert (
+        playbook
+        >> DataReceived(
```
errorxyz, line 1319: Do we want to add a test where client initiated empty data frames are actually passed through? It might be tested elsewhere but adding it here might help while debugging in case things break.

mitmproxy/proxy/layers/http/_http2.py
mhils: Yes. This is a bit of a limitation of the current approach. But empty data frames _with_ EOS are fine. The only thing that some clients don't like is empty data frames _without_ EOS.

mitmproxy/proxy/layers/http/_http2.py
mhils: The EOS empty data frame comes anyway from StreamEnded / ReceiveEndOfMessage.

test/mitmproxy/proxy/layers/http/test_http2.py
mhils: Goods idea, done!

PR #7603, opened 2025-03-11
Docs: Add Search Bar

mhils:
#### Description

This PR adds the first implementation of a search bar to the docs. The design is mostly based off what we do in pdoc:

image

Search is completely local. It's not 100% perfect (preview text is a bit tricky, so is linking to individual headings), but better than no search. Improvements welcome. :)

#### Checklist


lups2000 (approved):
Really nice as a first prototype :) I tested it and it worked fine! 🚀

Issue #3140, opened 2018-05-19
Add revocation information to certificates


Yepoleb:
libcurl with the WinSSL/Schannel backend fails with an SSL error if no revocation information is provided in the forged certificate. This is preventing me from intercepting an application that uses a statically linked libcurl.

I was able to reproduce this issue using curl 7.60.0 configured with `--without-ssl --with-winssl`, compiled with MSYS2 mingw-w64-x86_64-gcc 7.3.0 on Windows 7. The full error message is `curl: (35) schannel: next InitializeSecurityContext failed: Unknown error (0x80092012) - The revocation function was unable to check revocation for the certificate.`

I've not yet figured out what exactly is required by curl to accept the certificate, but I'll update the issue once I do. Adding an unresolved URL might just be enough to satisfy it, considering the awful state of revocation in general.

##### System information

Mitmproxy: 3.0.3 
Python:    3.6.5
OpenSSL:   OpenSSL 1.1.0h  27 Mar 2018
Platform:  Linux-4.16.0-1-amd64-x86_64-with-debian-buster-sid

Yepoleb:
I was able to make the revocation check pass by generating an empty CRL in DER format using OpenSSL and adding it to the certificate like this:

    cert.add_extensions(
        [OpenSSL.crypto.X509Extension(
            b"crlDistributionPoints", False, b"URI:http://mypc.lan/mitmproxy.crl")])

An unresolved URL was unfortunately not enough to pass, it needs to reachable.

cortesi:
Thanks for reporting this. One possibility is to provide a CRL distribution point within mitmproxy under the magic "mitm.it" domain we use for cert installation. There are a couple of things to think about:

- It could cause privacy concerns because of the DNS lookup of mitm.it, which would go to our public DNS server by default, and occur even for unrelated traffic. 
- It could cause hard-to-diagnose issues in isolated networks where mitm.it doesn't resolve. 
- I'd want to be sure that including a CRL distribution point doesn't cause flow-on effects for other clients. We'd have to test quite carefully.

@mitmproxy/devs Thoughts?

Yepoleb:
Thanks for your interest. I'm not sure what you mean by traffic going to your public DNS server. The initial domain query should never reach your servers, only the primary resolver which then anonymizes the request. I also can't think of a scenario where the certificate would get served but the CRL request still goes to the public mitm.it page.

cortesi:
[...]

PR #7609, opened 2025-03-16
Added CRL to dummy cert

JordanPlayz158:
#### Description

There has been a long standing issue #3140 so I tested out the code on the latest commit and it seems to still work :) so I figured I would try to get it merged

Here is a more detailed description from https://github.com/mitmproxy/mitmproxy/commit/0a095b28cd15abe3bbc2da25563bc7899a6b3309
> Properly integrate CRLs into the TLS flow
When establishing a connection the CRL distribution points are
extracted from the upstream cert. They are passed to the dummy
certificate creation function and their original path is replaced
with the CA's serial number. This makes it possible to detect the
modified URL without having to store any additional per-URL state.
>
> The certificate store has been extended to generate an empty CRL
at initialization. The tlsconfig addon has been extended to detect
a modified CRL URL and serve the empty CRL from the certificate
store.

#### Checklist


Yepoleb:
I wrote a detailed description in 0a095b28cd15abe3bbc2da25563bc7899a6b3309, it would be nice if you could preserve it in the commit message.

JordanPlayz158:
I'm sorry if this sounds silly but how would I do this, is there a way to rename an existing commit/message or do you mean when the merge commit is made or?

Yepoleb:
You can do an interactive git rebase to change the message of the initial commit. I just realized it should ideally be in a proper documentation file though.

JordanPlayz158:
Ohhh, I see, I'll see if there is a place that makes sense to add that, also because this contribution is 99.9999% of your code, do you want me to put your @ alongside mine in the CHANGELOG (probably first), that may sound silly, maybe even insulting given it is your work but my intention was to ensure it was okay with you before doing so, not to try to pass off your work as my own or anything in that realm

Yepoleb:
I would appreciate if you credit me in the changelog. I did not find a good place to put the documentation in the code, but I think if you put it in your PR description it will be easy enough to find.

JordanPlayz158:
Added you to changelog and alright, will do so!

mitmproxy/certs.py
```diff
@@ -337,10 +363,54 @@ def dummy_cert(
...
+    privkey: CA private key
+    cacert: CA certificate
+
+    Returns a CRL DER encoded
+    """
+    builder = x509.CertificateRevocationListBuilder()
+    builder = builder.issuer_name(cacert.issuer)
+
+    now = datetime.datetime.now()
+    builder = builder.last_update(now)
```
mhils, line 406: Please copy https://github.com/mitmproxy/mitmproxy/blob/6269898b434e1040c4ea46a425526c5d266c554b/mitmproxy/certs.py#L222 to account for clock mismatches.

mitmproxy/certs.py
```diff
@@ -281,12 +300,18 @@ def _fix_legacy_sans(sans: Iterable[x509.GeneralName] | list[str]) -> x509.Gener
...
+    crl_distribution_points: list[str] = field(default_factory=list)
+
+
 def dummy_cert(
     privkey: rsa.RSAPrivateKey,
     cacert: x509.Certificate,
     commonname: str | None,
     sans: Iterable[x509.GeneralName],
     organization: str | None = None,
+    revocation: RevocationInfo = RevocationInfo(),
```
mhils, line 314: Can we simplify this to `crl_url: str | None = None` please?

mitmproxy/addons/tlsconfig.py
```diff
@@ -616,7 +619,20 @@ def get_cert(self, conn_context: context.Context) -> certs.CertStoreEntry:
...
         # In other words, the Common Name is irrelevant then.
         cn = next((str(x.value) for x in altnames), None)
-        return self.certstore.get_cert(cn, altnames, organization)
+        return self.certstore.get_cert(cn, altnames, organization, revocation)
+
+    async def request(self, flow: http.HTTPFlow):
+        if not flow.live or flow.error or flow.response:
+            return
+        # Check if a request has a magic CRL token at the end
+        magic_token = str(self.certstore.default_ca.serial) + ".crl"
```
mhils, line 628: This can be a `crl_path` `@property` that can be reused between `request` and `get_cert`.

mitmproxy/certs.py
```diff
@@ -337,10 +363,54 @@ def dummy_cert(
     # https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.2 states
     # that SKI is optional for the leaf cert, so we skip that.
 
+    fake_distribution_points = []
+    for upstream_distribution_point in revocation.crl_distribution_points:
+        upstream_url = urllib.parse.urlparse(upstream_distribution_point)
+        # Replace original URL path with the CA cert serial number, which acts as a magic token
+        fake_url = upstream_url._replace(
```
mhils, line 370: I think there's a subtle difference between this and `request.path.endswith()` in tlsconfig: `request.path` may include a query string. https://github.com/mitmproxy/mitmproxy/blob/main/mitmproxy/net/http/url.py#L20 is the better alternative here.

mhils (commented):
Thank you two for pushing this forward! Looks great overall, left a few comments below.

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@
   ([#7570](https://github.com/mitmproxy/mitmproxy/pull/7570), @DenizenB)
 - Fix a bug where mitmweb would crash when viewing flows with undefined headers.
   ([#7595](https://github.com/mitmproxy/mitmproxy/pull/7595), @emanuele-em)
+- Add CRL entries to dummy cert when exists in upstream certificate.
```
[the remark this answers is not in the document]
JordanPlayz158: Hmmm, should it be a config option or just a blanket add crl regardless (as isn't mitmproxy designed to match the upstream cert details as much as possible? (at least that's what it seems to strive to do))

mitmproxy/certs.py
JordanPlayz158: So the dummy cert will only be able to account for one of the CRLs in the upstream cert? I guess that should be fine??

JordanPlayz158:
Ohhh, the codecov test goes on a per-file basis, interesting

JordanPlayz158:
Hmmm, seems like the tests may have a race condition that causes them to fail to bind some of the time, I encountered it only once in testing thus far and now the CI has done so on the latest pass it looks like which is why it is failing?

mitmproxy/certs.py
mhils: Function with docstring sounds good. Constant doesn't work because people run mitmdump for 2+ days. But I'm also fine with duplicating `now - datetime.timedelta(days=2)`, I don't anticipate any more usage. Your call. :)

mhils:
> Hmmm, seems like some of the tests may have a race condition that causes them to fail to bind some of the time, I encountered it only once in testing thus far and now the CI has done so on the latest pass it looks like which is why it is failing?

Tracked in https://github.com/mitmproxy/mitmproxy/issues/7614, thanks.

mitmproxy/certs.py
JordanPlayz158: Oh alright, will revise it to just `str`, sounds good

JordanPlayz158:
Ahh, thank you!

CHANGELOG.md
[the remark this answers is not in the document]
mhils: If we have upstream cert info, I think we should add a CRL entry if upstream has one. If upstream doesn't have a CRL, I have no strong opinion on whether we should have one.
If we don't have upstream cert info (e.g. because of https://docs.mitmproxy.org/dev/concepts/options/#connection_strategy), we should ideally add a CRL so that it still works with those clients that expect one. But this can be a second, follow-up PR.

Put differently, always adding a CRL would be fine with me (but it's a bit of a question which URL we should use). We can also make it a config option, but I'm not sure what we'd gain from that. Maybe to make the hostname configurable? I feel we're getting a bit sidetracked - let's focus on the case where upstream has a CRL first in this PR. 😃

mitmproxy/addons/tlsconfig.py
```diff
@@ -583,23 +586,46 @@ def _warn_seclevel_missing(self, side: Literal["client", "server"]) -> None:
                 f"for insecure TLS versions to work."
             )
 
+    @property
+    def crl_path(self) -> str:
+        if not self.certstore:
```
mhils, line 591: `certstore` should always be set, at least it's typed as `CertStore` and not as `CertStore | None`. The implicit contract (which we can't model very well) is that `configure()` is called right after init with the correct args.

 Long story short, we should not have this check here.

mitmproxy/addons/tlsconfig.py
```diff
@@ -583,23 +586,46 @@ def _warn_seclevel_missing(self, side: Literal["client", "server"]) -> None:
...
+            return ""
+
+        return "/" + str(self.certstore.default_ca.serial) + ".crl"
```
mhils, line 594: Would `"/mitmproxy-" + str(self.certstore.default_ca.serial) + ".crl"` make sense to make it clear what's happening here to someone who inspects the cert?

mitmproxy/addons/tlsconfig.py
```diff
@@ -583,23 +586,46 @@ def _warn_seclevel_missing(self, side: Literal["client", "server"]) -> None:
...
+
+                # I hope I am just doing something stupid and that this is not the intended way to actually go about doing this
+                crl_distribution_points.append(
+                    unparse(
+                        urlParts[0].decode("ascii"),
+                        urlParts[1].decode("idna"),
+                        urlParts[2],
+                        urlParts[3],
+                    )
+                )
```
mhils, line 627: Nit:
```suggestion
                scheme, host, port, _ = parse(url)
                url = unparse(scheme, host, port, self.crl_path)
                crl_distribution_points.append(url)
```

(+ the list[str] -> str change)

mhils (commented):
Thanks for the quick feedback loop!

CHANGELOG.md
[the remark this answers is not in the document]
JordanPlayz158: > We can also make it a config option, but I'm not sure what we'd gain from that. Maybe to make the hostname configurable? I feel we're getting a bit sidetracked - let's focus on the case where upstream has a CRL first in this PR. 😃

Yeah that is my bad, I was thinking of a theoretical scenario where an application would check the length of the CRLs the cert returned but if they were that strict they'd probably just check if the urls exactly match (but it was a "what-if" which is not very useful, at least in this case) what is expected anyway so... yeah I was just overthinking when writing that, and yeah, we could definitely save it for a second pr as, as mentioned, the URL which should be used is something to be considered

mitmproxy/certs.py
```diff
@@ -337,10 +363,54 @@ def dummy_cert(
...
+            path=str(cacert.serial_number) + ".crl"
+        ).geturl()
```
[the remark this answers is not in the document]
JordanPlayz158: I think I moved this to the right place?

mitmproxy/addons/tlsconfig.py
JordanPlayz158: Should be resolved?

mhils: Then you can just click resolve. :)

mitmproxy/certs.py
JordanPlayz158: Switched to the parse function in url.py as recommended

mitmproxy/addons/tlsconfig.py
JordanPlayz158: Apologies, I wasn't sure if it is considered impolite to do so instead of letting the original starter of the conversation confirm it is indeed resolved and have them resolve it, will mark it as resolved though, ty!

mitmproxy/addons/tlsconfig.py
JordanPlayz158: Hmmm, the weird thing is, if I remove this check, mitmproxy fails to load
```
  File "/home/jordan/Coding/Python/mitmproxy-crl/venv/bin/mitmproxy", line 8, in <module>
    sys.exit(mitmproxy())
             ~~~~~~~~~^^
  File "/home/jordan/Coding/Python/mitmproxy-crl/mitmproxy/tools/main.py", line 143, in mitmproxy
    run(console.master.ConsoleMaster, cmdline.mitmproxy, args)
    ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jordan/Coding/Python/mitmproxy-crl/mitmproxy/tools/main.py", line 137, in run
    return asyncio.run(main())
           ~~~~~~~~~~~^^^^^^^^
  File "/usr/lib64/python3.13/asyncio/runners.py", line 195, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/usr/lib64/python3.13/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/usr/lib64/python3.13/asyncio/base_events.py", line 725, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/home/jordan/Coding/Python/mitmproxy-crl/mitmproxy/tools/main.py", line 67, in main
    master = master_cls(opts)
  File "/home/jordan/Coding/Python/mitmproxy-crl/mitmproxy/tools/console/master.py", line 50, in __init__
    self.addons.add(*addons.default_addons())
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jordan/Coding/Python/mitmproxy-crl/mitmproxy/addonmanager.py", line 204, in add
    self.chain.append(self.register(i))
                      ~~~~~~~~~~~~~^^^
  File "/home/jordan/Coding/Python/mitmproxy-crl/mitmproxy/addonmanager.py", line 194, in register
    self.master.commands.collect_commands(a)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
  File "/home/jordan/Coding/Python/mitmproxy-crl/mitmproxy/command.py", line 177, in collect_commands
    o = getattr(addon, i)
  File "/home/jordan/Coding/Python/mitmproxy-crl/mitmproxy/addons/tlsconfig.py", line 591, in crl_path
    return "/" + str(self.certstore.default_ca.serial) + ".crl"
                     ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'default_ca'
```
mitmproxy.command.CommandManager#collect_commands calls the getattr method on every attribute of the addon and this includes `crl_path` so as a result self.default_ca is not yet set when mitmproxy attempts to access it... I guess (so this `collect_commands` comes before configure is called), I will look at other addons and see if any of them have similar things like properties dependent on things initialized in the configure function and see if they did anything to rectify it, or perhaps we just have a crl_path variable that is none and set by configure same as certstore?

mitmproxy/addons/tlsconfig.py
JordanPlayz158: Hmmm, the issue with that code is, it produces a cert with this type of url
```
X509v3 CRL Distribution Points: 
  Full Name:
    URI:b'http'://b'crl3.digicert.com'/mitmproxy-132645061682444739888094978381816565991132842371.crl
```
Which is why I had to do that weird way, does x509 not mind this mixing of `b'`/binary strings or...? As normal certs seem to just be a string without any binary strings in them

mhils: Ughs - I didn't realize that `url.parse` is super old terrible code. It looks like it returns `scheme, host, port, path` as `tuple[bytes, bytes, int, bytes]`, but `unparse` then expects `str` arguments. 🤦 

Let me fix that in a separate PR in the coming days.

mitmproxy/addons/tlsconfig.py
JordanPlayz158: Done!

mitmproxy/addons/tlsconfig.py
JordanPlayz158: Alrighty then! Ty!

mhils: https://github.com/mitmproxy/mitmproxy/pull/7624 should have you covered. The URI should of course be
```
http://crl3.digicert.com/mitmproxy-132645061682444739888094978381816565991132842371.crl
```
and not contain any other byte markers. :)

JordanPlayz158: I am confused on what the new code should be, looking at the changes, the current code I have seems to be the correct way still? (As it checks if scheme is string and if so, prints all as a string? (and as parse returns bytes for scheme and host, I still need to decode them back into strings from bytes?))

JordanPlayz158: Forgive me for doing so but it has been almost 2 weeks, so I am writing a message to "bump" the thread

Ashesh3:
Hey, is anything blocking this PR? I have the same issue, and it would be great to get this merged!

mhils:
It's only a lack of review capacity at the moment. It's one of the top entries on my review queue, sorry for the delay.

PR #7623, opened 2025-03-28
Add new contentview API

mhils:
#### Description

This PR is a first draft proposal for a new contentview API that makes it possible for users to _edit_ prettified views. This is particular useful for binary formats such as protobuf, which are otherwise hard to tinker with.

A first proof-of-concept for Protobuf lives at https://github.com/mhils/mitmproxy_rs/pull/4, but it's not wired up yet. :)

### DIFF
--- a/examples/addons/contentview-priority.py
+++ b/examples/addons/contentview-priority.py
@@ -1,0 +1,9 @@
+class CustomPriority(Contentview):
+    def render_priority(self, data: bytes, metadata: Metadata) -> float:
+        match metadata.content_type:
+            case "application/json":
+                return 1.0   # preferred contentview for JSON.
+            case "image/png":
+                return -1.0  # cannot handle images at all.
+            case _:
+                return 0.0
--- a/mitmproxy/contentviews/api.py
+++ b/mitmproxy/contentviews/api.py
@@ -1,0 +1,76 @@
+from __future__ import annotations
+from abc import ABC, ABCMeta, abstractmethod
+from dataclasses import dataclass
+from typing import ClassVar, Literal
+from mitmproxy import http
+from mitmproxy.flow import Flow
+
+
+class Contentview(ABC):
+    """A contentview that prettifies raw data."""
+
+    @property
+    def name(self) -> str:
+        """
+        The name of this contentview, e.g. "XML/HTML". 
+        Inferred from the class name by default.
+        """
+        return type(self).__name__.removesuffix("Contentview")
+    
+    @abstractmethod
+    def prettify(
+        self,
+        data: bytes,
+        metadata: Metadata,
+    ) -> str:
+        """Transform raw data into human-readable output."""
+
+    def render_priority(
+        self,
+        data: bytes,
+        metadata: Metadata,
+    ) -> float | None:
+        """
+        Return the priority of this view for rendering `data`.
+        If no particular view is chosen by the user, the view with the highest priority is selected.
+        If this view does not support the given data, return a float < 0.
+        """
+
+    syntax_highlight: ClassVar[Literal["yaml", "none"]] = "none"
+    """Optional syntax highlighting that should be applied to the prettified output."""
+
+    def __lt__(self, other):
+        assert isinstance(other, Contentview)
+        return self.name.__lt__(other.name)
+
+
+class InteractiveContentview(Contentview, metaclass=ABCMeta):
+    """A contentview that prettifies raw data and allows for interactive editing."""
+
+    @abstractmethod
+    def reencode(
+        self,
+        prettified: str,
+        original_data: bytes,
+        metadata: Metadata,
+    ) -> bytes:
+        """
+        Reencode the given (modified) `prettified` output into the original data format.
+        """
+    
+@dataclass
+class Metadata:
+    """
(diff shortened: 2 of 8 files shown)


errorxyz (approved):
Great stuff :rocket:

emanuele-em (approved):
💯

examples/addons/contentview-priority.py
```diff
@@ -0,0 +1,9 @@
+class CustomPriority(Contentview):
+    def render_priority(self, data: bytes, metadata: Metadata) -> float:
+        match metadata.content_type:
+            case "application/json":
+                return 1.0   # preferred contentview for JSON.
+            case "image/png":
+                return -1.0  # cannot handle images at all.
+            case _:
+                return 0.0
```
Prinzhorn, line 11: Not sure I understand why the examples are not combined as before? What purpose does the `CustomPriority` example view serve if it doesn't actually implement `prettify`? Isn't that a requirement of `@abstractmethod`?

mitmproxy/contentviews/api.py
```diff
@@ -0,0 +1,76 @@
...
+        data: bytes,
+        metadata: Metadata,
+    ) -> str:
+        """Transform raw data into human-readable output."""
+
+    def render_priority(
+        self,
+        data: bytes,
+        metadata: Metadata,
+    ) -> float | None:
```
Prinzhorn, line 34: Why does this allow `None` and doesn't default to `return 0`?

mhils: Fixed, thanks.

examples/addons/contentview-priority.py
mhils: Thinking about it... yeah. I did this so that I can have text in the middle in the docs, but we should rather fix the docs to allow partial regurgitation of examples.

mhils:
This should be good to go! 🚀 

I understand that it's incredibly hard to review, but I'd really appreciate a second look over docs/src/content/addons/contentviews.md and examples/. 😃

bburky:
For my SAML use case, I wanted my prettify to do things like autoformatting XML text with indentation, which makes it significantly more readable.

However, for editing, I actually wanted to operate on the maybe-deflated and base64-decoded data, because my prettify would break XML signatures.

1. Decode
2. Prettify _for UI visualization only_
3. edit (on the output of Decode) and re-encode when complete

Is this setup possible with this new feature?

mhils:
Yes. Here's an example for gRPC/Protobuf: https://asciinema.org/a/MivOWmlDxV3juRNvjhxFU7FH8

With the new API, you can implement:

1. `prettify(data: bytes, metadata: Metadata) -> str` to implement autoformatting/indentation, and
2. `reencode(prettified: str, metadata: Metadata) -> bytes` to implement reencoding + signature calculation.

bburky:
Not just syntax highlighting. Adding whitespace and things, which is actually destructive with XML signatures. Any other similar destructive feature is actually nice to do in prettify for better visualization. But, I want an edit feature to run on un-prettifyed text, because prettify may have destroyed data. But I still do want to have decoding (deflate and base64) performed even for edit.

Something like:
* `decode(data: bytes, metadata: Metadata) -> decode binary format for _editing only_
* `prettify(data: bytes, metadata: Metadata) -> str` to implement autoformatting/indentation
* `reencode(decoded: bytes, metadata: Metadata) -> bytes` to implement reencoding. Note: operates on edited decoded bytes, not prettified str

What you have is great for formats that roundtrip through prettify without losing data. Maybe it isn't meant to map exactly to my use case. I guess I could still use this new API by having two contentviews: one for editing that does non-destructive things, and a second that's better for visualization and does prettify aggressively.

mhils:
> But, I want an edit feature to run on un-prettifyed text, because prettify may have destroyed data.

I think this sounds like a very specific use case that may not warrant complicating the API. As you mention, a simple workaround would be to have two separate contentviews here. But more fundamentally, the reencoder could also undo the whitespace changes. That probably not super easy to implement... but feels very much possible. 😃  Let's keep this complexity in the individual contentview, and not in the API that's shared by all contentviews.

For context, while MsgPack reencoding is laughably simple ([msgpack.rs#L35-L46](https://github.com/mitmproxy/mitmproxy_rs/blob/main/mitmproxy-contentviews/src/msgpack.rs#L35-L46)), Protobuf is a [convoluted dance](https://github.com/mitmproxy/mitmproxy_rs/tree/main/mitmproxy-contentviews/src/protobuf). I think it's expected that some contentviews are a bit more complex than others.

test/mitmproxy/contentviews/test__view_raw.py
```diff
@@ -0,0 +1,15 @@
+from mitmproxy.contentviews._api import Metadata
+from mitmproxy.contentviews._view_raw import raw
+
+
+def test_view_raw():
+    meta = Metadata()
+    assert raw.prettify(b"foo", meta)
+    # unicode
+    assert raw.prettify("🫠".encode(), meta) == "🫠"
```
lups2000, line 9: your mood while writing tests 😅

emanuele-em (approved):

lups2000 (approved):
🚀 Really nice!

PR #7637, opened 2025-04-04
Log streamed bodies

mkiami:
#### Description

Add a new feature to store streamed bodies for requests and responses. The feature is disabled by default, and can be enabled by setting the option "store_streamed_bodies" to True.

#### Checklist


test/mitmproxy/proxy/layers/http/test_http.py
```diff
@@ -338,8 +338,10 @@ def test_disconnect_while_intercept(tctx):
 
 @pytest.mark.parametrize("why", ["body_size=0", "body_size=3", "addon"])
 @pytest.mark.parametrize("transfer_encoding", ["identity", "chunked"])
-def test_response_streaming(tctx, why, transfer_encoding):
+@pytest.mark.parametrize("store_streamed_bodies", [True, False])
+def test_response_streaming(tctx, why, transfer_encoding, store_streamed_bodies):
     """Test HTTP response streaming"""
+    tctx.options.store_streamed_bodies = store_streamed_bodies
```
mhils, line 344: I don't like how the changes here overload existing tests. Can we add one test that's entirely dedicated to `store_streamed_bodies=True` please? In that test, we can then assert that `.content` is set in the request and response hooks. We don't need the parametrization then. :)

mitmproxy/options.py
```diff
@@ -15,6 +15,12 @@ def __init__(self, **kwargs) -> None:
         self.add_option(
             "server", bool, True, "Start a proxy server. Enabled by default."
         )
+        self.add_option(
+            "store_streamed_bodies",
+            bool,
+            False,
+            "Store HTTP request and response bodies when streamed. ",
+        )
```
mhils, line 23: This one here seems unnecessary?

mitmproxy/addons/proxyserver.py
```diff
@@ -144,6 +144,12 @@ def register_connection(
             del self.connections[connection_id]
 
     def load(self, loader):
+        loader.add_option(
+            "store_streamed_bodies",
+            bool,
+            False,
+            "Store HTTP request and response bodies when streamed.",
```
mhils, line 151: Let's expand the docstring a bit. How about this?

```suggestion
            "Store HTTP request and response bodies when streamed (see `stream_large_bodies`). "
            "This increases memory consumption, but makes it possible to inspect streamed bodies.",
```

We should also point to it in the `stream_large_bodies` docstring (which also should mention _request_ bodies while you're at it 🙏 ).

mitmproxy/proxy/server.py
```diff
@@ -506,6 +506,12 @@ async def handle_hook(self, hook: commands.StartHook) -> None:
 
     opts = moptions.Options()
     # options duplicated here to simplify testing setup
+    opts.add_option(
+        "store_streamed_bodies",
+        bool,
+        False,
+        "Store HTTP request and response bodies when streamed.",
```
mhils, line 513: This is for testing only. Let's do empty string for all help texts to make it clear that these are not used anywhere in practice.

```suggestion
        "",
```

mhils (commented):
Very nice PR, thank you! 🍰

test/mitmproxy/proxy/layers/http/test_http.py
mkiami: Sure, I have added a test dedicated to the new feature, with both cases False and True, since we also want to test that the contents are not being stored when the feature is off.

mitmproxy/options.py
mkiami: I added the option here since I was not sure if it was needed, I've just removed it.

mitmproxy/addons/proxyserver.py
mkiami: Done in the latest commit

mitmproxy/proxy/server.py
mkiami: Done in the latest commit

nth347:
hello @mhils and @mkiami , thank you for the PR, how to enable the feature when using mitmproxy as library?

PR #7644, opened 2025-04-07
Fix `message_to_json` to always show raw payloads when message content is empty

themudhaxk:
This commit modifies the `message_to_json` function to ensure that raw message payloads are always shown when the parsed content is empty or consists of only empty parts. Previously, the function would not fallback to showing raw content if the parsed lines were empty, which led to missing data in some cases. 

With this fix, if the `lines` are empty or consist of only empty parts, the raw payload of the message is decoded and included in the output. This change improves the handling of edge cases where the content view parsing fails or doesn't generate any meaningful data, ensuring that the raw message content is always available for inspection.

Changes:
- Added a fallback to display raw payload if `lines` are empty or contain only empty parts.
- The raw message is decoded using UTF-8 and added to the lines if needed, with proper error handling in case of decoding issues.

#### Description

Fix `message_to_json` to always show raw payloads when message content is empty.

This change ensures that if the parsed message content (lines) is empty or consists of empty parts, the raw payload of the message will be shown. Previously, the function failed to display raw content in such cases, which could lead to missing message data. With this fix, raw content is decoded and included when necessary, providing a fallback for edge cases where content parsing fails.

#### Checklist

### DIFF
--- a/mitmproxy/tools/web/app.py
+++ b/mitmproxy/tools/web/app.py
@@ -587,6 +587,26 @@ class FlowContent(RequestHandler):
 
 
 class FlowContentView(RequestHandler):
+    # def message_to_json(
+    #     self,
+    #     viewname: str,
+    #     message: http.Message | TCPMessage | UDPMessage | WebSocketMessage,
+    #     flow: HTTPFlow | TCPFlow | UDPFlow,
+    #     max_lines: int | None = None,
+    # ):
+    #     description, lines, error = contentviews.get_message_content_view(
+    #         viewname, message, flow
+    #     )
+    #     if error:
+    #         logging.error(error)
+    #     if max_lines:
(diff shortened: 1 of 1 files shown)


themudhaxk:
#### Current Implementation

Screenshot 2025-04-07 030450

#### Previous Implementation

Screenshot 2025-04-07 030655

mhils:
Thank you for the PR! This is a great fix, but conflicts with https://github.com/mitmproxy/mitmproxy/pull/7623 (which is currently in-flight). I think #7623 should eventually fix this as well. I'll leave the PR open here for now to make sure we have this addressed. :)

mhils:
https://github.com/mitmproxy/mitmproxy/pull/7623 is merged and should fix this, thanks again! 🍰

PR #7673, opened 2025-04-25
Bump h11 dependency version for critical security fix

Antiz96:
A critical CVE affecting h11 < 0.16.0 has been published: [CVE-2025-43859](https://github.com/advisories/GHSA-vqfr-h8mv-ghfj)


mhils:
Thanks for the PR! This should not affect mitmproxy as we never keep chunking across connections.

PR #7685, opened 2025-05-02
Load script addon before dns resolver addon

Florigolo:
#### Description

Switch position of dns resolver and script in addon init.py to be able to intercept DNS requests properly.

#### Checklist


mhils (approved):
Thanks!

Issue #7650, opened 2025-04-10
Better error message for WireGuard port conflict


injust:
### Problem Description

I have mitmweb configured to use the WireGuard mode. If I try to run 2 instances, the 2nd instance fails as expected because the port is already used. But the error message could be improved.

#### Steps to reproduce the behavior:
1. `mitmweb --mode wireguard` twice

```
[12:23:18.282] Address already in use (os error 48)
Error logged during startup, exiting...
```

For example, this is the error message you get when the web UI port conflicts:

```
Traceback (most recent call last):
  File "mitmproxy/tools/web/master.py", line 111, in running
  File "tornado/tcpserver.py", line 183, in listen
  File "tornado/netutil.py", line 162, in bind_sockets
OSError: [Errno 48] Address already in use

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "mitmweb.py", line 4, in <module>
  File "mitmproxy/tools/main.py", line 167, in mitmweb
  File "mitmproxy/tools/main.py", line 137, in run
  File "asyncio/runners.py", line 194, in run
  File "asyncio/runners.py", line 118, in run
  File "asyncio/base_events.py", line 720, in run_until_complete
  File "mitmproxy/tools/main.py", line 134, in main
  File "mitmproxy/master.py", line 84, in run
  File "mitmproxy/tools/web/master.py", line 116, in running
OSError: [Errno 48] Web server failed to listen on 127.0.0.1:8081 with [Errno 48] Address already in use
Try specifying a different port by using `--set web_port=8083`.
[PYI-89638:ERROR] Failed to execute script 'mitmweb' due to unhandled exception!
```

### System Information

```raw
Mitmproxy: 11.1.3 binary
Python:    3.13.1
OpenSSL:   OpenSSL 3.4.1 11 Feb 2025
Platform:  macOS-15.4-x86_64-64bit-Mach-O
```

### Checklist

- [x] This bug affects the [latest mitmproxy release](https://github.com/mitmproxy/mitmproxy/releases).

errorxyz:
https://github.com/mitmproxy/mitmproxy/pull/7589 should fix this

turboOrange:
[...]

PR #7688, opened 2025-05-03
Fix better error message for wire guard port conflict

turboOrange:
#### Description

This relates to issue #7650. The current behavior is due to Rust errors not including a traceback by default. When using the PyO3 library, it's expected that exceptions may appear like this. That said, we can still improve the situation by enhancing error handling within the Rust code (mitmproxy_rs) and by adding more context on the Python side when an exception is raised. Below is the crash generated by the current version: `Traceback (most recent call last):
  File "/home/no_more_heroes/projet/mitmproxy/mitmproxy/proxy/mode_servers.py", line 386, in start_udp_based_server
    return await mitmproxy_rs.wireguard.start_wireguard_server(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<6 lines>...
    )
    ^
RuntimeError: Failed to bind UDP socket to 0.0.0.0:51820

Caused by:
    Address already in use (os error 98)
[20:37:25.913] WireGuard failed to start
Error logged during startup, exiting...`

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * In DNS proxy mode, user-provided addons now trigger before DNS resolution has taken place.
   ([#7685](https://github.com/mitmproxy/mitmproxy/pull/7685), @Florigolo)
 * Improve error message for WireGuard port conflict.
+  ([#7688](https://github.com/mitmproxy/mitmproxy/pull/7688), @turboOrange)
 
 ## 29 April 2025: mitmproxy 12.0.0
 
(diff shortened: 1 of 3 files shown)


mhils (changes requested):
Thanks for the PR! I think the better fix here is to properly raise an OSError on the Rust side, which we already handle correctly here:

https://github.com/mitmproxy/mitmproxy/blob/ad00a05d83cc26448bcab8ed5f05234ac6ed6f7a/mitmproxy/proxy/mode_servers.py#L260-L267

Issue #7651, opened 2025-04-10
Concurrent mitmweb instances overwrite each other's `mitmproxy-auth` cookie


injust:
### Problem Description

The default mitmweb UI binds to `http://127.0.0.1:8081`, so I tried to run a 2nd instance concurrently by passing `--set web_port=8082`.

As soon as the 2nd instance authenticates, the 1st instance is unable to load request/response bodies because of a Network Error (Error getting content: 403 Forbidden.)

#### Steps to reproduce the behavior:
1. `mitmweb --mode wireguard`
2. Stop the WireGuard server in the web UI (under Capture)
3. `mitmweb --mode wireguard --set web_port=8082`

### System Information

```raw
Mitmproxy: 11.1.3 binary
Python:    3.13.1
OpenSSL:   OpenSSL 3.4.1 11 Feb 2025
Platform:  macOS-15.4-x86_64-64bit-Mach-O
```

### Checklist

- [x] This bug affects the [latest mitmproxy release](https://github.com/mitmproxy/mitmproxy/releases).

mhils:
I don't think this is inherently fixable (but I'm all ears for suggestions). Some workarounds:

1. Specify the same web password for all instances.
2. Run one mitmproxy instances with multiple WireGuard modes active (you can pass --mode multiple times).
3. Use a different domain, e.g mitmproxy2.localhost or 127.0.0.2.

injust:
Maybe it'd suffice to add a blurb in the docs to say "hey, you're doing something valid, but it's not going to work like you expected because cookies are not isolated by port", try those 3 workarounds.

Since I'm running my mitmweb instance locally and not exposing it, I solved this for myself by setting an empty password using its argon2 hash (from https://github.com/mitmproxy/mitmproxy/issues/7551#issuecomment-2781367454) to my config.

Prinzhorn:
@mhils could you elaborate why this happens? I'm a little confused. These are two separate web apps, running on two separate origins with separate cookies. Why exactly do they conflict at all? The only explanation I could come up with is that the auth is not held in memory but stored in some shared location (`config`?)

injust:
> These are two separate web apps, running on two separate origins with separate cookies.

Not exactly. Both are running on 127.0.0.1 -- only the port is different, and cookies aren't isolated by port.

So when you authenticate on the newest instance, it sets the cookie, replacing the value being used by the older instance.

Prinzhorn:
[...]

PR #7690, opened 2025-05-04
Fix/concurrent mitmweb instances overwrite each others mitmproxy auth cookie

turboOrange:
#### Description

This is a fix for #7651.
The issue was that when two instances were running from the same domain or IP address, they shared cookies because cookies are tied to the domain, not the port. This made it difficult to run multiple instances on the same machine.

The solution proposed in this pull request is to include the port number in the cookie name. This ensures that even if cookies are technically shared at the domain level, they remain distinct and do not conflict.

#### Checklist

### DIFF
--- a/mitmproxy/tools/web/app.py
+++ b/mitmproxy/tools/web/app.py
@@ -213,7 +213,6 @@ class APIError(tornado.web.HTTPError):
 
 
 class AuthRequestHandler(tornado.web.RequestHandler):
-    AUTH_COOKIE_NAME = "mitmproxy-auth"
     AUTH_COOKIE_VALUE = b"y"
 
     def __init_subclass__(cls, **kwargs):
@@ -223,6 +222,11 @@ class AuthRequestHandler(tornado.web.RequestHandler):
             fn = getattr(cls, method)
             if fn is not tornado.web.RequestHandler._unimplemented_method:
                 setattr(cls, method, AuthRequestHandler._require_auth(fn))
+
+    def get_auth_cookie_name(self):
+        host = self.request.host
+        port = host.split(":")[-1]
+        return f"mitmproxy-auth-{port}"
 
     def auth_fail(self, invalid_password: bool) -> None:
         """
@@ -253,7 +257,7 @@ class AuthRequestHandler(tornado.web.RequestHandler):
                     self.auth_fail(bool(password))
                     return None
                 self.set_signed_cookie(
-                    self.AUTH_COOKIE_NAME,
+                    self.get_auth_cookie_name(),
                     self.AUTH_COOKIE_VALUE,
                     expires_days=400,
                     httponly=True,
@@ -265,7 +269,7 @@ class AuthRequestHandler(tornado.web.RequestHandler):
 
     def get_current_user(self) -> bool:
         return (
-            self.get_signed_cookie(self.AUTH_COOKIE_NAME, min_version=2)
+            self.get_signed_cookie(self.get_auth_cookie_name(), min_version=2)
             == self.AUTH_COOKIE_VALUE
         )
 
--- a/mitmproxy/tools/web/app.py
+++ b/mitmproxy/tools/web/app.py
@@ -224,8 +224,8 @@ class AuthRequestHandler(tornado.web.RequestHandler):
                 setattr(cls, method, AuthRequestHandler._require_auth(fn))
 
     def get_auth_cookie_name(self):
-        host = self.request.host
-        port = host.split(":")[-1]
+        _, port = tornado.httputil.split_host_and_port(self.request.host)
+        port = port if port is not None else 80
         return f"mitmproxy-auth-{port}"
 
     def auth_fail(self, invalid_password: bool) -> None:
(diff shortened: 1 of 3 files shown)


mitmproxy/tools/web/app.py
```diff
@@ -221,6 +220,11 @@ def __init_subclass__(cls, **kwargs):
             if fn is not tornado.web.RequestHandler._unimplemented_method:
                 setattr(cls, method, AuthRequestHandler._require_auth(fn))
 
+    def get_auth_cookie_name(self):
+        host = self.request.host
+        port = host.split(":")[-1]
```
injust, line 225: Is `request.host` guaranteed to specify a port?

Also maybe worth using [`tornado.httputil.split_host_and_port()`](https://www.tornadoweb.org/en/stable/httputil.html#tornado.httputil.split_host_and_port) instead.

turboOrange: You're right! It doesn't always specify the port. If it doesn't, it's because it's using port 80, since it's an HTTP server. Fun fact: on Linux, you can't bind to port 80 unless you're root. I added logic to default to port 80 if none is specified. Thanks for the suggestion to use the method from Tornado!

test/mitmproxy/tools/web/test_app.py
```diff
@@ -81,12 +81,16 @@ async def make_master() -> webmaster.WebMaster:
 
     @property
     def auth_cookie(self) -> str:
+        port = self.get_http_port()
+        cookie_name = f"mitmproxy-auth-{port}"
```
injust, line 85: nit: I would just do `cookie_name = f"mitmproxy-auth-{self.get_http_port()}"` here

mitmproxy/tools/web/app.py
```diff
@@ -221,6 +220,11 @@ def __init_subclass__(cls, **kwargs):
             if fn is not tornado.web.RequestHandler._unimplemented_method:
                 setattr(cls, method, AuthRequestHandler._require_auth(fn))
 
+    def get_auth_cookie_name(self):
+        _, port = tornado.httputil.split_host_and_port(self.request.host)
+        port = port if port is not None else 80
```
injust, line 225: The negated conditional is a bit hard to read.

Either `port = 80 if port is None else port`, or remove it entirely and `return f"mitmproxy-auth-{port or 80}"`.

injust:
LGTM aside from some nitpicks.

Putting the port in the cookie name isn't my idea of an _elegant_ solution, but it's probably the best way to solve https://github.com/mitmproxy/mitmproxy/issues/7651.

turboOrange:
> LGTM aside from some nitpicks.
> 
> Putting the port in the cookie name isn't my idea of an _elegant_ solution, but it's probably the best way to solve #7651.

I agree. Even if it's common to use a web page as UI for local tools, the technology is not really made for that so I think it's normal to need some workarounds. For that specific tool, I can really see the appeal of having more then one running. I just can't think of a cleaner solution.

turboOrange:
> > LGTM aside from some nitpicks.
> > Putting the port in the cookie name isn't my idea of an _elegant_ solution, but it's probably the best way to solve #7651.
> 
> I agree. Even if it's common to use a web page as UI for local tools, the technology is not really made for that so I think it's normal to need some workarounds. For that specific tool, I can really see the appeal of having more then one running. I just can't think of a cleaner solution.

Actually, I lie here. I have an other idea but that would involve giving the proxy an API that the web page would interact with. This way, all the instances would be accessible from the same web page. We would have to use some kind of discovery method to automatically connect the web page to the instances. We could explore that alternative in the future. I don't really know how it would work yet.

mitmproxy/tools/web/app.py
```diff
@@ -221,6 +220,10 @@ def __init_subclass__(cls, **kwargs):
             if fn is not tornado.web.RequestHandler._unimplemented_method:
                 setattr(cls, method, AuthRequestHandler._require_auth(fn))
 
+    def get_auth_cookie_name(self):
+        _, port = tornado.httputil.split_host_and_port(self.request.host)
```
mhils, line 224: I don't like to use user-provided data here (`self.request.host`). Can we use `self.master.options.web_port` instead?

mitmproxy/tools/web/app.py
```diff
@@ -221,6 +220,10 @@ def __init_subclass__(cls, **kwargs):
             if fn is not tornado.web.RequestHandler._unimplemented_method:
                 setattr(cls, method, AuthRequestHandler._require_auth(fn))
 
+    def get_auth_cookie_name(self):
```
mhils, line 223:
```suggestion
    @staticmethod
    def auth_cookie_name():
```
(and then reuse in the backend)

mhils (commented):
Thanks for the PR!

mitmproxy/tools/web/app.py
turboOrange: > I don't like to use user-provided data here (`self.request.host`). Can we use `self.master.options.web_port` instead?

It's not available in that context but I found self.application.settings.get("web_port")

Issue #7597, opened 2025-03-10
Docker use /root/.mitmproxy as conf dir if volume mapped /home/mitmproxy/.mitmproxy folder owned by host root user


ulion:
### Problem Description

Related to: https://github.com/mitmproxy/mitmproxy/issues/7156
It seems it was not fixed cleanly.

When use volume mapping, certainly we want ca file be reused when redeploy the docker image, the mitmproxy user's UID (maybe GID too) will be set to same with the owner of mapped /home/mitmproxy/.mitmproxy host folder.

So, if /home/mitmproxy/.mitmproxy owned by host root, the mitmproxy docker guest will set user mitmproxy UID to 0, then the mitmproxy will not use /home/mitmproxy/.mitmproxy, it will use /root/.mitmproxy, which cause the ca file re-generated each time re-deploy the docker image.

This is an undocumented issue, and it is annoy.

I have to start mitmproxy with --set confdir=/home/mitmproxy/.mitmproxy parameter to ensure it use the correct conf dir.

the Dockerfile or entrypoint has to consider what if the conf dir owned by host root, currently they do not, or, should document this unexpected behaviour.

### Checklist

- [x] This bug affects the [latest mitmproxy release](https://github.com/mitmproxy/mitmproxy/releases).

shellvon:
The root cause of this issue lies in how `mitmproxy` loads its configuration file based on the user during startup. In the Docker environment, to ensure that the container has the same permissions as the host without modifying the original permissions, the existing implementation uses [usermod](https://explainshell.com/explain/8/usermod) to adjust the UID/GID of the configuration files to match those on the host:
https://github.com/mitmproxy/mitmproxy/blob/4e8eef805b1620ba8a968d5fa841b46f8ccf3751/release/docker/docker-entrypoint.sh#L15-L19

However, when the UID/GID is set to `0`, the user is treated as `root`, which causes `mitmproxy` to load a different configuration file, leading to the problem

I believe the current permission-handling approach is a good solution (in contrast, fixing a specific UID/GID would modify the original host permissions, which I don’t recommend). Following this logic, we can resolve the issue by consistently specifying the configuration directory when starting the container. Specifically, we can enforce a fixed `confdir` in the `docker-entrypoint.sh` script, as suggested by @ulion.

[...]

PR #7699, opened 2025-05-08
Force confdir to /home/mitmproxy/.mitmproxy when volume is root-owned

shellvon:
Ensure mitmproxy uses /home/mitmproxy/.mitmproxy as the config directory even if the mapped volume is owned by host root, preventing fallback to /root/.mitmproxy and CA file regeneration on redeploy. Add `HOME` env confdir in entrypoint to enforce this behavior.

Fixes mitmproxy/mitmproxy#7597

#### Checklist


mhils (approved):
Thanks! 🎉

PR #7700, opened 2025-05-08
fix display of ipv6 addresses

mhils:


web/src/js/components/FlowView/Connection.tsx
```diff
@@ -14,7 +14,7 @@ export function ConnectionInfo({ conn }: ConnectionInfoProps) {
             <>
                 <tr>
                     <td>Address:</td>
-                    <td>{conn.address?.join(":")}</td>
+                    <td>{conn.address?.slice(0, 2).join(":")}</td>
```
lups2000, line 17: By inspecting the code I noticed that `conn.address` is always an array of 2 elements. The slice is then redundant. Am I missing something?

mhils: Yes. It's actually a 4-tuple for ipv6, the type annotation is wrong. Good call to fix this as well. :)

lups2000: apparently it's always a 2-tuple for ipv6 as well

mhils: I think my PR is just bad. Look at the resolved address, `:0:0` shouldn't be there. 🥲

lups2000 (approved):
Looks good now 🚀

PR #7701, opened 2025-05-09
Fix raw response export incorrectly zeroing non-zero `Content-Length` header for HEAD requests.

### LINKED ISSUE #7695: Content-Length header set to 0 in HEAD requests during Export
### Problem Description

Similar to [this one](https://github.com/mitmproxy/mitmproxy/issues/659). The `content-length` header value appears correctly in the web UI, but then it is incorrectly overwritten to `0` when executing a *raw* export.

1. Send a HEAD request that returns a non-zero `content-length` header value in the response (valid according to RFC)
> The Content-Length entity-header field indicates the size of the entity-body, in decimal number of OCTETs, sent to the recipient or, in the case of the HEAD method, the size of the entity-body that would have been sent had the request been a GET.

2. Observe the non-zero value in the web UI
3. Export the raw response and observe that the `content-length` header value in the response has been overwritten to `0`.

### System Information

```raw

```

### Checklist

sujaldev:
#### Description

Fixes #7695.

#### Checklist

### DIFF
--- a/mitmproxy/addons/export.py
+++ b/mitmproxy/addons/export.py
@@ -39,7 +39,8 @@ def cleanup_response(f: flow.Flow) -> http.Response:
         raise exceptions.CommandError("Can't export flow with no response.")
     assert isinstance(f, http.HTTPFlow)
     response = f.response.copy()  # type: ignore
-    response.decode(strict=False)
+    if response.content:
+        response.decode(strict=False)
     return response
 
 
--- a/test/mitmproxy/addons/test_export.py
+++ b/test/mitmproxy/addons/test_export.py
@@ -274,6 +274,13 @@ class TestRawResponse:
         with pytest.raises(exceptions.CommandError):
             export.raw_response(udp_flow)
 
+    def test_head_non_zero_content_length(self):
+        request = tflow.tflow(
+            req=tutils.treq(method=b"HEAD"),
+            resp=tutils.tresp(headers=((b"content-length", b"7"),), content=b""),
+        )
+        assert b"content-length: 7" in export.raw_response(request)
+
 
 def qr(f):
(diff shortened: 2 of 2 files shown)


mitmproxy/addons/export.py
```diff
@@ -39,7 +39,8 @@ def cleanup_response(f: flow.Flow) -> http.Response:
         raise exceptions.CommandError("Can't export flow with no response.")
     assert isinstance(f, http.HTTPFlow)
     response = f.response.copy()  # type: ignore
-    response.decode(strict=False)
+    if response.content:
+        response.decode(strict=False)
```
mhils, line 43: Should we rather fix `.decode()` to not update .content if .raw_content is empty?

sujaldev: That's what I thought too, but it seems `raw_content` being `None` and `b""` have different semantics and I'm not sure what I'm missing by treating both the same.

mhils: See https://github.com/mitmproxy/mitmproxy/pull/7702. We can just check for truthiness in this case.

sujaldev:
Not sure why coverage is complaining, all the lines I changed are covered.

mitmproxy/http.py
```diff
@@ -469,6 +469,8 @@ def decode(self, strict: bool = True) -> None:
         *Raises:*
          - `ValueError`, when the content-encoding is invalid and strict is True.
         """
+        if not self.raw_content:
+            return
```
mhils, line 473: Let's add tests for this in test_http.py:

1. One response with missing body (e.g. body streaming)
2. One response with empty body (e.g. a HEAD response)

mhils (approved):
Very nice, thank you! :cake:

PR #7692, opened 2025-05-04
web: increase typescript use for ducks

mhils:
#### Description

In preparation for https://github.com/mitmproxy/mitmproxy/pull/7617, this PR adds full typescript typing for all redux actions emitted server-side.

Commit of mitmproxy/tools/web/static/app.js is intentional as backend message format changed; otherwise mitmweb is broken on main.

#### Checklist

 - [x] I have updated tests where applicable.

PR #7617, opened 2025-03-22
Show/Hide flows on the client side

lups2000:
#### Description

In this PR, I have implemented the client side logic to show/hide the incoming flows based on the information that we get from the server. This should be the last part for finally completing this issue: #3609.
Once the code looks good, I will proceed with writing tests!

#### Checklist

 - [ ] I have updated tests where applicable.
 - [ ] I have added an entry to the CHANGELOG.

PR #7704, opened 2025-05-11
Web: search/highlight by body content

lups2000:
#### Description

In this PR, I'm implementing search and highlight by body content (#3609) in mitmweb.
This is a significant change in mitmweb as it updates how filtering and highlighting work. Previously, the logic was handled entirely on the frontend with the use of **filt.js**, but we're now moving it to the backend. 

Current status:
The backend logic has been implemented and already reviewed. I've just resolved conflicts and updated everything according to the new ducks structure (#7692). I also added directly here the last stuff that was implemented in #7617.
Once the code looks good, I will proceed with the last tests to improve the overall coverage.

Instead of continuing work in that old PR, I decided to open this new one directly to main and continue here; since this is what we'll eventually merge 😅.
Let me know what you think! :)

#### Checklist


web/src/js/ducks/flows/index.ts
```diff
@@ -0,0 +1,363 @@
...
+                    .map((id) => state.byId.get(id))
+                    .filter((f) => f !== undefined)
+                    .toSorted(makeSort(state.sort));
+                const _viewIndex = buildIndex(view);
+                return {
+                    ...state,
+                    view,
+                    _viewIndex,
+                };
+            }
```
lups2000, line 254: We have a bug here: if we apply a filter and then clear it, no flows are shown because `matching_flow_ids` is set to `[]`. Previously, I think I was also passing the filter expression, and if it was empty, it meant that `matching_flow_ids` could be ignored. Otherwise, we could pass `None` or `undefined` for the `matching_flow_ids`.

mhils: Great catch, fixed!

web/src/js/ducks/flows/index.ts
```diff
@@ -0,0 +1,366 @@
...
+        const { sort, selected, selectedIds } = state;
+        let { view, _viewIndex, highlighted } = state;
+        // Update list
+        const _listIndex = new Map(state._listIndex);
+        _listIndex.set(flow.id, state.list.length);
+        const list = [...state.list, flow];
+        const byId = new Map(state.byId);
+        byId.set(flow.id, flow);
+        // Update view
+        if (matching_filters[FilterName.Search] !== false) {
```
lups2000, line 110: for readability I would stick to `=== true` like below or simply `if (matching_filters[FilterName.Search])`

web/src/js/ducks/flows/index.ts
```diff
@@ -0,0 +1,366 @@
...
+        // Update list
+        const listPos = state._listIndex.get(flow.id)!;
+        const list = [...state.list];
+        list[listPos] = flow;
+        const byId = new Map(state.byId);
+        byId.set(flow.id, flow);
+        // Update view
+        const prevViewPos: number | undefined = _viewIndex.get(flow.id);
+        const hasOldFlow = prevViewPos !== undefined;
+        const hasNewFlow = !(matching_filters[FilterName.Search] === false);
```
lups2000, line 148: why not simply `const hasNewFlow = matching_filters[FilterName.Search] !== false` ?

web/src/js/ducks/flows/index.ts
```diff
@@ -0,0 +1,366 @@
...
+            if (selected.length === 1 && viewPos !== undefined) {
+                const fallback =
+                    view[viewPos /* no +1, already removed */] ??
+                    view[viewPos - 1];
+                selected = [fallback];
+            } else {
+                selected = selected.filter((f) => f.id !== flow_id);
+            }
+            selectedIds = buildLookup(selected);
+        }
```
lups2000, line 225: There is a problem here that arises when removing the last flow. Basically selected is equal to `[undefined]` and `buildLookup(selected)` crashes.
I think this check might help: `selected = fallback ? [fallback] : []`, but it doesn't fully solve the problem. If you try to clear all the flows, we no longer see the `mitmproxy is running...` message, which suggests that the `list` of flows is not actually empty and still contains elements. I don't have time to investigate further right now, but I can take a closer look later today or tomorrow morning.

web/src/js/ducks/flows/index.ts
mhils: https://github.com/mitmproxy/mitmproxy/pull/7704/commits/d28bd5ec4a7de1baffcf5b3a1c0b58e69de7b8f6 :)

web/src/js/ducks/flows/index.ts
mhils: Fixed somewhere in https://github.com/mitmproxy/mitmproxy/pull/7704/commits/bd7b35c7885bc1f25ba307af884205b59e0bcfc1! :)

web/src/js/ducks/flows/index.ts
```diff
@@ -0,0 +1,375 @@
...
+            selectedIds,
+            highlighted,
+        };
+    } else if (FLOWS_REMOVE.match(action)) {
+        const flow_id = action.payload;
+        const { sort } = state;
+        let { view, _viewIndex, selected, selectedIds } = state;
+        const listPos = state._listIndex.get(flow_id)!;
+        const list = state.list.toSpliced(listPos, 1);
+        const _listIndex = withKeyRemoved(state._listIndex, flow_id);
```
lups2000, line 208: There's still something not working as expected. When we have only one element and remove it, `list` doesn't become `[]`

mhils: How do you remove the last flow? Seems to work for me.

lups2000: This happens if we previously had more than one flow. For example, if we add 3 flows and then remove them one by one from the bottom up, when removing the last one (at the top), `list` still contains that top flow.

lups2000: https://github.com/user-attachments/assets/4c0a1602-f035-41dd-a9d3-d9e40a0e284f

lups2000 (commented):
I went through the code once again, and it looks good to me! Just left a few nits :)
I also tested it, and everything seems to be working fine 🚀

web/src/js/ducks/flows/index.ts
```diff
@@ -0,0 +1,373 @@
...
+        const _listIndex = new Map(state._listIndex);
+        _listIndex.set(flow.id, state.list.length);
+        const list = [...state.list, flow];
+        const byId = new Map(state.byId);
+        byId.set(flow.id, flow);
+        // Update view
+        if (
+            matching_filters[FilterName.Search] === true ||
+            matching_filters[FilterName.Search] === undefined
+        ) {
```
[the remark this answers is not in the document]
mhils: Disagree - this now spells out more clearly which cases we are handling, the previous version disguised that a bit with a clever trick. I'll extend the docstring to "Update view if filter matches (true) or is unset (undefined).". :)

web/src/js/ducks/flows/index.ts
```diff
@@ -0,0 +1,373 @@
...
+            ({ view, _viewIndex } = insertViewItem(
+                view,
+                _viewIndex,
+                flow,
+                makeSort(sort),
+            ));
+        }
+        // Update highlight
+        if (matching_filters[FilterName.Highlight] === true) {
```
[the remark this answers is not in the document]
mhils: Disagree, see above. :)

web/src/js/ducks/flows/index.ts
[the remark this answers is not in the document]
lups2000: okay okay fine :)

web/src/js/ducks/flows/index.ts
```diff
@@ -0,0 +1,373 @@
...
+        }
+        // Update selection
+        if (selectedIds.has(flow.id)) {
+            selected = selected.map((existing) =>
+                existing.id === flow.id ? flow : existing,
+            );
+        }
+        // Update highlight
+        if (matching_filters[FilterName.Highlight]) {
+            if (!highlighted.has(flow.id)) {
```
lups2000, line 180: here then we can specify `===true` as well

web/src/js/ducks/flows/index.ts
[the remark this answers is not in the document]
lups2000: added a comment below, so we stick to this version :)

web/src/js/ducks/flows/index.ts
```diff
@@ -0,0 +1,373 @@
...
+        const list = action.payload;
+        const _listIndex = buildIndex(list);
+        const byId = new Map(list.map((f) => [f.id, f]));
+        // No filter information yet, we expect that to come immediately after.
+        const view = list.toSorted(makeSort(sort));
+        const _viewIndex = buildIndex(view);
+        const selected = state.selected
+            .map((flow) => byId.get(flow.id))
+            .filter((f) => f !== undefined);
+        const selectedIds = buildLookup(selected);
```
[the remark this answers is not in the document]
mhils: No, clearing the flow list with File -> Clear All.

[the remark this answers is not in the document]
mhils: Ok, you _also_ get FLOWS_RECEIVE when loading a new set of flows. Which may have an overlap with the current set of flows. 😉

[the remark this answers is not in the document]
lups2000: Okay I see 😅

mhils (approved):
image

PR #7710, opened 2025-05-14
flowfilter: types++, add help strings

mhils:
Tests still need to cover the remaining `__str__` implementations.


lups2000:
If it's fine with you, I could take over and finalize the third step for rejuvenating mitmweb: **removing Filt.js** and moving descriptions and validation to the backend side (reusing the logic we implemented earlier for body-search filters). I think this PR started in this direction so it seems a good starting point! What do you think?

PR #7719, opened 2025-05-21
Do not add CRLF when setting x-manual-chunk in chunked transfer-encod…

gachikuku:
Sometimes the user wants to perform a request smuggling attack which requires 
modifying the request to the binary.

with this addon, it prevents the automatic addition of 0\r\n\r\n when using `Tranfer-Encoding: chunked`
together with `x-manual-chunk`

heres a [discussion](https://github.com/mitmproxy/mitmproxy/discussions/7462)

Please feel free to modify the code further.

This is just a *just works* script that I use.

### DIFF
--- a/examples/addons/chunkcontrol.py
+++ b/examples/addons/chunkcontrol.py
@@ -1,0 +1,81 @@
+"""Do not add CRLF when setting x-manual-chunk in chunked transfer-encoding."""
+
+from mitmproxy import http
+from mitmproxy.proxy.layers.http import _http1
+
+# Save the original send methods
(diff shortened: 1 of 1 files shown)


mhils:
Thanks for the PR! This is a cool example, but also a bit of a maintenance burden. Generally speaking, my recommendation for request smuggling research would be to force TCP mode (https://docs.mitmproxy.org/dev/concepts/options/#tcp_hosts or https://docs.mitmproxy.org/dev/concepts/modes/#protocol-specification) instead. It currently does not support replay... but that's something we could fix? 😅

PR #7724, opened 2025-05-23
Add portfile support

### LINKED ISSUE #7720: Support creating a pidfile/portfile
#### Problem Description

When driving mitmproxy / mitmdump externally, it can be useful to tell mitmproxy to use port `0` to ensure there is no port conflict and be able to run multiple instances concurrently.

While mitmproxy will bind fine,

- it can be difficult to know when it's started, stable, and ready to be interacted with
- it only outputs the port to stdout
- if the output mode is default or verbose (no output in quiet mode)
- and stdout being fully buffered makes it difficult to retrieve (the only way I've found is to run mitmdump under `PYTHONUNBUFFERED=1`)

Furthermore, once the port has been parsed out it requires additional infrastructure to ensure the pipe is read from to avoid the risk of mitmproxy blocking (no sure it actually happens in this case).

#### Proposal

The ability to specify a file to which (colloquially a pidfile, although in this case it wouldn't contain the pid, or not *just* the pid). Or a directory in which mitmproxy will create a well named file (that might be a bit less tricky in terms of ACLs). The file can be structured in order to provide more content (e.g. *both* pid and port), but ideally should be written atomically in order to ensure clients polling for the file can't do partial reads.

#### Prior Art

Under certain conditions, [Chrome creates a file called `DevToolsActivePort`](https://chromedevtools.github.io/devtools-protocol/#how-do-i-access-the-browser-target) in the user data dir.

xmo-odoo:
Convenient when manipulating mitmproxy programmatically, as it's often useful to specify port 0 (for all sorts of servers) but then the user needs a way to retrieve the actual port being used.

mitmproxy logs the bound ports out by default, however because it logs to the fully buffered stdout *and* can generate quite a bit of log, this tends to be inconvenient. Writing out a file at a user-specified location makes it much easier for them to poll, and using json even if there is a minor race the user can just keep reading until they see a valid JSON file.

Closes #7720

### DIFF
--- a/examples/contrib/portfile.py
+++ b/examples/contrib/portfile.py
@@ -1,0 +1,40 @@
+import json
+import pathlib
+from typing import Optional
+
+from mitmproxy import ctx
+
+
+class PortFile:
+    def load(self, loader):
+        loader.add_option(
+            name="datadir",
+            typespec=Optional[str],
+            default=None,
+            help="Creates `portfile` mapping proxies (by mode spec) to the port "
+            "they use in the provided directory.",
+        )
+
+    def running(self):
+        if not ctx.options.datadir:
+            return
+
+        datadir = pathlib.Path(ctx.options.datadir)
+        if not datadir.is_dir():
+            ctx.log.warning("%s is not a directory", datadir)
+            return
+
+        proxies = ctx.master.addons.get("proxyserver")
+        modemap = {
+            instance.mode.full_spec: addr[1]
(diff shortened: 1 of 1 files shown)


mhils (approved):
Very cool, thank you! If we see more upvotes on https://github.com/mitmproxy/mitmproxy/issues/7720 over time, I'd be willing to consider moving this directly into mitmproxy/addons. But for now this is perfect. 😃

PR #7729, opened 2025-05-25
web: account for race conditions during flow fetching

mhils:
#### Description

This should fix a regression in 12.1 when loading an existing flows file from disk. :)

#### Checklist


lups2000 (approved):
Yeah this works fine! 🚀

PR #7750, opened 2025-06-01
Web: interactive content views

lups2000:
#### Description

In this PR, I removed the edit/save toggle that previously allowed users to switch between viewing and editing a content view. Instead, I'm now using the CodeMirror editor directly, so users can modify the content inline. They can save their changes either by clicking **"Save"** or by pressing `Ctrl + S` (`Cmd + S`). Additionally, the **"Show more"** button has been removed; the full content is now always displayed by default to make the experience more seamless.
Demo:

https://github.com/user-attachments/assets/3bb154c2-b6ba-494c-9a2a-2a9f333819ec

I still need to modify and update the tests. First impressions?

#### Checklist


lups2000:
web tests fail here, but locally not. Do you have any idea why this happens?

PR #7752, opened 2025-06-02
Fix type for `Master.opts` to accept `None`

injust:
#### Description

I think the type hint for `Master.opts` is missing a `| None`. There's logic to handle a falsey value, and that's how this test instantiates `Master`:

https://github.com/mitmproxy/mitmproxy/blob/b5ac8e62ea633a7a7211d735fabe6008fb64ece0/test/mitmproxy/test_master.py#L15

#### Checklist

### DIFF
--- a/mitmproxy/master.py
+++ b/mitmproxy/master.py
@@ -26,7 +26,7 @@ class Master:
 
     def __init__(
         self,
-        opts: options.Options,
+        opts: options.Options | None,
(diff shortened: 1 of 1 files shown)


injust:
I encountered this while trying to run mitmproxy from an async service by using `Master.run()`.

mhils:
Thanks!

PR #7756, opened 2025-06-06
Web: Setting custom name when saving flows

lups2000:
#### Description

In this PR, I’m adding support for setting a custom file name when saving flows.
DEMO:

https://github.com/user-attachments/assets/0e37a912-2e73-47e4-9d1c-5f52d0680933

#### Checklist


lups2000:
Do you still think this change might be useful? otherwise we can close the PR :)

Issue #7476, opened 2025-01-14
TLS mitmproxy error:   `00D020E001000000:error:0A000126:SSL routines:ssl3_read_n:unexpected eof while reading:ssl/record/rec_layer_s3.c:323:`


xujimu:
#### Problem Description
I am using `openssl s_client` to perform packet capture testing with `mitmdump`.
```
openssl s_client -connect 192.168.1.123:666 -CAfile /Users/xjm/.mitmproxy/mitmproxy-ca-cert.pem -proxy localhost:8080
```
This `192.168.1.123` is my locally set up TLS server, and the Go code is as follows:
```
package main

import (
    "crypto/tls"
    "fmt"
    "io"
    "log"
    "net"
)

func main() {
    cert, err := tls.LoadX509KeyPair("ert.key")
    if err != nil {
       log.Fatalf("load err: %v", err)
    }
    config := &tls.Config{Certificates: []tls.Certificate{cert}}

    listener, err := tls.Listen("tcp", ":666", config)
    if err != nil {
       log.Fatalf("err: %v", err)
    }
    defer listener.Close()
    fmt.Println("list 666")

    for {
       conn, err := listener.Accept()
       if err != nil {
          log.Printf("err: %v", err)
          continue
       }
       go handleConnection(conn)
    }
}

func handleConnection(conn net.Conn) {
    defer conn.Close()
    buf := make([]byte, 512)
    n, err := conn.Read(buf)
    if err != nil && err != io.EOF {
       log.Printf("read: %v", err)
       return
    }
    fmt.Printf("data: %s\n", string(buf[:n]))
    conn.Write([]byte("Hello from TLS server"))
}

```
The `mitmproxy` code is as follows:
```
    print("server_connected")
def server_disconnected(data: mitmproxy.proxy.server_hooks.ServerConnectionHookData):
    print("server_disconnected")
def server_connect_error(data: mitmproxy.proxy.server_hooks.ServerConnectionHookData):
    print("server_connect_error")


def tcp_start(flow: mitmproxy.tcp.TCPFlow):
    print("tcp_start")
def tcp_message(flow: mitmproxy.tcp.TCPFlow):
    print("tcp_message")
def tcp_end(flow: mitmproxy.tcp.TCPFlow):
    print("tcp_end")
def tcp_error(flow: mitmproxy.tcp.TCPFlow):
    print("tcp_error")
def tls_clienthello(data: mitmproxy.tls.ClientHelloData):
[...]

PR #7781, opened 2025-06-19
Issue7476

makrinaemad:
#### Description

Fix(tls): Send TLS close_notify .

Mitmproxy's TLS layer previously did not send a `close_notify` alert
when closing TLS connections,
which led to "unexpected EOF while reading" errors in strict clients such as `openssl s_client` .

This commit fixes the issue by explicitly sending a `close_notify`record.
This ensures that TLS sessions are terminated cleanly.

Fixes: #7476

#### Checklist

 - [ x ] I have added an entry to the CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
   ([#7734](https://github.com/mitmproxy/mitmproxy/pull/7734), @lups2000)
 - Add syntax highlighting for CSS and JavaScript contentviews.
   ([#7749](https://github.com/mitmproxy/mitmproxy/pull/7749), @mhils)
+- Fix TLS to send close_notify alerts and avoid unexpected EOF errors with strict clients like OpenSSL s_client.
 
 ## 25 May 2025: mitmproxy 12.1.1
 
--- a/mitmproxy/proxy/layers/tls.py
+++ b/mitmproxy/proxy/layers/tls.py
@@ -241,6 +241,7 @@ class TlsFailedServerHook(StartHook):
 
 class TLSLayer(tunnel.TunnelLayer):
     tls: SSL.Connection = None  # type: ignore
+    peer_sent_close_notify: bool = False
     """The OpenSSL connection object"""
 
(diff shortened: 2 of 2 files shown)


errorxyz (approved):
LGTM

errorxyz:
Shouldn't we handle the `close_notify` here as well?
https://github.com/mitmproxy/mitmproxy/blob/a4d794c59a27472d193a592d8037505a1cf6ae93/mitmproxy/proxy/layers/tls.py#L455

makrinaemad:
Thank you for the review, @errorxyz!

I don’t think close_notify should be handled within the send_data() context. This block (except (SSL.ZeroReturnError, SSL.SysCallError)) is meant to catch cases where the peer unexpectedly closes the connection or a transmission error occurs.

Since send_data() is part of the active write phase, inserting a close_notify alert here would violate the TLS state machine — alerts must only be sent after the final successful SSL_write, as per RFC 8446.

In TLS 1.3, it is valid for the peer not to send a close_notify. We detect that on the read side via SSL.ZeroReturnError, and the recent changes explicitly track this in self.peer_sent_close_notify. If no alert is received, we synthesize one during send_close() to maintain spec compliance. So, I think no additional handling is needed in send_data().

errorxyz:
> alerts must only be sent after the final successful SSL_write, as per RFC 8446.

I can't seem to find this specified in the [RFC](https://datatracker.ietf.org/doc/html/rfc8446). Can you please provide a reference?

Also, as per [section 6.1](https://datatracker.ietf.org/doc/html/rfc8446#section-6.1),
> Note that this is a change from versions of TLS prior to TLS 1.3 in
   which implementations were required to react to a "close_notify" by
   discarding pending writes and sending an immediate "close_notify"
   alert of their own.
   
 we should be immediately sending a close_notify for TLS < 1.3.

makrinaemad:
Yes, you're right, thanks @errorxyz !

Updated: we now handle close_notify on ZeroReturnError in the send_data function for TLS < 1.3.

PR #7782, opened 2025-06-20
mitmweb: Add local timezone display option for timing info (fixes #7773)

### LINKED ISSUE #7773: mitmweb: show datetime with local timezone
Is there a way to specify the timezone in mitmweb? It uses UTC time by default and I can't find a way to change it.

Image

AyushPatiTripathi:
mitmweb: Add local timezone display option for timing info (fixes #7773)

## Description
Adds an option to display timestamps in the Timing tab using the local timezone instead of UTC. Users can toggle between UTC and local time via the Options modal.

## Changes
- Added preferences reducer to UI state with timezone setting
- Modified `formatTimeStamp` utility to support UTC/local conversion
- Updated Timing component to respect timezone preference
- Added UI control in OptionModal for timezone selection

## Testing
- Verified UTC and local time display in Timing component
- Tested with various flow types (HTTP, TCP, UDP)
- Confirmed setting persistence across modal open/close
- Checked formatting consistency across different timezones

Closes #7773.

### DIFF
--- a/web/src/js/components/FlowView/Timing.tsx
+++ b/web/src/js/components/FlowView/Timing.tsx
@@ -1,7 +1,7 @@
 import { Flow } from "../../flow";
 import * as React from "react";
 import { formatTimeDelta, formatTimeStamp } from "../../utils";
-import { useAppSelector } from '../../../ducks/hooks'; // Correct import path
+import { useAppSelector } from "../../../ducks/hooks"; // Correct import path
 
 export type TimeStampProps = {
     t: number;
@@ -9,21 +9,24 @@ export type TimeStampProps = {
     title: string;
 };
 
-export function TimeStamp({ t, deltaTo, title, timezone }: 
-    TimeStampProps & { timezone: 'utc' | 'local' }) 
-{
+export function TimeStamp({
+    t,
+    deltaTo,
+    title,
+    timezone,
+}: TimeStampProps & { timezone: "utc" | "local" }) {
     if (!t) return null;
-    
+
     return (
         <tr>
             <td>{title}:</td>
             <td>
                 {formatTimeStamp(t, { timezone })}
-                {deltaTo && 
+                {deltaTo && (
                     <span className="text-muted">
                         ({formatTimeDelta(1000 * (t - deltaTo))})
                     </span>
-                }
+                )}
             </td>
         </tr>
     );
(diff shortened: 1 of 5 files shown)


web/src/js/components/FlowView/Timing.tsx
```diff
@@ -1,40 +1,46 @@
...
             </td>
         </tr>
-    ) : (
-        <tr />
     );
 }
 
 export default function Timing({ flow }: { flow: Flow }) {
-    let ref: number;
-    if (flow.type === "http") {
```
Prinzhorn, line 31: Please remove unrelated changes, ternary expressions are almost always harder to read.

web/src/js/components/FlowView/Timing.tsx
```diff
@@ -1,40 +1,46 @@
 import { Flow } from "../../flow";
 import * as React from "react";
 import { formatTimeDelta, formatTimeStamp } from "../../utils";
+import { useAppSelector } from "../../../ducks/hooks"; // Correct import path
```
Prinzhorn, line 4: Remove unnecessary comment

web/src/js/components/FlowView/Timing.tsx
```diff
@@ -1,40 +1,46 @@
...
-        ref = flow.client_conn.timestamp_start;
-    }
+    // Determine reference timestamp
+    const ref =
+        flow.type === "http"
+            ? flow.request.timestamp_start
+            : flow.client_conn.timestamp_start;
 
-    const timestamps: Partial<TimeStampProps>[] = [
+    // Build timestamps array
```
Prinzhorn, line 42: Remove unnecessary comment

web/src/js/components/FlowView/Timing.tsx
```diff
@@ -71,51 +77,57 @@ export default function Timing({ flow }: { flow: Flow }) {
             deltaTo: ref,
         },
     ];
+
+    // Add HTTP-specific timestamps
```
Prinzhorn, line 81: Remove unnecessary comment (I'll stop repeating this comment, please remove all comments that don't add any value)

web/src/js/ducks/ui/index.ts
```diff
@@ -5,11 +5,31 @@ import optionsEditor from "./optionsEditor";
 import tabs from "./tabs";
 import filter from "./filter";
 
-// TODO: Just move ducks/ui/* into ducks/?
+const initialPreferencesState = {
+    timezoneDisplay: "utc" as "utc" | "local",
```
Prinzhorn, line 9: I'd default to local time, because I don't think most people want to see UTC. It feels more natural to work in local time.

web/src/js/ducks/ui/index.ts
```diff
@@ -5,11 +5,31 @@ import optionsEditor from "./optionsEditor";
 import tabs from "./tabs";
 import filter from "./filter";
 
-// TODO: Just move ducks/ui/* into ducks/?
```
Prinzhorn, line 8: Why was this comment removed?

Prinzhorn (changes requested):
Thanks a lot! I gave it a quick review. Most of my requested changes are just AI things (unnecessary changes and A LOT of useless comments).

Prinzhorn:
Please also include screenshots or even a video in your description. This makes it easier for maintainers to get a feeling for the changes without having to check it out locally.

Prinzhorn:
If we want to keep the option, it should probably be a mitmproxy option and not only in the web UI.

mitmproxy:

image

mitmweb:

image

But as I said in the original issue I don't think UTC  is desired at all. It would be good enough if mitmweb would just use local time like mitmproxy.

https://github.com/mitmproxy/mitmproxy/blob/72940c3eaf9ed328cdd485b770fdfe9f95a095c4/mitmproxy/tools/console/flowdetailview.py#L146

https://github.com/mitmproxy/mitmproxy/blob/72940c3eaf9ed328cdd485b770fdfe9f95a095c4/mitmproxy/tools/console/flowdetailview.py#L13

https://github.com/mitmproxy/mitmproxy/blob/72940c3eaf9ed328cdd485b770fdfe9f95a095c4/mitmproxy/utils/human.py#L75-L77

lups2000:
thanks for the PR! I also think that local time is enough. No need to switch :)

web/src/js/components/FlowView/Timing.tsx
AyushPatiTripathi: thanks for your input i will commit the changes ASAP

AyushPatiTripathi:
> Please also include screenshots or even a video in your description. This makes it easier for maintainers to get a feeling for the changes without having to check it out locally.

Will keep this in my mind and add an video /sceenshot in the pr

lups2000:
I have simplified things in #7804 :) feedback is welcome!

Prinzhorn:
https://github.com/mitmproxy/mitmproxy/pull/7804 was merged

PR #7799, opened 2025-07-12
feat: add menu_select_keys option

chedieck:
#### Description

Adds a new option `menu_select_keys to replace` the previously hardcoded shortcut characters (`"123456789abcdefghijklmnoprstuvwxyz"`) used in interactive menus.

This allows users to customize the keys used to select menu items, rather than being limited to the default 123456789ab... sequence.

A minimum length of 34 characters is enforced to match the original behavior. I'm not certain whether this limit is functionally necessary or simply a legacy safeguard — feedback on that is welcome.

Default behavior remains unchanged unless the option is explicitly set.

#### Checklist

 - [  ] I have updated tests where applicable.
 - [  ] I have added an entry to the CHANGELOG.


mhils:
Thank you for the PR! Can you explain what's the motivation here? Under which circumstances are the default keys an issue?

chedieck:
> Thank you for the PR! Can you explain what's the motivation here? Under which circumstances are the default keys an issue?

Sure!

I felt the need to change it while repeatedly modifying a request’s query to probe the endpoint’s behavior. I kept missing the `6` key, since it’s relatively far from the resting `ASDF` and `JKL;` keys. I'd often hit `5` or `7` by mistake, which meant I had to cancel and try again, something that caused me considerable frustration.

mhils:
> I felt the need to change it while repeatedly modifying a request’s query to probe the endpoint’s behavior. I kept missing the `6` key, since it’s relatively far from the resting `ASDF` and `JKL;` keys. I'd often hit `5` or `7` by mistake, which meant I had to cancel and try again, something that caused me considerable frustration.

I really appreciate you going the extra mile in implementing this change and opening a PR. But I think the added complexity from this change outweighs its benefits. For your personal workflow, you can totally use an addon that monkeypatches `ChooserListWalker.shortcuts`. Would that work for you?

chedieck:
> I really appreciate you going the extra mile in implementing this change and opening a PR. But I think the added complexity from this change outweighs its benefits. For your personal workflow, you can totally use an addon that monkeypatches `ChooserListWalker.shortcuts`. Would that work for you?

I see your point, and I get that the added complexity might not be worth it. I still think there's some value in exposing this option, since it makes the UI more flexible and customizable without breaking anything, but I understand the tradeoff.

If it’s not something that fits the project goals right now, I’m fine with monkeypatching it on my side. Thanks for reviewing.

mhils:
I think this isn't worth the complexity at the moment. Thank you again for the proposal though, appreciate it!

PR #7804, opened 2025-07-14
Web: Show Local timezone in the Timing tab

### LINKED ISSUE #7773: mitmweb: show datetime with local timezone

lups2000:
#### Description

Fixes: #7773. 
One note: I had to explicitly add `TZ=UTC` in the **test script** because keeping it in `setup-jest.js` didn’t work for some reason.

Screenshot 2025-07-14 at 20 52 19

#### Checklist

### DIFF
--- a/web/src/js/utils.ts
+++ b/web/src/js/utils.ts
@@ -35,9 +35,19 @@ export const formatTimeStamp = function (
     seconds: number,
     { milliseconds = true } = {},
 ) {
-    const utc = new Date(seconds * 1000);
-    let ts = utc.toISOString().replace("T", " ").replace("Z", "");
-    if (!milliseconds) ts = ts.slice(0, -4);
+    const local = new Date(seconds * 1000);
+
+    const year = local.getFullYear();
+    const month = String(local.getMonth() + 1).padStart(2, "0");
+    const day = String(local.getDate()).padStart(2, "0");
+    const hours = String(local.getHours()).padStart(2, "0");
+    const minutes = String(local.getMinutes()).padStart(2, "0");
+    const secondsPart = String(local.getSeconds()).padStart(2, "0");
+    const millisPart = String(local.getMilliseconds()).padStart(3, "0");
+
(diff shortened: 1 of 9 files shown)


Prinzhorn:
I tried to push to your PR branch but instead pushed a new branch lol, here's how to make `TZ` work:

https://github.com/mitmproxy/mitmproxy/commit/10bbe6d0726e85b11e3a4cc30b539ecdf749af74

lups2000:
> I tried to push to your PR branch but instead pushed a new branch lol, here's how to make `TZ` work:
> 
> [10bbe6d](https://github.com/mitmproxy/mitmproxy/commit/10bbe6d0726e85b11e3a4cc30b539ecdf749af74)

thank you! Updated :)

web/src/js/utils.ts
```diff
@@ -35,9 +35,19 @@ export const formatTimeStamp = function (
...
+    const month = String(local.getMonth() + 1).padStart(2, "0");
+    const day = String(local.getDate()).padStart(2, "0");
+    const hours = String(local.getHours()).padStart(2, "0");
+    const minutes = String(local.getMinutes()).padStart(2, "0");
+    const secondsPart = String(local.getSeconds()).padStart(2, "0");
+    const millisPart = String(local.getMilliseconds()).padStart(3, "0");
+
+    let ts = `${year}-${month}-${day} ${hours}:${minutes}:${secondsPart}`;
+    if (milliseconds) ts += `.${millisPart}`;
+
```
Prinzhorn, line 50: I think the variable names are confusing. You have `days` and `minutes` and then there's `milliseconds`, which is a boolean option, but has the same naming convention. Then there's also the `Part` suffix. Please make this consistent.

Prinzhorn (changes requested):

web/src/js/utils.ts
lups2000: done :)

Prinzhorn (approved):

Prinzhorn:
thanks!

PR #7810, opened 2025-07-31
fix missing content-length header in curl export

### LINKED ISSUE #7664: Export function strips the "Content-Length: 0" header from cURL/HTTPie
#### STEPS TO REPRODUCE:

1. start intercepting traffic 

2. run this on your Bash shell
`curl -H 'Host: example.com' -H 'User-Agent: Thisisatest' -H 'Accept: */*' -H 'Content-Length: 0' -H 'Banana: 0' http://example.com/`

3. the captured packet will show correctly in the web interface, but:
- doing `Flow -> Export -> Copy as cURL` (or `Copy as HTTPie`) will incorrectly omit `-H 'Content-Length: 0'` from the copied command
- meanwhile `-H 'Banana: 0'` gets correctly copied to the clipboard
- the other export options correctly include both `Content-Length: 0` and `Banana: 0`

Tested on `mitmproxy 11.1.3`. Possibly related to https://github.com/mitmproxy/mitmproxy/pull/3676

### System Information

```raw

```

### Checklist

mheguy:
#### Description

closes #7664

curl does not generate a content-length header when not given content. This can cause issues with some servers as they require a content-length header for certain http verbs (ex. POST). The changes in this PR adds some logic to add a 0-length content-length when needed.

I changed the signature of `pop_headers` to reflect that it changes the state of the received object in-place.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## Unreleased: mitmproxy next
 
+- fix: missing content-length header in curl export
 - fix: update log message with correct header name
   ([#7802](https://github.com/mitmproxy/mitmproxy/pull/7802), @kristof-mattei)
 - Update deprecated `windows-2019` runner to `windows-2025`.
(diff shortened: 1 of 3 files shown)


mhils (approved):
Thank you, also for the cleanup! 🍰

PR #7830, opened 2025-08-08
Remove deprecated event_loop fixture

errorxyz:


errorxyz:
Note to self: the `ModuleNotFoundError _curses` error in windows seems to come up for `hypothesis>=6.131.1`.

test/mitmproxy/tools/test_main.py
```diff
@@ -1,12 +1,9 @@
-import asyncio
-
 from mitmproxy.tools import main
 
 shutdown_script = "mitmproxy/data/addonscripts/shutdown.py"
 
 
-def test_mitmweb(event_loop, tdata):
-    asyncio.set_event_loop(event_loop)
+def test_mitmweb(tdata):
```
errorxyz, line 6: I'm not 100% confident if this is the right fix.

pyproject.toml
```diff
@@ -55,6 +55,7 @@ dependencies = [
     "typing-extensions>=4.13.2,<=4.14; python_version < '3.13'",
     "urwid>=2.6.14,<=2.6.16",
     "wsproto>=1.0,<=1.2.0",
+    "windows-curses==2.4.1; sys.platform == 'win32'",
```
errorxyz, line 58: for windows-curses<2.4.1, python 3.13 is not supported

mhils:
Can you clarify why we need a new dependency here?

errorxyz:
this test fails with a `ModuleNotFoundError: No module named '_curses'`: https://github.com/mitmproxy/mitmproxy/actions/runs/16829545375/job/47679608941
This error appears for hypothesis>=6.131.1 

In hindsight, should I have added this new dependency in the `dev` dependency-groups?

errorxyz:
> Can you reproduce on other OSes with `--hypothesis-seed=283474802224946286649981002652669788455`

I get this error only when I run tests using `uv run tox -e py` and not while running them via pytest i.e. `uv run pytest --hypothesis-seed=283474802224946286649981002652669788455 test/mitmproxy/proxy/layers/http/test_http_fuzz.py`. The error is seen only on windows however.

> I'm strongly against adding a new dependency that has "This project is not actively maintained and is looking for maintainers." in their GitHub repo.

Agreed, but I couldn't find an alternative.

mhils:
Hmm... in #7812, can we first update everything but hypothesis, and then see if indeed the hypothesis update is the culprit for this?

errorxyz:
> Hmm... in #7812, can we first update everything but hypothesis, and then see if indeed the hypothesis update is the culprit for this?

Locally all tests pass but for some reason pushing it to #7812 does not trigger the tests. I'll try digging deeper into this.

PR #7833, opened 2025-08-12
Console: Prevent showing the quit message when no flows are available in specific configurations

### LINKED ISSUE #7831: Filtering the flow list causes mitmproxy to prompt the user to quit the program
### Problem Description

When using vertical layout mode and displaying flow details, mitmproxy prompts the user to quit when filtering the list with a regex that produces no results. I believe this to be unintended behavior, and quite annoying.

1. Start mitmproxy with the `--set console_layout=vertical` flag
2. Populate the flow list with some requests
3. Configure the bottom pane to show flow details (Shift+Tab then P)
4. Change focus back to the flow list, then filter the list with a regex that produces no results
5. See how the user is met with a `Quit (yes,no)?` prompt

### System Information

```raw
Mitmproxy: 12.1.1 binary
Python:    3.13.3
Platform:  Linux-6.14.8-400.asahi.fc42.aarch64+16k-aarch64-with-glibc2.41
```

### Checklist

lups2000:
#### Description

This PR fixes: #7831 

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@
   ([#7749](https://github.com/mitmproxy/mitmproxy/pull/7749), @mhils)
 - Display local timezone in the Timing tab of mitmweb.
   ([#7804](https://github.com/mitmproxy/mitmproxy/pull/7804), @lups2000)
+- Prevent showing the quit message in the console when no flows are available under specific configurations.
+  ([#7833](https://github.com/mitmproxy/mitmproxy/pull/7833), @lups2000)
 
 ## 25 May 2025: mitmproxy 12.1.1
 
--- a/mitmproxy/tools/console/flowview.py
+++ b/mitmproxy/tools/console/flowview.py
@@ -104,7 +104,12 @@ class FlowDetails(tabs.Tabs):
                 ]
             self.show()
         else:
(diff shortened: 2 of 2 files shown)


Prinzhorn:
I know basically nothing about the console code. I have a hunch though that hardcoding `"flowlist"` is not the way. Can't we check the length to see if there is something to pop? Why does this require knowlege of which view is where? Isn't the logic "if we're not at the root, pop a window"?

lups2000:
> I know basically nothing about the console code. I have a hunch though that hardcoding `"flowlist"` is not the way. Can't we check the length to see if there is something to pop? Why does this require knowlege of which view is where? Isn't the logic "if we're not at the root, pop a window"?

I must admit that I also have limited knowledge on this. By inspecting the code, I noticed the logic was essentially **_“if we’re not at the root, pop a window"_**. However, if you applied a filter with no matches, or removed the last flow, the existing logic would still pop the window, which in this case triggered the **_“Do you want to quit?”_** prompt, even though you didn’t actually try to exit. I added a check for `"flowlist"` specifically because this window is a bit special and we shouldn't pop back just because there are no flows to show

mhils (approved):
Thanks! 🍰

PR #7837, opened 2025-08-19
Run pyupgrade and improve f-strings

injust:
#### Description

- Run pyupgrade for Python 3.12+
- Use `!r` flag instead of `repr()` in f-strings
- Remove redundant `str()` in f-strings
- Use self-documenting f-strings

#### Checklist


mhils (approved):
Thank you!

PR #7861, opened 2025-08-26
docs: Remove redundant "enabled by default" from option descriptions

injust:
#### Description

Some options have "enabled by default" in their description, which is redundant.

#### Checklist

### DIFF
--- a/mitmproxy/options.py
+++ b/mitmproxy/options.py
@@ -13,7 +13,7 @@ class Options(optmanager.OptManager):
     def __init__(self, **kwargs) -> None:
         super().__init__()
         self.add_option(
-            "server", bool, True, "Start a proxy server. Enabled by default."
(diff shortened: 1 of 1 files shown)


mhils:
Thank you! Can you clarify why you think those are redundant?

injust:
> Can you clarify why you think those are redundant?

All of these are booleans that are enabled by default.

On https://docs.mitmproxy.org/stable/concepts/options/, they read as "...enabled by default" and then "Default: True".

mhils:
Is it redundant in mitmproxy and mitmweb as well? These docstrings are displayed in multiple places.

injust:
> Is it redundant in mitmproxy and mitmweb as well? These docstrings are displayed in multiple places.

I see what you mean, e.g. the `--help` output doesn't show the options' default values.

PR #7864, opened 2025-08-30
Bump pytest dependency group

errorxyz:
#### Description

Bumps the various pytest-* dependencies.
Supersedes https://github.com/mitmproxy/mitmproxy/pull/7812

#### Checklist

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -890,11 +890,11 @@ dev = [
     { name = "pdoc", specifier = "==15.0.4" },
     { name = "pyinstaller", specifier = "==6.15.0" },
     { name = "pyinstaller-hooks-contrib", specifier = "==2025.8" },
-    { name = "pytest", specifier = "==8.3.5" },
-    { name = "pytest-asyncio", specifier = "==0.26.0" },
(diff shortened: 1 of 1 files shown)


test/mitmproxy/tools/test_main.py
```diff
@@ -1,12 +1,9 @@
-import asyncio
-
 from mitmproxy.tools import main
 
 shutdown_script = "mitmproxy/data/addonscripts/shutdown.py"
 
 
-def test_mitmweb(event_loop, tdata):
-    asyncio.set_event_loop(event_loop)
+def test_mitmweb(tdata):
```
errorxyz, line 6: I'm not 100% sure if its the right fix

mheguy:
@errorxyz 6cc3bf8 failed CI because you updated pyproject.toml but not uv.lock. uv threw an error because the two files were not in sync.
2bdd68c updated uv.lock automatically (nice github action)

test/mitmproxy/tools/test_main.py
mhils: This should be good enough, thanks!

PR #7881, opened 2025-09-14
docs: Add note about unused standard libraries.

adityaSiso:
#### Description

Added a note to the mitmproxy PyPI installation instructions regarding the skipping of unused standard library modules.

#### Checklist


adityaSiso:
Updated, Thanks for the information. 

PS: This is my first open source pull request, so I’m a bit nervous about the changes, appreciate any feedback!

injust:
I'd instead add a note after line 60. Something like:

> You might need to do this for standard library modules, as the standalone binaries only include a minimal Python environment.

The `uvx --with` option is already documented, so no need to repeat that.

adityaSiso:
> I'd instead add a note after line 60. Something like:
> 
> > You might need to do this for standard library modules, as the standalone binaries only include a minimal Python environment.
> 
> The `uvx --with` option is already documented, so no need to repeat that.

Thoughts on updating line 60 to something like:

> If your mitmproxy addons require additional Python packages, consider installing mitmproxy from PyPI. The standalone binaries include only a minimal Python runtime, which may require you to install dependencies manually.

PR #7896, opened 2025-10-02
address bcrypt error: migrate to libpass

crclark96:
#### Description

bcrypt 5.0 added a breaking change that causes passlib to throw and exception upon initialization; this is addressed in the most recent release of passlib from the libpass fork. Unfortunately, the original project has been out of maintenance for >5 years

#### Checklist

#### Related

[Passlib release notes describing bcrypt issue](https://github.com/notypecheck/passlib/releases/tag/1.9.2)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 ## Unreleased: mitmproxy next
 
 - Passlib version and fork updated to address bcrypt error.
-
+  ([#7896](https://github.com/mitmproxy/mitmproxy/pull/7896), @crclark96)
 
 ## 24 August 2025: mitmproxy 12.1.2
 


mhils:
Thank you for the PR! I don't feel comfortable switching to libpass. passlib is a 1) abandoned, 2) widely used, and 3) security critical. This makes it a prime target for a "Jia Tan"-style supply chain attack. 
To be clear, I have no other reasons to suspect something about libpass in particular, and the fork author very likely only has the best intentions. But I don't want to review their releases going forward. I think the best course of action for mitmproxy is to remove the htpasswd functionality, and/or replace it with a small custom htpasswd implementation that only supports a Python standard library hash format.

mhils:
Closing this in favor of https://github.com/mitmproxy/mitmproxy/pull/7896, thanks again!

crclark96:
Oh awesome, quick work!

I was planning to spend some more time looking at this but it seems like you're already done with a custom implementation. Thanks for addressing the underlying problem.

PR #7910, opened 2025-10-12
feat: Make TCP inactivity timeout configurable 

keshavkrishna:
#### Description
Make TCP inactivity timeout configurable 

Previously, mitmproxy used a hardcoded TCP inactivity timeout of 10 minutes (600 seconds) for all TCP connections. This change makes the timeout configurable through a new `tcp_timeout` option, allowing users to adjust the timeout based on their specific use cases.

**Changes:**
- Added `tcp_timeout` option to `mitmproxy/options.py` with default value of 600 seconds
- Updated `mitmproxy/proxy/server.py` to use the configurable timeout instead of hardcoded `TCP_TIMEOUT`
- Added command-line support for the new option in `mitmproxy/tools/cmdline.py`
- Added comprehensive tests for the new option in `test/mitmproxy/test_options.py`
- Updated web interface TypeScript definitions to include the new option

**Usage:**
```bash
# Set TCP timeout to 5 minutes
mitmproxy --set tcp_timeout=300

# Set TCP timeout to 30 minutes  
mitmproxy --set tcp_timeout=1800
```

This change maintains backward compatibility by keeping the same default timeout value while providing the flexibility to adjust it for different network environments and use cases.

#### Checklist

 - [x] I have added an entry to the CHANGELOG.md


web/gen/options_js.py
```diff
@@ -17,17 +17,17 @@
 
 
 def _ts_type(t):
-    if t is bool:
+    if t == bool:  # noqa: E721
```
mhils: What's up with this change here?

test/mitmproxy/test_options.py
```diff
@@ -3,3 +3,13 @@
...
+
+
+def test_tcp_timeout_default():
+    opts = options.Options()
+    assert opts.tcp_timeout == 600
+
+
+def test_tcp_timeout_settable():
+    opts = options.Options(tcp_timeout=123)
+    assert opts.tcp_timeout == 123
```
mhils: Remove these tests - they don't add value.

mhils (commented):
Thank you for the PR and sorry for taking a while to get to this! Can we please add a test that sets the timeout to 0.1s, then opens a connection, and verifies that we indeed get it closed right away?

keshavkrishna:
Hi @mhils,
  I've added the functional test, but tcp_timeout is currently typed as int (like http2_ping_keepalive), so it doesn't accept 0.1. hence, I used tcp_timeout = 1 (1 second) in the test instead. 
  Should I keep it as int with 1 second, or change the option type to float to support 0.1?

test/mitmproxy/test_options.py
keshavkrishna: removed these

web/gen/options_js.py
keshavkrishna: this was pushed by mistake, reverted it now.

test/mitmproxy/proxy/test_mode_servers.py
```diff
@@ -98,6 +98,42 @@ async def test_tcp_start_stop(caplog_async):
...
 
 
+async def test_tcp_timeout(caplog_async):
+    """Test that TCP connections are closed after the configured timeout period."""
+    caplog_async.set_level("INFO")
+    manager = MagicMock()
+
+    with taddons.context() as tctx:
+        # Set a very short timeout for testing (1 second)
+        tctx.options.tcp_timeout = 1
```
mhils: What happens if we set it to 0?

(I'm mostly concerned with test suite speed - if 0 works let's do that)

test/mitmproxy/proxy/test_mode_servers.py
```diff
@@ -98,6 +98,42 @@ async def test_tcp_start_stop(caplog_async):
...
+        await inst.start()
+        assert await caplog_async.await_log("proxy listening")
+
+        host, port, *_ = inst.listen_addrs[0]
+        reader, writer = await asyncio.open_connection(host, port)
+        assert await caplog_async.await_log("client connect")
+
+        # Keep connection open but inactive - it should timeout after 1s
+        # Wait a bit longer than the timeout to ensure it triggers
+        await asyncio.sleep(1.2)
```
mhils: Remove - the `await_log` statement below waits for the log message to appear anyhow.
```suggestion
```

test/mitmproxy/proxy/test_mode_servers.py
```diff
@@ -98,6 +98,42 @@ async def test_tcp_start_stop(caplog_async):
...
+
+        # Verify the connection was closed due to inactivity
+        assert await caplog_async.await_log("Closing connection due to inactivity")
+        assert await caplog_async.await_log("client disconnect")
+
+        # Try to read from the closed connection to confirm it's really closed
+        data = await reader.read(1)
+        assert data == b""  # EOF indicates connection is closed
```
mhils: Let's add a `async with asyncio.timeout(30):` for this block to ensure we've actually reduced the timeout?

mhils (commented):
Thank you! Let's keep it as an int, that makes things simpler.

test/mitmproxy/proxy/test_mode_servers.py
keshavkrishna: I tested tcp_timeout=0 and it works perfectly.

  With timeout=0, the TimeoutWatchdog does asyncio.sleep(0 - elapsed_time) which yields immediately, and the timeout condition last_activity + 0 < time.time() triggers right away. Connection closes on the next event loop iteration.

  Performance:
  - tcp_timeout=0: 0.11s 
  - tcp_timeout=1: 1.16s

PR #7923, opened 2025-10-19
Console: Show query parameters for empty-body requests

### LINKED ISSUE #7914: Query request format broken
### Problem Description

When viewing the request made with query parameters (encoded in URL) the request body is not shown neither in auto nor in manually select query mode.

1. Install mitmproxy version 12.1.2 
2.  Make GET request with query params
3.  Open the flow and notice that request body is empty

### System Information

```raw
Mitmproxy: 12.1.2
Python:    3.13.5
OpenSSL:   OpenSSL 3.5.2 5 Aug 2025
Platform:  Linux-6.12.48+deb13-amd64-x86_64-with-glibc2.41
```

### Checklist

lups2000:
#### Description

This PR fixes: #7914
We now explicitly check for query parameters and display them even when the request body is empty.

#### Checklist

### DIFF
--- a/mitmproxy/tools/console/flowview.py
+++ b/mitmproxy/tools/console/flowview.py
@@ -316,9 +316,8 @@ class FlowDetails(tabs.Tabs):
 
         if message.raw_content == b"":
             if isinstance(message, http.Request):
-                path = getattr(message, "path", "") or ""
-                parsed = urlparse(path)
-                if not parsed.query:
+                query = getattr(message, "query", "") or ""
+                if not query:
                     # No body and no query params
                     return "", [urwid.Text("No request content")]
                 # else: there are query params -> fall through to render them
--- a/mitmproxy/tools/console/flowview.py
+++ b/mitmproxy/tools/console/flowview.py
@@ -1,6 +1,5 @@
 import sys
 from functools import lru_cache
-from urllib.parse import urlparse
 
 import urwid
 
--- a/mitmproxy/tools/console/flowview.py
+++ b/mitmproxy/tools/console/flowview.py
@@ -1,5 +1,6 @@
 import sys
 from functools import lru_cache
+from urllib.parse import urlparse
 
 import urwid
 
@@ -310,34 +311,42 @@ class FlowDetails(tabs.Tabs):
     def content_view(
         self, viewmode: str, message: http.Message
     ) -> tuple[str, list[urwid.Text]]:
+
         if message.raw_content is None:
             return "", [urwid.Text([("error", "[content missing]")])]
-        elif message.raw_content == b"":
+
+        if message.raw_content == b"":
             if isinstance(message, http.Request):
-                return "", [urwid.Text("No request content")]
+                path = getattr(message, "path", "") or ""
+                parsed = urlparse(path)
+                if not parsed.query:
+                    # No body and no query params
+                    return "", [urwid.Text("No request content")]
+                # else: there are query params -> fall through to render them
             else:
                 return "", [urwid.Text("No content")]
-        else:
-            full = self.master.commands.execute(
-                "view.settings.getval @focus fullcontents false"
-            )
-            if full == "true":
-                limit = sys.maxsize
-            else:
-                limit = ctx.options.content_view_lines_cutoff
 
-            flow_modify_cache_invalidation = hash(
-                (
-                    message.raw_content,
-                    message.headers.fields,
-                    getattr(message, "path", None),
-                )
+        full = self.master.commands.execute(
+            "view.settings.getval @focus fullcontents false"
+        )
+        
+        if full == "true":
+            limit = sys.maxsize
+        else:
+            limit = ctx.options.content_view_lines_cutoff
+
+        flow_modify_cache_invalidation = hash(
+            (
+                message.raw_content,
+                message.headers.fields,
+                getattr(message, "path", None),
             )
-            # we need to pass the message off-band because it's not hashable
-            self._get_content_view_message = message
-            return self._get_content_view(
-                viewmode, limit, flow_modify_cache_invalidation
-            )
+        )
+        # we need to pass the message off-band because it's not hashable
+        self._get_content_view_message = message
+        return self._get_content_view(
+            viewmode, limit, flow_modify_cache_invalidation
+        )
 
     @lru_cache(maxsize=200)
     def _get_content_view(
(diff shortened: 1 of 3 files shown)


mitmproxy/tools/console/flowview.py
```diff
@@ -312,32 +313,37 @@ def content_view(
     ) -> tuple[str, list[urwid.Text]]:
         if message.raw_content is None:
             return "", [urwid.Text([("error", "[content missing]")])]
-        elif message.raw_content == b"":
+
+        if message.raw_content == b"":
             if isinstance(message, http.Request):
-                return "", [urwid.Text("No request content")]
+                path = getattr(message, "path", "") or ""
```
sujaldev: Maybe we can directly use `getattr(message, "query")` here?

lups2000: Hi, yes, you are absolutely right! I didn't notice that the class offers the query property :) thanks!

mitmproxy/tools/console/flowview.py
```diff
@@ -312,32 +312,36 @@ def content_view(
     ) -> tuple[str, list[urwid.Text]]:
         if message.raw_content is None:
             return "", [urwid.Text([("error", "[content missing]")])]
-        elif message.raw_content == b"":
+
+        if message.raw_content == b"":
             if isinstance(message, http.Request):
-                return "", [urwid.Text("No request content")]
+                query = getattr(message, "query", "") or ""
```
sujaldev:
```suggestion
                query = getattr(message, "query", "")
```

Since the next line already checks for a falsy value, this seems redundant.

sujaldev (commented):
Nevermind, I think this is fine.

sujaldev:
There might be a separate issue here with the `_get_content_view` function, it does not seem to check for the case where the query might exist but the message content is missing. Which can be reproduced by running the following bash script:

```bash
(
  printf 'GET /?a=b&c=d HTTP/1.1\r\n'
  printf 'Host: www.example.com\r\n'
  printf 'Content-Length: 1\r\n\r\n'
  sleep 100 # mitmproxy keeps waiting for the expected 1 byte of data for this time period.
) | nc 127.0.0.1 8080 # mitmproxy running on 8080

```

In this scenario even manually switching to query view will not allow rendering the query.

In anticipation for that issue, perhaps we should switch the sequence of checks to be something like this?

```python
query = getattr(message, "query", "")
if not query:
    if message.raw_content is None:
        return "", [urwid.Text([("error", "[content missing]")])]
    elif message.raw_content == b"":
        return "", [urwid.Text("No content")]
```

lups2000:
> In anticipation for that issue, perhaps we should switch the sequence of checks to be something like this?
> 
> ```python
> query = getattr(message, "query", "")
> if not query:
>     if message.raw_content is None:
>         return "", [urwid.Text([("error", "[content missing]")])]
>     elif message.raw_content == b"":
>         return "", [urwid.Text("No content")]
> ```

I tried switching the sequence to check for a query first, but the problem persists. If a query exists we continue with the normal flow and call `self._get_content_view(viewmode, limit, flow_modify_cache_invalidation)`. That calls `contentviews.prettify_message(message, self.flow, viewmode)`, which calls `data, enc = get_data(message)`. `data` is then `None` because the content we were waiting for is missing. I'm not sure where to handle this specific case. In the end, it might be acceptable to display a simple "Content is missing" message; what do you think?

sujaldev:
> I tried switching the sequence to check for a query first, but the problem persists.

I guess fixing that is outside the scope of this PR, so this should be ready to merge I think. I just mentioned that edge case so we wouldn't have to redo this later on when fixing that issue.

lups2000:
> > I tried switching the sequence to check for a query first, but the problem persists.
> 
> I guess fixing that is outside the scope of this PR, so this should be ready to merge I think. I just mentioned that edge case so we wouldn't have to redo this later on when fixing that issue.

sounds good to me! Thanks for pointing this out. Maybe you can already open an issue for that :)

lups2000:
0.02% missing coverage was annoying 🤯 and apparently it still fails 😓

test/mitmproxy/tools/console/test_flowview.py
```diff
@@ -31,3 +36,73 @@ async def test_edit(console, monkeypatch, caplog):
...
+        captured["limit"] = limit
+        return ("TEST_VIEW", [urwid.Text("ok")])
+
+    fd._get_content_view = fake_get_content_view
+
+    title, text_objs = fd.content_view("default", f.request)
+
+    assert title == "TEST_VIEW"
+    assert captured.get("limit") == sys.maxsize
+    assert isinstance(text_objs[0], urwid.Text)
```
sujaldev: We could make this shorter with `unittest.mock.MagicMock` here: 

```suggestion
    console.commands.execute("view.settings.setval @focus fullcontents true")

    fd._get_content_view = mock.MagicMock()

    fd.content_view("default", f.request)
    fd._get_content_view.assert_called_with("default", sys.maxsize, mock.ANY)
```

test/mitmproxy/tools/console/test_flowview.py
```diff
@@ -1,5 +1,10 @@
+import sys
+
```
sujaldev:
```suggestion
from unittest import mock

```

sujaldev (commented):
Looks great :rocket:! I've pointed out some nits (but feel free to ignore, only stylistic changes).

test/mitmproxy/tools/console/test_flowview.py
```diff
@@ -31,3 +36,73 @@ async def test_edit(console, monkeypatch, caplog):
...
+    await console.load_flow(f_missing)
+
+    fd = FlowDetails(console)
+
+    title, txt_objs = fd.content_view("default", f_missing.request)
+    assert title == ""
+
+    first_text = txt_objs[0].get_text()[0]
+    assert "[content missing]" in first_text
```
sujaldev: Maybe we could use `==` instead of `in` here (and other places)? `in` could hint towards it being a substring.

test/mitmproxy/tools/console/test_flowview.py
lups2000: yeah thanks for the tip :)

sujaldev:
Thanks for fixing this! Good to go from my end (just in case you were waiting on me).

lups2000:
> Thanks for fixing this! Good to go from my end (just in case you were waiting on me).

Thank you for your feedback :) let's wait for @mhils 's green light 🟢

mhils (approved):
Thanks! 🍰

PR #7928, opened 2025-10-22
Fix various issues in infer_content_encoding

xu-cheng:
#### Description

- Per spec, the byte order mark (BOM) always has the highest priority to determine various UTF related encodings. Ref:
    - https://www.w3.org/International/questions/qa-html-encoding-declarations#bom
    - https://html.spec.whatwg.org/multipage/semantics.html#charset
    - https://drafts.csswg.org/css-syntax/#css-decode-bytes
- For xml traffics such as `application/xml`, `image/svg+xml` etc, a XML declaration may contain the encoding information. Ref:
    - https://www.w3.org/TR/xml/#charencoding
    - https://html.spec.whatwg.org/multipage/semantics.html#charset
- For CSS, the default fallback encoding should be UTF-8. Ref:
    - https://drafts.csswg.org/css-syntax/#determine-the-fallback-encoding

#### Checklist

### DIFF
--- a/mitmproxy/net/http/headers.py
+++ b/mitmproxy/net/http/headers.py
@@ -70,7 +70,9 @@ def infer_content_encoding(content_type: str, content: bytes = b"") -> str:
             enc = meta_charset.group(1).decode("ascii", "ignore")
 
     if not enc and "xml" in content_type:
-        if xml_encoding := re.search(rb"""<\?xml[^\?>]+encoding=['"]([^'"\?>]+)""", content, re.IGNORECASE):
+        if xml_encoding := re.search(
+            rb"""<\?xml[^\?>]+encoding=['"]([^'"\?>]+)""", content, re.IGNORECASE
+        ):
             enc = xml_encoding.group(1).decode("ascii", "ignore")
 
     if not enc and "text/css" in content_type:


mitmproxy/net/http/headers.py
```diff
@@ -40,9 +40,25 @@ def infer_content_encoding(content_type: str, content: bytes = b"") -> str:
...
+        enc = "utf-8-sig"
+    else:
+        enc = None
+
     # Use the charset from the header if possible
-    parsed_content_type = parse_content_type(content_type)
-    enc = parsed_content_type[2].get("charset") if parsed_content_type else None
+    if not enc:
+        parsed_content_type = parse_content_type(content_type)
+        enc = parsed_content_type[2].get("charset") if parsed_content_type else None
```
mhils: Combine to this? 
```python
elif parsed_content_type := parse_content_type(content_type):
    enc = parsed_content_type[2].get("charset")
else:
    enc = None
```
 ?

mitmproxy/net/http/headers.py
```diff
@@ -55,12 +71,24 @@ def infer_content_encoding(content_type: str, content: bytes = b"") -> str:
...
+
     if not enc and "text/css" in content_type:
         # @charset rule must be the very first thing.
         css_charset = re.match(rb"""@charset "([^"]+)";""", content, re.IGNORECASE)
         if css_charset:
             enc = css_charset.group(1).decode("ascii", "ignore")
 
+        # Fallback to utf8 for css
+        # Ref: https://drafts.csswg.org/css-syntax/#determine-the-fallback-encoding
+        if not enc:
```
mhils:
```suggestion
        else:
            # Fallback to utf8 for css
            # Ref: https://drafts.csswg.org/css-syntax/#determine-the-fallback-encoding
```

Issue #7807, opened 2025-07-21
Could not extract BSD name from mount point


Urie96:
### Problem Description

When editing the request, nothing happened. After exiting mitmproxy, the error logs were printed. 

```
$ mitmproxy
Can't statfs , failed with error -1
Could not extract BSD name from mount point 
$ 
```

However, editing the response worked normally.

### System Information

```raw
Mitmproxy: 12.1.1
Python:    3.13.5
OpenSSL:   OpenSSL 3.5.1 1 Jul 2025
Platform:  macOS-14.7-arm64-arm-64bit-Mach-O
```

### Checklist

- [x] This bug affects the [latest mitmproxy release](https://github.com/mitmproxy/mitmproxy/releases).

mhils:
Thank you for your report. No idea what's going on here. Can you run mitmproxy with strace? Or maybe `RUST_LOG=TRACE mitmdump -vvv`?

Urie96:
> Can you run mitmproxy with strace? Or maybe `RUST_LOG=TRACE mitmdump -vvv`?

I tried `RUST_LOG=TRACE mitmproxy -vvv`, but the output was identical to what I’d already seen—no extra details appeared.

I’ve now found that the crash only happens when the request body contains Chinese characters.  
For example, with this curl command mitmproxy refuses to let me edit the request (the response also contains Chinese, yet the response can be edited without issue):
```sh
curl --proxy http://localhost:8080 -H "Content-Type: application/json" -d '{"key":"你好"}' https://httpbin.org/post
```
With this curl command everything works fine:
```sh
curl --proxy http://localhost:8080 -H "Content-Type: application/json" -d '{"key":"hello"}' https://httpbin.org/post
```

```console
$ /nix/store/nwval4ssnqfxmsivfkpbyzp6glr72vwz-python3.13-mitmproxy-12.1.1/bin/mitmproxy --version
Mitmproxy: 12.1.1
Python:    3.13.5
OpenSSL:   OpenSSL 3.5.1 1 Jul 2025
Platform:  macOS-14.7-arm64-arm-64bit-Mach-O
```

---

When I install mitmproxy via Homebrew the behavior is different: I cannot edit requests or responses even when they contain no Chinese characters. In this case the Events tab shows the following traceback:
```
info: [13:08:10.625] HTTP(S) proxy listening at *:8080.
info: [13:08:15.490][[::1]:59401] client connect
[...]

PR #7933, opened 2025-10-25
fix: is_mostly_bin support chinese characters

caiquejjx:
#### Description
Fixes: https://github.com/mitmproxy/mitmproxy/issues/7807
the `strutils.is_mostly_bin` function was mistakenly classifying chinese charactes as binary data causing a hex editor to be used instead of a text editor

#### Checklist

### DIFF
--- a/test/mitmproxy/utils/test_strutils.py
+++ b/test/mitmproxy/utils/test_strutils.py
@@ -85,14 +85,15 @@ def test_is_mostly_bin():
     assert not strutils.is_mostly_bin(b"")
     assert strutils.is_mostly_bin(b"\x00\x01\x02\x03\x04\x05\x06\x07\x08\x09")
     # shift UTF8 break point
-    assert not strutils.is_mostly_bin(
-        b"" + 50 * "𐍅".encode()
-    )  # four byte UTF8 character
+    # 𐍅 is four bytes in UTF-8, so we're breaking the 100 chars barrier.
+    assert not strutils.is_mostly_bin(b"" + 50 * "𐍅".encode())
     assert not strutils.is_mostly_bin(b"a" + 50 * "𐍅".encode())
     assert not strutils.is_mostly_bin(b"aa" + 50 * "𐍅".encode())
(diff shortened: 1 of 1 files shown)


mitmproxy/utils/strutils.py
```diff
@@ -126,10 +126,33 @@ def escaped_str_to_bytes(data: str) -> bytes:
...
+    Strategy:
+    1. Try decode as UTF-8:
+       - If decode fails, treat with byte-level heuristic (control bytes and >126).
+       - If decode succeeds, count control chars (ord < 32 or ==127) except \t,\n,\r.
+         If control-char ratio > threshold -> binary, else text.
+    """
     if not s or len(s) == 0:
         return False
+    try:
+        text = s[:100].decode("utf-8")
```
mhils: How does this interact with incomplete UTF-8 code points at around index 100? Does longer text fail to decode?

caiquejjx: no it doesn't fails, i was using the same sampling logic it had before where it would calculate up to 100 bytes:
`sum(i < 9 or 13 < i < 32 or 126 < i for i in s[:100]) / len(s[:100])`
but for sure we can decode the whole content

mhils: We can't decode the whole content for performance reasons.

My point is that this heuristic will fail for Chinese text longer than 100 characters. Here is a minimal repro payload:

```python3
(50*"你好").encode()[:100]
```

This payload will fail the `.decode("utf-8")` logic ('utf-8' codec can't decode byte 0xe5 in position 99: unexpected end of data) and fall back to the old logic. In other words, the new "is this utf8?" logic is not consistently applied for longer texts.

caiquejjx: oh I get it now, i think it could perhaps ignore errors on decoding so it decodes as much as it can, if nothing it's binary else apply the control chars threshold logic

caiquejjx: I'll look at it again soon, thanks for the review

caiquejjx:
@mhils yeah looks good to me, the heuristics look more concise, nice job

PR #7937, opened 2025-10-27
chore: remove bless from hex editors

caiquejjx:
#### Description
Complements: https://github.com/mitmproxy/mitmproxy/issues/7807
- Removes `bless` from the hex editors list to avoid issues with macOS own `bless` 
#### Checklist


caiquejjx:
@mhils I checked the failures but idk how it relates to this change

Issue #7795, opened 2025-07-09
Cannot decode a gzip response body


Karmylr:
### Problem Description

1、
```
if flow.request.pretty_url == "https://vd6.l.qq.com/proxyhttp":
    print(flow.response.get_text())
```
2、Open `https://v.qq.com/x/cover/mzc00200k9i34np/z0040nstq8h.html`

Logs:
```
Addon error: EOFError when decoding b'\x1f\x8b with 'gzip': EOFError('Compressed file ended before the end-of-stream marker was reached')
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/mitmproxy/net/encoding.py", line 62, in decode
    decoded = custom_decode[encoding](encoded)
  File "/usr/local/lib/python3.13/site-packages/mitmproxy/net/encoding.py", line 151, in decode_gzip
    return f.read()
           ~~~~~~^^
  File "/usr/local/lib/python3.13/gzip.py", line 340, in read
    return self._buffer.read(size)
           ~~~~~~~~~~~~~~~~~^^^^^^
  File "/usr/local/lib/python3.13/_compression.py", line 118, in readall
    while data := self.read(sys.maxsize):
                  ~~~~~~~~~^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/gzip.py", line 566, in read
    raise EOFError("Compressed file ended before the "
                   "end-of-stream marker was reached")
EOFError: Compressed file ended before the end-of-stream marker was reached

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/etc/mitmproxy/scrpit/adblock.py", line 786, in response
    print(flow.response.get_text())
          ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/site-packages/mitmproxy/http.py", line 432, in get_text
    content = self.get_content(strict)
  File "/usr/local/lib/python3.13/site-packages/mitmproxy/http.py", line 395, in get_content
    content = encoding.decode(self.raw_content, ce)
  File "/usr/local/lib/python3.13/site-packages/mitmproxy/net/encoding.py", line 71, in decode
    raise ValueError(
    ...<6 lines>...
    )
ValueError: EOFError when decoding b'\x1f\x8b with 'gzip': EOFError('Compressed file ended before the end-of-stream marker was reached')
```

### System Information

```raw
Mitmproxy: 12.1.1
Python:    3.13.4
[...]

PR #7940, opened 2025-10-30
fixes gzip decoding issue.

AdityaPatadiya:
#### Description

This PR fixes issue [#7795](https://github.com/mitmproxy/mitmproxy/issues/7795)
 where mitmproxy failed to decode certain raw binary payloads, resulting in errors such as:
```
Raw [cannot decode][failed to parse as JSON]
```

#### Checklist

### DIFF
--- a/test/mitmproxy/net/test_encoding.py
+++ b/test/mitmproxy/net/test_encoding.py
@@ -69,7 +69,9 @@ def test_encoders_strings(encoder):
 class TestDecodeGzip:
     def test_regular_gzip(self):
         # generated with gzip.compress(b"mitmproxy")
-        data = bytes.fromhex("1f8b0800e4a4106902ffcbcd2cc92d28caafa80400d21f9c9d09000000")
+        data = bytes.fromhex(
+            "1f8b0800e4a4106902ffcbcd2cc92d28caafa80400d21f9c9d09000000"
+        )
         assert encoding.decode_gzip(data) == b"mitmproxy"
 
     def test_zlib(self):
@@ -84,8 +86,8 @@ class TestDecodeGzip:
             "b2527ab17efbb38d4d4f7b5a9fec58fb6cd3c267733a934a3353946a01000000ffff"
         )
         assert encoding.decode_gzip(data) == (
-            b"{\"errCode\":-5, \"retMsg\":\"\xe8\xaf\xb7\xe6\xb1\x82\xe5\x8c\x85\xe4"
-            b"\xb8\xad\xe6\xb2\xa1\xe6\x9c\x89buid\"}"
+            b'{"errCode":-5, "retMsg":"\xe8\xaf\xb7\xe6\xb1\x82\xe5\x8c\x85\xe4'
+            b'\xb8\xad\xe6\xb2\xa1\xe6\x9c\x89buid"}'
         )
 
 


AdityaPatadiya:
Hello @mhils and @Prinzhorn , can you review this PR?

mitmproxy/net/encoding.py
```diff
@@ -147,8 +147,16 @@ def identity(content):
...
         return b""
-    with gzip.GzipFile(fileobj=BytesIO(content)) as f:
-        return f.read()
+
+    try:
+        with gzip.GzipFile(fileobj=BytesIO(content)) as f:
+            return f.read()
+    except (OSError, EOFError):
+        try:
+            decompressor = zlib.decompressobj(16 + zlib.MAX_WBITS)
```
mhils: Can we add a comment that explains the choice of `16 + zlib.MAX_WBITS` please?

Looking at https://docs.python.org/3/library/zlib.html#zlib.decompress, I'm wondering if +47 would be better and if that'd allow us to skip the `gzip.GzipFile` part. Could you investigate?

AdityaPatadiya: Thanks @mhils to review this PR.
I've investigate and conclude that +47 can cause unpredictable data masking issues due to overlapping header interpretation like if the data is malformed, mislabeled, or partially corrupted, it may:

- Silently succeed, giving you wrong or partial output, instead of raising a zlib.error (which would clearly indicate a problem).

and we could not use the zlib alone but I handled it more precisely that would always decode using zlib but if it couldn't then the gzip.GzipFile will handle it.

mhils: Can you cite a source for this please?

AdityaPatadiya: Thank you for pointing that out.

Upon re-reviewing the zlib documentation and tests across recent versions, I couldn't find an official or verified source describing "data masking" issue with `+47`. My earlier explaination about the overlapping header interpretation was based on my understanding of how `wbits` auto-detection works internally, but ir wasn't supported by any specific documentation or bug report.

To clarify, `+47` simply enables auth-detection of both zlib and gzip headers. There's no official record of it silently masking malformed data in current zlib versions.

mhils: Is this a human answer, or is this an AI-generated answer?

AdityaPatadiya: No, it is actually written by me not copy-pasted or generated by AI.

AdityaPatadiya: I apologize for the misunderstanding about the +47 issue.

I'll update the code with relevant changes as you said and that would also rely solely on zlib.

mitmproxy/net/encoding.py
```diff
@@ -145,10 +145,23 @@ def identity(content):
...
+    Reference:
+    https://docs.python.org/3/library/zlib.html#zlib.decompress
+    """
     if not content:
         return b""
-    with gzip.GzipFile(fileobj=BytesIO(content)) as f:
-        return f.read()
+
+    try:
+        decompressor = zlib.decompressobj(32 + zlib.MAX_WBITS)
```
mhils: Why decompressobj and not decompress?

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@
   Contributions for additional formats are welcome as long as they don't introduce new
   dependencies.
   ([#7906](https://github.com/mitmproxy/mitmproxy/pull/7906), @mhils)
+- Gracefully handle decoding of raw binary payloads that previously caused 
+  "Raw cannot decode" or "failed to parse as JSON" errors.
```
mhils: Wrong section

mitmproxy/net/encoding.py
mhils: Please use 47 as a constant, there's no added value in `32 + zlib.MAX_WBITS`

mitmproxy/net/encoding.py
```diff
@@ -145,10 +145,23 @@ def identity(content):
...
 
 def decode_gzip(content: bytes) -> bytes:
+    """
+    Decode gzip or zlib-compressed data using zlib's auto-detection.
+
+    Using wbits=47 (32 + 15) tells zlib to automatically detect both gzip and headers.
+    This simplifies decoding and avoids the need for a separate gzip.GzipFile fallback.
+
+    Reference:
+    https://docs.python.org/3/library/zlib.html#zlib.decompress
```
mhils: This belongs into a local comment, not the docstring.

test/mitmproxy/net/test_encoding.py
```diff
@@ -66,6 +67,21 @@ def test_encoders_strings(encoder):
         encoding.decode("foobar", encoder)
 
 
+def test_decode_gzip_truncated():
+    """
+    Test that decode_gzip() gracefully handles trucncated gzip streams.
```
mhils: typo

test/mitmproxy/net/test_encoding.py
```diff
@@ -66,6 +67,21 @@ def test_encoders_strings(encoder):
...
+    This ensures the fallback using zlib.decompressobj works correctly.
+    """
+    data = b"Hello, mitmproxy!"
+    compressed = gzip.compress(data)
+
+    truncated = compressed[:-5]
```
mhils: Let's not make up test vectors dynamically, but use the one from https://github.com/mitmproxy/mitmproxy/issues/7795#issuecomment-3061960259. This makes the test more robust.

test/mitmproxy/net/test_encoding.py
```diff
@@ -66,6 +67,21 @@ def test_encoders_strings(encoder):
         encoding.decode("foobar", encoder)
 
 
+def test_decode_gzip_truncated():
+    """
+    Test that decode_gzip() gracefully handles trucncated gzip streams.
+    This ensures the fallback using zlib.decompressobj works correctly.
```
mhils: Add a ref to #7795 here.

mhils (changes requested):

AdityaPatadiya:
Now the changes are made according to your requested changes.

AdityaPatadiya:
Hello @mhils , could you please review this PR
I made the changes.

AdityaPatadiya:
I think now it's ready to merge?

mhils:
Thank you. There's no need to be pushy on PR reviews, you're achieving the opposite of what you intend to achieve.

Issue #7859, opened 2025-08-24
Dependabot lockfile updates are broken


mhils:
#### Steps to reproduce the behavior:
1. Have a Dependabot PR that widens a version range (good)
2. The lockfile update in the PR will keep the old version (bad)

I can't reproduce this locally - we need to figure out why CI behaves differently. Maybe it's uv trying to be smart?

errorxyz:
I'm not too familiar with the uv environment, but does this help: https://github.com/mitmproxy/mitmproxy/actions/runs/17347910862/job/49250283535 ?

mheguy:
To explain this issue, see #7745.
Note how only 1 line in uv.lock is changed. The version specifier is widened, but the version that is actually used / specified in the lockfile is not updated.


I submitted #7873. It should help, but dependabot is a bit of a black box at times.
When testing locally, using https://github.com/dependabot/cli and specifying uv, the update failed because uv was unable to build mitmproxy. 🤷

mheguy:
@errorxyz I'll address your issue in your PR.

mheguy:
> I'm not too familiar with the uv environment, but does this help: https://github.com/mitmproxy/mitmproxy/actions/runs/17347910862/job/49250283535 ?

I'll take a look at your issue in that PR, but it's unrelated to dependabot.

PR #7946, opened 2025-11-02
Try to fix dependabot

mhils:
(Hopefully) fixes https://github.com/mitmproxy/mitmproxy/issues/7859


arjun810:
@mhils any chance you could put out a release with this? pip install mitmproxy fails on some systems without this change. installing from github now, but would appreciate a release!

PR #7961, opened 2025-11-04
infer_content_encoding: Fallback to UTF-8 for more content types

xu-cheng:
#### Description

For html, css, js, and xml content types, the official specs tell us to use UTF-8 as fallback when charset are not specified.

Relevant sections of the specs are included in the comments for the corresponding branches.

This commit also uncovered incorrect decoding in the following har tests:

test/mitmproxy/data/har_files/firefox.json
test/mitmproxy/data/har_files/safari.json

```diff
- * Licensed MIT © Zeno Rocha
+ * Licensed MIT  Zeno Rocha
```

Noted that the UTF-8 char `©` was missing due to the incorrect encoding.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 ## Unreleased: mitmproxy next
 
+- infer_content_encoding: Fallback to UTF-8 for more content types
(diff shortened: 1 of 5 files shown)


mhils (approved):
Thank you! 🍰

PR #7963, opened 2025-11-07
fix mitmweb --web-host with ipv6

Julien00859:
When using `--web-host ::1` the `web_url` was `http://::1:8081` instead of `http://[::1]:8081`. This lead for the wrong URL to be written in the logs, and the wrong URL to be opened with `--web-open-browser`.

#### Checklist

### DIFF
--- a/test/mitmproxy/tools/web/test_webaddons.py
+++ b/test/mitmproxy/tools/web/test_webaddons.py
@@ -49,14 +49,17 @@ class TestWebAuth:
             assert not a.is_valid_password("")
             assert not a.is_valid_password("test")
 
-    @pytest.mark.parametrize("web_host,web_port,expected_web_url", [
-        ('example.com', 8080, 'http://example.com:8080/'),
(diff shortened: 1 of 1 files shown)


mitmproxy/tools/web/webaddons.py
```diff
@@ -62,8 +62,11 @@ def web_url(self) -> str:
             auth = ""  # We don't want to print plaintext passwords (and it doesn't work for argon2 anyhow).
         else:
             auth = f"?token={self._password}"
+        web_host = ctx.options.web_host
+        if ':' in web_host:  # ipv6
+            web_host = f'[{web_host}]'
```
Julien00859: Alternatively:

```py
import ipaddress

try:
   ipaddress.IPv6Address(web_host)
except ValueError:
    pass
else:
    web_host = f'[{web_host}]'
```

But I've seen the `':' in host` pattern to tell if an host is an ipv6 address in some other softwares, I guess it is fine.

Julien00859:
I've tested it locally, tell me if you want an unittest. I don't think it is worth one.

PR #7964, opened 2025-11-07
Mitigate decompression bombs DoS

Cycloctane:
#### Description

mitmproxy does not limit the size and memory consumption when decoding http messages bodies. Clients or servers can easily crash mitmproxy by sending a single request/response with highly compressed data via proxy.

This pr makes mitmproxy use `body_size_limit` in option to limit decompressed body size of incoming http messages, protecting it from zip bombs DoS attacks.

http bodies decoding now raises `ValueError` as decoding fails if decompressed data length exceeds `body_size_limit`.

#### Checklist

### DIFF
--- a/mitmproxy/http.py
+++ b/mitmproxy/http.py
@@ -386,7 +386,9 @@ class Message(serializable.Serializable):
         else:
             self.headers["content-length"] = str(len(self.raw_content))
 
-    def get_content(self, strict: bool = True, size_limit: int | None = None) -> bytes | None:
+    def get_content(
+        self, strict: bool = True, size_limit: int | None = None
+    ) -> bytes | None:
         """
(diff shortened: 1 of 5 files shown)


Cycloctane:
@mhils This is the PR mentioned in that advisory. Please take a look :)
This feature needs brotli v1.2.0. So it can not pass the old-dependencies tests.

mhils (commented):
Thanks for the PR and sorry for the slow turnaround! Overall I'm happy to add this DoS mitigation, but I don't think it warrants a new property on `HttpMessage`. Can we just use an option and access that directly please?

Cycloctane:
Could you please expand on that? This feature needs size limit settings from context options. But it seems that not all places that decode http messages have access to the current context.

PR #7967, opened 2025-11-09
fix: add flush to buffered file handle

### LINKED ISSUE #7926: Flow file is not written in 12.2.0
### Problem Description

When using `-w` flag in `mitmdump`, it doesn't write the flow to the file. It comes out empty. I can reproduce the problem with the latest release `12.2.0`. When I downgrade to the previous version of `12.1.2`, `-w` flag works as expected and flow is written to the file specified.

Looks like there is a regression between 12.1.2 and 12.2.0.

1. `mitmdump --mode reverse:https://www.google.com -p 6000 -w foo.flow`
2. `curl http://localhost:6000/foobarbaz`
3. See in the console that the request has been forwarded.
4. Look at the contents of the file `foo.flow`, it's empty.
5. Ensure it by issuing `du -hs foo.flow` command.

### System Information

```raw
## Affected version

Mitmproxy: 12.2.0 binary
Python:    3.14.0
Platform:  macOS-15.6.1-arm64-arm-64bit-Mach-O


## Working version

Mitmproxy: 12.1.2 binary
Python:    3.13.7
OpenSSL:   OpenSSL 3.5.2 5 Aug 2025
Platform:  macOS-15.6.1-arm64-arm-64bit-Mach-O
```

### Checklist

caiquejjx:
#### Description
Closes #7926.
- Call `flush` in FlowWriter after every add

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 -->
 
 ## Unreleased: mitmproxy next
-- Disables buffering for flow file
+
+- Flush flow file after each flow to allow further processing.
   ([#7967](https://github.com/mitmproxy/mitmproxy/pull/7967), @caiquejjx)
 - infer_content_encoding: Fallback to UTF-8 for more content types
   ([#7961](https://github.com/mitmproxy/mitmproxy/pull/7961), @xu-cheng)


sujaldev:
LGTM, this should be even better for the mentioned usecase than having a small buffer.

sujaldev:
On second thought, this is equivalent to having no buffering at all (I tested both flush and no buffering with strace), so perhaps we should just disable it here instead of manually flushing everywhere:

https://github.com/mitmproxy/mitmproxy/blob/c356c40b652a5742bd2bd2b2cb64f89a1e3aeba1/mitmproxy/addons/save.py#L100

caiquejjx:
yeah I agree @sujaldev makes no sense to use buffering if we end up flushing on every write

mhils:
(CI failure is unrelated, ignore)

Can we please just call flush in `FilteredFlowWriter.add` after writing a flow? If I read `unbuffered=True` in two years I need to go look at the tnetstring sources to check if that's ok, if I read `.flush()`, I know it's fine.

caiquejjx:
yeah, done

mitmproxy/addons/save.py
```diff
@@ -97,7 +97,7 @@ def maybe_rotate_to_new_file(self) -> None:
         new_log_file = Path(path)
         new_log_file.parent.mkdir(parents=True, exist_ok=True)
 
-        f = new_log_file.open(_mode(ctx.options.save_stream_file))
+        f = new_log_file.open(_mode(ctx.options.save_stream_file), buffering=False)
```
sujaldev: I think you missed reverting this:
```suggestion
        f = new_log_file.open(_mode(ctx.options.save_stream_file))
```

mhils (approved):

mhils:
Thanks! 🍰 🚀

Issue #7969, opened 2025-11-09
mitmweb: Gulp -> Vite


mhils:
Breaking this out of https://github.com/mitmproxy/mitmproxy/pull/7784: Let's move mitmweb to Vite. 

- PEG compilation can be taken out of the compile flow. Eventually this should go anyhow.

Desired outcomes: HMR and at least equivalent bundle size/performance.

PR #7971, opened 2025-11-09
Move mitmweb to Vite

mhils:
This PR moves mitmweb from gulp + esbuild to Vite.

- To make XSRF Cookies work with the Vite server, we need to partially roll back #7491. That's not super terrible though, we keep the most important bits.
- I've decided to ignore the PEG parts for now, it's not part of the build pipeline anymore. We want to get rid of that anyhow.
- Cachebusting is now done by Vite and not tornado's `static_path`.
- A couple of annoying fixes: Vite's esbuild mangles function names by default, which broke some code relying on `.name`. Took that as an opportunity to refactor those bits instead of configuring Vite.

Fixes https://github.com/mitmproxy/mitmproxy/issues/7969.

### DIFF
--- a/release/release.py
+++ b/release/release.py
@@ -79,7 +79,10 @@ if __name__ == "__main__":
     print("➡️ Updating web assets...")
     subprocess.run(["npm", "ci"], cwd=root / "web", check=True, capture_output=True)
     subprocess.run(
-        ["npm", "run", "ci-build-release"], cwd=root / "web", check=True, capture_output=True
+        ["npm", "run", "ci-build-release"],
+        cwd=root / "web",
+        check=True,
+        capture_output=True,
     )
     subprocess.run(
         ["git", "add", "--all", f"mitmproxy/tools/web"], cwd=root, capture_output=True


mitmproxy/tools/web/app.py
```diff
@@ -918,11 +919,12 @@ def __init__(
             template_path=os.path.join(os.path.dirname(__file__), "templates"),
             static_path=os.path.join(os.path.dirname(__file__), "static"),
             xsrf_cookies=True,
-            xsrf_cookie_kwargs=dict(httponly=True, samesite="Strict"),
```
lups2000, line 921: why did you remove it?

mhils: Necessary to make the Vite server work. We used to pass the XSRF token in index.html, but that does not work if Vite controls index.html. See the change in web/src/js/utils.ts. :)

mhils:
> The code looks good to me! I did some testing and ran into issues related to XSRF.
> Specifically, I wasn’t able to add a new dummy flow by pressing ‘n’, edit an incoming flow, or delete one etc
> Example error:
> `XSRF cookie does not match POST argument`

Thanks, I can repro. The workaround is to currently reload the page once, but we shall fix that properly. :)

web/README.md
```diff
@@ -4,10 +4,12 @@
...
   [upgrade](https://github.com/nodesource/distributions/blob/master/README.md#installation-instructions).
-- Run `cd mitmproxy/web` to change to the directory with package.json
-- Run `npm install` to install dependencies
-- Run `npm start` to start live-compilation
-- Run `mitmweb` after activating your Python virtualenv (see [`../CONTRIBUTING.md`](../CONTRIBUTING.md)).
+- Run `cd ./web` to change to the directory with package.json.
+- Run `npm install` to install dependencies.
+
+
+- Run `uv run mitmweb`
```
sleeyax, line 9: Nitpick: this looks weird in text format and has no visual implications when rendered on GitHub.

```suggestion
- Run `npm install` to install dependencies.
- Run `uv run mitmweb`
```

web/README.md
```diff
@@ -4,10 +4,12 @@
 - Run `node --version` to make sure that you have at least Node.js 18 or above. If you are on **Ubuntu <= 22.04**, you
```
sleeyax, line 4: Note: Node.js v24+ includes [npm v11](https://docs.npmjs.com/cli/v11/using-npm/changelog), which introduces additional changes to `package-lock.json` (visible as soon as you run `npm install` using said version). 

This could be a separate issue, but we should consider one of the following:
- Document a specific supported Node.js version (e.g., v22) in the `README.md`
- Provide a `.nvmrc` file (or equivalent) to pin the Node.js version to v22 or below.
- Upgrade to v24+ `package-lock.json` and document that as the minimum required version.

web/package.json
```diff
@@ -3,8 +3,8 @@
     "private": true,
     "scripts": {
         "test": "eslint . && tsc --noEmit && jest --coverage",
-        "build": "gulp prod",
-        "start": "gulp",
+        "build": "rm -rf ../mitmproxy/tools/web/static && vite build",
```
sleeyax: AFAIK `rm` doesn’t exist in classic Command Prompt (CMD) on Windows. I'm a proud Arch Linux user myself, but we should take other operating systems into account.

A classic way to solve this is to install a cross-platform utility package such as [rimraf](https://www.npmjs.com/package/rimraf):

```
npm i --save-dev rimraf
```

```suggestion
        "build": "rimraf ../mitmproxy/tools/web/static && vite build",
```

web/src/css/vendor-bootstrap.less
```diff
@@ -5,7 +5,7 @@
 // Reset and dependencies
 @import "../../node_modules/bootstrap/less/normalize.less";
 @import "../../node_modules/bootstrap/less/print.less";
-@import "../../node_modules/bootstrap/less/glyphicons.less";
+// @import "../../node_modules/bootstrap/less/glyphicons.less";
```
sleeyax: Why is this commented out? If it's no longer in use we can delete it instead of commenting it out :)

web/src/js/ducks/options.ts
```diff
@@ -1,6 +1,7 @@
 import { fetchApi } from "../utils";
 import * as optionsEditorActions from "./ui/optionsEditor";
-import { defaultState, Option, OptionsState } from "./_options_gen";
+import { defaultState, OptionsState } from "./_options_gen";
```
sleeyax, line 3: Nitpick: for consistency with existing code in this PR (this syntax can/should also be enforced by a eslint rule):

```suggestion
import { defaultState, type OptionsState } from "./_options_gen";
```

sleeyax (changes requested):
Nice work! Left a few minor comments.

web/package.json
mhils: I'm in favor of making this cross-platform, but not at the cost of adding dependencies. Can't add a comment in the JSON file, so I renamed this to `ci-build-release`. :)

web/README.md
mhils: +1 to solving this outside of this PR, open to suggestions on what's most sensible. :)

web/src/js/ducks/options.ts
mhils: +1, but will be a separate PR due to size.

web/package.json
sleeyax: It's a dev dependency. Meaning it is only installed locally on your dev machine and never shipped to users. 

Unless I'm still missing something I don't see the problem?

PR #7990, opened 2025-11-23
Docs: Change 'uvx' to 'uv tool install' for Installation from PyPI

larpitect:
#### Description
Changed references to 'uvx' to 'uv tool install'.

'uvx' is an alias for 'uv tool run' that installs packages in a temporary virtual environment that is deleted when the uv cache is cleaned. The correct installation command should be 'uv tool install' which installs the virtual environment in the uv tools directory and makes the executable available on the PATH. See: https://docs.astral.sh/uv/concepts/tools/

#### Checklist


mhils (approved):
Thanks! 🍰

PR #7994, opened 2025-11-25
added clearer error hint for wireguard port conflicts

jagadeeshkaruturi11:
hey! i noticed in issue #7650 that the error message for WireGuard port conflicts wasn't super clear - it just said to try a different port but didn't mention that you might already have WireGuard running which is a common cause.

so i added a quick check that when we detect a port conflict AND we're in WireGuard mode, we now tell users "If you already have WireGuard running, stop it first or use a different port."

should help folks troubleshoot faster instead of scratching their heads about why the port's already taken.

let me know if you want me to change anything!


mhils:
https://github.com/mitmproxy/mitmproxy/pull/8170#issuecomment-4231673802

PR #8028, opened 2025-12-28
Android Magisk module: Add cert to cacerts_google

gucci-on-fleek:
#### Description

Currently, the Magisk module only copies the certificate to `/system/etc/security/cacerts/`. However, even after I installed this module on my Android device (Pixel 3a, Android 11, Magisk 23.0, Magisk Hide disabled, no other modules enabled), the mitmproxy CA certificate remained untrusted by all apps.

Looking through the filesystem, I found an additional folder `/system/etc/security/cacerts_google/` with about half as many certificates as `cacerts/`. Modifying the Magisk module to also copy the certificate to this folder made the mitmproxy CA trusted by all apps.

Note: I've tested this by manually adding this file to the `.zip` file created by mitmproxy, but I have not tried rebuilding mitmproxy with this change.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   `/system/etc/security/cacerts_google/`, which helps with compatibility
   on Pixel 3a devices running Android 11 and Magisk 23.0, and may
   potentially help with other Android devices as well.
+  ([#8028](https://github.com/mitmproxy/mitmproxy/pull/8028), @gucci-on-fleek)
 - Update optmanager value parsing exceptions to include the option name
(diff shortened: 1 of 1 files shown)


gucci-on-fleek:
@mhils

> Can you extend the comment to explain what kind of devices?

I've only tested this on a Pixel 3a running Android 11 and Magisk 23.0, so I'm not really sure what other devices this is necessary for. I can write something like the following if you want 

```diff
-   compatiblity on some devices.
+   compatiblity on Pixel 3a devices running Android 11 and Magisk 23.0 (and
+   possibly other unknown devices).
```

but I'm not sure how useful that is.

mhils:
That'd be great, thanks!

PR #8031, opened 2025-12-30
fix: content encoding inference from meta tag, especially for various unusual format

edenw97:
fix the content encoding inference from meta tag, especially for various unusual format(unquoted, whitespace), follows https://html.spec.whatwg.org/#extracting-character-encodings-from-meta-elements

#### Description
this PR address multiple edge case when we infer content-encoding in the meta tag

#### Checklist

### DIFF
--- a/test/mitmproxy/net/http/test_headers.py
+++ b/test/mitmproxy/net/http/test_headers.py
@@ -147,12 +147,12 @@ def test_assemble_content_type():
         ),
         (
             "text/html",
-            b"<meta charset=''",
+            b"<meta charset=''>",
             "utf8",
         ),
         (
             "text/html",
-            b"<meta charset=",
+            b"<meta charset=>",
             "utf8",
         ),
         # xml declaration encoding


mitmproxy/net/http/headers.py
```diff
@@ -63,11 +63,23 @@ def infer_content_encoding(content_type: str, content: bytes = b"") -> str:
...
     if not enc and "html" in content_type:
+        # Strict implementation of your provided algorithm
+        # 1. charset       : Matches literal "charset"
+        # 2. \s*=\s*       : Handles whitespace before AND after the equals sign
+        # 3. (?: ... )     : Non-capturing group for the OR logic
+        # 4. (["'])(.*?)\1 : Matches a quote, then anything lazily, then the MATCHING quote (\1)
+        # 5. ([^'"";\s]+)     : Unquoted - stops at quote, Semicolon (;) or Whitespace
         meta_charset = re.search(
-            rb"""<meta[^>]+charset=['"]?([^'">]+)""", content, re.IGNORECASE
+            rb"""<meta[^>]+charset\s*=\s*(?:(["'])(.*?)\1|([^'";\s]+))""",
```
mhils:
```suggestion
            rb"""<meta[^>]+charset\s*=\s*(?:(["'])(.*?)\1|([^'";>\s]+))""",
```
so that we correctly match `<meta charset=utf8>`?

mitmproxy/net/http/headers.py
```diff
@@ -63,11 +63,23 @@ def infer_content_encoding(content_type: str, content: bytes = b"") -> str:
...
+            content,
+            re.IGNORECASE,
         )
         if meta_charset:
-            enc = meta_charset.group(1).decode("ascii", "ignore")
+            match = meta_charset.group(2) or meta_charset.group(3)
+            if match:
+                enc = match.decode("ascii", "ignore").strip()
+            else:
```
mhils, line 81: We already validated that `meta_charset` is not None, so can this case ever happen?

(`charset=""` would be an option, but then we should move the `.strip()` before the `if` as well)

edenw97: Hi, do you mean move the strip() to before the decode?

PR #8057, opened 2026-01-30
Mitmweb: Add number of selected flows in the footer

skrattara:
#### Description

Added a quality of life feature where you can see both the number of selected and total flows at the bottom. It's handy if you have lots of flows selected and more get added in real-time.

I have tested the following cases:

- [x] When one flow is selected out of one or multiple
- [x] When multiple flows are selected out of one or multiple
- [x] When no flows are selected out of one or multiple
- [x] When one or more flows are added
- [x] When some or all flows are deleted/cleared

I haven't written tests for this because I'm not too familiar with the testing suite.

Screenshot:
Screenshot 2026-01-30 112231

#### Checklist

### DIFF
--- a/web/src/js/components/Footer.tsx
+++ b/web/src/js/components/Footer.tsx
@@ -24,7 +24,9 @@ export default function Footer() {
         ssl_insecure,
     } = useAppSelector((state) => state.options);
 
-    const selectedFlowsLength = useAppSelector((state) => state.flows.selected.length);
+    const selectedFlowsLength = useAppSelector(
+        (state) => state.flows.selected.length,
+    );
     const totalFlowsLength = useAppSelector((state) => state.flows.list.length);
 
     return (


skrattara:
I think this is ready for review.

web/src/js/components/Footer.tsx
```diff
@@ -24,6 +25,15 @@ export default function Footer() {
...
 
+    // Create a memoized selector to avoid failing tests
+    const num_flow_selector = createSelector(
+        (state) => state.flows.selected.length,
+        (state) => state.flows.list.length,
+        (selectedLength, listLength) => [selectedLength, listLength],
+    );
+
+    const num_flows = useAppSelector(num_flow_selector);
+
```
lups2000: question: why not simply something like this?
```
const selectedFlowsLength = useAppSelector((state) => state.flows.selected.length);
const totalFlowsLength = useAppSelector((state) => state.flows.list.length);
```
```
{totalFlowsLength > 0 && (
    <span className="label label-default">
          {selectedFlowsLength} of {totalFlowsLength} flows selected
    </span>
)}
```

skrattara: That could work, but the main reason I implemented it this way is because Github Actions complained about not memoizing the selector. This way, the component only re-renders if the number of flows change.

```
Error: Selector unknown returned a different result when called with the same parameters. This can lead to unnecessary rerenders.
    Selectors that return a new reference (such as an object or an array) should be memoized: https://redux.js.org/usage/deriving-data-selectors#optimizing-selectors-with-memoization
```

I guess I could do this, but I didn't want to pollute the commit history even more.
```js
const [selectedFlowsLength, totalFlowsLength] = useAppSelector(num_flow_selector);
```

skrattara:
Oops, accidentally clicked the merge button while trying to resolve the changes in CHANGELOG.md. Oh well, we ball 😅

Edit: it wasn't merged

web/src/js/components/Footer.tsx
lups2000: I guess it's because you were returning an array (reference). I think the proposed change is a bit clearer and simpler; it shouldn't raise any error :)

lups2000:
> Oops, accidentally clicked the merge button while trying to resolve the changes in CHANGELOG.md. Oh well, we ball 😅
> 
> Edit: it wasn't merged

No worries :) I think you don't have permissions to merge anyway 😅

web/src/js/components/Footer.tsx
skrattara: Done.

lups2000 (approved):
LGTM now thanks :)

PR #8082, opened 2026-02-12
fix: Resolve Sentry module import mismatch in bundled addon

hirak214:
## Summary
- Fixed `oximy_logger.py` importing a stale `sentry_service` from site-packages instead of the local bundle copy when running in the Mac app
- Added second sed pattern in `sync-addon.sh` to convert `from mitmproxy.addons.oximy import X` → `import X` (previously only `from mitmproxy.addons.oximy.X` was handled)
- Added try/except import fallback in `oximy_logger.py` matching the pattern used in addon.py, process.py, etc.

## Test plan
- [x] 447 tests pass
- [x] `make build` succeeds
- [x] Bundled `oximy_logger.py` uses `import sentry_service` (not `from mitmproxy.addons.oximy import`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/OximyMac/Resources/oximy-addon/collector.py
+++ b/OximyMac/Resources/oximy-addon/collector.py
@@ -30,7 +30,7 @@ except ImportError:
     HAS_WATCHFILES = False
 
 try:
-    from mitmproxy.addons.oximy import sentry_service
+    import sentry_service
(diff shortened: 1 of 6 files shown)


hirak214:
Opened against wrong repo by mistake, closing.

PR #8104, opened 2026-02-27
web: Reduce FlowTable Redux subscriptions from O(rows) to O(1)

### LINKED ISSUE #8106: web: Reduce FlowTable Redux subscriptions from O(rows) to O(1)
### Problem

The current FlowTable implementation creates a separate Redux subscription for every rendered row via `useAppSelector` in `FlowRow`, and an additional subscription per cell in the `index` column via `useAppSelector` inside `FlowColumns.index`. With 100 visible rows and 8 columns, this creates ~100 subscriptions from FlowRow plus ~100 from the index column — each running their selector on every Redux state change.

This causes measurable jank and dropped frames during rapid flow updates, particularly noticeable when:
- The table has 500+ rows
- Flows are updating in real-time (e.g., proxy capturing live traffic)
- Lower-powered machines or tabs with constrained CPU

### Root Cause

In `FlowRow.tsx`, `displayColumnNames` is fetched via `useAppSelector((state) => state.options.web_columns)` **inside each row**. Since this value is the same for every row, this creates N identical subscriptions.

In `FlowColumns.tsx`, the `index` column calls `useAppSelector((state) => state.flows._listIndex.get(flow.id))` to look up the row's display index — one subscription per visible row.

### Proposed Solution

1. **Lift `displayColumnNames` to `FlowTable`**: Fetch `state.options.web_columns` once in `FlowTable`'s `mapStateToProps` and pass it down as a prop. Eliminates N identical subscriptions.

2. **Lift `rowNumber` to `FlowTable`**: Compute the row number from the `.slice()` offset (`vScroll.start + idx`) and pass it as a prop. Eliminates per-cell subscription in the index column.

3. **Wrap `FlowRow` in `React.memo()`**: With `displayColumnNames` and `rowNumber` as props instead of internal subscriptions, `memo()` can effectively skip re-renders when the row data hasn't changed.

### Files Changed

| File | Change |
|------|--------|
| `FlowTable.tsx` | Add `displayColumnNames` to `mapStateToProps`, pass `rowNumber` to `FlowRow` |
| `FlowTable/FlowRow.tsx` | Accept `displayColumnNames` and `rowNumber` as props, wrap in `memo()`, remove `useAppSelector` |
| `FlowTable/FlowColumns.tsx` | Accept required `rowNumber` prop in `index` column, remove `useAppSelector` |

### Impact

- **Before**: ~2N Redux subscriptions (N rows × 2 selectors)
- **After**: 1 Redux subscription (at FlowTable level)
- Zero behavioral change — columns, sorting, selection all work identically

ariel42:
#### Description

Reduces unnecessary Redux subscriptions in the FlowTable by lifting shared state to the parent component level.

Fixes #8106

Each `FlowRow` independently subscribes to `state.options.web_columns` via `useAppSelector`, creating N identical subscriptions across all visible rows. The `index` column additionally subscribes to `state.flows._viewIndex` per cell. With large flow tables (500+ rows), this creates measurable selector overhead on every Redux state change.

**Changes:**
- **`FlowTable.tsx`**: Fetch `displayColumnNames` from `state.options.web_columns` in `mapStateToProps` and pass it down to `FlowRow` as a prop.
- **`FlowRow.tsx`**: Accept `displayColumnNames` and `rowNumber` as props instead of using `useAppSelector`. Wrap component in `React.memo()` for efficient re-render skipping.
- **`FlowColumns.tsx`**: Accept optional `rowNumber` prop in the `index` column component, falling back to `flow.id` if not provided. Remove the per-cell `useAppSelector` call.

Verified with 1000+ flows — no visual or behavioral regressions. Column display, sorting, selection, and virtual scrolling all work identically.

#### Checklist


ariel42:
This PR is ready to merge.

web/src/js/__tests__/components/FlowTable/FlowRowSpec.tsx
```diff
@@ -3,15 +3,37 @@ import FlowRow from "../../../components/FlowTable/FlowRow";
 import { fireEvent, render, screen } from "../../test-utils";
 import { TStore } from "../../ducks/tutils";
 
+const defaultColumns = [
```
mhils: Why is this necessary?

web/src/js/components/FlowTable/FlowColumns.tsx
```diff
@@ -42,11 +43,12 @@ export const tls: FlowColumn = ({ flow }) => {
...
-export const index: FlowColumn = ({ flow }) => {
-    const index = useAppSelector(
-        (state) => state.flows._listIndex.get(flow.id)!,
-    );
-    return <td className="col-index">{index + 1}</td>;
+export const index: FlowColumn = ({ flow, flowIndex }) => {
+    // Use the index passed from the parent when available, avoiding a
+    // per-cell Redux subscription that would otherwise fire on every
+    // state change.
+    const displayIndex = flowIndex !== undefined ? flowIndex + 1 : flow.id;
```
mhils: Why would flowIndex ever be undefined?

web/src/js/components/FlowTable/FlowRow.tsx
```diff
@@ -1,25 +1,26 @@
...
+import { useAppDispatch } from "../../ducks";
 import { select, selectRange, selectToggle } from "../../ducks/flows";
 import * as columns from "./FlowColumns";
 
 type FlowRowProps = {
     flow: Flow;
     selected: boolean;
     highlighted: boolean;
+    displayColumnNames: string[];
+    flowIndex: number;
```
mhils: Is `rowNumber` maybe better than `flowIndex`?

web/src/js/components/FlowTable/FlowTableHead.tsx
```diff
@@ -15,7 +15,7 @@ export default React.memo(function FlowTableHead() {
     );
 
     const sortType = sortDesc ? "sort-desc" : "sort-asc";
-    const displayColumns = displayColumnNames
+    const displayColumns = (displayColumnNames || [])
```
mhils: Why is this necessary?

mhils (commented):
Thank you for the PR! This looks really promising overall, I've only left a few small nits below. How did you measure performance? Can you share some rough before/after screenshots of your profiler?

ariel42:
Thanks for the review! I've addressed all feedback in 61fc416:

- **`rowNumber`**: Renamed `flowIndex` → `rowNumber` as suggested.
- **Required prop**: `rowNumber` is now required in `FlowColumnProps` (not optional), so the undefined check and fallback are gone.
- **`|| []` guard removed**: `web_columns` is always initialized by the options reducer, so the null guard in `FlowTableHead` was unnecessary — reverted.
- **Test constant removed**: `FlowRowSpec` now reads `displayColumnNames` from the test store (`store.getState().options.web_columns`) instead of a hardcoded array.

Regarding profiler screenshots — I don't currently have a convenient setup to capture before/after profiles on two branches (my environment uses a fork with additional infrastructure that makes branch-switching non-trivial). That said, the improvement here is structural rather than a micro-optimization: we're replacing 2N Redux subscriptions (one `useAppSelector` per row for `web_columns`, plus one per row for `_listIndex`) with a single selector at the `FlowTable` level. The complexity reduction from O(rows) to O(1) holds regardless of the workload. In my day-to-day usage with hundreds to thousands of concurrent flows, scrolling has been smooth — but the unnecessary selector overhead was there nonetheless, and it compounds on lower-powered devices or under CPU pressure.

Happy to capture profiler screenshots if you consider them essential for merging, but I hope the structural argument is sufficient.

web/src/js/__tests__/components/FlowTable/FlowRowSpec.tsx
ariel42: Removed — the test now reads `web_columns` from the store instead of hardcoding column names. See main comment for details.

web/src/js/components/FlowTable/FlowColumns.tsx
ariel42: You're right, it wouldn't — made `rowNumber` required and removed the fallback. See main comment.

web/src/js/components/FlowTable/FlowRow.tsx
ariel42: Agreed — renamed to `rowNumber` throughout.

web/src/js/components/FlowTable/FlowTableHead.tsx
ariel42: It wasn't — `web_columns` is always initialized by the reducer. Reverted.

ariel42:
Good catch @themavik — you're right, `vScroll.start + idx` gives the view position (after sort/filter), not the original flow number. The old code used `_listIndex` (insertion order), so the `#` column would drift when sorted.

Fixed in 9e4b9f2: `rowNumber` now comes from `_listIndex.get(flow.id)!` (fetched once at the `FlowTable` level via `mapStateToProps`), preserving the original behavior. Also added a test for the index column display.

ariel42:
@mhils you are welcome! Thank you for your awesome project!

PR #8107, opened 2026-02-28
feat: Upload worker thread — prevent internet blocking when API is down

hirak214:
## Summary
- **Upload worker thread**: Moves all upload I/O off mitmproxy's event loop into a dedicated daemon thread. The `response()` hook now does O(1) buffer append + event signal (~microseconds) instead of blocking up to ~16s on retries.
- **Fail-open on buffer full**: When memory buffer is full, traces are dropped instead of blocking on synchronous disk I/O. Upload worker is signalled to drain ASAP.
- **Events Pending UI fix**: In-memory buffer count now exposed via `remote-state.json` → Swift `RemoteStateService`, so the macOS app shows accurate pending event count instead of always 0.
- **Force sync fix**: `upload_all(force=True)` now correctly bypasses the upload circuit breaker.
- **E2E test suite**: 6 scenarios (19/19 checks) covering normal upload, API failure, recovery, buffer full, cold start, and force sync.

## Test plan
- [x] 856 unit tests pass (`python -m pytest mitmproxy/addons/oximy/tests/ -v`)
- [x] E2E scenarios 1-6 all pass (19/19 checks, 67s)
- [x] Internet stays fast (~0.6s for 5 requests) with API down (was ~48s before fix)
- [x] macOS app builds successfully and shows buffer count in Events Pending

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/mitmproxy/addons/oximy/addon.py
+++ b/mitmproxy/addons/oximy/addon.py
@@ -4947,7 +4947,6 @@ class OximyAddon:
         def upload_loop():
             logger.info("Upload worker thread started")
             while not self._upload_stop.is_set():
-                # Wait for signal or timeout — interruptible sleep
                 self._upload_trigger.wait(timeout=self._upload_interval_seconds)
                 self._upload_trigger.clear()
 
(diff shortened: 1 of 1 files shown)


lups2000:
Please stop submitting these large AI-generated PRs, as they will simply be closed.
Before submitting, please review your changes carefully and avoid sending PRs like this in the future. Thanks!

PR #8126, opened 2026-03-12
Default text encoding to utf8 instead of (very) legacy latin-1

emaheuxPEREN:
#### Description

In `infer_content_encoding`, we set the fallback encoding to the UTF-8 standard instead of the (very) legacy Latin-1 encoding.

As a consequence, `get_text()` outputs (and thus HAR dumps) may differ from previous version, advocating for more than a patch release.

(Note that to reach full coverage I had to perform a mini-refact in `mitmproxy.io.har` -> extract a little untested snippet into an unitary testable function)

#### Checklist


themavik (commented):
utf8 default plus HAR fixture updates match the behavioral shift; worth watching for any scripts that still assumed latin-1 on missing charset.

emaheuxPEREN:
> worth watching for any scripts that still assumed latin-1 on missing charset.

could not find any more script assuming latin-1 as default charset

PR #8129, opened 2026-03-15
Add individual exit codes for different error conditions

### LINKED ISSUE #5696: Add individual exit codes for different errors

kovan:
## Summary

- Replace all `sys.exit(1)` calls with specific exit codes so parent processes can distinguish failure reasons
- Add `mitmproxy/utils/exit_codes.py` defining grouped exit codes:
  - `10` — startup error (addon load failure, etc.)
  - `20` — no TTY (console requires interactive shell)
  - `30` — invalid arguments (unrecognized CLI flags)
  - `31` — invalid options (bad `--set` values)
  - `40` — cannot print (stdout broken)
  - `41` — cannot write to file (disk full, etc.)
- Tests for all testable exit code paths, including INVALID_ARGS and INVALID_OPTIONS which stumped previous attempts

Picks up from #5133 and #5882.

Fixes #5696

## Test plan

- [x] `test_exit_codes.py` — exit codes are distinct, SUCCESS is 0
- [x] `test_errorcheck.py` — startup errors exit with STARTUP_ERROR
- [x] `test_termlog.py` — print failure exits with CANNOT_PRINT
- [x] `test_save.py` — write failure exits with CANNOT_WRITE_TO_FILE
- [x] `test_main.py` — bad flags exit with INVALID_ARGS, bad options exit with INVALID_OPTIONS
- [x] All 23 tests pass
- [x] ruff clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ## Unreleased: mitmproxy next
 
 - Add individual exit codes for different error conditions instead of always exiting with 1.
+  ([#8129](https://github.com/mitmproxy/mitmproxy/pull/8129), @kovan)
 - Update optmanager value parsing exceptions to include the option name
   ([#8016](https://github.com/mitmproxy/mitmproxy/pull/8016), @zdwg42)
 - mitmweb: show intercept filter tag at the bottom for default options


kovan:
Valid concern. Well-written shell scripts already check for non-zero (`if cmd; then` or `|| exit`), but scripts that explicitly check `== 1` would need updating. This is a backwards-incompatible change and should be called out in the changelog. I'll add a note in the PR description about the migration path.

PR #8144, opened 2026-03-25
fix: preserve newlines in multipart form data decoding

### LINKED ISSUE #4466: Error parsing multipart with newline in the content
#### Problem Description

```python
from mitmproxy import http
from mitmproxy import ctx

    if flow.request.path:
                if k == b"data":
                    ctx.log.info(v)
```

#### System Information

$ mitmproxy --version
Python:    3.8.5
(venv)

Krishnachaitanyakc:
## Summary

Fixes #4466 — multipart form data values containing newlines (`\n` or `\r\n`) were silently stripped during decoding, corrupting text fields with line breaks and binary file uploads (e.g. JPEG images).

- **Root cause**: `decode_multipart` used `bytes.splitlines()` to parse each part and then joined body lines with `b""`, which discarded all newline characters from the body content.
- **Fix**: Replace the `splitlines()` + join approach with `bytes.find()` to locate the header/body separator (`\r\n\r\n` or `\n\n`), keeping the body as a single contiguous `bytes` object. Only the trailing CRLF/LF that RFC 2046 defines as part of the boundary delimiter is stripped.
- **Tests**: Updated the existing `test_decode` to use RFC-compliant CRLF delimiters and added new test cases for:
  - Embedded CRLF newlines in values
  - Bare LF newlines in values
  - Binary content with embedded `\r\n` and `\n` bytes
  - Encode/decode round-trip key recovery

## Test plan

- [x] Verify existing `test_decode` passes with CRLF-delimited payloads
- [x] Verify `test_decode_with_lf` passes for bare-LF payloads
- [x] Verify `test_decode_content_preserves_newlines` catches the original bug (value `a\r\nb` not flattened to `ab`)
- [x] Verify `test_decode_binary_content` ensures binary data with embedded `\r\n`/`\n` bytes is preserved
- [x] Verify `test_decode_roundtrip` confirms encode then decode recovers field names
- [x] Verify `test_encode` still passes unchanged

### DIFF
--- a/mitmproxy/net/http/multipart.py
+++ b/mitmproxy/net/http/multipart.py
@@ -30,7 +30,6 @@ def encode_multipart(content_type: str, parts: list[tuple[bytes, bytes]]) -> byt
                     hdrs.append(b"Content-Type: %b" % file_type.encode("utf-8"))
                     hdrs.append(b"")
                     hdrs.append(value)
-                hdrs.append(b"")
 
                 if value is not None:
                     # If boundary is found in value then raise ValueError
--- a/mitmproxy/net/http/multipart.py
+++ b/mitmproxy/net/http/multipart.py
@@ -64,13 +64,39 @@ def decode_multipart(
         r = []
         if content is not None:
             for i in content.split(b"--" + boundary):
-                parts = i.splitlines()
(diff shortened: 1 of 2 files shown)


Krishnachaitanyakc:
@mhils can you please review this?

PR #8147, opened 2026-03-28
Web: Remove Bootstrap dependency and keep mitmweb styling parity

lups2000:
#### Description

Bootstrap has become technical debt and we were still relying on Bootstrap 3. This PR **completely removes** the Bootstrap dependency from mitmweb and replaces it with local styles, while preserving the existing UI look and behaviour. Class names can be subjective, so feel free to suggest any improvements in that sense.

To properly review this change, the best approach is to run the dev server and verify that the behavior and appearance remain identical.

This marks the first step in improving the mitmweb UI. Now that we are no longer dependent on Bootstrap, we have the flexibility to decide how to move forward, whether by using vanilla CSS or adopting more modern solutions like Tailwind CSS.

#### Checklist


web/src/css/global.less
```diff
@@ -0,0 +1,489 @@
...
+    padding-right: 24px;
+}
+
+textarea.input {
+    height: auto;
+}
+
+.has-error .input {
+    border-color: #a94442;
+}
```
mhils, line 489: This file was presumably copied over from bootstrap somehow? Is this a blanket copy, or only copying the relevant parts?

mhils (commented):
Fantastic - great job at eliminating Bootstrap! 🥳🥳🥳

I assume you did some manual testing to hunt for major rendering differentials?

web/src/css/global.less
lups2000: This file was created to include only the subset of selectors that were still effectively Bootstrap-dependent and used across the website. I kept the original class names to preserve existing behaviour and avoid a rename across buttons, forms, and other UI elements.

lups2000:
> Fantastic - great job at eliminating Bootstrap! 🥳🥳🥳
> 
> I assume you did some manual testing to hunt for major rendering differentials?

Yes, exactly! I had to manually review and fix the styling across the components so it would be completely identical. It is basically the old mitmweb, but no Bootstrap, only CSS :) I can upload a screen recording if you want to show it

mhils (approved):
Amazing! No screen recording needed. Let's ship a patch release and then get this in right after?

lups2000:
> Amazing! No screen recording needed. Let's ship a patch release and then get this in right after?

Yes. Makes sense :)

lups2000:
Can we merge this? :) Now the dilemma: plain css or tailwind and co ? 😅 I am oriented towards plain css to be honest. 
Btw not directly related to this, but we should maybe consider to upgrade/remove fontawesome (and its icons) as well.

mhils:
> plain css or tailwind and co ? 😅 I am oriented towards plain css to be honest.

Then let's do that! We can always adopt more frameworks if we need to, the other way is difficult.

> Btw not directly related to this, but we should maybe consider to upgrade/remove fontawesome (and its icons) as well.

👍👍👍

lups2000:
Awesome 🙌

lups2000:
Hey Max, I think this can be merged, right? Unfortunately I don't have the permissions to merge it in case there are failing checks 😅😢

mhils (approved):
🎉 🎉🎉

PR #8149, opened 2026-03-31
fix(web): allow clearing body content to empty in editor

ariel42:
#### Description

The body editor's save function in `HttpMessage.tsx` uses `editedContent || content || ""` which treats empty string as falsy. This makes it impossible to clear a body to empty via the Edit button — the save always falls back to the original content.

Changed to `editedContent ?? content ?? ""` so that only `undefined` (user didn't edit) falls back, while empty string is preserved as a valid edit.

Added `HttpMessageEditSpec.tsx` with two tests:
1. Saving empty body sends `""`, not the original content
2. Saving unedited body correctly falls back to original via `undefined ??`

#### Checklist


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 
 ## Unreleased: mitmproxy next
 
+- mitmweb: Fix body editor unable to clear content to empty.
```
lups2000:
```suggestion
- mitmweb: Fix editors not allowing content to be cleared to an empty string
```

lups2000 (changes requested):
Thanks for the PR :) This indeed resolves the issue!
Regarding the tests, could you please include them in the `HttpMessageSpec.tsx` file?
Other than that, this looks good to me 😄

ariel42:
Thanks for the review! Tests are now in `HttpMessageSpec.tsx` as requested.

Since the tests need a CodeEditor mock to capture `onChange` (CodeMirror isn't drivable in jsdom), and `jest.mock` is hoisted to file scope, the existing edit-mode snapshot was updated — the 86-line CodeMirror DOM dump is replaced by a lightweight `<textarea>` stub. All 304 tests pass.

web/src/js/__tests__/components/contentviews/__snapshots__/HttpMessageSpec.tsx.snap
```diff
@@ -114,95 +114,9 @@ exports[`HttpMessage 2`] = `
          Cancel
       </button>
     </div>
-    <div
-      class="codeeditor"
-    >
```
lups2000, line 119: I think the snapshot can remain unchanged, rendering the normal editor. Only the edit/save behavior tests should switch to the textarea mock for deterministic input control.

web/src/js/__tests__/components/contentviews/HttpMessageSpec.tsx
```diff
@@ -3,11 +3,34 @@ import * as React from "react";
...
     ViewImage,
 } from "../../../components/contentviews/HttpMessage";
-import { fireEvent, render, screen, waitFor } from "../../test-utils";
+import { act, fireEvent, render, screen, waitFor } from "../../test-utils";
 import fetchMock, { enableFetchMocks } from "jest-fetch-mock";
 
 enableFetchMocks();
 
+// Mock CodeEditor — CodeMirror is not drivable in jsdom.
+let capturedOnChange: ((content: string) => void) | null = null;
```
lups2000: you can use a flag: `mockUseCodeEditor` to control the behaviour


```suggestion
let mockUseCodeEditor = false,
    mockCapturedOnChange: ((content: string) => void) | null = null;

```

web/src/js/__tests__/components/contentviews/HttpMessageSpec.tsx
```diff
@@ -3,11 +3,34 @@ import * as React from "react";
...
+    }: {
+        initialContent: string;
+        onChange: (c: string) => void;
+    }) => {
+        capturedOnChange = onChange;
+        return (
+            <textarea data-testid="mock-editor" defaultValue={initialContent} />
+        );
+    },
+}));
```
lups2000:
```suggestion
jest.mock("../../../components/contentviews/CodeEditor", () => {
    const actual = jest.requireActual(
        "../../../components/contentviews/CodeEditor",
    );
    return {
        __esModule: true,
        default: ({
            initialContent,
            onChange,
        }: {
            initialContent: string;
            onChange: (c: string) => void;
        }) => {
            if (!mockUseCodeEditor) {
                return actual.default({ initialContent, onChange });
            }
            mockCapturedOnChange = onChange;
            return (
                <textarea data-testid="mock-editor" defaultValue={initialContent} />
            );
        },
    };
});
```

web/src/js/__tests__/components/contentviews/HttpMessageSpec.tsx
```diff
@@ -115,3 +138,70 @@ describe("HttpMessage Copy Button", () => {
         );
     });
 });
+
+describe("HttpMessage body edit", () => {
+    const CVD = { view_name: "Raw", description: "", syntax_highlight: "none" };
```
lups2000: nit: call it `cvd`, only for consistency across the file

web/src/js/__tests__/components/contentviews/HttpMessageSpec.tsx
```diff
@@ -115,3 +138,70 @@ describe("HttpMessage Copy Button", () => {
...
+
```
lups2000, line 151:
```suggestion
beforeEach(() => {
        mockUseCodeEditor = true;
        mockCapturedOnChange = null;
    });

    afterEach(() => {
        mockUseCodeEditor = false;
        mockCapturedOnChange = null;
    });
```

lups2000 (changes requested):
Thanks for addressing the comments so quickly 🙂 I’ve left a couple more.

The main thing to highlight is the snapshot; we could probably keep it unchanged and introduce a flag to switch the behaviour. Otherwise, the CodeMirror editor would always be rendered as a textarea. 

What do you think?

ariel42:
Thanks for the thorough review! All feedback has been addressed:

- **Conditional CodeEditor mock**: Introduced `mockUseCodeEditor` flag with `jest.requireActual` — the real CodeEditor renders by default, and only the body-edit tests switch to the textarea mock via `beforeEach`/`afterEach`.
- **Snapshot unchanged**: The existing snapshot now renders the real CodeMirror editor as before.
- **`CVD` → `cvd`**: Renamed for consistency.

All CI checks are green.

CHANGELOG.md
ariel42: Applied in bad6ad9.

ariel42: Applied in bad6ad9.

web/src/js/__tests__/components/contentviews/__snapshots__/HttpMessageSpec.tsx.snap
ariel42: Done — snapshot is now unchanged.

web/src/js/__tests__/components/contentviews/HttpMessageSpec.tsx
ariel42: Done.

web/src/js/__tests__/components/contentviews/HttpMessageSpec.tsx
ariel42: Done.

web/src/js/__tests__/components/contentviews/HttpMessageSpec.tsx
ariel42: Done.

web/src/js/__tests__/components/contentviews/HttpMessageSpec.tsx
ariel42: Done.

lups2000 (approved):
LGTM 🚀
CC: @mhils

PR #8150, opened 2026-04-01
Treat ALPN "grpc-exp" as HTTP/2

AlexTech01:
#### Description

This PR makes mitmproxy treat ALPN "grpc-exp" as HTTP/2, making it show properly as HTTP/2 in mitmproxy/mitmweb instead of as raw TCP. ALPN "grpc-exp" is used by some gRPC applications/servers instead of ALPN "h2", so this change makes it easier for users to mitm these applications without needing to patch them to use ALPN "h2" instead.

#### Checklist


mhils:
Are there any docs/standards for "grpc-exp"? I don't really love the idea of supporting random ALPNs.

PR #8164, opened 2026-04-04
Fix: HTTP/S proxy not starting on programmatic WebMaster launch (#8146)

### LINKED ISSUE #8146: Programmatic Python launch does not enable HTTP proxy
### Problem Description

Launching mitmweb or mitmdump from Python code (e.g. with `WebMaster`) doesn't enable the HTTP(S) proxy by default, even if the `mode` is set to `regular`. Opening the (alpha) options window in mitmweb, changing the `mode` to `regula` and immediately back to `regular` resolved the issue, even if no other options are changed.

1. Run the following Python code:
   ```python
   from asyncio import run, sleep
   
   from mitmproxy.options import Options
   from mitmproxy.tools.web.master import WebMaster
   
   async def main():
       master = WebMaster(Options())
       await master.running()
       for i in range(60):
           print(i + 1)
           await sleep(1)
       master.shutdown()
       await master.done()
   
   if __name__ == "__main__":
       run(main())
   ```
2. Open a (separate) browser with HTTP and HTTPS proxies set to `localhost:8080`. Navigate to any page, say `https://github.com/` You'll receive an error akin to "The proxy server is refusing connections".
3. In the mitmweb interface, go to the tab "Capture". Notice that "Run HTTP/S Proxy" is disabled.
4. In the mitmweb interface, go to the tab "Options". Click "Edit Options (alpha)". Scroll down to "Mode". Change the value "Regular" to "Regula". Immediately change the value back to "Regular".
5. In the separate browser, refresh the page. The page will now load correctly.

Note: Changing `WebMaster(Options())` to `WebMaster(Options(mode=["regular"]))` does not resolve the issue.

### System Information

```raw
Mitmproxy: 12.2.1
Python:    3.14.2
OpenSSL:   OpenSSL 3.5.5 27 Jan 2026
Platform:  Windows-11-10.0.26200-SP0
```

### Checklist

NETIZEN-11:
# Description

This PR fixes an issue where launching mitmweb programmatically using `WebMaster(Options())` does not start the HTTP/S proxy, even when the mode is set to "regular".

# Problem

When mitmweb is started via Python code, the proxy server remains inactive by default. The proxy only starts after manually toggling the mode in the UI, which indicates that the required initialization logic is not executed during programmatic startup.

# Root Cause

The proxy server initialization is not triggered when WebMaster is launched programmatically, even if a valid proxy mode is already configured.

# Fix

- Ensured that proxy server initialization is triggered during WebMaster startup
- Reused the same logic that is executed when the mode is changed via the UI
- Enabled automatic proxy activation when a valid mode (e.g., "regular") is set

# Result

- HTTP/S proxy starts automatically on programmatic launch
- No manual UI interaction required
- Consistent behavior between UI-based and programmatic usage

# Testing

- Ran the provided Python reproduction script
- Configured browser proxy to `localhost:8080`
- Verified that requests are successfully proxied without toggling settings in the UI

Fixes #8146

### DIFF
--- a/mitmproxy/addons/proxyserver.py
+++ b/mitmproxy/addons/proxyserver.py
@@ -237,16 +237,6 @@ class Proxyserver(ServerManager):
 
     def running(self):
         self.is_running = True
-        # If servers haven't been set up yet (e.g., when running() is called directly
-        # without going through master.run()), set them up now.
-        # This ensures programmatic usage works even if setup_servers() wasn't called.
-        if not self._servers_initialized and ctx.options.mode and ctx.options.server:
-            logger.info("Servers not initialized, setting up now...")
-            asyncio_utils.create_task(
-                self.setup_servers(),
-                name="setup_servers_from_running",
-                keep_ref=True,
-            )
 
     def configure(self, updated) -> None:
         if "stream_large_bodies" in updated:
--- a/mitmproxy/addons/proxyserver.py
+++ b/mitmproxy/addons/proxyserver.py
@@ -64,11 +64,15 @@ class Servers:
                 for spec in modes:
                     if spec in self._instances:
                         instance = self._instances[spec]
-                        logger.debug(f"Keeping existing server instance for mode: {spec.full_spec}")
+                        logger.debug(
+                            f"Keeping existing server instance for mode: {spec.full_spec}"
+                        )
                     else:
                         instance = ServerInstance.make(spec, self._manager)
                         start_tasks.append(instance.start())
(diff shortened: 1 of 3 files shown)


fwdekker-pc:
I don't think this is the correct solution for the issue I opened. The original issue was because I used the MitmProxy API wrongly. I have written what I think is the correct usage in [a comment on the issue](https://github.com/mitmproxy/mitmproxy/issues/8146#issuecomment-4476396069). I have closed the issue, and @NETIZEN-11 I think you should also close this PR.

PR #8167, opened 2026-04-07
correct a parameter and strengthen cleanup

nameearly:
- [√] I have updated tests where applicable.
 - [√] I have added an entry to the CHANGELOG.

### DIFF
--- a/mitmproxy/proxy/server.py
+++ b/mitmproxy/proxy/server.py
@@ -322,8 +322,7 @@ class ConnectionHandler(metaclass=abc.ABCMeta):
             writer.close()
         except OSError:
             pass
-        finally:
-            self.transports.pop(connection, None)
+        self.transports.pop(connection)
 
         if cancelled:
             raise cancelled
--- a/mitmproxy/proxy/server.py
+++ b/mitmproxy/proxy/server.py
@@ -322,7 +322,8 @@ class ConnectionHandler(metaclass=abc.ABCMeta):
             writer.close()
         except OSError:
             pass
-        self.transports.pop(connection)
+        finally:
+            self.transports.pop(connection, None)
 
         if cancelled:
             raise cancelled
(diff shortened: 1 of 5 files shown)


lups2000 (commented):
Thanks for the PR :) I think it currently mixes heterogeneous changes. Ideally, each PR should focus on a single feature or bug fix.

nameearly:
> Thanks for the PR :) I think it currently mixes heterogeneous changes. Ideally, each PR should focus on a single feature or bug fix.

"Strengthen cleanup" is not very needed to be a inditional pr and they are both simple, so I mix them.

mitmproxy/proxy/server.py
```diff
@@ -322,7 +322,8 @@ async def handle_connection(self, connection: Connection) -> None:
             writer.close()
         except OSError:
             pass
-        self.transports.pop(connection)
+        finally:
+            self.transports.pop(connection, None)
```
mhils: Are there cases where something else than an OSError is raised? How did you come up with this change?

nameearly: Nothing else, just did a static analysis.
Thanks for being willing to take the time to talk with me.

mhils: Thanks. In this case please remove this change and the one from `mitmproxy/proxy/layers/tls.py`. 

- `finally` gives a wrong impression here that we expect other types of errors. If anything there would need to be a big try: finally: block around most of the method, but that's too much.
- Similary, the API contract of `self.event_to_child` is to never raise.

mhils:
> "Strengthen cleanup" is not very needed to be a inditional pr and they are both simple, so I mix them.

+1 to what @lups2000 said - separate small PRs for separate changes are better!

nameearly:
I have removed those changes.

lups2000 (changes requested):
I think you forgot to merge main and solve the conflicts :)

nameearly:
> I think you forgot to merge main and solve the conflicts :)

To solve the conflict, I opended another. https://github.com/mitmproxy/mitmproxy/pull/8172
I would close this one.

PR #8171, opened 2026-04-10
SECURITY.md: Exclude DoS from scope

mhils:
See diff - what do you folks think?

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -12,7 +12,8 @@
 
  - **Denial of Service (DoS):** We consider DoS vectors to be ordinary bugs and not security vulnerabilities.
    You may post them openly on the issue tracker. We will not issue any advisories or CVEs for them.
-   The reasoning here is that mitmproxy is a single-threaded Python application. It can typically be overwhelmed by sending many requests anyways.
+   The reasoning here is that mitmproxy is an interactive inspection tool, not a web server for high volume traffic.
+   It can typically be overwhelmed by sending too many requests; any DoS is just a variation of this.
  - **All other vulnerabilities:** Please report them using the process below.
 
 ## Reporting a Vulnerability
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -8,6 +8,13 @@
 | most recent release | :white_check_mark: |
 | older releases      | :x:                |
 
+## Scope
+
+ - **Denial of Service (DoS):** We consider DoS vectors to be ordinary bugs and not security vulnerabilities.
+   You may post them openly on the issue tracker. We will not issue any advisories or CVEs for them.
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -8,6 +8,13 @@
 | most recent release | :white_check_mark: |
 | older releases      | :x:                |
 
+## Scope
+
+ - **Denial of Service (DoS):** We consider DoS vectors to be ordinary bugs and not security vulnerabilities.
```
lups2000, line 13: Should we maybe be more specific here and include some examples?

lups2000 (approved):
Makes sense to me :)

SECURITY.md
mhils: Examples such as? :)

lups2000: I’m not sure 😅  for example, are we including both simple DoS attacks (ex. high request volume overwhelming the application) as well as DoS issues that may expose unintended behaviour (ex. crashes or memory corruption)?

SECURITY.md
```diff
@@ -8,6 +8,13 @@
...
+   You may post them openly on the issue tracker. We will not issue any advisories or CVEs for them.
+   The reasoning here is that mitmproxy is a single-threaded Python application. It can typically be overwhelmed by sending many requests anyways.
```
emanuele-em:
```suggestion
   The reasoning here is that mitmproxy is a single-threaded Python application designed for traffic inspection and analysis, not for production use. It can typically be overwhelmed by sending many requests anyways.
```

I'd add this

Kriechi: I feel like the "Python single-threaded app" is an excuse, not a reason.

The reason IMO is derived from our GitHub "About" tag line: 
> An interactive TLS-capable intercepting HTTP proxy for penetration testers and software developers.

mitmproxy is **not** intended to serve as reverse proxy or deep-packet inspection layer in front of public facing Internet services or networks. By that deduction, DoS is not a relevant class of security issues, because mitmproxy is not intended to operate in such an environment, where uptime or availability of a service is of concern to the operator (SRE mentality trying to achieve an SLO, or offering any form of SLA).

SECURITY.md
mhils: Memory Corruption is not a DoS anymore. :-) Crashes fall into the same category IMO. Not sensitive, but something we probably want to prioritize fixing.

SECURITY.md
mhils: Fair point. I've revised the wording and made it more general, PTAL. :-)

lups2000 (approved):

errorxyz (approved):

Prinzhorn (approved):

Kriechi (approved):

PR #8177, opened 2026-04-12
accept leading zeros in Content-Length behind validate_inbound_headers=false

### LINKED ISSUE #7786: Unable to parse HTTP response: invalid content-lenght header : '0000344'
### Problem Description

We are calling a soap service that ("wrongly") returns a content-length header with leading 0 (zeroes) and we get an error and the flow stops.

Our colleagues use Fiddler and it looks like they strip the zeros ;-)

### System Information

```raw

```

### Checklist

emanuele-em:
#### Description

- Accept leading zeros in `Content-Length` headers (e.g. `0012`) when `validate_inbound_headers` is disabled
- Thread `validate_inbound_headers` through `parse_content_length()` and `expected_http_body_size()`
- Strict validation (default) is unchanged — leading zeros are still rejected to prevent "request smuggling" (closes #7786)

#### Checklist

### DIFF
--- a/mitmproxy/net/http/http1/read.py
+++ b/mitmproxy/net/http/http1/read.py
@@ -45,7 +45,9 @@ def connection_close(http_version, headers):
 
 
 def expected_http_body_size(
-    request: Request, response: Response | None = None
+    request: Request,
+    response: Response | None = None,
+    validate_inbound_headers: bool = True,
 ) -> int | None:
     """
     Returns:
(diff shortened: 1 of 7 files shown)


mitmproxy/net/http/http1/read.py
```diff
@@ -45,7 +45,9 @@ def connection_close(http_version, headers):
 
 
 def expected_http_body_size(
-    request: Request, response: Response | None = None
+    request: Request,
+    response: Response | None = None,
+    validate_inbound_headers: bool = True,
```
mhils, line 50: I think conceptually it doesn't make sense for `expected_http_body_size` to take a `validate_inbound_headers` argument. What even is inbound? The request? The response?

Maybe it should be something like `insecure_allow_malformed`?

Issue #8188, opened 2026-04-21
Index out of range error when converting to .har file


T-aut:
### Problem Description

When converting .mitm flow data to a .har file, an exception is thrown in the `mitmproxy/utils/strutils.py` source file for specific response payloads.

#### Steps to reproduce the behavior:
1. Download attached minimal reproduction `.mitm` flow
2. Run `mitmdump -n -r scrubbed.mitm --set hardump=./test.har`
3. Observe output

Minimal repro file:
[scrubbed.mitm.zip](https://github.com/user-attachments/files/26930631/scrubbed.mitm.zip)

Stacktrace:
```
[13:00:52.694] Addon error: index out of range
Traceback (most recent call last):
  File "mitmproxy/addons/savehar.py", line 128, in done
  File "mitmproxy/command.py", line 322, in wrapper
  File "mitmproxy/addons/savehar.py", line 39, in export_har
  File "mitmproxy/addons/savehar.py", line 58, in make_har
  File "mitmproxy/addons/savehar.py", line 205, in flow_entry
  File "mitmproxy/utils/strutils.py", line 136, in is_mostly_bin
IndexError: index out of range
```

Issue is also present in `mitmweb` when using the command line and invoking `save.har`, since it uses the same `/strutils.py` source.

I believe the issue may have been introduced with https://github.com/mitmproxy/mitmproxy/pull/7933

### System Information

```raw
Mitmproxy: 12.2.2 binary
Python:    3.14.3
OpenSSL:   OpenSSL 3.5.5 27 Jan 2026
Platform:  macOS-15.7.4-arm64-arm-64bit-Mach-O
```

### Checklist

- [x] This bug affects the [latest mitmproxy release](https://github.com/mitmproxy/mitmproxy/releases).

PR #8189, opened 2026-04-21
Fix issue 8188: index out of range when saving .har file

T-aut:
#### Description

Proposal for fix regarding an issue I've encountered https://github.com/mitmproxy/mitmproxy/issues/8188

After examining the source it seems that it's just an unchecked direct access to an array of bytes. We assure we're>= 100 bytes, but the lines after attempt to access bytes up to index 104.

#### Checklist


mhils:
Thanks a bunch! I missed this older PR before merging #8196, apologies. In either case, this is fixed now. 😃

PR #8196, opened 2026-04-25
fix: avoid IndexError in is_mostly_bin for short tails

### LINKED ISSUE #8188: Index out of range error when converting to .har file

juliosuas:
fixes #8188.

reproducible with any bytes payload of len 101–103 where index 100 is a UTF-8 continuation byte:

```python
>>> from mitmproxy.utils import strutils
>>> strutils.is_mostly_bin(b'a' * 100 + b'\x80')
  ...
IndexError: index out of range
```

the lookahead loop `range(100, 104)` reads `s[100]` through `s[103]` to find the start of the next UTF-8 char and avoid chopping a multibyte code point in half. when len(s) is between 101 and 103 and the first byte past the 100-byte cutoff is a continuation byte, the loop continues into out-of-range indices.

cap the loop end at `len(s)`. the existing `s[:100]` fallback still covers the case where every byte in the lookahead window is a continuation byte.

added a small regression assertion for the three affected lengths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 ## Unreleased: mitmproxy next
 
+- Fix `IndexError` in `is_mostly_bin` when exporting flows to HAR with payloads
+  that have a UTF-8 continuation byte at the 100-byte cutoff.
 
 ## 12 April 2026: mitmproxy 12.2.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 - Fix `IndexError` in `is_mostly_bin` when exporting flows to HAR with payloads
   that have a UTF-8 continuation byte at the 100-byte cutoff.
+  ([#8196](https://github.com/mitmproxy/mitmproxy/pull/8196), @juliosuas)
 
 ## 12 April 2026: mitmproxy 12.2.2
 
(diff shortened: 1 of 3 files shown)


mhils (approved):
Thanks!

juliosuas:
🫡

PR #8197, opened 2026-04-28
Fix IP blocking and merge additional DoH blocklists

Olexandr88:
Fixed broken IP-based DoH blocking by extracting the host string from the server_conn.address tuple, and ensured additional_doh_names and additional_doh_ips are actually merged into the lookup sets at runtime.


mhils (approved):
Thanks!

PR #8202, opened 2026-05-01
Fix query content view for missing request bodies

### LINKED ISSUE #7941: Mitmproxy/web do not allow rendering other content views while content is missing.
### Problem Description

Mitmproxy and mitmweb cannot render other content views while the request body is still being received. The check for other content views is preceded by a check for missing content, which causes other views such as the query view (which can be rendered much before receiving the entire body) to render to "[content missing]".

1. Run mitmproxy.
2. Run this bash script:
   ```bash
   (
     printf 'GET /?a=b&c=d HTTP/1.1\r\n'
     printf 'Host: www.example.com\r\n'
     printf 'Content-Length: 1\r\n\r\n'
     sleep 100 # mitmproxy keeps waiting for the expected 1 byte of data for this time period.
   ) | nc 127.0.0.1 8080 # mitmproxy running on 8080
   ```
3. Open the flow generated from the above request and try manually switching to query view (press m), which will still display "[content missing]".

(Refs #7923).

### System Information

```raw
Mitmproxy: 12.2.0
Python:    3.14.0
Platform:  Linux-6.17.4-300.fc43.x86_64-x86_64-with-glibc2.42
```

### Checklist

emanuele-em:
#### Description

Fixes #7941.

This allows the Query content view to render request query parameters even while the request body is still missing or being received. Previously, missing request content short-circuited content view rendering before the Query view had a chance to render data that is already available from the request URL.

I am not super happy with having such a specific hard-coded exception for the Query view, but I do not have a better idea here.

The change covers both mitmproxy console and mitmweb, and keeps other explicit body views such as Raw showing `[content missing]` when body content is unavailable.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## Unreleased: mitmproxy next
 
+- Render the Query content view for requests whose body content is still being received.
 - Fix `IndexError` in `is_mostly_bin` when exporting flows to HAR with payloads
   that have a UTF-8 continuation byte at the 100-byte cutoff.
   ([#8196](https://github.com/mitmproxy/mitmproxy/pull/8196), @juliosuas)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ## Unreleased: mitmproxy next
 
 - Render the Query content view for requests whose body content is still being received.
+  ([#8202](https://github.com/mitmproxy/mitmproxy/pull/8202), @emanuele-em)
 - Fix `IndexError` in `is_mostly_bin` when exporting flows to HAR with payloads
   that have a UTF-8 continuation byte at the 100-byte cutoff.
   ([#8196](https://github.com/mitmproxy/mitmproxy/pull/8196), @juliosuas)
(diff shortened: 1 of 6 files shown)


lups2000 (commented):
Yes, this seems to solve the issue :) I agree that it looks a bit hardcoded, but I don't see many alternatives here.
Maybe we can improve the readability a bit by adding a few more comments and extracting some code (or maybe finding a balance between the two):

```
def _missing_content() -> ContentviewResult:
    return ContentviewResult(
        text="Content is missing.",
        syntax_highlight="error",
        description="",
        view_name=None,
    )


if data is None:
    if not isinstance(message, http.Request):
        return _missing_content()

    # The query view only depends on request metadata, not body content.
    metadata = Metadata(flow=flow, http_message=message)
    view = registry.get_view(b"", metadata, view_name)

    if view.name != query.name:
        return _missing_content()

    data = b""
else:
    metadata = make_metadata(message, flow)

```

mhils:
Hmm. I'd rather remove the query view entirely from `mitmproxy.contentviews` than to pile more workarounds on top of it. How are other tools handling query view display? 🤔

PR #8213, opened 2026-05-04
view: warn when view_order_reversed is set on mitmweb (#5520)

### LINKED ISSUE #5520: Can't reverse the order of captured requests even with view_order_reversed set to true
#### Problem Description
I'm using mitmweb and new requests are put at the end of the table so I have to scroll to see new ones. So I searched online and found view_order_reversed option which in mitmproxy docs website states that this applies to both `mitmproxy` and `mitmweb`. I enabled this in 3 different ways using config.yaml, directly from mitmweb and through using --set command and nothing works. The viewing order is not reversed. The newest requests are always at the end of the table and I have to scroll for them.

Is there anyway I can view newest requests first? if not, how about autoscrolling to the last request?

1. start mitmweb with this command: `mitmweb --set view_order_reversed`
2. go to the opened browser tab and wait for new requests.
3. requests are still being displayed oldest to newest, so newest requests are still hidden and require scroll.

#### System Information
Mitmproxy: 6.0.2
Python:    3.10.5
OpenSSL:   OpenSSL 3.0.5 5 Jul 2022
Platform:  Linux-5.10.0-13-amd64-x86_64

ChrisJr404:
## Summary

Closes #5520.

The \`view_order_reversed\` option is documented as applying to all of mitmproxy, mitmweb, and mitmdump, but mitmweb's React flow table sorts by the clicked column and ignores the option entirely. The reporter on #5520 ran \`mitmweb --set view_order_reversed\`, expected newest-first, and got nothing back. @Prinzhorn diagnosed the same docs/runtime mismatch in the thread:

> Maybe the \"bug\" is in the docs listing mitmweb, because I don't see any code handling this in mitmweb. ... right now we are in an in-between state, where \`view_order_reversed\` does somewhat apply to mitmweb.

This PR addresses the mismatch without changing the option's existing console behaviour:

1. **Option description** now states honored-only-by-the-console-TUI and links the issue, so readers of \`mitmweb --options\` / docs see the limit before they pass the flag.
2. **Runtime warning** when the option is set to True on a mitmweb master. Setting it on the console master, or to False on mitmweb, is silent.

## Why a warning, not a hard error

Per @Prinzhorn (\"in-between state\"), the option still partially affects mitmweb (the underlying \`_view\` ordering changes) but the user-visible UI doesn't reflect it. A hard error would punish people whose existing config files have \`view_order_reversed: true\` carried over from mitmproxy and now run mitmweb. A one-time warning at configure time is enough to keep the user from being silently fooled, and is safe to pull out later if mitmweb ever grows a timeline-sort affordance (per @Prinzhorn's followup remark).

## Detection

I needed to recognise mitmweb's \`WebMaster\` from inside the \`view\` addon. A direct import would create a cycle (the addon is loaded by every Master subclass, including \`WebMaster\`). The helper \`_master_is_mitmweb\` walks \`master.__class__.__module__\` instead — a class whose module path begins with \`mitmproxy.tools.web\` is mitmweb. The fallback name check (\`__name__ == \"WebMaster\"\`) keeps the test fixtures simple.

## Tests

\`test/mitmproxy/addons/test_view.py\`:

- \`test_master_is_mitmweb\` — helper recognises a class whose module path matches \`mitmproxy.tools.web.*\` and rejects an unrelated console-style master.
- \`test_view_order_reversed_warns_in_mitmweb\` — caplog asserts a WARNING line containing both \"mitmweb\" and \"view_order_reversed\" when the option flips True on a fake \`WebMaster\`.
- \`test_view_order_reversed_silent_in_console\` — same option flip on a console master, no warning, and \`view.order_reversed\` is updated as before.

```
$ pytest test/mitmproxy/addons/test_view.py
35 passed in 0.49s
```

(3 new + 32 existing.)

## Notes for reviewers

- I deliberately kept \`set_reversed\` behaviour in mitmweb mode so the underlying \`_view\` order still flips — that's what the existing flow has done for years and what Prinzhorn called the \"in-between state.\" This PR just makes that state honest in the help text and audible in the log.
- If you'd rather make the option console-only at the option level (i.e. reject the value on a mitmweb master), that's a single \`raise exceptions.OptionsError(...)\` in \`configure\` instead of the warning. Happy to flip on review.

### DIFF
--- a/mitmproxy/addons/view.py
+++ b/mitmproxy/addons/view.py
@@ -56,7 +56,9 @@ def _master_is_mitmweb(master: Any) -> bool:
     enough to recognise the mitmweb tools tree without re-importing it.
     """
     cls = master.__class__
-    return cls.__module__.startswith("mitmproxy.tools.web") or cls.__name__ == "WebMaster"
+    return (
+        cls.__module__.startswith("mitmproxy.tools.web") or cls.__name__ == "WebMaster"
+    )
 
 
 class _OrderKey:
--- a/mitmproxy/addons/view.py
+++ b/mitmproxy/addons/view.py
@@ -47,6 +47,18 @@ from mitmproxy.utils import signals
 # when they are updated.
 
 
+def _master_is_mitmweb(master: Any) -> bool:
+    """Return True when the running master is mitmweb's WebMaster.
+
+    We can't import mitmproxy.tools.web.master directly here because the view
+    addon is loaded by every Master subclass (including mitmweb's), and a
+    direct import would create a cycle. Inspecting the class' module path is
+    enough to recognise the mitmweb tools tree without re-importing it.
+    """
+    cls = master.__class__
+    return cls.__module__.startswith("mitmproxy.tools.web") or cls.__name__ == "WebMaster"
+
+
 class _OrderKey:
     def __init__(self, view):
         self.view = view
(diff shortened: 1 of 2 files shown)


mhils:
I think the correct fix is to make the option work, not to add a cranky warning for it.

ChrisJr404:
Makes sense. The hop from "addon warns" to "mitmweb actually honors the option" crosses into the frontend table-sort code, which is invasive enough that the warning approach was the wrong shape. Closing this one. I'll take a look at the frontend side and open a fresh PR if I can land a clean fix.

Issue #7703, opened 2025-05-10
authority and subject key identifier mismatch


wtlwang2019:
#### Steps to reproduce the behavior:
1. ubuntu 22.04 
2. pip install mitmproxy 
3. system trust ca-cert.pem
4. curl --proxy https://127.0.0.1:8080 https://www.baidu.com
5. proxy failed: certificate error, authority and subject key identifier mismatch 

### System Information

```raw

```

### Checklist

- [ ] This bug affects the [latest mitmproxy release](https://github.com/mitmproxy/mitmproxy/releases).

sujaldev:
Likely not a mitmproxy issue, can you connect to www.baidu.com without mitmproxy?

wtlwang2019:
@sujaldev yes, i can connect to www.baidu.com without proxy

sujaldev:
I cannot reproduce this, are you able to connect to any other https website via mitmproxy? If not, ensure that you've followed the certificate installation instructions on [mitm.it](http://mitm.it/) under linux > ubuntu/debian.

wtlwang2019:
thx, I reconfigure it，then it works.
`
sudo mkdir /usr/local/share/ca-certificates/extra
Copy the CA .crt file to this directory:

sudo cp foo.crt /usr/local/share/ca-certificates/extra/foo.crt
(For Ubuntu 20.04 or higher the path is /usr/share/ca-certificates/extra/foo.crt)

Let Ubuntu add the .crt file's path relative to /usr/local/share/ca-certificates to /etc/ca-certificates.conf:

sudo dpkg-reconfigure ca-certificates
To do this non-interactively, run:

sudo update-ca-certificates
`

PR #8214, opened 2026-05-04
Copy issuer's SubjectKeyIdentifier into leaf AuthorityKeyIdentifier (RFC 5280 §4.2.1.2)

unique-jakub:
#### Description

`mitmproxy/certs.py::dummy_cert` builds the leaf certificate's `AuthorityKeyIdentifier` with:

```python
x509.AuthorityKeyIdentifier.from_issuer_public_key(cacert.public_key())
```

`cryptography`'s `from_issuer_public_key` always derives the keyIdentifier as the **SHA-1** of the issuer's BIT STRING public key. It ignores whatever `SubjectKeyIdentifier` the issuer cert actually has stored. This violates RFC 5280 §4.2.1.2:

> The value of the subject key identifier MUST be the value placed in the key identifier field of the authority key identifier extension of certificates issued by the subject of this certificate.

It also makes mitmproxy incompatible with any CA whose stored SKI was not derived as SHA-1 — an entirely legal choice per RFC 7093 (truncated SHA-256/384/512) and RFC 5280 (any unique value).

Strict TLS chain builders (`X509_V_FLAG_X509_STRICT`, Python `ssl`, Go `crypto/x509`, modern Java) all use AKI to find the issuer candidate during chain construction, so an AKI/SKI mismatch is rejected with `authority and subject key identifier mismatch` — even though the issuer's subject DN matches and the signature verifies.

##### Why this is hitting more users now

- **cert-manager ≥ 1.18** (Oct 2025) defaults to truncated SHA-256 SKIs for FIPS 140-3 compliance. Any mitmproxy deployment whose CA was provisioned by cert-manager (very common in Kubernetes egress / corporate intercept setups) breaks.
- **Go ≥ 1.25** (Aug 2025) does the same when generating self-signed CAs (`GODEBUG=x509sha256skid=1`, on by default). Tools that wrap Go's `crypto/x509` to produce CAs (cert-manager being one) inherit this.
- Hand-rolled CAs with `subjectKeyIdentifier=hash:sha256` in the OpenSSL config trigger it too.

The closest existing report is #7703, which was closed without investigation after the OP misdiagnosed it as a CA-trust-store problem.

##### Reproduction

```bash
# 1) Make a CA with a non-SHA-1 SKI (mimics cert-manager output).
cat > /tmp/ca.cnf <<'CNF'
[req]
distinguished_name = req
x509_extensions    = v3_ca
prompt             = no
CN                 = Test CA
[v3_ca]
basicConstraints       = critical,CA:TRUE
keyUsage               = critical,keyCertSign,cRLSign
subjectKeyIdentifier   = hash:sha256
CNF
openssl req -x509 -newkey rsa:2048 -nodes -days 3650 \
  -keyout /tmp/ca.key -out /tmp/ca.crt -config /tmp/ca.cnf

# 2) Hand it to mitmproxy as the CA.
cat /tmp/ca.crt /tmp/ca.key > ~/.mitmproxy/mitmproxy-ca.pem
mitmdump -p 8080 &

# 3) Trust /tmp/ca.crt system-wide, then:
curl --proxy http://127.0.0.1:8080 https://example.com
# curl: (60) SSL certificate problem: authority and subject key identifier mismatch
```

##### Fix

Prefer the issuer's stored SKI verbatim, falling back to the previous SHA-1-of-public-key behavior only when the issuer has no `SubjectKeyIdentifier` extension at all. The change is fully backward-compatible: when the issuer cert was generated by mitmproxy itself (or any other tool that does in fact use SHA-1 of the public key for the SKI), `from_issuer_subject_key_identifier(stored_ski)` and `from_issuer_public_key(public_key)` produce byte-identical results.

##### Tests

Added two regression tests in `test/mitmproxy/test_certs.py::TestDummyCert`:

- `test_aki_copies_issuer_ski_non_sha1` — builds a CA whose stored SKI is a truncated SHA-256 (as cert-manager / Go 1.25+ would produce) and asserts the leaf's AKI is byte-equal to the CA's stored SKI.
- `test_aki_falls_back_when_issuer_has_no_ski` — pins the legacy fallback path: if the CA has no SKI extension, the leaf's AKI is still computed from the issuer's public key and matches the prior behavior.

All existing `test_certs.py` tests pass (36 total). I also re-ran `test/mitmproxy/proxy/layers/test_tls.py` to make sure nothing downstream of `dummy_cert` was affected (62 tests total, all pass). `ruff check`, `ruff format --check`, and `mypy mitmproxy/certs.py` are clean.

##### Verification with real clients

Outside the unit tests I also reproduced the bug end-to-end on a Kubernetes minikube cluster with a cert-manager-issued CA, then re-ran the same tests with a `mitmweb` instance whose `dummy_cert` had been runtime-patched with this same change. Strict TLS chain validation through Python `requests`, Go `net/http`, and `curl --cacert` all flipped from `authority and subject key identifier mismatch` to a successful intercept, with no other code changes.

#### Checklist


mhils (approved):
Thank you for the excellent PR!

PR #8216, opened 2026-05-06
feat(mitmweb): Add Chinese (zh) localization support

HULANG-BTB:
#### Description

mitmweb: Add Chinese (zh) localization support. English remains the default language.
  A language selector is available in the header bar.

#### Checklist

 - [ √] I have updated tests where applicable.
 - [√ ] I have added an entry to the CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 ## Unreleased: mitmproxy next
 
+- mitmweb: Add Chinese (zh) localization support. English remains the default language.
+  A language selector is available in the header bar.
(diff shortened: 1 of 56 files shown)


lups2000 (commented):
Thanks for the PR! I'm not sure if we want to add localization support though. I couldn't find any previous discussion about this, but I guess it has come up at least once in the past. Maybe @mhils or @Prinzhorn can provide more reasons here.

mhils (commented):
Thank you for the PR! As much as I'd like to welcome more non-English users, we don't have the capacity to maintain translations at this point. The in-browser translation tools work fairly well in my experience and should hopefully be a good workaround.

Issue #8194, opened 2026-04-24
X-XSRFToken: undefined, unable to view data in mitmweb


prostraction:
#### Steps to reproduce the behavior:
1. start mitmweb --web-port 8082
2. open UI
3. toggle any checkbox in Options

Expected: PUT /options succeeds
Actual: 403, request header "X-XSRFToken: undefined" while cookie "_xsrf" is set correctly

```
mitmweb --web-port 8082
[09:05:20.787] HTTP(S) proxy listening at *:8080.
[09:05:20.790] Web server listening at http://127.0.0.1:8082/?token=70ff08a4fab4640d07bb2825cc40287b
[09:06:06.166] 403 PUT /options (127.0.0.1): XSRF cookie does not match POST argument
[09:06:06.166] 403 PUT /options (127.0.0.1) 0.00ms
```

<img width="1873" height="1029" alt="Image" src="https://github.com/user-attachments/assets/d9aae9e6-68d7-448d-9bf9-6df52e62fd51" />

<img width="953" height="329" alt="Image" src="https://github.com/user-attachments/assets/103f09a6-279f-4196-a541-9af503640bd2" />

### System Information

```raw
mitmproxy --version
Mitmproxy: 12.2.2
Python:    3.12.5
OpenSSL:   OpenSSL 3.4.1 11 Feb 2025
Platform:  Windows-11-10.0.22631-SP0
```

### Checklist

- [x] This bug affects the [latest mitmproxy release](https://github.com/mitmproxy/mitmproxy/releases).

Prinzhorn:
Can confirm _in Firefox_ not in Chromium. Broke in 12.2.1, works with 12.2.0.

However, I could not reproduce this in a private window. Turns out I had a `_xsrf` cookie with the `HttpOnly` flag, which the client could not read to double submit as header. Removing that cookie fixed the issue.

This commit broke it @mhils , we previously read the value from a data attribute https://github.com/mitmproxy/mitmproxy/commit/26074879818ed547d3ca44b02880fb78ee62bbb3#diff-df88fc1c7e495af8f18824d3f2d56d72d0f7a1865d1d231ea041810102c67ec3

So everyone who used mitmweb before and has the cookie is in a broken state.

Can you confirm that a private window works for you as well?

lups2000:
Yes @Prinzhorn, I think there was a similar problem in #8000; removing the token and refreshing seems the solution here.

Prinzhorn:
[...]

PR #8217, opened 2026-05-08
Fix mitmweb XSRF token header

juliosuas:
## Summary

Fixes the mitmweb XSRF header regression from #8194.

The frontend was reading `_xsrf` exclusively from `document.cookie`. If a browser had an older `_xsrf` cookie marked `HttpOnly`, mitmweb could not read it, so non-GET API requests were sent with `X-XSRFToken: undefined` and Tornado rejected them with 403.

This restores the previous behavior of rendering the current XSRF token into the page and having the frontend prefer that value. The cookie fallback remains in place, and the Vite dev-server placeholder is ignored so local frontend development does not accidentally send `{{ xsrf_token }}` as a token.

## Testing

- `npm test -- --runTestsByPath src/js/__tests__/utilsSpec.tsx`
- `npm run ci-build-release`
- `git diff --check`

I also tried `python3 -m pytest test/mitmproxy/tools/web/test_app.py -q`, but the local environment has an incompatible pytest/config combination and exits while parsing `pyproject.toml` before collecting tests.


mhils:
Thanks! Superseded by https://github.com/mitmproxy/mitmproxy/pull/8224, which is a bit simpler.

PR #8219, opened 2026-05-09
Bump upper bounds for cryptography and pyOpenSSL

reaperhulk:
Updates the cryptography and pyopenssl bounds so we can move forward with removal of some functionality in cryptography. https://github.com/pyca/cryptography/pull/14798


mhils (approved):
Thanks!

PR #8221, opened 2026-05-10
Fix multipart parser stripping \n and \r bytes from binary content

### LINKED ISSUE #4466: Error parsing multipart with newline in the content

quick123-666:
## Description

Closes #4466

### Problem

When sending multipart requests containing `\n` or `\r` bytes in the content (e.g., binary files or text with newlines), `decode_multipart()` strips those bytes because it uses `bytes.splitlines()` which removes line-ending characters.

For example, `curl -x 'http://localhost:9000' -F $'data=a\nb' http://some-site` results in `v == b"ab"` instead of expected `b"a\nb"`.

### Root Cause

In `mitmproxy/net/http/multipart.py`, the `decode_multipart()` function does:
```python
parts = i.splitlines()
...
value = b"".join(parts[3 + parts[2:].index(b"") :])
```

`splitlines()` splits on `\n`, `\r\n`, etc. and **removes** those characters. When the body content is reconstructed via `b"".join()` on the split segments, all newline/CR bytes in the body are lost.

### Fix

Instead of using `splitlines()` which destroys the body content, the fix:

1. Strips only the leading newline(s) that prefix each boundary-delimited part
2. Finds the header/body separator (first `\r\n\r\n` or `\n\n`)
3. Preserves the body content as-is, with all internal newline characters intact
4. Strips only the trailing `\r\n` or `\n` that belongs to the boundary delimiter

### Testing

Existing tests continue to pass. Added a new test case that specifically verifies binary content with `\n` and `\r\n` bytes is preserved.

### DIFF
--- a/mitmproxy/net/http/multipart.py
+++ b/mitmproxy/net/http/multipart.py
@@ -82,12 +82,12 @@ def decode_multipart(
                 sep_idx = part.find(b"\r\n\r\n")
                 if sep_idx != -1:
                     header_section = part[:sep_idx]
-                    body = part[sep_idx + 4:]
+                    body = part[sep_idx + 4 :]
                 else:
                     sep_idx = part.find(b"\n\n")
                     if sep_idx != -1:
                         header_section = part[:sep_idx]
-                        body = part[sep_idx + 2:]
+                        body = part[sep_idx + 2 :]
                     else:
                         continue
 
--- a/mitmproxy/net/http/multipart.py
+++ b/mitmproxy/net/http/multipart.py
@@ -50,6 +50,9 @@ def decode_multipart(
 ) -> list[tuple[bytes, bytes]]:
     """
     Takes a multipart boundary encoded string and returns list of (key, value) tuples.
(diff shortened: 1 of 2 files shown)


quick123-666:
CI revealed two issues, both fixed:

1. **`test_decode_preserves_newlines`** - str+bytes concatenation error in test case 3 (`f"\r\n" + b"..."`). Fixed to use `b"\r\n"`.

2. **`test_set_multipart_form`** - `encode_multipart` adds an empty `hdrs.append(b"")` after each value, which produces an extra `\r\n` in the encoded output. The old `splitlines()` ate these, but the new parser only stripped one `\r\n`. Fixed by using a `while` loop to strip all trailing `\r\n` sequences.

PR #8231, opened 2026-05-14
Preserve buffered request/response body on protocol error

Halle-Astra:
#### Description

  When a connection is interrupted mid-stream, the body data already
  buffered in request_body_buf / response_body_buf was discarded,
  causing the Web UI to show 'Content is missing.' instead of the
  partial content that was received.

  This fix saves any buffered data to flow.request.data.content /
  flow.response.data.content in handle_protocol_error before the
  error hook is triggered, so users can still inspect what was
  received before the connection broke.

#### Checklist

### DIFF
--- a/mitmproxy/proxy/layers/http/__init__.py
+++ b/mitmproxy/proxy/layers/http/__init__.py
@@ -733,6 +733,15 @@ class HttpStream(layer.Layer):
             or self.server_state in (self.state_done, self.state_errored)
         )
 
+        # Save any buffered request/response body data before we lose it,
+        # so users can see what was received before the connection broke.
+        if self.request_body_buf and self.flow and self.flow.request:
+            if self.flow.request.raw_content is None:
+                self.flow.request.data.content = bytes(self.request_body_buf)
+        if self.response_body_buf and self.flow and self.flow.response:
+            if self.flow.response.raw_content is None:
+                self.flow.response.data.content = bytes(self.response_body_buf)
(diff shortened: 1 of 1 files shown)


mhils:
I'm not sure if I like this change. `None` clearly signals that we don't have the full body. Now we have a partial body but no good signalling that it's just partial.

Halle-Astra:
> I'm not sure if I like this change. `None` clearly signals that we don't have the full body. Now we have a partial body but no good signalling that it's just partial.

In my practice, I need the buffered stream to analyze when my LLM agent disconnected request (shown in mitmweb as 'client disconnected').  Such as, something like timeout when agent received partial stream but think it's too long to receive full stream content and raise the timeout Error.
In this scence, I hope to know what content LLM wants to produce and makes it so long. 

But current behaviour is discarding the buffered stream content. So I cannot analyze it even with a hook script method.

Also, I know in most practice, the response of failed request is None or meaningless(maybe partial stream but nobody care). 

current performance, though this post is disconnected by client, but still show partial stream
image
image
image

Halle-Astra:
> I'm not sure if I like this change. `None` clearly signals that we don't have the full body. Now we have a partial body but no good signalling that it's just partial.

Thanks for your reply and good point. I've updated the PR — when partial body data is saved, the error message now says e.g. "Client disconnected. (partial body saved)" so it's clear the body is incomplete. How about this?

PR #8232, opened 2026-05-14
fix(web): render AVIF images and IANA icon favicons in the response tab

ariel42:
#### Description

The `ViewImage` regex in `HttpMessage.tsx` rejects two valid image content types, so the response tab does not render them inline:

1. `image/avif` is missing from the alternation. AVIF is widely served by image CDNs and decoded natively in `` by Chrome 85+, Firefox 93+, and Safari 16.1+.
2. `image/vnd.microsoft.icon` (the IANA-canonical `.ico` MIME) never matches because the alternation has `vnc.microsoft.icon` — a `vnc`/`vnd` typo. The de-facto legacy `image/x-icon` continued to work, which made the bug easy to miss. The dots are also unescaped, so the regex coincidentally matches `image/vncXmicrosoftXicon` for any single character `X`.

The fix adds `avif`, corrects `vnc` → `vnd`, and escapes the dots.

Added a `ViewImage.matches` unit test covering both the additions and the existing image MIME types.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## Unreleased: mitmproxy next
 
+- mitmweb: Fix AVIF images and `image/vnd.microsoft.icon` favicons not rendering in the response tab.
 
 ## 12 May 2026: mitmproxy 12.2.3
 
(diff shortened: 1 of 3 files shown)


mhils (approved):
Thanks! 🍰

PR #8233, opened 2026-05-14
fix(web): avoid infinite componentDidUpdate loop in FlowTable

ariel42:
#### Description

`FlowTable.componentDidUpdate` calls `this.onViewportUpdate()` on every update, even when the inputs to `calcVScroll` (the flow list and the row height) did not change. `onViewportUpdate` reads the current `viewport.scrollTop`, recomputes `vScroll`, and `setState`s when `state.viewportTop !== scrollTop` or `vScroll` shape changed. The new `state.viewportTop` is `Math.min(scrollTop, vScroll.end * rowHeight)`, so when `vScroll.end * rowHeight < scrollTop` the new state stays strictly below `scrollTop`, the inequality holds on the next iteration, and the cycle spins until the browser kills the JS thread.

Gated the call on `prevProps.flowView !== this.props.flowView || prevProps.rowHeight !== this.props.rowHeight`. The other `onViewportUpdate` call sites — `componentDidMount`, the window resize listener, the viewport `onScroll` handler, and the post-scroll-into-view call — are unchanged, so user scrolling and resizing still drive the virtual-scroll window normally.

Added a `FlowTableSpec.tsx` regression test that spies on `PureFlowTable.prototype.onViewportUpdate` and asserts it is NOT called after a `select` dispatch (which changes connect-mapped props but leaves `flowView` and `rowHeight` alone). The test fails without the fix.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## Unreleased: mitmproxy next
 
+- mitmweb: Fix an infinite update cycle in `FlowTable` by only recomputing the virtual-scroll window in `componentDidUpdate` when `flowView` or `rowHeight` actually change.
 
 ## 12 May 2026: mitmproxy 12.2.3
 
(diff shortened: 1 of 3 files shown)


mhils (approved):
Good catch - thank you!

PR #8238, opened 2026-05-17
Improve TypeScript typing across mitmweb

lups2000:
#### Description

This PR tightens TypeScript typing in the `web/` frontend, focusing mainly on UI components that previously produced editor or tsc warnings.
Behaviour is intended to stay the exact same; the goal was better type safety, fewer red underlines, and cleaner tsc output :)

#### Checklist


web/src/js/components/FlowView/HttpMessages.tsx
```diff
@@ -167,16 +167,25 @@ function Trailers({ flow, message }: TrailersProps) {
...
-}: {
-    flow: HTTPFlow;
-    message: HTTPMessage;
-}) {
-    const part = flow.request === message ? "request" : "response";
-    const FirstLine = flow.request === message ? RequestLine : ResponseLine;
-
+    part,
+}:
+    | {
```
mhils, line 172: This looks wrong?

(overall I like the change, let's also add `part: "request" | "response"` (or whatever the correct typescript syntax is)?)

web/src/js/components/EventLog/EventList.tsx
```diff
@@ -47,7 +47,13 @@ export default class EventLogList extends Component<
...
     }
 
-    componentDidUpdate(prevProps, prevState, snapshot) {
+    componentDidUpdate(
+        prevProps: EventLogListProps,
+        prevState: EventLogListState,
+        snapshot: boolean,
+    ) {
+        void prevProps;
+        void prevState;
```
mhils: What does this void thing accomplish? :)

web/src/js/components/EventLog/EventList.tsx
```diff
@@ -107,12 +113,12 @@ export default class EventLogList extends Component<
...
-            warn: "exclamation-triangle",
-            error: "ban",
-        }[event.level] || "info";
+    const icon = {
+        web: "html5",
+        debug: "bug",
+        info: "info",
+        warn: "exclamation-triangle",
+        error: "ban",
+    }[event.level];
```
mhils: What about "alert" level? Was the previous fallback bad in any way?

web/src/js/components/Modal/Modal.tsx
```diff
@@ -2,9 +2,16 @@ import * as React from "react";
...
+    return name in ModalList;
+}
+
 export default function PureModal() {
     const activeModal = useAppSelector((state) => state.ui.modal.activeModal);
-    const Modal = activeModal ? ModalList[activeModal] : undefined;
+    const Modal =
+        activeModal && isModalName(activeModal)
+            ? ModalList[activeModal]
+            : undefined;
```
mhils, line 14: Not objecting to the change, but did the previous implementation miss anything?

web/src/js/utils.ts
```diff
@@ -64,32 +71,36 @@ export function formatAddress(address: [string, number]): string {
...
 
-let xsrf = () => {
+let xsrf: () => string | undefined = () => {
     const cached = getCookie("_mitmproxy_xsrf");
     xsrf = () => cached;
     return xsrf();
 };
 
-export function fetchApi(
+export const fetchApi = function (
```
mhils: Why this change?

web/src/js/components/FlowView.tsx
```diff
@@ -12,37 +12,112 @@ import WebSocket from "./FlowView/WebSocket";
...
+        displayName: TcpMessages.displayName,
+    },
+);
+const UdpMessagesTab = Object.assign(
+    ({ flow }: TabProps) =>
+        flow.type === "udp" ? <UdpMessages flow={flow} /> : null,
+    {
+        displayName: UdpMessages.displayName,
+    },
+);
```
mhils: This here above (`has...` and `...Tab`) doesn't convince me - it just seems like a lot of boilerplate that doesn't really improve anything material?

web/src/js/components/FlowView.tsx
```diff
@@ -51,9 +126,8 @@ export function tabsForFlow(flow: Flow): string[] {
             tabs = ["udpmessages"];
             break;
         case "dns":
-            tabs = ["request", "response"]
-                .filter((k) => flow[k])
-                .map((s) => "dns" + s);
+            tabs = ["dnsrequest"];
+            if (flow.response) tabs.push("dnsresponse");
```
mhils, line 108: 👍

mhils (commented):
Thanks! Lots of mostly nitpicky comments below, overall very happy with this change. :)

web/src/js/components/FlowView/HttpMessages.tsx
lups2000: Yeah, I also thought it looked weird 😅 but apparently it’s just formatting. The idea here is that we’re defining a union type, like A | B, but since it starts on a new line, the syntax becomes:
```
}:
    | A
    | B
```

what do you mean with: add`part: "request" | "response"` ? the union already narrows it

lups2000: Screenshot 2026-05-19 at 08 12 17

web/src/js/components/EventLog/EventList.tsx
lups2000: It was just a way to mark those parameters as intentionally unused (and silence TS warning). If you prefer, this is equivalent:

```
componentDidUpdate(
    _prevProps: EventLogListProps,
    _prevState: EventLogListState,
    snapshot: boolean,
) {
    if (snapshot) {
        autoscroll.adjustScrollTop(this.viewport);
    }
    this.onViewportUpdate();
}
```

web/src/js/components/EventLog/EventList.tsx
lups2000: Do we have an `"alert"` level? I didn’t know that 😅
I removed the fallback because I assumed the possible values were fixed, so I wasn’t sure we still needed it.

web/src/js/components/Modal/Modal.tsx
lups2000: No nothing is missing. This was simply added to remove:

```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{ OptionModal: () => Element; }'.
No index signature with a parameter of type 'string' was found on type '{ OptionModal: () => Element; }'.
```

on `ModalList[activeModal]`; so basically we cannot guarantee that activeModal is actually a valid key of ModalList

web/src/js/utils.ts
lups2000: no reasons, let me revert that. Just same identical meaning :)

web/src/js/components/FlowView.tsx
lups2000: Okay, let me try to simplify things while keeping the important parts, such as the `TabId` type. I agree that it probably adds too much code for little benefit

web/src/js/components/FlowView.tsx
lups2000: This is the other change, together with the `TabId` type, that needs to be maintained :)

web/src/js/components/FlowView.tsx
lups2000: better now? [4307079](https://github.com/mitmproxy/mitmproxy/pull/8238/commits/430707914acd81b22292a59060c4e8fad456e202)

web/src/js/components/EventLog/EventList.tsx
lups2000: I checked and the options are only that ones, do we still need a fallback?

mhils: https://github.com/mitmproxy/mitmproxy/blob/2bb021371653b1379a30cb1454d2d44be81da80a/mitmproxy/log.py#L16-L23

 - let's add a fallback :)

lups2000: done, I also merged main and solved the conflicts :)

mhils (approved):
Awesome, thanks! 😃

PR #8242, opened 2026-05-20
http1: don't emit chunked terminator for empty data events

Osezno-byte:
## Summary

`Http1Server.send` and `Http1Client.send` format every `ResponseData` /
`RequestData` event into a chunked-encoding frame via:

```python
raw = b"%x\r\n%s\r\n" % (len(event.data), event.data)
```

When `event.data == b""` and the response/request uses
`transfer-encoding: chunked`, the format string produces `b"0\r\n\r\n"`
— the HTTP/1.1 body-end terminator. The peer parses it as end-of-body
and stops reading. Anything emitted afterwards (legitimate chunks, the
real `0\r\n\r\n` from `ResponseEndOfMessage`) is silently discarded.

This bites `response.stream` / `request.stream` addons that buffer
incoming chunks and emit `b""` while waiting for a later flush — a
natural pattern when an addon needs to re-frame across event boundaries
(e.g. SSE rehydration that joins split tokens spanning multiple deltas).

On HTTP/2 and HTTP/3 the same `b""` is harmless: hyper-h2 / aioquic
emit a zero-byte `DATA` frame without `END_STREAM`, which the peer
ignores. The bug is HTTP/1.1-specific.

## Fix

Skip empty `ResponseData` / `RequestData` events early in
`Http1Server.send` / `Http1Client.send`, before any framing decision.

## Tests

Two new tests in `test/mitmproxy/proxy/layers/http/test_http1.py`:
- `TestServer.test_chunked_response_empty_data_no_terminator`: empty
  `ResponseData` between two real chunked-encoded chunks must produce
  no `SendData`.
- `TestClient.test_chunked_request_empty_data_no_terminator`: same on
  the request side.

Both fail without the fix (Playbook strict-mismatch on the unexpected
`SendData(b"0\r\n\r\n")`) and pass with it.

## How I found this

`response.stream` addon for a custom SSE-format PII rehydrator was
buffering events while waiting for a partial `[[LABEL` token to close.
While buffering, the callable returned `b""`. Through mitmproxy, the
upstream-side stream truncated at exactly the first `b""` return
(client only saw the chunks emitted before that point). Through a
direct connection to the same upstream, the full stream came through.

I narrowed it via:
1. tcpdump on the upstream interface: mitmproxy initiated the FIN.
2. ALPN logger via a `tls_established_server` hook: upstream ALPN
   `b''` → HTTP/1.1.
3. Variant return-value sweep: `b""` truncates, `b"\n"` (single byte)
   works fine. That isolates the bug to the empty-event chunked-encode
   path.

The `\n` chunk gets encoded as `b"1\r\n\n\r\n"` — a valid 1-byte chunk.
The `b""` chunk gets encoded as `b"0\r\n\r\n"` — the body terminator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## Unreleased: mitmproxy next
 
+- Fix `Http1Server`/`Http1Client` emitting the chunked-encoding body-end terminator `0\r\n\r\n` when a `ResponseData`/`RequestData` event carries an empty payload. Previously, a `response.stream` addon callable that returned `b""` (e.g. while buffering for a later flush) would make the client treat the response as complete and stop reading mid-stream.
 - mitmweb: Fix the filter input losing half-typed text on unrelated parent re-renders.
   ([#8234](https://github.com/mitmproxy/mitmproxy/pull/8234), @ariel42)
 - mitmweb: Fix an infinite update cycle in `FlowTable` by only recomputing the virtual-scroll window in `componentDidUpdate` when `flowView` or `rowHeight` actually change.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ## Unreleased: mitmproxy next
 
 - Fix `Http1Server`/`Http1Client` emitting the chunked-encoding body-end terminator `0\r\n\r\n` when a `ResponseData`/`RequestData` event carries an empty payload. Previously, a `response.stream` addon callable that returned `b""` (e.g. while buffering for a later flush) would make the client treat the response as complete and stop reading mid-stream.
+  ([#8242](https://github.com/mitmproxy/mitmproxy/pull/8242), @Osezno-byte)
 - mitmweb: Fix the filter input losing half-typed text on unrelated parent re-renders.
   ([#8234](https://github.com/mitmproxy/mitmproxy/pull/8234), @ariel42)
 - mitmweb: Fix an infinite update cycle in `FlowTable` by only recomputing the virtual-scroll window in `componentDidUpdate` when `flowView` or `rowHeight` actually change.
--- a/mitmproxy/proxy/layers/http/_http1.py
+++ b/mitmproxy/proxy/layers/http/_http1.py
@@ -250,6 +250,14 @@ class Http1Server(Http1Connection):
             yield commands.SendData(self.conn, raw)
         elif isinstance(event, ResponseData):
             assert self.response
+            if not event.data:
+                # Empty ResponseData is a no-op. Under chunked transfer-encoding
+                # we must NOT emit a zero-length chunk — `0\r\n\r\n` is the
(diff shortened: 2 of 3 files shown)


mitmproxy/proxy/layers/http/_http1.py
mhils, line 265: This condition is always true now, correct?

If so, let's remove it.

mitmproxy/proxy/layers/http/_http1.py
```diff
@@ -250,6 +250,14 @@ def send(self, event: HttpEvent) -> layer.CommandGenerator[None]:
             yield commands.SendData(self.conn, raw)
         elif isinstance(event, ResponseData):
             assert self.response
+            if not event.data:
+                # Empty ResponseData is a no-op. Under chunked transfer-encoding
+                # we must NOT emit a zero-length chunk — `0\r\n\r\n` is the
+                # body-end terminator and would make the client treat the
+                # response as complete mid-stream. See: stream callable in a
+                # response.stream addon returning b"" while buffering for a
+                # later flush.
```
mhils, line 259:
```suggestion
                # When using transfer-encoding: chunked, we must not 
                # emit a zero-length chunk. `0\r\n\r\n` would prematurely
                # close the stream.
```

test/mitmproxy/proxy/layers/http/test_http1.py
```diff
@@ -111,8 +111,76 @@ def test_upgrade_denied(self, tctx):
...
+            )
+            << SendData(
+                tctx.client,
+                b"HTTP/1.1 200 OK\r\ntransfer-encoding: chunked\r\n\r\n",
+            )
+            >> ResponseData(1, b"hello")
+            << SendData(tctx.client, b"5\r\nhello\r\n")
+            # Empty ResponseData must produce no SendData. Playbook is strict:
+            # any unexpected SendData here would fail the test.
+            >> ResponseData(1, b"")
```
mhils, line 144:
```suggestion
            >> ResponseData(1, b"")
            << None  # Must not generate premature close
```

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 
 ## Unreleased: mitmproxy next
 
+- Fix `Http1Server`/`Http1Client` emitting the chunked-encoding body-end terminator `0\r\n\r\n` when a `ResponseData`/`RequestData` event carries an empty payload. Previously, a `response.stream` addon callable that returned `b""` (e.g. while buffering for a later flush) would make the client treat the response as complete and stop reading mid-stream.
```
mhils, line 10: Let's focus a bit more on the user perspective:

```suggestion
- Fix a bug where addons using `request.stream` or `response.stream` callbacks would inadvertently close transfer-encoded streams when returning empty output.
```

mhils (commented):
Thanks! Good catch, only minor comments below. :)

mhils:
Hi, any progress here?

PR #8243, opened 2026-05-21
Treat carriage return as whitespace in strutils.is_xml

ADiTyaRaj8969:
## Summary

`strutils.is_xml` walks the input bytes skipping `(9, 10, 32)` — tab, LF, space — before checking for an opening `<`. CR (`\r`, `0x0D`) is also XML whitespace per [XML 1.0 §2.3](https://www.w3.org/TR/REC-xml/#NT-S), but it's missing from that tuple. So a response body that begins with a stray `\r` or `\r\n` before the root element returned `False`, and the XML/HTML content view auto-detection score (`mitmproxy/contentviews/_view_xml_html.py:273`) dropped from 0.4 to 0.

## Reproduction

```python
>>> from mitmproxy.utils import strutils
>>> strutils.is_xml(b"\r<foo>")
False
>>> strutils.is_xml(b"\r\n<foo>")
False
```

Both should be `True`. After the change:

```python
>>> strutils.is_xml(b"\r<foo>")
True
>>> strutils.is_xml(b"\r\n<foo>")
True
>>> strutils.is_xml(b"\r\nfoo")   # still rejected
False
```

## Change

One byte added to the whitespace tuple in `is_xml`, and a short comment pointing at the spec so the next reader doesn't reintroduce the gap. Test additions cover the two CR cases plus a negative case that confirms the "must be `<`" check still rejects non-XML.

```diff
-        if char in (9, 10, 32):  # is space?
+        if char in (9, 10, 13, 32):  # is whitespace?
```

## Verification

```
$ uv run pytest test/mitmproxy/utils/test_strutils.py::test_is_xml -v
test/mitmproxy/utils/test_strutils.py::test_is_xml PASSED
```

Ruff lint + format both clean on the touched files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Treat carriage return as whitespace in `strutils.is_xml`, so XML response
   bodies that arrive with a leading CR or CRLF before the root element are
   still detected by the XML/HTML content view.
+  ([#8243](https://github.com/mitmproxy/mitmproxy/pull/8243), @ADiTyaRaj8969)
 - mitmweb: Fix the filter input losing half-typed text on unrelated parent re-renders.
   ([#8234](https://github.com/mitmproxy/mitmproxy/pull/8234), @ariel42)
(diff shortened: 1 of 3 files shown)


mhils (approved):
Thanks!

PR #8245, opened 2026-05-23
Add Flowfilter description strings

lups2000:
#### Description

This PR revives the old draft in [#7710](https://github.com/mitmproxy/mitmproxy/pull/7710) and completes the missing test coverage for the new __str__ implementations in flowfilter.py.

The broader goal behind this work is to move filter descriptions to the backend so that mitmweb can eventually stop relying on **filt.js**. I initially thought that backend descriptions plus the removal of filt.js could be done together in one PR, but after digging into the mitmweb flow I think that second step needs a bit more discussion. For that reason, I’d prefer to merge this first and follow up separately with the mitmweb changes.

One thing to note is that CI currently fails on old-dependencies (https://github.com/mitmproxy/mitmproxy/actions/runs/26331517591/job/77518316616?pr=8245). I guess we can't ignore that, but how should we fix that? never encountered this type of error.

#### General Discussion for next steps

This PR is only the first step. The remaining open design question is how mitmweb should validate filter expressions once filt.js is removed.

Today mitmweb has three places that use filter expressions:

1. Search
2. Highlight
3. Intercept

Search and Highlight are closely related: they use the websocket flow-filter path and already have dedicated update messages. Intercept is different: it is configured through the /options endpoint.

Because of that, moving validation from the frontend to the backend is not entirely straightforward. The main implementation options I see are:

1. Keep Search and Highlight on the existing websocket filter-update path, and add a separate backend validation path only for Intercept.
This avoids duplicated work for Search/Highlight, because validation and application are effectively the same operation there.

2. Introduce one unified backend validation channel for all three inputs (Search, Highlight, Intercept), and keep the existing websocket//options paths only for committing valid values.
This is more uniform conceptually, but it introduces some duplication for Search and Highlight, because the same expression may be parsed once for validation and again for application.

3. Enrich the existing Search/Highlight websocket messages with validation metadata, and extend the /options endpoint so that backend validation errors can be surfaced directly in the UI. The downside is that this changes the semantics of the general options update path, not only the intercept.

Thoughts?

#### Checklist

### DIFF
--- a/mitmproxy/flowfilter.py
+++ b/mitmproxy/flowfilter.py
@@ -520,9 +520,7 @@ class FUrl(_StrRex):
         if isinstance(f, http.HTTPFlow):
             return bool(self.re.search(f.request.pretty_url))
         assert isinstance(f, dns.DNSFlow)
-        return bool(
-            f.request.questions and self.re.search(f.request.questions[0].name)
-        )
+        return bool(f.request.questions and self.re.search(f.request.questions[0].name))
 
     def __str__(self) -> str:
         return f"url matches {self.regex_str}"
--- a/mitmproxy/flowfilter.py
+++ b/mitmproxy/flowfilter.py
@@ -519,11 +519,10 @@ class FUrl(_StrRex):
             return False
         if isinstance(f, http.HTTPFlow):
             return bool(self.re.search(f.request.pretty_url))
-        elif isinstance(f, dns.DNSFlow):
-            return f.request.questions and bool(
-                self.re.search(f.request.questions[0].name)
-            )
-        return False
+        assert isinstance(f, dns.DNSFlow)
+        return bool(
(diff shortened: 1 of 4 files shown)


mhils (commented):
LGTM % CI failure! You can just bump the lower version bound for pyparsing to the current upper bound if that's easiest. :)

lups2000:
> LGTM % CI failure! You can just bump the lower version bound for pyparsing to the current upper bound if that's easiest. :)

easy 😄 regarding the next steps do you have any thoughts? should we move the conversation on discord?

mhils:
Thanks! Let's finally get this in. 😃 

Going forward, I think approaches 2-3 are both reasonable. 2) probably makes most sense. But no strong opinions at this point, I take anything that removes the legacy cruft. 😃

lups2000:
Perfect 🙌 thanks Max

PR #8249, opened 2026-05-30
Fix/#8185

amamarante923:
#### Description

Feature: Added option to use Microsoft Windows Certificate Store as a source of trusted CAs for mitmproxy. 

#### Checklist

 - [yes ] I have updated tests where applicable.
 - [yes ] I have added an entry to the CHANGELOG.


mhils (commented):
Thanks! I'm generally open to adding support for this, but this sort of code needs to live in pyOpenSSL. Our current use of `SSL._lib` already isn't ideal, and we don't want to pile more of it on top of it.

amamarante923:
> Thanks! I'm generally open to adding support for this, but this sort of code needs to live in pyOpenSSL. Our current use of `SSL._lib` already isn't ideal, and we don't want to pile more of it on top of it.

No problem. I'm closing it then. Feel free to direct me to any other issue that I can solve at the moment.

PR #8251, opened 2026-05-31
Fix #8138: log warning when load_pem_private_key retries without password

ErnestHysa:
## Description

When `load_pem_private_key` encounters a TypeError (e.g., a passphrase is provided but the private key is not actually encrypted), it silently retries with `password=None`. This hides a potential misconfiguration where a user specifies a passphrase for a non-encrypted key.

This PR adds a `logger.warning` message to inform users about this fallback, helping them identify certificate configuration issues.

## Fix

In `mitmproxy/certs.py`, the `load_pem_private_key` function now logs a warning before retrying without a password.

## Testing

- Code compiles and passes basic checks
- Tests pass (CI)

### DIFF
--- a/mitmproxy/certs.py
+++ b/mitmproxy/certs.py
@@ -737,5 +737,9 @@ def load_pem_private_key(data: bytes, password: bytes | None) -> rsa.RSAPrivateK
         return serialization.load_pem_private_key(data, password)  # type: ignore
     except TypeError:
         if password is not None:
+            logger.warning(
+                "Unable to load private key with the provided password, "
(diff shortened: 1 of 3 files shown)


mhils (commented):
Thank you for the PR! Can you 1) remove the unrelated changes, and 2) change the logging so that it only triggers if loading without the password is successful? I.e. something like this:

```python3
key = load_pem_private_key(data, None)
logger.warning("A password was specified, but the provided private key did not require a password.")
return key
```

lups2000:
Closing this since it has unrelated changes and it seems inactive. There is this now: #8345

PR #8276, opened 2026-06-07
Handle comma-separated Content-Encoding in decode/encode

### LINKED ISSUE #2246: KeyError: 'sdch, gzip'

1:     #----------------------------------------------------------------------
    def response(self, flow):
        # Gather outgoing data size
        if flow.response.content:
            self.outgoing_data_size += len(flow.response.raw_content)
2. Go to google.com and search "test"
3. Red text flys by saying:
Script error: Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/mitmproxy/net/http/encoding.py", line 48, in decode
    decoded = custom_decode[encoding](encoded)
KeyError: 'sdch, gzip'

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.5/dist-packages/mitmproxy/net/http/encoding.py", line 50, in decode
    decoded = codecs.decode(encoded, encoding, errors)
LookupError: unknown encoding: sdch, gzip

During handling of the above exception, another exception occurred:

  File "main.py", line 791, in response
    if flow.response.content:
  File "/usr/local/lib/python3.5/dist-packages/mitmproxy/net/http/message.py", line 97, in get_content
    content = encoding.decode(self.raw_content, ce)
  File "/usr/local/lib/python3.5/dist-packages/mitmproxy/net/http/encoding.py", line 61, in decode
    repr(e),
ValueError: LookupError when decoding b'\x1f\x8b with 'sdch, gzip': LookupError('unknown encoding: sdch, gzip',)

Looks like gzip is supported, but I don't see anything about sdch. Looks to be missing

clmoon2:
Content-Encoding can technically be a list like "sdch, gzip", and right now decode just does one dict lookup on the whole string and blows up with a KeyError/LookupError. so if a server sends something like that, decoding the body throws instead of working.

fixed it by splitting on commas and applying each codec in turn (reverse order for decode, forward for encode). also treat sdch as identity since we never actually decode it and it was the thing tripping people up here. went with handling it in encoding.py per the thread where mhils said we should just deal with this.

added a test covering the sdch, gzip case plus whitespace and casing, full suite passes locally. closes #2246

### DIFF
--- a/mitmproxy/net/encoding.py
+++ b/mitmproxy/net/encoding.py
@@ -48,6 +48,16 @@ def decode(
         return None
     encoding = encoding.lower()
 
+    # Content-Encoding can technically be a list like "sdch, gzip", so if we
+    # see commas we just peel off one codec at a time. decode order is the
+    # reverse of how it was applied, so handle the last token first.
+    # https://github.com/mitmproxy/mitmproxy/issues/2246
+    if "," in encoding:
+        encodings = [e.strip() for e in encoding.split(",") if e.strip()]
+        for e in reversed(encodings):
+            encoded = decode(encoded, e, errors)
+        return encoded
+
     global _cache
     cached = (
         isinstance(encoded, bytes)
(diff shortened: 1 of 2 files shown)


clmoon2:
ci was just individual_coverage, the sdch no-op on the encode side never got exercised so encoding.py sat at 99%. added a one line assert covering it, green now

PR #8287, opened 2026-06-20
Add on-the-fly zstandard compression for --save-stream-file

### LINKED ISSUE #8286: Add on-the-fly compression support for --save-stream-file
#### Problem Description

 When using `mitmdump --save-stream-file` to capture flows over long periods or high-traffic proxying, the resulting files can grow very large. There's no built-in way to compress them on the fly - users must compress after the fact, which doesn't work well with strftime-rotated files or disk space constraints during capture.

 #### Proposal

 Add a `--set save_stream_compression=gz|bz2|xz` option that compresses stream files on the fly. Optionally accepts a level (e.g. `gz=9`). On the read side, auto-detect compressed files via magic bytes so `-r`, `view.flows.load`, replay addons, and the Python API transparently read gz/bz2/xz files without any extra flags.

 #### Alternatives

 - Infer compression from file extension (`.gz`) - rejected because this would be a breaking change if someone happens to use these extensions today for uncompressed files.
 - Post-hoc compression - wastes disk I/O and doesn't help with disk space constraints during capture.

 #### Additional context

 I have a working implementation with tests ready to open as a PR.

joelverhagen:
#### Description

Closes #8286.

Adds a `save_stream_compress` boolean option that compresses stream files on the fly using zstandard.

On the read side, `FlowReader` auto-detects zstandard-compressed files via magic bytes — no extra flags needed. This works transparently with `-r`, `view.flows.load`, replay, and `read_flows_from_paths()`.

**Write side:**
- New `save_stream_compress` option (bool, default false)
- Works with strftime rotation and append mode (`+` prefix)
- Guards against appending compressed data to uncompressed files (and vice versa)

**Read side:**
- `FlowReader` sniffs the first 4 bytes for zstd magic and wraps in a decompressor transparently
- Graceful error handling for corrupt compressed files (`ZstdError` -> `FlowReadException`)

#### Checklist

- [x] I have added an entry to the CHANGELOG


mitmproxy/addons/save.py
```diff
@@ -64,6 +71,12 @@ def load(self, loader):
             None,
             "Filter which flows are written to file.",
         )
+        loader.add_option(
+            "save_stream_compression",
```
mhils: This should probably be `save_stream_compress`?

test/mitmproxy/tools/console/test_integration.py
```diff
@@ -7,6 +7,25 @@ def test_integration(tdata, console):
     assert "http://example.com/" in console.screen_contents()
 
 
+def test_load_compressed_zstd(tmp_path, console):
```
mhils: This doesn't need an extra integration test, remove.

test/mitmproxy/addons/test_save.py
```diff
@@ -197,3 +205,85 @@ def _raise(*_):
...
+        sa.request(f2)  # second request already started.
+        tctx.configure(sa, save_stream_file=str(tmp_path / "b"))
+        sa.response(f2)
+        sa.done()
+
+        assert len(rd_zstd(tmp_path / "a")) == 1
+        assert len(rd_zstd(tmp_path / "b")) == 1
+
+
+def test_tcp_zstd(tmp_path):
```
mhils: Why do we need a special test for TCP?

mhils (changes requested):
This looks slopped and not integrated into what's already there. FlowReader needs to support zstd instead of adding `open_flow_file`/`read_flows_from_paths` wrappers that complicate control flow.

mitmproxy/addons/save.py
joelverhagen: Sure, happy to change it. The other options looked more like noun-ish not verbs. But I guess technically filter and file can also verbs 😅.

test/mitmproxy/addons/test_save.py
joelverhagen: In my manual testing I was able to get HTTP and HTTP websocket flows populated to disk. This allowed me to exercise the TCP flow type as well. But to be fair I didn't get DNS or UDP in my manual testing and I didn't add UTs for that. I'll remove it.

test/mitmproxy/tools/console/test_integration.py
joelverhagen: Sounds good. Sorry, I didn't know the bar for new integration test vs. only UTs.

joelverhagen:
> This looks slopped and not integrated into what's already there. FlowReader needs to support zstd instead of adding open_flow_file/read_flows_from_paths wrappers that complicate control flow.

`read_flows_from_paths` is pre-existing, but I'll try out what you're suggesting.

mitmproxy/addons/save.py
joelverhagen: Done.

test/mitmproxy/tools/console/test_integration.py
joelverhagen: Done.

test/mitmproxy/addons/test_save.py
joelverhagen: Done.

joelverhagen:
I didn't eliminate the `read_flows_from_paths` since it was pre-existing but `FlowReader` now handles the zstd internally. I also improved the error handling when a the user attempts to mix uncompressed and compressed in the same append file.

PR #8288, opened 2026-06-21
mitmweb: honor view_order_reversed for live flows

### LINKED ISSUE #5520: Can't reverse the order of captured requests even with view_order_reversed set to true

hexbinoct:
Fixes #5520.

## Problem

In mitmweb the `view_order_reversed` option only affected the initial flow list. Flows that arrive afterwards over the WebSocket were always appended to the bottom of the table, so enabling the option did not keep the newest flow visible at the top. This is the behaviour reported in the issue, and it matches what Prinzhorn observed there (the option is ignored and new rows are appended).

## Cause

The initial `/flows` response is sent in the server view order, which already respects `view_order_reversed`, and the client keeps that order when no column sort is active. Live `flows/add` events go through `insertViewItem` instead, and with no active column sort the comparator is a no-op, so `findInsertPos` always returns the end of the list. The client never consulted `view_order_reversed`.

## Change

The flows reducer now tracks `view_order_reversed` from the options state. When no column sort is active, a newly added flow is inserted at the front if the option is set and at the back otherwise, which is the same order the server uses for the initial list. An active column sort still takes precedence and is unchanged.

The fix is limited to newly added flows, which are unambiguously the most recent. The order restored when a column sort is toggled back off is left as it was.

## Tests

Added reducer tests covering the option being read from `OPTIONS_RECEIVE` and `OPTIONS_UPDATE`, front insertion when reversed, back insertion when not, and a column sort taking precedence over the option. `npm test` (eslint, tsc, jest) passes locally.

### DIFF
--- a/web/src/js/ducks/flows/index.ts
+++ b/web/src/js/ducks/flows/index.ts
@@ -84,7 +84,7 @@ export default function flowsReducer(
     action: UnknownAction,
 ): FlowsState {
     if (FLOWS_RECEIVE.match(action)) {
-        const { sort } = state;
+        const { sort, orderReversed } = state;
         const list = action.payload;
         const _listIndex = buildIndex(list);
         const byId = new Map(list.map((f) => [f.id, f]));
@@ -104,7 +104,7 @@ export default function flowsReducer(
             view,
             _viewIndex,
             sort,
-            orderReversed: state.orderReversed,
+            orderReversed,
             selected,
             selectedIds,
             highlightedIds,
@@ -114,7 +114,7 @@ export default function flowsReducer(
         if (state._listIndex.has(flow.id)) {
             return state; // WebSocket/HTTP race
         }
-        const { sort, selected, selectedIds } = state;
+        const { sort, selected, selectedIds, orderReversed } = state;
         let { view, _viewIndex, highlightedIds } = state;
         // Update list
         const _listIndex = new Map(state._listIndex);
@@ -132,7 +132,7 @@ export default function flowsReducer(
                 _viewIndex,
                 flow,
                 sort,
-                state.orderReversed,
+                orderReversed,
             ));
         }
         // Update highlight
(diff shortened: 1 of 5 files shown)


web/src/js/ducks/flows/index.ts
```diff
@@ -97,6 +104,7 @@ export default function flowsReducer(
             view,
             _viewIndex,
             sort,
+            orderReversed: state.orderReversed,
```
lups2000: just for consistency, I would define a variable like the others above

web/src/js/ducks/flows/index.ts
```diff
@@ -139,6 +148,7 @@ export default function flowsReducer(
...
```
lups2000: same thing, this and the ones below

lups2000 (changes requested):
Thanks a lot for the PR 🚀 this looks good to me, just left a couple of nits. 
I tested it and it seems working well :)

hexbinoct:
Thanks for the review and for testing it! Good call. I pushed d7da5b5 that destructures orderReversed from state at the top of each branch and uses the shorthand in the returns, matching the style of the surrounding variables. Let me know if you spot anything else.

lups2000 (approved):
LGTM 🚀

Issue #8253, opened 2026-06-01
CSS and XML/HTML prettifiers produce incorrect output on Windows due to CRLF line endings


hhh714:
## Bug Description
On Windows, the CSS and XML/HTML content view prettifiers fail to format 
correctly because Windows uses CRLF (\r\n) line endings, which are not 
normalized before processing.

## Steps to Reproduce
1. Clone the repository on Windows
2. Run: pytest test/mitmproxy/contentviews/test__view_css.py
3. Run: pytest test/mitmproxy/contentviews/test__view_xml_html.py

## Expected Behavior
All tests pass.

## Actual Behavior
4 tests fail with assertion errors showing mismatched line endings.

## Fix
Add line ending normalization at the start of the prettify functions:
data = data.replace("\r\n", "\n").replace("\r", "\n")

A PR with the fix is ready.

### System Information

```raw

```

### Checklist

- [x] This bug affects the [latest mitmproxy release](https://github.com/mitmproxy/mitmproxy/releases).

PR #8294, opened 2026-06-29
net.tls: handle SSL.Error from set_min/max_proto_version in is_supported_version

### LINKED ISSUE #8264: is_supported_version raises unhandled OpenSSL.SSL.Error for disabled TLS versions (e.g. SSL3)
## Bug Description
`is_supported_version()` in `mitmproxy/net/tls.py` crashes with an unhandled 
`OpenSSL.SSL.Error` when called with `Version.SSL3` on modern OpenSSL builds.

## Steps to Reproduce
1. Clone the repository on a system with modern OpenSSL (1.1.0+)
2. Run: pytest test/mitmproxy/net/test_tls.py test/mitmproxy/addons/test_tlsconfig.py

## Root Cause
Modern OpenSSL has removed SSLv3 support following the POODLE attack 
(CVE-2014-3566). When `set_min_proto_version(SSL3_VERSION)` is called, 
OpenSSL raises SSL.Error immediately at context setup time.
The existing try/except only wraps `client_conn.recv()`, missing this 
earlier failure point.

## Fix
Wrap `set_min_proto_version` and `set_max_proto_version` in a try/except 
SSL.Error block, returning False when the version is rejected at context 
setup time.

A PR with the fix is ready.

### System Information

```raw

```

### Checklist

gaurav0107:
## Description

`mitmproxy.net.tls.is_supported_version()` builds an `SSL.Context`, sets the
min/max protocol version, then probes support with `client_conn.recv()`. Only
the `recv()` probe was wrapped in `try/except SSL.Error`.

On OpenSSL builds that have dropped a protocol version entirely (e.g. SSLv3
after POODLE / CVE-2014-3566, see pyca/cryptography#9523),
`set_min_proto_version(SSL3_VERSION)` raises `SSL.Error` immediately at
context-setup time — *before* the guarded `recv()`. The exception then
propagated unhandled and crashed callers. The only production caller,
`TlsConfig._warn_unsupported_version`, iterates over every `Version` to build
the "supported versions" list, so on those builds option validation crashed
outright.

This moves the protocol-version setters and connection setup inside the
existing `try`, so a setup-time `SSL.Error` is treated the same as a
probe-time one: the version is reported as unsupported (`return False`) instead
of raising. Behavior for currently-supported versions is unchanged.

A regression test (`test_supported_setup_error`) monkeypatches
`SSL.Context.set_min_proto_version` to raise `SSL.Error`, reproducing the crash
path deterministically on any OpenSSL build (the existing `test_supported[SSL3]`
only exercises the `recv()` path on builds that still allow SSL3 context setup).

Closes #8264

Note: #8254 also touches this function but bundles it with an unrelated CRLF
prettifier fix (#8253) across four files. This PR is a focused, single-concern
alternative scoped to #8264 only; happy to defer to whichever the maintainers
prefer.

## Checklist

- [x] I have updated the CHANGELOG (`## Unreleased: mitmproxy next`).


gaurav0107:
Heads up: the only red check is `build (ubuntu-latest, linux-x86_64, debian)`, which failed at the `Run apt-get update` container-setup step (before checkout) — a transient apt-mirror hiccup unrelated to this change. Every code check is green: lint, mypy, individual_coverage, the full test matrix across Linux/macOS/Windows and Python 3.12–3.14, codecov/patch, and docs. A re-run of that one job should clear it. Happy to adjust anything on request.

gaurav0107:
Update: I squashed the branch to a single commit (`ad5f556`) and re-pushed to re-trigger CI. The previously-red `build (ubuntu-latest, linux-x86_64, debian)` job — which had failed at the `apt-get update` container-setup step, before checkout — is now green, and the `check` gate passes as well. All checks are green and the code diff is unchanged (still the same net.tls fix + regression test); verified locally with the `test/mitmproxy/net/test_tls.py` suite, `ruff`, and `mypy`. Ready for review whenever you have a moment.

test/mitmproxy/net/test_tls.py
```diff
@@ -15,6 +15,23 @@ def test_supported(version):
...
+    def raise_ssl_error(self, version):
+        raise SSL.Error("unsupported protocol version")
+
+    monkeypatch.setattr(SSL.Context, "set_min_proto_version", raise_ssl_error)
+    tls.is_supported_version.cache_clear()
+    try:
+        assert tls.is_supported_version(tls.Version.SSL3) is False
+    finally:
+        # Don't leave a result cached under the monkeypatched setter.
+        tls.is_supported_version.cache_clear()
```
mhils: Let's remove this test - it's not really adding value.

PR #8295, opened 2026-06-30
Don't stall QUIC flows when host ignoring is enabled

tbodt:
If allow_hosts or ignore_hosts was set, any QUIC connections would stall forever after TLS handshake and never get assigned an HttpLayer or e.g. appear in the flow list in mitmweb. The codepath goes like this:

1. After client TLS handshake completes, next layer hook is invoked with data_client = b'', i.e. the data so far inside the QUIC stream
2. _next_layer first checks using _ignore_connection
3. _ignore_connection uses _get_client_hello, passing b'' data_client
4. quick_parse_client_hello_from_datagram raises a type of ValueError because b'' is definitely not a valid QUIC packet
5. ValueError is skipped, so dtls_parse_client_hello is tried, which also raises a type of ValueError for about the same reason
6. This is now converted into NeedsMoreData, which bails out of the entire _next_layer process.

This repeats each time next_layer is called and never makes progress, even if data is read on the QUIC stream, it's still not going to be a valid QUIC or DTLS packet.

This fix matches the tcp case and only tries to parse the packet as quic or dtls if it heuristically looks like that.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@
   ([#8232](https://github.com/mitmproxy/mitmproxy/pull/8232), @ariel42)
 - mitmweb: Fix correctly displaying multiple blank lines in content renderer.
   ([#8248](https://github.com/mitmproxy/mitmproxy/pull/8248), @vincentdehaan)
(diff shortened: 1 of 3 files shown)


mhils (approved):
Thanks! 🍰

Issue #185, opened 2013-12-02
mitmproxy: page down does not work on Macbook Pro


djensen47:
Since the Macbook Pro does not have a Page up/down key, I have to use Fn+Down and Fn+Up. However, it does not work when mitmproxy is running and I'm viewing a flow using the Terminal.app. If I page up (fn+up), it will page up the terminal window but not the response window area.

May I suggest alternate keys? Ctrl-U and Ctrl-D is what I use in vim for paging.

Thanks.

PR #8312, opened 2026-07-06
fix(web): avoid infinite componentDidUpdate loop in the event log

### LINKED ISSUE #8292: mitmweb event log "Maximum update depth exceeded" (React #185) on scroll; already fixed for FlowTable
### Problem Description

This was previously reported and fixed for FlowTable but not event log:
**#5198**  fixed by **#5507**  plus follow-up **#5661** 
Separately, another recent fix in **#8233**

Steps to reproduce:
In mitweb, expand event log and scroll up and down when there is enough text to make it scrollable

### System Information

```raw
Mitmproxy: 12.2.3
Python:    3.14.5
OpenSSL:   OpenSSL 4.0.0 14 Apr 2026
Platform:  Linux-5.4.0-73-generic-x86_64-with-glibc2.41
```

### Checklist

hexbinoct:
#### Description

Fixes #8292.

Expanding the mitmweb event log and scrolling it (with enough lines to be scrollable) throws React's "Maximum update depth exceeded" (#185). This is the same class of bug that was reported and fixed for `FlowTable` in #8233, but the event log was never given the same treatment.

`EventList.componentDidUpdate` calls `this.onViewportUpdate()` on every update, even when the inputs to `calcVScroll` (the event list and the row height) have not changed. `onViewportUpdate` reads the current scroll position, recomputes the virtual-scroll window, and calls `setState` when the window shape changes. Because the event log measures per-row heights, a `setState` it produces can feed itself: rendering a new window remeasures rows, which shifts the computed window, which triggers `setState` again. With variable heights the window can ping-pong between two shapes and never converge, so the cycle runs until React aborts it.

The fix gates that call on `prevProps.events !== this.props.events || prevProps.rowHeight !== this.props.rowHeight`, mirroring #8233. The other call sites still drive updates as needed: `componentDidMount`, the `resize` listener, the viewport `onScroll` handler, and `setHeight` when a row is first measured. Scrolling and the autoscroll-to-bottom behavior are unaffected.

I added a regression test that spies on `onViewportUpdate` and asserts it does not run when a re-render leaves `events` and `rowHeight` unchanged (it fails without the gate), plus a complement test proving it still runs when the events list changes.

`eslint`, `tsc --noEmit`, `prettier --check`, and the full `jest` suite (316 tests) pass locally.


mhils (approved):
Thanks! 🍰

Issue #7789, opened 2025-07-01
Mitmweb rewrite progress


sleeyax:
This issue exists to keep track of progress on https://github.com/mitmproxy/mitmproxy/pull/7784. If I missed anything let me know (or update the issue if you have the right permissions to do so)!

---

# Checklist

## Legend

- [x] = done
- [ ] = not done
- [!] = almost done

### Content views

- [x] Code editor (codemirror)
- [x] View HTTP message
- [ ] Edit HTTP request (e.g. during interception)
- [x] View image
- [x] Content view selector

#### Additional features

- [x] 'Show all' button on the code editor to show the full message

### Flow table

- [x] View flow table
- [x] Sort by clicking on header columns
- [x] Virtualized list (using react-virtuoso instead of a custom implementation)

#### Additional features

- [ ] Configure active web columns (the `web_columns` option) by right clicking / opening a menu somewhere on the table header itself
- [!] Context menu on flow table rows. This replaces the 'flow menu' component for individual and multiple selected flows.
  - The context menu exists, but is not complete. It currently only supports the 'replay' action for an individual flow.
- [ ] Adjustable column width. Would be nice to have a 'drag' handle on the column header to allow users to adjust the width of a column.

#### Known bugs

- The context menu (i.e. the pop up that shows when the user R mouse clicks on a row) closes itself when a new flow is added to the table.
- Keyboard navigation works, but the list doesn't scroll as you navigate through the items. This is because react-virtuoso expects a specific implementation through the use of `ref` and `scrollToIndex`, which is not implemented yet. This can be fixed by excluding the arrow up and down keys from the keyboard component and implementing the keyboard navigation logic according to the `react-virtuoso` documentation instead.

#### Performance

Performance is not up to par compared to the current web implementation.

The issue can be best explained by seeing it occur in development mode. Proxy your browser traffic and navigate back and forth different websites quickly. You should see that it looks like the list is 'lagging behind' on the actual requests that are being sent at the moment of navigation.

[...]

PR #8316, opened 2026-07-08
feat(web): add semantic CSS custom-property color layer

sleeyax:
Introduces a `:root` set of semantic color tokens and routes recurring colors through them. This is the foundation for the dark theme in the follow-up PR. Colors consumed by LESS functions and one-off decorative colors are intentionally left as literals. 

Part of #7789.

### DIFF
--- a/web/src/css/command.less
+++ b/web/src/css/command.less
@@ -1,5 +1,5 @@
 .command-title {
-    background-color: #f2f2f2;
+    background-color: var(--mitmweb-bg-alt);
     border: 1px solid #aaa;
 }
 
--- a/web/src/css/command.less
+++ b/web/src/css/command.less
@@ -1,12 +1,12 @@
 .command-title {
     background-color: var(--mitmweb-bg-alt);
-    border: 1px solid #aaa;
+    border: 1px solid var(--mitmweb-border-strong);
 }
 
 .command-result {
     display: block;
     margin: 0px;
-    background-color: #fcfcfc;
+    background-color: var(--mitmweb-bg);
     height: 100px;
     max-height: 100px;
     overflow: auto;
--- a/web/src/css/command.less
+++ b/web/src/css/command.less
@@ -12,12 +12,8 @@
     overflow: auto;
 }
 
-.command-suggestion {
-    background-color: #9c9c9c;
-}
-
 .argument-suggestion {
-    background-color: hsla(209, 52%, 84%, 0.5) !important;
+    background-color: var(--mitmweb-highlight) !important;
 }
 
 .command > .popover {
--- a/web/src/css/dropdown.less
+++ b/web/src/css/dropdown.less
@@ -39,8 +39,8 @@
     > li > a:hover,
     > li > a:focus {
         text-decoration: none;
-        color: #262626;
-        background-color: #f5f5f5;
+        color: var(--mitmweb-fg-strong);
+        background-color: var(--mitmweb-bg-muted);
     }
 
     max-height: 250px;
--- a/web/src/css/dropdown.less
+++ b/web/src/css/dropdown.less
@@ -13,7 +13,6 @@
     text-align: left;
     background-color: var(--mitmweb-bg);
     border: 1px solid var(--mitmweb-border);
-    border: 1px solid rgba(0, 0, 0, 15%);
     border-radius: 4px;
     box-shadow: 0 6px 12px rgba(0, 0, 0, 17.5%);
     background-clip: padding-box;
--- a/web/src/css/dropdown.less
+++ b/web/src/css/dropdown.less
@@ -11,8 +11,8 @@
     list-style: none;
     font-size: 14px;
     text-align: left;
-    background-color: #fff;
-    border: 1px solid #ccc;
+    background-color: var(--mitmweb-bg);
+    border: 1px solid var(--mitmweb-border);
     border: 1px solid rgba(0, 0, 0, 15%);
     border-radius: 4px;
     box-shadow: 0 6px 12px rgba(0, 0, 0, 17.5%);
@@ -31,7 +31,7 @@
         clear: both;
         font-weight: 400;
         line-height: 1.42857143;
-        color: #333;
+        color: var(--mitmweb-fg);
         white-space: nowrap;
         text-decoration: none;
     }
@@ -51,5 +51,5 @@
     height: 1px;
     margin: 9px 0;
     overflow: hidden;
-    background-color: #e5e5e5;
+    background-color: var(--mitmweb-border-lighter);
 }
--- a/web/src/css/flowdetail.less
+++ b/web/src/css/flowdetail.less
@@ -71,7 +71,7 @@
 
     hr {
         border: 0;
-        border-top: 1px solid #eee;
+        border-top: 1px solid var(--mitmweb-border-lighter);
         margin: 0;
     }
 }
@@ -128,7 +128,7 @@
     word-break: break-all;
 
     tr:not(:first-child) td {
-        border-top: 1px solid #f7f7f7;
+        border-top: 1px solid var(--mitmweb-border-lighter);
     }
 
     td {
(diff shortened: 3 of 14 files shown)


lups2000 (commented):
Thanks for the PR :) Overall, the changes look good to me! 🙌 

My only concern is that there are still some border, text, and background colors hardcoded directly in the CSS. Some of them may only be used in one place, so rather than introducing new design tokens for each one, we could potentially reuse existing tokens where appropriate.

For example, I found these:

```css
.command-title {
    background-color: var(--mitmweb-bg-alt);
    border: 1px solid #aaa;
}

.command-result {
    display: block;
    margin: 0;
    background-color: #fcfcfc;
    height: 100px;
    max-height: 100px;
    overflow: auto;
}

.command-suggestion {
    background-color: #9c9c9c;
}
```

These colors look very similar to some of the ones we have already defined.

I assume the main goal of this change is to define the main colors in a single place, which would also make dark mode easier to support.

What do you think? Could you perhaps go through the Less files and check for any remaining hardcoded colors? Thanks :)

sleeyax:
Thanks! Good call. I went through all the LESS files and routed the remaining background, text, and border colors through the existing tokens where it fits 

That covers the `.command-*` examples you pointed out plus a few others.

Light-mode stays visually the same. I left a handful of one-off decorative colors as literals (e.g. `.command-suggestion`, the `.first-line` banner, the mode category markers), along with the colors that LESS functions like `lighten()` consume since those can't take a `var()`.

web/src/css/command.less
lups2000: just noticed that this seems dead style

lups2000: should we define a new design token for that? it seems used in **header.less** as well

web/src/css/dropdown.less
```diff
@@ -11,8 +11,8 @@
     list-style: none;
     font-size: 14px;
     text-align: left;
-    background-color: #fff;
-    border: 1px solid #ccc;
+    background-color: var(--mitmweb-bg);
+    border: 1px solid var(--mitmweb-border);
     border: 1px solid rgba(0, 0, 0, 15%);
```
lups2000: what happens here?

web/src/css/flowdetail.less
lups2000: maybe we can reuse `--mitmweb-accent` ?

web/src/css/mode.less
lups2000: this?

web/src/css/tabs.less
```diff
@@ -5,22 +5,22 @@
...
-            color: #337ab7;
-            background-color: white;
+            color: var(--mitmweb-accent);
+            background-color: var(--mitmweb-bg);
             border-color: @separator-color;
-            border-bottom-color: white;
+            border-bottom-color: var(--mitmweb-bg);
         }
         &.special {
             @special-color: #396cad;
```
lups2000: maybe `--mitmweb-accent` ?

web/src/css/flowtable.less
lups2000: just wondering, can't we define them as design tokens in **global.less**?

lups2000 (commented):
Just left a couple more nits :) In general, I would like to have as many colors as possible defined in a single place and referenced through `var()`. This should also make dark mode easier to maintain and review.

We could perhaps leave standard colors such as `green` or `red` as literals, but even then, if their values need to change in dark mode, I would introduce a design token for them. Does that make sense?

web/src/css/command.less
sleeyax: Good catch. `.command-suggestion` is unused (only `.argument-suggestion` is referenced, in `CommandBar.tsx`), it's been dead since the command bar landed. Removed.

sleeyax: Agreed.  It's the same value used for the row hover-highlight in header.less. Pulled both into a `--mitmweb-highlight` token.

web/src/css/dropdown.less
sleeyax: This is a leftover Bootstrap fallback: the `rgba(0, 0, 0, 15%)` line always overrode the `var(--mitmweb-border)` line above it, so the token was actually dead and the border was rendering as translucent black. That would disappear against a dark background, so I dropped the rgba line and let `--mitmweb-border` drive the border like it does everywhere else. 

Heads up that this nudges the light-mode border by about one shade (~`#d9d9d9` -> `#cccccc`), which is imperceptible in practice and brings the dropdown in line with all the other bordered surfaces.

web/src/css/command.less
lups2000: Maybe we can use a different name, depending on what we decide here: [https://github.com/mitmproxy/mitmproxy/pull/8316#discussion_r3568585890](https://github.com/mitmproxy/mitmproxy/pull/8316#discussion_r3568585890)

`--mitmweb-highlight` sounds more like a flow table highlight to me, no? Just use whatever you think makes the most sense; just make sure the token names are clear and consistent in the end :) Thanks 🙌

web/src/css/flowdetail.less
sleeyax: `#428bca` is close to `--mitmweb-accent` (`#337ab7`) but not exactly the same, so reusing it would be a visible color change to the light theme. It's a one-off decorative color that reads fine on both light and dark backgrounds, so the follow-up dark-mode PR leaves it as-is too. 

I'm inclined to keep it a literal in this PR so the light theme remains strictly visually the same, but happy to reuse the accent (or give it its own token) if you prefer that.

web/src/css/mode.less
sleeyax: Good spot.  The `#b2b2b2` color is used a couple times in this file for grays and muted text, so I've pulled it into a `--mitmweb-gray` token.

 I've left the green `#77c77a` border as a literal since it only appears once, but we can tokenize that too if you'd prefer everything consistent!

web/src/css/tabs.less
sleeyax: Two reasons to leave this one as-is:
-  `#396cad` isn't equal to `--mitmweb-accent` (`#337ab7`), so reusing it would be a visible color change. 
-  `@special-color` is fed to `lighten(@special-color, 10%)` for the hover state, and LESS's `lighten()` can't operate on a CSS `var()`. That's why the token layer intentionally leaves colors consumed by LESS functions as literals. The follow-up dark-mode PR leaves this one as a literal too. Let me know if you'd still like me to change it, or if I'm missing something regarding this last point.

web/src/css/flowtable.less
sleeyax: These table shades are each used in exactly one place and their dark values are tuned per state rather than a 1:1 swap, so they aren't really shared semantic tokens. The dark-mode PR themes them with a small targeted `[data-theme="dark"] .flow-table tr` override instead, which keeps `global.less` for genuinely **global** reusable colors. I'd lean toward leaving them as literals here, but happy to tokenize if you'd prefer everything centralized.

lups2000:
@mhils do you have any thoughts on this? I know you weren’t a big fan of dark mode 😅, but nowadays with AI agents, it might be easier to maintain than it was in the past. This PR is preparing the UI for dark mode.

Personally, I would try to centralize the color definitions as much as possible so that we can reference them through tokens. However, some colors are currently used in only one place, so the options seem to be:

1. Leave them as they are, defined directly with their color values.
2. Reuse an existing token when the color is sufficiently similar.
3. Introduce a new token.

Perhaps you have a different approach or other suggestions here :)

sleeyax:
Refactored the last hardcoded colors in the `.less` files into `global.less` CSS tokens, reusing `--mitmweb-accent` for the flow first line and special tab, and adding `--mitmweb-green` and `--mitmweb-row-*` tokens. That way every color is now maintained in one place.

lups2000 (approved):
Nice! LGTM :) There’s just one conflict to resolve. We can either remove the changelog entry and keep only the dark mode one later, or resolve the conflict wth main. You can decide :)

sleeyax:
> Nice! LGTM :) There’s just one conflict to resolve. We can either remove the changelog entry and keep only the dark mode one later, or resolve the conflict wth main. You can decide :)

Removed the changes to the changelog. This is kind of an "in-between" stage anyways, the dark theme addition is far more interesting to put in there.

PR #8317, opened 2026-07-08
feat(web): add dark theme

### LINKED ISSUE #7828: Dark Theme Mode
#### Problem Description

I use mitmweb personally a lot in reverse engineering Android apps API calls. One thing that really bothers me is the lack of a dark theme mode. My whole systems & apps work in dark mode, as soon as I open mitmweb it burns my eyes and I have to tolerate the pain for quit some time till I switch back to the IDE or anything else.

#### Proposal
I am proposing to add dark theme mode and the ability to choose between different themes (Light, Dark).

sleeyax:
**Requires https://github.com/mitmproxy/mitmproxy/pull/8316 to be merged first.**

Adds a `web_theme` option (`system`/`dark`/`light`, default `system`) and a dark palette that overrides the color tokens under `[data-theme="dark"]`. `system` follows the OS `prefers-color-scheme` and reacts to live changes. The CodeMirror content editor uses the `one-dark` theme in dark mode. 

Historically, dark mode was declined on maintenance-cost grounds (#3886), but this implementation keeps the cost to a single token-override block plus only a handful of targeted overrides. Hopefully that's simple enough of an implementation to convince you to merge this :)

Closes #7828.

Part of #7789.

## Screenshots

image

### DIFF
--- a/mitmproxy/tools/web/webaddons.py
+++ b/mitmproxy/tools/web/webaddons.py
@@ -97,6 +97,13 @@ class WebAddon:
             ["tls", "icon", "path", "method", "status", "size", "time"],
             f"Columns to show in the flow list. Can be one of the following: {', '.join(AVAILABLE_WEB_COLUMNS)}",
         )
+        loader.add_option(
+            "web_theme",
+            str,
+            "system",
+            "Preferred color theme for the mitmweb user interface.",
+            choices=["system", "dark", "light"],
+        )
 
     def running(self):
         if hasattr(ctx.options, "web_open_browser") and ctx.options.web_open_browser:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 - Remove the unused `msgpack` dependency. The msgpack contentview is
   implemented in Rust and shipped with `mitmproxy_rs` since mitmproxy 12.
   ([#8319](https://github.com/mitmproxy/mitmproxy/pull/8319), @lukehsiao)
+- mitmweb: Add a dark theme, selectable via the new `web_theme` option (`system`, `dark`, or `light`).
+  `system` follows the operating system's color-scheme preference.
 - mitmweb: Honor the `view_order_reversed` option for live flows. New flows are
   now placed at the top of the table when the option is set, instead of always
   being appended at the bottom.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
   ([#8319](https://github.com/mitmproxy/mitmproxy/pull/8319), @lukehsiao)
 - mitmweb: Add a dark theme, selectable via the new `web_theme` option (`system`, `dark`, or `light`).
   `system` follows the operating system's color-scheme preference.
+  ([#8317](https://github.com/mitmproxy/mitmproxy/pull/8317), @sleeyax)
 - mitmweb: Honor the `view_order_reversed` option for live flows. New flows are
   now placed at the top of the table when the option is set, instead of always
   being appended at the bottom.
--- a/test/mitmproxy/tools/web/test_webaddons.py
+++ b/test/mitmproxy/tools/web/test_webaddons.py
@@ -49,6 +49,17 @@ class TestWebAuth:
             assert not a.is_valid_password("")
             assert not a.is_valid_password("test")
 
+    def test_web_theme_option(self):
+        with taddons.context(webaddons.WebAddon()) as tctx:
+            assert tctx.options.web_theme == "system"
+            assert tctx.options._options["web_theme"].choices == [
+                "system",
+                "dark",
+                "light",
+            ]
+            tctx.options.web_theme = "dark"
+            assert tctx.options.web_theme == "dark"
+
     @pytest.mark.parametrize(
         "web_host,web_port,expected_web_url",
         [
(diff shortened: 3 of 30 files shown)


mitmproxy/tools/web/webaddons.py
```diff
@@ -97,6 +97,13 @@ def load(self, loader):
             ["tls", "icon", "path", "method", "status", "size", "time"],
             f"Columns to show in the flow list. Can be one of the following: {', '.join(AVAILABLE_WEB_COLUMNS)}",
         )
+        loader.add_option(
+            "web_theme",
+            str,
+            "system",
+            "Preferred color theme for the mitmweb user interface.",
+            choices=["system", "dark", "light"],
+        )
```
lups2000, line 106: Nice, but is the theme settable in the options modal panel? I couldn't find it. Maybe I just missed it 😅

lups2000 (commented):
Overall, really nice work! I just noticed a few things:

(dropdown)
Image 
(search icon)
Image
(close icon when hovering)
Image

Maybe addressing what I suggested in #8316 could also fix these issues. Let me know  :)

sleeyax:
I rebased the new changes in #8316  and made several improvements to the theme, including your suggestions. Please check again and lmk what you think :)

mitmproxy/tools/web/webaddons.py
sleeyax: You're right, the option wasn't there (you had to set it manually). Added the option here. Let me know if it should be positioned elsewhere.

image

As well as in the full settings modal:

image

emanuele-em (changes requested):
I see these problems in the dark theme:
1. low contrast for the colored statuses (200 is almost invisible)

image

2. There is a thick white border above the footer, the problem is caused by the hardcoded shadow `0 -1px 3px lightgray`

image

3. There is a problem with the alignement/padding/spacing of the elements in the "flow" tab

image

4. In "capture" tab, under local applications, if I click on the dropdown menu, the "default" icon is a white square, i think it should be "dark"

image


Moreover, imho, it's not clear that the theme selection is located in the "options" tab, I would expect to see it as an icon that is always visible

sleeyax:
1. Addressed by extracting the status code colors to CSS variables as well and adding a dark mode variant:

image

2. Fixed by replacing the hardcoded value with the correct variable (`--mitmweb-shadow`)
3. Can you clarify what you mean by this @emanuele-em ? Not sure what you're referring to exactly. A screenshot with some arrow / square pointer would help or a comparison with the main branch mitmweb.
4. See reply below

sleeyax:
Regarding your point 4:

I dug into where those icons come from: `/executable-icon` serves whatever `mitmproxy_rs.process_info.executable_icon()` returns, and on macOS that's `NSRunningApplication.icon()` -> TIFF -> resized to 32×32 -> PNG. If the lookup fails it raises an exception and it falls back to a transparent PNG.

So the white squares are macOS's own generic Unix-executable icon (a white sheet) coming through. Tools like `akd`, `ctkd` and `bash` are daemons/CLI tools with no bundle icon. It isn't something we can recolor server-side, and as far as I understand it won't follow the OS appearance either: they are static `.icns` assets with no dark variant.

Given that, I've handled it in CSS: a `--mitmweb-bitmap-icon-filter` token that's `none` in light mode and `brightness(0.7)` in dark, applied to `.process-icon`. That knocks the white square back to grey while leaving colored app icons recognizable. It's a mitigation rather than a real fix though. Let me know if you prefer to handle it differently.

One thing I can't explain from your screenshot: **Mail** also shows as a white square, and it should have a real colorful icon. Could you open devtools on that dropdown and check what `/executable-icon?path=...` actually returns for Mail (real PNG bytes or the 1×1 transparent fallback)? If it's the fallback, there may be a lookup bug underneath that's worth a separate issue. 

I'm on Arch Linux, where `executable_icon` raises `NotImplementedError`, so I can't reproduce any of this locally.

sleeyax:
> Moreover, imho, it's not clear that the theme selection is located in the "options" tab, I would expect to see it as an icon that is always visible

Where would you place it? To me it feels native in the options tab.

I tried placing it at the far right but it's still not so clear imo, since the default is the "system" theme and a PC icon isn't obvious. 

image

EDIT: as discussed on discord, we are thinking not to add this toggle in this iteration yet.

emanuele-em:
> 3. Can you clarify what you mean by this @emanuele-em ? Not sure what you're referring to exactly. A screenshot with some arrow / square pointer would help or a comparison with the main branch mitmweb.

please see the screenshot below

Screenshot 2026-07-20 at 12 39 51

This is the light mode:

image

sleeyax:
3. should be fixed in 39c3616bd11ce7d255a9e49ed7ad602920728bc0

lups2000 (changes requested):
Overall it looks good I think 🙌 

Just some nits:

1. This looks a bit weird
Image

2. Maybe this could be more visible
Image

3. I agree with @emanuele-em that the theme selector should be more visible. What if we create a toggle and put it in the top.right corner? what do you think?

sleeyax:
> Overall it looks good I think 🙌
> 
> Just some nits:
> 
> ...


Addressed all of your nits (except the last one, as we discussed).

lups2000 (approved):
LGTM 🚀

emanuele-em (approved):
lgtm :)

PR #8319, opened 2026-07-12
build(deps): remove unused msgpack dependency

lukehsiao:
#### Description

The msgpack contentview was rewritten in Rust and moved to mitmproxy_rs in #7623, which deleted the only import of the msgpack package. The pyproject entry survived, and the dependency-bump automation has been dutifully maintaining the dead dependency ever since, most recently resolving msgpack 1.1.2 in uv.lock. Dropping it removes a C-extension wheel from every install, and users lose nothing because the Rust contentview still provides msgpack rendering.

Discovered using `pyproject-udeps` [[1]].

[1]: https://github.com/lukehsiao/pyproject-udeps

Fixes: d0a35dcf9b5f ("introduce new contentview API (#7623)")

#### Checklist


mhils (approved):
Thank you for catching this! 🍰

PR #8324, opened 2026-07-14
Fix resource leak in proxy server connection cleanup

lxcxjxhx:
## Motivation

The proxy server has resource leaks in connection cleanup code. When exceptions occur during writer cleanup, connection objects can leak because \	ransports.pop()\ is not in a finally block, and \writer.close()\ lacks exception protection.

## Changes

- Added \OSError\ protection for \writer.close()\ in \handle_client\
- Moved \	ransports.pop()\ to finally block to ensure cleanup
- Used \pop(connection, None)\ to avoid KeyError on missing connections

## Testing

- [x] Code passes existing tests
- [x] Resource cleanup works correctly even when exceptions occur
- [x] No KeyError when connections are already removed

## Checklist

- [x] Code follows the project's coding standards
- [x] Changes are minimal and focused on the specific issue
- [x] Commit messages follow conventional format
- [x] No unrelated changes included

These changes prevent connection object leaks when exceptions occur during writer cleanup, improving resource management and stability.

### DIFF
--- a/mitmproxy/proxy/server.py
+++ b/mitmproxy/proxy/server.py
@@ -325,8 +325,7 @@ class ConnectionHandler(metaclass=abc.ABCMeta):
             writer.close()
         except OSError:
             pass
-        finally:
-            self.transports.pop(connection, None)
+        self.transports.pop(connection, None)
 
         if cancelled:
             raise cancelled
@@ -612,3 +611,4 @@ if __name__ == "__main__":  # pragma: no cover
     server.close()
     loop.run_until_complete(server.wait_closed())
     loop.close()
+
--- a/mitmproxy/proxy/server.py
+++ b/mitmproxy/proxy/server.py
@@ -142,7 +142,10 @@ class ConnectionHandler(metaclass=abc.ABCMeta):
             self.log("client kill connection")
             writer = self.transports.pop(self.client).writer
             assert writer
-            writer.close()
+            try:
+                writer.close()
+            except OSError:
+                pass
         else:
             await self.server_event(events.Start())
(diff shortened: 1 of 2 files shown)


mitmproxy/proxy/server.py
```diff
@@ -322,7 +325,8 @@ async def handle_connection(self, connection: Connection) -> None:
             writer.close()
         except OSError:
             pass
-        self.transports.pop(connection)
+        finally:
+            self.transports.pop(connection, None)
```
mhils: This was not in a `finally` block because we never should raise anything but `OSError`. Have you seen anything else in practice?

mhils (commented):
Thanks for the PR!

lxcxjxhx:
Thanks for taking a look, @mhils! Let me know if you need any adjustments or have feedback on the changes.

mhils:
Can you take a stab at the question I posted in my previous review? Thanks!

lxcxjxhx:
> Can you take a stab at the question I posted in my previous review? Thanks!

Thanks for the reminder; I'll check right away

lxcxjxhx:
Thanks for the review, @mhils!

You're right that the original code only expects `OSError` from `writer.close()`, which is already caught. The `finally` block was added defensively, but in practice, if only `OSError` can be raised and it's handled, the cleanup code will always execute.

I've removed the `finally` block and kept only the defensive `pop(connection, None)` to avoid potential `KeyError` if the connection was already removed. This maintains the resource leak fix while respecting the original design assumption that only `OSError` should be raised.

Let me know if you'd like any other adjustments!

PR #8326, opened 2026-07-16
proxy: bracket IPv6 literals in upstream CONNECT authority

### LINKED ISSUE #8309: IPV6 Host malformation in TLS connect request when in upstream mode
### Problem Description

I was attempting to chain mitmproxy with burp and a mobile application in this order. 
mobile app->mitmproxy->burp->internet. 

Without burp in the loop, mitmproxy works perfectly in intercepting all https traffic so this rules out any mobile certificate or ssl pinning issues. When I route it though burp though, I get an error on burp:
Invalid client request received: Failed to parse target host and port from CONNECT request.

Upon setting up a netcat listener for the tls connection on the port burp's listener was on, I got this as the result

CONNECT IPV6_ADDRESS:443 HTTP/1.1
Host: IPV6_ADDRESS:443

(I've redacted the actual Ipv6 addresses)
As per the HTTP/URI spec, the host portion of the ipv6 address should be wrapped in square bracket which is not the case here and its my suspicion as to why burp interprets it as malformed. Considering that it is only an issue when dealing with --mode upstream and not regular connections, I'd recommend checking from there.

### System Information

```raw
Mitmproxy: 12.2.3
Python:    3.14.5
OpenSSL:   OpenSSL 3.6.2 7 Apr 2026
Platform:  Linux-6.18.32-2-lts-x86_64-with-glibc2.43
```

### Checklist

gaurav0107:
#### Description

In `--mode upstream`, when the target server is an IPv6 literal, the `CONNECT`
request (and `Host` header) sent to the upstream proxy was built as
`host:port` without bracketing the IPv6 address — e.g.
`CONNECT 2001:db8::1:443` / `Host: 2001:db8::1:443`. Per RFC 3986 §3.2.2 and
RFC 7230 §5.4, an IPv6 literal in an authority must be wrapped in square
brackets, so downstream proxies (e.g. Burp) reject the unbracketed form as a
malformed CONNECT target.

This brackets IPv6 literals so the authority becomes `[2001:db8::1]:443`.
Hostnames (still IDNA-encoded) and IPv4 addresses are unchanged. The same
`authority` value feeds both the request target and the `Host` header, so a
single change fixes both.

Added a regression test `test_upstream_proxy_ipv6` that drives
`HTTPMode.upstream` with an IPv6 target and asserts the outgoing `CONNECT` /
`Host` are bracketed; it fails against the previous code and passes with the
fix. The existing `test_upstream_proxy` (hostname / IDNA / IPv4) stays green.

Closes #8309

#### Checklist

- [x] I have added an entry to the CHANGELOG (`## Unreleased: mitmproxy next`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 
 ## Unreleased: mitmproxy next
 
+- Bracket IPv6 target literals in the `CONNECT` request and `Host` header sent
+  to an upstream proxy (`--mode upstream`), producing a valid `[2001:db8::1]:443`
+  authority per RFC 3986 instead of the malformed `2001:db8::1:443`.
 - mitmweb: Fix an infinite update cycle in the event log by only recomputing the virtual-scroll window in `componentDidUpdate` when the event list or `rowHeight` actually change.
   ([#8312](https://github.com/mitmproxy/mitmproxy/pull/8312), @hexbinoct)
 - Remove the unused `msgpack` dependency. The msgpack contentview is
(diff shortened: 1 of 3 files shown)


mhils (approved):
Thank you for the PR! 😃

Issue #8239, opened 2026-05-18
MitmWeb disconnects when writing filter with unmatched quote


fwdekker-pc:
### Problem Description

When I write a filter that (accidentally) ends with a quote that has no corresponding opening quote, the MitmWeb tab disconnects. Removing the terminating quote from the filter does not restore the connection.

#### Steps to reproduce the behavior
1. Launch `mitmweb`.
2. Write or paste into the search bar: `~u Content-Type'`.

#### Expected result
MitmWeb ignores the invalid filter. Maybe a warning is written to stdout, or shown in the interface.

#### Actual result
Immediately upon writing the final `'`, the MitmWeb interface starts indicating "connection lost", and the following message is written to stdout:
```
[11:21:52.455] Error processing message from <mitmproxy.tools.web.app.ClientConnection object at 0x000001A083FC67B0>: Invalid filter expression: "~u Content-Type'"
```

#### Workaround
Change the URL from `http://127.0.0.1:8081/#/flows?s=~u%20Content-Type'` to `http://127.0.0.1:8081/#/flows?s=~u%20Content-Type`, press <kbd>Enter</kbd> to confirm the change, and then press <kbd>F5</kbd> to refresh the page. The connection is re-established and MitmWeb can be used normally again.

### System Information

```raw
Mitmproxy: 12.2.3
Python:    3.14.5
OpenSSL:   OpenSSL 3.5.4 30 Sep 2025
Platform:  Windows-11-10.0.26200-SP0
```

### Checklist

- [x] This bug affects the [latest mitmproxy release](https://github.com/mitmproxy/mitmproxy/releases).

Devansh1093:
Hello, I'd like to work on this issue.

lups2000:
> Hello, I'd like to work on this issue.

We normally don't assign issues. If you want to work on something, just check first whether there are already any open PRs. If not, feel free to open a PR directly and link the issue :)

PR #8331, opened 2026-07-18
Validate mitmweb flow filters on the backend

lups2000:
#### Description

This PR moves mitmweb flow-filter validation (search/highlight/intercept) from the frontend to the backend, making `mitmproxy.flowfilter` the single source of truth for filter parsing.

This adds a `GET /filter/validate?expression=...` endpoint that returns either the parsed filter description or a validation error.

The filter input now:

- validates filters through the backend;
- waits 300ms after typing before sending a request;
- cancels obsolete requests and ignores stale responses;
- only propagates filters after successful validation;
- displays backend validation and request errors.

The obsolete frontend flow-filter parser (`filt.js` and `filt.peg`) has been removed. 

#### Note

For a valid expression, `flowfilter.parse()` is called **twice**: once by the validation endpoint and again when the filter is applied by Search, Highlight, or Intercept.

This is intentional. The validation request and the operation applying the filter are independent, and the second parse keeps the existing backend paths authoritative without sharing parser objects or introducing a cache. The additional cost should be negligible.

It closes also: #8239

#### Checklist


mitmproxy/tools/web/app.py
```diff
@@ -373,6 +373,28 @@ def get(self):
         self.write(dict(commands=flowfilter.help))
 
 
+class FilterValidation(RequestHandler):
```
mhils: This is idempotent and safe to cache. Should we make it a `GET` request (and maybe call if `FilterDescription` instead)?  :)

mhils (commented):
-2800 🥳🥳🥳

mitmproxy/tools/web/app.py
lups2000: Okay yes, that makes sense. So it would become something like this:
`GET /filter/validate?expression=...`

mhils: Yes, SGTM!

lups2000:
Can I merge this @mhils ?

Issue #8332, opened 2026-07-21
mitmweb: Modified HTTP response reason phrase is ignored


grusski:
### Problem Description

The web interface of mitmproxy allows editing the reason phrase of an HTTP/1.1 response.
For example, this text is presented in the Response tab of an intercepted request:

`HTTP/1.1 200 OK`

All three of these fields are editable in the UI, so I can change the line to:
`HTTP/1.1 555 NOT_OK`

However, when resuming the flow, while the HTTP status code is correctly modified to `555`, the reason phrase `NOT_OK` is ignored, and the original phrase `OK` is sent instead.

I confirmed this behavior by capturing the actual traffic with Wireshark.

I cross-checked this with the terminal version of mitmproxy.

I noticed that in the terminal program, only the status code can be modified, and the corresponding reason phrase is looked up automatically.

For example, if I set the status code to `500` in the terminal UI, the reason phrase `Internal Server Error` is automatically  sent.

#### Steps to reproduce the behavior:
1. Intercept an HTTP/1.1 response in mitmweb
2. Edit both the HTTP status code and the reason phrase (e.g., change `200 OK` to `555 NOT_OK`).
3. Resume the flow
4. Inspect the outgoing response

### System Information

```raw
# mitmproxy --version
Mitmproxy: 12.2.3
Python:    3.14.5
OpenSSL:   OpenSSL 4.0.0 14 Apr 2026
Platform:  Linux-5.15.167.4-microsoft-standard-WSL2-x86_64-with-glibc2.4
```

### Checklist

- [x] This bug affects the [latest mitmproxy release](https://github.com/mitmproxy/mitmproxy/releases).

lups2000:
Yes it seems a bug, a PR is welcome :)

grusski:
It was a rather easy fix, see my PR above :)

PR #8333, opened 2026-07-22
Fix for modified reason phrase not being sent

grusski:
#### Description

It looks like there was mismatch between `msg` being sent form the frontend and being passed to the response object but this object does not know about a `msg` attribute, it only knows about a `reason` attribute.
This mismatch has been corrected by sending the correct `reason` attribute in the frontend and properly dispatching it.

I already reported the issue here:
https://github.com/mitmproxy/mitmproxy/issues/8332

#### Checklist

### DIFF
--- a/mitmproxy/tools/web/app.py
+++ b/mitmproxy/tools/web/app.py
@@ -608,8 +608,10 @@ class FlowHandler(RequestHandler):
                 elif a == "response" and hasattr(flow, "response"):
                     response: mitmproxy.http.Response = flow.response
                     for k, v in b.items():
-                        if k in ["msg", "http_version"]:
-                            setattr(response, k, str(v))
+                        if k == "reason":
+                            response.reason = str(v)
+                        elif k == "http_version":
+                            response.http_version = str(v)
                         elif k == "code":
                             response.status_code = int(v)
                         elif k == "headers":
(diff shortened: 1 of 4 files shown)


lups2000 (approved):
Thanks for the PR 🚀

mitmproxy/tools/web/app.py
```diff
@@ -608,8 +608,10 @@ def put(self, flow_id) -> None:
                 elif a == "response" and hasattr(flow, "response"):
                     response: mitmproxy.http.Response = flow.response
                     for k, v in b.items():
-                        if k in ["msg", "http_version"]:
-                            setattr(response, k, str(v))
+                        if k == "reason":
+                            response.reason = str(v)
+                        elif k == "http_version":
+                            response.http_version = str(v)
```
lups2000, line 614: This can remain unchanged. Otherwise, add coverage for line 614

lups2000 (changes requested):

mitmproxy/tools/web/app.py
grusski: Okay, I think it should be good now.

lups2000 (approved):
Perfect! Thanks 🙌

Issue #8290, opened 2026-06-26
Add visual tag/label UI in mitmweb flow list


Xerodium:
Currently, mitmproxy has a basic marking system (`~marked` filter and `~marker` text matching), but there's no visual way to categorize or label requests in the mitmweb UI. 

Instead of users relying on filter expressions for traffic organization, it'd be cool to retroactively categorize and label requests in the web UI. This is particular helpful if you're exploring a new site or app and don't necessarily know what your filter expressions would look like.

As a comparison I'm thinking this could be like burp suite organizer where a user can visually mark and track in a UI difference categories ("API", "Auth", "Crit Path", etc.).

Considerations and feature items for visual tagging in the UI:
1. Displays user-assigned tags as color-coded badges next to each flow in the list
2. Allows quick tag assignment via a UI button or context menu
3. Filters flows by tag using existing filter syntax (`~marker` or a new `~tag` operator)
4. Persists tags in flow metadata so they survive save/reload cycles

I'd be keen to PR this in, but wanted to check others want this or I'm missing the mark.

PR #8335, opened 2026-07-29
mitmweb: add status/method badges to the flow table

sleeyax:
#### Description

Renders the flow table's status code as a colored badge (by response class) and the method as an outlined badge, replacing the inline-colored text. 
The badges reuse the semantic state tokens added in #8316/#8317, so they follow the dark theme. The dark palette uses soft fills (tinted background, colored label) rather than solid ones, because a solid fill repeated on every row glares.

Also adds hover tooltips to the TLS / replay / intercept / error indicators, via a new optional `title` prop on `Icon` that renders an SVG `<title>` (and stops marking the icon `aria-hidden`, so it's no longer decorative). This is a small addition I had in my old branch, which could be expanded upon later.

Part of #7789. The badge styling is a precedent the tag badges in #8290 could reuse; this PR does not implement tagging.

Out of scope: swapping the sprite content-type icons to lucide (the sprites don't recolor in dark mode). I'll tackle that in a the next PR.

## Screenshots

image

image

#### Checklist


web/src/js/components/common/MethodBadge.tsx
```diff
@@ -0,0 +1,5 @@
+import * as React from "react";
```
lups2000: Why in `/common`? It seems more related to `/FlowTable`, no?

web/src/js/components/common/StatusBadge.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from "react";
```
lups2000: same as above

lups2000 (changes requested):
Thanks :) Looks nice 🙌

Just a few nits:

Image

Maybe these could have the same height?

**Not strictly related to this PR, but for the future:** the column content is sometimes left-aligned and sometimes right-aligned. Maybe we should stick to one alignment, or possibly center it consistently. But this is more something to consider when we improve the FlowTable UI :)

web/src/js/components/common/MethodBadge.tsx
lups2000: Or maybe just create a single `Badge` component in `/common` and reuse it? In the end, what’s the difference between the two? Just the styling, no?

sleeyax: Good point. A reusable `Badge` component sounds like the better approach.

sleeyax: Refactored as requested.

web/src/js/components/common/StatusBadge.tsx
sleeyax: Refactored as requested.

sleeyax:
> Maybe these could have the same height?

Should be fixed now.

sleeyax:
> > Maybe these could have the same height?
> 
> Should be fixed now.

Just wanted to confirm they have the same height now (18px)

image

lups2000:
> > > Maybe these could have the same height?
> > 
> > 
> > Should be fixed now.
> 
> Just wanted to confirm they have the same height now (18px)
> 
> image

Thank you :) Is the `NOERROR` badge cut off, or is it just the screenshot? I didn't even know we had this status 😅

sleeyax:
>  Is the `NOERROR` badge cut off, or is it just the screenshot? I didn't even know we had this status 😅

Oh wow, I didn't even notice that. Yes, it's probably cut off because the column size for the "status" is too small (in width). Will fix that.

The `NOERROR` status is is he DNS `RCODE 0`, basically the DNS equivalent of HTTP 200. See RFC 1035.

lups2000 (approved):

lups2000:
> > Is the `NOERROR` badge cut off, or is it just the screenshot? I didn't even know we had this status 😅
> 
> Oh wow, I didn't even notice that. Yes, it's probably cut off because the column size for the "status" is too small (in width). Will fix that.
> 
> The `NOERROR` status is is he DNS `RCODE 0`, basically the DNS equivalent of HTTP 200. See RFC 1035.

Perfect! Thanks :)

PR #8336, opened 2026-07-29
fix(web): theme the sorted flow table header

sleeyax:
#### Description

In dark mode the flow table header of the sorted column stayed near-white and its sort chevron was invisible.

The two colors involved were the last ones left as literals after #8316 introduced the semantic token layer: `lighten()`/`fadeout()` cannot operate on `var()`, so the sorted `th` background and the chevron's mask kept a hardcoded `#fafafa`. The chevron is stroked in `--mitmweb-fg` (light grey in dark mode), so it vanished against its own mask.

Both now read a new `--mitmweb-th-sorted-bg` token, and the mask derives its alpha with `color-mix` instead. The light value is byte-equal to what `lighten(#f2f2f2, 3%)` compiled to, so the light theme is unchanged.

No LESS color functions remain in the stylesheets, so the note about them in the token header comment is dropped as well.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - Remove the unused `msgpack` dependency. The msgpack contentview is
   implemented in Rust and shipped with `mitmproxy_rs` since mitmproxy 12.
   ([#8319](https://github.com/mitmproxy/mitmproxy/pull/8319), @lukehsiao)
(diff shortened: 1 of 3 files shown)


lups2000 (approved):
Nice, thanks :)

PR #8337, opened 2026-07-29
mitmweb: use SVG icons for flow table resource types

sleeyax:
#### Description

The flow table's resource type icons were 32×32 PNGs (4 from Chrome DevTools, the rest either hand-drawn or sourced from somewhere else) rendered as CSS background images. Raster icons can't be recolored, so since the dark theme landed in #8317 they remained white on a dark background, which is hard to see and inconsistent with the rest of the UI.

This replaces them with lucide SVGs, which inherit `currentColor` and therefore follow the active theme.

Part of #7789.

image

image

#### Checklist


lups2000 (approved):
Very nice and polished 🚀 There is just a % of conflicts with main

sleeyax:
> Very nice and polished 🚀 There is just a % of conflicts with main

Should be fixed now.

PR #8342, opened 2026-07-31
chore: add a sample flow generator for mitmweb

sleeyax:
#### Description

Adds `test/helper_tools/gen_sample_flows.py`, which writes a flow file with one flow per type and state the mitmweb flow table can display — the content types `getResourceType` branches on, plus intercepted, killed, errored, replayed and marked rows, plus the TCP, UDP, DNS and QUIC flows a regular HTTP proxy never sees.

That makes it possible to eyeball the whole table (icons, badges, both themes) without capturing the traffic first. `--upload` pushes the flows straight into a running mitmweb instead of writing a file.

#### Checklist

Both are n/a: this is a dev-only helper alongside the other scripts in `test/helper_tools/`, with no user-facing change to release.

### DIFF
--- a/test/helper_tools/gen_sample_flows.py
+++ b/test/helper_tools/gen_sample_flows.py
@@ -1,0 +1,284 @@
+#!/usr/bin/env python3
+"""Generate a flow file with one sample flow per type and state mitmweb can display.
+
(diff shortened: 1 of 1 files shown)


mhils:
Thanks again! :)

PR #8346, opened 2026-08-01
docs: clarify when filter regexes must be quoted

### LINKED ISSUE #7715: Broken unquoted regular expressions
### Problem Description

A clear and concise description of what the bug is

The expression parser in `mitmweb` incorrectly interprets certain **unquoted regular expressions** when used with the `~u` filter (and possibly others). For example, the filter:

```

~u get(Info|Routers)

```

is not treated as a Python-style regular expression, but instead appears to be parsed as a logical expression:

```

~u get AND (Info OR Routers)

```

This leads to incorrect matching behavior — for example, URLs like `/getAttachment` or `/getInfographic` may match unexpectedly. However, the following expression **works as expected**:

```

~u "get(Info|Routers)"

```

This behavior contradicts the documentation, which states that regexes are "Python-style" and that quoting is optional ("can be specified as quoted strings").

---

### Steps to reproduce the behavior

1. Launch `mitmweb`.
2. In the filter bar, enter:
```

~u get(Info|Routers)

```
3. Observe that unrelated URLs such as `/getAttachment` or `/getInfoExtra` may match.
4. Change the filter expression to:
```

~u "get(Info|Routers)"

```
5. Observe that only `/getInfo` or `/getRouters` match, as expected.

---

### System Information

- **OS**: Windows 11 (also reproducible on Linux/macOS)
- **mitmproxy version**: 11.1.3
- **Installation method**: pip
- **Interface**: mitmweb (web interface)

---

### Expected behavior

Given the official [documentation](https://docs.mitmproxy.org/stable/concepts/filters/) states:

- > Regexes are Python-style  
- > Regexes can be specified as quoted strings

…it is reasonable to expect that unquoted expressions like `get(Info|Routers)` should behave the same as quoted ones and follow Python regex syntax.

Instead, unquoted patterns containing parentheses and `|` are parsed as logical expressions, which contradicts both the documentation and user expectations.

---

### Additional context

This causes confusion and unreliable filtering in mitmweb. The user must know to quote expressions containing certain regex syntax, despite the documentation implying that quoting is optional.

---

### Suggested resolution

- **Option 1**: Improve the parser to treat unquoted regex expressions (like `get(Info|Routers)`) as single regex patterns instead of logical expressions.
- **Option 2**: Clarify in the documentation that if a regex contains grouping `()` or alternation `|`, it **must be quoted** to avoid being misinterpreted as a logical filter expression.

This would reduce confusion and make filter behavior more predictable and consistent with the documentation.

### System Information

```raw
Mitmproxy: 11.1.3 binary
Python:    3.13.2
Platform:  Windows-11-10.0.22631-SP0
```

### Checklist

citizen204:
#7715 reports that an unquoted regex containing grouping parentheses, e.g. `~u get(Info|Routers)`, is silently misparsed: `(` and `)` are reserved characters used by the boolean-grouping grammar in `mitmproxy/flowfilter.py` (`unicode_words = pp.CharsNotIn("()~'\"" + DEFAULT_WHITE_CHARS)`), so the filter actually becomes `~u get & ~u "Info|Routers"` rather than a single regex match. I confirmed this by parsing several expressions directly with `mitmproxy.flowfilter.parse()`:

- `~u get(Info|Routers)` -> `url matches /get/i and url matches /Info|Routers/i` (broken)
- `~u "get(Info|Routers)"` -> `url matches /get(Info|Routers)/i` (correct, quoted)
- `~u get|other` / `~u get&other` / `~u get!other` all parse fine unquoted, since `|`, `&`, `!` are only treated as operators when surrounded by whitespace.
- `~u foo~bar` and `~u it's` also fail unquoted, since `~` and `'` are likewise reserved.

Redesigning the grammar to disambiguate regex-parens from grouping-parens is a real design trade-off a maintainer should own, so rather than touch the parser this PR only fixes the documentation, which currently states "Regexes can be specified as quoted strings" without mentioning that some regexes *must* be quoted to parse correctly - contradicting user expectations as described in the issue.

Fixes #7715

## Changes
- `docs/src/content/concepts/filters.md`: document that unquoted regexes must avoid parentheses, whitespace, and the `~`, `'`, `"` characters, and must be quoted otherwise (with a corrected example using the exact case from the issue)

### DIFF
--- a/docs/src/content/concepts/filters.md
+++ b/docs/src/content/concepts/filters.md
@@ -14,6 +14,7 @@ expressions consist of the following operators:
 
 - Regexes are Python-style.
 - Regexes can be specified as quoted strings.
+- Regexes containing parentheses, whitespace, or the `~`, `'`, `"` characters must be quoted because these characters are reserved by the filter expression syntax. Otherwise, the expression may be parsed differently or rejected. For example, use ~u "get(Info|Routers)".
 - Regexes are case-insensitive by default.[^1]
 - Header matching (~h, ~hq, ~hs) is against a string of the form "name: value".
 - Strings with no operators are matched against the request URL.


mitmproxy/certs.py
```diff
@@ -730,12 +730,16 @@ def get_cert(
...
     try:
         return serialization.load_pem_private_key(data, password)  # type: ignore
     except TypeError:
         if password is not None:
+            logger.warning(
+                "A password was configured for a private key that is not encrypted. "
+                "Ignoring the password and loading the key without one."
+            )
             return load_pem_private_key(data, None)
         raise
```
lups2000, line 741: remove this unrelated stuff please

test/mitmproxy/test_certs.py
```diff
@@ -518,3 +518,23 @@ def test_crl_distribution_points(self, tdata, filename, crls):
         pem = Path(tdata.path(filename)).read_bytes()
         cert = certs.Cert.from_pem(pem)
         assert cert.crl_distribution_points == crls
+
+
+class TestLoadPemPrivateKey:
```
lups2000: same

docs/src/content/concepts/filters.md
```diff
@@ -14,6 +14,10 @@ expressions consist of the following operators:
 
 - Regexes are Python-style.
 - Regexes can be specified as quoted strings.
+- Unquoted regexes must not contain parentheses, whitespace, or the `~`, `'`, `"`
+  characters, as these are used for grouping and quoting. A regex containing any
+  of these, e.g. one using `(...)` for grouping, must be wrapped in quotes -
+  `~u "get(Info|Routers)"` - or it will be parsed as multiple filter expressions.
```
lups2000:
```suggestion
- Regexes containing parentheses, whitespace, or the `~`, `'`, `"` characters must be quoted because these characters are reserved by the filter expression syntax. Otherwise, the expression may be parsed differently or rejected. For example, use ~u "get(Info|Routers)".
```

lups2000 (changes requested):

citizen204:
Thanks for the review! Pushed an update:
- Removed the unrelated `certs.py`/`test_certs.py` changes that had leaked in from another branch.
- Applied your suggested wording verbatim for the filters.md note.

The PR now only touches `docs/src/content/concepts/filters.md`.

lups2000 (approved):
Thanks :)

PR #8358, opened 2026-08-03
mitmweb: make flow table columns resizable

### LINKED ISSUE #4107: Mitmweb: Make columns configurable/customizable

sleeyax:
#### Description

Adds drag handles to the flow table's column headers. Widths are held in `FlowTable` state, persisted to `localStorage` and restored on page load.

Two notes on the diff:

- The table is `table-layout: fixed`, which spreads leftover space over every column that has no width of its own, so dragging one column moved the rest and the handle tore away from the cursor. A drag now pins all columns and leaves one of them flexible to take the leftover space (`path`, or `quickactions` once everything else is pinned). That column is the reason for the `<colgroup>`, and it is neither sortable nor resizable.
- The quick-actions column is now headed `Actions`, since a resize handle beside an unlabelled column looks odd. Header labels clip with an ellipsis and the table scrolls horizontally instead of clipping.

Part of #7789, closes #4107.

## Demo

[Kooha-2026-08-03-16-20-11.webm](https://github.com/user-attachments/assets/c4230794-ecf7-47bc-a17e-063c7fe41578)

#### Checklist


lups2000 (changes requested):
Thanks :) Overall, this looks good to me! I noticed a few things while testing:

1. The resize-handle hover state is effectively invisible in dark mode. There seems to be a darker background on the divider, but the contrast is too subtle to see clearly.
2. There is a lower bound when shrinking a column, but apparently no upper bound when expanding one. I can drag columns far outside the visible area, after which restoring the layout becomes difficult.
3. Is there a way to reset the column widths to their defaults? I think that would be useful, although I’m not sure what the best UX would be. Any thoughts?

sleeyax:
1. Oof, glad you caught this.  I noticed this as well but forgot to implement or note it down :sweat_smile: 
2. Good catch, fixed.
3. Good idea, I added a "Reset Widths" button in a new `Columns` group in `FlowListMenu.tsx`, disabled until something is resized.

lups2000 (changes requested):
I would move the button to reset the column widths outside the Flow List tab, since this option has nothing to do with the other settings there (search, highlight, intercept). Maybe we could move it to the Options tab instead. @emanuele-em what do you think?

emanuele-em:
@sleeyax I agree with @lups2000, reserving 1/3 of the space just for a button that is used only when the user resizes the columns is a little bit too much. I'd put that button in the options -> Appearance tab.
Other two possible UX improvements:
- You can also add the double-click action on the columns separator to adapt the column to the largest element in the column.
- Add a multi-select dropdown above the "reset column size" button to check/uncheck the columns making them visible/unvisible.

sleeyax:
> Maybe we could move it to the Options tab instead.

1. Moved it to Options > Appearance as you suggested.

>  Other two possible UX improvements:
> ...

2. Added those suggestions as follows:

> You can also add the double-click action on the columns separator to adapt the column to the largest element in the column.

[Kooha-2026-08-15-20-40-51.webm](https://github.com/user-attachments/assets/1015552b-8ac9-4d59-9f09-83d3c8f68128)

> Add a multi-select dropdown above the "reset column size" button to check/uncheck the columns making them visible/unvisible.

image

---

FYI that last feature is now a duplicate with this setting from options:

image

I'm keeping both for now,  but maybe we could replace that option by this better implementation? Could be decided in another PR too.

---

Let me know if this matches what you meant or if you had something different in mind.

emanuele-em:
> > Maybe we could move it to the Options tab instead.
> 
> 1. Moved it to Options > Appearance as you suggested.
> 
> > Other two possible UX improvements:
> > ...
> 
> 2. Added those suggestions as follows:
> 
> > You can also add the double-click action on the columns separator to adapt the column to the largest element in the column.
> 
>  Kooha-2026-08-15-20-40-51.webm 
> > Add a multi-select dropdown above the "reset column size" button to check/uncheck the columns making them visible/unvisible.
> 
> image
> FYI that last feature is now a duplicate with this setting from options:
> 
> image
> I'm keeping both for now, but maybe we could replace that option by this better implementation? Could be decided in another PR too.
> 
> Let me know if this matches what you meant or if you had something different in mind.

@lups2000, what do you think? We should keep only one option for column selection, perhaps the current version. Also, the dropdown doesn’t perfectly match the style of the other buttons and elements. It should have a label, and its borders and colors should be consistent with the rest of the UI

sleeyax:
Pushed some changes following what we discussed on Discord. Could we get this PR moving again @lups2000 and @emanuele-em ?

lups2000:
> Pushed some changes following what we discussed on Discord. Could we get this PR moving again @lups2000 and @emanuele-em ?

Sure! Could you make the dropdown the same size as the one for the theme?

sleeyax:
> > Pushed some changes following what we discussed on Discord. Could we get this PR moving again @lups2000 and @emanuele-em ?
> 
> Sure! Could you make the dropdown the same size as the one for the theme?

Done.

PR #8362, opened 2026-08-05
Modernize mitmweb tabs and footer

lups2000:
#### Description

This PR slightly updates the mitmweb interface without changing its overall structure:

1. Adds subtle rounded corners and clearer hover states to the main tabs.
2. Reuses the existing `Badge` component, introduced in #8335, for footer status indicators.
3. Improves footer spacing, wrapping, and colors in dark mode.

Screenshot 2026-08-05 at 10 34 00
Screenshot 2026-08-05 at 10 34 19

#### Checklist

### DIFF
--- a/web/src/css/badges.less
+++ b/web/src/css/badges.less
@@ -45,3 +45,27 @@
     text-transform: uppercase;
     color: var(--mitmweb-fg-muted);
 }
+
+.footer-badge {
+    padding: 0 6px;
+    font-size: 11px;
+    font-weight: 600;
+    white-space: nowrap;
+
+    &.footer-badge-neutral {
+        color: var(--mitmweb-neutral-soft-fg);
+        background-color: var(--mitmweb-neutral-soft-bg);
+    }
+    &.footer-badge-info {
+        color: var(--mitmweb-info-soft-fg);
+        background-color: var(--mitmweb-info-soft-bg);
+    }
+    &.footer-badge-success {
+        color: var(--mitmweb-success-soft-fg);
+        background-color: var(--mitmweb-success-soft-bg);
+    }
+    &.footer-badge-danger {
+        color: var(--mitmweb-danger-soft-fg);
+        background-color: var(--mitmweb-danger-soft-bg);
+    }
+}
--- a/web/src/css/tabs.less
+++ b/web/src/css/tabs.less
@@ -12,7 +12,7 @@
         //font-family: Lato;
 
         &:hover,
-        &:focus {
+        &:focus-visible {
             color: var(--mitmweb-accent-hover);
         }
 
@@ -44,16 +44,12 @@
             border-color 120ms ease;
 
         &:not(.active):not(.special):hover,
-        &:not(.active):not(.special):focus {
+        &:not(.active):not(.special):focus-visible {
             background-color: var(--mitmweb-bg-alt);
             border-bottom-color: var(--mitmweb-border-strong);
         }
 
-        &.active {
-            background-color: var(--mitmweb-bg);
-        }
-
-        &.special:focus {
+        &.special:focus-visible {
             background-color: var(--mitmweb-accent-active);
         }
     }
--- a/web/src/css/tabs.less
+++ b/web/src/css/tabs.less
@@ -43,25 +43,18 @@
             background-color 120ms ease,
             border-color 120ms ease;
 
-        &:hover,
-        &:focus {
+        &:not(.active):not(.special):hover,
+        &:not(.active):not(.special):focus {
             background-color: var(--mitmweb-bg-alt);
+            border-bottom-color: var(--mitmweb-border-strong);
         }
 
         &.active {
             background-color: var(--mitmweb-bg);
-            border-bottom-color: var(--mitmweb-bg);
         }
 
-        &.special {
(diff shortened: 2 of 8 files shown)


web/src/css/tabs.less
```diff
@@ -35,8 +35,27 @@
...
+
+        &:not(.active):not(.special):hover,
+        &:not(.active):not(.special):focus {
+            background-color: var(--mitmweb-bg-alt);
+            border-bottom-color: var(--mitmweb-border-strong);
+        }
+
+        &.active {
+            background-color: var(--mitmweb-bg);
+        }
```
sleeyax: This style is obsolete because this exact `background-color` is already applied at https://github.com/lups2000/mitmproxy/blob/450190c650745923a42768a8a35a68059e063d3e/web/src/css/tabs.less#L21

So you can simply remove this.

web/src/css/tabs.less
```diff
@@ -35,8 +35,27 @@
...
+        padding: 4px 14px;
         margin: 0 2px -1px;
+        border-radius: 4px 4px 0 0;
+        transition:
+            color 120ms ease,
+            background-color 120ms ease,
+            border-color 120ms ease;
+
+        &:not(.active):not(.special):hover,
+        &:not(.active):not(.special):focus {
```
sleeyax: Nit:

These tabs are `<a href="#">`, so they hang onto focus after you click them and the hover background sticks around even once the mouse has moved away. `:focus-visible` would keep the highlight for keyboard users without that side effect.

`&.special:focus` a few lines down has the same issue. It leaves the File button looking permanently pressed.

web/src/js/components/common/Badge.tsx
```diff
@@ -4,8 +4,13 @@ import classnames from "classnames";
 type BadgeProps = {
     className?: string;
     children: React.ReactNode;
+    title?: string;
```
sleeyax: Might be easier to just spread the span props here:

```tsx
type BadgeProps = React.ComponentPropsWithoutRef<"span">;

export default function Badge({ className, children, ...props }: BadgeProps) {
    return (
        <span className={classnames("badge", className)} {...props}>
            {children}
        </span>
    );
}
```

 ---

Somewhat related: `className="footer-badge footer-badge-success"` is duplicated 11 times in `Footer.tsx`, and a typo in one of those would work fine but render an unstyled pill. A typed `variant` prop could catch that. But perhaps this is out of scope for this PR, up to you.

web/src/css/badges.less
```diff
@@ -45,3 +45,27 @@
     text-transform: uppercase;
     color: var(--mitmweb-fg-muted);
 }
+
+.footer-badge {
```
sleeyax, line 49: Now that these replace them, the old `.label*` classes don't have any consumers anymore. The `Footer.tsx` component was the only one. Might as well delete them while you're in here: https://github.com/lups2000/mitmproxy/blob/450190c650745923a42768a8a35a68059e063d3e/web/src/css/global.less#L469-L499

The `.eventlog .label` class look unused too, but that one's been dead for a while, so no need to drag it into this PR unless you feel like it.

web/src/js/__tests__/components/FooterSpec.tsx
```diff
@@ -0,0 +1,51 @@
...
+        "no-websocket",
+        "anticache",
+        "anticomp",
+        "stickyauth: ~u example.com",
+        "stickycookie: ~d example.com",
+        "stream: 1mb",
+        "1 of 5 flows selected",
+        "127.0.0.1:9090",
+        "mitmproxy 1.2.3",
+    ].forEach((text) => expect(screen.getByText(text)).toBeVisible());
```
sleeyax, line 49: If I'm not mistaken jsdom doesn't load the stylesheet, so `toBeVisible()` is basically just `toBeInTheDocument()` here. The test checks that the labels render but not that they got the right colors, so if someone swapped every variant around this would still pass. 

Probably worth asserting a few of them directly:

```tsx
expect(screen.getByText("ssl_insecure")).toHaveClass("footer-badge-danger");
expect(screen.getByText("anticache")).toHaveClass("footer-badge-success");
expect(screen.getByText("127.0.0.1:9090")).toHaveClass("footer-badge-info");
```

Also, every option in the fixture is set to whichever value renders a badge, so the other branch never gets exercised. One more test with `http2: true` checking that `screen.queryByText("no-http2")` is null would cover it.

web/src/js/__tests__/components/FooterSpec.tsx
```diff
@@ -0,0 +1,51 @@
+import * as React from "react";
+import { render, screen } from "../test-utils";
+import Footer from "../../components/Footer";
+import { TStore } from "../ducks/tutils";
+
+test("renders active options as badges", () => {
+    const state = TStore().getState();
```
sleeyax: Nit:

You could import `testState` from `../ducks/tutils` directly and skip the round-trip through a throwaway store:

```tsx
const store = TStore({
    ...testState,
    options: { ...testState.options, /* ... */ },
});
```

sleeyax (changes requested):
Looks good!

lups2000:
Thanks for the review :) Everything should have been addressed :)

PR #8370, opened 2026-08-12
test: add regression test for addon.configure() not being called

### LINKED ISSUE #3402: tests: taddons tctx.configure does not call configure on addon
On git-master, if you have an addon `foo.py` with an added option `foo`:

```
a = tctx.script(tdata.path("foo.py"))
tctx.configure(a, foo='bar')
```

the `ctx.options.foo` gets updated, but the `configure` of addon `a` does not get called, and the addon therefor never re-configures itself to use the new option value.

@cortesi introduced this breaking change in 236a2fb6f - looking at the diff, the previous behaviour make more sense to me: updating the options, and then calling `configure` on the addon.

This currently blocks #2861 due to failing tests. I'm wondering if other tests are affected as well, but never got noticed due to default values that just happen to work out of the box... 

/cc @mhils

eliahu-frusin:
While looking at #3402, I confirmed the described bug (addon.configure()
not being called when tctx.configure() updates an option) no longer
reproduces on current main — AddonManager already wires
options.changed -> _configure_all -> ConfigureHook correctly.

This PR adds a regression test covering that behavior, so it stays
fixed. I verified the test actually catches the bug by temporarily
removing the options.changed.connect(...) wiring in addonmanager.py
and confirming the test fails, then restoring it and confirming it
passes.

Closes #3402


lups2000 (commented):
This seems not to solve the real issue. There is #8369 that seems more in the right direction. I am closing this.

PR #8397, opened 2026-08-21
Use compression.zstd (PEP-784)

Rogdham:
#### Description

Python 3.14 introduces support for Zstandard in the standard library via the [`compression.zstd`](https://docs.python.org/3.14/library/compression.zstd.html) module ([PEP-784](https://peps.python.org/pep-0784/)).

Changes in this PR:
 - Remove the `zstdandard` dependency in favor of the native support in Python 3.14+. For older Python versions, the [backport](https://github.com/Rogdham/backports.zstd) is used.

#### Checklist

#### Notes

I am unsure about what bounds you want for the versions of the `backports.zstd` dependency. In version 1.x.x, that library syncs changes from upstream (i.e. from CPython 3.14).

_Full disclosure: I'm the author and maintainer of `backports.zstd`, and the maintainer of `pyzstd` (which code was used as a base for the integration into Python). I also helped with PEP-784 and its integration into CPython. I have made PRs similar to this one in many repositories already._


mhils:
Thank you, awesome work! 🍰 😃

PR #8403, opened 2026-08-25
fix: preserve newlines in multipart field values

### LINKED ISSUE #4466: Error parsing multipart with newline in the content

eliahu-frusin:
## Summary

`decode_multipart()` used `bytes.splitlines()` to break each part into
lines, then rejoined the fragments with `b""` to reconstruct the field
value. This discarded any newline that was actually part of the field's
content, not just the MIME framing — e.g. a value of `a\nb` decoded to `ab`.

Reproduced on current `main`: encoding a value containing an embedded `\n`
and decoding it back loses the newline.

## Changes

- `mitmproxy/net/http/multipart.py`: `decode_multipart()` now finds the end
  of each part's headers by byte position (`\r\n\r\n`, falling back to
  `\n\n`) instead of splitting the whole part into lines. Only the single
  line-break structurally separating the value from the next boundary is
  stripped from the end; everything else is preserved untouched.
  - Side effect, called out explicitly: a malformed part with no
    header/body separator is now skipped instead of raising an uncaught
    `ValueError` that previously aborted parsing of the *entire* multipart
    body.
- `test/mitmproxy/net/http/test_multipart.py`: regression tests for a
  newline inside a value, a genuine trailing newline in a value, and a
  malformed part being skipped without crashing the rest of the parse.

Verified against both hand-built multipart bodies and a full round-trip
through the real `encode_multipart()` (`test_http.py::test_set_multipart_form`),
confirming the fix doesn't just satisfy hand-written test bytes.

Fixes #4466


noqt (commented):
Disclosure: I work on [NOQT's Lumi Trace](https://github.com/noqt/Lumi-Trace). I ran Lumi Trace 0.10.1 against this exact head (`a6a6c716`); it ranked `decode_multipart()` first, and I then reproduced the parser behavior directly.

There's another value-preservation case this rewrite still drops. `content.split(b"--" + boundary)` treats the boundary token as a delimiter anywhere in a field value, but [RFC 2046 section 5.1.1](https://www.rfc-editor.org/rfc/rfc2046.html#section-5.1.1) requires a boundary delimiter to begin a line.

With one valid field whose value is `prefix--abc123suffix` and boundary `abc123`, this head returns:

```json
{"decoded": [["field", "prefix"]], "expected": [["field", "prefix--abc123suffix"]], "value_preserved": false}
```

The token is mid-line, so it is data rather than a delimiter; everything after it is currently lost. I'd scan for delimiter lines (start-of-body or immediately after a line break, with the boundary-line suffix validated) instead of splitting on the raw token, and add this mid-line case alongside the newline tests.

eliahu-frusin:
@noqt Thanks for catching this — confirmed and reproduced your exact case locally.

Fixed in the latest commit: `decode_multipart()` now requires a boundary
delimiter to begin a line (start of body, or immediately after a CRLF/LF),
per RFC 2046 §5.1.1, rather than splitting on the raw boundary token
wherever it appears. Your `prefix--abc123suffix` example now round-trips
correctly, and I added it as a regression test
(`test_decode_multipart_value_contains_mid_line_boundary_token`) along
with a related case for a `--` that doesn't match the boundary at all.

All existing tests still pass. Let me know if you spot anything else.

PR #8413, opened 2026-09-01
fix(web): reject unmatched quotes in filters

### LINKED ISSUE #8239: MitmWeb disconnects when writing filter with unmatched quote

RRXXZZYY:
#### Description

Fixes #8239.

MitmWeb's client-side PEG parser allowed quote characters inside an unquoted filter token, while the Python backend rejects them. As a result, an input such as `~u Content-Type'` was considered valid in the browser and sent over the WebSocket; the backend then rejected it and closed the live connection.

This change:

- excludes single and double quotes from unquoted filter values, matching the backend grammar;
- keeps properly quoted values valid;
- regenerates the checked-in parser;
- adds regression coverage showing that the invalid text remains in the input but is not propagated to the backend.

#### Validation

- `npm test -- --runInBand` — ESLint and TypeScript checks passed; 82 suites, 339 tests, and 116 snapshots passed.
- `uv run tox -e lint` — passed (`ruff check .`).
- Focused `FilterInputSpec.tsx` run — 14 tests passed.
- Regenerated `filt.js` was compared with fresh PEG.js 0.9.0 output; the generated bodies match exactly.

I did not run a separate manual browser session; the regression test exercises the exact input and verifies that no filter update is emitted.

#### AI assistance

OpenAI Codex assisted with the investigation and implementation. I reviewed the final diff, reproduced the failure before the fix, and ran the checks listed above.

#### Checklist


lups2000 (commented):
Hi, thanks for the PR but filt.js and filt.peg are being removed in #8331

RRXXZZYY:
Thanks for the heads-up. #8331 supersedes this change by removing the frontend parser and making backend validation authoritative, so this patch would no longer apply. Closing this PR in favor of #8331.

PR #8441, opened 2026-09-19
docs: fix relative link to version.py in the release README

1816586742-stack:
## What

Fix the relative link to `mitmproxy/version.py` in the release README.

## Why

The last step of the release checklist links to the version file like this:

```markdown
[...](mitmproxy/version.py) and add a `.dev` suffix.
```

Relative to `release/README.md`, that target resolves to `release/mitmproxy/version.py`, which does not exist — the link 404s. The file is at the repository root:

```
$ git ls-files | grep 'mitmproxy/version.py$'
mitmproxy/version.py
```

One character class of fix — the target now starts with `../` and resolves to the file that actually exists.

## Verification

- `mitmproxy/version.py` exists at the repository root
- the only other link in this file is an absolute `https://` URL, so nothing else in it is affected
- checked every relative link in `release/*.md`: this was the only broken one

I kept the visible link text unchanged so the diff stays to one line; if you would rather simplify the text to `version.py` at the same time, say the word and I will fold it in.


lups2000 (approved):
Thanks :)

